PR #72, opened 2015-01-12
Update launchd template

tommeier:
(using latest agent).
Fixed existing launchd template:
  * agent is now within a `bin/` directory
  * token is now `BUILDBOX_AGENT_TOKEN `

I've also added a 'private' launchd plist example (which is what I'm using).


templates/launchd.plist
```diff
@@ -11,6 +11,9 @@
     # Set the user to run buildbox-agent as (usually `whoami`). It should be a full OS X user created via System Prefs, and the one you installed buildbox-agent under.
     sudo sed -i '' "s/your-build-user/[insert username here]/g" /Library/LaunchDaemons/io.buildbox.buildbox-agent.plist
 
+    # Set the agent's name
+    sudo sed -i '' "s/your-agent-name/[insert agent name]/g" /Library/LaunchDaemons/io.buildbox.buildbox-agent.plist
+
```
kouky: @tommeier can you change the destination folder for the plist to `/Library/LaunchAgents` as discussed. Moving it to that location solved my fork exec errors on system reboot.

tommeier: Yup, people who don't care about a GUI would be fine in `LaunchDaemons` though, if they're running normal tests, perhaps we need a gui_launchd_template and non_gui_launchd_template.

tommeier: @kouky done :)

toolmantim:
Yoohoo!

PR #73, opened 2015-01-14
Updating bootstrap.sh to build the result of a Github PR

samw-bilue:
If `$BUILDBOX_PULL_REQUEST` is anything other than `false`, we'll now check out the result of that merge into its destination branch. Works for PRs inside the project repo, or from an external fork. Not tested for anything other than Github.

### DIFF
--- a/templates/bootstrap.sh
+++ b/templates/bootstrap.sh
@@ -92,13 +92,21 @@ fi
 buildbox-run "git clean -fdq"
 buildbox-run "git fetch -q"
 
-# Only reset to the branch if we're not on a tag
-if [ "$BUILDBOX_TAG" == "" ]
+if [ "$BUILDBOX_PULL_REQUEST" != "false" ]
 then
-buildbox-run "git reset --hard origin/$BUILDBOX_BRANCH"
+  echo "building result of pull request $BUILDBOX_PULL_REQUEST"
+  buildbox-run "git fetch origin +refs/pull/$BUILDBOX_PULL_REQUEST/merge:"
+  buildbox-run "git checkout -qf FETCH_HEAD"
+else
+  if [ "$BUILDBOX_TAG" == "" ]
+  then
+    # Only reset to the branch if we're not on a tag
+    buildbox-run "git reset --hard origin/$BUILDBOX_BRANCH"
+  fi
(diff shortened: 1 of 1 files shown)


toolmantim:
Hmm… so this will test against `refs/[pr]/merge` each time regardless of what the commit was that triggered the build. 

If the PR was updated in the mean time, the build would be for different commit (and the wrong commit would have their build commit status changed).

Hrmm…

samw-bilue:
Yeah fair point. I guess ideally if we detect an attempt to schedult a build for a PR which was already being build, I'd want the running build to be cancelled. Similar behaviour to the 'skip for multiple commits on the same branch'?

keithpitt:
Maybe I'm completley wrong, but can we do this:

```bash
buildbox-run "git fetch origin +refs/pull/$BUILDBOX_PULL_REQUEST/merge:"
buildbox-run "git reset --hard origin/+refs/pull/$BUILDBOX_PULL_REQUEST/merge:"
buildbox-run "git checkout -qf \"$BUILDBOX_COMMIT\""
```

Or something?

keithpitt:
Is there some docs about this somewhere?

samw-bilue:
Not that I've read. I'm just reconstructing this from how Travis does it. 

Re: the above comment:

    buildbox-run "git fetch origin +refs/pull/$BUILDBOX_PULL_REQUEST/merge:"

Gets the imaginary result of a merge from the PR into the destination branch, and sticks it in a temp branch called `FETCH_HEAD`.

    buildbox-run "git reset --hard origin/+refs/pull/$BUILDBOX_PULL_REQUEST/merge:"

Doesn't work :( you can do `git reset --hard FETCH_HEAD`, but it has the same effect as the current `"git checkout -qf FETCH_HEAD"` (except with the `git reset` we're merging the changes onto the local `develop` branch, as opposed to a separate detached branch you get when you `git checkout`).

    buildbox-run "git checkout -qf \"$BUILDBOX_COMMIT\""

Does work. But just rewinds you to the commit on the PR branch, not the result of that commit being merged to the destination.

keithpitt:
Hey @samw-bilue, thanks for figuring out all the things! I've moved your changes (tweaked a little) into https://github.com/buildbox/agent/pull/74. My version doesn't handle testing merge commits, but at least it's a step in the right direction! :smile:

PR #76, opened 2015-01-29
Option to automatically accept RSA fingerprints when git cloning.

keithpitt:
This change will allow the agent to automatically accept the RSA fingerprints when cloning from an ssh based git repository:

image

- [ ] Add support for `BUILDKITE_REPO_SSH_HOST` on buildkite.com
- [ ] Decide on the right argument name. Is `--no-ssh-keyscan` correct? Or is there another name? `--no-ssh-accept` or something?

### DIFF
--- a/buildkite/agent.go
+++ b/buildkite/agent.go
@@ -26,6 +26,9 @@ type Agent struct {
 	// The path to the run the builds in
 	BuildPath string
 
+	// Whether or not the agent is allowed to run an ssh keyscan
+	SSHKeyScan bool
+
 	// Run jobs in a PTY
 	RunInPty bool
 
--- a/buildkite/agent.go
+++ b/buildkite/agent.go
@@ -26,8 +26,9 @@ type Agent struct {
 	// The path to the run the builds in
 	BuildPath string
 
-	// Whether or not the agent is allowed to run an ssh keyscan
-	SSHKeyScan bool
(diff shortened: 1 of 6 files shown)


toolmantim:
`--no-ssh-keyscan` sounds right!

toolmantim:
I guess that'd go with `--scripts-only` for security-restricted setups, once we do arbitrary command execution…

keithpitt:
The only thing is, if we come up with a better way (instead of keyscan) then the option becomes wrong. `--no-ssh-rsa-accept` maybe?

toolmantim:
Yeah good point. What about `--no-ssh-key-detection` or `--disable-ssh-key-detection`

keithpitt:
I'm gonna run with `--no-ssh-fingerprint-verification` I think, it's a bit long, but it seems like the language that https://help.github.com/articles/generating-ssh-keys/ uses.

toolmantim:
hrmm… reading it now, it sounds like it disables SSH clone's fingerprint detection altogether. Perhaps `--no-automatic-ssh-fingerprint-verification` or `--disable-automatic-ssh-fingerprint-verification`

keithpitt:
`--no-automatic-ssh-fingerprint-verification` :D:D

toolmantim:
`--please-do-not-attempt-to-automatically-verify-an-ssh-hosts-keys-fingerprint-mmkay`

PR #77, opened 2015-01-31
Allow scripts sent from Buildkite to be evaluated

keithpitt:
This will allow you to store your scripts on Buildkite, and have the agent execute them (instead of having to commit them to your source repository).

Because this could be a potential security risk, this feature can be turned off with `--no-script-eval`


toolmantim:
Wooooo!

PR #80, opened 2015-02-12
Bootstrap Hooks

keithpitt:
Adds support for: `pre-checkout`, `pre-script` and `post-script`. You add these scripts to: `~/.buildkite/hooks` folder. It is also configurable using the `--hooks-path` argument, or the `BUILDKITE_HOOKS_PATH` environment variable.

image


toolmantim:
Woo! Now, for the hooks, some use cases:

* Do our own repository checkout, I don't want your git clone
* Run scripts my way, I want to tweak your Docker/fig setup

For this, perhaps we could have a `checkout` and `run-script`… if they exist, use them, if not go default.

What about:

* `pre-checkout`
* `checkout` (overrides standard behaviour)
* `post-checkout`
* `pre-run-script`
* `run-script` (overrides standard behaviour)
* `post-run-script`

And I guess if they want to cancel the build in `pre-checkout` and `pre-script` they just exit 1, and the `-e` will cause the whole thing to fail?

plasticine:
I am so hot for this.

abr

plasticine:
Also, just wanted to add that these would be awesome if they could be included on a per-project basis.

I can see `pre-run-script`, `run-script`, and `post-run-script` being handy. We (ferocia) already do something a little like this and have pre and post step hooks that live in `PROJECT_ROOT/.buildkite/hooks`

keithpitt:
Oh, as in if you commit a `.buildkite` folder to git, those hooks will automatically run?

keithpitt:
@toolmantim I've added `run-command` and `checkout` hooks. In agent 1.0 world, scripts have been renamed to `commands`

keithpitt:
@plasticine what do you use the hooks for?

plasticine:
> Oh, as in if you commit a .buildkite folder to git, those hooks will automatically run?

Yep.

> what do you use the hooks for?

Per-step cleanup stuff, logging step duration, etc.

toolmantim:
@plasticine can you paste some code examples of the logging step duration and cleanup stuff?

keithpitt:
Let's push this puppy into beta testing...

PR #81, opened 2015-02-16
Track timing of output headers

keithpitt:
- [x] Track when a marker is found
- [x] Store timing markers when the script outputs `---|+++|~~~`
- [x] Send markers to BK


joho:
use of `io.MultiWriter` looks legit to me.

PR #85, opened 2015-02-17
Continuously deploy the homebrew recipe

toolmantim:
This updates the https://github.com/buildkite/homebrew-buildkite/blob/master/buildkite-agent.rb when a Github release is made. Not 100% sure on how the Github versioning will work with homebrew…


keithpitt:
Let's give it a go! :+1:

PR #92, opened 2015-02-22
Default value for BUILDKITE_COMMIT

joelmoss:
bootstrap failed if no `BUILDKITE_COMMIT` was given, as most of the time, you would only specify the branch name, this PR set the `BUILDKITE_COMMIT` to be the `BUILDKITE_BRANCH` by default.

And unless I'm missing something, line 197 seems to be redundant in the case of `BUILDKITE_COMMIT` being the same as `BUILDKITE_BRANCH`.


toolmantim:
Thanks for the PR! But if you don't pin a pipeline to a specific commit on `master`, then if `master` changed during the build pipeline then one build step will be running against one commit, and another step against another commit. We'll probably try and solve this a different way.

PR #98, opened 2015-02-27
Correct encode backslashes in URLs

toolmantim:
When using custom S3 creds an invalid URL can be generated if filenames contain invalid URL characters (e.g. `\` or ` `). This ensures they're encoded correctly.


keithpitt:
Seems legit!

keithpitt:
(btw, is this still valid @toolmantim?)

toolmantim:
Yep, we do this on the Ruby side now too.

But I'm thinking for Windows we should gsub '\' for '/' in the upload path, and '/' for '\' in the search and download path. It's a bit tricky, but that'll mean on Windows you're using the native paths '\' and it "just works", but internally we're storing them all with Unix path style.

If you were to mix Windows and *nix it's look like:

Win: artifact upload dir1\*
*nix: artifact download dir1/*

At the moment (and with this patch) and least it works, but it's:

Win: artifact upload dir1\file.txt
*nix: artifact download dir1\* (and the filename that gets created is called "dir1\file.txt" - the dir is part of the filename)

toolmantim:
Doh, Github ate some slashes and stars there.

toolmantim:
Fixed the slashes and \*'s

PR #106, opened 2015-03-11
Build does not report failure on Windows

samritchie:
Sorry, there are a couple of changes in here:

1) Added some quotes around %BUILDKITE_BUILD_DIR% in case BuildKite is installed in Program Files. Who would put it there though, right? :)
2) Converted line endings to CRLF (sorry). I often only have notepad available on a server & this is a bit more user-friendly. I always run autocrlf false and trust contributors to do the right thing on a per-file basis, but if you guys use a different policy I can redo the PR.
3) %ERRORLEVEL% does not work the way you’d expect inside IF blocks (it’s expanded prior to the block being executed). This requires `SETLOCAL ENABLEDELAYEDEXPANSION` and the use of `!ERRORLEVEL!` inside the blocks. See http://stackoverflow.com/a/4368104/47636. Yes, we all learned something today.

This bootstrap is running okay on my build server now, although no doubt something else will crop up.

### DIFF
--- a/templates/bootstrap.bat
+++ b/templates/bootstrap.bat
@@ -1,5 +1,8 @@
 @echo off
 
+REM enable delayed expansion so that ERRORLEVEL is evaluated properly inside IF blocks
+SETLOCAL ENABLEDELAYEDEXPANSION
+
 REM echo --- Environment Variables
 REM SET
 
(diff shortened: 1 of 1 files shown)


keithpitt:
Thanks @samritchie! I'll review shortly and let you know - but from just a quick scan, looks good :) I'm not too fussed about the line endings, just makes it hard to see what's actually changed in the diff.

samritchie:
It would have made much more sense to do two separate commits. I’m older and wiser now.

PR #128, opened 2015-04-07
Stream logs to BK in chunks

keithpitt:
This is a work in progress of the new log streaming gear. Buildkite has no API to support this yet, so I'm just working on the agent implementation first.

- [x] Stop uploading chunks when the process wants to exit (this requires us sending a cancel channel I think to the uploader)
- [x] Copy more from `client.go` to `http/request.go`
- [x] Fix path concatenation in `request.go`
- [x] Tweak default connection timeouts


buildkite/job.go
```diff
@@ -121,29 +120,35 @@ func (j *Job) Run(agent *Agent) error {
...
-	j.StartedAt = time.Now().UTC().Format(time.RFC3339Nano)
-	_, err := agent.Client.JobUpdate(j)
-	if err != nil {
-		// We don't care if the HTTP request failed here. We hope that it
-		// starts working during the actual build.
-	}
+	// The HTTP request we'll be sending it to
+	logStreamerRequest := agent.Client.GetSession().NewRequest("POST", "jobs/"+j.ID+"/log")
+
+	// Chunks will keep on retrying until they upload
```
keithpitt: Not sure what logic to choose here. I can say: "the chunk will forever keep trying" or "if it doesn't upload within 5 minutes of retrying, give up".

keithpitt: I can detect the later in the BK API and we can present an error saying that the not all the logs are visible due to uploading problems. This is very edge case though.

toolmantim: I don't have too much to add, apart from "what happens if we have downtime"

keithpitt:
Most of the code for this is now done - @wolfeidau mind having a quick squiz? :)

keithpitt:
So essentially every second we grab the build output, compare it to the previous time we checked it (so grab the difference) - chop that difference into 100kb parts, and ship that to BK.

wolfeidau:
So I have a few questions.

1. What is the reason for the custom HTTP client?
2. Can you separate this out into another project and provide some interface between it and this buildbox agent?

The reasons I say that are:

1. This is a useful thing 
2. it is going to be something you iterate on i think as it is hard
3. The agent doesn't give a shit about where the logs go, it just wants the old "hey chuck this over the wall bro" thing.

And just think your first library Keith!

keithpitt:
Hey @wolfeidau - it's less about a custom HTTP client, and I wanted to move the gear inside the `FormUploader` class and `Client` class into something more HTTP'ish. I want to put gzip compression in there as well at some point. So the code isn't anything new to the agent, just abstracted away into a folder really. Also there's a common pattern of stuff like this:

https://github.com/buildkite/agent/blob/master/buildkite/job.go#L193

Where some requests must succeed, and abstracting that away is something I want to do. Also making it easier to set timeouts - I still need to increase the TLS handshake timeout - as it's very small and on crappy internet connections (like the one I'm on at the moment in the EU) - it happens alot.

The idea I was going for with the log streamer was: "hey, here's the output, and here's a request template, make that request every time you have a chunk to send"

Do you mean separating the HTTP gear into a new library? If so, that's was kind of my plan. I've looked at some of the ones out there, and they don't do all the things I'm after. And using the stdlib is way to verbose.

keithpitt:
I also made the `logstreamer` a new package because I'm just trying to write more abstracted things, and less specific BK agent stuff.

wolfeidau:
Yeah you need to ensure internally that each of the pieces have clear boundaries, that will make enhancements later easier :) As long as all this stuff is behind a Client, preferably an interface, then it should be OK.

Normally you will just write a new transport to tweak timeouts https://github.com/golang/go/blob/master/src/net/http/transport.go#L33-L40 you can see them there. The existing http client is very extensible, just from the inside, rather than the outside :)

So why separate this out? Well I would like to see what happens when you pipe a few hundred MB of logs into one side, while the other side is trying send data out a congested link. Ideally this would result in the writer being blocked while stuff in transit is pushed, otherwise your agent is going to use a lot of ram.

Given the current setup, I need to test the entire agent, rather than just the uploader.

Hopefully this helps you understand my previous points :)

buildkite/job.go
```diff
@@ -176,19 +223,41 @@ func (j *Job) Run(agent *Agent) error {
...
+	j.uploadHeaderTimes(agent)
+
+	// Stop the log streamer
+	logStreamer.Stop()
+
+	// Save how many chunks failed to upload
+	j.ChunksFailedCount = int(logStreamer.ChunksFailedCount)
+
+	if j.ChunksFailedCount > 0 {
+		logger.Warn("%d chunks failed to upload for this job", j.ChunksFailedCount)
```
toolmantim: I wonder if we should store the log to disk at this point, so they've got a copy of the log? `/var/log/buildkite-agent/failed-logs/{account.slug}-{proj.slug}-{build.number}-{job.uuid}.log`

keithpitt: Yeah - it's probably a good idea. The storage location is tricky (windows would be different) and there's log rotation to worry about. Can you open an issue and I can take a look at doing this in another PR?

PR #130, opened 2015-04-10
Catch Docker failures

toolmantim:
Now that we pipe to `cat` (as of 1c4241c3c9153dda168748c0821d927773aaf73b) any failure exit status codes in docker and docker-compose are masked. This change sets `pipefail` so non-zero exit statuses are bubbled through the command pipeline.


keithpitt:
Ah yes, great catch! :+1:

PR #132, opened 2015-04-10
Docker hooks - pre-docker-build & pre-docker-run

deanperry:
The `pre-command` hook is handy but for what I want to do (check if MySQL is running), it needs to be ran before the docker run command

(I'm on Windows at the mo so I haven't tested it as it's just a quick change using the GH site)


deanperry:
Actually, thinking about it, the script would more than likely need to be run actually in the docker container?

PR #135, opened 2015-04-17
Remove all the pre-1.0 Readme content

toolmantim:
This removes all the pre-1.0 readme content, bringing the focus of the docs to "how to contribute". The official user/usage docs will live at [buildkite.com/docs](https://buildkite.com/docs)


keithpitt:
:shipit:

Issue #126, opened 2015-04-02
Move homebrew to using the new config file support


toolmantim:

keithpitt:
It does now!! Booyah!

Issue #125, opened 2015-04-02
Move apt package's buildkite-agent.env to the new config file format


toolmantim:

keithpitt:
This is done too :) :boar:

PR #136, opened 2015-04-18
Switch upstart to use the new cfg file

toolmantim:
This is the first stab at changing our apt package to use the new config file. This way the brew and windows packages can use the same base cfg file, and we have a nice doc online explaining each of the properties… and that way we can refer to that doc in any guides, docs or elsewhere, no matter the install method.

Not yet run or tested…

Also #125 and #126


keithpitt:
I've pulled these changes into master :family:

PR #138, opened 2015-04-21
Use Buildkite beta agent for plist templates

tommeier:
Using this for my own needs (so i can autoupdate the beta agent via boxen), close if you don't need/want it :)


keithpitt:
@tommeier we also support a new config file using `--config` if you want to store stuff in there (and not have to edit the plist ENV vars)

keithpitt:
Example here: https://github.com/buildkite/agent/blob/switch-to-cfg-file/templates/buildkite-agent.cfg

PR #142, opened 2015-05-04
Add a docker cleanup script to the linux packages

toolmantim:
This adds a cleanup script for Docker which people can use to keep their builds snappy and disk space under control


keithpitt:
:+1:

DylanFM:
:heartbeat:

PR #143, opened 2015-05-04
Fix template launchd.plist

kouky:
Just setup a mac OS X machine with a new agent and had to make these changes to the template plist to get it to work.

- Fix incorrect raw github address for template `launchd.plist`
- Fix missing program arguments
- Fix incorrect env var `BUILDKITE_AGENT_ACCESS_TOKEN`
- Rename destination plist to `com.buildkite.buildkite-agent.plist`

### DIFF
--- a/templates/launchd.plist
+++ b/templates/launchd.plist
@@ -52,6 +52,8 @@
     <array>
       <string>/Users/your-build-user/.buildkite/bin/buildkite-agent</string>
       <string>start</string>
+      <string>--token</string>
+      <string>BUILDKITE_AGENT_ACCESS_TOKEN</string>
     </array>
 
     <key>KeepAlive</key>
--- a/templates/launchd.plist
+++ b/templates/launchd.plist
@@ -6,23 +6,23 @@
   To install:
 
     # Download the launchd config to /Library/LaunchAgents/ (it needs to be owned by root:wheel)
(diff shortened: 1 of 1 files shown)


templates/launchd.plist
```diff
@@ -73,17 +75,17 @@
...
     <key>StandardErrorPath</key>
-    <string>/usr/local/var/log/buildkite-agent.log</string>
+    <string>/Users/your-build-user/.buildkite/log/buildkite-agent.log</string>
 
     <key>EnvironmentVariables</key>
     <dict>
       <key>PATH</key>
       <string>/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/bin</string>
 
-      <key>BUILDKITE_AGENT_TOKEN</key>
```
keithpitt: I think it's supposed to be `BUILDKITE_AGENT_TOKEN` because that's the registration token - I think?

kouky: @keithpitt is it meant to match the env var from this [list in the docs](https://buildkite.com/docs/guides/environment-variables)?

keithpitt: I actually think you can get rid of this entire ENV all together :) You don't need it since you've got https://github.com/buildkite/agent/pull/143/files#diff-d2f5f813d0c1aed057afb6b8b014be21R56

keithpitt:
Thanks @kouky!! <3<3<3

tommeier:
:heart:

PR #145, opened 2015-05-06
honour --build-dir and chdir across filesystems

silarsis:
The BUILDKITE_BUILD_PATH environment variable wasn't being honoured - this assumes that one will be provided and works with it. This also adds a "/D' to cd to make sure we change directory across filesystems, in case you have a data drive mounted you want to build in.


keithpitt:
Looks great, thanks for the PR! :heart_decoration:

PR #149, opened 2015-05-14
Expand the command output by default

toolmantim:


toolmantim:
@keithpitt what you reckon?

keithpitt:
The problem was with the log renderer itself. Just pushed a fix so this should all be good now! :heart:

PR #155, opened 2015-05-21
Don't set defaults in the cfg that don't need to be set

toolmantim:
This moves the bootstrap config line down to the bottom, because it's not important and we don't want people changing it. This also comments out the defaults that are picked up by the agent anyways, which prevents me from commenting them out using bash voodoo in the homebrew (which should "just work" without any custom bootstrap path set in the config file)


toolmantim:
This doesn't make sense unless the binaries showed the default in `--help` etc. Closing for now!

PR #164, opened 2015-06-10
Add health check endpoint

wolfeidau:
This is the health check we discussed.

I have the basic health check working, I need to add a flag, rejig the structure and add some tests.

The aim is to expose the last job, and it's status and the last api request and it's status.

Ideally you would add some other information as requested in the future.

Looks like: 

```json
{
  "lastApiCall": {
    "status": "POST https://agent.buildkite.com/v2/register got response with status 200 OK",
    "timeTaken": "1.421656691s"
  }
}
```

### DIFF
--- a/agent/agent_pool.go
+++ b/agent/agent_pool.go
@@ -2,12 +2,14 @@ package agent
 
 import (
 	"fmt"
+	"os"
+	"time"
+
 	"github.com/buildkite/agent/api"
+	"github.com/buildkite/agent/health"
 	"github.com/buildkite/agent/logger"
 	"github.com/buildkite/agent/retry"
 	"github.com/buildkite/agent/signalwatcher"
-	"os"
-	"time"
 )
 
 type AgentPool struct {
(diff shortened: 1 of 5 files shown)


keithpitt:
Awesome! With the port thing, how does that work when you have multiple agents installed on the same machine?

wolfeidau:
To start with it may be better to leave this off by default.

My plan is to add a switch to enable it and change the bind address, this lets people with many agents bind them on different ports if required, or just do nothing and it won't bind any of them.

Switch is an empty string by default, and can be overridden at the command line, so `:9090` as an example.

As you would have to configure something to monitor the state of the health check it doesn't seem to much to request the user also enable it.

keithpitt:
I'm going to close this until we can both find some time to take another crack at it @wolfeidau! :dancers:

keithpitt:
Thanks for the work you've done on it so far! :heart:

wolfeidau:
Yeah agreed maybe when you come back to Australia!?

keithpitt:
Sounds good!

PR #165, opened 2015-06-15
Added defer'd flush of the body to ensure connection reuse.

wolfeidau:
Pretty small change to ensure the Body is ALWAYS flushed, it seems the JSON decoder is only reading enough to satisfy the decoder and not triggering the final EOF on the stream.

From the golang http library documentation.

```
	// The default HTTP client's Transport does not
	// attempt to reuse HTTP/1.0 or HTTP/1.1 TCP connections
	// ("keep-alive") unless the Body is read to completion and is
	// closed. 
```

So this change makes sure that happens.


keithpitt:
Awesome, thanks!! :dancers:

Issue #52, opened 2014-09-23
bootstrap.sh should check out submodules by default


MSch:
basically

```
buildbox-run "git checkout -qf \"$BUILDBOX_COMMIT\""
buildbox-run "git submodule update --init"
```

keithpitt:
@ahawkins are you doing a `git submodule update --init` in your bootstrap.sh file?

keithpitt:
I've added it to our `bootstrap.sh` for the upcoming 1.0 release of the agents: https://github.com/buildkite/agent/blob/master/templates/bootstrap.sh#L107

keithpitt:
If you'd like similar behaviour now, just copy those 3 lines into your current `bootstrap.sh`

Thanks everyone for your help!! :dancers:

blueimp:
Sorry to comment on a closed issue, but would it be possible to disable the git submodule calls with a buildkite configuration setting or an environment variable?

This would be useful for minimal agents without perl installed (git submodules require perl).

Tasks on an agent without perl don't break, but they still display error messages like this:
```sh
/usr/libexec/git-core/git-submodule: line 1: /usr/bin/perl: not found
```

Currently I'm disabling git submodule calls with the following command, but that's rather brittle:
```sh
sed -i 's/\(git submodule\)/true \1/g' /buildkite/bootstrap.sh
```

toolmantim:
@blueimp can you give the bootstrap from #168 a go? I've added a `BUILDKITE_DISABLE_GIT_SUBMODULES` environment variable for disabling the git submodule checkout commands.

blueimp:
Thanks, I've added my comment to #168

PR #168, opened 2015-07-04
Add ability to disable git submodules

toolmantim:
Allows using the agent in environments that don't have Perl, which the git submodule commands rely upon.

Related to #52

### DIFF
--- a/templates/bootstrap.sh
+++ b/templates/bootstrap.sh
@@ -215,7 +215,10 @@ else
   fi
 
   buildkite-run "git clean -fdq"
-  buildkite-run "git submodule foreach --recursive git clean -fdq"
+
+  if [[ -z "${BUILDKITE_DISABLE_GIT_SUBMODULES:-}" ]]; then
+    buildkite-run "git submodule foreach --recursive git clean -fdq"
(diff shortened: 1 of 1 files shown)


blueimp:
Thanks for the quick implementation, @toolmantim!

If you change the two lines with
```sh
if [[ ! -z "${BUILDKITE_DISABLE_GIT_SUBMODULES:-}" ]]; then
```
to 
```sh
if [[ -z "${BUILDKITE_DISABLE_GIT_SUBMODULES:-}" ]]; then
```
it will work fine. :)

From the naming of the variable, I guess you wanna check if `BUILDKITE_DISABLE_GIT_SUBMODULES` is **empty**, rather than **not empty** before executing the `git submodule` calls.

toolmantim:
Whoops, fixed. Give that a go and happy to merge into master if it works :+1:

blueimp:
Tested and works! :+1:

toolmantim:
@blueimp awesome! Thanks for testing. This will go out in the next agent release, which should be within the next week.

blueimp:
Thanks! :)

PR #174, opened 2015-08-07
Adds prune flag to git fetch

EmilS:
Adds prune flag to `git fetch`. On active repos we receive warnings from `git fetch` to run `git prune`.

```
$ git fetch -q
Auto packing the repository for optimum performance. You may also
run "git gc" manually. See "git help gc" for more information.
Nothing new to pack.
Checking connectivity: 1429273, done.
warning: There are too many unreachable loose objects; run 'git prune' to remove them.
``` 

@keithpitt @xthexder

### DIFF
--- a/templates/bootstrap.sh
+++ b/templates/bootstrap.sh
@@ -220,7 +220,7 @@ else
     buildkite-run "git submodule foreach --recursive git clean -fdq"
   fi
 
-  buildkite-run "git fetch -q"
+  buildkite-run "git fetch -qp"
 
   # Allow checkouts of forked pull requests on GitHub only. See:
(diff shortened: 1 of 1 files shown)


xthexder:
:+1:

xthexder:
Hm... didn't realize this was in the actual buildkite agent.
This is definitely good for our use case, but I can't really speak for everyone else.

@EmilS I think we have our own bootstrap / pre-hook script that we can change instead if this isn't something that goes upstream.

EmilS:
The more appropriate location for this is in the global pre-checkout hook which is what we ended up doing.

PR #175, opened 2015-08-16
Fetch only the required commit

byroot:
Problem:

On very large repositories with many contributors, branches tend to accumulate and `git fetch` have a lot of objects to download making it very slow.

And then the local cache tend to accumulate a lot of loose objects and we need to prune it.

Solution:

We can fetch only the commit we need (and the missing ancestors) with `git fetch origin SHA1`.

### DIFF
--- a/templates/bootstrap.sh
+++ b/templates/bootstrap.sh
@@ -220,17 +220,18 @@ else
     buildkite-run "git submodule foreach --recursive git clean -fdq"
   fi
 
-  buildkite-run "git fetch -q"
-
   # Allow checkouts of forked pull requests on GitHub only. See:
   # https://help.github.com/articles/checking-out-pull-requests-locally/#modifying-an-inactive-pull-request-locally
   if [[ "$BUILDKITE_PULL_REQUEST" != "false" ]] && [[ "$BUILDKITE_PROJECT_PROVIDER" == *"github"* ]]; then
     buildkite-run "git fetch origin \"+refs/pull/$BUILDKITE_PULL_REQUEST/head:\""
-  elif [[ "$BUILDKITE_TAG" == "" ]]; then
-    # Default empty branch names
-    : "${BUILDKITE_BRANCH:=master}"
+  else
+    buildkite-run "git fetch -q origin $BUILDKITE_COMMIT 2> /dev/null || git fetch -q"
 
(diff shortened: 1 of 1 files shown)


toolmantim:
This sounds like a very sensible default. And if you needed access to more branches you could always add another `git fetch` in `after_checkout` hook…

keithpitt:
I like this idea! I just did some testing locally and it seems that it works for most cases `git fetch origin my-branch`, but doesn't seem to work with partial SHA1's. Some services only send us the first 8 characters of the SHA in their webhooks >,<. So I tested with a partial SHA1, and git fetch doesn't seem to like it:

```
$ git fetch ed3d8916082a4ab00d76ef062dbea22b8755bc94
From github.com:buildkite/buildkite
 * branch            ed3d8916082a4ab00d76ef062dbea22b8755bc94 -> FETCH_HEAD

$ git fetch ed3d89160
fatal: Couldn't find remote ref ed3d89160
```

I'd be happy to merge this PR in if you handle the case where `git fetch` fails, and then tries to do a a full git fetch (so we don't break backwards compat). If you check the bootstrap around https://github.com/buildkite/agent/blob/master/templates/bootstrap.sh#L243 you'll see an example of how to run a command and grab the failure.

templates/bootstrap.sh
```diff
@@ -220,17 +220,18 @@ else
...
-
   # Allow checkouts of forked pull requests on GitHub only. See:
   # https://help.github.com/articles/checking-out-pull-requests-locally/#modifying-an-inactive-pull-request-locally
   if [[ "$BUILDKITE_PULL_REQUEST" != "false" ]] && [[ "$BUILDKITE_PROJECT_PROVIDER" == *"github"* ]]; then
     buildkite-run "git fetch origin \"+refs/pull/$BUILDKITE_PULL_REQUEST/head:\""
-  elif [[ "$BUILDKITE_TAG" == "" ]]; then
-    # Default empty branch names
-    : "${BUILDKITE_BRANCH:=master}"
+  else
+    buildkite-run "git fetch -q origin $BUILDKITE_COMMIT 2> /dev/null || git fetch -q"
```
byroot: @keithpitt I believe that should do it ^

keithpitt:
:dango:

keithpitt:
Thanks!! It'll be part of the next agent release!

PR #182, opened 2015-09-09
Handle repository URL updates

### LINKED ISSUE #79: Changing project repository URL breaks bootstrap.sh
If you change the project URL (say, you accidentally paste the https instead of ssl version) and run a build, the bootstrap.sh doesn't notice the change and doesn't try to do any update, or fresh checkout.

One fix would be:

```bash
if [[ ! -d ".git" ]]; then
  buildkite-run "git clone \"$BUILDKITE_REPO\" . -qv"
else
  buildkite-run "git remote set-url origin \"$BUILDKITE_REPO\""
fi
```

alexjurkiewicz:
Fixes #79

### DIFF
--- a/templates/bootstrap.sh
+++ b/templates/bootstrap.sh
@@ -226,7 +226,9 @@ else
   export GIT_TERMINAL_PROMPT=0
 
   # Do we need to do a git checkout?
-  if [[ ! -d ".git" ]]; then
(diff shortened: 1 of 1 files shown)


keithpitt:
Just tested this out locally. Works like a treat, thanks!

Issue #108, opened 2015-03-18
Artifact download syntax


toolmantim:
After reading `buildkite-agent artifact --help` I just tried to do a `buildkite-agent artifact download pkg/buildkite-agent-darwin-386.tar.gz pkg/buildkite-agent-darwin-386.tar.gz` in the homebrew release script, expecting it to download the file, and received:

```
2015-03-18 03:53:23 FATAL  Could not find information about destination: /go/src/github.com/buildkite/agent/pkg/buildkite-agent-darwin-386.tar.gz
```

I wonder if we should support this syntax, and/or provider a better error message, and/or improve the help.

```
Example:

   $ buildkite-agent artifact download "pkg/*.tar.gz" . --build xxx

   This will search across all the artifacts for the build with files that match that part.
   The first argument is the search query, and the second argument is the download destination.
```

curl works like this (the reverse):

```
$ curl "https://www.google.com.au/?gfe_rd=cr&amp;ei=bfoIVf7zLKSN8Qf_ioCIDw" -o /tmp
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0Warning: Failed to create the file /tmp: Is a directory
curl: (23) Failed writing body (0 != 515)
$ curl "https://www.google.com.au/?gfe_rd=cr&amp;ei=bfoIVf7zLKSN8Qf_ioCIDw" -o /tmp/google.com
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 17860    0 17860    0     0  24064      0 --:--:-- --:--:-- --:--:-- 24037
```

perhaps we should support both (as in you can specify the full file path, or just the directory), like `cp` does?

keithpitt:
This is related: https://github.com/buildkite/agent/issues/91

lox:
Closing in favour of #91.

Issue #96, opened 2015-02-24
Artifact path normalization


underscorediscovery:
Paths from Windows agent (1.0-beta.12.317) display in the dash board as `\` while other agents (mac/linux) display as `/`. The real problem is trying to download those from an agent - an artifact path to match `files/*` fails to find Windows built artifacts that would match the glob/path. I think all paths should be normalized to / as to retain parity/consistency at all times. 

Reproduce :
- add build step with artifact path in pipeline such as `files/*`
- observe artifacts from a Windows agent are uploaded just fine
- add second build step (with wait for previous steps)
- try download the artifacts from the step using `buildkite-agent build-artifact download "files/*"`
- result:
  - Windows agent artifacts not found
  - Other agents artifacts found

Also tested:
- Tried `files\*` and `files\\*`, no luck.

underscorediscovery:
Newer fixes by @toolmantim have solved the downloads being found by the tool. \o/ 

However, while on a unix machine downloading the artifacts from Windows agents, it results in files only found by `\`. This is ok, but the resulting disk-file being named as `path\to\item.file` literally, as the filename.

samritchie:
+1. I’ve observed the following additional behaviour:

For uploads:

* `files/**/*` only matches one level deep, i.e. `files/foo/bar.txt`, but not `files/foo/bar/baz.txt`
* `files\**\*` matches the entire tree

For downloads:

* `files\**\*` seems to match all artifacts
* `files\foo\*` doesn’t match anything

Looks like I’m currently running 2.0.3.761.

keithpitt:
We've done some testing with Artifacts and the latest v3 release of the agent, and everything seems to be working A-OK! We're gonna close this for now. If you see any further issues, hit us up!

Issue #91, opened 2015-02-19
Artifact path downloads


keithpitt:
Double check the behaviour of the artifact downloading gear:

![image](https://cloud.githubusercontent.com/assets/25882/6260291/8f367104-b817-11e4-83c2-736c2bd8807f.png)

And make sure it's consistent with the rest of the unix world with how tools like `cp` work

keithpitt:
Also it may be worth adding options like `-R` for recursive copies

lox:
Y'know, on looking at this, I actually don't think I agree:

It makes sense to me that `buildkite-agent artifact upload foo/bar/bang.txt` and then `buildkite-agent artifact download y/` would result in `y/foo/bar/bang.txt`.

I'd be more inclined to look to rsync (http://man7.org/linux/man-pages/man1/rsync.1.html) for conventions than I would `cp` and `mv`.

keithduncan:
I’ve been running in to a version of this in my pipelines 😅

Right now there’s a 1:1 correlation between the file system at upload time / the glob given to the upload command and the "artifact key space". I often rearrange the file system (or cd) so that I can `buildkite-agent artifact upload` without a prefix, but that doesn’t work when using the `artifact_paths` step property which I’d prefer to use over adding an entry to my `commands` array. Personally I like to drop the path prefix to my artifacts between steps to decouple the artifact name from how the repository is laid out today, so I can change the layout without having to think about my pipelines, this might just be me though 😄 

What would you think about a destination syntax or a new option for where in the artifact key space things should be placed when uploading?

PR #186, opened 2015-09-20
New artifact upload/download syntax

keithpitt:
Currently the syntax for uploading and downloading files is inconsistent and confusing. This PR switches us to a model that replicates `cp` and `mv`

- [x] Add option
- [ ] Rewrite file collection
- [ ] Rewrite file downloading
- [ ] Ensure download logic is contained within a seperate module so it can be used by the S3 downloader

This change will address: https://github.com/buildkite/agent/issues/91 https://github.com/buildkite/agent/issues/96 and https://github.com/buildkite/agent/issues/108

### DIFF
--- a/agent/agent_configuration.go
+++ b/agent/agent_configuration.go
@@ -7,4 +7,5 @@ type AgentConfiguration struct {
 	AutoSSHFingerprintVerification bool
 	CommandEval                    bool
 	RunInPty                       bool
+	ArtifactUnixSyntax             bool
 }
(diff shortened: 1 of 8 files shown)


keithpitt:
This PR is also pretty dead now. Closing and will try again in another agent version

lox:
I wonder if we should address this in a future version @keithpitt? At the very least, we should document the difference in semantics somewhere.

keithpitt:
@lox yeah, we still gotta fix it. Probably look at adding it next year when I've got a bit more time. Artifact download/upload semantics always confuse me and I wrote them!

PR #187, opened 2015-09-22
Use a lock when reading or writing to artifactsStates

grosskur:
I see the agent hanging occasionally after all artifacts have been uploaded. It appears the background goroutine that uploads state changes using the Buildkite API is waiting for more state changes.

Looking into this, I noticed that the `artifactsStates` map is accessed and modified in different goroutines without first acquiring a lock, which appears to be unsafe. See:

https://golang.org/doc/faq#atomic_maps

Adding `p.Lock()` and `p.Unlock()` fixes the hangs for me.

Note: This uses the pool's `sync.Mutex`. I think we only need a `sync.RWMutex`, which allows multiple parallel readers---but that would require a bigger change.

### DIFF
--- a/agent/artifact_uploader.go
+++ b/agent/artifact_uploader.go
@@ -192,7 +192,9 @@ func (a *ArtifactUploader) upload(artifacts []*api.Artifact) error {
 			// them from the tracking map
 			for id, state := range artifactsStates {
 				statesToUpload[id] = state
+				p.Lock()
 				delete(artifactsStates, id)
+				p.Unlock()
 			}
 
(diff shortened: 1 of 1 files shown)


toolmantim:
Nice catch, and thanks for the contribution! @keithpitt will have to review this when he's back from holidays in a week or so.

keithpitt:
Hey @grosskur, great catch! Yeah, I think what you found would definitely cause that sort of behaviour. Thanks for spending the time debugging and making the fix! I'll merge this now, and include it in the next beta release of 2.1 :dancers:

PR #194, opened 2015-10-09
Use git clone --single-branch

byroot:
This avoid pulling all the branch during the clone.

It does require Git `1.7.10` thought: https://github.com/Shopify/shipit-engine/blob/45d0225276715b6352159b043c3bb4aa07845c2a/lib/stack_commands.rb#L50-L53

@keithpitt @toolmantim for review please.

### DIFF
--- a/templates/bootstrap.sh
+++ b/templates/bootstrap.sh
@@ -243,7 +243,11 @@ else
   if [[ -d ".git" ]]; then
     buildkite-run "git remote set-url origin \"$BUILDKITE_REPO\""
   else
-    buildkite-run "git clone \"$BUILDKITE_REPO\" . -qv"
+    if git clone --help | grep -- --single-branch &> /dev/null; then
+      buildkite-run "git clone -qv --single-branch \"$BUILDKITE_BRANCH\" -- \"$BUILDKITE_REPO\" ."
+    else
+      buildkite-run "git clone -qv -- \"$BUILDKITE_REPO\" ."
+    fi
   fi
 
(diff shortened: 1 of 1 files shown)


byroot:
So I checked why we had to check for the git version. And it's because Ubuntu 12.04 ship with git `1.7.9`, so close... http://packages.ubuntu.com/precise/git

If it's a requirement to support it, I think I can handle it. Let me know.

keithpitt:
So the first clone will just get the branch it needs for the build, and subsequent builds will just fetch only commits/branches it needs?

keithpitt:
And I think we need to gracefully handle older git versions unfortunately - that could be tricky. I went down the rabbit hole a while ago trying to have some smarts around git version and doing stuff conditionally on that: https://github.com/buildkite/agent/blob/master/templates/bootstrap.sh - but I couldn't figure it out.

You could maybe: `git clone --help | grep single-branch` something something to do it?

byroot:
> So the first clone will just get the branch it needs for the build, and subsequent builds will just fetch only commits/branches it needs?

Yep. It's basically just the continuation of the previous PRs.

> And I think we need to gracefully handle older git versions unfortunately

No problem. I'll handle it.

keithpitt:
:sunglasses:

byroot:
@keithpitt I tested on 12.04 `git clone --help | grep -- --single-branch &> /dev/null` does indeed work.

Does it look good to you now?

keithpitt:
Yeah yeah!!! Thanks! :+1:

PR #195, opened 2015-10-13
add arm support for install script

mirdhyn:
Currently the install script fallback to 386 unless the arch is x86_64.
As arm binaries are officially released, a simple case statement in the install script fix the issue.


keithpitt:
Looks good to me @mirdhyn, thanks! :+1:

PR #198, opened 2015-10-15
Convert `bootstrap.sh` to Golang

keithpitt:
- [x] Finalize Windows support


agent/bootstrap.go
```diff
@@ -0,0 +1,845 @@
...
+}
+
+// Shows a buildkite boostrap warning
+func warningf(format string, v ...interface{}) {
+	printf("\033[33m⚠️ Buildkite Warning: %s\033[0m", fmt.Sprintf(format, v...))
+	printf("^^^ +++")
+}
+
+// Shows the error text and exits the bootstrap
+func fatalf(format string, v ...interface{}) {
```
wolfeidau: I would call this `exitf` just to because it doesn't show a trace like `fatalf` (which uses panic("message"))

shell/const.go
```diff
@@ -0,0 +1,8 @@
+// +build !windows
+
+package shell
+
+const (
+	ENVIRONMENT_KEY_PATH      = "PATH"
```
wolfeidau: These should be `EnvironmentKeyPath` if your following golang style. Snakes don't live here! :snake: :no_entry_sign:

agent/bootstrap.go
```diff
@@ -0,0 +1,845 @@
...
+
+		absolutePathToHook, err := filepath.Abs(hookPath)
+		if err != nil {
+			fatalf("Failed to find absolute path to \"%s\" (%s)", hookPath, err)
+		}
+
+		// Create the hook runner code
+		var hookScript string
+		if runtime.GOOS == "windows" {
+			hookScript = "@echo off\n" +
```
wolfeidau: You should do this using a golang template as illustrated below as it will save the mess of escaping.

see https://github.com/wolfeidau/aws-ek-setup/blob/master/cmd/aws-ek-setup/command_create.go#L27-L39 and executed https://github.com/wolfeidau/aws-ek-setup/blob/master/cmd/aws-ek-setup/command_create.go#L75-L92

toolmantim:
Wow nice one!

wolfeidau:
:+1: great work!

keithpitt:
So much gaoling.

keithpitt:
Also golang

PR #200, opened 2015-10-19
Added experimental support for plugins

keithpitt:
Initial cut of a plugin implementation for the agent. Still very much a WIP.

### DIFF
--- a/agent/bootstrap.go
+++ b/agent/bootstrap.go
@@ -226,29 +226,37 @@ func checkShellError(err error, cmd *shell.Command) {
 }
 
 // Aquires a lock on the folder
-func aquireLock(path string, seconds int) (lockfile.Lockfile, error) {
-	lock, _ := lockfile.New(path)
+func aquireLock(path string, seconds int) (*lockfile.Lockfile, error) {
+	lock, err := lockfile.New(path)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to create lock \"%s\" (%s)", path, err)
+	}
 
 	// Aquire the lock. Keep trying the lock until we get it
 	attempts := 0
 	for true {
 		err := lock.TryLock()
 		if err != nil {
-			// Keey track of how many times we tried to get the
-			// lock
-			attempts += 1
-			if attempts > seconds {
-				return lock, fmt.Errorf("Gave up trying to aquire a lock on \"%s\"", path)
-			}
+			switch err {
+			case lockfile.ErrBusy:
+				// Keey track of how many times we tried to get the
+				// lock
+				attempts += 1
+				if attempts > seconds {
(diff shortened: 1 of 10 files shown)


toolmantim:
This is such a good idea! Love it.

What do you think about requiring people to `bundle-agent plugin install` a plugin before it's allowed to be referenced in a `pipeline.yml`? Then people can script/automate the plugin install/updating process, rather than it just happening live in job runs.

keithpitt:
@toolmantim I thought about for install, but then realised it's a bit of a PITA when you spin up a new agent in your cluster (oh we can't run test on `agent-1` yet because we need to manually install all these plugins). `plugin update` was another thing I thought about, but I haven't come up with a good way of knowing if a plugin version is out of date (currently plugins don't really have versions). They can be locked to commits or branches though (therefor you can't update commit-locked plugins, but branches you can). For now you can just `rm -rf /var/lib/buildkite-agent/plugins` to get a clean update of them all.

toolmantim:
Also we could pre-install plugins as part of installs, such as https://github.com/buildkite/buildkite-aws-stack, baked into their aws images. For that sort of thing I don't think @lox would want plugins auto-install/updating on machines, he'd cut a new build for people.

I could imagine also adding some args to the homebrew recipe for installing some plugins on your behalf (fastlane & xcode) so the installation for Xcode peeps becomes one command.

Also for Github: repo-per-plugin or one repo with all the plugins.

toolmantim:
The idea of being able to test out a new plugin by just branching and changing the yml file (like it currently works) is pretty neat tho!

keithpitt:
I hear you about pre-installing plugins during docker builds! There's probably a use case there.

agent/bootstrap.go
```diff
@@ -220,29 +226,38 @@ func checkShellError(err error, cmd *shell.Command) {
 }
 
 // Aquires a lock on the folder
-func aquireLock(path string, seconds int) (lockfile.Lockfile, error) {
-	lock, _ := lockfile.New(path)
+func aquireLock(path string, seconds int) (*lockfile.Lockfile, error) {
```
plasticine: `acquireLock`? :trollface:

keithpitt: Lolololol

plasticine:
:fire: 4 dis! :dancer:

keithpitt:
For anyone looking at the codez, the plugin implementation and Golang version of the bootstrap are very messy. My first step was just everything in there, from there I can start to refactor into smaller parts (I didn't want to spend a lot of time trying to modularise everything and then be like "lol, didn't need this after all").

There are a couple of open source packages that will fall out of the agent I think. One will be the `shell` package, which contains a bunch of helpers for working with the shell cross platform. Also more of the bootstrap will be merged into the `shell` package.

PR #201, opened 2015-10-19
Single branch fetch take 2

byroot:
Another take at https://github.com/buildkite/agent/commit/d19df370bca173a8d41abe5d536545057f413b7e

Can be tested like this:

```
git clone --single-branch -b offical-aws-lib -- \
  git@github.com:buildkite/agent.git
cd agent
git fetch origin refs/heads/offical-aws-lib
git reset --hard refs/heads/offical-aws-lib
```

NB: `--single-branch` doesn't create a shallow repository, only `--depth` does. But it's a repository that doesn't track all upstream branches.

Also I totally understand if you think this is too complicated, we can run our own checkout script, we don't do it for simplicity though.

### DIFF
--- a/templates/bootstrap.sh
+++ b/templates/bootstrap.sh
@@ -240,7 +240,7 @@ else
     # If the commit is HEAD, we can't do a commit-only fetch, we'll need to use
     # the branch instead.
     if [[ "$BUILDKITE_COMMIT" == "HEAD" ]]; then
-      buildkite-run "git fetch -q origin $BUILDKITE_BRANCH 2> /dev/null || git fetch -q"
+      buildkite-run "git fetch -q origin refs/heads/$BUILDKITE_BRANCH 2> /dev/null || git fetch -q"
     else
       # First try to fetch the commit only (because it's usually much faster).
       # If that doesn't work, just resort back to a regular fetch.
@@ -250,7 +250,7 @@ else
     if [[ "$BUILDKITE_TAG" == "" ]]; then
       # Default empty branch names
       : "${BUILDKITE_BRANCH:=master}"
-      buildkite-run "git reset --hard origin/$BUILDKITE_BRANCH"
+      buildkite-run "git reset --hard refs/heads/$BUILDKITE_BRANCH"
     fi
   fi
 
--- a/templates/bootstrap.sh
+++ b/templates/bootstrap.sh
@@ -218,7 +218,12 @@ else
   if [[ -d ".git" ]]; then
     buildkite-run "git remote set-url origin \"$BUILDKITE_REPO\""
   else
-    buildkite-run "git clone -qv -- \"$BUILDKITE_REPO\" ."
+    # Attempt to clone just a single branch so first time clones are fast
+    if git clone --help | grep -- --single-branch &> /dev/null; then
+      buildkite-run "git clone -qv --single-branch -b \"$BUILDKITE_BRANCH\" -- \"$BUILDKITE_REPO\" ."
+    else
+      buildkite-run "git clone -qv -- \"$BUILDKITE_REPO\" ."
+    fi
   fi
 
   buildkite-run "git clean -fdq"


keithpitt:
Here's the test I tried:

Create `master/HEAD` build

```
$ git clone -qv --single-branch -b master -- https://github.com/buildkite/agent-tests.git .
Cloning into '.'...
POST git-upload-pack (152 bytes)
remote: Counting objects: 714, done.
remote: Compressing objects: 100% (149/149), done.
remote: Total 714 (delta 95), reused 0 (delta 0), pack-reused 535
Receiving objects: 100% (714/714), 10.03 MiB | 587.00 KiB/s, done.
Resolving deltas: 100% (330/330), done.
Checking connectivity... done.
$ git clean -fdq
$ git submodule foreach --recursive git clean -fdq
$ git fetch -q origin refs/heads/master
$ git reset --hard refs/heads/master
HEAD is now at e3630a8 Need git@ for private repos
$ git checkout -qf HEAD
$ git submodule sync --recursive
$ git submodule update --init --recursive
$ git submodule foreach --recursive git reset --hard
```

Then create a `windows-pr/HEAD` build. I then got this error:

```
$ git remote set-url origin https://github.com/buildkite/agent-tests.git
$ git clean -fdq
$ git submodule foreach --recursive git clean -fdq
$ git fetch -q origin refs/heads/windows-pr
$ git reset --hard refs/heads/windows-pr
fatal: ambiguous argument 'refs/heads/windows-pr': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
exit status 128
```

TimMoore:
@keithpitt I got a similar error in the released 2.1.4 agent:

```
$ mkdir -p "/var/lib/buildkite-agent/builds/ip-10-0-40-169-2/culture-amp/murmur"
$ cd "/var/lib/buildkite-agent/builds/ip-10-0-40-169-2/culture-amp/murmur"
$ git remote set-url origin "git@github.com:cultureamp/murmur.git"
$ git clean -fdq
$ git submodule foreach --recursive git clean -fdq
$ git fetch -q origin cc7eed67f5f3c9ec216ab114837adbf34a552d13 2> /dev/null || git fetch -q
$ git reset --hard origin/release_064
fatal: ambiguous argument 'origin/release_064': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]’
```

I'm thinking it should be fetching the branch and resetting to the commit instead of the other way around?

keithpitt:
@TimMoore weird.. I knew we had a problem in the `2.1` release, but I thought I'd fixed that in `2.1.4`. Does your `bootstrap.sh` file have the phrase `single-branch` in it?

byroot:
Could be because that repo was initially cloned with `--single-branch` and then later the agent was upgraded to 2.1.4?

If so the simplest course of action might be to do a clean checkout.

Sorry for that :/

keithpitt:
Ah yes, that could be it!

No sweat, it worked for me when I tested it :P (just forgot to test with a fresh repo)

byroot:
We went with a custom bootstrap script in the end, so closing.

keithpitt:
I've had another idea about this by the way.

Instead of having each build folder do it's own `git clone git@github.com:/...`, what if the agent by default, cloned repositories to `/var/buildkite-agent/repositories/git-github-com-x-y`, and each build just `git fetched` and `git cloned` from that directory, instead of the internet.

We could do some locking where only 1 agent does the `git fetch` for the entire build, and the rest just clone again from the cached repo. What do you think?

(Idea inspired from https://github.com/square/git-fastclone)

byroot:
It's precisely what we are doing in our custom script.

We don't lock though but I think we should because we had a couple corrupt cache repos (I'm still not entirely sure what the root cause is, and it doesn't happen a lot but I keep an eye on it).

keithpitt:
@toolmantim just pointed out to me that git has a lock file, but maybe that's not very reliable?

keithpitt:
I'm glad that we've come to the same idea though :) I'm keen to roll this into the agent as the default behaviour. Would you be able to share what you've done?

byroot:
It mostly is yes. That's why I'm not sure the corrupted repositories were because of the concurrency. Also the 2 corruptions we had were on the same host in a very short time, so I blame network / ec2 fuckery even though I have no proof.



> Would you be able to share what you've done?

Sure there is really no secret sauce in there, but I'm not sure if it would be super useful, I started from scratch in ruby, and I use modern git options (`uploadpack.allowReachableSHA1InWant `)

But just in case it would help: https://gist.github.com/byroot/2b4b6d02af1b123b2554

byroot:
Also note that local clones, if on the same physical drive, will automatically use hardlinks to save space and also be faster.

keithpitt:
Oh wow! You rewrote the bootstrap in ruby!

keithpitt:
Where does the hardlink stuff happen? Maybe I'm blind and can't see it.

keithpitt:
Ah! @toolmantim just explained to me that hard links are a git thing :)

byroot:
In git itself, my message probably wasn't clear. You have an option to force it, but usually git will do it if it can:


from `git help clone`
```
       --local, -l
           When the repository to clone from is on a local machine, this flag bypasses the normal "Git aware" transport mechanism and clones the repository by
           making a copy of HEAD and everything under objects and refs directories. The files under .git/objects/ directory are hardlinked to save space when
           possible.

           If the repository is specified as a local path (e.g., /path/to/repo), this is the default, and --local is essentially a no-op. If the repository is
           specified as a URL, then this flag is ignored (and we never use the local optimizations). Specifying --no-local will override the default when
           /path/to/repo is given, using the regular Git transport instead.
```

PR #203, opened 2015-10-20
Use TrimPrefix instead of TrimLeft

### LINKED ISSUE #202: S3 bucket names starting with an "s" don't work
All the leading `s`'es in a bucket name are chopped off for some reason leading uploads to a bucket with a name starting with `s` to fail.

Reproduce:
- Create an s3 bucket called `something-is-not-right-here-<random string>`
- Give it the correct permissions for your access keys
- Try uploading to it, you'll get a 'bucket does not exist error' and notice that it's actually looking for `omething-is-not-right-here-<random string>`.

Edit: this is most likely due to using `TrimLeft` [here](https://github.com/buildkite/agent/blob/master/agent/s3_uploader.go#L122). See http://stackoverflow.com/a/29187137 for more info.

joostdevries:
See http://stackoverflow.com/questions/29187086/why-trimleft-doesnt-work-as-expected

Fixes #202 
NOT TESTED (by me)

### DIFF
--- a/agent/s3_downloader.go
+++ b/agent/s3_downloader.go
@@ -42,7 +42,7 @@ func (d S3Downloader) Start() error {
 	}
 
 	// Split apart the bucket
-	bucketParts := strings.Split(strings.TrimLeft(d.Bucket, "s3://"), "/")
+	bucketParts := strings.Split(strings.TrimPrefix(d.Bucket, "s3://"), "/")
 	bucketName := bucketParts[0]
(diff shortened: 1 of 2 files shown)


keithpitt:
@joostdevries oh wow! Thanks for finding this bug! I'll merge now and it'll be available in the next beta if you wanted to try it out there.

keithpitt:
I'm going to replace other references to `TrimLeft` now! Thanks again for pointing it out!

PR #207, opened 2015-10-23
Add support for artifacts on Google Cloud Storage

grosskur:
The Buildkite agent currently allows storing artifacts in an S3 bucket
with the `s3://` scheme. This change adds support for Google Cloud
Storage buckets with the `gs://` scheme. Details:

* GS authentication is handled using Google's OAuth2 library, which
  searches for the credentials in various locations. See:

    https://godoc.org/golang.org/x/oauth2/google#DefaultClient

* GS ACLs are similar to S3 ACLs. The default GS ACL is `publicRead`,
  which is analagous to S3's `public-read`. I'd actually prefer if it
  was safer and defaulted to `private` for both GS and S3, but I left
  it this way for consistency.

* GS downloading is handled by reusing `download.go`, similar to S3
  downloading. An `http.Client` field is now passed in via the
  `Download` struct. This allows the GS code to pass in a client that
  was created by Google's OAuth2 library and automatically adds an
  appropriate `Authorization` header to requests.

* There are basically no test cases yet for GS uploading and
  downloading. I borrowed a few from S3 in `gs_uploader_test.go`, but
  I'm not sure if they make sense for GS.

* This change adds dependencies on the following imports:

    `golang.org/x/net/context`
    `golang.org/x/oauth2/google`
    `google.golang.org/api/storage/v1`

  These should be commited under the `vendor` directory, probably in a
  separate change.

### DIFF
--- a/agent/artifact_downloader.go
+++ b/agent/artifact_downloader.go
@@ -75,7 +75,7 @@ func (a *ArtifactDownloader) Download() error {
 				} else if strings.HasPrefix(artifact.UploadDestination, "gs://") {
 					err = GSDownloader{
 						Path:        artifact.Path,
-						URL:         artifact.URL,
+						Bucket:      artifact.UploadDestination,
 						Destination: downloadDestination,
 						Retries:     5,
 						DebugHTTP:   a.APIClient.DebugHTTP,
--- a/agent/artifact_downloader.go
+++ b/agent/artifact_downloader.go
@@ -63,11 +63,19 @@ func (a *ArtifactDownloader) Download() error {
 			p.Spawn(func() {
 				var err error
 
-				// Handle downloading from S3
+				// Handle downloading from S3 and GS
 				if strings.HasPrefix(artifact.UploadDestination, "s3://") {
 					err = S3Downloader{
 						Path:        artifact.Path,
 						Bucket:      artifact.UploadDestination,
+						Destination: downloadDestination,
+						Retries:     5,
+						DebugHTTP:   a.APIClient.DebugHTTP,
+					}.Start()
+				} else if strings.HasPrefix(artifact.UploadDestination, "gs://") {
+					err = GSDownloader{
+						Path:        artifact.Path,
+						URL:         artifact.URL,
 						Destination: downloadDestination,
 						Retries:     5,
 						DebugHTTP:   a.APIClient.DebugHTTP,
--- a/agent/artifact_uploader.go
+++ b/agent/artifact_uploader.go
@@ -139,6 +139,8 @@ func (a *ArtifactUploader) upload(artifacts []*api.Artifact) error {
 	if a.Destination != "" {
 		if strings.HasPrefix(a.Destination, "s3://") {
 			uploader = new(S3Uploader)
+		} else if strings.HasPrefix(a.Destination, "gs://") {
+			uploader = new(GSUploader)
 		} else {
 			return errors.New("Unknown upload destination: " + a.Destination)
 		}
--- a/agent/download.go
+++ b/agent/download.go
@@ -15,6 +15,9 @@ import (
 )
 
 type Download struct {
+	// The HTTP client to use for downloading
+	Client http.Client
+
 	// The actual URL to get the file from
 	URL string
 
@@ -74,7 +77,7 @@ func (d Download) try() error {
 	logger.Debug("Downloading %s to %s", d.URL, targetFile)
 
 	// Start by downloading the file
-	response, err := http.Get(d.URL)
+	response, err := d.Client.Get(d.URL)
 	if err != nil {
 		return fmt.Errorf("Error while downloading %s (%T: %v)", d.URL, err, err)
 	}
--- a/agent/gs_downloader.go
+++ b/agent/gs_downloader.go
@@ -36,7 +36,7 @@ func (d GSDownloader) Start() error {
 	// We can now cheat and pass the URL onto our regular downloader
 	return Download{
 		Client:      *client,
-		URL:         d.URL + "?alt=media",
+		URL:         d.URL,
 		Path:        d.Path,
 		Destination: d.Destination,
 		Retries:     d.Retries,
--- a/agent/gs_downloader.go
+++ b/agent/gs_downloader.go
@@ -1,0 +1,45 @@
+package agent
+
+import (
+	"errors"
+	"fmt"
+
+	"golang.org/x/net/context"
+	"golang.org/x/oauth2/google"
+	storage "google.golang.org/api/storage/v1"
+)
+
+type GSDownloader struct {
+	// The URL
+	URL string
+
+	// The root directory of the download
+	Destination string
+
+	// The relative path that should be preserved in the download folder,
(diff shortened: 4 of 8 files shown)


keithpitt:
Nice! I'll check this out soon and will let you know how it all looks. I had a quick scan already and it all looks pretty spot on (just gotta do some manual testing)

keithpitt:
Hey @grosskur, I just checked out the PR locally and it all works a dream! Thanks so much for taking the time to do this!!

Here a couple of changes I think we should make:

1. You're right about the default permission, it should be `private`. Let's change it, and we'll start to default the permission in S3 as well.
2. Since you can define `S3_` credentials with the `BUILDKITE_` prefix, it'd be nice to be able to define `GOOGLE_APPLICATION_CREDENTIALS` with `BUILDKITE_GS_APPLICATION_CREDENTIALS` maybe?
3. I've got a change pending to push to Buildkite that will return the URL for GS artifacts already with the `?alt=media` suffix. I needed to change this so when you click on an artifact in the UI, it'll download it, and not take you to the info screen. I haven't pushed this yet just in case you were already using this GS code in production. If you're not, I can push this change now, and you can remove the `?alt=media` line from the `GSDownloader`

Thanks again!!! :dancers:

keithpitt:
Once it hits master I'll vendor the new packages

grosskur:
Thanks for taking the time to look this over! A few of thoughts/questions:

1. I've been thinking about the default permission. Google Cloud Storage has a concept of "default ACL" that can be set on a per-bucket basis. See the [gsutil defacl](https://cloud.google.com/storage/docs/gsutil/commands/defacl) command. Ideally, I'd like to make `buildkite-agent` respect the default ACL and *not* call `PredefinedAcl()` when creating the object, unless `BUILDKITE_GS_ACL` explicitly is set.

   For S3, I'd also like to remove the explicit default. S3 [defaults to private](http://docs.aws.amazon.com/AmazonS3/latest/dev/s3-access-control.html) anyway, and Amazon seems to recommend controlling access via [policies](http://docs.aws.amazon.com/AmazonS3/latest/dev/access-policy-alternatives-guidelines.html).

   Overall, I think the recommended practice for using an S3 or GS bucket with Buildkite should be to set the permissions externally via ACL or IAM policy and just let Buildkite use that. With this model `BUILDKITE_GS_ACL` and `BUILDKITE_S3_ACL` probably aren't even necessary, but maybe it's good to still support them as overrides for awhile. What do you think?

2. Sure, I can make the code look for the `BUILDKITE_GS_APPLICATION_CREDENTIALS` environment variable and call [JWTConfigFromJSON](https://godoc.org/golang.org/x/oauth2/google#JWTConfigFromJSON) if present, and otherwise fall back to [DefaultClient](https://godoc.org/golang.org/x/oauth2/google#DefaultClient).

3. Sounds good. The other option is `GSUploader` could explicitly set the URL to contain `?alt=media`. Then the Buildkite server code doesn't need to add `?alt=media`. IMO, it seems a little weird for the API to change the URL to something different than `GSUploader` set it to. But maybe that solves other problems? Let me know what you think---I can change the code to work either way.

keithpitt:
1. Yeah, I think it's good to default what the bucket says, but I'm keen to leave in the `BUILDKITE_S3_ACL` as an override option (because it's a bit easier to do then futz around with ACL permissions)
2. :+1:
3. Yeah, I think you're option is better :) I didn't push the code I wrote, so I'll ditch I wrote, and if you add the `?alt=media` stuff to the Uploader, the UI should just work I think! Just double check by clicking on one of the artifacts you upload in the Buildkite UI (it should be downloadable)

grosskur:
Cool, I'll update this PR shortly...

keithpitt:
Awesome, thanks @grosskur! :+1:

grosskur:
I just pushed another commit to cover the things we talked about. I didn't remove the default `public` ACL for S3 yet---I'd kind of prefer to do that in a separate change.

One thing still doesn't feel right, though. When I tested a URL by downloading it via the Buildkite web interface, it didn't work for a private object (I got a 401). It did work for a public object, but (1) the filename wasn't what I expected, and (2) the file didn't get displayed inline in the browser even though the mime type was `text/plain`.

The reason the filename wasn't what I expected is because the form of the URL is:

`https://www.googleapis.com/storage/v1/b/<bucket>/o/<object>?alt=media`

If the object is in a subdirectory, `i.e.`, it has slashes in the name, that subdirectory will get prepended to the downloaded filename. I tried downloading `foo/bar/blah.log` and I got a file called `foo-bar-blah.txt`.

Looking at [Concepts and Techniques](https://cloud.google.com/storage/docs/concepts-techniques?hl=en), it seems like there are two kinds of URLs for a file on Google Cloud Storage:

1. **API authentication.** This is the one that has `?alt=media` at the end. To access this kind of URL, you need to set the `Authorization` header appropriately. This is what `buildkite-agent` uses to download artifacts.

2. **Cookie-based authenticated.** This one looks like `https://storage.cloud.google.com/<bucket>/<object>` and, if necessary, it will walk you through the OAuth flow to download it. This sounds like what the Buildkite web interface should use, I think.

So now I actually think we should store the cookie-auth version in the artifact's URL field, since then everything will Just Work in the web interface. The `GSDownloader` code can ignore this field and construct the API-auth version of the URL using the bucket name and object name. How does that sound?

keithpitt:
Yeah, so downloading private artifacts via the Buildkite UI is broken, and it's OK to be broken. It's the same with S3. For most companies, they'd implement a "bucket policy" that allows them to access artifacts from their office IP address. It's a bit suckey, but ideally we wouldn't store any sort of keys in the artifact URL (if you're using a private bucket, it's because you don't want us access to them).

So as long as `artifact download` can get to them, that's OK. If they make the artifacts publicly readable, then they'll be able to access them in the UI.

keithpitt:
> If the object is in a subdirectory, i.e., it has slashes in the name, that subdirectory will get prepended to the downloaded filename. I tried downloading foo/bar/blah.log and I got a file called foo-bar-blah.txt.

Yeah, our downloading file system logic is crappy. I have plans to rewrite the rules around the downloads to be more unix like. As long as it works how the current S3 one works, that's OK for now.

keithpitt:
@grosskur I reckon we can merge this? :dancer:

grosskur:
Sorry for the long delay. I have one more commit I'm working on---hopefully done today. Basically, I can make private artifacts downloadable in the Buildkite UI after all. The `https://storage.cloud.google.com/<bucket>/<object>` form of the URL has no private information in it, and when you click the link, it's protected by Google OAuth. So I think this is the best of both worlds.

grosskur:
Sorry again for the delay---I've updated this PR and, pending testing on your side, I think it's ready to merge. The main difference now is that publicly-accessible artifacts can be downloaded properly in the Buildkite UI and have the correct filename on disk. Private artifacts still get 403, which is consistent with how private S3 artifacts behave in the Buildkite UI. Thanks for your patience!

keithpitt:
This is awesome! Thanks so much for doing this work @grosskur! I'll include it in the next beta release of the agent :D

PR #212, opened 2015-11-12
Update launchd templates to use .buildkite-agent dir

julianwa:
Should match dir structure created by https://raw.githubusercontent.com/buildkite/agent/master/install.sh

### DIFF
--- a/templates/launchd.plist
+++ b/templates/launchd.plist
@@ -14,7 +14,7 @@
 
     <key>ProgramArguments</key>
     <array>
-      <string>/Users/your-build-user/.buildkite/bin/buildkite-agent</string>
+      <string>/Users/your-build-user/.buildkite-agent/bin/buildkite-agent</string>
       <string>start</string>
       <!-- <string>--debug</string> -->
     </array>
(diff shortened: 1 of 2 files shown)


toolmantim:
Thanks for the catch here!

julianwa:
NP. Should also update these docs:

https://buildkite.com/docs/agent/osx#starting-on-login

`s/LaunchAgent/LaunchAgents/`
`s/.buildkite/.buildkite-agent/`

toolmantim:
Thanks! I've just fixed those in buildkite/docs@a2570cd3342610269bd93bd69a6a1e6000e9209f and buildkite/docs@a2570cd3342610269bd93bd69a6a1e6000e9209f, and it's been updated on https://buildkite.com/docs/agent/osx - let me know if that all looks good :+1:

julianwa:
LGTM, thanks! (Sorry, didn't realize the docs were in the repo or I would have included in my PR.)

One other note: it's not clear that the `sudo`s are necessary since all of the files affected should be owned by the current user.

toolmantim:
:+1: No worries, they're a separate repo anyhoo. And I've removed those `sudo`s now too, nice catch (again)

PR #215, opened 2015-11-20
Send desired signal instead of always SIGTERM

nikyoudale:
I've been observing agents not successfully killing build jobs on timeout recently. The process of `SIGTERM`, wait 10 seconds the `SIGKILL` sounds fine. This fix should make that work as intended.


keithpitt:
Ah, this is a good pickup! Thanks!

PR #218, opened 2015-12-01
Add bootstrap-script path back to default config

dgoodlad:
This was removed in 80373be618fc6aacbda88f756e16dcc5c3725612, but is still required in 2.1.x agents. Without this, after `apt-get install buildkite-agent` you will _not_ be able to start the agent.

### DIFF
--- a/packaging/linux/root/usr/share/buildkite-agent/buildkite-agent.cfg
+++ b/packaging/linux/root/usr/share/buildkite-agent/buildkite-agent.cfg
@@ -13,6 +13,11 @@ name="%hostname-%n"
 # Populate the meta data from the current instances EC2 Tags
 # meta-data-ec2-tags=true
 
+# Path to the bootstrap script. You should avoid changing this file as it will
+# be overridden when you update your agent. If you need to make changes to this
+# file: use the hooks provided, or copy the file and reference it here.
+bootstrap-script="/usr/share/buildkite-agent/bootstrap.sh"
+
 # Path to where the builds will run from
 build-path="/var/lib/buildkite-agent/builds"
 


keithpitt:
Ah, sorry about that @dgoodlad! Does the latest `2.1.x` agent have a broken config?

keithpitt:
Hey @dgoodlad - I just installed the latest stable agent from `apt`, and it looks like it's got the right config.  How are you installing the agent?

dgoodlad:
@keithpitt I actually pulled the config from this repo here, customised it with the token, and installed it when building my agent AMI. So in reality, the package's default config probably got overwritten before it tried to start.

Where does the package get it's default config from when you build it?

keithpitt:
So the package gets it's default config from the file you changed, but the latest stable has the older version of this file packaged in it, with that line existing in it. I'll merge this in, and just comment out the line (instead of removing it all together) so if anyone copy/pastes it from master, they'll know what's going on (since we're planning on removing the bootstrap script all together)

keithpitt:
Sorry that it caused you some pain @dgoodlad!

dgoodlad:
Where does the current stable package get built from? e.g. how does it have "the old version" of the default config? Is there a branch I should've grabbed it from instead?

keithpitt:
So the current stable package was built from master a few months ago (when it was released). We don't have a great branch/versioning strategy at the moment - mostly because I'm not sure what the best approach is. Perhaps we should have branches for each version or something? We default the branch on the GH UI to the latest release or something?

dgoodlad:
Tags plus a `stable` branch that you cut new stable package releases from is my favoured approach. The `stable` branch could be `master`, and then you maintain a separate `dev` or `unstable` branch, whatever, but if you've fundamentally got two separate release tracks for packages, you should have two separate branches.

sj26:
@dgoodlad I've added a 2-1-stable branch which will track the 2.1 line for now. We're going to cherry pick some stuff across and do another release.

PR #232, opened 2016-01-14
Remove gitignore'd files on clean

### LINKED ISSUE #173: Ability to change `git clean` behaviour to also include `-x`
`-x` will also remove files that are included in the `.gitignore` list. So essentially a `-x` clean is a true clean of the repo.

toolmantim:
For ages we've spoken about changing the default behaviour so that it cleans all the files between jobs. This updates the golang bootstrap to remove all files, even those that are git ignored.

Something to consider: how do you achieve wanting to keep files around between builds? Do you just use the artifact gear, or your own `cp` commands on the file system, if you want to assume a single box? We could add it as an agent config if we really wanted.

Closes #173

### DIFF
--- a/agent/bootstrap.go
+++ b/agent/bootstrap.go
@@ -844,11 +844,11 @@ func (b *Bootstrap) Start() error {
 		}
 
 		// Clean up the repository
-		b.runCommand("git", "clean", "-fdq")
+		b.runCommand("git", "clean", "-fdqx")
 
 		// Also clean up submodules if we can
 		if b.GitSubmodules {
-			b.runCommand("git", "submodule", "foreach", "--recursive", "git", "clean", "-fdq")
+			b.runCommand("git", "submodule", "foreach", "--recursive", "git", "clean", "-fdqx")
 		}
 
 		// Allow checkouts of forked pull requests on GitHub only. See:


jules2689:
We'd really appreciate these changes. 

In a recent issue, one of our teams had a gitignored folder `iOS/builds`, which was rightfully ignored from github. This folder was used for various output, but needs to be cleared between builds. 

`git clean -fdq` doesn't take into account gitignored files/folders - but `-fdqx` does. So :+1:

While I realize this may break some behaviours on which people relied in the past, I don't think that files should be persistent in the repo between builds. If they need persistent files - for whatever reason - then a folder outside of the repo can be used for these purposes. This gives a bonus of making sure the repo is in a totally clear state before each build.

So, Tl;Dr; This is good and a step in the right direction IMO

keithpitt:
We'll probably need to wrap this up in an option, default it to the old behaviour in the next release, then default it to the on in the release after that, then in v3, remove the option, and just have it be the default behaviour.

keithpitt:
Another thing we need to add to this branch is an additional `-f` to the git clean, so it'd make the final command `git clean --ffdxq`.

> Git will refuse to delete directories with .git sub directory or file unless a second -f is given. This affects also git submodules where the storage area of the removed submodule under .git/modules/ is not removed until -f is given twice.

dkubb:
>  then in v3, remove the option, and just have it be the default behaviour.

It would be nice to have this stay configurable. I would like to have a few `--exclude` switches to prevent some cached data around between builds.

>  `git clean --ffdxq`

There's only one `--`, so it'd be `git clean -ffdxq.`

toolmantim:
This is being dealt with in #240 #264

PR #237, opened 2016-02-12
bootstrap: Add support for BUILDKITE_REFSPEC

grosskur:
This is useful to integrating Buildkite with Gerrit. Basically, in the Gerrit workflow commits for patchsets are not done on branches. Instead, they each have a ref of the form:

    refs/changes/XX/XX/XX

In order to build the commit, you must first pull the ref. So add support for a new environment variable BUILDKITE_REFSPEC that can be set to indicate the refspec that Buildkite should pull.

This is a generalization of the support for building pull requests from GitHub. In that case Buildkite figures out the ref automatically from GITHUB_PULL_REQUEST environment variable.

### DIFF
--- a/agent/bootstrap.go
+++ b/agent/bootstrap.go
@@ -41,6 +41,9 @@ type Bootstrap struct {
 	// The tag of the job commit
 	Tag string
 
+	// A refspec to fetch
+	RefSpec string
+
 	// Plugin definition for the job
 	Plugins string
 
@@ -858,6 +861,8 @@ func (b *Bootstrap) Start() error {
 		// https://help.github.com/articles/checking-out-pull-requests-locally/#modifying-an-inactive-pull-request-locally
 		if b.PullRequest != "false" && strings.Contains(b.PipelineProvider, "github") {
 			b.runCommand("git", "fetch", "-q", "origin", "+refs/pull/"+b.PullRequest+"/head:")
+		} else if b.RefSpec != "" {
+			b.runCommand("git", "fetch", "-q", "origin", b.RefSpec)
 		} else {
 			// If the commit is HEAD, we can't do a commit-only
 			// fetch, we'll need to use the branch instead.  During
--- a/clicommand/bootstrap.go
+++ b/clicommand/bootstrap.go
@@ -31,6 +31,7 @@ type BootstrapConfig struct {
 	Commit                       string `cli:"commit" validate:"required"`
 	Branch                       string `cli:"branch" validate:"required"`
 	Tag                          string `cli:"tag"`
+	RefSpec                      string `cli:"refspec"`
 	Plugins                      string `cli:"plugins"`
 	PullRequest                  string `cli:"pullrequest"`
 	GitSubmodules                bool   `cli:"git-submodules"`
@@ -91,6 +92,12 @@ var BootstrapCommand = cli.Command{
 			Value:  "",
 			Usage:  "The tag the commit",
 			EnvVar: "BUILDKITE_TAG",
+		},
+		cli.StringFlag{
+			Name:   "refspec",
+			Value:  "",
+			Usage:  "The refspec this commit belonged to",
+			EnvVar: "BUILDKITE_REFSPEC",
 		},
 		cli.StringFlag{
 			Name:   "plugins",
--- a/templates/bootstrap.sh
+++ b/templates/bootstrap.sh
@@ -231,6 +231,8 @@ else
   # https://help.github.com/articles/checking-out-pull-requests-locally/#modifying-an-inactive-pull-request-locally
   if [[ "$BUILDKITE_PULL_REQUEST" != "false" ]] && [[ "$BUILDKITE_PIPELINE_PROVIDER" == *"github"* ]]; then
     buildkite-run "git fetch origin \"+refs/pull/$BUILDKITE_PULL_REQUEST/head:\""
+  elif [[ "${BUILDKITE_REFSPEC:-}" != "" ]]; then
+    buildkite-run "git fetch origin \"$BUILDKITE_REFSPEC\""
   else
     # If the commit is HEAD, we can't do a commit-only fetch, we'll need to use
     # the branch instead.


sj26:
Makes sense. So this is useful when you want to build anything that's not a default head or tag, and doesn't fit the specific pull request scenario already handled? Does this mean you have something like `BUILDKITE_BRANCH="refs/changes/XX/XX/XX"`? Or do you have a `BUILDKITE_BRANCH="changes/XX/XX/XX"` and you want a refspec of `+refs/changes/*:changes/*`?

Maybe we can consolidate the pull request behaviour with this as well. We might be able to get away without the option at all. Buildkite strips `refs/heads/` and `refs/tags/` from `BUILDKITE_BRANCH`, but if the value still has a `refs/` prefix we could special-case the fetch. Maybe something like:

```sh
  # Allow checkouts of forked pull requests on GitHub only. See:
  # https://help.github.com/articles/checking-out-pull-requests-locally/#modifying-an-inactive-pull-request-locally
  if [[ "$BUILDKITE_PULL_REQUEST" != "false" ]] && [[ "$BUILDKITE_PIPELINE_PROVIDER" == *"github"* ]]; then
    buildkite-run "git fetch origin \"+refs/pull/$BUILDKITE_PULL_REQUEST/head:\""

  # Allow checkout of non-standard refs with explicit fetch
  elif [[ "${BUILDKITE_BRANCH##refs/}" != "${BUILDKITE_BRANCH}" ]]; then
    buildkite-run "git fetch origin \"+$BUILDKITE_BRANCH:\""

  # Fall back to a standard fetch
  else
    buildkite-run "git fetch origin"
  fi
```

Would this work for you?

grosskur:
Thanks for taking a look! Right---this is useful for building any ref that isn't of the form `refs/heads/*`, `refs/tags/*`, or `refs/pulls/*`.

With Gerrit, you are still logically working on a branch (usually `master`) and you make a local topic branch. When you run `git-review`, your changes in the topic branch are pushed to the central repository as refs of the form `refs/changes/XX/XX/XX`. In our case, when we build, set actually both `BUILDKITE_BRANCH=master` and `BUILDKITE_REFSPEC=refs/changes/XX/XX/XX`.

You're right that `BUILDKITE_BRANCH` doesn't get used in the Git checkout code if `BUILDKITE_REFSPEC` is set, so they seem somewhat redundant. Personally, I think it's useful to have both, though, for two reasons:

1. We sometimes have parallel development happening where people are building Gerrit patchsets based on different branches, *e.g.*, `master` and `my-feature`. Having `BUILDKITE_BRANCH=master` or `BUILDKITE_BRANCH=my-feature` allows us to easily differentiate builds in the web UI and API. It also lets us take advantage of Buildkite's [branch filtering](https://buildkite.com/docs/guides/branch-configuration) in pipeline steps to, *e.g.*, add `branches: master` to some steps. Or we could use the value of `$BUILDKITE_BRANCH` to upload a different set of pipeline steps altogether.

2. A Git [refspec](https://git-scm.com/book/en/v2/Git-Internals-The-Refspec) is a more general concept than a branch. The format of the refspec is an optional `+`, followed by `<src>:<dst>`, where `<src>` and `<dest>` are both patterns. For some strange reason, I might want a refspec of `refs/heads/*:refs/foobar/*`. If I put this value in `BUILDKITE_BRANCH` and Buildkite naively strips `refs/heads/` and `refs/tags/`, I'll end up with `BUILDKITE_BRANCH=*:refs/foobar/*`, which won't work. You could of course make Buildkite smarter about how it strips prefixes, but at this point it seems more straightforward to just use separate variables, each with clearly-defined semantics.

sj26:
Interesting! Do you also push a `BUILDKITE_COMMIT`? We can also explicitly fetch a commit, which I feel like we should be doing anyway. I kind of want to change all of this to be:

```sh
git fetch origin "$BUILDKITE_COMMIT"
git checkout -B "$BUILDKITE_BRANCH" "$BUILDKITE_COMMIT"
git config "branch.$BUILDKITE_BRANCH.remote" origin
```

I think this should actually work for most scenarios. Would this work for yours? I might do some testing myself as well.

sj26:
I was hoping you could just fetch a commit but it does not seem to be so. :-(

grosskur:
Yes, we also set `BUILDKITE_COMMIT` to the commit SHA. But I couldn't get `git fetch` to work on a single commit. I tried with both git-1.9.1 (from Ubuntu 14.04) and git-2.7.1 (latest).

[This stackoverflow answer](http://stackoverflow.com/questions/14872486/retrieve-specific-commit-from-a-remote-git-repository/30701724#30701724) seems to indicate that it's possible to fetch a single commit if the server has the config option `uploadpack.allowReachableSHA1InWant`. Our Git server is Gerrit, which is based on JGit, and its look like that config option is available when using [Gerrit 2.12 (latest) and the download-commands plugin](https://gerrit.googlesource.com/plugins/download-commands/+/87075138f5b27e9aba91a3af56ddcf1cda3be6b1). We're running Gerrit 2.11.x right now and probably won't upgrade for awhile. And I'm also not sure if our Gerrit admins will support the `download-commands` plugin.

So overall, fetching a single commit sounds nice but will probably not be possible for us for awhile, if at all.

sj26:
Right. So we circle around to your original suggestion to make the refspec overridable. And it appears git fetch will ignore an empty refspec. The only piece missing from your diff, then, is adding the logic to the go version of the bootstrap file (bootstrap.sh will be deprecated).

grosskur:
You're right---I forgot to include `agent/bootstrap.go`. I've updated the diff now---let me know what you think. Thanks!

sj26:
Thanks @grosskur, and sorry for the delay. The git logic here is a bit sticky and we've been trying to balance a few different concerns. We're preparing a new release of the 2.1 stable agent in [2-1-stable](https://github.com/buildkite/agent/commits/2-1-stable). I've proposed some [changes to the bootstrap script](https://github.com/buildkite/agent/pull/243) for this branch which would give you some more flexibility. Then we could forward port it into the bleeding edge master agent, too.

sj26:
Hiya @grosskur, I've integrated this into both the 2-1-stable agents and on master. Give it a go?

grosskur:
Thanks! I'll try out the latest master and let you know if I see any issues.

PR #238, opened 2016-02-12
vendor: Move files to work with GO15VENDOREXPERIMENT=1

grosskur:
Go 1.5 introduces support for an experimental `GO15VENDOREXPERIMENT` environment variable that tells Go to automatically look in the `vendor/` subdirectory for dependencies.

This doesn't work with the current directory structure because the dependencies are in `vendor/src/`, rather than `vendor/`. So move them into `vendor/` and update two import paths

    github.com/mitchellh/go-homedir
    github.com/mattn/go-shellwords

so that Go can find them properly when `GO15VENDOREXPERIMENT=1`.

Also update `README.md`.


keithpitt:
Nice! I've been meaning to do this exact thing for a while now!

grosskur:
Cool! BTW, the `vendor/` stuff is now enabled by default in Go 1.6, which was [released today](https://blog.golang.org/go1.6).

PR #240, opened 2016-02-18
bootstrap: Make "git clean" behavior configurable

### LINKED ISSUE #173: Ability to change `git clean` behaviour to also include `-x`

grosskur:
The default `git clean` flags are `-fdq`. It's useful to be able to
configure this. For example, `-x` can be used to remove gitignore'd
files. Or `-e` can be used to exclude specific directories.

Introduce an environment variable `BUILDKITE_GIT_CLEAN_FLAGS` to
control this behavior. This can then be set in an `environment` hook.

Closes #173.

### DIFF
--- a/agent/bootstrap.go
+++ b/agent/bootstrap.go
@@ -65,6 +65,9 @@ type Bootstrap struct {
 	// Should the bootstrap remove an existing checkout before running the job
 	CleanCheckout bool
 
+	// Flags to pass to "git clean" command
+	GitCleanFlags string
+
 	// Whether or not to run the hooks/commands in a PTY
 	RunInPty bool
 
@@ -608,6 +611,7 @@ func (b *Bootstrap) pluginHookExists(plugins []*Plugin, name string) bool {
 func (b *Bootstrap) applyEnvironmentConfigChanges() {
 	artifactPathsChanged := false
 	artifactUploadDestinationChanged := false
+	gitCleanFlagsChanged := false
 
 	if b.env.Exists("BUILDKITE_ARTIFACT_PATHS") {
 		envArifactPaths := b.env.Get("BUILDKITE_ARTIFACT_PATHS")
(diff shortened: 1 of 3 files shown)


sj26:
I like this idea, but I feel like maybe we could simplify it a bit and make it a "clean everything including ignored files" option or something, instead of passing raw command line arguments. It might be a little easier conceptually. I'm not sure passing raw arguments gives too much more control. What do you think?

sj26:
I think we also want to try and keep the agent as agnostic about the actual version control system as possible, too. So perhaps we could generalise to `BUILDKITE_CLEAN_IGNORED="true"` or something?

keithpitt:
Yup, @sj26 that was my plan. I think that `-x` will actually become the default in v3.0, but it'll be an option until then. I think if peeps want to have files persist when `-x` is the default, they'll just need to write some code to move them out at the end of the job, and move them back in at the beginning (or use the artifacts system).

So in terms of the option, I think `BUILDKITE_GIT_CLEAN_IGNORED=true` (a mix of both) is probably the setting we should go with for the rest of 2.x, and then remove it in 3.0, and make the included `-x` the default.

keithpitt:
Also, we could probably add another hook `prepare` or something, and it's job is to setup the repo for the job. A user could also just override the `git clean` behaviour in that hook.

grosskur:
I definitely understand the desire to make this flag agnostic to the source control system. For background, the problem I'm really trying to solve is this: How can I stash away some files across builds in a cache directory? For example, a `ccache` directory or files downloaded with `pip`.

Some build environments like Heroku provide the build with its own [CACHE_DIR](https://devcenter.heroku.com/articles/buildpack-api#caching). I couldn't figure out a good way to do this in Buildkite, so I've been writing these files to a `.cache` directory inside the build tree and setting `BUILDKITE_GIT_CLEAN_FLAGS=-ffdqxe.cache` to preseve them.

Perhaps what I should do instead is create a separate set of cache directories outside the build directories...

toolmantim:
I just added https://github.com/buildkite/agent/pull/264 — what do you think?

toolmantim:
So the hooks weren't great.

The advantage of making the `git clean` flags explicit for this version, is that it forces you to understand what you're doing as well… seeing as it's going to be very different from the default. And that it covers @grosskur’s case.

I'm thinking this PR is the best to merge, and then in agent 3.0 we change the default (which should cover most cases) and then have this for people wanting an explicit flag?

agent/bootstrap.go
```diff
@@ -847,11 +864,11 @@ func (b *Bootstrap) Start() error {
 		}
 
 		// Clean up the repository
-		b.runCommand("git", "clean", "-fdq")
+		b.runCommand("git", "clean", b.env.Get("BUILDKITE_GIT_CLEAN_FLAGS"))
 
 		// Also clean up submodules if we can
 		if b.GitSubmodules {
-			b.runCommand("git", "submodule", "foreach", "--recursive", "git", "clean", "-fdq")
+			b.runCommand("git", "submodule", "foreach", "--recursive", "git", "clean", b.env.Get("BUILDKITE_GIT_CLEAN_FLAGS"))
```
toolmantim: This should be `b.GitCleanFlags` yeah, in case it's changed in a hook?

toolmantim:
Moved to #270

PR #243, opened 2016-02-25
Bootstrap script fetch logic tweaks

sj26:
Change the git fetch and checkout behaviour to be a little more explicit.

There are a few discrete scenarios which are commented. There's also an additional ability to specify your own refspec, as suggested in #237.

I've tested this in several scenarios, with and without clean checkouts:

 * Building a master push using a github webhook
 * Building "master" at "HEAD" using web ui
 * Building a pushed tag with a commit unreachable from any head ref via webhook
 * Building a "proposed" branch triggerd via push
 * Building a "proposed" branch with "HEAD" triggered via web UI
 * Building a pull request via webhook
 * Building a pull request from a fork via webhook
 * Building a weird ref using `BUILDKITE_REFSPEC` in env and commit `FETCH_HEAD` via web UI

Hopefully this will provide more resilient checkout results, and more debug information if things go awry.

This should fix at least these scenarios:

* Fetching a git commit directly by sha then trying to reset hard to a branch we didn't actually fetch
* Fetching tags which don't appear on any head

### DIFF
--- a/templates/bootstrap.sh
+++ b/templates/bootstrap.sh
@@ -227,11 +227,17 @@ else
     buildkite-run "git submodule foreach --recursive git clean -fdq"
   fi
 
+  # If a refspec is provided then use it instead.
+  # i.e. `refs/not/a/head`
+  if [[ -n "${BUILDKITE_REFSPEC:-}" ]]; then
+    buildkite-run "git fetch origin \"$BUILDKITE_REFSPEC\""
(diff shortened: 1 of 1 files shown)


toolmantim:
I just gave this test on fresh checkout of just `HEAD/master` and that worked nicely (with usual warning of a detached head). Then a random commit and an invalid branch, and that worked fine. If you're cool with your tests too, then I'd say this is :+1:

Issue #205, opened 2015-10-22
Pipeline default executables and environment substitution


toolmantim:
Say you had the following `.buildkite/pipeline.template.yml` file:

```yml
steps:
  -
    name: Step 1
    command: echo hola
    agents:
      name: "$BUILDKITE_AGENT_META_DATA_NAME"
  -
    name: Step 2
    command: echo bonjour
    agents:
      name: "$BUILDKITE_AGENT_META_DATA_NAME"
  -
    name: Step 3
    command: echo konnichiwa
    agents:
      name: "$BUILDKITE_AGENT_META_DATA_NAME"
```

and the following executable script in `.buildkite/pipeline`:

```bash
#!/bin/bash

set -eu

# Makes sure all the steps run on this same agent
sed "s/\$BUILDKITE_AGENT_META_DATA_NAME/$BUILDKITE_AGENT_META_DATA_NAME/" pipeline.template.yml
```

it'd be neat if the `buildkite-agent pipeline upload` detected the presence of the executable file `.buildkite/pipeline` and executed it and read in it's STDIN, basically a default/defacto way of doing generated build pipelines the same as the default location for the yml file.

toolmantim:
Also, stupid idea: a default `pipeline.template.yml` location that also-subs environment variables in

keithpitt:
Another idea is to support a {{}} syntax, where ENV variables are read in:

```
  -
    name: Step 3
    command: echo konnichiwa
    agents:
      name: "{{$BUILDKITE_AGENT_META_DATA_NAME}}"
```

lox:
Yeah, I'd be tempted to mirror the docker-compose syntax:
https://github.com/docker/compose/pull/1765

keithpitt:
I was waiting for someone to come along with the perfect answer @lox :) It shall be done!

keithpitt:
@lox I'm trying to find their current docs on the feature, to see if they handled the default value thing. Do you know if it handles default values?

lox:
https://github.com/docker/compose/pull/1765#issuecomment-125617042

lox:
tl;dr No, but they will support POSIX something something syntax: `${HOST_PORT-8080} or ${HOST_PORT:-8080}`

keithpitt:
Yeah cool, this is ace. Will get it into the next release with that syntax, and I'll support the default gear swell.

toolmantim:
🛰

PR #246, opened 2016-02-28
Added support for environment variable substitution in pipeline files

keithpitt:
Added support for POSIX like environment variable substitution. This fixes https://github.com/buildkite/agent/issues/205

We now support:

```bash
$REGULAR_ENV
${ENV_LIKE_THIS}

export ONES_WITH_DEFAULT_IF_NULL="" # this is "set" but it's value is null

${ONES_WITH_DEFAULT_IF_NULL:-this value will show}
${ONES_WITH_DEFAULT_IF_NULL-but this value won't}

${EVEN_REQUIRE_ONES?}
#=> $EVEN_REQUIRE_ONES: not set

export AND_CUSTOM_ERROR_MESSAGES="" # nulls are treated as empty

${AND_CUSTOM_ERROR_MESSAGES?y u no set me}
#=> $AND_CUSTOM_ERROR_MESSAGES: y u no set me
```

:warning: This change introduces a somewhat breaking change. Any customer who already have `$SOMETHING` and `${SOMETHING}` in their pipeline files will need to escape them to `\$SOMETHING` and `\${SOMETHING}`.

We should probably release this in a 2.2? Maybe scrap the idea of the current 2.2? Or do you think that we could get away with just another 2.1 release?

### DIFF
--- a/agent/pipeline_parser.go
+++ b/agent/pipeline_parser.go
@@ -1,0 +1,126 @@
+package agent
+
+import (
+	"fmt"
+	"os"
+	"regexp"
+	"strings"
+)
+
+type PipelineParser struct {
+	Data []byte
+}
+
+var variablesWithBracketsRegex = regexp.MustCompile(`([\\\$]?\$\{(.+[^\\])})`)
+var variablesWithNoBracketsRegex = regexp.MustCompile(`([\\\$]?\$[a-zA-Z0-9_]+)`)
+
+func (p PipelineParser) Parse() (parsed []byte, err error) {
+	// Do a parse and handle ENV variables with the ${} syntax, i.e. ${FOO}
+	parsed = variablesWithBracketsRegex.ReplaceAllFunc(p.Data, func(part []byte) []byte {
+		v := string(part[:])
+
+		if !p.isPrefixedWithEscapeSequence(v) && err == nil {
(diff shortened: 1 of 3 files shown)


sj26:
Yeah this feels like at least a minor version change. Maybe this should be 2.2.

There's a lot going on in the current master (2.2) release. Perhaps it should be a 3.0 with plugins and go bootstrap?

keithpitt:
Yeah, I think you're right. We should push out a new 2.2 stable with plugins and bootstrap.go removed. Then we can push this functionality into 2.2. Maybe we make a 2.2 branch off our 2.1 now, cherry pick some bits into it, and add this?

toolmantim:
Version 2.2 is master yeah? If so we can just merge… 2.1 releases happen off 2-1-stable I believe

keithpitt:
Ah yes!

keithpitt:
:rocket:

sj26:
I've been merging new PRs into master, then cherry-picking to 2-1-stable if applicable.

toolmantim:
Yep, I just did the same. Turns out you can do a whole list of commit ids on `git cherry-pick`

We had a third-party one against 2-1-stable so I just did it in reverse for that… but the policy should be to base PRs off master unless there's a good reason to do so.

PR #247, opened 2016-02-29
Read BUILDKITE_S3_ACCESS_URL and use to override URL of artifact sent…

### LINKED ISSUE #236: Artifacts uploading to a custom S3 bucket should have SSL enabled by default
Currently the `http://` is hardcoded here: https://github.com/buildkite/agent/blob/master/agent/s3_uploader.go#L65

We should either allow SSL to be toggled, or just change it to `https`

daveoxley:
Modification to accept environment variable BUILDKITE_S3_ACCESS_URL so that artifacts uploaded to a custom S3 bucket can be linked from buildkite using the base URL specified in that environment variable instead of http://{bucketName}.s3.amazonaws.com.

This allows static web hosting to be turned off on the bucket and instead serving artifacts from http://www.s3auth.com/ or using https://github.com/pottava/aws-s3-proxy so that artifacts can be more securely restricted than just be IP address. This will also fix #236 by specifying BUILDKITE_S3_ACCESS_URL as https://{bucketName}.s3.amazonaws.com.

### DIFF
--- a/agent/s3_uploader.go
+++ b/agent/s3_uploader.go
@@ -62,7 +62,11 @@ func (u *S3Uploader) Setup(destination string, debugHTTP bool) error {
 }
 
 func (u *S3Uploader) URL(artifact *api.Artifact) string {
-	url, _ := url.Parse("http://" + u.bucketName() + ".s3.amazonaws.com")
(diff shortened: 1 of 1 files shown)


daveoxley:
Had previously targeted pull request at master but had meant to target 2.1-stable.

PR #248, opened 2016-02-29
Bootstrap tweaks (on master)

sj26:
This applies the tweaks from #243 on master to both the bootstrap script and the go boostrap. It also includes custom refspecs from #237 which are related.

Here are some examples:

screenshot 2016-02-29 16 00 53

screenshot 2016-02-29 16 01 32

screenshot 2016-02-29 16 02 27

screenshot 2016-02-29 16 02 59

screenshot 2016-02-29 16 04 03

screenshot 2016-02-29 16 05 42

/cc @keithpitt

### DIFF
--- a/agent/bootstrap.go
+++ b/agent/bootstrap.go
@@ -41,6 +41,9 @@ type Bootstrap struct {
 	// The tag of the job commit
 	Tag string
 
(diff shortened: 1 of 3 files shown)


keithpitt:
:+1:

PR #250, opened 2016-03-03
Fix git fetch tags

sj26:
Prior to git 1.9.0, `git fetch --tags` would only fetch tags, not in addition to the default refspec. So replace it with an explicit set of refspecs.

Also, turn on verbose fetching for easier diagnostics.


toolmantim:
Ace :banana:s!

PR #251, opened 2016-03-07
Install arch flag

underscorediscovery:
This PR proposes :

- BUILDKITE_INSTALL_ARCH be added for explicitly choosing an arch on install.sh
- Using `uname -m` instead of `uname -p` which can report false info.

Each commit includes detailed descriptions, I can split the commits to separate PR's if preferred.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -20,6 +20,7 @@ echo -e "\033[33m
 
 echo -e "Finding latest release..."
 
+ARCH="386"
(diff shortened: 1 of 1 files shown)


underscorediscovery:
OS X uname comparison:
screen 2016-03-07 at 5 47 19 pm

With both commits:
screen 2016-03-07 at 6 09 43 pm

toolmantim:
Thanks @underscorediscovery!

PR #252, opened 2016-03-07
Add BUILDKITE_DOCKER_COMPOSE_LEAVE_VOLUMES

daveslutzkin:
At present the agent is eager in deleting volumes attached to containers, which is the right thing to do in general to save space.

For speed reasons, we use a persistent Named Volume (holding ruby gems) and that volume shouldn't be deleted. But docker doesn't natively provide a way to say that it's persistent. So we need a flag to tell the agent not to ask docker to delete volumes. Hence `BUILDKITE_DOCKER_COMPOSE_LEAVE_VOLUMES`.

In general this is dangerous because it's bad for disk space, so those using the flag need to acknowledge the risks.


toolmantim:
Ace! This is great.

PR #254, opened 2016-03-09
Log chunk as request body

sj26:
This uses much less memory on the receiving end than processing multipart form uploads because rails. We're using http content-encoding for gzip compression, too, so can eventually transparently offload into a terminator instead of the app.


toolmantim:
:thumbsup:

Let's do this.

PR #255, opened 2016-03-09
Static link

sj26:
This doesn't seem to have much effect for windows or mac, but it stops linux from dynamically linking libc so we can restore Debian 7 compatibility. It doesn't seem to add a huge about of weight to the binary.

If all good I'll backport to 2-1-stable.


toolmantim:
:+1: I think this is a good idea

sj26:
## Before

sj26:
### Before

screenshot 2016-03-09 22 19 44

### After

screenshot 2016-03-09 22 19 57

keithpitt:
Sounds great to me! :+1:

PR #259, opened 2016-03-10
Increase timeouts

toolmantim:
This increases our HTTP response timeout to 1 minute, and decreases retry frequency for API calls from once every 1 second to once every 5 seconds.

This means the agent will now wait 10 seconds for a http connection, 60 seconds for a response, and then it'll wait 5 seconds before retrying. For network situations, or downtime, this means things will attempting to retry every 5 seconds instead of every second.

### DIFF
--- a/agent/agent_pool.go
+++ b/agent/agent_pool.go
@@ -154,7 +154,8 @@ func (r *AgentPool) RegisterAgent(agent *api.Agent) (*api.Agent, error) {
 		return err
 	}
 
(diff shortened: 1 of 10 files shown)


toolmantim:
@keithpitt given you chose the initial values, I'll need your eyes on this. I don't see many user-facing problems with the increases?

toolmantim:
Once this is merged I'll also backport to stable

keithpitt:
:+1:

PR #264, opened 2016-03-17
Add git-clean hook support

toolmantim:
Based on #240 this adds support for a `git-clean` hook.

Is this better than simply supporting a `BUILDKITE_GIT_CLEAN_FLAGS` environment variable?

### DIFF
--- a/agent/bootstrap.go
+++ b/agent/bootstrap.go
@@ -849,6 +849,15 @@ func (b *Bootstrap) Start() error {
 			b.runCommand("git", "clone", "-v", "--", b.Repository, ".")
 		}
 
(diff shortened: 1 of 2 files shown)


keithpitt:
Perhaps a less git specific name, like just `clean` or `prepare`?

toolmantim:
Hmm… it's inside the standard git checkout code… so it's expected to be git specific?

toolmantim:
That's why I think the environment variable might be better…

keithpitt:
Yeah, I think I prefer just the ENV var. Close this?

toolmantim:
Yep!

PR #270, opened 2016-03-17
Add BUILDKITE_GIT_CLEAN_FLAGS

toolmantim:
:rocket: #240 

> The default `git clean` flags are `-fdq`. It's useful to be able to configure this. For example, `-x` can be used to remove gitignore'd files. Or `-e` can be used to exclude specific directories.
>
> Introduce an environment variable `BUILDKITE_GIT_CLEAN_FLAGS` to control this behavior. This can then be set in an `environment` hook.

### DIFF
--- a/agent/bootstrap.go
+++ b/agent/bootstrap.go
@@ -867,11 +867,11 @@ func (b *Bootstrap) Start() error {
 		}
 
 		// Clean up the repository
-		b.runCommand("git", "clean", b.env.Get("BUILDKITE_GIT_CLEAN_FLAGS"))
+		b.runCommand("git", "clean", b.GitCleanFlags)
 
 		// Also clean up submodules if we can
 		if b.GitSubmodules {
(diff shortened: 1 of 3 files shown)


toolmantim:
I just tested it with the following environment hook, and it works a treat:

```
#!/bin/bash

set -eu

export BUILDKITE_GIT_CLEAN_FLAGS="-fdx"
```

And in the build output:

```
$ git clean "-fdx"
Removing .env
Removing buildkite-script-8c8a6870-349c-4af7-a5c7-49b79f2f67fc
```

`.env` was a git ignored file :+1:

grosskur:
Nice, thanks for merging! :+1:

toolmantim:
Do you all think we should remove the `-q` by default? It's all under the "Preparing build header" and I think it might help people know what's going on. If you want it to be quiet you can now change it too, but I'd assume most people would want to see more clearly what's getting cleaned.

keithpitt:
I think @sj26 made something else not quiet recently?

PR #276, opened 2016-03-18
Build debs once, publish in stages

toolmantim:
@sj26 & @toolmantim paired on fixing our debian archives becoming corrupted. Turns out we were building the debian packages multiple times but they go in the same place on the archive so were overwriting each other and causing the wrong file sizes/hashes. So we build them once instead, then publish the same packages into each subsequent archive: experimental, unstable, stable.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -17,6 +17,16 @@ steps:
   - wait
 
   -
+    name: ":debian:"
+    command: "scripts/build-debian-packages.sh"
+    artifact_paths: "deb/**/*"
+    branches: "master"
+    agents:
+      queue: "deploy"
+
+  - wait
+
+  -
     name: ":redhat: experimental"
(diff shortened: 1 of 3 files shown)


keithpitt:
Nice! I looked through the current building scripts (trying to figure out if there was a reason why I didn't build once before) and I couldn't find anything.

A long time ago the `--version` call to the agent used to return different things, depending on which Debian channel you downloaded from (stable would say: `1.2` where as unstable would have `1.2.523` (here `523` would be the build number). Which is why I think I did the builds per `$CODENAME` - But later on I changed the `--version` command to just always return build information.

So if this works, :rocket:

toolmantim:
Ahh… I was trying to remember what we used to do about different versions! Because I was mentioning to @sj26 that I thought it was diff for diff stages.

toolmantim:
There was a bug in the artifact download syntax (damn that syntax!). Should be fixed, will merge to master to give it a go

PR #278, opened 2016-03-18
Make sure we initialize the GitCleanFlags

toolmantim:
Unfortunately #270 was broken in that it didn't initialize `GitCleanFlags`, so `git clean` would error with:

```
$ git clean
fatal: clean.requireForce defaults to true and neither -i, -n, nor -f given; refusing to clean
```

This fixes it so it works as expected.

### DIFF
--- a/clicommand/bootstrap.go
+++ b/clicommand/bootstrap.go
@@ -232,6 +232,7 @@ var BootstrapCommand = cli.Command{
 			Plugins:                      cfg.Plugins,
 			GitSubmodules:                cfg.GitSubmodules,
 			PullRequest:                  cfg.PullRequest,
+			GitCleanFlags:								cfg.GitCleanFlags,
 			AgentName:                    cfg.AgentName,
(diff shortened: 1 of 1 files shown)


clicommand/bootstrap.go
```diff
@@ -232,6 +232,7 @@ var BootstrapCommand = cli.Command{
 			Plugins:                      cfg.Plugins,
 			GitSubmodules:                cfg.GitSubmodules,
 			PullRequest:                  cfg.PullRequest,
+			GitCleanFlags:								cfg.GitCleanFlags,
```
keithpitt: Crazy indentation!!

toolmantim: Ugh, will fix in master…

toolmantim: Fixed in 805b277

keithpitt: :+1:

PR #282, opened 2016-03-31
Fix bug I introduced with (BK)_(DC)_LEAVE_VOLUMES

daveslutzkin:
No idea what I was thinking with the initial version of this change, given the semantics were inversed and it was very broken on non-bash shells.


toolmantim:
I TRUSTED YOU :)

Ok… so you’ve triple tested this yeah?

toolmantim:
Ok, this is all good

PR #284, opened 2016-03-31
Strip ANSI colors when checking to see if a line is a header

keithpitt:
Backports https://github.com/buildkite/agent/pull/279


toolmantim:
👍

PR #285, opened 2016-04-01
Build only the specified Docker image

caiofbpa:
Some people have more buildable services in their docker-compose.yml, and building images other than the specified in `BUILDKITE_DOCKER_COMPOSE_CONTAINER` takes unnecessary build time.

### DIFF
--- a/templates/bootstrap.sh
+++ b/templates/bootstrap.sh
@@ -436,7 +436,11 @@ else
     # Build the Docker images using Compose, namespaced to the job
     echo "~~~ Building Docker images"
 
-    buildkite-run "$COMPOSE_COMMAND build $BUILDKITE_DOCKER_COMPOSE_CONTAINER"
+    if [[ ! -z "${BUILDKITE_DOCKER_COMPOSE_BUILD_ALL:-}" ]] && [[ "$BUILDKITE_DOCKER_COMPOSE_BUILD_ALL" != "" ]]; then
+      buildkite-run "$COMPOSE_COMMAND build"
+    else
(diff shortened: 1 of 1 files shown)


toolmantim:
Oh, why didn't I do this in the first place? Thanks for the PR!

I doubt there’s many people relying on the existing behaviour, but perhaps we should provide an option for them?

caiofbpa:
I added the variable `BUILDKITE_DOCKER_COMPOSE_BUILD_ALL`, which when set to `true` triggers the old `docker-compose build` command. Is that ok?

I'm no shell master, let me know if there are any obvious improvements I should make so we can merge this. :smile:

templates/bootstrap.sh
```diff
@@ -436,7 +436,11 @@ else
     # Build the Docker images using Compose, namespaced to the job
     echo "~~~ Building Docker images"
 
-    buildkite-run "$COMPOSE_COMMAND build"
+    if [[ "${BUILDKITE_DOCKER_COMPOSE_BUILD_ALL:-false}" == "true" ]]; then
```
caarlos0: you can safely remove the `:-false` and leave only one `=` sign if you want :)

caiofbpa: Thanks for the ProTip:tm:, @caarlos0! The `:-false` and `==` seem to follow the coding style present in the file, so I'll wait for @toolmantim to have a final say about this.

lox:
I always wondered why this was the case too. 👍

toolmantim:
Awesome, thanks for the update! We'll give it a test and merge

caiofbpa:
Any updates on this?

toolmantim:
Thanks, just tasted! 👍

PR #290, opened 2016-04-17
Fix up init scripts for CentOS-6 and Amazon Linux

jnewbigin:
Tested on Amazon Linux, CentOS-6 and CentOS-7 (systemd). Have not tested upstart version != 0.6.5

### DIFF
--- a/packaging/linux/scripts/after-install-and-upgrade.sh
+++ b/packaging/linux/scripts/after-install-and-upgrade.sh
@@ -58,9 +58,14 @@ BK_IS_UBUNTU_14_10=$?
 command -v initctl > /dev/null
 BK_UPSTART_EXISTS=$?
 
-# Check if the system is Amazon linux which has an old and broken upstart.
-[ -f /etc/system-release ] && grep -qi "Amazon Linux" /etc/system-release
-BK_IS_AMAZON_LINUX=$?
+# Check if upstart is version 0.6.5 as seen on Amazon linux, RHEL6 & CentOS-6
+BK_UPSTART_TOO_OLD=0
+if [ $BK_UPSTART_EXISTS -eq 0 ]; then
+  BK_UPSTART_VERSION="$(initctl --version | awk 'BEGIN{FS="[ ()]"} NR==1{print $4}')"
+  if [ "$BK_UPSTART_VERSION" = "0.6.5" ]; then
+    BK_UPSTART_TOO_OLD=1
+  fi
+fi
 
 # Install the relevant system process
 if [ $BK_SYSTEMD_EXISTS -eq 0 ] && [ $BK_IS_UBUNTU_14_10 -eq 1 ]; then
@@ -69,7 +74,7 @@ if [ $BK_SYSTEMD_EXISTS -eq 0 ] && [ $BK_IS_UBUNTU_14_10 -eq 1 ]; then
   fi
 
   START_COMMAND="sudo systemctl enable buildkite-agent && sudo systemctl start buildkite-agent"
-elif [ $BK_UPSTART_EXISTS -eq 0 ] && [ $BK_IS_AMAZON_LINUX -eq 1 ]; then
+elif [ $BK_UPSTART_EXISTS -eq 0 ] && [ $BK_UPSTART_TOO_OLD -eq 0 ]; then
   if [ ! -f /etc/init/buildkite-agent.conf ]; then
     # If the system has the old .env file, install the old upstart script, and
     # let them know they should upgrade. Because the upstart script is no
--- a/packaging/linux/scripts/after-install-and-upgrade.sh
+++ b/packaging/linux/scripts/after-install-and-upgrade.sh
@@ -121,6 +121,7 @@ elif [ $BK_UPSTART_EXISTS -eq 0 ] && [ $BK_UPSTART_TOO_OLD -eq 0 ]; then
 elif [ -d /etc/init.d ]; then
   if [ ! -f /etc/init.d/buildkite-agent ]; then
     cp /usr/share/buildkite-agent/lsb/buildkite-agent.sh /etc/init.d/buildkite-agent
+    command -v chkconfig > /dev/null && chkconfig --add buildkite-agent
   fi
 
   START_COMMAND="sudo /etc/init.d/buildkite-agent start"
--- a/packaging/linux/scripts/after-remove.sh
+++ b/packaging/linux/scripts/after-remove.sh
@@ -1,8 +1,10 @@
 # Remove the system service we installed
 if command -v systemctl > /dev/null; then
   rm -f /lib/systemd/system/buildkite-agent.service
-elif [ -d /etc/init ]; then
+fi
+if [ -f /etc/init/buildkite-agent.conf ]; then
   rm -f /etc/init/buildkite-agent.conf
-elif [ -d /etc/init.d ]; then
+fi
+if [ -f /etc/init.d/buildkite-agent ]; then
   rm -f /etc/init.d/buildkite-agent
 fi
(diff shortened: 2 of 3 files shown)


jnewbigin:
Not sure which branch you want this on either...

keithpitt:
Thanks for this @jnewbigin! Will check it out today!

packaging/linux/scripts/after-install-and-upgrade.sh
```diff
@@ -58,9 +58,14 @@ BK_IS_UBUNTU_14_10=$?
...
 BK_UPSTART_EXISTS=$?
 
-# Check if the system is Amazon linux which has an old and broken upstart.
-[ -f /etc/system-release ] && grep -qi "Amazon Linux" /etc/system-release
-BK_IS_AMAZON_LINUX=$?
+# Check if upstart is version 0.6.5 as seen on Amazon linux, RHEL6 & CentOS-6
+BK_UPSTART_TOO_OLD=0
+if [ $BK_UPSTART_EXISTS -eq 0 ]; then
+  BK_UPSTART_VERSION="$(initctl --version | awk 'BEGIN{FS="[ ()]"} NR==1{print $4}')"
+  if [ "$BK_UPSTART_VERSION" = "0.6.5" ]; then
```
keithpitt: Here's a question, what happens if the version is `0.6.4`?

jnewbigin: There is no version 0.6.4 :-)
http://upstart.ubuntu.com/cookbook/#releases
I was originally going to do a version check against the required minimum version for you upstart config. Alas, I don't know what version /is/ required. CentOS-6 and Amazon Linux both use exactly 0.6.5.
It may be that we can just look at the major version number but I don't have any machines on upstart 1.x

toolmantim: So we could merge this in, and if it turns out 0.6.5/0.6.6/0.6.7 need it we could update the conditional? And eventually the AWS images won't need the hack…

lox:
```
[ec2-user@ip-10-0-3-26 ~]$ initctl version
init (upstart 0.6.5)
[ec2-user@ip-10-0-3-26 ~]$ cat /etc/issue
Amazon Linux AMI release 2016.03
Kernel \r on an \m
```

toolmantim:
This should be good to merge

toolmantim:
Ok, I just tested this release on Amazon Linux…

```bash
# Install
sudo sh -c 'echo -e "[buildkite-agent]\nname = Buildkite Pty Ltd\nbaseurl = https://yum.buildkite.com/buildkite-agent/experimental/x86_64/\nenabled=1\ngpgcheck=0\npriority=1" > /etc/yum.repos.d/buildkite-agent.repo'
sudo yum install -y buildkite-agent-2.1.10-1095
sudo sed -i "s/xxx/INSERT_AGENT_TOKEN/g" /etc/buildkite-agent/buildkite-agent.cfg
sudo /etc/init.d/buildkite-agent start
```

It doesn't seem to stop the agent gracefully when the system shuts down, which means we went up with a ghost agent:

```bash
# Emulate an instance termination
$ sudo shutdown -r now

# Agent still shows as connected in Buildkite UI
```

Strangely, once the system restarts `init.d` says it's stopped, but it is actually running (and appears in the Buildkite UI):

```bash
$ sudo /etc/init.d/buildkite-agent status
Stopped
$ ps aux | grep buildkite-agent
root      2245  0.0  0.8 184516  4348 ?        S    09:05   0:00 sudo -u buildkite-agent /usr/bin/buildkite-agent start
498       2253  0.0  2.3 321924 11704 ?        Sl   09:05   0:00 buildkite-agent v2.1.10 [idle]           
ec2-user  2505  0.0  0.4 110456  2028 pts/0    S+   09:07   0:00 grep --color=auto buildkite-agent
```

The PID looks odd too:

```bash
$ ls -l /var/run/*.pid | grep buildkite
-rw-r--r-- 1 root    root     5 May  9 09:05 /var/run/S50buildkite-agent.pid
```

And if you start it manually, you end up with two agents running:

```
[ec2-user@ip-172-31-57-59 ~]$ sudo /etc/init.d/buildkite-agent start
Starting buildkite-agent
[ec2-user@ip-172-31-57-59 ~]$ ls -l /var/run/*.pid | grep buildkite
-rw-r--r-- 1 root    root     5 May  9 09:12 /var/run/buildkite-agent.pid
-rw-r--r-- 1 root    root     5 May  9 09:05 /var/run/S50buildkite-agent.pid
```

Am I running the correct init/install commands?

toolmantim:
Some more debug info…

```bash
$ sudo chkconfig --list | head
acpid          	0:off	1:off	2:on	3:on	4:on	5:on	6:off
atd            	0:off	1:off	2:off	3:on	4:on	5:on	6:off
auditd         	0:off	1:off	2:on	3:on	4:on	5:on	6:off
blk-availability	0:off	1:on	2:on	3:on	4:on	5:on	6:off
buildkite-agent	0:off	1:off	2:on	3:on	4:on	5:on	6:off
cgconfig       	0:off	1:off	2:off	3:off	4:off	5:off	6:off
cgred          	0:off	1:off	2:off	3:off	4:off	5:off	6:off
cloud-config   	0:off	1:off	2:on	3:on	4:on	5:on	6:off
cloud-final    	0:off	1:off	2:on	3:on	4:on	5:on	6:off
cloud-init     	0:off	1:off	2:on	3:on	4:on	5:on	6:off
```

```bash
$ head -n 50 /etc/init.d/buildkite-agent 
#!/bin/sh
#
### BEGIN INIT INFO
# Provides:          buildkite-agent
# Required-Start:    $network $local_fs $remote_fs
# Required-Stop:     $remote_fs
# Should-Start:      $named
# Should-Stop:
# Default-Start:     2 3 4 5
# Default-Stop:      0 1 6
# Short-Description: The Buildkite Build Agent
# Description:       The Buildkite Build Agent
### END INIT INFO

user="buildkite-agent"
cmd="/usr/bin/buildkite-agent start"

name=`basename $0`
pid_file="/var/run/$name.pid"
log="/var/log/$name.log"
stderr_log="/var/log/$name.err"

get_pid() {
    cat "$pid_file"
}

is_running() {
    [ -f "$pid_file" ] && ps `get_pid` > /dev/null 2>&1
}

case "$1" in
    start)
    if is_running; then
        echo "Already started"
    else
        echo "Starting $name"
        sudo -u "$user" $cmd >>"$log" 2>&1 &
        echo $! > "$pid_file"
        if ! is_running; then
            echo "Unable to start, see $log"
            exit 1
        fi
    fi
    ;;
    stop)
    if is_running; then
        echo -n "Stopping $name.."
        kill `get_pid`
        for i in {1..10}
        do
```

jnewbigin:
I'll have a look

toolmantim:
@jnewbigin thanks!

jnewbigin:
Hi, I have identified at least one issue.

(First I had to edit /etc/buildkite-agent/buildkite-agent.cfg and enable the bootstrap-script script before the agent would start (looks like a v2.2 configfile for a v2.1 build))

There is a bug in the init script where name= does not dereference the symlink which causes pid file problems (multiple starts & no stop). I replaced with
name=$(basename $(readlink -f $0))

After doing that, the agent starts & stops fine with these commands
service buildkite-agent [start|stop]

Still not sure why the ghost appears after a shutdown/reboot. I suspect a timing issue where the network goes away before the agent has shutdown. Not sure why but I'll try and dig into it

jnewbigin:
The init script also needs to touch touch /var/lock/subsys/$name or the shutdown script will not be called.
I can create a PR in the morning.

PR #293, opened 2016-04-21
Make sanitised agent name alpha-numeric only

rprieto:
A build agent's working dir is `/usr/local/var/buildkite-agent/builds/{{name}}`.

This change makes it a little safer by making sure there are no weird characters, especially emojis with sneaky `:` characters. Docker-compose doesn't seem to like them too much when mounting volumes.

| Agent | Folder |
|-------|--------|
| hello world | /usr/local/var/buildkite-agent/builds/hello_world |
| :taco: builder | /usr/local/var/buildkite-agent/builds/_taco_builder |

On the downside this will change many working dirs, so agents will need to re-download every repo... thoughts?


keithpitt:
Thanks for this! We may need to be a smarter about this change. We should check the existence of the older named folders, then move them and use the new name. The reason being that by changing how they're named, after people upgrade their agent, they'll get hit with a slow build because it'll clone to a new folder (also leave the old one to go stale and take up space). I hope that made sense :)

I'm happy to make this change if you want!! Or you can if you're feeling adventurous :)

Thanks again for the PR!!

rprieto:
Yep made total sense, check+move sounds like a good option :+1: I'll give it a shot tomorrow.
In the meantime I added this to my `pre-checkout` hook to make `docker-compose` happy.

rprieto:
What do you think the behaviour should be if there's conflict, e.g. the folder exists with the old name, but the new name is already taken too?

keithpitt:
@rprieto that's a good question! What do you think?

keithpitt:
We could `rm -rf` the old one and move on perhaps?

keithpitt:
Or maybe we just use the new folder and leave the old one alone?

rprieto:
What does the existing agent do in case of conflict, e.g. if there are 2 agents with the same name? BuildKite doesn't seem to reject same-name agents, but I assume it would cause problems if they run on the same server.

keithpitt:
Yeah... So currently if you have 2 agents with the same name, bad things happen. We try to give people tools to avoid that (like `--name "foo-agent-%n"` which will assign it a unique number) but clashes sometimes still happen. That's a future problem though I think :)

keithpitt:
So I think the safest bet is to just use the newer named folder if it exists, and leave the old one alone? Just because the likelihood of this is happening is pretty rare I think? So the code would be something like this...

```
if !folder_with_new_name.exists?
  if folder_with_old_name.exists?
    mv folder_with_old_name folder_with_new_name
  else
    make_new_folder
  end
end

test party!
```

toolmantim:
This is probably ok to merge as is for v3.0 (this PR is against master).

Okay to merge @keithpitt?

keithpitt:
@toolmantim yeah, after thinking about the edge case, I think it's fine to merge into master as-is.

toolmantim:
@keithpitt cool, I reckon so too!

rprieto:
Thanks for that! Sorry I didn't have a chance to look at the smart rename yet. So are we 👍 about the fact agents will need to clone from scratch again?

toolmantim:
@rprieto yep! If you wanted to backport this to 2-1-stable branch to go out in a 2.1 point release, we'd need to make sure they wouldn't re-clone.

Actually… I just released the bootstrap.sh doesn't apply to master, it has the bootstrap in go. So I just merged this into oblivion. We're removing this in #303 to avoid this confusion.

So for master we need to update:
https://github.com/buildkite/agent/blob/3c6d8ccca71fd918de9272c004187c4a66452ea0/agent/bootstrap.go#L674-L676

@rprieto We can fix the golang version. Did you want this in 2-1-stable so you can emoji on stable?

toolmantim:
here's the change in master: https://github.com/buildkite/agent/pull/310

rprieto:
Thanks that made sense. I made that change to our local `pre-checkout` so we're OK for now. Just saw your new PR on the golang version, looks great!

PR #301, opened 2016-05-04
Add pipefail to generated scripts

toolmantim:
A nice catch by @npaufler

We should consider adding -u as well


toolmantim:
I just tested the master golang version, and it fails on both error'ing commands failed pipes, no need to port these to master

PR #303, opened 2016-05-09
Remove the old bootstrap files

toolmantim:
👋


keithpitt:
Excellent! Should we update the default branch as well?

toolmantim:
So instead of changing the default branch… I thought it'd be better to just update the Readme to point out "This is for the upcoming 3.0 version of the agent. See [2-1-stable](2-1-stable) for the maintenance branch for the current stable release." type thing. Put it as the first para of the Readme. We can use that moving forward to point people to where they should create pull requests for the diff versions. And will encourage people to submit PRs for master too.

keithpitt:
That works as well!

I actually removed the bootstrap before, but some peeps actually download the bootstrap.sh in scripts they have for their agent setup automation gear, so I put it back in. The links they use would break, but I suppose it needs to happen eventually!

toolmantim:
Cool… just updated the Readme. It'd be nice if GitHub had a better way of doing this… so the Readme doesn't have to get changed every time there's a new branch. We could perhaps point to the wiki index, and include that information there? That way each Readme doesn't have it's version number embedded, and it's easier to tag and branch etc diff versions.

Or, how about we point to our own docs… a simple https://buildkite.com/docs/agent/developing page? That can mention Github, pull requests, and list the maintenance branches. And we can update the readmes to always just point to that page.

keithpitt:
I think what you have should be fine!

PR #304, opened 2016-05-09
Ensure git clean removes gitignore'd files

toolmantim:
The new default is to clean all files, including those in the `.gitignore`, using `git clean -fdqx`

This can be overridden using the config var or `BUILDKITE_GIT_CLEAN_FLAGS`


keithpitt:
🚀

PR #305, opened 2016-05-09
CentOS-6/Amazon Linux initscript

jnewbigin:
Fix .pid and lock file handling

### DIFF
--- a/packaging/linux/root/usr/share/buildkite-agent/lsb/buildkite-agent.sh
+++ b/packaging/linux/root/usr/share/buildkite-agent/lsb/buildkite-agent.sh
@@ -41,7 +41,7 @@ case "$1" in
             echo "Unable to start, see $log"
             exit 1
         fi
-	touch "$lock_file"
+        touch "$lock_file"
     fi
     ;;
     stop)
@@ -67,7 +67,7 @@ case "$1" in
             if [ -f "$pid_file" ]; then
                 rm "$pid_file"
             fi
-	    rm -f "$lock_file"
+        rm -f "$lock_file"
         fi
     else
         echo "Not running"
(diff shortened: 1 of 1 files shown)


packaging/linux/root/usr/share/buildkite-agent/lsb/buildkite-agent.sh
```diff
@@ -15,10 +15,11 @@
 user="buildkite-agent"
 cmd="/usr/bin/buildkite-agent start"
 
-name=`basename $0`
-pid_file="/var/run/$name.pid"
-log="/var/log/$name.log"
-stderr_log="/var/log/$name.err"
+name=$(basename "$(readlink -f "$0")")
```
toolmantim: Is there any reason we couldn't just hardcode "buildkite-agent" here?

toolmantim:
Thanks so much! I'll merge this in for testing.

packaging/linux/root/usr/share/buildkite-agent/lsb/buildkite-agent.sh
jnewbigin: Using basename does allow you to run multiple instances with (a bit) less fuss:
cp /etc/rc.d/init.d/buildkite-agent  /etc/rc.d/init.d/buildkite-agent-2
chkconfig --add buildkite-agent-2
service buildkite-agent-2 start

toolmantim:
Ok, I just tested and this works a treat. Thanks so much!

packaging/linux/root/usr/share/buildkite-agent/lsb/buildkite-agent.sh
toolmantim: Right right! Yeah… let's leave that in then :)

PR #308, opened 2016-05-10
Use SHA512 for Debian signing

### LINKED ISSUE #295: Issue after updating to ubuntu 16.04
After updating the server we have the agent running on to Ubuntu 16.04
we get the following error:

W: https://apt.buildkite.com/buildkite-agent/dists/stable/Release.gpg:
Signature by key 32A37959C2FA5C3C99EFBC32A79206696452D198 uses weak
digest algorithm (SHA1)

It seems to be related to apt dropping support for SHA1 keys. The same
is supposed to happen with Debian as of January 1st 2017. To my
understanding the fix should be rather straight forward:

https://wiki.debian.org/Teams/Apt/Sha1Removal

boiling down to

The repository owner needs to pass --digest-algo SHA512 or --digest-algo
SHA256 (or another SHA2 algorithm) to gpg when signing the file.
Repositories with DSA keys need to be migrated to RSA first.

Migrating from DSA to RSA is best done by signing the repository with
two keys (old and new one) and shipping the new one to the users. A
relatively safe way to ship the key would be to embed it in the package.
Some months after those changes, it is OK to drop the old key from the
repository and the users machines (if shipped with a package).

toolmantim:
SHA1 is being removed, and is already broken on Ubuntu 16.04:
https://wiki.debian.org/Teams/Apt/Sha1Removal

Fixes #295

Needs backporting to 2-1-stable

### DIFF
--- a/scripts/utils/publish-debian-package.sh
+++ b/scripts/utils/publish-debian-package.sh
@@ -57,7 +57,7 @@ GPG_PASSPHRASE=`openssl aes-256-cbc -k "$GPG_PASSPHRASE_PASSWORD" -in "$GPG_PASS
 bundle exec deb-s3 upload \
   --preserve-versions \
   --sign $GPG_SIGNING_KEY \
-  --gpg-options "\-\-passphrase $GPG_PASSPHRASE" \
+  --gpg-options "\-\-digest-algo SHA512 \-\-passphrase $GPG_PASSPHRASE" \
   --bucket $DEB_S3_BUCKET \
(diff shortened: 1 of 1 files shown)


keithpitt:
As long as the SHA 512 ones still work on older Ubuntus, 🚀

If you didn't check, I can for you?

toolmantim:
Yeah, I have done zero testing on this…

keithpitt:
Testing on old ubuntu now..

PR #310, opened 2016-05-11
Use a sanitized agent name for git checkout

toolmantim:
Ports #293 to the golang bootstrap.

I've opted to keep the case, and not collapse down any leading or duplicate `-` characters to stop unexpected clashes. `my-agent` will be different to `My-Agent` (on Linux). Is that sensible?


rprieto:
👍 from me as little as it counts :) Keeping the case and not collapsing feels less magical, I like it better that way actually.

PR #312, opened 2016-05-11
Use - in place of : in agent names for build checkouts

toolmantim:
Another take on #293 for stable, something a little less aggressive than #310 which is what we do on 3.0. This will only affect people that use `:` in their agent names.

Thoughts @keithpitt?


toolmantim:
Eh, let's leave this out of stable for now…

PR #313, opened 2016-05-11
Add the EC2 instance ID to the automatic metadata

toolmantim:
It's not one of the standard meta-data values… but it'd be super handy for debugging, lining up with cloudwatch logs, etc etc

### DIFF
--- a/agent/ec2_tags.go
+++ b/agent/ec2_tags.go
@@ -41,5 +41,8 @@ func (e EC2Tags) Get() (map[string]string, error) {
 		tags[tag.Key] = tag.Value
 	}
 
+	// We set this manually, it's not a standard tag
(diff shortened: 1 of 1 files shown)


lox:
👍

lox:
Would be great to get this exposed to whatever % formatter that agent name uses.

keithpitt:
@lox what about this?

```
buildkite-agent start --name "agent-$(wget -q -O - http://instance-data/latest/meta-data/instance-id)"
```

:)

toolmantim:
We're going to add a `--meta-data-ec2` command line flag to grab the few things from meta-data that are important. They don't need a role, unlike the `--meta-data-ec2-tags` flag.

For now we can just do it in the stack directly (buildkite/buildkite-aws-stack#67)

PR #317, opened 2016-05-12
Output the build directory in the log

toolmantim:
Also renames folder -> directory for consistency

### DIFF
--- a/agent/bootstrap.go
+++ b/agent/bootstrap.go
@@ -86,7 +86,7 @@ type Bootstrap struct {
 	// Path to the global hooks
 	HooksPath string
 
-	// Path to the plugins folder
+	// Path to the plugins directory
 	PluginsPath string
 
 	// Paths to automatically upload as artifacts when the build finishes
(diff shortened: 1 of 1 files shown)


agent/bootstrap.go
```diff
@@ -844,7 +844,7 @@ func (b *Bootstrap) Start() error {
 	}
 
 	// Switch the internal wd to it
-	commentf("Switching working directroy to build folder")
+	commentf("Switching working directory build directory")
```
keithpitt: Maybe it'd be better if we put it here? Since it's a header, it'd be visible each build, and for the most part that information isn't relevant when scanning down the headers.

```
commentf("Switching working directory to: %s", b.env.Get("BUILDKITE_BUILD_CHECKOUT_PATH"))
```

toolmantim: Hah, that's where I originally had it. Will change!

toolmantim: kk fixed

keithpitt:
👍

Issue #288, opened 2016-04-14
buildkite-agent meta-data set crash


daveoxley:
Since about 1:30pm 14/4 buildkite-agent has been crashing at the end of our builds while doing `buildkite-agent meta-data set`.

https://buildkite.com/intelematics/android-module-ubi/builds/199#ca493f25-c730-4fc5-8043-976746821594

```
fatal error: unexpected signal during runtime execution
[signal 0xb code=0x1 addr=0x0 pc=0x7c2e6c]
 
runtime stack:
runtime.throw(0xaa5e40, 0x2a)
	/usr/local/go/src/runtime/panic.go:530 +0x90
runtime.sigpanic()
	/usr/local/go/src/runtime/sigpanic_unix.go:12 +0x5a
 
goroutine 1 [syscall, locked to thread]:
runtime.cgocall(0x7c2db0, 0xc82004bdf8, 0xc800000000)
	/usr/local/go/src/runtime/cgocall.go:123 +0x11b fp=0xc82004bda8 sp=0xc82004bd78
github.com/ErikDubbelboer/gspt._Cfunc_spt_init2(0x5, 0x130f740)
	github.com/ErikDubbelboer/gspt/_obj/_cgo_gotypes.go:89 +0x3a fp=0xc82004bdf8 sp=0xc82004bda8
github.com/ErikDubbelboer/gspt.init.1()
	/go/src/github.com/buildkite/agent/vendor/src/github.com/ErikDubbelboer/gspt/gspt.go:53 +0x427 fp=0xc82004bf18 sp=0xc82004bdf8
github.com/ErikDubbelboer/gspt.init()
	github.com/ErikDubbelboer/gspt/_obj/_cgo_import.go:1 +0x4a fp=0xc82004bf20 sp=0xc82004bf18
github.com/buildkite/agent/proctitle.init()
	/go/src/github.com/buildkite/agent/proctitle/proctitle_linux_amd64.go:21 +0x40 fp=0xc82004bf28 sp=0xc82004bf20
github.com/buildkite/agent/agent.init()
	/go/src/github.com/buildkite/agent/agent/version.go:22 +0x7e fp=0xc82004bf48 sp=0xc82004bf28
main.init()
	/go/src/github.com/buildkite/agent/main.go:101 +0x45 fp=0xc82004bf50 sp=0xc82004bf48
runtime.main()
[...]

PR #327, opened 2016-05-26
Allow metadata value to be read from stdin to workaround issue #288.

daveoxley:
Introduces new parameter `--value-from-file` for metadata set. When this parameter exists it treats the value as a path to a file which is read. The contents of the file are then set as the metadata value.

### DIFF
--- a/clicommand/meta_data_set.go
+++ b/clicommand/meta_data_set.go
@@ -9,6 +9,7 @@ import (
 	"github.com/buildkite/agent/logger"
 	"github.com/buildkite/agent/retry"
 	"github.com/codegangsta/cli"
+	"io/ioutil"
 )
 
 var MetaDataSetHelpDescription = `Usage:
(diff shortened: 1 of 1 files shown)


sj26:
Oh, this looks great!

Perhaps instead of adding an argument and opening the file ourselves we could presume that if no value is provided on the command line then a value is being provided via stdin, which leaves piping etc as options for file input?

sj26:
Similiar to https://github.com/buildkite/agent/blob/20d0c6eee8367dd3788aec672cd2d0bb5ca17ff9/clicommand/pipeline_upload.go#L105-L111

daveoxley:
Latest commit attempts to read from stdin if no metadata value is provided. Do you want me to remove the `--value-from-file` functionality or leave it in?

sj26:
Yes please. I think we don't need it if you can pipe a file in, and keeping it simpler is probably better. :-)

sj26:
Thanks @daveoxley! I tweaked it a little to still accept empty values (which are valid use-cases) and added some documentation. There should be an experimental build available shortly.

sj26:
What platform are you on? These experimental builds should include this change:

- http://download.buildkite.com/agent/experimental/2.1.13.1200/buildkite-agent-linux-amd64
- http://download.buildkite.com/agent/experimental/2.1.13.1200/buildkite-agent-darwin-amd64

daveoxley:
Yeah, I've confirmed the experimental build resolves the issue. Thanks very much :+1:

keithpitt:
@daveoxley just wanted to give you a heads up this is now in the latest stable!

keithpitt:
Thanks again for helping us fix it!!

daveoxley:
@keithpitt I've migrated to the release version now and confirm it's working well. Thanks. :+1:

PR #335, opened 2016-06-02
Bootstrap interpolation fixes

sj26:
This is a fairly extensive review of the legacy bootstrap.sh. There are no functional changes. This should be a drop-in for existing stable agents.

Some highlights:

- Many commands have been changed to use positional parameter passing so that distinct words are preserved and isn't broken by whitespace or quotes.
- Printing arbitrary escape codes is now avoided.
- Where values must be printed or used in eval'd scripts we now use printf with automatic escaping.
- Jobs which are running a single command corresponding to a local file have been unwrapped from shell scripts, too, as it was unnecessary and makes Docker workflows a little more painful.
- Some constraints have been made a little stricter, i.e. print ENV starting with (not containing) "BUILDKITE"

I've tested this and compared to 2-1-stable bootstrap.sh and master go bootstrap running jobs which are:

- a command
- a script
- a command running via docker
- a script running via docker
- a command running via docker compose 1.6.2
- a script running via docker compose 1.6.2
- a command running via docker compose 1.7.1
- a script running via docker compose 1.7.1
- a local hook command
- a global hook command

using: https://github.com/sj26/test

/cc @keith @toolmantim


toolmantim:
Will have to give this a good test run!

sj26:
I've rebased this against latest 2-1-stable.

templates/bootstrap.sh
```diff
@@ -463,7 +463,7 @@ else
     fi
 
     # Run the build script command in the service specified in BUILDKITE_DOCKER_COMPOSE_CONTAINER
-    echo "+++ $BUILDKITE_COMMAND_ACTION (in Docker Compose container)"
+    echo "~~~ $BUILDKITE_COMMAND_ACTION (in Docker Compose container)"
```
toolmantim: I recently added this in https://buildkite.com/changelog/10-buildkite-agent-2-dot-1-9 — perhaps we need to update the other parts of the bootstrap to do this instead?

sj26: We only want the "Running command" line to expand if no other "---" or "+++" sections are supplied by the build command output, because most of the time it's just the output of the command execution:

```
~~~ Running command...
./script/something
--- Running specs...
...
```

We make sure that "Running command..." or similar is expanded if no other sections are open in the Buildkite frontend, and I've verified that the section is expanded for the all the scenarios I've listed in the original pull request message, including the docker compose ones. :-)

toolmantim: Ah, I must not have been testing it properly back in the day… from memory it wasn't automatically expanding for simple commands, such as running `ls` or things if people hadn't got the stage of adding `echo '---'` style output headers. Not sure why it wasn't working back then! I'll leave this to you 👍

keithpitt:
🚀

keithpitt:
Let's do it!

PR #337, opened 2016-06-07
Fix ssh keyscan

sj26:
There were a few edge cases here which meant that we could be scanning every single time we run a job, and the ssh known hosts file would fill with duplicates.

These changes should make sure that we only scan for know hosts once, whether HashKnownHosts is set or not (see [ssh_config(5)](http://man.openbsd.org/ssh_config)).


keithpitt:
Works for me! 🚀

Issue #338, opened 2016-06-09
Build using existing working copy doesn't fetch tags correctly


ivovh:
I'm trying to build code that lives at a certain tag in the git repository. When creating a new working copy this works fine:
![image](https://cloud.githubusercontent.com/assets/1636963/15933597/a0ed0f32-2e60-11e6-91a7-727ca9b3a553.png)

Unfortunately, when the working copy is already there it doesn't work:
![image](https://cloud.githubusercontent.com/assets/1636963/15933510/4aaf79a2-2e60-11e6-8e68-b0583ed176b7.png)

It seems like the tags aren't fetched properly. If I use `git fetch -tv` instead of `git fetch -v` it does work. We're using agent v3.0-beta3.

sj26:
Ah, this is a tricky one. Tags!

We can't use the `--tags` argument because we need to support git versions before it was introduced.

Here's the code path which seems to be running in your screenshot:

https://github.com/buildkite/agent/blob/v3.0-beta.3/agent/bootstrap.go#L942-L953

It looked like fetching a tag will work using the first attempted checkout, but because it's not mapped to the same ref locally (because it's a tag) it can't be checked out using its' symbolic name. It would work fine if we checked out the FETCH_HEAD instead.

Can you change your `commit` in however you're triggering this build to be `HEAD` instead of the tag name? That should make it use this code path which should work:

https://github.com/buildkite/agent/blob/v3.0-beta.3/agent/bootstrap.go#L934-L937

In the mean time I'll see if I can patch this to detect when a single commit fetch succeeds and use the `FETCH_HEAD` in that case as well.

sj26:
Another potential hurdle: if the commit is set to a symbolic ref instead of a fully-qualified commit sha then you might not get the same commit across all jobs in your pipeline because Buildkite won't update the job's commit value.

ivovh:
> Can you change your commit in however you're triggering this build to be HEAD instead of the tag name?

Unfortunately this isn't possible. We trigger the build from Phabricator. Once a _diff_ (it's kind of like a versioned PR) is created, the code is uploaded to a certain tag belonging to the diff id. Each update will be tagged with a new diff id. The actual commit hash is not saved anywhere as far as I know.

mgood:
@ivovh you may be interested to know Phabricator is adding an extension for Buildkite support: https://secure.phabricator.com/T12173

[...]

PR #339, opened 2016-06-14
Commit fetch means we can checkout fetch head

sj26:
This also fixes a case where we build a branch or tag and put the symbolic ref name into both the branch and commit field (like #338).


keithpitt:
🚀

keithpitt:
Whoops, found a regression in this. Will revert and we can fix up in another branch.

PR #341, opened 2016-06-16
Expect plugin hooks to live in a hooks dir

toolmantim:
This changes the plugin gear to expect hooks in a hooks dir instead of the repo root:

```
buildkite-plugins/docker-compose/
buildkite-plugins/docker-compose/hooks/
buildkite-plugins/docker-compose/hooks/command
```


keithpitt:
👍

PR #342, opened 2016-06-16
Use correct slashes on Windows

### LINKED ISSUE #328: `buildkite-agent pipeline upload` has wrong default path for Windows
When running `buildkite-agent pipeline upload` on Windows the default path should be `.buildkite\pipeline.yml` but we're using the wrong slash currently `.buildkite/pipeline.yml`

sj26:
We need to translate relative paths like this for use on Windows. This should fix #328.


keithpitt:
slack_for_ios_upload

keithpitt:
Just tested this in my VM, works a treat!

PR #343, opened 2016-06-16
Revert "Commit fetch means we can checkout fetch head"

keithpitt:
Reverts buildkite/agent#339


sj26:
Ah, sorry about this! It just needed the checkout line copied. I'll re-apply the patch and fix the problem, then test it a little more thoroughly.

PR #349, opened 2016-06-30
[RFC] Adding a hooks/secrets

toolmantim:
The idea of a secret hook is that any environment values it exports could be automatically scrubbed from build output for you. It's implemented as a standard Go `io.Writer` that knows what secrets were setup in the secrets hook.

The test cases in this first cut bring up some interesting questions… like:

* How do you sanitise partial values?
* What happens if the output gets split between write byte calls? Buffering? How does it work?

### DIFF
--- a/agent/bootstrap.go
+++ b/agent/bootstrap.go
@@ -462,7 +462,7 @@ func (b *Bootstrap) addRepositoryHostToSSHKnownHosts(repository string) {
 	}
 }
 
-// Executes a hook and applyes any environment changes. The tricky thing with
+// Executes a hook and applies any environment changes. The tricky thing with
 // hooks is that they can modify the ENV of a bootstrap. And it's impossible to
 // grab the ENV of a child process before it finishes, so we've got an awesome
 // ugly hack to get around this.  We essentially have a bash script that writes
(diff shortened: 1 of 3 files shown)


toolmantim:
So the partial values case shouldn't be an issue I think, as long as the buffer was initialised with enough size for the largest thing we're searching for. As for how to do buffering… there must be some good Golang examples?

The goal though I think shouldn't be to provide 100% protection, but just "improved protection" and the ability to use `env` for debugging build scripts (which is pretty common) without having to worry about secrets being exposed.

Another option is we provide some other mechanism for exposing secrets which isn't exporting environment variables, and try some other approach altogether?

toolmantim:
To be clear, this wouldn't help to distribute secrets between agents, or to help protect malicious people with access to builds from getting access to secrets—this hook only makes sense because `env` and such commands are so common in build script and plugin debugging, and we recommend setting secrets by exporting environment variables. This would only decrease, not completely eliminate, the chance of someone exposing creds. Environment variables are just such a common way that secrets are configured.

Completely eliminating the probability of someone sending sensitive creds to Buildkite probably isn't possible.

We could come up with some other pattern to make exposing creds less common?

If you don't use environment variables, but normal variables, then they're not exported in `env`:

```bash
var1="$(aws s3 cp "$S3_FILE_URL" -)"
var2="$(some-secret-vault-storage-facility)"
```

We also can't scrub for their value… because they're not exposed to agent process (unlike exported env vars).

Hmm…

toolmantim:
Closing for lack of interest…

PR #352, opened 2016-07-01
Support plugin array configs

toolmantim:


toolmantim:
@keithpitt wasn't sure how to do arrays of ints, and other funky stuff?

keithpitt:
This looks pretty close! Does `[]int` work? We can just use the exact same setup that you currently have and just convert them to strings.

toolmantim:
Yeah we can do straight ints easy. Any idea how you'd do a mix of strings and ints? My Google fu last night was failing me

keithpitt:
I'm actually not sure what the JSON parser will do (because you can't have an array of mixed types I think). But give it a crack and see what happens!

keithpitt:
(It may all just work)

toolmantim:
Dang, so the int stuff never worked in the first place! Because of this: http://stackoverflow.com/questions/22343083/json-marshaling-with-long-numbers-in-golang-gives-floating-point-number

PR #353, opened 2016-07-04
Trigger Docker pipeline via triggers

toolmantim:
No more need for `curl`!

Also axed the `steps.deprecated.json` which doesn't seem to be referenced anywhere


keithpitt:
Looks good! 🚀

PR #355, opened 2016-07-13
Remove giant dockerfile

toolmantim:
This removes the giant slow Dockerfile (and pre-build + push step optimisation), and now it just uses the standard golang docker image in the docker-compose.yml file, which is snappy even on completely blank/cold machines.

It also splits out the binary building into separate steps, so it's parallelized and and runs in 1 minute instead of 4.5 minutes.

I think this pipeline should probably be split up a bit… perhaps Agent Build + Agent Release? All the release steps run on the deploy agent at the moment… and it's just assumed it has all the necessary things (`ruby`, `fpm`, etc)


keithpitt:
Sounds good to me!

PR #361, opened 2016-08-03
Fix Content-Type for Google Cloud Storage API calls

grosskur:
The Google Cloud Storage team rolled out a change today to their service that checks the Content-Type header more strictly against the GS object metadata. Previously, the Buildkite agent would use the Google API library in the normal way, which would allow GS to sniff the content type and fill in the metadata header. But since we also set the Content-Type header when uploading, this unfortunately results in errors like:

    Failed to PUT file "foo" (googleapi: Error 400: Content-Type specified in the upload (text/plain; charset=utf-8) does not match Content-Type specified in metadata (application/json). If it was a simple upload (uploadType=media), the Content-Type was specified as a bare header. If it was a multipart upload (uploadType=multipart), then the Content-Type was specified in the second part of the multipart. If it was a resumable upload (uploadType=resumable), then the Content-Type was specified with the X-Upload-Content-Type header with the start of the resumable session. For more information, see https://cloud.google.com/storage/docs/json_api/v1/how-tos/upload., invalid) (Attempt 1/10 Retrying in 1s)

The Google Cloud Storage team recommends fixing this by updating to the latest Google API client library and disabling Content-Type auto-sniffing with:

    Media(file, googleapi.ContentType(""))
   
See google/google-api-go-client#134 for background.


keithpitt:
Nice! Thanks for fixing this for us! I'll merge it in now.

Are you using the 3.0 beta series to 2.1.x?

grosskur:
Thanks! We're now using the latest master (3.0 beta series).

keithpitt:
@grosskur cool cool, I just pushed out a new beta with your changes: https://github.com/buildkite/agent/releases/tag/v3.0-beta.7

Let me know if it's all good now!

PR #362, opened 2016-08-05
Remove submodules, extraneous and with no .gitmodules folder

tekacs:


keithpitt:
Ah thanks - I was just about to do this :) Thanks!!

PR #363, opened 2016-08-05
Have buildkite-agent install a systemd template for instances.

tekacs:
I comment on this in buildkite/docs#46 as well, but this is a change which would have buildkite-agent install a systemd 'template' unit, which would allow users to start additional agents by simply running:

```bash
sudo systemctl start buildkite-agent@2
sudo systemctl start buildkite-agent@3
sudo systemctl start buildkite-agent@arbitrary-string
...
```

Note how the user doesn't have to copy the unit file in order to spawn multiple workers.

Note that they can be stopped or monitored en-masse by using (for example):
`sudo systemctl status 'buildkite-agent@*'`

This also has the advantage that the 'buildkite-agent@.service' file can be version controlled/package managed by buildkite-agent, which the currently suggested ad-hoc copies (buildkite-agent-2.service, etc.) cannot.

In this pull request I leave the original 'buildkite-agent.service' around for compatibility reasons.

### DIFF
--- a/packaging/linux/scripts/after-install-and-upgrade.sh
+++ b/packaging/linux/scripts/after-install-and-upgrade.sh
@@ -71,6 +71,7 @@ fi
 if [ $BK_SYSTEMD_EXITS -eq 0 ] && [ $BK_IS_UBUNTU_14_10 -eq 1 ]; then
   if [ ! -f /lib/systemd/system/buildkite-agent.service ]; then
     cp /usr/share/buildkite-agent/systemd/buildkite-agent.service /lib/systemd/system/buildkite-agent.service
+    cp /usr/share/buildkite-agent/systemd/buildkite-agent.service /lib/systemd/system/buildkite-agent@.service
   fi
 
   START_COMMAND="sudo systemctl enable buildkite-agent && sudo systemctl start buildkite-agent"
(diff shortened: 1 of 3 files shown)


tekacs:
If someone has more time than me, note that systemd substitutes the string `%i` for the arbitrary string after the `@` sign in the instance name, so that could be used to name the buildkite-agent instances for display purposes on buildkite.com.

(i.e. `systemctl start buildkite-agent@cheesecake` => `%i == 'cheesecake'` inside of `buildkite-agent@.service`)

keithpitt:
Thanks for the PR @tekacs! @sj26 is our local systemd expert, so I'll let him review this one :)

keithpitt:
Related: https://github.com/buildkite/docs/pull/46 and https://github.com/buildkite/docs/pull/40

sj26:
Thanks for the contribution @tekacs! Sorry for the late response.

We're already doing something very similar for our ops agents: https://gist.github.com/sj26/bccfac9142bf804113b4c1f4026d105a

I'd love to do as you say and make this the recommended way to do things, i.e. add an /etc/systemd/buildkite-agent.service like this one to override the default which has dependencies on instances of a packaged template.

We'll need to add some docs here before we merge. I'll take a look when I can and keep you updated.

keithpitt:
@sj26 do these docs suffice for us to be able to merge this in? https://github.com/buildkite/docs/pull/40 and https://github.com/buildkite/docs/pull/46

sj26:
Not quite. I think we should include a service instance template with the agent, but not enable it by default, then recommend overriding the main service in /etc/ with something invoking the service instances if you wish. I'll draft up a PR.

sj26:
Thanks, I've rebased and merged this with some additions! 💖

PR #364, opened 2016-08-08
`sha1` has been deprecated in latest homebrew

keithpitt:
I believe the `sha1` call in homebrew has been deprecated! (see: https://github.com/buildkite/homebrew-buildkite/issues/4)

This PR removes the auto-updating of it - next step will be to remove it entirely from the homebrew repo.

@toolmantim does this look OK to you?


toolmantim:
I'd remove the sha1 from the homebrew recipe first, then this will run and update everything AOK :rocket:

PR #365, opened 2016-08-09
Make bootstrap actually use the global command hook if it exists

skevy:
`globalCommandHookPath` was referencing the local command hook path.

### DIFF
--- a/agent/bootstrap.go
+++ b/agent/bootstrap.go
@@ -1072,7 +1072,7 @@ func (b *Bootstrap) Start() error {
 	// We need to manually run these hooks so we can customize their
 	// `exitOnError` behaviour
 	localCommandHookPath := b.localHookPath("command")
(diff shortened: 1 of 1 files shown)


keithpitt:
Oh whoops, good find! :)

keithpitt:
Thanks for the fix :)

skevy:
Not to be a pain, but is there any chance you could release a new version of the beta with this change? I suppose I can fork and build my own copy, but would prefer not to do that if I don't have to :)

keithpitt:
Of course! I'm quickly gonna go grab a snack, and will do it when I sit back down at my desk ☕

skevy:
Thanks so much!

keithpitt:
@skevy SHIPPED! :) https://github.com/buildkite/agent/releases/tag/v3.0-beta.8

PR #366, opened 2016-08-11
fix a few typos

cab:


keithpitt:
Oh whoops! Thanks for catching those!! <3

PR #373, opened 2016-08-26
Disable CGO

sj26:
We really want to ship an entirely static binary to make it as easy as possible for folks to run the buildkite-agent without worrying about glibc etc. So let's disable CGO entirely.

This means the system's networking configuration may not be entirely honoured. On *nix systems Go will read /etc/resolv.conf and use those dns servers, but otherwise skip any other name resolution mechanisms like mDNS. This includes losing Bonjour/mDNS on macOS.

At the moment this also means we lose process titles, but they've only been available on amd64 linux until now and we have [a pure Go solution to that in the works](https://github.com/buildkite/agent/pull/373).

For Windows, this shouldn't cause any change.

/cc @toolmantim

### DIFF
--- a/scripts/utils/build-binary.sh
+++ b/scripts/utils/build-binary.sh
@@ -36,6 +36,9 @@ if [[ "$GOOS" == "windows" ]]; then
   BINARY_FILENAME="$BINARY_FILENAME.exe"
 fi
 
+# Disable CGO completely
+export CGO_ENABLED=0
+
 mkdir -p $BUILD_PATH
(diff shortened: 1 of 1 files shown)


toolmantim:
👏 this would unblock our Docker gear. I'm cool with shipping this in master/experimental whilst we work on process title changes.

keithpitt:
I thiiiiink some peeps are relying on the process titles at the moment... I have memories of talking about it with Shopify in the Buildkite Chat. We should check!

keithpitt:
I remember because I went to do this before and was like: "i'll just get rid of the process titles because they don't really work anyway" and someone was like: "oh...but we use those..."

keithpitt:
(I do agree that this is the correct change to make though!!)

sj26:
Let's make this change to fix master, then we'll fix proctitle.s

PR #375, opened 2016-08-30
Add env vars for ec2 and gcp config options

toolmantim:
As part of https://github.com/buildkite/docs/commit/18e02116a9ab29887ab7cf5cea6db21aedc341d4 I noticed that the ec2 and gcp options were the only ones without an environment variable equivalent. This adds the environment variable setting support to these options.


keithpitt:
👍

PR #380, opened 2016-09-13
Enable http2 for Agent API

sj26:
This will hopefully make the agent api interactions even faster! The agent endpoint should soon have http2-aware layer 7 load balancing.

### DIFF
--- a/agent/api_client.go
+++ b/agent/api_client.go
@@ -7,6 +7,7 @@ import (
 	"runtime"
 	"time"
 
+	"golang.org/x/net/http2"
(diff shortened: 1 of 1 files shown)


keithpitt:
Oh nice! Is this backwards compatible as well?

sj26:
Yep, should just be a transparent upgrade where possible, and fall back to http 1.1 or what have you. This is crisply how the default golang http client works now, but it doesn't automatically http2 because we're using a custom dialer.

keithpitt:
:rocket:

PR #381, opened 2016-09-26
retries for grabbing ec2 metadata

davidk-zenefits:
When starting several agents in quick succession some of the agents fail to populate with the proper EC2 metadata for tags. I suspect it has something to do with grabbing IAM credentials and throttling on AWS side. Retries with some random delay should fix the problem.

### DIFF
--- a/agent/agent_pool.go
+++ b/agent/agent_pool.go
@@ -113,7 +113,7 @@ func (r *AgentPool) CreateAgentTemplate() *api.Agent {
 			tags, err := EC2MetaData{}.Get()
 			if err != nil {
 				// retry.Do will sleep 1s per configuration and we tack on a few random ms
-				time.Sleep(1000 * r.Float32() * time.Millisecond)
+				time.Sleep(time.Duration(1000 * r.Float32()) * time.Millisecond)
 			} else {
 				for tag, value := range tags {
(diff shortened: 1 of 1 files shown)


keithpitt:
Hey @davidk-zenefits - this is a good change! Thanks so much for the PR!

We have a our own `retry` function that can use here: https://github.com/buildkite/agent/blob/805b277a1eb75077fc583ecac890b1376119b094/clicommand/meta_data_get.go#L76

If you're happy to update the code to use that (if you want to extend retry to add some randomness to it, then that's cool to) I'd happily merge this in and push out a stable release :)

davidk-zenefits:
Cool. I'll try.

davidk-zenefits:
I think I got all the errors. At least running `go build main.go` no longer gives any errors. Let me know if there are more changes that are necessary @keithpitt.

keithpitt:
Thanks @davidk-zenefits! I just merged these changes into master. I made a few tweaks: added some extra logging and moved your sleep stuff directly into our retry gear - since it's a really good idea!

Which version of the agent are you currently running? I'll make sure to get a new release out in the next few days.

davidk-zenefits:
Fantastic. Thanks a lot. According to agent overview page we are running `User Agent: buildkite-agent/2.2 (linux; amd64)`. We've already worked around the issue on our end by adding a sleep between each agent start call to avoid the AWS throttling so the metadata is populated correctly now.

PR #386, opened 2016-09-29
Fix escaping bug in publish-debian-package.sh

### LINKED ISSUE #385: Potential bug in `publish-debian-package.sh`
Lines [21](https://github.com/buildkite/agent/blob/master/scripts/utils/publish-debian-package.sh#L21), [26](https://github.com/buildkite/agent/blob/master/scripts/utils/publish-debian-package.sh#L26), and [31](https://github.com/buildkite/agent/blob/master/scripts/utils/publish-debian-package.sh#L31) of [`publish-debian-package.sh`](https://github.com/buildkite/agent/blob/master/scripts/utils/publish-debian-package.sh) look like they might have a subtle escaping bug.

If I'm not mistaken, the variable names following "Error: Missing ENV variable" are interpolated, thus the string "Error: Missing ENV variable " is echoed.

If they don't, then the validations below should likely be updated for consistency.

colinrymer:
The variable names following "Error: Missing ENV variable" in the error messages should not be interpolated, so that the missing variable can be identified.

Fixes #385

### DIFF
--- a/scripts/utils/publish-debian-package.sh
+++ b/scripts/utils/publish-debian-package.sh
@@ -18,17 +18,17 @@ COMPONENT="main"
 
 # Some validations
 if [ -z "$GPG_SIGNING_KEY" ]; then
-  echo "Error: Missing ENV variable $GPG_SIGNING_KEY"
+  echo "Error: Missing ENV variable GPG_SIGNING_KEY"
   exit 1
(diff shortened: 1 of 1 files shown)


keithpitt:
Oh, good find! Thanks!

PR #387, opened 2016-10-04
Substring expansion

sj26:
Enables substring expansion in pipeline.yml files, like this practical example:

```yaml
steps:
  - name: ":docker: :hammer:"
    plugins:
      docker-compose:
        build: app
        tags:
          - ${BUILDKITE_COMMIT:0:7}
```

This follows the conventions of [linux shell parameter substring expansion](http://www.tldp.org/LDP/abs/html/parameter-substitution.html#EXPREPL1) and embraces the [bash version](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html) which allows negative offsets and lengths, too.

### DIFF
--- a/agent/pipeline_parser.go
+++ b/agent/pipeline_parser.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"regexp"
+	"strconv"
 	"strings"
 )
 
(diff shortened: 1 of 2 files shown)


jufemaiz:
Hi @sj26 does this imply we can start using custom tags with docker-compose?

sj26:
@jufemaiz not quite yet! But that's coming in buildkite-plugins/docker-compose#7

jufemaiz:
Should I get @L226 to break down his PR (https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/pull/8) into the atomic parts?

sj26:
@jufemaiz I think those changes might already be in progress as https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/pull/3 and https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/pull/7. I've asked for clarification in https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/pull/8. :-)

Issue #391, opened 2016-10-09
Multiple commands error detection on Windows


essen:
Hello,

A followup to #223, this time on Windows 7. Agent version: 3.0-beta.11

I have setup a pipeline with the following two commands to run:

```
cat badfile
cat Makefile
```

Output is:

```
> cat badfile
cat: badfile: No such file or directory
> cat Makefile
# See LICENSE for licensing information.
 
PROJECT = ranch
[snip]
```

Despite the failure of the first command, the build succeeds.

Putting the commands in reverse order makes the build fail, hinting that maybe only the last command matters on Windows.

While the above commands run MSYS2 programs, this seems to be unrelated to MSYS2, because the exact same problem can be observed with the following commands:

```
set abc
set
```

And reversing those also makes the test fail instead of succeeding.

keithpitt:
Hey @essen, thanks for reporting!

I just had a quick look into this, and the script we generate for *nix machines is done here: https://github.com/buildkite/agent/blob/master/agent/bootstrap.go#L1135 The `set -e` ensures that the script stops if any of the commands fail.

Here we generate the script we run on Windows: https://github.com/buildkite/agent/blob/master/agent/bootstrap.go#L1126

Is there an equivalent `set -e` type directive we can use in Batch to make it bail if a command fails?

essen:
I have no idea but I will look into it.

essen:
I got it working for the simple case (set abc; set). Now trying against a full test suite with many commands, and will open a PR if all goes well.

essen:
PR is open! #392 

For future reference (mostly for me I guess), I had to do the following to get the environment setup under MSYS2 and properly test the builds:

* Install Go 1.7 from the official site
* export GOROOT=/c/Go (where Go was installed)
* export GOPATH=/mingw64 (though any folder would do)
* export PATH=$PATH:/c/Go/bin
* go get github.com/buildkite/agent
* cd /mingw64/src/github.com/buildkite/agent
* go build
* mv agent.exe buildkite-agent.exe
* PATH=$PATH:. buildkite-agent.exe start --debug --token "mytoken" --build-path /tmp/ --meta-data "os=dev" (PATH because I had path issues otherwise, and metadata to easily select the agent I was testing)

It took me a little of trial and error and is probably not the most efficient way to do it but at least I could debug from SSH. Had to stop the agent from the Web UI in between tries though.

keithpitt:
Merged! Thanks for the PR @essen!

PR #392, opened 2016-10-10
Fix error detection on Windows

essen:
Windows builds failed to report an error if it didn't occur
on the last command of the build. Batch files have no equivalent
for "set -e"; instead we need to check after each command for
the return value and exit if it is non-zero.

Fix for #391

### DIFF
--- a/agent/bootstrap.go
+++ b/agent/bootstrap.go
@@ -1128,7 +1128,8 @@ func (b *Bootstrap) Start() error {
 					if k != "" {
 						buildScriptContents = buildScriptContents +
 							fmt.Sprintf("ECHO %s\n", windows.BatchEscape("\033[90m>\033[0m "+k)) +
-							k + "\n"
+							k + "\n" +
+							"if %errorlevel% neq 0 exit /b %errorlevel%\n"
 					}
 				}
 			} else {


keithpitt:
@essen oh nice! I wasn't expecting a PR to fix the issue, this is awesome. I'll get it tested today and if all goes well, I'll publish another beta for you to check out.

essen:
Me neither. :-)

I've had it running all day (copied the .exe onto the previous beta install). Double checked every time, haven't seen a problem so far. Looking forward to new beta!

ticky:
I’d suggest when we test this we should make sure it doesn’t have issues if flow control is used in the supplied script.

Also worth noting that this won’t do anything about batch scripts which delimit commands with `;`, while our bash `set -e` behaviour does - is that a problem?

essen:
I did not know batch had `;` so the problem didn't even occur to me. My needs are very small (simple list of commands, one per line) so I would not be surprised if many cases are not covered. And my knowledge of batch even smaller.

keithpitt:
I did some testing of this yesterday, and @ticky is right that we have problems when it comes to `;` - however - our current implementation has this problem already. Since we `ECHO` each line, it will already break up command that are across multiple lines (like conditionals) and things with `;`.

So since we have this problem already, I'm gonna merge this in and push out another beta to get @essen's stuff working. Perhaps in 3.1 we can take another look at how those commands are handled and make it better.

An idea I had was:

```yaml
steps:
  - command:
     - "ECHO first command"
     - "./foo.batch"
```

So if peeps had complex command setups (and didn't want to put it all in a script) we could support this syntax, and then put the "ECHO" and exit status checking after each.

essen:
Cheers!

keithpitt:
@essen sorry for the lateness, but I've just pushed out another beta with your fix! https://github.com/buildkite/agent/releases/tag/v3.0-beta.12

essen:
No worries, it was week-end after all. I'll install it tonight, thanks!

PR #395, opened 2016-10-18
Pre-process pipelines before uploading

keithpitt:
This PR is making a few big changes to how pipelines are uploaded to the BK API:

- Instead of uploading the pipeline files "as-is", we parse them first (either as YAML or as JSON) and upload them to the Agent API as JSON all the time. This gives us the benefit of catching JSON or YAML syntax errors before even hitting the Agent API
- Instead of doing an Environment Variable Interpolation across the entire pipeline as a whole, we recursively go through the entire config and do it for strings only, after the first parse. This solves many quoting issues
- When piping a pipeline in via STDIN, you must now tell the agent what format it's in `--format yaml` or `--format json`
- We now accept `.buildkite/pipeline.yaml` (not just `.yml`) files as default file. Apparently `.yaml` is actually the _official_ extension, but no one seems to care?
- `.buildkite/steps.json` has officially been removed

There's a few more things I need to do before I can merge this in:

- [x] Add the YAML parsing library to `master` and merge in
- [ ] Fix the tests
- [ ] Write some tests
- [x] Move the EnvironmentVariableInterpolator into the it's own package (done in master)


keithpitt:
Closing in support of another cleaner PR

PR #397, opened 2016-10-18
Pre-process pipelines before uploading

### LINKED ISSUE #346: ENV interpolation and invalid YAML
Given this:

```yaml
  - trigger: "deploy"
    label: ":shipit:"
    branches: "master"
    async: true
    build:
      message: ${BUILDKITE_MESSAGE}
      commit: ${BUILDKITE_COMMIT}
      branch: ${BUILDKITE_BRANCH}
```

And a `BUILDKITE_MESSAGE` of

"Bumped frontend (:sasslang: emoji)"

becomes...

```yaml
  - trigger: "deploy"
    label: ":shipit:"
    branches: "master"
    async: true
    build:
      message: Bumped frontend (:sasslang: emoji)
      commit: ...
      branch: ...
```

you end up getting an error from Buildkite "There was an error parsing the config file. Maybe it's invalid YAML?"

I think after we do the interpolation we end up creating an invalid YAML file. Instead of doing an interpolation over the entire file (just treat the whole thing as text) I think we need to parse the YAML locally, recursively search over the structure, and interpolate each value individually.

keithpitt:
This PR is making a few big changes to how pipelines are uploaded to the BK API:

- Instead of uploading the pipeline files "as-is", we parse them first (either as YAML or as JSON) and upload them to the Agent API as JSON all the time. This gives us the benefit of catching JSON or YAML syntax errors before even hitting the Agent API
- Instead of doing an Environment Variable Interpolation across the entire pipeline as a whole, we recursively go through the entire config and do it for strings only, after the first parse. This solves many quoting issues (Fixes #346)
- When piping a pipeline in via STDIN, you must now tell the agent what format it's in `--format yaml` or `--format json`
- We now accept `.buildkite/pipeline.yaml` (not just `.yml`) files as default file. Apparently `.yaml` is actually the _official_ extension, but no one seems to care?
- `.buildkite/steps.json` has officially been removed

There's a few more things I need to do before I can merge this in:

- [x] Add the YAML parsing library to `master` and merge in
- [x] Fix the tests
- [x] Write some tests
- [x] Move the EnvironmentVariableInterpolator into the it's own package (done in master)

### DIFF
--- a/agent/pipeline_parser.go
+++ b/agent/pipeline_parser.go
@@ -1,0 +1,209 @@
+package agent
+
+import (
+	"encoding/json"
+	"fmt"
+	"path/filepath"
+	"reflect"
+
+	"github.com/buildkite/agent/envvar"
+	"github.com/buildkite/agent/logger"
+	"github.com/ghodss/yaml"
+)
+
+type PipelineParser struct {
+	Filename string
+	Format   string
+	Pipeline []byte
+}
+
+func (p PipelineParser) Parse() (pipeline interface{}, err error) {
+	format := p.Format
+
+	// If no format is passed, figure it out based on the filename
+	if format == "" {
+		logger.Debug("Pipeline format not supplied, inferring from filename `%s`", p.Filename)
+
+		format, err = inferFormatFromFilename(p.Filename)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// Unmarshal the pipeline into an actual data structure
+	unmarshaled, err := unmarshal(p.Pipeline, format)
+	if err != nil {
+		return nil, err
+	}
+
+	// Recursivly go through the entire pipeline and perform environment
+	// variable interpolation on strings
+	interpolated, err := interpolate(unmarshaled)
+	if err != nil {
+		return nil, err
+	}
+
+	return interpolated, nil
+}
+
+func inferFormatFromFilename(filename string) (string, error) {
+	// Make sure we've got a filename in the first place
+	if filename == "" {
+		return "", fmt.Errorf("No filename to infer a format from")
+	}
+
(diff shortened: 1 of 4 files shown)


toolmantim:
Wooo! For STDIN can we just default to YAML?

keithpitt:
Yeah! I think I can make that change!

keithpitt:
@toolmantim so I had a stab at defaulting to YAML, and the problem is the experience for folks that use JSON is a bit poo poo. So the current behaviour:

`thing.sh | buildkite-agent pipeline upload`

Will fail with a "hey, please tell us the format", and the developer we go: "oh, that's annoying...I should fix that!", then the change it to:

`thing.sh | buildkite-agent pipeline upload --format json`

However if we default to YAML, the error they'd get is something like: "error parsing YAML, invalid token \n blah blah" sort of thing. And they may be like: "but it's JSON...why is it doing YAML? I didn't tell it to do that"

But maybe this problem could be solved with a better error message? I figured it was better to be more specific with this sort of thing, instead of a weird default that they may not understand.

keithpitt:
Actually...I has an idea...

toolmantim:
Oh, so the current default from STDIN is JSON? I just assumed the current default was YAML so was suggesting we keep that (also, because all our docs are YAML)

keithpitt:
@toolmantim so I ended up just getting rid of the whole `--format` thing all together: https://github.com/buildkite/agent/pull/397/files#diff-bfaa789ca90c13e83df7b0e86b66520fR43

Currently in the back end we automatically figure out if it's JSON by looking at the first character (if it's a `[` or a `{`) so I've just replicated the behaviour here. It's not entirely "accurate", but it seems to have worked for us up to this point!

PR #401, opened 2016-11-08
Update path to ssh tools for Windows

bendrucker:
Spent some time trying to set up an agent on Windows and ran into this. Here's my VM showing that `ssh-keygen.exe` is in `C:\Program Files\Git\usr\bin` and didn't match the path traversal that the agent is trying to do. This is from Windows Server 2012 running in Vagrant:

screen shot 2016-11-07 at 22 07 33

Git was installed via Chef. Not sure that this is the right fix since the current path could be right for other install methods. Just wanted to attach some code in case it turns out to be right.

### DIFF
--- a/agent/bootstrap.go
+++ b/agent/bootstrap.go
@@ -446,10 +446,6 @@ func (b *Bootstrap) addRepositoryHostToSSHKnownHosts(repository string) {
 					break
 				}
 			}
-
-			if sshToolBinaryPath == "" {
-				warningf("Could not find `ssh-keygen`")
-			}
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


keithpitt:
Hey @bendrucker thanks for the PR! I think on my system they were at that original location (I installed git using GitHub for Windows). So I think what we _may_ need to do here is have an array of possible locations, and then loop through them to try and find it.

I don't suppose you'd be up for giving that change a go? :)

bendrucker:
I'd be happy to give that a shot. Will update.

bendrucker:
Wrote some gross code that I thought would handle this but can't seem to actually generate a build for Windows that runs it. It compiles ok into `pkg/` and I'm able to copy the exe to Windows and start an agent. But the warning about the `ssh-keygen` path is identical and it seems like the code that's running is unchanged.

This is how I'm building:

```sh
./scripts/utils/build-binary.sh windows 386 1
```

Happy to jump on Slack if you'd like.

keithpitt:
@bendrucker just tested this locally and works a treat, thanks so much for the fix!

bendrucker:
Awesome, glad to help!

keithpitt:
@bendrucker I've just pushed a new beta release out https://github.com/buildkite/agent/releases/tag/v3.0-beta.14 with this included!

bendrucker:
Awesome, should have some time early next week to finalize our Windows builds. Nice to have SSH automatically handled!

PR #405, opened 2016-11-17
Use an older version of the init.d script for Debian 7

keithpitt:


keithpitt:
Even though this feels a bit dirty, it's probably the easiest thing we can do no so peeps can get their agents running again!

PR #406, opened 2016-11-17
Conditional locks

sj26:
Instead of an entirely new startup script we can just detect if the lock directory exists at all and only perform locking if it does. Hopefully this is a little more resilient.

/cc @keithpitt


sj26:
We an also make sure that the `su` has a login shell, regardless of what the user database says.

PR #412, opened 2016-11-24
Add an example to environment hook for providing inline_image function

ineffyble:
Extend the commented examples in the environment hook sample, with a sample export of the inline_image function, so that it can be easily called in bash scripts.

### DIFF
--- a/packaging/linux/root/usr/share/buildkite-agent/hooks/environment.sample
+++ b/packaging/linux/root/usr/share/buildkite-agent/hooks/environment.sample
@@ -4,8 +4,14 @@
 # to set up secrets, data, etc. Anything exported in hooks will be available
 # to the build script.
 #
-# For example:
+# Examples:
 #
+### Set an environment variable
 # export SECRET_VAR=token
+#
+### Export inline_image function for bash scripts, to embed images in logs
+# function inline_image { printf '\033]1338;url='"$1"';alt='"$2"'\a\n'; }
+# export -f inline_image
+#
 
 set -e


keithpitt:
Thanks for the PR @ineffyble! Unfortunately this won't be possible in v3 of the agent 😢

In v2 of the agent (which is latest stable) the job bootstrapper (bootstrap.sh) is written in Bash. This worked incredibly well for us for ages, but we reached the limit as to what can be done with it.

In v3 we wanted to add plugins, but we didn't want to write a plugin architecture in Bash. We also had many bugs when it came to the Windows version of the same file (bootstrap.bat) so we rewrote the whole thing in Go (which is available as `buildkite-agent bootstrap` in v3). The fixed a ton of outstanding bugs and allowed us to add new features we'd be meaning to add for ages.

We managed to come up with a method of allowing hooks to export variables like in v2, but functions on the other hand didn't make the cut (mostly because we couldn't figure out how to do it).

The change has been good and bad. Bad in the sense that you can't export functions anymore, but good in the sense that stuff doesn't leak as much from one hook to the other (also if a hook exits that means the rest of the bootstrap process doesn't have to exit as well).

Thanks for the suggestion though! I think in your case you might need a `common.sh` type file that you source in your own build scripts.

ineffyble:
Calculon

> Oh, how cruel and melodramatic fate is

...now I want to add a Calculon emoji

keithpitt:
We would deffintily accept a `:calculon:` emoji...

PR #421, opened 2017-01-05
Artifact upload destination env

sj26:
@mcgain was trying to setup a custom s3 bucket for artifact uploads and noticed that `$BUILDKITE_ARTIFACT_UPLOAD_DESTINATION` is only respected in the bootstrap's default artifact uploader, and not when using `buildkite-agent artifact upload ...` manually. It seems the destination arg is not a normal flag and doesn't ingest environment variable values by default, so if the argument is not supplied positionally then it always uses the default buildkite artifact bucket. I've added support to our arg parser so that positional args can also fall through to environment variables, which doesn't seem to be available in the codegangsta cli package.

@mcgain, care to give it a go and confirm it's working now?

### DIFF
--- a/clicommand/artifact_upload.go
+++ b/clicommand/artifact_upload.go
@@ -38,7 +38,7 @@ Example:
 
 type ArtifactUploadConfig struct {
 	UploadPaths      string `cli:"arg:0" label:"upload paths" validate:"required"`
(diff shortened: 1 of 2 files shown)


mcgain:
I tried this out locally and it works fine now. Thanks!

PR #423, opened 2017-01-17
[WIP] `buildkite-agent pipeline validate` command

ticky:
This command does the same things as `pipeline upload`, apart from the actual uploading part. This is useful for catching syntax errors in scripts when testing locally.

- [x] Basic command
- [ ] Break shared stuff from `pipeline upload` out into a shared method
- [ ] Perhaps validate the structure, rather than just the validity of the serialisation format used?

### DIFF
--- a/clicommand/pipeline_validate.go
+++ b/clicommand/pipeline_validate.go
@@ -1,0 +1,145 @@
+package clicommand
+
+import (
+  "io/ioutil"
+  "os"
+  "path"
+  "path/filepath"
+  "strings"
+
+  "github.com/buildkite/agent/agent"
+  "github.com/buildkite/agent/cliconfig"
+  "github.com/buildkite/agent/logger"
+  "github.com/buildkite/agent/stdin"
+  "github.com/codegangsta/cli"
+)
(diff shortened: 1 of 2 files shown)


keithpitt:
Oh this neat! I always figured we'd have this sort of command (I was thinking `buildkite-agent pipeline lint` or something, but `validate` works probably better!

ticky:
I’m happy to take direction on the name, just felt like this’d be nicer to point at than “when you upload, it’ll validate!” lol

keithpitt:
Oh yeah, totally. I think `validate` makes much more sense, since that's the word you use when speaking anyway. You don't say: "hey is there a way to lint the config file" - you usually ask "is there a way to validate that my yaml is right?"

keithpitt:
Although there's more validation that happens on Buildkite's (like "hey, `cmdfun` isn't a real property!") so it's more about validating that the YAML syntax is correct I think?

keithpitt:
I'm just gonna close this PR for now until we come back to it. I'll leave the branch alone though!

PR #428, opened 2017-02-03
Fix lock not found error

sj26:
Some folks noticed a problem when trying to use plugins on hosts with many agents:

```
--- Setting up plugins
# Could not aquire lock on "/var/lib/buildkite-agent/plugins/github-com-buildkite-plugins-docker-compose-buildkite-plugin.lock" (Locked by other process)
# Trying again in 1 second...
# Could not aquire lock on "/var/lib/buildkite-agent/plugins/github-com-buildkite-plugins-docker-compose-buildkite-plugin.lock" (Locked by other process)
# Trying again in 1 second...
🚨 Buildkite Error: lstat /var/lib/buildkite-agent/plugins/github-com-buildkite-plugins-docker-compose-buildkite-plugin.lock: no such file or directory
```

Looks like we had a slightly older version of [the lockfile library we're using](https://github.com/nightlyone/lockfile). The new version seems to handle this case now, so let's try updating as a fix.

This uses govendor to keep track of the new revision, which we should probably start doing more. 🚀


vendor/vendor.json
```diff
@@ -0,0 +1,13 @@
+{
```
keithpitt: What generates this file? Perhaps if we're gonna do a `vendor.json` thing  we should do them all at once in a seperate PR?

agent/bootstrap.go
```diff
@@ -265,9 +265,8 @@ func acquireLock(path string, seconds int) (*lockfile.Lockfile, error) {
 	for true {
 		err := lock.TryLock()
 		if err != nil {
-			if err == lockfile.ErrBusy {
-				// Keey track of how many times we tried to get the
-				// lock
+			if te, ok := err.(interface{ Temporary() bool }); ok && te.Temporary() {
```
keithpitt: Oh boo, I hate having to do error inspection like this. Makes for such an annoying API.

So reading through the docs, `Temporary` is the "you should try again" error?

keithpitt: Also `TemporaryError("Lockfile created, but doesn't exist")` is a weird error. I wonder why you'd get that?

vendor/vendor.json
sj26: govendor.

I tried doing it all at once and it is very difficult. I think if we slowly update the dependencies and track where they came from at what revision we'll get there eventually.

agent/bootstrap.go
sj26: Yeah, it's just the author's way of saying "this error means try again" I think. It's a big of a pig, but it works!

PR #430, opened 2017-02-16
Add option to timestamp lines

stefanmb:
***Background:*** 
The issue in https://github.com/buildkite/feedback/issues/80 has become a showstopper for me. I am supporting a Buildkite based container building service which allows users to provide their own complex build scripts. There are often performance problems that must be diagnosed. Since I do not control the individual build steps I cannot add group header to their build output in order to get timing information. I require an option to timestamp each line of output so we can profile any part of the build.

I've tried ad-hoc solutions involving `ts` and `unbuffer` but they are not drop-in replacements, this feature is needed in the agent itself.

***Changes:***
This PR adds a `BUILDKITE_TIMESTAMP_LINES` environment variable. If set to one of ` 1, t, T, TRUE, true, True, 0, f, F` each line of output will be prefixed with an RFC3339 UTC timestamp. 

Note that the header groups are not affected.

I've tried to avoid introducing perf hits (due to single-threaded regex parsing) if the option is not enabled. The code became a bit complex as the header API appears to only transmit a list of timestamps and their matching headers must be parsed a second time by the backend, so they cannot be prefixed. Unfortunately this approach does introduce more coupling between `process.go` and `header_times_streamer.go`.

I made this PR against `2-3-stable` because that's what we're using. If you want me to port it to `master` please let me know.

***Review:***
Would greatly appreciate some comments from @keithpitt @toolmantim.
cc: @camilo @n1koo

### DIFF
--- a/agent/header_times_streamer.go
+++ b/agent/header_times_streamer.go
@@ -73,18 +73,16 @@ func (h *HeaderTimesStreamer) Scan(line string) {
 	h.scanWaitGroup.Add(1)
 	defer h.scanWaitGroup.Done()
 
-	if h.lineIsHeader(line) {
-		logger.Debug("[HeaderTimesStreamer] Found header %q", line)
+	logger.Debug("[HeaderTimesStreamer] Found header %q", line)
 
-		// Aquire a lock on the times and then add the current time to
-		// our times slice.
-		h.timesMutex.Lock()
-		h.times = append(h.times, time.Now().UTC().Format(time.RFC3339Nano))
-		h.timesMutex.Unlock()
+	// Aquire a lock on the times and then add the current time to
+	// our times slice.
+	h.timesMutex.Lock()
+	h.times = append(h.times, time.Now().UTC().Format(time.RFC3339Nano))
+	h.timesMutex.Unlock()
 
-		// Add the time to the wait group
-		h.uploadWaitGroup.Add(1)
-	}
+	// Add the time to the wait group
+	h.uploadWaitGroup.Add(1)
 }
 
 func (h *HeaderTimesStreamer) Upload() {
@@ -136,7 +134,7 @@ func (h *HeaderTimesStreamer) Stop() {
 	h.streamingMutex.Unlock()
 }
 
-func (h *HeaderTimesStreamer) lineIsHeader(line string) bool {
+func (h *HeaderTimesStreamer) LineIsHeader(line string) bool {
 	// Make sure all ANSI colors are removed from the string before we
 	// check to see if it's a header (sometimes a color escape sequence may
 	// be the first thing on the line, which will cause the regex to ignore
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 	"sync"
 	"time"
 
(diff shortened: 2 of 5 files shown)


agent/header_times_streamer.go
```diff
@@ -73,18 +73,16 @@ func (h *HeaderTimesStreamer) Scan(line string) {
 	h.scanWaitGroup.Add(1)
 	defer h.scanWaitGroup.Done()
 
-	if h.lineIsHeader(line) {
```
keithpitt: I'm probably missing something, but I think this should still be there? Otherwise each line is probably considered a `---header` right?

keithpitt:
Hey @stefanmb  - this is awesome! I'm just doing a code review now and we'll probably need to do some manual testing on our end. But I'm happy to add it to stable once we've done everything we need to!

agent/job_runner.go
```diff
@@ -59,13 +60,20 @@ func (r JobRunner) Create() (runner *JobRunner, err error) {
 	// the Buildkite Agent API
 	runner.logStreamer = LogStreamer{MaxChunkSizeBytes: r.Job.ChunksMaxSizeBytes, Callback: r.onUploadChunk}.New()
 
+	timestamp, err := strconv.ParseBool(os.Getenv("BUILDKITE_TIMESTAMP_LINES"))
```
keithpitt: Maybe we should check this in `AgentConfiguration`? Or are you thinking that `BUILDKITE_TIMESTAMP_LINES` could be turned on/off from the BK UI at runtime?

agent/header_times_streamer.go
stefanmb: I had to move that check out of `header_times_streamer.go`, it introduces some more coupling but I didn't want the check to happen twice. Unfortunately I have to know if something is a header before I can timestamp it, so it's done here now: 

https://github.com/buildkite/agent/pull/430/files#diff-14ee62401fee1966227df2f4ab955c3cR192

process/process.go
```diff
@@ -171,11 +178,37 @@ func (p *Process) Start() error {
...
-				defer lineCallbackWaitGroup.Done()
-				p.LineCallback(line)
-			}(string(line))
+			// If we're timestamping this main thread will take
+			// the hit of running the regex so we can build up
+			// the timestamped buffer without breaking headers,
+			// otherwise we let the goroutines take the perf hit.
+
+			checkedForCallback := false
+			lineHasCallback := false
```
keithpitt: This bit here has me a little confused. `lineHasCallback` always seems to be false unless timestamping has been turned on?

agent/job_runner.go
stefanmb: Sure, I can move it there!

process/process.go
stefanmb: Yeah, that variable is not strictly needed but I left it for clarity. 

The idea was to use a short circuit evaluation to prevent doing the check twice in case the timestamping option is enabled: https://github.com/buildkite/agent/pull/430/files#diff-14ee62401fee1966227df2f4ab955c3cR207

I admit it's not super elegant.

agent/job_runner.go
```diff
@@ -59,13 +60,20 @@ func (r JobRunner) Create() (runner *JobRunner, err error) {
...
-		Env:           r.createEnvironment(),
-		PTY:           r.AgentConfiguration.RunInPty,
-		StartCallback: r.onProcessStartCallback,
-		LineCallback:  runner.headerTimesStreamer.Scan,
+		Script:             r.AgentConfiguration.BootstrapScript,
+		Env:                r.createEnvironment(),
+		PTY:                r.AgentConfiguration.RunInPty,
+		Timestamp:          timestamp,
+		StartCallback:      r.onProcessStartCallback,
+		LineCallback:       runner.headerTimesStreamer.Scan,
```
keithpitt: Ohhh right right, I see how this works now. Took a while for it to click :)

So essentially, `LineCallback` will only be called if the `LineCallbackFilter` function returns true for a particular line, is that right?

stefanmb: That's exactly the idea! The ugly bit of coupling is that lines are not timestamped if there is *any* callback handler. However, the only callback handler that exists right now is used for the header timestamps.

stefanmb:
@keithpitt I've moved the env variable to `AgentConfiguration` (and added a corresponding command line switch). Let me know if there is anything else I can do to help! Thanks!

keithpitt:
Rad, I've just done some local testing and it works great. Let's merge it!

toolmantim:
I think we forgot to port this to master…

mdub:
Is this something that can be enabled per-pipeline? Or can it only be set at agent level?

stefanmb:
Agent level as I implemented it.

PR #432, opened 2017-02-20
Switch to govendor for dependency tracking

sj26:
[Govendor](https://github.com/kardianos/govendor) is a tool for managing the vendor directory in Go projects. It can calculate and fetch the exact packages your project requires. It also removes extraneous test files. And it tracks the exact revision of the packages you're using.

Fetch a new dependency with:

```
$ govendor fetch cloud.google.com/go/compute
```

Fetch and fix an exact version of a dependency with:

```
$ govendor fetch cloud.google.com/go/compute@v0.5.0
```

Update existing dependencies the same way:

```
$ govendor fetch cloud.google.com/go/compute # or cloud.google.com/go/compute@v0.6.0 etc.
```

Prune unused dependencies:

```
$ govendor remote +unused
```

But my favourite feature is that you can just import a package in your go code:

```go
package agent

import "cloud.google.com/go/compute/metadata"

# ...
```

Then tell govendor to figure it out:

```
$ govendor fetch +outside
```

:tada:

Heroku have [a pretty good user guide](https://devcenter.heroku.com/articles/go-dependencies-via-govendor).

This PR includes adding the `vendor/vendor.json` file (like a `Gemfile.lock`, basically) and updates a bunch of our dependencies. The code compiles, and I can successfully start an agent and run a build. I've also created an equivalent branch for 2.3: https://github.com/buildkite/agent/compare/2-3-stable...2-3-govendor. :ok_hand:


sj26:
Huh, it seems I broke artifacts. AdrRoll/goamz seems to be colliding awkwardly with the official [aws-sdk-go](https://github.com/aws/aws-sdk-go).

PR #436, opened 2017-03-11
Add --disconnect-after-job command line arg

toolmantim:
When creating container-based schedulers and runners you often want the runner to exit after running the job (e.g. [hyper-buildkite-agent](https://github.com/toolmantim/hyper-buildkite-agent))

This adds a command line flag that causes the agent to shut down once it's completed a job.

- [x] Add `--disconnect-after-job` command line flag
- [x] Add `--disconnect-after-job-timeout` command line flag

### DIFF
--- a/agent/agent_configuration.go
+++ b/agent/agent_configuration.go
@@ -11,4 +11,5 @@ type AgentConfiguration struct {
 	CommandEval                bool
 	RunInPty                   bool
 	DisconnectAfterJob         bool
+	DisconnectAfterJobTimeout  int
 }
--- a/agent/agent_configuration.go
+++ b/agent/agent_configuration.go
@@ -10,5 +10,5 @@ type AgentConfiguration struct {
 	SSHFingerprintVerification bool
 	CommandEval                bool
 	RunInPty                   bool
-	ExitAfterJob               bool
+	DisconnectAfterJob         bool
 }
--- a/agent/agent_configuration.go
+++ b/agent/agent_configuration.go
@@ -10,4 +10,5 @@ type AgentConfiguration struct {
 	SSHFingerprintVerification bool
 	CommandEval                bool
 	RunInPty                   bool
+	ExitAfterJob               bool
 }
--- a/agent/agent_pool.go
+++ b/agent/agent_pool.go
@@ -258,4 +258,8 @@ func (r *AgentPool) ShowBanner() {
 	if !r.AgentConfiguration.RunInPty {
 		logger.Debug("Running builds within a pseudoterminal (PTY) has been disabled")
 	}
+
+	if r.AgentConfiguration.ExitAfterJob {
+		logger.Debug("Agent will exit after a job run has completed")
+	}
 }
--- a/agent/agent_pool.go
+++ b/agent/agent_pool.go
@@ -63,7 +63,13 @@ func (r *AgentPool) Start() error {
 
 	logger.Info("Agent successfully connected")
 	logger.Info("You can press Ctrl-C to stop the agent")
-	logger.Info("Waiting for work...")
+
+	if r.AgentConfiguration.DisconnectAfterJob {
+		logger.Info("Waiting for first job to be assigned...")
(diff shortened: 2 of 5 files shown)


toolmantim:
I forgot to mention, adding this flag makes the hyper-buildkite-agent-runner image completely unnecessary.

keithpitt:
Rad! I think this may require 1 more bit of technology. I did some thinking about this recently and realised there's a bit of a race condition...

Let's say you spin up an agent specifically to run `queue=$BUILDKITE_JOB_ID` for the cases like Hyper. There's a potential race condition where the agent could get stuck waiting for the job, if something happens to the job in-between rescheduling it, and the agent starts up.

So:

1. Job with build on "feature/blah" has tags `queue=hyper`
2. Job is rescheduled to run on `queue=$BUILDKITE_JOB_ID`
3. API calls are made to setup a Hyper agent with `--tag queue=$BUILDKITE_JOB_ID`
4. A new build is pushed to "feature/blah" and auto-cancelation/skipping technology kicks into gear and cancels the job
5. Agent finally connects and get's stuck in limbo since the job it wants to run will never get to it

I'd recommend maybe a 60 second timeout where if an agent is in `ExitAfterJob` mode, and still hasn't gotten a job, it'll shut itself down. Or if not a hardcoded timeout, a `--exit-after-job-timeout` option?

sj26:
@keithpitt oh, nice catch. I can see you've been thinking about this deeply for a while. ⏰

toolmantim:
Oh yeah, good catch @keithpitt! What about this?

```
Options:

   --config value                  Path to a configuration file [$BUILDKITE_AGENT_CONFIG]
   --token value                   Your account agent token [$BUILDKITE_AGENT_TOKEN]
   --name value                    The name of the agent [$BUILDKITE_AGENT_NAME]
   --priority value                The priority of the agent (higher priorities are assigned work first) [$BUILDKITE_AGENT_PRIORITY]
   --exit-after-job                Shutdown the agent after running a job [$BUILDKITE_AGENT_EXIT_AFTER_JOB]
   --exit-after-job-timeout value  When --exit-after-job is specified, the number of seconds to wait for a job before shutting down (default is "120") [$BUILDKITE_AGENT_EXIT_AFTER_JOB_TIMEOUT]
   --meta-data value               Meta-data for the agent (default is "queue=default") [$BUILDKITE_AGENT_META_DATA]
   --meta-data-ec2                 Include the host's EC2 meta-data (instance-id, instance-type, and ami-id) as meta-data [$BUILDKITE_AGENT_META_DATA_EC2]
   ...
```

I also sort of ignored the worker pool design… but perhaps that can be revisited at a future time when there is actual support for multiple workers?

keithpitt:
`--exit-after-job-timeout` is 👌 

The only other bit of feedback is that maybe it should be `--shutdown-after-job` instead of `exit`. Since we don't really "exit" agents, and agents don't really "exit". They do "disconnect" and "stop" though! Maybe it should be `--disconnect-after-job`?

keithpitt:
> I also sort of ignored the worker pool design… but perhaps that can be revisited at a future time when there is actual support for multiple workers?

Worker pool design? As in, 1 agent running multiple jobs at once?

toolmantim:
@keithpitt yeah!

keithpitt:
Oh right, I didn't realise that was a thing that was needed? For Hyper, I think our 1 job per agent is fine?

toolmantim:
I just meant that the place where I put the code, in the worker, felt a little at odds with the worker pool design. But I think it's something we just clean up / address later once we actually support that.

toolmantim:
Nice work @keithpitt! Should we add support for `-1` if someone wants to make sure it never times out?

keithpitt:
@toolmantim in that case they could probably just make it `100000` or something? I did think of that case, but couldn't think of an actual "no, please don't timeout" use case. What ya think?

toolmantim:
Yeah, I couldn't think of a solid case for `-`. Let's ship it, and we can add support for disabling the timeout later if necessary!

PR #438, opened 2017-03-17
Do not timestamp header expansion directives

stefanmb:
**What:**
In https://github.com/buildkite/agent/pull/430 I introduced the line timestamp feature, but there is a bug where `^^^ +++` gets timestamped and consequently fails to expand the previous group. The PR resolves this issue.

**How:**
The [documentation](https://buildkite.com/docs/builds/managing-log-output) indicates `^^^ +++` can be used to expand a header group.

Currently this line will be time stamped because it does not pass the header regex. However, this line is not actually a header, and should not be reported as one to the Buildkite backend. Instead, a special check is made to avoid timestamping these special lines.

*Pros:* Fixes the bug!
*Cons:* Even more coupling between `header_times_streamer` and `process`.

**Review:**
@keithpitt

### DIFF
--- a/agent/header_times_streamer.go
+++ b/agent/header_times_streamer.go
@@ -134,14 +134,16 @@ func (h *HeaderTimesStreamer) Stop() {
 	h.streamingMutex.Unlock()
 }
 
-func (h *HeaderTimesStreamer) LineIsHeader(line string) bool {
+func (h *HeaderTimesStreamer) LinePreProcessor(line string) string {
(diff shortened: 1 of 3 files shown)


keithpitt:
Ah whoops, yeah I should have picked up that before. Thanks for the fix! Will push out a 2.4.1 today.

keithpitt:
@stefanmb 🚀 a new release!

PR #443, opened 2017-03-28
freebsd: fix architecture detection

dch:
FreeBSD's `uname -sm` returns `FreeBSD amd64` not `x86_64`, so we need 
to ensure that the agent installs the 64-bit version on 64-bit flavours.


keithpitt:
@dch thanks! 👍

PR #451, opened 2017-04-13
Use /bin/sh instead of /bin/bash

toolmantim:
Backport #448 to 2.6

### DIFF
--- a/templates/bootstrap.sh
+++ b/templates/bootstrap.sh
@@ -425,7 +425,7 @@ else
 
     # We'll actually run a temporary file with pipefail and exit-on-fail
     # containing the full script body, printed literally through printf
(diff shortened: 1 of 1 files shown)


templates/bootstrap.sh
```diff
@@ -425,7 +425,7 @@ else
 
     # We'll actually run a temporary file with pipefail and exit-on-fail
     # containing the full script body, printed literally through printf
-    printf "#!/bin/bash\nset -eo pipefail\n%s\n" "$BUILDKITE_COMMAND" > "$BUILDKITE_COMMAND_PATH"
+    printf "#!/bin/sh\nset -eo pipefail\n%s\n" "$BUILDKITE_COMMAND" > "$BUILDKITE_COMMAND_PATH"
```
lox: pipefail is a `bash` thing, not a `sh` thing

PR #455, opened 2017-04-18
Use /bin/sh instead of /bin/bash

keithpitt:
Opening a PR to try making this change again. Will do some local testing to see if it all works.

### DIFF
--- a/agent/bootstrap.go
+++ b/agent/bootstrap.go
@@ -355,16 +355,16 @@ func (b *Bootstrap) runScript(command string) int {
 		cmd = b.newCommand(command)
 	} else {
 		// If you run a script on Linux that doesn't have the
-		// #!/bin/bash thingy at the top, it will fail to run with a
+		// #!/bin/sh thingy at the top, it will fail to run with a
 		// "exec format error" error. You can solve it by adding the
-		// #!/bin/bash line to the top of the file, but that's
+		// #!/bin/sh line to the top of the file, but that's
 		// annoying, and people generally forget it, so we'll make it
 		// easy on them and add it for them here.
 		//
 		// We also need to make sure the script we pass has quotes
-		// around it, otherwise `/bin/bash -c run script with space.sh`
+		// around it, otherwise `/bin/sh -c run script with space.sh`
 		// fails.
-		cmd = b.newCommand("/bin/bash", "-c", `"`+strings.Replace(command, `"`, `\"`, -1)+`"`)
+		cmd = b.newCommand("/bin/sh", "-c", `"`+strings.Replace(command, `"`, `\"`, -1)+`"`)
 	}
 
 	process, err := shell.Run(cmd, &shell.Config{Writer: os.Stdout, PTY: b.RunInPty})
@@ -481,7 +481,7 @@ func (b *Bootstrap) addRepositoryHostToSSHKnownHosts(repository string) {
 // Executes a hook and applyes any environment changes. The tricky thing with
 // hooks is that they can modify the ENV of a bootstrap. And it's impossible to
 // grab the ENV of a child process before it finishes, so we've got an awesome
-// ugly hack to get around this.  We essentially have a bash script that writes
+// ugly hack to get around this.  We essentially have a shell script that writes
 // the ENV to a file, runs the hook, then writes the ENV back to another file.
 // Once all that has finished, we compare the files, and apply what ever
 // changes to our running env. Cool huh?
@@ -518,7 +518,7 @@ func (b *Bootstrap) executeHook(name string, hookPath string, exitOnError bool, 
 				"SET > \"" + tempEnvAfterFile.Name() + "\"\n" +
 				"EXIT %BUILDKITE_LAST_HOOK_EXIT_STATUS%"
 		} else {
-			hookScript = "#!/bin/bash\n" +
+			hookScript = "#!/bin/sh\n" +
 				"env > \"" + tempEnvBeforeFile.Name() + "\"\n" +
 				". \"" + absolutePathToHook + "\"\n" +
 				"BUILDKITE_LAST_HOOK_EXIT_STATUS=$?\n" +
@@ -1165,7 +1165,7 @@ func (b *Bootstrap) Start() error {
 					}
 				}
 			} else {
-				buildScriptContents = "#!/bin/bash\nset -e\n"
+				buildScriptContents = "#!/bin/sh\nset -e\n"
 				for _, k := range strings.Split(b.Command, "\n") {
 					if k != "" {
 						buildScriptContents = buildScriptContents +


keithpitt:
@sj26 can you think of any reason why this wouldn't work?

sj26:
Many of our hooks and plugins rely on running in bash and use bash only features I think. The hook scripts are sourced, not executed, so they would fail. For instance, most rely on `pipefail` being enabled:

https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/blob/master/hooks/command#L3

This also made me realised that our stable and beta command scripts aren't quite compatible. In stable we enable pipefail in genrated scripts:

https://github.com/buildkite/agent/blob/2-6-stable/templates/bootstrap.sh#L428

sj26:
I think we may need to re-think plugin communication. I don't think the agent should require bash to use plugins. Windows, non-bash *nix systems, etc. We may have to switch to a back-channel socket or something.

toolmantim:
Really, this change was just for executing the generated command bash file in Docker container (nothing related to hooks etc). For the Buildkite Agent 3 Docker plugin it won't even generate the bash file, as it implements its own customer command hook. So I think we just close this for now, and simply leave it as bash.

keithpitt:
I just wanted to re-open this so I can get a further understanding around why this may be a bad idea :)

So this line here:

```
cmd = b.newCommand("/bin/bash", "-c", `"`+strings.Replace(command, `"`, `\"`, -1)+`"`)
```

Is how we run scripts.... What's the difference between that and:

```
cmd = b.newCommand("/bin/sh", "-c", `"`+strings.Replace(command, `"`, `\"`, -1)+`"`)
```

The command we just end up running on the system is:

```
 /bin/sh -c ".buildkite/deploy.sh --with-arg true"
```

If `.deploy.sh` has `#!/bin/bash` at the top, won't it "just work"?

ticky:
Shebangs are interpreted by the system’s program loader, not by a shell you pass them to. In this case you’re explicitly asking the `sh` executable to interpret a script which wants `/bin/bash`!

keithpitt:
@ticky I think that's where my brain fails..because say I've got `my-awesome-script` which has this:

```
#!/usr/bin/env ruby

puts "hello"
```

And then I do this:

```
/bin/sh -c "./my-awesome-script"
```

It runs the Ruby! I interpret what you said as "if you do /bin/sh - what ever you pass to it should be run as `sh`, not the thing in the shebang at the top"

sj26:
I think @keithpitt is right, because of `-c`. `/bin/sh $COMAND` would be a problem but `/bin/sh -c $COMMAND` should be fine, but then so should `./command`. _Except_ for the comment directly above that line, which is also why this causes everything to break:

```
		// If you run a script on Linux that doesn't have the
		// #!/bin/sh thingy at the top, it will fail to run with a
		// "exec format error" error. You can solve it by adding the
		// #!/bin/sh line to the top of the file, but that's
		// annoying, and people generally forget it, so we'll make it
		// easy on them and add it for them here.
```

If we switch from `/bin/bash` to `/bin/sh` for scripts which don't have a shebang then suddenly those scripts are running in a different shell, and scripts written which use bash might break because `/bin/sh` might still link to bash but will run it in compat mode with all the bash fanciness turned off.

This is more problematic for the hook scripts later. The docker compose plugin for instance heavily relies on pipefail and advanced traps to work.

keithpitt:
@sj26 so....

> This is most problematic for the hook scripts and plugins later. The docker compose plugin for instance heavily relies on pipefail and advanced traps to work. And the hooks have to source their scripts into the same shell environment so we can capture the env after the script has run, we can't rely on a shebang to change the hook's shell.

That means it's only a problem for scripts that *don't* have a shebang already? If hook scripts add `/bin/bash` at the top, then everything will "just work"?

sj26:
@keithpitt unfortunately, no.

Hook scripts are _sourced_ not _executed_. They are loaded into the existing shell script so their environment changes can be grabbed when finished. You currently cannot use anything but a bash script as a hook script. Other shebangs are not respected and compiled executables simply do not work.

sj26:
This will also change how any snippets of script typed into pipeline.yml or supplied in the web ui are run with not ability to change them because we insert our own shebang above any user-supplied input.

dch:
what about shipping with a default shebang for those scripts and snippets that don't have their own already, and allowing the default to be overridden in the pipeline?

lox:
For reference: https://docs.docker.com/engine/reference/builder/#shell

lox:
Could we have a agent parameter for specifying the shell, with a default of `/bin/sh -c` for nix and `cmd /S /C` for windows?

lox:
@sj26 are you sure we source the hook scripts? Wasn't that a 2.x thing? https://github.com/buildkite/agent/blob/master/agent/bootstrap.go#L522

lox:
Also, this makes things interesting: 

> with a single command to execute, bash execve’s it directly

On quite a few systems, `/bin/sh` is an alias for bash, or dash.

sj26:
@lox [the line running the script](https://github.com/buildkite/agent/blob/f7fec37b3b1232bc107f5d524561e0e750dcb9b4/agent/bootstrap.go#L523) uses [`.`](https://www.gnu.org/software/bash/manual/html_node/Bourne-Shell-Builtins.html#index-_002e) (`source` is [a non-posix synonym](https://www.gnu.org/software/bash/manual/html_node/Bash-Builtins.html#index-source)) to load the script. It's because we dump the env before the script, run the script in the same shell context, then dump the env after and do a diff. If we execute the script instead then we lose any environment changes.

It's quite clever, but it means hooks can only be written in the same shell.

This is why I think we need a mechanism beyond shell for making env changes from within hooks. I think that mechanism could potentially allow other things, too, like fetching more data than provided in the env, maybe eventually a limited graphql query capability for build/job data etc, and for streamlining operations like pipeline upload, meta-data, and artifact handling over the same http contexts.

When bash is invoked as `/bin/sh` it runs in a [posix compatibility mode](https://www.gnu.org/software/bash/manual/html_node/Bash-POSIX-Mode.html), disabling most bash-specific features.

toolmantim:
I think we should stay with requiring/assuming `bash` everywhere for now, and in a post-v3 release we can tackle all the bash'isms. I'd be happy to close this for now.

keithpitt:
@toolmantim yeah cool, will close now.

I wanted to re-open just so I fully understood why we couldn't do it, and think we've figured it out now. At least we have somewhere to point if someone asks in the future!

JensRantil:
> I think we should stay with requiring/assuming bash everywhere for now, and in a post-v3 release we can tackle all the bash'isms. I'd be happy to close this for now.

@toolmantim I think I am the Buildkite customer that originally brought up this issue (over support e-mail). I just wanted to chime in here that we are fine with that workaround as long as that dependency is documented (which it wasn't at the time - is it?) in documentation. It was a really messy to debug the issue when our Docker container didn't contain `bash` because Docker doesn't give very friendly error messages.

PR #457, opened 2017-04-20
Fix a typo in Readme

mousavian:


keithpitt:
Ah, nice catch @mousavian!

PR #460, opened 2017-05-01
Retry EC2 tags fetch in case of rate limits

sj26:
If we're asked to fetch EC2 tags as agent tags/meta-data then try a couple of times in the face of failure in case we hit rate limits or network problems.

### DIFF
--- a/agent/ec2_tags.go
+++ b/agent/ec2_tags.go
@@ -7,6 +7,9 @@ import (
 
 	"github.com/AdRoll/goamz/aws"
 	"github.com/AdRoll/goamz/ec2"
+
+	"github.com/buildkite/agent/logger"
+	"github.com/buildkite/agent/retry"
 )
 
 type EC2Tags struct {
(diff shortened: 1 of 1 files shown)


sj26:
@keithpitt I think this is fine, it's just a copy/paste, but would you mind doing a quick review? I'm not sure how to test the failure case. I have manually tested that it still works when running an agent on an ec2 instance with tags.

keithpitt:
@sj26 try running it on your local machine :) That should trigger the error state!

keithpitt:
Oh, that won't work since it auth's AWS just above. You can test it on a machine that has disallows the "get tags" IAM permission

sj26:
@keithpitt good idea, thanks!

sj26:
Oh, it looks like the latest beta agent actually already does this, thanks to

https://github.com/buildkite/agent/commit/1d671548fd24bc3c40fe46152394e097a8b7e189

I'll backport that to stable instead.

keithpitt:
@sj26 oh sheet! Sorry, I totally forgot that this PR existed - I think I even helped write a little bit of it. My brain totally broke - sorry!!

sj26:
See #461.

PR #454, opened 2017-04-15
Pass signals on to bootstrap

lox:
Fix for #389

Previously the bootstrap was bash based, which meant it did some magical things with signal handling. Specifically, it propagated signals down to it's child processes.

The introduction of the golang bootstrap broke this, as signals weren't propagated to sub-processes. This fixes that.

PR #464, opened 2017-05-04
If possible, send signals to process group

lox:
This follows up on #454 and makes our agent behave more like a local shell by sending signals to it's process group, rather than just to the child process.

I believe this will work fine on windows, as the cast will fail.


toolmantim:
Woo, nice one! Looks like the Windows builds are failing at the moment.

lox:
I was wrong, `syscall.Kill` doesn't exist on windows. Have split it out into conditionally compiled files.

lox:
This is good to go, although it questionably introduces a breaking change, as the stable/beta agent doesn't do this currently.

PR #467, opened 2017-05-09
`buildkite-agent annotate` CLI command

keithpitt:
- [x] Add API classes
- [x] Use in the CLI command action
- [x] Write cli docs

Different uses:

```
buildkite-agent annotate "my-annotation"
./script/annotation-generator | buildkite-agent annotate Readme.md
buildkite-agent annotate --style "success"
```


clicommand/annotate.go
```diff
@@ -0,0 +1,155 @@
...
+
+   You can update an existing annotation's style by calling the annotate command
+   again without an annotation body.
+
+Example:
+
+   $ buildkite-agent annotate "All tests passed! :rocket:"
+   $ cat annotation.md | buildkite-agent annotate --style "warning"
+   $ buildkite-agent annotate --style "success" --context "junit"
+   $ ./script/dynamic_annotation_generator | buildkite-agent annotate --style "success'`
```
sj26: Small typo: I think there's a single comma here which should be a double?

clicommand/annotate.go
```diff
@@ -0,0 +1,155 @@
...
+
+   - Links to artifacts generated by your jobs
+   - Test result summaries
+   - Graphs that include analysis about your codebase
+   - Helpful information for team members about what happened during a build
+
+   Annotations can be written in either Markdown or HTML.
+
+   You can update an existing annotation's style by calling the annotate command
+   again without an annotation body.
```
sj26: Worth mentioning that annotations with the same context override each other?

clicommand/annotate.go
```diff
@@ -0,0 +1,155 @@
...
+				return err
+			}
+
+			// Show the unexpected error
+			if err != nil {
+				logger.Warn("%s (%s)", err, s)
+			}
+
+			return err
+		}, &retry.Config{Maximum: 5, Interval: 1 * time.Second})
```
sj26: 34421

sj26: This could use jitter? I feel like maybe we should introduce a little more jitter across this codebase.

keithpitt:
@sj26 thanks! I've addressed all the things! Turns out we had `Jitter` built into the retry gear already!

PR #469, opened 2017-05-11
Move bootstrap to top level subpackage

lox:
As part of learning my way around the agent codebase, I did some minor refactoring. I'm completely happy for these to be rejected, they don't do much beyond consolidate some things and move some things around.

* `agent/bootstrap` moves to the top level `bootstrap` package.
* `shell/process` moves into `bootstrap/subprocess` to make it's intent and use-case clearer
* Removed an unused error return from `shell.EnvironmentFromSlice`. 
* Copy the std library `exec.LookPath` rather than using the env locking stuff in `bootstrap.subprocess`. 

Wrote this on 🌴 so take with a grain of salt :)


keithpitt:
This salt is tasty! Once the build is green I'll do a proper review and then we can have a merge party!

lox:
Cool, fixed @keithpitt. I finally got my IDE to not remove imports that are only in vendors :)

lox:
:shipit: ?

keithpitt:
:rocket:

PR #185, opened 2015-09-20
Replace AdRoll/goamz with official AWS SDK

keithpitt:
This is dependent on: https://github.com/buildkite/agent/pull/186

- [x] Import the offical AWS library
- [x] Rewrite uploader
- [ ] Rewrite downloader
- [ ] Make changes to the Buildkite Agent API so we can use the offical S3 gear to upload/download artifacts from Buildkite's S3 Servers (makes it faster and we can remove the URL Downloader and Form Uploader classes)

In addition to using the official SDK, it also should:

- Fixes https://github.com/buildkite/agent/issues/179
- Also hopefully fixes issues with uploading larger (4gig) files

PR #474, opened 2017-05-18
Migrate the aging goamz library to the latest aws-sdk

lox:
This is take 2 of https://github.com/buildkite/agent/pull/185. 

It aims to be a minimal update to the latest aws-sdk. There are lots of nice things that can be done with parallel uploads and downloads, but in the meantime there are bug fixes that are needed.

* [x] Update ec2-metadata
* [x] Update ec2-tags
* [x] Update s3-uploader
* [x] Update s3-downloader


keithpitt:
@lox is this still a WIP? Or ready for a review party?

lox:
Pretty sure it's actually ready.

agent/s3.go
```diff
@@ -2,75 +2,109 @@ package agent
...
+	e.retrieved = true
 	return
 }
 
-func awsS3Region() (region aws.Region, err error) {
+func (e *credentialsProvider) IsExpired() bool {
+	return !e.retrieved
+}
+
+func awsS3Credentials() *credentials.Credentials {
```
keithpitt: Does this support the same ENV vars as this?

https://github.com/goamz/goamz/blob/master/aws/aws.go#L340

lox: Yup! https://docs.aws.amazon.com/sdk-for-go/api/aws/credentials/#EnvProvider

agent/s3.go
```diff
@@ -2,75 +2,109 @@ package agent
...
+	return credentials.NewChainCredentials(
+		[]credentials.Provider{
+			&credentialsProvider{},
+			&credentials.EnvProvider{},
+			&ec2rolecreds.EC2RoleProvider{},
+		})
+}
+
+func awsS3RegionFromEnv() (region string, err error) {
 	regionName := "us-east-1"
```
lox: I reckon this should probably be based on the region of the bucket in question.

agent/s3.go
keithpitt: Radical.

agent/s3.go
keithpitt: How can you tell a region based on a bucket name?

lox: https://docs.aws.amazon.com/sdk-for-go/api/service/s3/s3manager/#GetBucketRegion I assume magic?

keithpitt: > GetBucketRegion will attempt to get the region for a bucket using the regionHint to determine which AWS partition to perform the query on.
> 
> The request will not be signed, and will not use your AWS credentials.

That's some cool sounding magic! Yeah, I reckon give that a go and if it doesn't work, fall back to the default.

lox:
I've given this an initial test against Minio and it works well. Next up is some testing against S3.

muhlba91:
@lox i am struggling to find out how to configure the artifacts storage for a private minio bucket where i need to set the endpoint url. would you be able to provide me an example on how i'd need to set the configuration for this use case as you tested against minio?

PR #478, opened 2017-05-26
Support for exporting multi-line environment variables

keithpitt:
- [x] Make the rest of it work

### DIFF
--- a/shell/environment.go
+++ b/shell/environment.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"regexp"
-	"runtime"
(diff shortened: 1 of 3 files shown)


shell/environment.go
```diff
@@ -50,29 +160,6 @@ func (e *Environment) Exists(key string) bool {
 
 // Sets a key in the environment
 func (e *Environment) Set(key string, value string) string {
-	// Trim the values
```
keithpitt: I remove all this code because it's all wrong. We shouldn't be doing any of that stuff with environment variable values.

PR #479, opened 2017-05-26
Rewrote environment variable parsing to handle multi-line values

keithpitt:


lox:
Might be worth adding a test for something with a $ in it?

```
declare -x PROMPT="%(?..%B%F{red}✗ exit %?%f%b
)
 \$(directory_name)\$(git_dirty)\$(need_push)\$(author)› "
```

keithpitt:
@lox fixed - also realised I had the wrong export syntax!

PR #483, opened 2017-05-31
Show the file it tried to find when sowing the no-command-eval error message.

### LINKED ISSUE #416: Misleading error when command eval disabled
We are getting a number of questions where people are mistyping their script paths and being confused by a message about command eval. Here's an example:

**Given** I have an agent command eval disabled
**And** I have a repo with the script `bar`
**And** I have a pipeline step with the mistyped command `baz`
**When** my agent runs the step
**Then** my step fails with the message:
> "This agent is not allowed to evaluate console commands. To allow this, re-run this agent without the `--no-command-eval` option, or specify a script within your repository to run instead (such as scripts/test.sh)."

A message such as this would be much easier to diagnose:

> no such file: ./baz
> This agent is not allowed to evaluate console commands. To allow this, re-run this agent without the `--no-command-eval` option, or specify a script within your repository to run instead (such as scripts/test.sh).

keithpitt:
Fixes #416

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -1133,19 +1133,22 @@ func (b *Bootstrap) Start() error {
 			exitf("No command has been defined. Please go to \"Pipeline Settings\" and configure your build step's \"Command\"")
 		}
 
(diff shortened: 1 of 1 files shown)


KevinGrandon:
Thanks for doing this! 👍

Issue #458, opened 2017-04-26
buildkite-agent.cfg token - file path vs integer


erikng:
It would be great if we could set the token to a file path instead of the token integer.

We want to control our cfg file via chef, but this means we have to expose the token in our code repo. Having it link to a file would allow us to deploy the secret, but dynamically control the config file in chef.

toolmantim:
Hey Erik! Interesting! With Chef, it's not possible to embed/compose configs as files? Because the agent config itself is a file, it just has the token inside of it?

erikng:
If we embed the file, our token is inside of it.

We have an internal secret system that we call via chef to write out the file currently, but the limitation is chef doesn't know what the file structure of the config should look like. If we could give the config a path to a file that just contains the token, chef can obtain that secret (and doesn't need to know what the secret is) and we can make changes to our config file as needed.

toolmantim:
Right! So you'd have to store the entire config file itself inside a "secret" in your secret store? Because there's no way to gsub/sed the config with the secret?

erikng:
Yeah, exactly.

toolmantim:
Thanks for clarifying @erikng!

@keithpitt @sj26 is this something you see as okay to add to the agent? If so, I can look at adding support. I think it also fits well with Kubernetes and Docker Swarm secrets too, as they're also files on the filesystem.

sj26:
Hmm, I guess we could. This feels like a very specific use case which could have many other solutions which don't require adding such a specific new feature into the core agent. For example, the way I might do this would be to write out a systemd drop-in which sources an environment file placed somewhere by a secrets system, like:

```
# /etc/systemd/system/buildkite-agent.service.d/secrets.conf
[Service]
EnvironmentFile=/etc/buildkite-agent/secrets
```

```
# /etc/buildkite-agent/secrets
BUILDKITE_AGENT_TOKEN=abc123
```

Would something like that work?

toolmantim:
Is using environment variables for sensitive data considered an anti-pattern these days? Same as command line args? I thought it was, and have been wondering how we can encourage people to move away from specifying the agent token as an env for cases like Docker. I thought this would be the same kind of use case. But I might have over-generalised there.

sj26:
Heh, I think my answer to that is a pretty hefty "it depends."

Sorry for the brain dump, but some thoughts:

[...]

PR #486, opened 2017-06-07
Add `token-script` param for start command that allows external agent-token loading

lox:
After some discussion with the GitHub folks and #458, I thought it might be interesting to look at adding a mechanism to load the agent-token from an external script.

Whilst I don't love the idea of adding complexity for agent-token over a more general solution(for instance a wrapper around buildkite-agent that sets env, or a PreStart script in systemd), I think that the nature of the agent token is that it's currently the "keys to the kingdom" and I think that this approach allows for much stricter security constraints around agent tokens.

I was inspired by openssh's recent addition of [AuthorizedKeysCommand](https://linux.die.net/man/5/sshd_config) that let's you specify an external command for loading AuthorizedKeys. 

Usage of it would be something like:

```bash
buildkite-agent start --token-script ./get_token_from_vault.sh
```

### DIFF
--- a/agent/agent_pool.go
+++ b/agent/agent_pool.go
@@ -17,6 +17,7 @@ type AgentPool struct {
 	APIClient          *api.Client
 	Token              string
 	Token              TokenReader
+	TokenScript        string
 	ConfigFilePath     string
 	Name               string
 	Priority           string
@@ -32,8 +33,19 @@ func (r *AgentPool) Start() error {
 	// Show the welcome banner and config options used
 	r.ShowBanner()
 
+	var token TokenReader = StringToken(r.Token)
+	var err error
+
+	// Support lazily reading agent token from a script / executable
+	if r.TokenScript != "" {
+		token, err = ScriptTokenFromString(r.TokenScript)
+		if err != nil {
+			return err
+		}
+	}
+
 	// Create the agent registration API Client
-	r.APIClient = APIClient{Endpoint: r.Endpoint, Token: r.Token}.Create()
+	r.APIClient = APIClient{Endpoint: r.Endpoint, Token: token}.Create()
 
 	// Create the agent template. We use pass this template to the register
 	// call, at which point we get back a real agent.
--- a/agent/agent_pool.go
+++ b/agent/agent_pool.go
@@ -15,9 +15,7 @@ import (
 
 type AgentPool struct {
 	APIClient          *api.Client
-	Token              string
 	Token              TokenReader
-	TokenScript        string
 	ConfigFilePath     string
 	Name               string
 	Priority           string
@@ -33,19 +31,8 @@ func (r *AgentPool) Start() error {
 	// Show the welcome banner and config options used
 	r.ShowBanner()
 
-	var token TokenReader = StringToken(r.Token)
-	var err error
-
-	// Support lazily reading agent token from a script / executable
-	if r.TokenScript != "" {
-		token, err = ScriptTokenFromString(r.TokenScript)
-		if err != nil {
-			return err
-		}
-	}
-
 	// Create the agent registration API Client
-	r.APIClient = APIClient{Endpoint: r.Endpoint, Token: token}.Create()
+	r.APIClient = APIClient{Endpoint: r.Endpoint, Token: r.Token}.Create()
 
 	// Create the agent template. We use pass this template to the register
 	// call, at which point we get back a real agent.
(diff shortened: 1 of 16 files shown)


lox:
Parking this for now.

lox:
IMO this was a good idea and we probably should have done it. Feels @keithpitt / @sj26?

keithpitt:
I'm to remember why we parked this back then. From memory, it was because there are already solutions to getting keys onto machines (Chef, puppet, server management technology x) and that we should just leave it up to them to get the key onto the machine?

lox:
Yup, although the thing about that is that you need to customize your agent config file for each machine, which makes upgrades hard. If you could defer to an external file for your agent token then you could have the same config shared across lots of machines.

keithpitt:
Couldn't you also achieve it with something like:

```
export BUILDKITE_AGENT_REGISTRATION_TOKEN=${./my_other_script.sh}
```

And we just have a "pre-startup" script somewhere that we source as part of the agent startup?

lox:
> And we just have a "pre-startup" script somewhere that we source as part of the agent startup?

Remember how complicated that was to do in a portable way across init systems? It also leaks the token in the environment for the agent process.

keithpitt:
> It also leaks the token in the environment for the agent process.

We have that problem regardless anyway if you define the registration token with an environment variable. Going back to the original issue that prompted this and this comment here: https://github.com/buildkite/agent/issues/458#issuecomment-298790881 I read:

> Store the entire config in a secret store and continually drop it down each chef run since chef cannot understand the content of the config

I wonder why Chef doesn't understand the format of the config? Maybe if we just supported a different config format systems like Chef could just do the right thing?

lox:
> I wonder why Chef doesn't understand the format of the config? Maybe if we just supported a different config format systems like Chef could just do the right thing?

It's not that it doesn't understand the format, it's that Chef (or most config systems) work on a hash of the config files. If the hash differs then the config file gets updated. You still end up with a different file hash for every system though, as you need per-host customization, or at least per-host-per-token customization.

keithpitt:
I thought Chef had some magical "Just make sure x=y in this config file kthnx" and Chef will parse, check the value, and update it if necessary?

sj26:
I always feel like services should be kept super simple and be managed by something which can take care of this stuff, especially when you want those services to be cross platform. That way you can configure those services the way in which you are familiar, or which suits your platform. i.e. I think forking is the responsibility of the init system or supervisor, not the service. I also think it's a spiral of complexity which shouldn't exist within the service itself, and shouldn't be replicated in every service implementation.

I think I feel the same way here. If you want to feed a token into the agent from a fancy secret store then I think you should use a wrapper script, or something, aws-vault–style.

Passing ENV into a process is not inherently insecure. ENV is not special beyond being a bit of process memory in a known location when that process is started. The process might move it around. This token is going to live in the process memory, however it gets there. `/proc/N/environ` can only be seen if you can read all of the process memory. If we're concerned about the token being sniffed we could even [unset the env](https://golang.org/pkg/os/#Unsetenv) which should clear it in this memory. And we're very careful not to pass too much environment into subprocesses started by the agent.

lox:
I hear what you are both saying, but it seems like a "slippery slope" argument to me. Yeah, if we did this for everything it would suck, but we aren't going to. This is the most high risk piece of config in there and I think it merits an easier way to configure it. We've had feedback from customers that existing init mechanisms (upstart, systemd, launchd, etc) are complicated to get right with tools like chef, ansible, et al and I agree, they are a pain. This seems like a good compromise to me.

lox:
I guess a good mental exercise would be, what if this was an RSA private key and not a short token? Would we still have it in the config file?

sj26:
I really don't like the idea of supervising a process which is supposed to feed us back a token. It just feels wrong to me. It doesn't feel UNIX-y, and I haven't seen it done anywhere else. I'm used to doing this stuff in wrapper scripts, or as part of init systems.

If you have this level of investment in ops then I think it's reasonable to write a wrapper script which puts the token in ENV, and modify or create a new service (i.e. in [Chef](https://docs.chef.io/resource_service.html) or [Puppet](https://puppet.com/docs/puppet/5.3/types/service.html)). This is how I'd expect to prime a service with an RSA key — using a wrapper script.

But ultimately I yield to your discretion, @low. 🙇

lox:
Ok, ok :) Let's park this again and see whether we get any more requests for similar.

keithpitt:
Yeah, this issue is one of the more tricky ones. I'd be keen to reach out to customers to see what they think and how they solve this problem so far. Next time it comes up again I might do that to get some more data!

Issue #491, opened 2017-06-21
AWS EC2 Instance Tags aren't always available immediately on boot


lox:
Currently we are fetching EC2 tags when the agent boots and then providing them to the registration API. Unfortunately sometimes an ASG takes some time to apply the tags to the instance. 

It seems like this is a race condition that sometimes shows up as agents not getting any tags applied to them.

lox:
Closed in #492, available in beta 28.

PR #492, opened 2017-06-21
Wait for EC2 tags to be available before registering the agent

lox:
This addresses #491. TL;DR is that AWS often takes quite some time to apply tags to a running instance, specifically when in an Autoscaling Group.

**Update**
Based on feedback this has changed to be the default behaviour with a timeout to opt-out of. This means if you enable `tags-from-ec2-tags` then we will wait up to 15 seconds for tags (any tags) to be made available. If you don't set tags, but you set `tags-from-ec2-tags` YOU WILL WAIT 15 SECONDS.

### DIFF
--- a/agent/agent_pool.go
+++ b/agent/agent_pool.go
@@ -14,17 +14,18 @@ import (
 )
 
 type AgentPool struct {
-	APIClient          *api.Client
-	Token              string
-	ConfigFilePath     string
-	Name               string
-	Priority           string
-	Tags               []string
-	TagsFromEC2        bool
-	TagsFromEC2Tags    bool
-	TagsFromGCP        bool
-	Endpoint           string
-	AgentConfiguration *AgentConfiguration
+	APIClient             *api.Client
+	Token                 string
+	ConfigFilePath        string
+	Name                  string
+	Priority              string
+	Tags                  []string
+	TagsFromEC2           bool
+	TagsFromEC2Tags       bool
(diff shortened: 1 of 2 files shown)


keithpitt:
Ah, so that was the problem! Here's a question. Is there a scenario where you'd _not_ want that behaviour? It seems that if you've enabled EC2 tags, it's cause you want them. Not because "only get them if you can". Perhaps it should just be the default and not an option?

keithpitt:
Oh I see, you actually don't know if there _are_ any tags. `Wait` actually means "keep asking for tags until you actually get some"

lox:
Yup, it's possible you haven't set any tags for your instance, in which case this would wait forever. I can imagine a future where this is the default if you enabled EC2 tags though.

The other option is to periodically update tags, but I wasn't sure how that would work with re-registering the agent.

agent/agent_pool.go
```diff
@@ -152,9 +157,8 @@ func (r *AgentPool) CreateAgentTemplate() *api.Agent {
 				}
 				s.Break()
 			}
-
 			return err
-		}, &retry.Config{Maximum: 5, Interval: 1 * time.Second, Jitter: true})
+		}, &retry.Config{Maximum: 5, Interval: r.WaitForEC2TagsTimeout / 5, Jitter: true})
```
keithpitt: Neat!

PR #495, opened 2017-06-23
Support branch and pipeline scope for metadata

lox:
Based on a conversation with @keithpitt and I, we realized that supporting metadata scoped wider than the current build scope would be pretty straightforward. 

The thinking was to add "scope" as a concept, which currently defaults to `build`. In other words, currently metadata that you set is scoped to builds. Anyone with access to the build can read it. 

We are adding a `branch` scope, which means that you can shared metadata between different builds/jobs in the same branch of the same pipeline.

The `pipeline` scope is available from anything running in the same pipeline.

Future scopes might include `team` or `organization`, but there are permissions dragons there.

Llama bouncing

### DIFF
--- a/api/meta_data.go
+++ b/api/meta_data.go
@@ -10,10 +10,14 @@ type MetaDataService struct {
 	client *Client
 }
 
+// DefaultMetadataScope is the default scope used for metadata
+const DefaultMetadataScope = "build"
+
 // MetaData represents a Buildkite Agent API MetaData
 type MetaData struct {
 	Key   string `json:"key,omitempty"`
 	Value string `json:"value,omitempty"`
+	Scope string `json:"scope,omitempty"`
 }
 
 // MetaDataExists represents a Buildkite Agent API MetaData Exists check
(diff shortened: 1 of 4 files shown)


toolmantim:
I think this is really exciting!

sj26:
It might be neat to go even finer and allow job-based scopes, too. I can see scenarios where you want to bring a piece of meta-data between docker containers or something.

ineffyble:
This sounds amazing - makes it simple to do things like only run jobs if a checksum of a folder has changed since the last build.

lastthyme (approved):

jdub:
I'm just about to start building a little utility that will do this for me, as it appears you did [with Shopify's Buildkit](https://buildkite.com/blog/terraform-techniques-with-buildkite) a while back.

Any idea why this wasn't merged? Five years is a long time to have a good idea hanging around unfinished.

sj26:
It's a good idea, but the agent api does not currently support these queries.

jdub:
Ha, oh. 😢

PR #498, opened 2017-07-04
Add a message to build output when the agent is killed

lox:
The agent kills it's subprocesses when it's forcefully killed. This currently just looks like "Agent exited with -1" in the user interface.

This adds a message "The agent has been forcefully shutdown, all process will be terminated".


lox:
Decided this wasn't a good idea. It's not something we do anywhere else, debug will do the same.

PR #502, opened 2017-07-12
Add a CHANGELOG.md file

lox:
We are going to follow http://keepachangelog.com/en/1.0.0/

This ports all the release notes into this file. Future releases will happen by diffing this file and putting it in the release notes.


keithpitt:
Looks so impressive when you write it all out like that :)

keithpitt:
Did you use a program to generate it? Or was it done by hand?

lox:
By hand!

Issue #503, opened 2017-07-12
Agent is prompted to verify remote server authenticity when cloning submodule from unkown host


Regner:
Buildkite Agent Version:
```
$ buildkite-agent -v
buildkite-agent version 3.0-beta.28, build 1618
```

When cloning a Git repo from BitBucket that has submodules in GitHub the agent is prompted to add the remote host to the known hosts:
```
Cloning into 'redacted'...
The authenticity of host 'github.com (192.30.255.112)' can't be established.
RSA key fingerprint is 16:27:ac:a5:76:28:2d:36:63:1b:56:4d:eb:df:a6:48.
Are you sure you want to continue connecting (yes/no)?
```

Most of our repos are in BitBucket and cloning those works fine. It appears the agent preemptively adds the remote host to the known hosts (https://github.com/buildkite/agent/blob/master/bootstrap/bootstrap.go#L972) and things are grand. However it doesn't add submodule hosts to the known hosts. I assume if on that agent machine we had a pipeline that used GitHub for the repo, and not the submodule, that would work and then this pipeline would pass.

Regner:
For now I have just manually added the GitHub host to the agents known hosts.

lox:
Implemented! Will get this into a beta soon.

Regner:
Thanks for the quick fix. :)

jam13:
I seem to be hitting this problem on latest buildkite agent (3.4.0).

I've got a Bitbucket repo that has a single submodule from GitHub. When it tries to update the submodules it prompts to accept the GitHub keys and so prevents the build from continuing.

It doesn't fail on every build because an agent instance might have added the GitHub keys to wellknown through another pipeline job, but on newly deployed instances (AWS stack) it fails every time.

Looking at the code, the problem seems to be that it uses the following command to get the submodule URLs which it then uses to update the known_hosts:

```
git submodule foreach --recursive git ls-remote --get-url
```

As this is run _before_ the submodules are initialised, the foreach always returns empty and so no URLs are returned.

PR #509, opened 2017-07-21
Add submodule remotes to known_hosts

lox:
Add submodule remotes to known_hosts as part of the checkout process. 

Fix for #503

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -1066,9 +1066,13 @@ func (b *Bootstrap) Start() error {
 			// submodules might need their fingerprints verified too
 			if b.SSHFingerprintVerification {
 				commentf("Checking to see if submodule urls need to be added to known_hosts")
-				submoduleRepos, _ := b.gitEnumerateSubmoduleURLs()
-				for _, repository := range submoduleRepos {
-					b.addRepositoryHostToSSHKnownHosts(repository)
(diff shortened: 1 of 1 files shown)


bootstrap/bootstrap.go
```diff
@@ -776,6 +776,26 @@ func (b *Bootstrap) gitClean() {
 	}
 }
 
+func (b *Bootstrap) gitEnumerateSubmoduleURLs() ([]string, error) {
+	urls := []string{}
+
+	gitSubmoduleOutput, err := b.runCommandSilentlyAndCaptureOutput(
```
keithpitt: Can we get a copy/paste of what the output looks like?

lox:
More comments added @keithpitt!

bootstrap/bootstrap.go
```diff
@@ -1035,6 +1063,15 @@ func (b *Bootstrap) Start() error {
 		}
 
 		if b.GitSubmodules {
+			// submodules might need their fingerprints verified too
+			if b.SSHFingerprintVerification {
+				commentf("Checking to see if submodule urls need to be added to known_hosts")
+				submoduleRepos, _ := b.gitEnumerateSubmoduleURLs()
```
keithpitt: We should probably show a warning if it fails for what ever reason I think, even if we just show it an continue on our merry way.

Issue #510, opened 2017-07-25
Agent 3.x cannot handle valid YAML pipeline definitions


stefanmb:
Attempting to use `buildkite-agent pipeline upload` in beta 29 with a file like this:

```
base_step: &base_step
  type: script
  agent_query_rules:
    - queue=default

steps:
- <<: *base_step
  name: ':docker: building image'
  command: docker build .
  agents:
    queue: default
```

Results in:

> FATAL  Failed to upload and process pipeline: POST https://agent.buildkite.com/v3/jobs/####/pipelines: 422 `<<` is not a valid property on the `command` step. Please check the documentation to get a full list of supported 

@toolmantim @keithpitt 
/cc: @DazWorrall @wvanbergen

lox:
Investigating this, seems to parse to json ok. Output is:

```json
{
  "steps": [
    {
      "type": "script", 
      "command": "docker build .", 
      "name": ":docker: building image", 
      "agents": {
        "queue": "default"
      }, 
      "agent_query_rules": [
        "queue=default"
      ]
    }
  ], 
  "base_step": {
    "type": "script", 
    "agent_query_rules": [
      "queue=default"
    ]
  }
}
```

stefanmb:
Yeah, I think the issue is it's literally sending "<<", e.g. 
> "steps":[{"\u003c\u003c"

lox:
It's possibly a quirk in the golang yaml parser. I recall there being some nuance around how the yaml extension stuff is handled.

stefanmb:
This used to work ok in Buildkite 2.6

lox:
Yup, entirely possible, I believe we changed how parsing works between them. I will dive into it and get back to you soon.

keithpitt:
@lox I'll have a sneaky look at this now!

PR #511, opened 2017-07-26
Switch to yaml.v2 as the YAML parser

keithpitt:
To fix: https://github.com/buildkite/agent/issues/510

We use github.com/ghodss/yaml as our YAML parser, which in turn uses https://github.com/go-yaml/yaml. For some reason (probably me, but I can't remember) we're using a modified version of ghodss/yaml that uses cloudfoundry-incubator/candiedyaml instead.

Switching to the proper ghodss/yaml version seems to fix the YAML parsing issue as reported in #510. I need to do some more digging why I did it like this in the first place...


keithpitt:
Heh, turns out `ghodss/yaml` tried `cloudfoundry-incubator/candiedyaml` for a while, then switch back to the old version. I guess we added this package at the time they were trailing the new one :)

sj26:
I remember having a go (hah!) at this while back and it turned out more difficult than I thought... but I don't remember why. We should definitely update it, though!

lox:
go-yaml for sure!

stefanmb:
@lox @sj26 @keithpitt Thank you, I can confirm this fixed the problem!

keithpitt:
@stefanmb excellent! I even wrote a test with your example to make sure we don't break it again. It's a good YAML example to make sure that our parser can handle the more gnarly scenarios :)

PR #512, opened 2017-07-26
Cleanup various AWS SDK Bugs

lox:
This adds support for `BUILDKITE_ARTIFACT_UPLOAD_DESTINATION` containing a path (e.g s3://bucket/path). 

It also adds some more debugging around where the region is detected from and clears up a typo with env var names.

### DIFF
--- a/agent/s3.go
+++ b/agent/s3.go
@@ -116,23 +116,36 @@ func awsS3RegionFromEnv() (region string, err error) {
 	return "", fmt.Errorf("Unknown AWS S3 Region %q", regionName)
 }
 
+func awsLogger(args ...interface{}) {
+	if message, ok := args[0].(string); ok {
+		logger.Debug(strings.TrimPrefix(message, "DEBUG: "))
+	}
+}
+
 func newS3Client(bucket string) (*s3.S3, error) {
-	// Generate the AWS config used by the S3 client
 	region, err := awsS3RegionFromEnv()
 	if err != nil {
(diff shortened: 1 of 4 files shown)


agent/s3.go
```diff
@@ -1,41 +1,51 @@
...
 }
 
 func (e *credentialsProvider) Retrieve() (creds credentials.Value, err error) {
 	e.retrieved = false
 
-	creds.AccessKeyID = os.Getenv("BUILDKITE_S3_ACCESS_KEY_ID")
-	if creds.AccessKeyID == "" {
-		creds.AccessKeyID = os.Getenv("BUILDKITE_S3_ACCESS_KEY")
+	if v := os.Getenv(envS3AccessKeyID); v != "" {
+		logger.Debug("Found s3 access key id from " + envS3AccessKeyID)
```
lox: Accidentally dropped support for these variants. Misread this code.

stefanmb:
@lox This fixes the auth problems I've encountered with the Beta agent, thanks!

agent/s3.go
keithpitt: My brain did struggle to understand those changes! I was like "something looks wrong here, but I can't figure it out" - was hard to understand in the first place :)

lox:
Going to split this into a few PRs. Conflated too many things!

lox:
We will get a release out today @stefanmb.

PR #515, opened 2017-08-02
Stop making hook files executable

yeungda-rea:
Hooks are invoked using the `source` keyword in `bootstrap.sh`. Before they are sourced, they are made executable. I believe this is unnecessary:

```
# echo 'date' > /tmp/sourceme
# chmod -x /tmp/sourceme
# source /tmp/sourceme
Wed Aug  2 06:58:06 UTC 2017
```

This unnecessary `chmod` is causing me some trouble when mounting my hooks on a read only filesystem.

Firstly, this error appears before every hook:

> chmod: changing permissions of '/buildkite/hooks/environment': Read-only file system

Secondly, the UI indents things weirdly:

screen shot 2017-08-02 at 4 40 32 pm

Lastly, here's a cat in a sink:

cat-sinks-08

### DIFF
--- a/templates/bootstrap.sh
+++ b/templates/bootstrap.sh
@@ -115,9 +115,6 @@ function buildkite-hook {
   HOOK_SCRIPT_PATH="$2"
 
   if [[ -e "$HOOK_SCRIPT_PATH" ]]; then
-    # Make sure the script path is executable
-    chmod +x "$HOOK_SCRIPT_PATH"
-
     # Print to the screen we're going to run the hook
     echo "~~~ Running $HOOK_LABEL hook"
     buildkite-prompt "$HOOK_SCRIPT_PATH"


lox:
Thanks for this @yeungda-rea, especially the cat picture, I lol'd :) 

We make hook files executable because often people forget to do that and it saves a lot of failed builds. I can see how a read-only filesystem would complicate that though. Perhaps we could check first and only make the change if needed?

yeungda-rea:
Hey @lox, I certainly don't want builds to fail either! The snippet above attempts to show that `source` is safe to run on files without executable permission. It's okay to forget to make the hook executable because it doesn't need to be. Are you sure that builds will fail if we don't `chmod` hooks? What would make them fail?

Keep in mind that this change is different from the [command `chmod`](https://github.com/yeungda-rea/agent/blob/7fb96a1784330db85e9d0205b89f8684d1cccb7f/templates/bootstrap.sh#L433-L436). The command [executes a file specifically](https://github.com/yeungda-rea/agent/blob/7fb96a1784330db85e9d0205b89f8684d1cccb7f/templates/bootstrap.sh#L63) and for that `chmod` is required. Not a problem for us because because the source code does not have any read only constraints.

Anyway, if we can't determine that this change is safe, I'd be happy to wrap that `chmod` in some test for executability instead.

lox:
I should have read the diff first. I think you're correct, the executable flag should be unneeded if it's sourced. @keithpitt can you think of any reason why we'd need the executable flag on hooks?

toolmantim:
I've noticed [we do this in master too](https://github.com/buildkite/agent/blob/6a9608216cb395cda21ce42cf6f50f2437fc78f3/bootstrap/bootstrap.go#L506-L507)

lox:
In master it's an actual script though. Will be interesting to see whether v3 works on the read-only fs.

yeungda-rea:
Here is a brief description of the process on `master`:

1. create a temporary file buildkite-agent-bootstrap-hook-runner

```
tempHookRunnerFile := createTempFile(normalizeScriptFileName("buildkite-agent-bootstrap-hook-runner"))
```

2. make the temporary file executable

```
addExecutePermissiontoFile(tempHookRunnerFile.Name())
```

3. write a script into the temporary file that sources the hook

```
tempHookRunnerFile.WriteString(hookScript)
```

4. launch a shell that launches the temporary file:

```
cmd = b.newCommand("/bin/bash", "-c", `"`+strings.Replace(command, `"`, `\"`, -1)+`"`)
```

yeungda-rea:
By my interpretation of the process on `master`, the original hook script is left alone, which is the behavior I desire for a hook that lives on a read only filesystem.

yeungda-rea:
FWIW, the v3 implementation described will cause problems for anyone who runs a hook and mounts their `TMPDIR` with `noexec`. I believe mounting `/tmp` with `noexec` is a common hardening practice.

This example shows that if `/tmp` is mounted `exec`, I can copy `date` to `/tmp/date` and execute it:

```
# docker run --tmpfs /tmp:exec busybox sh -c 'mount | grep '/tmp' && cp /bin/date /tmp/date && /tmp/date'
tmpfs on /tmp type tmpfs (rw,nosuid,nodev,relatime)
Fri Aug  4 02:15:23 UTC 2017
```

If I do the same, using `noexec` (docker's default for tmpfs) I am not permitted to run `/tmp/date`:

```
# docker run --tmpfs /tmp busybox sh -c 'mount | grep '/tmp' && cp /bin/date /tmp/date && /tmp/date'     
tmpfs on /tmp type tmpfs (rw,nosuid,nodev,noexec,relatime)
sh: /tmp/date: Permission denied
```

lox:
The scripts are interpretted with /bin/bash though, rather than directly executed, so they shouldn't need either an executable bit and they shouldn't have problems with `noexec`

keithpitt:
That indentation is pretty weird..not sure what's going on there!

I'm 👍 on this change. The `chmod` was introduced when we rewrote how hooks are handled (see this commit: https://github.com/buildkite/agent/commit/2f064eae4d940877ba04b0a65df4d621315a8568#diff-9f790e29c0786af0f7faff0d01153a21R101)

When I first build hooks, I `source`'d them. Then someone was like "if a hook exits, it exits the whole thing" so I tried changing it to `execute` - where I added the `chmod`. Then we realised that environment variables stopped working between hooks and the main script, so we reverted (but left in the chmod) as seen here: https://github.com/buildkite/agent/commit/b9ea282f5fa787d4ccc508558f6fd9818d118fe7

So I'm like 99.5% sure that the `chmod` is a left over from a time where we used to execute hooks in the bootstrap. I'm not sure of what it does in bootstrap.go just yet, but that'll need to be handled in a seperate change.

keithpitt:
@yeungda-rea do you need another Agent release? Or are you happy to wait for this to be included with the next one?

yeungda-rea:
@keithpitt we'll be maintaining a patched `bootstrap.sh` until the next release. I don't mind when you release. Just glad to know that we won't have to maintain it forever.

yeungda-rea:
Thank you all for help in reviewing and helping to get this change merged. I really appreciate it.

Issue #504, opened 2017-07-14
More streamlined checkout options for phabricator repos


KevinGrandon:
Currently for Phabricator the first thing checkout does is: `git fetch -v origin <commit>`

If the commit is not found, it fetches all tags with:
`git fetch -v origin +refs/heads/*:refs/remotes/origin/* +refs/tags/*:refs/tags/*`

This fetch can take a long time for larger repos and in the case of phabricator should not be necessary. Instead we would like to be able to directly fetch the tag with something like:
`git fetch origin "${BUILDKITE_BRANCH}:"${BUILDKITE_BRANCH}"`

I think we may be able to do this with a custom checkout hook, but since it may be beneficial to everyone using Phabricator, an improvement in the agent would be nice.

toolmantim:
Sorry you're having troubles with this.

I think it's because your Phabricator git server doesn't supporting [fetching a specific git commit](https://stackoverflow.com/questions/14872486/retrieve-specific-commit-from-a-remote-git-repository/30701724#30701724), which is making [this](https://github.com/buildkite/agent/blob/6a9608216cb395cda21ce42cf6f50f2437fc78f3/bootstrap/bootstrap.go#L1058) return a non-zero status. I don't think is specific to Phabricator at all… any git server that doesn't support that, but has lots of tags, would have this performance problem :(

One workaround for now is to use the bootstrap’s [custom refspec support](https://github.com/buildkite/agent/blob/6a9608216cb395cda21ce42cf6f50f2437fc78f3/bootstrap/bootstrap.go#L1025-L1030).

You can't set the `BUILDKITE_REFSPEC` environment variable in a hook, but you can do it by wrapping the bootstrap. Which is nicer and easier than overriding the `checkout` hook.

Here's an example, which I've tested and it seems to work okay:
https://gist.github.com/toolmantim/43de11658602b3f7c0e1921fcea7ac5f

Would that suit as a workaround for now?

I just opened PR #517 to see if we could instead just add a `fetch` hook, rather than having to rely on this custom refspec support and bootstrap wrapper.

KevinGrandon:
Thank you for the detailed response. We are indeed running an older git server so that makes sense why it isn't able to fetch the commit. I'm going to put in a request to upgrade this server, as well as implement the bootstrap script in parallel.

KevinGrandon:
We have a workaround for now, so going to close this. Hopefully folks might find this useful in the future if they stumble across this ticket.

PR #517, opened 2017-08-03
Add support for a fetch hook

toolmantim:
If people want to optimise the `git fetch` for performance sensitive or quirky environments (e.g. #504), it requires overriding the entire `git checkout` hook. Even if we exposed the original checkout hook for calling, it wouldn't allow you to optimise just the `git fetch`.

To help, this adds a new hook called `fetch` allowing you to provide your own logic.

- [x] Code
- [ ] Feedback
- [ ] Run some tests

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -1022,9 +1022,19 @@ func (b *Bootstrap) Start() error {
 		// Git clean prior to checkout
 		b.gitClean()
 
+		// If they've provided their own optimized fetch
+		if fileExists(b.globalHookPath("fetch")) {
+			b.executeGlobalHook("fetch")
+			b.runCommand("git", "checkout", "-f", b.Commit)
+
+		// If a plugin has provided it's own optimized fetch
+		} else if b.pluginHookExists(plugins, "fetch") {
+			b.executePluginHook(plugins, "fetch")
+			b.runCommand("git", "checkout", "-f", b.Commit)
+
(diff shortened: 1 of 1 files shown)


lox:
Seems sensible to me! I assume if there is a checkout hook it trumps the fetch hook?

keithpitt:
Hrm... I'm not 100% sold that adding another hook is the best way to approach this one. Our fetch logic is already quite complex - and tries to do the right thing depending on the situation. In this particular case, you'd need to reimplement the whole thing in bash, just to change how the "catch-all" method works.

Maybe we should try `git fetch origin "${BUILDKITE_BRANCH}:"${BUILDKITE_BRANCH}"` in the bootstrap before then doing a "fetch all the things and hope the commit is in there" approach? I'm not 100% over the fetch logic, @sj26 is the expert there - but my vote would be to just improve the current fetch logic - instead of a custom hook.

toolmantim:
Yeah, and given the particular problem in #504 is for older git servers, and wanting to disable the "fetch all tags" behaviour, it's probably not a great hook to add moving forward!

Probably better to try to add some configuration behaviour, or all you to use an `environment` hook to set the `BUILDKITE_REFSPEC` environment variable?

keithpitt:
@toolmantim https://github.com/buildkite/agent/releases/tag/v3.0-beta.30

toolmantim:
@keithpitt nice!! Thank you 👏🏼

Issue #507, opened 2017-07-21
"Preparing build directory" step should also (probably) prune remote branches


SEAPUNK:
Sometimes, if I create a branch called `foo/bar/baz`, where branch `foo/bar` exists, I cannot push the branch to the remote Github repo. So, because `foo/bar` on the Github repo isn't used, I delete that branch, run `git remote prune origin`, and then push my branch. This causes a problem running `git fetch` on the build box:

```
  | $ git clean -fdq
  | $ git submodule foreach --recursive git clean -fdq
  | $ git fetch -v origin aaaaaaaaa442166ea29ce572f4cbc6385b5a9266
  | $ git fetch -v origin +refs/heads/\*:refs/remotes/origin/\* +refs/tags/\*:refs/tags/\*
  | remote: Counting objects: 25, done.
  | remote: Compressing objects: 100% (5/5), done.
  | remote: Total 25 (delta 21), reused 24 (delta 20), pack-reused 0
  | Unpacking objects: 100% (25/25), done.
  | error: 'refs/remotes/origin/foo/bar' exists; cannot create 'refs/remotes/origin/foo/bar/baz'
  | From github.com:some/repo
  | ! [new branch]      foo/bar/baz -> origin/foo/bar/baz  (unable to update local ref)
```

causing all of the subsequent builds to fail.

Because of this, to resolve this issue, maybe the agent should run `git remote prune origin` shortly after running `git clean`, to prevent this error from occurring?

cysp:
Alternatively, `git fetch` takes a `--prune` option which could combine those steps.

lox:
Closed in #518

PR #518, opened 2017-08-04
Prune when fetching with a refspec

cysp:
https://github.com/buildkite/agent/issues/507

The `Fetch and checkout commit` codepath (the last one) is the one that I actually saw the problem affect and this change fix, but the other paths look like they might get caught by the same scenario.
I left the `refs/pull/…/head` and `commit-sha` fetches unchanged because I figure that they wouldn't need the pruning.

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -1026,7 +1026,7 @@ func (b *Bootstrap) Start() error {
 		// i.e. `refs/not/a/head`
 		if b.RefSpec != "" {
 			commentf("Fetch and checkout custom refspec")
-			b.runCommand("git", "fetch", "-v", "origin", b.RefSpec)
+			b.runCommand("git", "fetch", "-v", "--prune", "origin", b.RefSpec)
 			b.runCommand("git", "checkout", "-f", b.Commit)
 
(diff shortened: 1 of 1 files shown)


keithpitt:
Ah cool, thanks for the change! Which versions of Git is `--prune` available in? Do we need to do any sort of version checking?

keithpitt:
@sj26 has done a ton of work in that section of the bootstrap, so I'll let him chime in and see if it's cool to merge. But looks OK to me! 👌

cysp:
I took a look at when it was introduced and saw that it was eight years ago so figured it would be okay.
https://github.com/git/git/commit/f360d844de4b752b4cba2343b95592ae4d55d09a

Is that okay?

keithpitt:
Oh yeah, I think that's totally fine (I was just wondering because if was a feature included with the latest release of Git, it'd be hard to get everyone to convert). I'm actually not sure what our lowest supported version is, but we should find out and make that clear in the bootstrap!

sj26:
We should backport this to 2-6-stable in the bootstrap.sh, too.

lox:
Thanks @cysp!

pda:
Thanks — I just ran into this issue, saves a pull request ;)

PR #522, opened 2017-08-15
Consolidate Environment into env package

lox:
As part of gradually refactoring the bootstrap package, this consolidates shell.Environment and envvar into an env package, which aims to provide a map for environment vars and then ways to interact with that env, e.g Interpolate.

This takes the guts out of `shell`, which means I can refactor what is left into the bootstrap package as a `Shell` struct and a lightweight `Subprocess`.


env/environment.go
```diff
@@ -0,0 +1,142 @@
...
+// There's a mix of both CamelCase and UPPERCASE, but the can all be accessed
+// regardless of the case you use. So PATH is the same as Path, PAth, pATH,
+// etc.
+//
+// os.Environ() in Golang returns key/values in the original casing, so it
+// returns a slice like this:
+//
+//     { "Path=...", "PROCESSOR_IDENTIFIER=...", "SystemRoot=..." }
+//
+// Users of shell.Environment shouldn't need to care about this.
```
keithpitt: Gotta fix comment here

PR #523, opened 2017-08-16
Move shell package to bootstrap/shell

lox:
This is an incremental step towards more of a "virtual shell", like we discussed @keithpitt. It's meh as it is, but it starts to herd shell functionality into the shell subpackage under the bootstrap.


process/process.go
```diff
@@ -18,7 +18,7 @@ import (
 	"time"
 
 	"github.com/buildkite/agent/logger"
-	"github.com/buildkite/agent/shell"
+	"github.com/mattn/go-shellwords"
```
keithpitt: I think we have `shellx.Split` as well which probably does the same thing?

lox: Hrm, this was what was in the Command.ParseString:

```go
// Creates a command by parsing a string like `ls -lsa`
func CommandFromString(str string) (*Command, error) {
	args, err := shellwords.Parse(str)
	if err != nil {
		return nil, err
	}

	return &Command{Command: args[0], Args: args[1:]}, nil
}
```

PR #524, opened 2017-08-18
Disallow local hooks if BUILDKITE_NO_LOCAL_HOOKS is set.

solemnwarning:
Useful for preventing hostile hooks in the repository from escaping whatever
chroot/container/etc mechanism may be in use around the build commands.

### DIFF
--- a/templates/bootstrap.sh
+++ b/templates/bootstrap.sh
@@ -150,6 +150,10 @@ function buildkite-global-hook {
 }
 
 function buildkite-local-hook {
+  if [[ -e ".buildkite/hooks/$1" ]] && [[ "${BUILDKITE_NO_LOCAL_HOOKS:-}" == "true" ]]; then
+    buildkite-error "Attempted to run .buildkite/hooks/$1, but BUILDKITE_NO_LOCAL_HOOKS is enabled"
+  fi
+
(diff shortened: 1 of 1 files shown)


lox:
I'm all for this, thoughts @keithpitt? 

Eventually I'd love to roll together some security settings in to a `--hardened` mode.

keithpitt:
This is interesting! Couldn't a hostile person just change the build script contents and pretty much circumvent the whole thing anyway?

i.e I want to get a `rm -rf` into the build process. You've disabled hooks, so I can't insert one there. But I can just add it to `script/ci` and do the same thing. If you've got access to the code, you can pretty much do what evs...

Unless this is protect against another sort of vulnerability that I'm not seeing? I may need more ☕️

solemnwarning:
The pipeline configuration runs the build scripts/etc inside a chroot/container-y thing, so whatever they do it can't have any influence outside of the build directory.

lox:
The thing I like about this is that local hooks are a way to override global hooks, which if you are trying to lock things down is a concern. You can run your command hook in a chroot/container/VM, but unless you run your checkout in there then there is a spot for rogue hooks to execute outside that protection.

lox:
@keithpitt and I discussed this and are 👍 

Will port to master too. Thanks @solemnwarning

PR #525, opened 2017-08-19
Expand shell abstraction

lox:
This builds on #523 and establishes shell.Shell as a "virtual shell" that handles all of the logging and command execution for the bootstrap. It's far larger a PR then I am normally comfortable with, but it's easier to evaluate as a whole then #523 was in isolation. Happy to take suggestions on splitting it up.

Major changes:
 * Moves the `shell` package into `bootstrap/shell` (#523)
 * Adds a `shell.Shell` abstraction which represents a shell that things are executed in. Handles output, and the pretty logging output.   
* Split `bootstrap.Start` into `phase` methods that returns errors rather than exit
* Split out `bootstrap.executeHook` into a `Hook` struct. 

The design goals I've been trying to hold onto throughout this is are:

* Generalize the execution of hooks so that we can more gracefully handle cancellation signals in future
* Make shell.Shell really useful for third-party (plugin) consumption
* Provide a good spot for mocking out commands VCR-style to allow integration testing the `bootstrap` package. 
* Allow for subcommands of `bootstrap` to invoke particular phases, along with setup and teardown

Major questions I have left:

 1. I hate `shell.RunCommand` and `shell.RunCommandSilentlyAndCaptureOutput`. I'd love one method that could do both, or a better name for the latter. Suggestions welcomed.
 2. Is passing the `shell.Shell` instance into things like the `Hook` and `knownHosts` structs gross?
 3. What abstraction could we wrap up the finding and executing of hooks in? I tried out a `HookSet` with an `Apply` method, but got caught up on wondering about question 2. 
4. Should I defer the `Hook` change to a later PR once 4 is resolved?


sj26:
I really like this idea! I also dislike the name "RunCommandSilentlyAndCaptureOutput" but don't have better alternatives. Ooh, `CaptureCommandOutput`?

lox:
Thanks @sj26, check out #530.

PR #526, opened 2017-08-22
Add .editorconfig and set indent_size to 4

marius92mc:
I saw that the repository doesn't have a config file for specifying GitHub to treat tabs with less than 8 spaces, which is GitHub's default behavior.

This PR changes this default behavior and introduces the possibility to specify the number of spaces GitHub to use as indent size for `*.go` files.


lox:
Thanks! I didn't realize that GitHub respected .editorconfig.

PR #527, opened 2017-08-23
Add a dockerfile for local development

lox:
This adds a local Dockerfile that will compile the agent during development. 

It doesn't aim to replace https://github.com/buildkite/docker-buildkite-agent

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -2,7 +2,7 @@ steps:
   - name: ":hammer:"
     command: "scripts/tests.sh"
     plugins:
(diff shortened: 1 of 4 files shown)


lox:
@toolmantim https://docs.docker.com/compose/extends/#multiple-compose-files

> By default, Compose reads two files, a docker-compose.yml and an optional docker-compose.override.yml file. By convention, the docker-compose.yml contains your base configuration. The override file, as its name implies, can contain configuration overrides for existing services or entirely new services.

I tend to use it for development specific config that you don't want applied in CI.

PR #529, opened 2017-08-25
Support git clean and git clone flags with spaces

lox:
Previously the GIT_CLEAN_FLAGS and GIT_CLONE_FLAGS were wrapped in quotes, which doesn't work if there are multiple flags or they have spaces in them. This wordsplits them into an array and then interpolates them as multiple words.

An example of what this breaks is the long args string that the secrets plugin puts in:

https://github.com/buildkite-plugins/s3-secrets-buildkite-plugin/blob/85347a8ca961e0eb43356188cd5ab3704a134961/hooks/environment#L111-L112

### DIFF
--- a/templates/bootstrap.sh
+++ b/templates/bootstrap.sh
@@ -158,11 +158,11 @@ function buildkite-local-hook {
 }
 
 function buildkite-git-clean {
-  BUILDKITE_GIT_CLEAN_FLAGS=${BUILDKITE_GIT_CLEAN_FLAGS:--fdq}
-  buildkite-run git clean "$BUILDKITE_GIT_CLEAN_FLAGS"
+  BUILDKITE_GIT_CLEAN_FLAGS=(${BUILDKITE_GIT_CLEAN_FLAGS:--fdq})
+  buildkite-run git clean "${BUILDKITE_GIT_CLEAN_FLAGS[@]}"
 
   if [[ -z "${BUILDKITE_DISABLE_GIT_SUBMODULES:-}" ]]; then
-    buildkite-run git submodule foreach --recursive git clean "$BUILDKITE_GIT_CLEAN_FLAGS"
+    buildkite-run git submodule foreach --recursive git clean "${BUILDKITE_GIT_CLEAN_FLAGS[@]}"
   fi
 }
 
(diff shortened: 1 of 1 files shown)


ticky:
Hmm, I’m not familiar with this use of bash arrays, does this duplicate the quote marks for each element?

lox:
Yup, it's bash black magic, but it's the standard way of doing this:

http://www.gnu.org/software/bash/manual/bashref.html#Special-Parameters

> *
> ($*) Expands to the positional parameters, starting from one. When the expansion is not within double quotes, each positional parameter expands to a separate word. In contexts where it is performed, those words are subject to further word splitting and pathname expansion. When the expansion occurs within double quotes, it expands to a single word with the value of each parameter separated by the first character of the IFS special variable. That is, "$*" is equivalent to "$1c$2c…", where c is the first character of the value of the IFS variable. If IFS is unset, the parameters are separated by spaces. If IFS is null, the parameters are joined without intervening separators.

lox:
Urgh I pasted the wrong one.

lox:
> @
> ($@) Expands to the positional parameters, starting from one. When the expansion occurs within double quotes, each parameter expands to a separate word. That is, "$@" is equivalent to "$1" "$2" …. If the double-quoted expansion occurs within a word, the expansion of the first parameter is joined with the beginning part of the original word, and the expansion of the last parameter is joined with the last part of the original word. When there are no positional parameters, "$@" and $@ expand to nothing (i.e., they are removed).

ticky:
Sounds legit, how many bash versions back is it compatible with? (Remembering macOS ships 3.2!)

lox:
I think since the early 90's :)

ticky:
Sounds legit, although given we’re converting a string into an array, splitting on spaces, is this any different, functionally, than just passing the unquoted string to the command?

lox:
It should make it considerably harder to do things like:

export GIT_CLEAN_FLAGS="; rm -rf /"

ticky:
That sort of command logic shouldn’t be possible from the middle of a string argument, though?

lox:
Blergh, this is harder than I thought. http://rextester.com/edit/DUAAT17379

lox:
> That sort of command logic shouldn’t be possible from the middle of a string argument, though?

Wouldn't it? If it wasn't quoted it would work.

PR #530, opened 2017-08-26
Bootstrap rewrite

lox:
This is rewrite/refactor of the bootstrap to split it into layers of abstraction. This will let us write tests and also invoke things like `bootstrap checkout` and `bootstrap command` directly. 

The shell abstraction should also be broadly useful in plugins.

Builds on (and replaces) #523 and #525.


bootstrap/hook.go
```diff
@@ -0,0 +1,126 @@
...
+// environment set for them and then we capture any extra environment variables
+// that are exported in the script.
+//
+// The tricky thing is that it's impossible to grab the ENV of a child process
+// before it finishes, so we've got an awesome (ugly) hack to get around this.
+// We write the ENV to file, run the hook and then write the ENV back to another file.
+// Then we can use the diff of the two to figure out what changes to make to the
+// bootstrap. Horrible, but effective.
+
+type hookRunner struct {
```
lox: hookScriptBuilder

bootstrap/hook.go
```diff
@@ -0,0 +1,126 @@
...
+	return h.scriptFile.Name()
+}
+
+func (h *hookRunner) Close() {
+	os.Remove(h.scriptFile.Name())
+	os.Remove(h.beforeEnvFile.Name())
+	os.Remove(h.afterEnvFile.Name())
+}
+
+// Environment returns and environment variables exported during the hook run
```
lox: More story

bootstrap/git.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"strings"
+
+	"github.com/buildkite/agent/bootstrap/shell"
+)
+
+func gitClean(sh *shell.Shell, gitCleanFlags string, gitSubmodules bool) error {
+	// Clean up the repository
+	if err := sh.Run("git clean %s", gitCleanFlags); err != nil {
```
lox: Needs to be Quoted

Issue #528, opened 2017-08-24
Add BUILDKITE_GCS_ACCESS_URL env variable


mcgain:
There doesn't seem to be an equivalent of `BUILDKITE_S3_ACCESS_URL` for GCS. At the moment, GCS returns a url of `storage.googleapis.com` which are designed for API usage. Humans can fetch objects on GCS with a `storage.cloud.google.com` url. Respecting `BUILDKITE_GCS_ACCESS_URL` would allow human users to just click on the links in buildkite and be able to download artifacts. 

For machine users, we would still have to substitute `storage.googleapis.com` for `storage.cloud.google.com` but that's ok. To solve the problem completely, it would be nice to differentiate human download urls from machine ones. `BUILDKITE_GCS_ACCESS_URL` vs `BUILDKITE_GCS_API_ACCESS_URL` for example. This is a secondary concern though.

(GCS was added in #207 so cc @grosskur )

lox:
Closing now that we have #532, @mcgain let me know if this doesn't provide a workable solution in the short term.

PR #532, opened 2017-08-29
Add BUILDKITE_GCS_ACCESS_HOST for GCS Host choice

jules2689:
Problem
---
- The URL for artifacts uploaded to GCS are api links.
- Following the link in browser yields an XML error page.
- The URL buildkite puts is under `storage.googleapis.com` but to leverage the oath prompt mentioned above users need to access objects under `storage.cloud.google.com`.
- I couldn't find anything from buildkite's doc equivalent to BUILDKITE_S3_ACCESS_URL for Google Cloud.

Alternatives
---
- When a user clicks a link in browser, you should use `storage.cloud.google.com`.
  - This flow is for humans, not machines - so it should be optimized for that experience
- When a user fetches a link via API, you should use `storage.googleapis.com`
  - This flow is for machines, not humans - so it should be optimized for that experience

This change would be simple but allow for us to easily leverage oAuth and API when needed.

Looking at our internal issue again, I think @mcgain has asked about this - so best to get his input 😄  I saw this pattern in S3 Uploader, so I copied. 

cc @keithpitt @toolmantim

### DIFF
--- a/agent/gs_uploader.go
+++ b/agent/gs_uploader.go
@@ -50,10 +50,14 @@ func (u *GSUploader) Setup(destination string, debugHTTP bool) error {
 }
 
 func (u *GSUploader) URL(artifact *api.Artifact) string {
-	// If not publicly accessible, GET on this URL results in 403.
+	host := "storage.googleapis.com"
+	if os.Getenv("BUILDKITE_GCS_ACCESS_HOST") != "" {
+		host = os.Getenv("BUILDKITE_GCS_ACCESS_HOST")
+	}
+
 	var artifactURL = &url.URL{
 		Scheme: "https",
-		Host:   "storage.googleapis.com",
+		Host:   host,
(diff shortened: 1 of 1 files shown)


mcgain:
In #528 I suggested `BUILDKITE_GCS_ACCESS_URL` to be consistent with `BUILDKITE_S3_ACCESS_URL`.

I agree that the alternatives are preferable, but I'm not sure whether they scale to other providers.

jules2689:
Ah yea, I chose `_HOST` here to be consistent with the use of the env

lox:
Sorry for the slow response!

This seems like a good incremental solution to me. We have similar issues with `BUILDKITE_S3_ACCESS_URL` where we need to come up with a way to serve that url, but signed for authentication. When we tackle that, we can figure out a way to improve the click to download experience for the GCS stuff too.

PR #538, opened 2017-09-02
Add a windows testing step

lox:
This adds a step that runs the tests on a Windows 2016 w/ Containers instance in another asynchronous pipeline. This means we can still see windows failures while we iterate on fixing them.

For the actual infrastructure, I created a hand-rolled m4.large windows instance running Windows 2016 Server. Will replace with something more scalable when we need it.


toolmantim:
👏🏼

lox:
The windows tests are failing, but my thinking is if we get this in then I can incrementally ship fixes for the various breakages.

PR #539, opened 2017-09-05
Replace glob package with go-zglob

lox:
This replaces our glob package with the excellent https://github.com/mattn/go-zglob. It's faster, well tested and supports windows.

I moved the old `glob.Root` function logic around relative paths into a story in the `artifact_uploader`.


agent/artifact_uploader.go
```diff
@@ -53,55 +58,79 @@ func (a *ArtifactUploader) Upload() error {
...
+			// Ignore directories, we only want files
+			if isDir(absolutePath) {
+				logger.Debug("Skipping directory %s", file)
+				continue
+			}
 
-				if isDir, _ := glob.IsDir(absolutePath); isDir {
-					logger.Debug("Skipping directory %s", file)
-					continue
+			// If a glob is absolute, we need to make it relative to the root so that
```
lox: /cc @keithpitt

lox:
Note that this also closes out the file handles open bug that Airbnb surfaced.

PR #540, opened 2017-09-05
Add --no-plugins option to agent

lox:
This adds a `--no-plugins` option to an agent, which causes as step to fail with an error if there is a plugin present:

```
Error: This agent is isn't allowed to run plugins. To allow this, re-run this agent without the `--no-plugins` option.
```

A little bit of extra technology was needed to clean up how errors are wrapped and returned from the bootstrap. Previously there was a duplication of errors from when the command failed and when the phase failed.


.gitignore
```diff
@@ -1,13 +1,12 @@
...
+/vendor/bundle
+/vendor/pkg
 .bundle
 .envrc
-.agent.conf
-.agent.*.conf
-.buildkite-agent.cfg
+/.agent.conf
+/.agent.*.conf
+/.buildkite-agent.cfg
```
sj26: :+1:

bootstrap/bootstrap.go
```diff
@@ -363,10 +358,19 @@ func (b *Bootstrap) PluginPhase() error {
 		return fmt.Errorf("Can't checkout plugins without a `plugins-path`")
 	}
 
+	if b.Debug {
+		b.shell.Commentf("Plugin JSON is %s", b.Plugins)
+	}
+
+	// Check if we can run plugins (disabled via --no-plugins)
+	if b.Plugins != "" && !b.Config.PluginsEnabled {
+		return fmt.Errorf("This agent is isn't allowed to run plugins. To allow this, re-run this agent without the `--no-plugins` option.")
```
sj26: Is there an extra "is" in here? ;-)

lox: There is is. Good catch.

bootstrap/shell/shell.go
```diff
@@ -42,7 +43,7 @@ type Shell struct {
 func New() (*Shell, error) {
 	wd, err := os.Getwd()
 	if err != nil {
-		return nil, fmt.Errorf("Failed to find current working directory: %v", err)
+		return nil, errors.Wrapf(err, "Failed to find current working directory: %v")
```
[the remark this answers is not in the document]
lox: Yeah, me too. They are really nice.

PR #543, opened 2017-09-06
Remove confusing docker compose override

lox:
Based on a conversation with @toolmantim, the `docker-compose.override.yml` is surprising and results in different behaviour from CI. I agree. 

This rolls it back to something simpler.


lox:
I'm working on my stories, thank you for noticing :)

PR #545, opened 2017-09-06
Support docker environment vars from v2

lox:
This adds support for everything in https://buildkite.com/docs/builds/docker-containerized-builds, but also displays warnings.

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -364,8 +364,8 @@ func (b *Bootstrap) tearDown() error {
 	}
 
 	// Support deprecated BUILDKITE_DOCKER* env vars
-	if hasDockerEnvironment(b.shell) {
-		return tearDownDockerIntegration(b.shell)
+	if hasDeprecatedDockerIntegration(b.shell) {
+		return tearDownDeprecatedDockerIntegration(b.shell)
 	}
 
 	return nil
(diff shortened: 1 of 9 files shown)


bootstrap/bootstrap.go
```diff
@@ -96,26 +96,6 @@ func (b *Bootstrap) Start() int {
 	return exitStatus
 }
 
-// executeScript executes a script in a Shell, but the target is an interpreted script
```
lox: I moved this to the Shell, as it's shell related and a pretty common thing to want to do.

bootstrap/bootstrap.go
```diff
@@ -379,6 +363,11 @@ func (b *Bootstrap) tearDown() error {
 		return err
 	}
 
+	// Support deprecated BUILDKITE_DOCKER* env vars
+	if hasDockerEnvironment(b.shell) {
```
keithpitt: Can we call it `hasDeprecatedDockerEnvironment` just to make it clear that it shouldn't be considered the future?

lox: Hah yup, I called it that initially but was worried it was too verbose.

keithpitt: Verbosity is my fav.

lox: Ok, done.

PR #546, opened 2017-09-13
Stable Agent Bootstrap Integration tests

lox:
This adds integration tests for the bootstrap, which will also be runnable on the v3 agent. 

Feels welcomed.


toolmantim:
I don't have much to add on this one I'm afraid. I'll leave it to you and @keithpitt

Issue #544, opened 2017-09-06
Global environmental variables are not immediately available for interpolation in pipeline.yml files. files.


neil-m-donnelly:
Hi there!

Whilst using a version 3 buildkite agent; if I define a pipeline.yml file like so:

```
env:
  VERSION: "0.0.1"
steps:
  - command: build-stuff
    label: "build ${VERSION} from ${BUILDKITE_BRANCH}"
```
Then I'd expect that the globally defined environmental variables in the pipeline file would be available for interpolation within the file so that the label would display the defined version and the branch. As it stands only the BUILDKITE_BRANCH variable is resolved.

I can get around this by defining a second pipeline file, which calls onto my original file like so:
```
env:
  VERSION: "0.0.1"
steps:
  - command: buildkite-agent pipeline upload .buildkite/pipeline.yml
    label: "Init Env vars"
```

At this point, the environmental variable has been exported into the shell and is available for interpolation. I think it might be cuter if globally defined environmental vars in a pipeline file are first exported into the shell before the interpolation step is performed on the v3 file.

Many thanks! :D

lox:
Thanks for the detailed report! This is by design, but ends up being confusing, I've run into this myself several times as well. 

I'll see how hard this would be to fix.

neil-m-donnelly:
Thanks for looking into it. :)

keithpitt:
Yeah I've actually hit this issue myself, and I wrote the interpolation gear!

I don't think it'd be too hard since we parse the YAML file in Golang, @lox we just need to copy the `env` and special case it.

Although, here's a funky issue (maybe)

```
env:
  MESSAGE: "omg tasty $FOOD"

steps:
  - command: "echo $MESSAGE"
```

We'll need to parse `env` first, get the resulting variables - then do a parse of `steps`

neil-m-donnelly:
I think parsing the env first and then doing the steps would be a way to go with it. As a mostly dumb user; if I'm writing a file where I can specify global environmental variables and interpolate shell variables then I'd expect that I could use the shell variables to make up part of the other variables.

In fact, being a completely dumb user I'd probably think I could do something like this:

```
env:
  TEAM1: "England"
  TEAM2: "Australia"
  HEADLINE: "${TEAM1} smashes ${TEAM2} to win the ashes in ${YEAR_FROM_SHELL}!!"
steps:
  - command: "echo ${HEADLINE}"
```

I could then re-use the composite headline in every build step. ~.^x

[...]

PR #552, opened 2017-09-21
Interpolate env block first

lox:
A quick implementation of #544


lox:
Depends on https://github.com/buildkite/agent/pull/555

PR #555, opened 2017-09-25
Make Interpolate a method of Environment

lox:
This removes another spot where we rely on system environment variables. Makes testing lots easier.

### DIFF
--- a/env/interpolator.go
+++ b/env/interpolator.go
@@ -2,7 +2,6 @@ package env
 
 import (
 	"fmt"
-	"os"
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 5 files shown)


env/interpolator.go
```diff
@@ -2,156 +2,162 @@ package env
...
 )
 
 var variablesWithBracketsRegex = regexp.MustCompile(`([\\\$]?\$\{([^}]+?)})`)
 var variablesWithNoBracketsRegex = regexp.MustCompile(`([\\\$]?\$[a-zA-Z0-9_]+)`)
-
 var substringRegexp = regexp.MustCompile(`\A\s*:\s*(\-?\s*\d+)(?:\s*:\s*(\-?\s*\d+))?\s*\z`)
 
-func Interpolate(str string) (interpolated string, err error) {
+// Interpolate takes a string and interpolates shell-like variable expansions in the provided string
+func (e Environment) Interpolate(str string) (string, error) {
```
lox: This is functionally the same, I just used the go pattern of "fail fast" to reduce nesting depth.

env/interpolator.go
```diff
@@ -2,156 +2,162 @@ package env
...
+			// In the second form, length is the number of characters from offset
+			// to select. If negeative, length is instead an offset from the end
+			// of $parameter.
+			//
+			match := substringRegexp.FindStringSubmatch(option)
+			lenvv := int64(len(vv))
+
+			offset, parseErr := strconv.ParseInt(match[1], 10, 0)
+			if err != nil {
+				err = parseErr
```
lox: Actually this is a change, if the parseInt fails, it returns an error, before it just logged it

PR #556, opened 2017-09-27
Remove corrupted previous checkouts

### LINKED ISSUE #533: Issue with code checkout on buildkite agents
There seems to be an issue with code checkout on agents on rare occasion where the .git folder is not created correctly. When this happens on an agent, it means that every build for that pipeline on that agent will fail.

This can be fixed by adding a global hook to check and fix the workspace if it detects it is in this state, but it should really be resolved at the root of the issue.

I have attached a snippet of one of my build logs for reference. I am unsure how it gets into this state (possibly cancelling a build at a bad time) but it is inconvenient non-the-less.

```

$ mkdir -p /var/lib/buildkite-agent/builds/buildkite_agent-0d36ebbf02a599f51/company/pipeline
$ cd /var/lib/buildkite-agent/builds/buildkite_agent-0d36ebbf02a599f51/company/pipeline
$ git remote set-url origin git@bitbucket.org:company/repository.git
fatal: Not a git repository (or any of the parent directories): .git

```

lox:
Closes #533 

Currently if a previous checkout doesn't have a .git folder, it breaks future checkouts. This detects that, as well as checking that a basic `git rev-parse` works on it.

It doesn't detect corrupted git databases though. We'd need to do a `git fsck` for that, which would take a few seconds. I don't think it's worth it in stable, perhaps we could do that in future if we detect a corrupted git repository during checkout or clone.


lox:
Any feels @toolmantim?

test/integration/checkout_integration_test.go
```diff
@@ -159,3 +159,92 @@ func TestForcingACleanCheckout(t *testing.T) {
...
+	agent := tester.MustMock(t, "buildkite-agent")
+	agent.
+		Expect("meta-data", "exists", "buildkite:git:commit").
+		AndExitWith(0)
+
+	tester.RunAndCheck(t)
+}
+
+func TestCheckoutOnABrokenExistingRepository(t *testing.T) {
+	t.Skipf("This is currently broken")
```
toolmantim: This meant to be skipped?

lox: Only in that we don't currently recover from it. I write the tests expecting that we would, and then figured out that `git rev-parse` won't detect corrupted objects.

lox: I figure having a skipped test reminds us that we don't support it, but it could possibly just be removed.

toolmantim: Oh right. I thought `"This is currently broken"` meant that the test was currently broken! Perhaps that just needs to be clarified with your words above.

toolmantim:
Merge town?

toolmantim:
I notice this is for stable?

lox:
Yeah, it still feels like we are approaching it wrong. What if instead we destroyed the checkout directory after a failed checkout and retried the checkout? That would cover all of the failed checkout scenarios we've seen and not require super-git-specific magic.

toolmantim:
Sounds good to me! A single retry?

lox:
> Sounds good to me! A single retry? 

Or possibly no retry? Just removing the checkout dir on checkout hook failure would at least mean we get rid of the thing where it causes that agent to fail for future builds. I feel like if checkout fails the build should fail.

notana:
@lox I don't understand why the build should fail in this case, it is outside of the users control so shouldn't it try and recover if possible? Would save some pain of having to come back and check/retry builds when it happens.

Just a thought.

toolmantim:
I think the change as it is probably handles the most common git failures we've seen. Seems like an okay approach to me?

lox:
Yeah, I think I'm probably ok with it.

notana:
Thanks for the help with this, much appreciated!

lox:
No problems @notana, sorry it's taken us a while!

lox:
> @toolmantim I notice this is for stable?

I'll follow-up with a patch to master and we'll get this into some new releases soon.

PR #560, opened 2017-10-06
Prototype of `secret get` command

lox:
This adds a `secret get` command for fetching a secret. The currently supported backend is Amazon's SSM, via https://github.com/segmentio/chamber.

This is intended to replace ALL secret consumption, and would look something like this:

```
buildkite-agent secret get "github_ssh_key" | ssh-add -
MY_OTHER_SECRET=$(buildkite-agent secret get "my_other_secret")
```

Basically secrets will exist in a tuple of `{service, key, version}`. The "service" will be `buildkite_{org}_{pipeline}` or `buildkite_{org}`. This is hidden behind the `--scope pipeline` or `scope organization` param to `secret get`. This service will be secured with SSM/AWS via an IAM permission that only allows the bearer to access certain SSM key prefixes. 

Writing these secrets would be done via whatever backend the secrets command supported:

```
aws-vault exec my-profile -- chamber write "buildkite_myorg_mypipeline" "github_ssh_key" <(~/.ssh/id_rsa) 
aws-vault exec my-profile -- chamber write "buildkite_myorg_mypipeline" "my_other_secret" "blah"
```

We could use roughly the same schema to support other backend channels too, e.g s3 like we currently offer for https://github.com/buildkite-plugins/s3-secrets-buildkite-plugin or local encrypted files in https://github.com/Shopify/ejson.


keithpitt:
Nice! I've got a ton of ides/questions about this feature, and I'm not entirely sure if I should just throw them here - or talk in RL.

My first question that I'll ask here though is: Is this something you're thinking would be part of 3.0? Or maybe part of the 3.1 release?

keithpitt:
It also looks like some other libraries (codegansta) made it into the vendor folder. Happy for you to cherry-pick those into `master` to reduce the changeset of this PR./

lox:
I’d say 3.1. Reckon we should feature freeze 3.0.

keithpitt:
Cool, I think that's a good idea. Just wanted to double check!

PR #561, opened 2017-10-06
Test ssh known_hosts bug and refactor surrounding code

lox:
This adds some tests and cleans up the code around interacting with ssh `known_hosts` in the bootstrap. We've occasionally seen issues where `ssh-keyscan` doesn't return anything, this should at least provide some more debugging.

~~Notably, it removes the file locking around known_hosts.~~ This narrows the scope of the locking down to just the Add operation.  

Additionally, this replaces calling `ssh-keygen` to check the `known_hosts` file with manual parsing of the file, as `ssh-keygen` doesn't appear to support hosts with ports. Relates to #565.


keithpitt:
Don't we need to lock it if another agent is trying to read + write to it at the same time?

lox:
@keithpitt yeah, suspect you are right. Was hoping to eliminate the need for that, but don't think I'll be able to get it entirely atomic.

lox:
I am tempted to move it out of the knownHosts struct and into the bootstrap though. Perhaps we should expose file locking as a "Shell" thing? `

sj26:
Tests 😍

bootstrap/shell/shell.go
```diff
@@ -143,6 +143,10 @@ func (s *Shell) RunAndCapture(command string, arg ...string) (string, error) {
 		return "", err
 	}
 
+	if s.Debug {
```
lox: When debug is enabled, this prints commands being executed by RunAndCapture.

bootstrap/ssh.go
```diff
@@ -0,0 +1,54 @@
...
+func sshKeygen(sh *shell.Shell, knownHostsfile, host string) (string, error) {
+	sshKeygen, err := sh.AbsolutePath("ssh-keygen")
+	if err != nil && runtime.GOOS == "windows" {
+		var winErr error
+		if sshKeygen, winErr = findPathToSSHToolsOnWindows(sh); winErr != nil {
+			return "", err
+		}
+	}
+
+	return sh.RunAndCapture(sshKeygen, "-f", knownHostsfile, "-F", host)
```
sj26: We need to capture stdout and stderr separately — only stdout should be appended to the known hosts file.

bootstrap/ssh.go
```diff
@@ -0,0 +1,54 @@
...
+func sshKeyScan(sh *shell.Shell, host string) (string, error) {
+	sshKeyscan, err := sh.AbsolutePath("ssh-keyscan")
+	if err != nil && runtime.GOOS == "windows" {
+		var winErr error
+		if sshKeyscan, winErr = findPathToSSHToolsOnWindows(sh); winErr != nil {
+			return "", err
+		}
+	}
+
+	return sh.RunAndCapture(sshKeyscan, host)
```
sj26: We need to capture stdout and stderr separately — only stdout should be appended to the known hosts file.

Later versions of ssh-keyscan will exit non-zero, but earlier versions just won't output anything to stdout. We should probably catch these cases as errors, or at least warnings.

lox: Hmmm. I'd say we should ignore stderr entirely in RunAndCapture then.

sj26: Or always capture stdout and leave stderr connected to the actual stderr if Debug? Or something?

lox:
This is probably ready to merge.

lox:
Any last requests before merge @sj26?

bootstrap/knownhosts.go
```diff
@@ -33,147 +28,117 @@ func findKnownHosts(sh *shell.Shell) (*knownHosts, error) {
...
+	// # A CA key, accepted for any host in *.mydomain.com or *.mydomain.org
+	// @cert-authority *.mydomain.org,*.mydomain.com ssh-rsa AAAAB5W...
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		fields := strings.Split(scanner.Text(), " ")
+		if len(fields) != 3 {
+			continue
+		}
+		for _, addr := range strings.Split(fields[0], ",") {
+			if addr == normalized {
```
sj26: Do we are that this no longer supports hashed hosts?

sj26:
This is awesmoe, thanks @lox!

bootstrap/knownhosts.go
lox: Yeah, good point. I'll add that.

PR #565, opened 2017-10-18
Fix ssh keyscan for urls with ports

### LINKED ISSUE #564: SSH Keyscanning breaks for git repositories with non-standard ports
See https://github.com/buildkite/elastic-ci-stack-for-aws/issues/348:

```
$ ssh-keyscan scm.xxx
$ git clone -v -- ssh://git@scm.xxx:7999/yyy/zzz.git .
Cloning into '.'...
The authenticity of host '[scm.xxx]:7999 ([11.22.33.44]:7999)' can't be established.
RSA key fingerprint is 28:e7:1e:d8:c0:a7:9c:c9:f1:c3:6d:fd:0b:1c:c2:00.
Are you sure you want to continue connecting (yes/no)?
```

sj26:
[Hostname()](https://golang.org/pkg/net/url/#URL.Hostname) returns Host without any port number.

This shouldn't clash with #561 looking at the diff, but should be complementary.

Closes #564.

### DIFF
--- a/bootstrap/knownhosts.go
+++ b/bootstrap/knownhosts.go
@@ -124,6 +124,14 @@ func newGittableURL(ref string) (*url.URL, error) {
 	return url.Parse(ref)
 }
 
+// Clean up the SSH host and remove any key identifiers. See:
+// git@github.com-custom-identifier:foo/bar.git
+// https://buildkite.com/docs/agent/ssh-keys#creating-multiple-ssh-keys
+var repoSSHKeySwitcherRegex = regexp.MustCompile(`-[a-z0-9\-]+$`)
+func repoSSHKeySwitcherStrip(host string) (string) {
+	return repoSSHKeySwitcherRegex.ReplaceAllString(host, "")
+}
+
 // AddFromRepository takes a git repo url, extracts the host and adds it
 func (kh *knownHosts) AddFromRepository(repository string) error {
 	// Try and parse the repository URL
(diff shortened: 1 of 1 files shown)


bootstrap/knownhosts.go
```diff
@@ -124,6 +124,14 @@ func newGittableURL(ref string) (*url.URL, error) {
 	return url.Parse(ref)
 }
 
+// Clean up the SSH host and remove any key identifiers. See:
+// git@github.com-custom-identifier:foo/bar.git
+// https://buildkite.com/docs/agent/ssh-keys#creating-multiple-ssh-keys
+var repoSSHKeySwitcherRegex = regexp.MustCompile(`-[a-z0-9\-]+$`)
+func repoSSHKeySwitcherStrip(host string) (string) {
```
lox: Wouldn't this strip these values from legit hostnames?

lox: Perhaps `stripAliasesFromGitHost` might be clearer?

sj26: I don't mind what it's called, just extracted it out the way :-)

sj26:
Of course this doesn't actually work. ssh-keyscan doesn't know which port to scan the remote host on. We need to supply a `-p $PORT` to ssh-keyscan.

But it's more complicated again becasue `ssh-keygen -F` which tests the presence of the host in the existing known hosts file doesn't understand port numbers.

We're also dumping the comments and errors which `ssh-keyscan` outputs to stderr into the known_hosts file which we shouldn't be doing. I had invalid lines in my known hosts after scanning a host which couldn't be connected to.

I'm thinking we might need to own this stuff entirely and use something like [x/net/ssh/knownhosts](https://godoc.org/golang.org/x/crypto/ssh/knownhosts) or [fleet's known_hosts](https://github.com/coreos/fleet/blob/master/ssh/known_hosts.go) to test for the presence of the host, then parsing the output of ssh-keyscan and only adding good entries to the file. Or something.

lox:
Hmmm. Sounds deeply unpleasant.

lox:
My vote would be getting `x/net/ssh/knownhosts` into https://github.com/buildkite/agent/pull/561

sj26:
Yeah, that's where I was leaning.

PR #566, opened 2017-10-18
Only show "Running hook" if one exists

lox:
Currently we show a LOT of headers in the 3.x agent. This removes "Running hook" headers if the header doesn't exist.


toolmantim:
😻

PR #569, opened 2017-10-20
Trim environment variables to avoid unexpected line endings on windows

m-pavel:
When running on windows machine S3* variables initialized from environment.bat are incorrect. They have tailing '\r' symbol. Not depending on what line endings used in bat file. Errors are like following:
 FATAL  Failed to upload artifacts: Unknown AWS S3 Region "us-east-1\r"
Changes below just trim strings.

### DIFF
--- a/agent/s3.go
+++ b/agent/s3.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials"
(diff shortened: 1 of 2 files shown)


lox:
Thanks for the pull request @m-pavel! Which version of the agent are you seeing this issue with?

m-pavel:
Hi,
That's for the v3.0-beta.34.

lox:
The current beta test suite isn't passing on windows, so I'm going to get that working and figure out the cause of this bug! There shouldn't be line endings on ENV, so I suspect it's a deeper problem.

lox:
We figured out the bug that is causing this, including it in #538. Thanks for raising the issue and PR @m-pavel!

PR #570, opened 2017-10-27
Fix tests on windows

lox:
This gets our test suites passing under Windows!


lox:
Consolidating with #538

PR #573, opened 2017-11-07
Add a check for ensuring the changelog is updated

toolmantim:


toolmantim:
I wanted to ask if we should fail the build… because the error annotation isn't visible from the GitHub pull request commit statuses? Or if we should fail the build only on a tag release? But in that case, we'd need to force push the tag to fix? Hrmmm…

lox:
Yeah, I reckon fail the build.

lox:
Actually, how will this work, given that we bump the version to the next version?

toolmantim:
That’s kinda why I didnt. Maybe we need the click to unblock step to actually ask “Have you updated the Changelog? Yes no”?

PR #577, opened 2017-11-08
Pass working dir between hooks

lox:
It's common in v2 agent hooks to change directories and expect it to affect subsequent hooks, for instance you might want to change the working directory in a `pre-command` hook for a monorepo. 

This PR makes that happen.

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -139,24 +139,24 @@ func (b *Bootstrap) executeHook(name string, hookPath string, extraEnviron *env.
 	}
 
 	// Get changed environent
-	changes, err := script.ChangedEnvironment()
+	changes, err := script.Changes()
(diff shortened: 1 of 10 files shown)


lox:
Nope, it was never implemented in v3.

lox:
I don't think, anyway.

lox:
I think I figured out how we thought this worked @keithpitt. The example you mentioned of agent-tests uses a `post-checkout` hook and sets `BUILDKITE_BUILD_CHECKOUT_PATH`:

https://github.com/buildkite/agent-tests/blob/60dff36cffd83ed3b8b389b475359b7b4484f53e/.buildkite/hooks/post-checkout.bat#L6

There is special handling in the v3 bootstrap for this changing in post-checkout hooks. The v2 bootstrap will respect any directory change in a hook.

PR #579, opened 2017-11-08
Support hashed hostnames in .ssh/known_hosts

lox:
I forgot to support this in the initial implementation, this adds support for hashed hostnames.


plasticine:
Awesome—I ran into this just yesterday! <3

PR #580, opened 2017-11-08
Handle upgrading bootstrap-path from old 2.x shell script

lox:
This adds a bootstrap.sh shim that gets installed at the same location as the old v2.x bootstrap that prints a warning and then runs `buildkite-agent bootstrap`. This should mean that users that upgrade from v2 -> v3 don't run into problems because their config still refers to the v2 bootstrap.sh.

### DIFF
--- a/packaging/linux/root/usr/share/buildkite-agent/bootstrap.sh
+++ b/packaging/linux/root/usr/share/buildkite-agent/bootstrap.sh
@@ -1,0 +1,8 @@
+#!/bin/sh
+
+# This is a shim to make the upgrade process from v2 -> v3 smoother as
+# v2 configurations will still have bootstrap-path=/usr/share/buildkite-agent/bootstrap.sh
+# which will cause all sorts of wierd behaviour
+
+echo "+++ :warning: Your agent configuration contains an outdated bootstrap-path reference. It can be safely removed."
+exec buildkite-agent bootstrap "$@"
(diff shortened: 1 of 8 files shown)


packaging/linux/root/usr/share/buildkite-agent/bootstrap.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+# This is a shim to make the upgrade process from v2 -> v3 smoother as
+# v2 configurations will still have bootstrap-path=/usr/share/buildkite-agent/bootstrap.sh
+# which will cause all sorts of wierd behaviour
+
+echo "+++ :warning: Your agent configuration contains an outdated bootstrap-path reference. It can be safely removed."
```
toolmantim: It'd be great to make this a bit more useful. Do we have the path to the configuration file?

Maybe something like:

```bash
echo "+++ warning: Agent configuration file deprecation warning"
echo "Your agent configuration file references a bootstrap-script compatibility shim, which will be removed in Buildkite Agent 4.x+"
echo ""
echo "To fix this, remove the 'bootstrap-script' line from your agent configuration file located at:"
echo "/var/lib/buildkite-agent/buildkite-agent.cfg"
echo ""
echo "For more information, see our Buildkite Agent 3.0 upgrade guide:"
echo "https://buildkite.com/docs/agent/some-upgrade-doc"
```

lox: 👌🏻

keithpitt: https://buildkite.com/docs/agent/upgrading-to-v3

PR #583, opened 2017-11-12
Add vagrant for windows

lox:
This adds a Vagrantfile and docs for creating a Window's based docker environment.


README.md
```diff
@@ -66,36 +68,29 @@ To test the commands locally:
 go run main.go start --debug --token "abc123"
 ```
 
-### Dependency management
```
toolmantim: As a very infrequent golang user, I'm curious why this is gone?

lox: Yeah, fair point. I guess I removed it because adding a dependency is something that is done very infrequently. Previously we didn't commit `vendors/`, so it was something that you ran into more. We don't include steps on how to add a dependency on our other repos, so it seemed like overkill here.

lox: That said, it probably was poor form to tailgate that change into this PR.

toolmantim: Cool. I know in Ruby land there's pretty much only one way, so no need to say "We use this style" — I think once upon a time there were multiple approaches to vendor'ing in golang? It's pretty standard now?

lox: Yeah, I guess that is true.

lox: Shall I put it back to make this PR less controversial? In future I'd like to move those sorts of things into a DEVELOPMENT doc.

toolmantim: No it's all good, I was just curious how dependencies are done in golang, and if a third-party contributor would know what to do?

lox: No, you are probably right. Things are better, but not at the same state of convergence.

PR #587, opened 2017-11-15
Show the version to be deployed in block steps

lox:
This splits the pipeline.yml into two phases so we can interpolate AGENT_VERSION into the block steps. 

My hope is that this will prevent myself from accidentally clicking that button before bumping the version again.


.buildkite/pipeline.release-experimental.yml
```diff
@@ -0,0 +1,36 @@
+steps:
+  - wait
+
+  - name: ":s3: publish binaries"
+    command: "scripts/publish-to-s3.sh"
+    branches: "master 2-1-stable"
```
toolmantim: The 2-1-stable stuff makes no sense in master, you can remove all those. Each stable release branch has it's own version: https://github.com/buildkite/agent/blob/2-6-stable/.buildkite/pipeline.yml#L100

lox:
I'm going to move this into a better named branch and do some rebasing.

PR #588, opened 2017-11-17
Split into different pipelines for release

lox:
I accidentally half-released a beta version that overwrote v3.0-beta.34, so as part of preventing it happening again I thought it would be a good time to split our agent pipeline into different build and release pipelines. This does that. It also:

* Only shows a beta step if there is no existing tag
* Only shows a stable step if the version doesn't match `(beta|rc)`. 
* Fails if there isn't a changelog entry
* Adds a `DRY_RUN` variable that prevents releases from actually being released, for testing.

The block steps exist in this pipeline, which then trigger builds in `Agent - Release - Experimental`, `Agent - Release - Unstable` and`Agent - Release - Stable`.  

screentshot

TODO:

* [x] Remove `echo`from commands that do things
* [x] Add dry-run support to homebrew/github-release steps
* [x] Remove references to `show-version-in-block-steps` branch


scripts/github-release.sh
```diff
@@ -1,6 +1,14 @@
 #!/bin/bash
 set -e
 
+dry_run() {
+  if [[ "${DRY_RUN:-}" == "false" ]] ; then
+    echo "$@"
```
sj26: I think this line maybe shouldn't have an `echo`?

scripts/upload-release-steps.sh
```diff
@@ -0,0 +1,72 @@
...
+}
+
+output_steps_yaml() {
+  echo "steps:"
+
+  trigger_step \
+    "edge ${agent_version}.${build_version}" \
+    "agent-release-experimental"
+
+  block_step "beta?"
```
sj26: Is there any way of adding a pop-up confirmation to the block step with some more description, like "This will release version v3.0.0-beta.23 and publish it to all customers" or something? Or can we only do pop-ups with fields? Might just be nice to be very explicit with verification. Especially for stable.

scripts/upload-release-steps.sh
```diff
@@ -0,0 +1,72 @@
...
+
+  trigger_step \
+    "beta ${agent_version}" \
+    "agent-release-unstable"
+
+  if [[ ! $agent_version =~ (beta|rc) ]] ; then
```
sj26: Maybe `"$agent_version"` in case special chars creep in?

lox:
I'm thinking I will ship this unless you object @sj26.

toolmantim:
Awesome! 👏🏼

PR #589, opened 2017-11-20
Don't retry pipeline uploads on 422 responses

DazWorrall:
Some example 422 responses on pipeline upload:

```
WARN   POST https://agent.buildkite.com/v3/jobs/<uuid>/pipelines: 422 One of the steps you provided was invalid: can't be blank, The step type supplied is not supported (Attempt 1/5 Retrying in 1s)
WARN   POST https://agent.buildkite.com/v3/jobs/<uuid>/pipelines: 422 One of the steps you provided was invalid: The reason provided to `skip` is too long. The maximum length is 70 characters (Attempt 1/5 Retrying in 1s)
```

These will never succeed on retry, so save some time by bailing early.


lox:
Thanks @DazWorrall!

PR #590, opened 2017-11-21
Add a `--spawn` option to spawn multiple parallel agents

lox:
Inspired by the long dormant https://github.com/buildkite/agent/tree/lol-workers and the need for high-concurrency agents for testing lockfiles, I added a hidden ~~`--workers`~~`--spawn` flag that specifies how many concurrent workers to run. 

Have been giving it a workout with `-race` on and have sorted out the various leaks.

### DIFF
--- a/agent/agent_pool.go
+++ b/agent/agent_pool.go
@@ -34,6 +34,7 @@ type AgentPool struct {
 	DisableHTTP2          bool
 	AgentConfiguration    *AgentConfiguration
 	MetricsCollector      *metrics.Collector
+	Workers               int
 
 	interruptCount int
 	signalLock     sync.Mutex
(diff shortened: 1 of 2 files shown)


toolmantim:
That branch name 😂

lox:
I figure I'll land this in 3.1 unless anyone objects. We can perhaps consider proper public support in 4.0?

keithpitt:
Ah this old thing! Maybe a different term here would be better? What about `--pool`? Was there an aversion to using `--workers?` We already use `concurrency` for step parallelism, and might be a bit confusing using a shared term?

I also wonder if it's better to have multiple agents within the 1 process, or have another program managing multiple processes? Feels?

lox:
Yeah, I like `pool-size` or `workers`, agree `concurrency` is confusingly overloaded at this stage.

lox:
The worker pool refactor got pulled out into #599, so this will get simpler once that lands.

lox:
Ok, this is now hidden behind a `MultipleWorkers` experiment flag too. Feels on shipping it as part of 3.0 in that state @keithpitt?

lox:
> I also wonder if it's better to have multiple agents within the 1 process, or have another program managing multiple processes? Feels?

I think the model in this PR is great, it's simple and has a minimal footprint in terms of system configuration. It should just nicely replace all the stuff people do with multiple init scripts.

lox:
Thought it over, reckon 3.1 is safer.

keithpitt:
Going to dismiss my review here as well for now!

toolmantim:
I don't have much else to add to this! But I'm keen to hear from @keithpitt and @sj26 for how it works conceptually w/ Buildkite

lox:
I believe they will use there own http connections, we could definitely look to consolidate them, I wonder if that would have conurrency issues between the go routines.

lox:
I was aiming for "no worse than running the same concurrency of top-level processes"

sj26:
@lox yep, I reckon don't worry about it for now.

The signals and figuring out stopping are the only blockers on my mind. Tweaking the log verbosity is a nice-to-have.

keithpitt:
@lox the other "nice to have" with this change is if we could send some sort of identifier to the registration API that this agent was spun up along side others.. maybe it'd be called `SpawnUUID` or something? We'd only send it if the agent was run with `spawn > 1`.

Then in the BK backend (when stopping an agent) we'd be able to also stop the other agents it was registered along side.

sj26:
I'm happy with this direction, ask for review again when we're ready to let it loose?

keithpitt:
Just gonna remove myself from the review requests until we do another around of looking at this :)

lox:
I investigated adding a prefix to the logger, it's super hard as the logger gear uses globals and package-level methods. I will add another PR with a refactor of the logging gear after this lands.

sj26:
I know it's probably too late for naming, but what about `--count`?

```sh
buildkite-agent start --token abc123 --count 5
```

PR #593, opened 2017-11-22
Improve plugin git transport support

### LINKED ISSUE #520: Support for privately-hosted Buildkite plugins
I am currently trying to create a Buildkite plugin that utilises an internally-built tool that helps developers interact with & deploy to our AWS environments. As this is an internal tool, the source code is private, hosted on our Github Enterprise instance.

This plugin would be optionally called by developers in their build pipelines - for example it might look something like this:

```
label: 'deploy'
command: 'deploy'
plugins:
  my-private-plugin:
    environment: 'production'
```

Currently there are a few issues impeding my ability to do this:

1. If you do not provide a fully qualified path as the plugin name, it will automatically download the plugin from `https://github.com/buildkite-plugins/<name>`.

2. If you _do_ provide a fully qualified path as the plugin name (e.g., `private.repo.com/my-org/my-private-plugin`), it will try to git clone as HTTPS. However, as this is not a publicly available repository, this fails unless provided git credentials. Even if you DO provide git credentials (such as with the S3 secrets buildkite plugin), this fails due to the fact the plugin clone step explicitly ignores provided credentials: https://github.com/buildkite/agent/blob/master/bootstrap/bootstrap.go#L946

3. If you take the approach of baking the plugins into the agent AMI, the plugins must still be referred to by their fully qualified name (i.e., `private.repo.com/my-org/my-private-plugin` rather than simply `my-private-plugin`) as using the shortened name will assume it comes from github.com/buildkite-plugins. 

Additionally the naming is very particular, as plugins must be named according to an internal standard in the buildkite-agent `~/plugins` directory as something like `my-private-repo-com-my-org-my-private-plugin` to simply avoid doing a `git clone` :) I was able to get this working in the end though it feels brittle and slows/limits the ability to test new versions of the plugin...

Ideally I would like to be able to refer to private plugins with the shortened name (i.e., `my-private-plugin` rather than `private.repo.com/my-org/my-private-plugin`) and additionally specify an alternate location for the plugin as an option. It would also need to honour the git credentials set or alternatively support ssh download.

Something like this, roughly:

```
plugins:
  my-private-plugin:
    location: private.repo.com/my-org/my-private-plugin-repo
    use-git-credentials: true
```

In the minimal case, simply allowing me to https clone using stored git credentials would be more than workable :)

toolmantim:
Using plugins from non-http/https git repos seems broken at the moment. This adds support for a bunch of things listed by `git clone --help`

Fixes #520

- [x] Tests
- [ ] Fixes

### DIFF
--- a/agent/plugin_test.go
+++ b/agent/plugin_test.go
@@ -191,13 +191,41 @@ func TestRepositoryAndSubdirectory(t *testing.T) {
 	assert.Equal(t, sub, "docker-compose/beta")
 	assert.Nil(t, err)
 
-	plugin = &Plugin{Location: "/Users/keithpitt/Development/plugins.git/test-plugin"}
+	// Check that all the various git protocols are supported
+
+	plugin = &Plugin{Location: "ssh://user@myrepo.com:99/test-plugin"}
 	repo, err = plugin.Repository()
-	assert.Equal(t, repo, "/Users/keithpitt/Development/plugins.git")
+	assert.Equal(t, "ssh://user@myrepo.com:99", repo)
 	assert.Nil(t, err)
 	sub, err = plugin.RepositorySubdirectory()
-	assert.Equal(t, sub, "test-plugin")
+	assert.Equal(t, "test-plugin", sub)
 	assert.Nil(t, err)
+
+	plugin = &Plugin{Location: "https://user@myrepo.com:99/test-plugin"}
+	repo, err = plugin.Repository()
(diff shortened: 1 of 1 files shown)


toolmantim:
As per https://github.com/buildkite/agent/issues/520#issuecomment-346248920 maybe this isn't broken and I'm just going crazy?

toolmantim:
I don't think this is actually an issue, as they seem to work AOK. Closing this for now.

Issue #291, opened 2016-04-19
Agent cannot be run under most recent amazon linux / Trace/breakpoint trap


mbj:
The latest amazon linux release disables the `modify_ldt` systemcall golang used.

> The kernel config CONFIG_MODIFY_LDT_SYSCALL, which control
> enable/disable modify_ldt, is disabled on Amazon Linux 2016.03.

Golang fixes this in: https://github.com/golang/go/commit/0f86d1edfb87fddb2c7f50d177b67f48219151f3

I'd hope you could do a new agent release build with a fixed golang.

keithpitt:
Oh! Thanks for letting us know. I just checked the latest Golang release and it looks like they haven't done an official release yet with this fix?

mbj:
@keithpitt AFAIK there is no such release yet. I wounder if golang master is stable enough to use it? Their review process seems to be strict enough. Right now I downgraded to the 2015.09 AMIs but these need to get security fixes applied that need a reboot to get fully active (else the fixed glibc is not globally used), so not a very nice workaround.

mbj:
@lox Was this fixed?

lox:
Actually, now I double check and it's golang 1.9.2. We build with `golang:1.9`, which might be older.

lox:
I'll update the image and the next release should work.

lox:
This is in Beta 36!

PR #596, opened 2017-11-23
Bump golang version to 1.9.2 and version to 3.0-beta.35

lox:


dch:
I have a nascent FreeBSD port in the works. Is there something specific that `agent` requires from 1.9.2
or can I continue to build against 1.9.0?

lox:
Just updated the description, this was motivated by https://github.com/buildkite/agent/issues/291.

lox:
Should be fine to keep building against 1.9.0 for BSD @dch!

PR #599, opened 2017-11-25
Refactor job runner

lox:
As part of #590 I discovered that the callback `StartCallback` used in `process.Process` triggered the golang race detector. This replaces that callback with a `WaitStarted` method, which allowed for some cleanup in other places.

Beyond that, I noticed that we seem to implement a lot of what go's `https://golang.org/pkg/bufio/#Scanner`, so I figured we could just use that, it cuts down on quite a bit of code.

### DIFF
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -63,16 +63,15 @@ func (r JobRunner) Create() (runner *JobRunner, err error) {
 	runner.logStreamer = LogStreamer{MaxChunkSizeBytes: r.Job.ChunksMaxSizeBytes, Callback: r.onUploadChunk}.New()
 
 	// The process that will run the bootstrap script
-	runner.process = &process.Process{
-		Script:             r.AgentConfiguration.BootstrapScript,
-		Env:                r.createEnvironment(),
-		PTY:                r.AgentConfiguration.RunInPty,
-		Timestamp:          r.AgentConfiguration.TimestampLines,
-		StartCallback:      r.onProcessStartCallback,
-		LineCallback:       runner.headerTimesStreamer.Scan,
-		LinePreProcessor:   runner.headerTimesStreamer.LinePreProcessor,
-		LineCallbackFilter: runner.headerTimesStreamer.LineIsHeader,
-	}
+	process := process.NewProcess()
+	process.Script = r.AgentConfiguration.BootstrapScript
+	process.Env = r.createEnvironment()
+	process.PTY = r.AgentConfiguration.RunInPty
+	process.Timestamp = r.AgentConfiguration.TimestampLines
+	process.LineCallback = runner.headerTimesStreamer.Scan
+	process.LinePreProcessor = runner.headerTimesStreamer.LinePreProcessor
+	process.LineCallbackFilter = runner.headerTimesStreamer.LineIsHeader
+	runner.process = process
 
 	return
 }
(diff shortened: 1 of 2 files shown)


lox:
@stefanmb looks like you spent some time with this code previously, would welcome your feels on it if any!

process/process.go
```diff
@@ -132,82 +139,44 @@ func (p *Process) Start() error {
...
-		reader := bufio.NewReader(lineReaderPipe)
+	// Notify other goroutines that are blocked on our Started condition.
+	p.startedCond.L.Unlock()
+	p.startedCond.Broadcast()
 
-		var appending []byte
-		var lineCallbackWaitGroup sync.WaitGroup
+	scanner := bufio.NewScanner(lineReaderPipe)
 
-		for {
```
stefanmb: 👍  Great to see this part abstracted/removed.

agent/job_runner.go
```diff
@@ -249,53 +264,37 @@ func (r *JobRunner) finishJob(finishedAt time.Time, exitStatus string, failedChu
...
+	logger.Debug("[JobRunner] Routine that processes the log has finished")
+}
 
-			// Check for cancellations
-			time.Sleep(time.Duration(r.Agent.JobStatusInterval) * time.Second)
+// Start a routine that will constantly ping Buildkite to see if the
+// job has been canceled or changed in state
+func (r *JobRunner) pollJobStateChanges() {
+	for r.process.IsRunning() {
+		// Re-get the job and check it's status to see if it's been cancelled
```
stefanmb: Nitpick: Old typo "its status" is correct.

agent/job_runner.go
```diff
@@ -249,53 +264,37 @@ func (r *JobRunner) finishJob(finishedAt time.Time, exitStatus string, failedChu
...
+		jobState, _, err := r.APIClient.Jobs.GetState(r.Job.ID)
+		if err != nil {
+			// We don't really care if it fails, we'll just try again soon anyway
+			logger.Warn("Problem with getting job state %s (%s)", r.Job.ID, err)
+		} else if jobState.State == "canceling" || jobState.State == "canceled" {
+			_ = r.Kill()
```
stefanmb: Seems safe to call repeatedly 👍

lox:
@keithpitt do you remember if there was a reason for using `line, isPrefix, err := reader.ReadLine()` vs the higher-level `Scanner`?

agent/job_runner.go
```diff
@@ -63,16 +63,15 @@ func (r JobRunner) Create() (runner *JobRunner, err error) {
...
-		Script:             r.AgentConfiguration.BootstrapScript,
-		Env:                r.createEnvironment(),
-		PTY:                r.AgentConfiguration.RunInPty,
-		Timestamp:          r.AgentConfiguration.TimestampLines,
-		StartCallback:      r.onProcessStartCallback,
-		LineCallback:       runner.headerTimesStreamer.Scan,
-		LinePreProcessor:   runner.headerTimesStreamer.LinePreProcessor,
-		LineCallbackFilter: runner.headerTimesStreamer.LineIsHeader,
-	}
+	process := process.NewProcess()
```
lox: Is this needed?

lox:
Closing this in favour of some smaller pieces.

PR #604, opened 2017-12-13
S3 streaming and zero size fixes

sj26:
Removes our wholesale buffering of s3 uploaded artifacts into memory by using PutReader (which is what goamz is doing under the hood anyway), and patches an issue introduced by newer Golang's net/http which was fixed upstream in https://github.com/goamz/goamz/pull/135. 

From [net/http docs](https://golang.org/pkg/net/http/):

> ContentLength records the length of the associated content.
>
> For client requests, a value of 0 with a non-nil Body is  also treated as unknown.

This is not an issue in master (v3 beta) because we've replaced goamz with the official aws-sdk.

### DIFF
--- a/agent/s3_uploader.go
+++ b/agent/s3_uploader.go
@@ -3,7 +3,6 @@ package agent
 import (
 	"errors"
 	"fmt"
-	"io/ioutil"
 	"net/url"
 	"os"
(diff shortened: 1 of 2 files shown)


sj26:
Thanks for merging, @lox!

PR #605, opened 2017-12-14
Add agent-wide plugin support

lox:
One outstanding issue with our plugins is that there is a chicken-and-egg issue with plugins that are needed before you have your source code, for instance getting git credentials for checkout. 

This adds a new top-level configuration and env for adding agent plugins that are then loaded for every job that the agent processes:

```
--plugins value                              
A comma-separated list of plugins to preload for all jobs on this agent (e.g s3-secrets#1.1.0) [$BUILDKITE_PLUGINS]
```

This is then merged with whatever plugins are brought in via the pipeline. The agent-wide ones trump the pipeline ones. 

I'm not entirely comfortable with the name for these two. An agent plugin vs a pipeline plugin? Feels @keithpitt @toolmantim @sj26?

### DIFF
--- a/agent/agent_configuration.go
+++ b/agent/agent_configuration.go
@@ -5,6 +5,7 @@ type AgentConfiguration struct {
 	BuildPath                  string
 	HooksPath                  string
 	PluginsPath                string
+	Plugins                    []string
 	GitCloneFlags              string
 	GitCleanFlags              string
 	SSHFingerprintVerification bool
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -1,6 +1,7 @@
 package agent
 
 import (
+	"encoding/json"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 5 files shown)


toolmantim:
Interesting!

Any ideas how you’d configure the plugins, or the config file syntax?

lox:
You'd have to use something like `BUILDKITE_PLUGIN_SECRETS_BUCKET` style env I think.

lox:
The alternative would be something like `buildkite-agent plugins load s3-secrets#1.1.0` in your `environment` hook. The security implications of that get a bit sketchy.

toolmantim:
🤔 What about a `plugins.yml`?

lox:
Along side `buildkite-agent.cfg`? Would probably be a bit weird wouldn't it?

lox:
But definitely interesting.

lox:
The other option would be allowing plugins to be added at a pipeline level.

lox:
That would probably be a more natural place to have a `plugins.yml`, it makes the agents less special-snowflake.

toolmantim:
Yeah, it’s a bit weird. But I’m just thinking for end users of plugins… to go from a readme, to installing it. We pretty much just focus on the YML version.

🤔 In the env version, do you reckon you’d use an env hook to configure them?

I guess the env config methods aren’t a secret, they’re just not at all documented unless you’re a plugin author.

We should probably document the exact env name on every plugin readme next to the option name too. And show how to do arrays etc?

Or we just lock in the v3 yml plugin spec and implement it in the agent too?

lox:
Could we just support a top-level plugins block in the new YAML-editor based pipelines?

lox:
And maybe have some way to apply plugins on pipeline upload?

lox:
> I guess the env config methods aren’t a secret, they’re just not at all documented unless you’re a plugin author.

Yeah, I agree, it's awkward.

lox:
What if we offered an `buildkite-agent plugin <name> get/set` command for interacting with the config values?

lox:
That might get gnarly I guess.

lox:
Reminds me of this: https://gist.github.com/lox/1dd425819b4e3874ba394346cb56a74c

lox:
```yaml

    
steps:
  - name: ":pipeline:"
    command: buildkite-agent pipeline upload --inherit-plugins
    plugins:
      s3-secrets#v1.0:
        bucket: my-buildkite-secrets
```

lox:
That is still probably weird semantics in that the plugin would be loaded and executed for the pipeline step.

toolmantim:
I like the top level plugins block, and also the idea of a plugins.yml at the agent level

lox:
Closing this for now, thing we will go a different way.

PR #609, opened 2017-12-25
fix "multi" typo

deoxxa:


lox:
Christmas doc's fixes! 🎄 Thanks @deoxxa. How is your buildkite emulator going?

PR #610, opened 2018-01-06
Port features from bootstrap.sh to bootstrap.bat

solemnwarning:
Respect the following environment variables:

 - BUILDKITE_CLEAN_CHECKOUT

 - BUILDKITE_GIT_CLONE_FLAGS

 - BUILDKITE_GIT_CLEAN_FLAGS


lox:
Thanks! We haven't ported a lot of the features from bash bootstrap to the batch one because we made the decision to focus on the 3.x bootstrap rewrite in golang. Appreciate these, will get them in a release early this week.

We are hoping to get the 3.x agent released stable in the next 4-6 weeks too.

PR #612, opened 2018-01-09
Update buildkite-agent.cfg to show new defaults for git-clean-flags

lox:
As part of v3, we updated the defaults for `git-clean-flags` to add `-x`. It doesn't look like this is reflected in the packaging default configs. This updates them.


lox:
Urgh so it did.

PR #613, opened 2018-01-09
Add a config option to retry checkouts

lox:
There are several situations in which checkouts can be flakey, sometimes it's a overloaded central server and sometimes it's something like XCode being slow to release file locks (causing git clean to fail). 

This adds a generalized exponential backoff on checkouts. It is off by default.

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -14,7 +14,13 @@ import (
 	"github.com/buildkite/agent/agent"
 	"github.com/buildkite/agent/bootstrap/shell"
 	"github.com/buildkite/agent/env"
+	"github.com/buildkite/agent/experiments"
+	"github.com/buildkite/agent/retry"
 	"github.com/pkg/errors"
+)
+
+const (
+	retryCheckoutOnFailure = `RetryCheckoutOnFailure`
 )
 
 // Bootstrap represents the phases of execution in a Buildkite Job. It's run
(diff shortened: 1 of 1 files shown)


jgavris:
We build an Xcode project (putting its DERIVED_DATA_PATH) in the git repository. Sometimes builds are interrupted because we're pushing new code to that branch, and Xcode is a little slow at unlocking all the build files. When the next job starts, git fails to complete the checkout because of the locked files. We've been manually retrying this for a while, but it is fairly annoying. Because this step runs before we get a chance to intervene, this fix needs to placed here. Thanks!

keithpitt:
@lox oh cool, any reason why this isn't just built in by default?

jgavris:
@keithpitt what’s a reasonable default?

keithpitt:
@jgavris as in, instead of having to "opt-in" to retries, it's just part of the default checkout process. I can't think of an example where you _wouldn't_ want us to retry if it fails?

keithpitt:
Perhaps if the repository is _really_ large, and doing another checkout is really slow - then it could be a bad idea? But the job will fail anyway, and someone will login to BK, see the result and be like "eughhh stupid internet" and just hit retry. At least if we retry by default that'll save them a step?

keithpitt:
Oh, also what's the behaviour when a checkout fails because of a security check?

lox:
The most common failure would be not having permission to checkout a repository, in which case I reckon a retry would be pretty annoying.

lox:
We've seen checkouts fail from Phabricator due to the server being overwhelmed. 

I dunno, I'd tend to have this off by default initially. I think it's a solution to an uncommon problem that has the side effect of making permissions errors much slower to fail.

jgavris:
I'm totally fine with turning this on in the config...our linux build agents for instance probably won't need this behavior.

keithpitt:
> The most common failure would be not having permission to checkout a repository, in which case I reckon a retry would be pretty annoying.

Totally. In that case, could we perhaps inspect the output of the checkout and not do a retry?

keithpitt:
I'm just trying to avoid configurations where users contact us and email support with "I have this problem" and we're like "oh, go turn this thing on". When ever I have problem with software, my first point of call isn't usually looking in the config for a magical flag to change.

(I'm just trying to make my case that this should just be on by default no config. I think the change is a really good one)

lox:
> I'm just trying to avoid configurations where users contact us and email support with "I have this problem" and we're like "oh, go turn this thing on". When ever I have problem with software, my first point of call isn't usually looking in the config for a magical flag to change.

Yup, totally on the same page with avoiding proliferation of knobs and dials. I just suspect there are enough completely common reasons for a checkout to fail that we will have a hard time catching them all. I guess it would be pretty annoying to get retries for some of the common ones if you turned this on regardless of whether it's default or not.

jgavris:
If you could catch more specifically the `git clean` portion, that would probably be safe to retry a few times?

jgavris:
Can we try this out as opt-in for a release / version?

lox:
We'll ship this in the next beta as an "experiment". Will do a write up on how to enable it when we ship it @jgavris. Thanks for your patience, all going well this will either become a default or a config option in the v3 stable release.

jgavris:
Is it possible to install the beta from homebrew?

PR #617, opened 2018-01-14
Replace process.Process's bufio.Reader with a bufio.Scanner

lox:
The bufio.Reader is a low-level primitive, and it should be a bufio.Scanner. 

Originally part of https://github.com/buildkite/agent/pull/599.

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -30,17 +30,22 @@ type Process struct {
 	Env        []string
 	ExitStatus string
 
-	buffer  outputBuffer
-	command *exec.Cmd
-
-	// This callback is called when the process offically starts
+	// StartCallback is called when the process is started
 	StartCallback func()
 
-	// For every line in the process output, this callback will be called
-	// with the contents of the line if its filter returns true.
-	LineCallback       func(string)
-	LinePreProcessor   func(string) string
+	// LinePreProcessor is called with ever line of output and can be used to modify each line
+	// The modified value is then made available to LineCallbackFilter and LineCallback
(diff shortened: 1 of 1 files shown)


lox:
This build shows the race condition I was talking about @keithpitt: https://buildkite.com/buildkite/agent/builds/1989#1809e9a0-6279-4180-9f28-39a4c623f5ee

lox:
There is some history in:

https://github.com/buildkite/agent/pull/430 and https://github.com/buildkite/agent/commit/e1d590a81b3e84525e0660f13cd146715b84c399

keithpitt:
I'm 10% nervous about this change, mostly because it's super core and we're merging it right before we ship the last beta as stable. I suppose this is one of those things though that "works" or it doesn't.

Here's a question I've just thought of. How does it handle lines that are currently in progress? I.e. `rspec` appends to the current line, and when it's done, sends a "\n". Will the scanner get access to the last unfinished line?

lox:
Yeah, I'm slightly nervous about it too. Let's defer it till 3.1! I got the tests and race fixes in the other PR, so this one is much lower down my list of cares.

keithpitt:
@lox I'll dismiss my review for now and do a proper look when we have our 3.1 party.

lox:
This should have no impact at all @keithpitt, it does exactly the same thing as it did before, just using the go standard library more and less code in agent. 

If you're worried about it, we could drop it in an experiment and run that for a while?

keithpitt:
Sounds fine to me!

PR #618, opened 2018-01-14
Fix race condition in process.Process

lox:
This adds test coverage for `process.Process` and fixes a race that is caused by the jobRunner calling `process.Output()` in go routine: https://github.com/buildkite/agent/blob/e3fa0d1ed2f46fd001382586dfcd478bed97ce99/agent/job_runner.go#L259-L273
The fix is a `bytes.Buffer` that is wrapped with a RWMutex. 

Originally part of https://github.com/buildkite/agent/pull/599.

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -31,8 +31,7 @@ type Process struct {
 	Env        []string
 	ExitStatus string
 
-	buffer bytes.Buffer
-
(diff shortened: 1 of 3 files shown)


lox:
Woop woop!

PR #620, opened 2018-01-15
Fix bug where a failing glob would cause later globs to be ignored

### LINKED ISSUE #619: Agent artifact upload too agressive with missing artifacts
I'm running agent v3.0-beta.38.

I'm having issues with multiple artifact targets. These two builds are identical except for the artifacts definition.

```
$ buildkite-agent artifact upload build/playBinary/reports/test/xml/*.xml;build/distributions/playBinary.zip
--
  | 2018-01-15 00:28:12 DEBUG  Debug mode enabled
  | 2018-01-15 00:28:12 DEBUG  Searching for build/playBinary/reports/test/xml/*.xml
  | 2018-01-15 00:28:12 DEBUG  Searching for build/distributions/playBinary.zip
  | 2018-01-15 00:28:12 INFO   No files matched paths: build/playBinary/reports/test/xml/*.xml;build/distributions/playBinary.zip
```

```
buildkite-agent artifact upload build/playBinary/reports/test/xml/*.xml
--
  | 2018-01-15 00:30:31 INFO   Found 3 files that match "build/playBinary/reports/test/xml/*.xml"
  | 2018-01-15 00:30:31 INFO   Creating (0-3)/3 artifacts
  | 2018-01-15 00:30:32 INFO   Uploading artifact 672b7d8b-f7f3-4863-848b-824f3b0cf9b6 build/playBinary/reports/test/xml/TEST-services.1.xml (932 bytes)
  | 2018-01-15 00:30:32 INFO   Uploading artifact 3ac216f4-0bb9-4f3a-890f-9a7925ba1c2a build/playBinary/reports/test/xml/TEST-controllers.1.xml (30005 bytes)
  | 2018-01-15 00:30:32 INFO   Uploading artifact 1057ffe2-21bc-405c-9c0d-a6b5353c60a9 build/playBinary/reports/test/xml/TEST-services.2.xml (619 bytes)
```

I think the issue is here (https://github.com/buildkite/agent/blob/master/agent/artifact_uploader.go#L86). Because the build failed at tests - the second artifact doesn't exist. That line I linked is a `return nil` so if any of the multiple artifacts is missing it will return `nil` for all the given paths. It should probably be a `continue`instead.

lox:
This also brings in a revamped test for the uploader that I've been working on for windows compatibility. 

Closes #619.

### DIFF
--- a/agent/artifact_uploader.go
+++ b/agent/artifact_uploader.go
@@ -80,10 +80,11 @@ func (a *ArtifactUploader) Collect() (artifacts []*api.Artifact, err error) {
 
 		logger.Debug("Searching for %s", globPath)
 
-		// Resolve the globs (with * and ** in them)
+		// Resolve the globs (with * and ** in them), if it's a non-globbed path and doesn't exists
+		// then we will get the ErrNotExist that is handled below
 		files, err := zglob.Glob(globPath)
 		if err == os.ErrNotExist {
-			logger.Info("No files matched glob: %s", globPath)
+			logger.Info("File not found: %s", globPath)
 			continue
 		} else if err != nil {
 			return nil, err
--- a/agent/artifact_uploader.go
+++ b/agent/artifact_uploader.go
@@ -83,7 +83,8 @@ func (a *ArtifactUploader) Collect() (artifacts []*api.Artifact, err error) {
 		// Resolve the globs (with * and ** in them)
 		files, err := zglob.Glob(globPath)
 		if err == os.ErrNotExist {
-			return nil, nil
+			logger.Info("No files matched glob: %s", globPath)
+			continue
 		} else if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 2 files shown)


agent/artifact_uploader.go
```diff
@@ -83,7 +83,8 @@ func (a *ArtifactUploader) Collect() (artifacts []*api.Artifact, err error) {
 		// Resolve the globs (with * and ** in them)
 		files, err := zglob.Glob(globPath)
 		if err == os.ErrNotExist {
-			return nil, nil
+			logger.Info("No files matched glob: %s", globPath)
```
bharrisau: Technically - `zglob` only returns `os.ErrNotExist` when you pass in a non-globbed path. So the message would look this `INFO - No files matched glob: some/path/with/no/globs`. Maybe you'd want a slight re-wording of the log?

`"File not found: %s"`

PR #623, opened 2018-01-16
Show stderr from command output

lox:
Otherwise errors from commands get suppressed.

### DIFF
--- a/bootstrap/shell/shell.go
+++ b/bootstrap/shell/shell.go
@@ -311,7 +311,7 @@ func (s *Shell) executeCommand(cmd *exec.Cmd, w io.Writer, flags executeFlags) e
 		}
 	} else {
 		cmd.Stdout = w
-		cmd.Stderr = nil
+		cmd.Stderr = w
 		cmd.Stdin = nil
 
 		if s.Debug {


keithpitt:
Something something keyscan going into authorised keys something something? We may need to fix parts of the bootstrap process that do things with shell output if they now include errors.

lox:
Ah yeah, that is right. Good thinking! I'll rethink this.

lox:
The reasoning for this is that in some situations `git clone` outputs it's errors to stderr, so when it fails you can't see why in the bootstrap 🤔

keithpitt:
Oh I wasn't recommending you close it! I think it's a good change, and the things that do inspect the output should also consider errors. There may only be 2-3 places that you need to consider.

Also perhaps it can be an option to the shell runner thingy, and default it's on?

lox:
No you were totally correct! I think it probably needs a different approach. At the minute there is just `shell.Run` and `shell.RunAndCapture`, both explicitly drop stderr.

lox:
One option would be to pass stderr directly to the system stderr, but that makes it hard to test and reason about what is going on.

PR #628, opened 2018-01-16
Support older git versions with submodules

### LINKED ISSUE #602: Bootstrap v3 git submodule error/warning
I noticed when we upgraded one of our agents to a v3 beta on an Ubuntu 14.04.5 LTS host we now get this during "Preparing build directory":

```
# Checking to see if submodule urls need to be added to known_hosts
$ git submodule foreach --recursive git remote get-url origin
⚠️ Warning: Failed to enumerate git submodules: Error running `/usr/bin/git submodule foreach --recursive git remote get-url origin`: exit status 1
```

Running this command on the host in a build directory reveals:

```
$ git submodule foreach --recursive git remote get-url origin
Entering 'vendor/docs'
error: Unknown subcommand: get-url
usage: git remote [-v | --verbose]
   or: git remote add [-t <branch>] [-m <master>] [-f] [--tags|--no-tags] [--mirror=<fetch|push>] <name> <url>
   or: git remote rename <old> <new>
   or: git remote remove <name>
   or: git remote set-head <name> (-a | --auto | -d | --delete |<branch>)
   or: git remote [-v | --verbose] show [-n] <name>
   or: git remote prune [-n | --dry-run] <name>
   or: git remote [-v | --verbose] update [-p | --prune] [(<group> | <remote>)...]
   or: git remote set-branches [--add] <name> <branch>...
   or: git remote set-url [--push] <name> <newurl> [<oldurl>]
   or: git remote set-url --add <name> <newurl>
   or: git remote set-url --delete <name> <url>

    -v, --verbose         be verbose; must be placed before a subcommand

Stopping at 'vendor/docs'; script

$ git --version
git version 1.9.1
```

This is the latest version of git available for trusty.

1. We should probably be printing error output or something
2. We may need to look for the url in a more compatible way?
3. Or we need to have a minimum version of git and complain loudly when it's too old.

lox:
Previously `git remote get-url origin` was used to determine the origin url for submodules. This was only introduced in git 2.7, so it caused errors on older installations (e.g Ubuntu 14.04). 

This replaces it with `git config --get remote.origin.url`, which from what I can tell should be backwards compatible and acomplish the same thing.

Closes #602.

### DIFF
--- a/bootstrap/git.go
+++ b/bootstrap/git.go
@@ -89,7 +89,7 @@ func gitEnumerateSubmoduleURLs(sh *shell.Shell) ([]string, error) {
 	// Entering 'vendor/frontend/vendor/emojis'
 	// git@github.com:buildkite/emojis.git
 	output, err := sh.RunAndCapture(
-		"git", "submodule", "foreach", "--recursive", "git", "remote", "get-url", "origin")
+		"git", "submodule", "foreach", "--recursive", "git", "config", "--get", "remote.origin.url")
 	if err != nil {
 		return nil, err
 	}
--- a/bootstrap/git.go
+++ b/bootstrap/git.go
@@ -89,7 +89,7 @@ func gitEnumerateSubmoduleURLs(sh *shell.Shell) ([]string, error) {
 	// Entering 'vendor/frontend/vendor/emojis'
 	// git@github.com:buildkite/emojis.git
 	output, err := sh.RunAndCapture(
-		"git", "submodule", "foreach", "--recursive", "git", "config", "--get", "remote.origin.url")
+		"git", "submodule", "foreach", "--recursive", "git", "ls-remote", "--get-url")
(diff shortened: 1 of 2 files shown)


sj26:
There's quite a bit of logic behind calculating remote urls. Using `git config ...` should be enough for most cases. But could we use `git ls-remote --get-url` instead? This has been around since at least [git v2.0.0](https://github.com/git/git/blob/v2.0.0/builtin/ls-remote.c#L7-L8) and seems to fully consider the remote config.

lox:
Pretty sure older Ubuntu's are still git 1.9! 

We only use this to get a hostname to add to ssh known hosts. I think it's ok if it doesn't handle aliases.

lox:
This is lots faster too, as it doesn't require a network lookup.

lox:
Yup, I'm ok with that!

PR #631, opened 2018-01-19
Fix parsing of git urls with windows volumes in them

lox:
This change handles parsing urls like `file:///C:/Users/vagrant/repo`.

### DIFF
--- a/bootstrap/git_test.go
+++ b/bootstrap/git_test.go
@@ -1,54 +1,72 @@
 package bootstrap
 
 import (
-	"os"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
 )
 
-func TestParsingGittableRepository(t *testing.T) {
+func TestParsingGittableRepositoryFromFilesPaths(t *testing.T) {
 	t.Parallel()
 
-	wd, _ := os.Getwd()
-
-	var testCases = []struct {
-		Ref      string
-		Expected string
-		IsErr    bool
-	}{
-		{wd, wd, false},
-		{"git@github.com:buildkite/agent.git", "ssh://git@github.com/buildkite/agent.git", false},
-		{"git@github.com-alias1:buildkite/agent.git", "ssh://git@github.com-alias1/buildkite/agent.git", false},
-		{"ssh://git@scm.xxx:7999/yyy/zzz.git", "ssh://git@scm.xxx:7999/yyy/zzz.git", false},
-		{"ssh://root@git.host.de:4019/var/cache/git/project.git", "ssh://root@git.host.de:4019/var/cache/git/project.git", false},
+	u, err := parseGittableURL(`/home/vagrant/repo`)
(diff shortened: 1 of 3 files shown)


lox:
This was a particularly gross usage of table tests. I think the result post discussion in https://github.com/buildkite/agent/pull/630 tells a much better story. Feels @keithpitt?

bootstrap/git_test.go
```diff
@@ -1,54 +1,72 @@
...
-		Expected string
-		IsErr    bool
-	}{
-		{wd, wd, false},
-		{"git@github.com:buildkite/agent.git", "ssh://git@github.com/buildkite/agent.git", false},
-		{"git@github.com-alias1:buildkite/agent.git", "ssh://git@github.com-alias1/buildkite/agent.git", false},
-		{"ssh://git@scm.xxx:7999/yyy/zzz.git", "ssh://git@scm.xxx:7999/yyy/zzz.git", false},
-		{"ssh://root@git.host.de:4019/var/cache/git/project.git", "ssh://root@git.host.de:4019/var/cache/git/project.git", false},
+	u, err := parseGittableURL(`/home/vagrant/repo`)
+	if err != nil {
```
keithpitt: Should this be...

```go
 assert.Nil(err)
```

..instead?

keithpitt: I also learnt about this syntax:

```
func TestSomething(t *testing.T) {
  assert := assert.New(t)

  assert.Nil(err)
}
```

So you don't need to pass `t` around everywhere.

lox: The problem with that is that assert.Nil isn't actually fatal, it doesn't stop the test.  I tend to think the if block with a Fail is a lot clearer here. Go's error handling is meant to look like this :)

lox: The alternative would be:

```
if !assert.NoError(err) {
  t.FailNow()
}
```

Which is ok I guess? Doesn't seem worth the cognitive overhead.

keithpitt: Orrr....

```go
import (
  "github.com/stretchr/testify/require"
)

func TestSomething(t *testing.T) {
  require := require.New(t)

  require.Nil(err)
}
```

That'll blow up the test (instead of running) if it fails! Maybe better/worse?

keithpitt: (See: https://godoc.org/github.com/stretchr/testify/require)

lox: It makes me sad :(

lox: Reckon I'm going to stick with this as is rather than bring in a whole new package.

keithpitt: Can you just use `require` instead of `assert` for the whole thing?

lox: Yup, we could. I guess if we are doing the whole testify thing we might as well bring it all in.

PR #636, opened 2018-01-21
Support bash hooks on windows

lox:
@keithpitt had a neat suggestion for how to support plugin hooks on windows, which was to use the bash.exe bundled with Git for Windows. This generalizes that and searches for the best option under windows, either batch, powershell or if neither of those exist it uses the bash/sh options.


toolmantim:
Nice!!

keithpitt:
Rad! Would this mean our plugins "just work" for the most part then?

lox:
> Rad! Would this mean our plugins "just work" for the most part then?

Yeah, hopefully :)

lox:
```bash
# Executing "C:\Users\ContainerAdministrator\AppData\Local\Temp\buildkite-agent-bootstrap-hook-runner-066954642"
--
  | > C:\Program Files\Git\usr\bin\bash.exe -c C:/Users/ContainerAdministrator/AppData/Local/Temp/buildkite-agent-bootstrap-hook-runner-066954642
  | This is an example of a pre-command hook from .buildkite/hooks/pre-command
```

lox:
I'm struggling with this error:

```
> buildkite-agent pipeline upload
--
  | 2018-01-21 20:06:12 DEBUG  Debug mode enabled
  | 2018-01-21 20:06:12 INFO   Reading pipeline config from STDIN
  | 2018-01-21 20:06:12 FATAL  Config file is empty
  | Exited with error: exit status 1
```

Any ideas @keithpitt?

lox:
Ah, right. It seems to be mistakenly detecting that there is a STDIN.

bootstrap/bootstrap.go
```diff
@@ -195,28 +198,64 @@ func (b *Bootstrap) applyEnvironmentChanges(environ *env.Environment, dir string
...
 }
 
-// Returns the absolute path to a global hook
-func (b *Bootstrap) globalHookPath(name string) string {
-	return filepath.Join(b.HooksPath, normalizeScriptFileName(name))
+// Returns the absolute path to the best matching hook file in a path, or os.ErrNotExist if none is found
+func (b *Bootstrap) findHookFile(hookDir string, name string) (string, error) {
+	if runtime.GOOS == "windows" {
+		// check for windows types first
+		if p, err := shell.LookPath(name, hookDir, ".BAT;.CMD;.EXE"); err == nil {
```
lox: Unfortunately we can't support powershell hooks yet, as need to write a environment hook magic thing for them.

bootstrap/bootstrap.go
```diff
@@ -231,19 +270,15 @@ func fileExists(filename string) bool {
...
-	}
-	return filename
-}
-
 func dirForAgentName(agentName string) string {
 	badCharsPattern := regexp.MustCompile("[[:^alnum:]]")
 	return badCharsPattern.ReplaceAllString(agentName, "-")
 }
 
+func buildkiteAgentBinary() string {
```
lox: This replaces hardcoded `buildkite-agent`

clicommand/annotate.go
```diff
@@ -108,7 +108,7 @@ var AnnotateCommand = cli.Command{
 
 		if cfg.Body != "" {
 			body = cfg.Body
-		} else if stdin.IsPipe() {
+		} else if terminal.IsTerminal(int(os.Stdin.Fd())) {
```
keithpitt: Ah, is this the cross platform way to do it?

lox: Yup, or at least it let's it be someone else's problem. Our implementation didn't seem to work on Windows, despite it saying it did.

bootstrap/hook.go
```diff
@@ -97,12 +105,11 @@ func newHookScriptWrapper(hookPath string) (*hookScriptWrapper, error) {
 			"SET > \"" + h.afterEnvFile.Name() + "\"\n" +
 			"EXIT %" + hookExitStatusEnv + "%"
 	} else {
-		script = "#!/bin/bash\n" +
```
lox: I ditched this shebang, we always invoke these scripts with an interpreter and it was causing problems on windows bash.

PR #640, opened 2018-01-22
Set Content-Disposition on uploads to GCS

### LINKED ISSUE #634: Set a more sensible Content-Disposition on GCS objects
We’ve migrated our build artifacts from S3 to GCS (closer to our agents), I think on s3 the default filename when downloading via the browser was `basename(key)`, but on GCS it’s just `key`, and these can be quite long. In some cases they're truncated when a user downloads a file, which isn't ideal.

[Content-Disposition](https://cloud.google.com/storage/docs/metadata#content-disposition) can be set on [`Object`](https://godoc.org/google.golang.org/api/storage/v1#Object), would it be acceptable to set a more sensible value on upload [in the agent](https://github.com/buildkite/agent/blob/35aa57a58736eff33842fa935db8f63e3b822635/agent/gs_uploader.go#L86-L88)? No disposition header is set [by default](https://gist.github.com/DazWorrall/af31c3cbe6842b2033098ea3dd53e419), so I would suggest `inline; filename=#{basename(key)}`. From the [RFC](https://tools.ietf.org/html/rfc6266):

>  On the other hand, if it matches "inline" (case-insensitively), this implies default processing.  Therefore, the disposition type "inline" is only useful when it is augmented with additional parameters, such as the filename.

DazWorrall:
Closes #634 

This makes the default filenames more sensible when downloading from GCS in a browser:

2018-01-22 at 14 03

I have tested this with directory trees and files with spaces in the name, they are handled correctly.

### DIFF
--- a/agent/gs_uploader.go
+++ b/agent/gs_uploader.go
@@ -84,8 +84,9 @@ func (u *GSUploader) Upload(artifact *api.Artifact) error {
 			u.artifactPath(artifact), u.BucketName(), permission)
 	}
 	object := &storage.Object{
-		Name:        u.artifactPath(artifact),
-		ContentType: u.mimeType(artifact),
+		Name:               u.artifactPath(artifact),
+		ContentType:        u.mimeType(artifact),
+		ContentDisposition: u.contentDisposition(artifact),
 	}
 	file, err := os.Open(artifact.AbsolutePath)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


lox:
Looks great! Any idea if we should be doing the same thing for our S3 integration?

DazWorrall:
@lox I don't know if we were doing anything special in our stack, but we never had this issue while using S3 so it's possible the defaults are just more sensible there.

lox:
We will get this out in the next beta!

PR #641, opened 2018-01-24
Fix incorrect polarity on IsTerminal check

lox:
I broke pipeline uploads from stdin in #636. I removed the previous `IsStdin` technology in favour of  https://godoc.org/golang.org/x/crypto/ssh/terminal#IsTerminal, but I neglected to account for them having opposite meanings. 😓


keithpitt:
Whoops, sorry I missed this in review!

Issue #615, opened 2018-01-10
Add 'buildkite-runner env' command to get the job's environment variables.


sander-m:
It would be useful if `buildkite-agent` has a command to output its current set of environment variables, without host/system variables mixed in. I am aware of the REST API to do this (https://buildkite.com/docs/rest-api/jobs#get-a-jobs-environment-variables). This API gives the desired functionality but requires an API call for every job we run.

For example, `buildkite-agent env` could return the current Buildkite environment associated to the running job.
```json
{
  "CI": "true",
  "BUILDKITE": "true",
  "BUILDKITE_TAG": "",
  "BUILDKITE_REPO": "git@github.com:my-great-org/my-repo.git",
  "BUILDKITE_BRANCH": "master",
  "BUILDKITE_COMMIT": "a65572555600c07c7ee79a2bd909220e1ca5485b",
  "BUILDKITE_JOB_ID": "bde076a8-bc2c-4fda-9652-10220a56d638",
  "BUILDKITE_COMMAND": "buildkite-agent pipeline upload",
  "BUILDKITE_MESSAGE": ":llama:",
  "BUILDKITE_BUILD_ID": "c4e312cb-e734-4f0a-a5bd-1cac2535c57e",
  "BUILDKITE_BUILD_URL": "https://buildkite.com/my-great-org/my-pipeline/builds/15",
  "BUILDKITE_AGENT_NAME": "ci-1",
  "BUILDKITE_COMMAND": "buildkite-agent pipeline upload",
  "BUILDKITE_BUILD_NUMBER": "15",
  "BUILDKITE_ORGANIZATION_SLUG": "my-great-org",
  "BUILDKITE_PIPELINE_SLUG": "my-pipeline",
  "BUILDKITE_PULL_REQUEST": "false",
  "BUILDKITE_BUILD_CREATOR": "Keith Pitt",
  "BUILDKITE_REPO_SSH_HOST": "github.com",
  "BUILDKITE_ARTIFACT_PATHS": "",
  "BUILDKITE_PIPELINE_PROVIDER": "github",
  "BUILDKITE_BUILD_CREATOR_EMAIL": "keith@buildkite.com",
  "BUILDKITE_AGENT_META_DATA_LOCAL": "true"
}
```

lox:
Yup, I've always wanted this too :) What if it was more like the [meta-data](https://buildkite.com/docs/agent/cli-meta-data) command and allowed `get`/`set`/`list`?

lox:
[...]

PR #643, opened 2018-01-26
Add `--write-env-file` cli option

DazWorrall:
Partially addresses #615

This fulfils our desire to present only a clean environment to docker container (our runner gear can check for the presence of this file), and a future `buildkite-agent env` command can simply echo this back if it exists locally, before falling back to agent API.

WDYT @lox?

### DIFF
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -187,6 +187,17 @@ func (r *JobRunner) createEnvironment() []string {
 		env[key] = value
 	}
 
+	// Write out the job environment to a file if configured, in k=v format.
+	// We do this early to present only the clean environment, i.e excluding
+	// any varibales set or inherited here.
+	if r.envFile != nil {
+		for key, value := range(env) {
+			r.envFile.WriteString(fmt.Sprintf("%s=%s\n", key, value))
+		}
+		r.envFile.Close()
+		env["BUILDKITE_ENV_FILE"] = r.envFile.Name()
+	}
+
 	// Add agent environment variables
 	env["BUILDKITE_AGENT_ENDPOINT"] = r.Endpoint
 	env["BUILDKITE_AGENT_ACCESS_TOKEN"] = r.Agent.AccessToken
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -66,14 +66,12 @@ func (r JobRunner) Create() (runner *JobRunner, err error) {
 	// the Buildkite Agent API
 	runner.logStreamer = LogStreamer{MaxChunkSizeBytes: r.Job.ChunksMaxSizeBytes, Callback: r.onUploadChunk}.New()
 
-	// If configured to write the job env to a file, create that file
-	if runner.AgentConfiguration.WriteEnvFile {
-		if file, err := ioutil.TempFile("", fmt.Sprintf("job-env-%s", runner.Job.ID)) ; err != nil {
-			return runner, err
-		} else {
-			logger.Debug("[JobRunner] Created env file: %s", file.Name())
-			runner.envFile = file
-		}
+	// Prepare a file to recieve the given job environment
+	if file, err := ioutil.TempFile("", fmt.Sprintf("job-env-%s", runner.Job.ID)) ; err != nil {
+		return runner, err
+	} else {
+		logger.Debug("[JobRunner] Created env file: %s", file.Name())
+		runner.envFile = file
 	}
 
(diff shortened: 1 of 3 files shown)


lox:
Hey @DazWorrall, thanks for taking the time to implement this and also being patient whilst we thought it over! 

I know I suggested it initially, but reckon we should ditch the config option and just expose a env file to each job with it's path in `BUILDKITE_ENV_FILE`. Happy to collab, or if you want to have a shot at it, that would be amazing :)

lox:
(What you suggested in https://github.com/buildkite/agent/issues/615#issuecomment-360485623 😓)

lox:
One question we've been talking back and forward on is whether you'd expect environmental changes from hooks to be reflected in this file. Thoughts?

DazWorrall:
We wouldn't I don't think - we're explicitly looking to see what was configured on the job (i.e. the equivalent of calling the `jobs/env` [REST api](https://buildkite.com/docs/rest-api/jobs#get-a-jobs-environment-variables)), to provide a clean environment to a docker container. We may whitelist some other variables from the agent environment as well, but we still need to know what was given to the job.

DazWorrall:
@lox removed the config option.

agent/job_runner.go
```diff
@@ -62,6 +66,14 @@ func (r JobRunner) Create() (runner *JobRunner, err error) {
 	// the Buildkite Agent API
 	runner.logStreamer = LogStreamer{MaxChunkSizeBytes: r.Job.ChunksMaxSizeBytes, Callback: r.onUploadChunk}.New()
 
+	// Prepare a file to recieve the given job environment
+	if file, err := ioutil.TempFile("", fmt.Sprintf("job-env-%s", runner.Job.ID)) ; err != nil {
+		return runner, err
+	} else {
```
lox: This doesn't need the else block, see https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow.

agent/job_runner.go
```diff
@@ -130,6 +142,15 @@ func (r *JobRunner) Run() error {
 	logger.Debug("[JobRunner] Waiting for all other routines to finish")
 	r.routineWaitGroup.Wait()
 
+	// Remove the env file, if any
+	if r.envFile != nil {
+		if err := os.Remove(r.envFile.Name()); err != nil {
+			logger.Warn("[JobRunner] Error cleaning up env file: %s", err)
+		} else {
```
lox: Same as above with else block.

lox:
Thanks @DazWorrall will talk over with @keithpitt!

agent/job_runner.go
DazWorrall: Without the else block `file` will be out of scope:

```.go
if file, err := ioutil.TempFile("", fmt.Sprintf("job-env-%s", runner.Job.ID)) ; err != nil {
    return runner, err
}
logger.Debug("[JobRunner] Created env file: %s", file.Name())
runner.envFile = file
```
Will fail to compile with `agent/job_runner.go:73:51: undefined: file`.

I could do:

```.go
file, err := foo()
if err != nil { }
```

But that doesn't match the style elsewhere.

lox: > But that doesn't match the style elsewhere.

Yup, that's fair. Match the style of the surrounding code and I'll deal with it holistically to match the golang style at some point.

lox:
Cool @keithpitt and I discussed it. I think we will subsequently write env updates to that file in our bootstrap, but that shouldn't affect you, or if it does you could take a snapshot of the env file in a bootstrap script or the env hook.

Beyond that, the only concern that popped up was how we'd handle newlines in env variables. Docker is notoriously bad at this https://github.com/moby/moby/issues/12997 https://github.com/docker/compose/issues/3527. This can be hacked around if needed at least. 

I'm going to merge this and then do some testing. Thanks so much driving this discussion and for the contribution! We'll endeavour to get this out in a release soon.

DazWorrall:
We already sanitise variables before passing them to docker FYI, not our first brush with newlines 🙂

sander-m:
@lox Thanks for shipping this quickly. I have done some testing and we're running into issues with the newlines in environment variables (as you pointed out already). We have concluded that encoding the output file as JSON would solve our issues. That would make it trivial to parse for our use case.

lox:
Another option would be the export format we parse here https://github.com/buildkite/agent/blob/master/env/export.go

PR #648, opened 2018-02-09
Allow the shell interpreter to be configured

lox:
Currently the agent guesses the shell based on your operating system, either `/bin/bash` on nix or `cmd` on windows.

This allows for the shell used for interpreting `$BUILDKITE_COMMAND` to be configured. 

IMO eventually we should default to `/bin/sh`, but for backwards compat, we will go with `/bin/bash`.

This controversially also ditches the behaviour where we write a shellscript for each command to be executed. 

image

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -41,6 +41,7 @@ type AgentStartConfig struct {
 	BuildPath                    string   `cli:"build-path" normalize:"filepath" validate:"required"`
 	HooksPath                    string   `cli:"hooks-path" normalize:"filepath"`
 	PluginsPath                  string   `cli:"plugins-path" normalize:"filepath"`
+	Shell                        string   `cli:"shell"`
 	Tags                         []string `cli:"tags"`
 	TagsFromEC2                  bool     `cli:"tags-from-ec2"`
 	TagsFromEC2Tags              bool     `cli:"tags-from-ec2-tags"`
@@ -63,6 +64,15 @@ type AgentStartConfig struct {
 	MetaDataEC2     bool     `cli:"meta-data-ec2" deprecated-and-renamed-to:"TagsFromEC2"`
 	MetaDataEC2Tags bool     `cli:"meta-data-ec2-tags" deprecated-and-renamed-to:"TagsFromEC2Tags"`
 	MetaDataGCP     bool     `cli:"meta-data-gcp" deprecated-and-renamed-to:"TagsFromGCP"`
+}
+
+func DefaultShell() string {
+	if runtime.GOOS == "windows" {
+		return "C:\\Windows\\System32\\CMD.EXE /S /C"
+	} else {
+		return "/bin/bash -c"
+
+	}
 }
 
 func DefaultConfigFilePaths() (paths []string) {
@@ -129,6 +139,12 @@ var AgentStartCommand = cli.Command{
 			Value:  120,
 			Usage:  "When --disconnect-after-job is specified, the number of seconds to wait for a job before shutting down",
 			EnvVar: "BUILDKITE_AGENT_DISCONNECT_AFTER_JOB_TIMEOUT",
+		},
+		cli.StringFlag{
+			Name:   "shell",
+			Value:  DefaultShell(),
+			Usage:  "The shell to use to interpret build commands",
+			EnvVar: "BUILDKITE_SHELL",
 		},
 		cli.StringSliceFlag{
 			Name:   "tags",
@@ -277,6 +293,11 @@ var AgentStartCommand = cli.Command{
 			cfg.BootstrapScript = fmt.Sprintf("%q bootstrap", os.Args[0])
 		}
 
+		// Guess the shell if none is provided
+		if cfg.Shell == "" {
+			cfg.Shell = DefaultShell()
+		}
+
 		// Make sure the DisconnectAfterJobTimeout value is correct
 		if cfg.DisconnectAfterJob && cfg.DisconnectAfterJobTimeout < 120 {
 			logger.Fatal("The timeout for `disconnect-after-job` must be at least 120 seconds")
@@ -316,6 +337,7 @@ var AgentStartCommand = cli.Command{
 				TimestampLines:             cfg.TimestampLines,
 				DisconnectAfterJob:         cfg.DisconnectAfterJob,
 				DisconnectAfterJobTimeout:  cfg.DisconnectAfterJobTimeout,
+				Shell: cfg.Shell,
 			},
 		}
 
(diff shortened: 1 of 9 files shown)


dch:
would switching to `/bin/sh` be worth considering as part of the 3.0 breaking changes,or will you bum that to a future release?

lox:
@dch the issue is that lots of folks tend to use bash semantics without realising it in their command steps (for better or worse). 

We'll talk it over. Let's see if I can get this PR right first :)

lox:
In terms of directly invoking the commands vs writing a tmp script file, this was my pros/cons list:

**PROS**

* Radically simpler, less moving parts
* No dependency on the local filesystem, which makes things much simpler for environments like docker
* Plays well with switching your shell to something else, for instance Powershell on Windows

**CONS**

* Length limits for how long a set of commands can be, e.g ARG_MAX
* If folks have shebangs at the start of their "slab of commands", it just won't work
* We need to pass the -e to bash to provide the same semantics as we previously had
* Maybe it looks a bit less tidy at a process-level view?
* Users might need to understand shells a bit better?

keithpitt:
> If folks have shebangs at the start of their "slab of commands", it just won't work

This seems like a pretty major con? As in, some people stuff will just stop working?

keithpitt:
> Maybe it looks a bit less tidy at a process-level view?

Can you paste a screenshot of the before/after? Part of the bootstrap magic is how awesome the output looks - I wouldn't really want to compromise on that.

keithpitt:
> No dependency on the local filesystem, which makes things much simpler for environments like docker

How so? I thought your "executor" thingy would have solved that bit? And the "local executor" we have no just needs file system access?

lox:
Oh, scratch the shebang one off the list:

```bash
bash -c '#!/bin/bash                                                                                        
echo hello world'
hello world
```

lox:
> Can you paste a screenshot of the before/after? Part of the bootstrap magic is how awesome the output looks - I wouldn't really want to compromise on that.

@keithpitt I meant how the `ps -auxxx` would look on the host running the agent. The bootstrap output ends up with considerably less cruft in it after this (which I will now add to PROS).

lox:
> How so? I thought your "executor" thingy would have solved that bit? And the "local executor" we have now just needs file system access?

Nope, it would be an annoying and manual process for all pluggable executors. Each one would have to have it's own set of techniques for getting the temporary script over.

clicommand/agent_start.go
```diff
@@ -65,6 +66,14 @@ type AgentStartConfig struct {
...
 }
 
+func DefaultShell() string {
+	if runtime.GOOS == "windows" {
+		return `C:\Windows\System32\CMD.EXE /S /C`
+	} else {
+		return `/bin/bash -e -c`
+	}
+}
+
```
dch: Do you feel like adding another block here? or maybe it's time to use a switch?

```
if runtime.GOOS == "freebsd" { 
return `/usr/local/bin/bash -e -c`
} else if {
...
```

or
```
switch runtime.GOOS {
case "windows":
	shell_path = `C:\Windows\System32\CMD.exe /S /C`
case "freebsd", "openbsd", "netbsd":
    shell_path = `/usr/local/bin/bash -e -c`
default:
  # assume linux or is it time to add illumos/smartos/... ;-)
   shell_path = `/bin/bash -e -c`
}
...
```

BTW I'd expect there are a couple of other places, at present I "rewrite" this with a blunt object https://github.com/ideegeo/ports/blob/master/devel/buildkite-agent/Makefile#L30-L34

lox: Yeah that sounds pretty reasonable to me. Alternately we could just use `bash` and do a path lookup on it.

lox: Or maybe we just do drop back to `/bin/sh`.

lox:
There are some kinks to work out 🤔

before
after

I'm torn on echoing each line. Feels?

lox:
I think I prefer keeping it minimal and just showing the shell command and the output. It lets the output speak for itself.

clicommand/agent_start.go
lox: Nah, I think defaulting to /bin/sh would result in too many breakages for us.

lox: Added a switch.

keithpitt:
I'm not particularly tied to the approach of how to run commands (writing the script I mean). If there are more pros than cons then I'm totally fine with it.

Does this approach allow us to run slightly complex pipelines like this?

```yaml
steps:
  - command: <
      #!/bin/bash
      echo "The \$BUILDKITE_RETRY_COUNT value is $BUILDKITE_RETRY_COUNT"
      if [[ "$BUILDKITE_RETRY_COUNT" == "0" ]]; then
        echo "Gonna fail 👎"
        exit 1
      elif [[ "$BUILDKITE_RETRY_COUNT" == "1" ]]; then
        echo "Gonna fail 👎 1 last time"
        exit 2
      else
        echo "Horray! It works! 👍"
        exit 0
      fi
```

Hrm...just thinking about it more...if I do a `export FOO=bar` in a command, will that environment variable be available in a `post-command` hook/plugin? I think that's why I did the script writing in the first place (so there's a place to capture that stuff).

lox:
> Hrm...just thinking about it more...if I do a export FOO=bar in a command, will that environment variable be available in a post-command hook/plugin?

Hmmm... no, I don't think so. Did we ever support that?

lox:
The command you posted would run fine! The `#!/bin/bash` is ignored though.

keithpitt:
> Hmmm... no, I don't think so. Did we ever support that?

We didn't, I'm remembering lies. I think we supported it in the pure bootstrap.sh version of the file, but not in the improved Golang version. Do you think we should support it? (knowing that it kinda makes your proposed change not possible).

I kinda would expect any `exports` to carry across, but there are many caveats to it working (just like now with hooks).

Whatta ya rekon?

keithpitt:
>  ends up around 128k characters

I think that's totally fine. Anything longer and you can just put in a script.

lox:
>  I think we supported it in the pure bootstrap.sh version of the file, but not in the improved Golang version. 

Nah, that was where I looked first. I couldn't find any prior art for normal commands. If you have a command hook, it will do the env dance, but not a non-hook command.

The thing that weirds me out about passing env is that for lots of folks it's unexpected (me included), because it's not how shells work. Subshells get a copy of env, but they can't influence shells that the parent process subsequently creates. 

My take would be to not add magic env passing to the command, unless there is prior art in the 2.x series. I'd love to see that env passing mechanism be more explicit via something like `buildkite-agent env set LLAMAS "always"`.

lox:
Thanks @keithpitt! I'm sure we'll run into reasons why it's a bad idea, but for the moment I think it's a good simplification!

PR #652, opened 2018-02-19
Check hook script write

sj26:
If a disk fills up or for some other reason we fail to write this hook file contents the bootstrap just plods along anyway, running an empty script file. So check for errors during the write. Also a couple of other little tweaks nearby.

### DIFF
--- a/bootstrap/hook.go
+++ b/bootstrap/hook.go
@@ -115,7 +115,10 @@ func newHookScriptWrapper(hookPath string) (*hookScriptWrapper, error) {
 	}
 
 	// Write the hook script to the runner then close the file so we can run it
-	h.scriptFile.WriteString(script)
+	_, err = h.scriptFile.WriteString(script)
+	if err != nil {
+		return nil, err
+	}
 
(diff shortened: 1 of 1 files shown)


sj26:
I don't understand what's happening with the build. Will have to investigate.

sj26:
The windows agents had run out of disk space. I now know more than I ever wanted to about running windows on aws and resizing volumes. Which is to say: very little. But the build seems to be going again.

lox:
More error handling is good, been meaning to cleanup these linter errors. Thanks!

bootstrap/hook.go
```diff
@@ -64,6 +64,7 @@ func newHookScriptWrapper(hookPath string) (*hookScriptWrapper, error) {
 	if err != nil {
 		return nil, err
 	}
+	defer h.scriptFile.Close()
```
lox: Should we be catching this error?

sj26: I'm not sure! I might merge what we've got and leave that for your perusal later?

PR #659, opened 2018-02-25
Move windows agent tests back to asynch

lox:
Due to recent issues with our windows instances being flakey, this moves things back to an asynch pipeline trigger of the windows-agent pipeline.


lox:
Gonna sneaky merge this.

PR #662, opened 2018-02-27
Update to latest version of bintest

lox:
This moves from the old lox/bintest repo to the new buildkite/bintest repo.

Amongst other improvements, this lets us use the bintest compile cache which I am hoping will result in a considerable test speed improvement.


lox:
Gonna sneaky merge this given it passes.

PR #664, opened 2018-03-07
add submodule checkout to windows agent

lankaapura:
Seems submodule checkout is missing in 2.6 but it is available in non-windows version.
https://github.com/buildkite/agent/issues/52.

I'd personally prefer to disable submodule by default but I've enabled it following the non-windows implementation.


lox:
Thanks for this! Looks like the indentation isn't right, are you able to tweak it and we'll merge it?

Btw, the v3 agent has much stronger windows support!

lankaapura:
@lox tabs vs spaces 😄 fixed it. Great, I'll check it out. It should work with the existing pipeline.yml right?

lox:
@lankaapura yup, the v3 agent will work with the existing pipeline.yml!

PR #668, opened 2018-03-09
Replace github.com/ghodss/yaml with go-yaml

lox:
Currently we use github.com/ghodss/yaml for parsing YAML in a way that easily roundtrips to JSON. The issue that this brings is that there is no mechanism for ordered maps like yaml.MapSlice (see https://blog.labix.org/2014/09/22/announcing-yaml-v2-for-go#mapslice) in https://godoc.org/gopkg.in/yaml.v2. This has caused us problems when trying to pre-parse things like the top level env map.

The previous code went to a lot of trouble to guess whether code was JSON or YAML. The thing is that JSON is a subset of YAML, and we can basically treat everything at YAML. The only compromise here is that parse errors talk about YAML. Technically this is correct, it's all YAML. IMO this is fine. 

The other annoying compromise here is that we can't just use gopkg.in/yaml.v2 because there is a bug with anchors and yaml.MapSlice at https://github.com/go-yaml/yaml/issues/184. This uses a fork from https://github.com/go-yaml/yaml/pull/220. We can go back to the main library once this gets merged.


agent/pipeline_parser_test.go
```diff
@@ -70,24 +73,18 @@ steps:
...
+	assert.Equal(t, `Failed to parse yaml: found character that cannot start any token`, fmt.Sprintf("%s", err))
 
 	// It parses unknown JSON objects
 	result, err = PipelineParser{Pipeline: []byte("\n\n     \n  { \"foo\": \"bye ${ENV_VAR_FRIEND}\" }\n"), Env: environ}.Parse()
 	assert.Nil(t, err)
 	j, err = json.Marshal(result)
 	assert.Equal(t, `{"foo":"bye \"friend\""}`, string(j))
 
-	// It parses unknown JSON arrays
+	// It returns an error for arrays
```
lox: Any idea why we ever wanted to support this @keithpitt?

ticky:
A major issue with this is that it seems to coerce boolean values to strings, which causes them to fail validation in the pipeline uploader. We specifically look for a real `true` or `false` in the `required` field of `block` steps, producing this error on a previously-valid pipeline:

```
POST http://agent.buildkite.localhost/v3/jobs/bb078d93-7cd0-4269-8aaf-a90c46da6292/pipelines: 422 `required` must be a boolean (`true` or `false`) (Attempt 1/5 Retrying in 1s)
```

Debugging revealed this was passed to this point as a string.

PR #670, opened 2018-03-13
Retry failed checkouts

### LINKED ISSUE #477: Retry/sleep Git checkouts
Sometimes GitHub web hooks sender faster than the time it takes for their Git servers to update. So builds end up starting before the commit exists.

This often surfaces as this error:

```
$ git checkout -f <hash>
fatal: reference is not a tree: <hash>
```

It would be good if the agent could "guess" if the checkout failed because of this, and do an automatic sleep/retry (with a sensible "give up" limit).

This should be pretty straight forward since we moved the Bootstrap to "Golang" - we already have a `retry` package that can help here too.

lox:
@keithpitt finally convinced me that we should be doing checkout retries by default. 

Closes #477.

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -13,14 +13,9 @@ import (
 	"github.com/buildkite/agent/agent"
 	"github.com/buildkite/agent/bootstrap/shell"
 	"github.com/buildkite/agent/env"
-	"github.com/buildkite/agent/experiments"
 	"github.com/buildkite/agent/retry"
 	shellwords "github.com/mattn/go-shellwords"
 	"github.com/pkg/errors"
-)
-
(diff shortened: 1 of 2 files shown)


bootstrap/bootstrap.go
```diff
@@ -599,21 +594,15 @@ func (b *Bootstrap) CheckoutPhase() error {
...
 			if err != nil {
-				return err
-			}
-		} else {
-			if err := doCheckout(); err != nil {
-				return err
+				b.shell.Warningf("%s (%s)", err, s)
 			}
+			return err
+		}, &retry.Config{Maximum: 10, Interval: 5 * time.Second})
```
keithpitt: Tweak values here

PR #680, opened 2018-03-13
JSON-encode environment file

sander-m:
The way we currently write the environment file, as implemented by https://github.com/buildkite/agent/pull/643, is difficult to parse because environment variables could contain newlines.

Sourcing this file is therefore not safe, and parsing it into a map-like structure is hard. I'd suggest writing this file in JSON format. This makes it easy to parse.

### DIFF
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -1,6 +1,7 @@
 package agent
 
 import (
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -188,9 +189,8 @@ func (r *JobRunner) createEnvironment() []string {
 	// We present only the clean environment - i.e only variables configured
 	// on the job upstream - and expose the path in another environment variable.
 	if r.envFile != nil {
-		for key, value := range env {
-			r.envFile.WriteString(fmt.Sprintf("%s=%s\n", key, value))
-		}
+		encoded, _ := json.Marshal(d)
+		r.envFile.WriteString(string(encoded))
 		r.envFile.Close()
 		env["BUILDKITE_ENV_FILE"] = r.envFile.Name()
 	}


agent/job_runner.go
```diff
@@ -188,9 +189,8 @@ func (r *JobRunner) createEnvironment() []string {
 	// We present only the clean environment - i.e only variables configured
 	// on the job upstream - and expose the path in another environment variable.
 	if r.envFile != nil {
-		for key, value := range env {
-			r.envFile.WriteString(fmt.Sprintf("%s=%s\n", key, value))
-		}
+		encoded, _ := json.Marshal(d)
```
sander-m: Should we error here if Marshal fails?

sj26:
Hi @sander-m, thanks for the contribution! Could we wrap the values in quotes instead so they can contain newlines, but can still be sourced by shell scripts? Or use "\n" to escape? There are probably other characters to escape, maybe we should use something like [shellquote](https://github.com/kballard/go-shellquote)?

lox:
Is there any other env-file format for docker / docker-compose I wonder?

sander-m:
@sj26 Wrapping the values in quotes would be a good improvement. With the quotes added, I think we would be able to parse into key-value pairs with a regex. For our use case, we don't directly pass the file to Docker.

lox:
It's got to be a broadly useful feature for us though @sander-m. The idea of the env-file was that it could be loaded directly into docker. For your use-case I think we'd probably want to look at an agent subcommand to read env like was initially suggested.

sander-m:
Makes sense.  We would be happy with something like a `buildkite-agent env --json` command too.

sander-m:
Adding quotes to the values might be sufficient to make the current format `source`-able? That seems like an obvious improvement. In addition to that, did you mean an agent subcommand for json env output @lox ?

lox:
I've thought about it a bit. There is no clear standard anywhere for env files. I think in the face of that, going with something like JSON makes sense. At least then there is a clear path to converting to different dialects of env files from that format with `jq` or something like https://www.npmjs.com/package/envfile. 

I'm in favour of this change. Let's just double check if there is any prior-art anywhere for the json framing of the file.

lox:
We've discussed it, it came down to two options:

 * Env-file format with KEY="blah\nblah\nblah"
 * Env-file with JSON format

In the end, we decided on the former, because although it isn't supported by docker, it's easy to parse with shell primitives and is the format expected by most envfile parsers. 

I'm going to close this and follow up with a PR to this effect. Thoughts @sander-m?

sander-m:
Makes sense to me, thanks for the quick response :)

PR #682, opened 2018-03-14
Quote and escape env-file values

lox:
Follow on from #680 and #643. Changes the env-file format to KEY="VALUES" with values in the quotes escaped. 

This won't be compatible with docker (see the horror of https://github.com/moby/moby/issues/12997), but it's the generally accepted standard (https://github.com/bkeepers/dotenv/blob/master/lib/dotenv/parser.rb and https://github.com/joho/godotenv/blob/master/godotenv.go, for example) and it's easy to parse with standard shell tools.

### DIFF
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -74,10 +74,15 @@ func (r JobRunner) Create() (runner *JobRunner, err error) {
 		runner.envFile = file
 	}
 
+	env, err := r.createEnvironment()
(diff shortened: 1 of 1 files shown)


lox:
> I do also like the idea of a --json option for a json version, too.

@sj26 yeah, me too.

DazWorrall:
Thanks for the quick work here @lox ❤️

PR #685, opened 2018-03-14
Show plugin commit if it's already installed

### LINKED ISSUE #359: More plugin information would be nice
I just merged an update to the docker-compose plugin into master. Triggering a build which uses the plugin has a nice "Setting up plugins" section in the job log which lists the plugin in use, but it doesn't say which commit of the plugin is actually being used so I'm not sure if it's using an old, cached version or my new version.

It would be nice to output which commit of the plugin is being used in this log output for diagnosis purposes.

keithpitt:
Fixes #359

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -452,9 +452,9 @@ func (b *Bootstrap) checkoutPlugin(p *agent.Plugin) (*pluginCheckout, error) {
 		// let's figure that out.
 		headCommit, err := gitRevParseInWorkingDirectory(b.shell, directory, "--short=7", "HEAD")
 		if err != nil {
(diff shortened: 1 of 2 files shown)


bootstrap/bootstrap.go
```diff
@@ -448,7 +448,15 @@ func (b *Bootstrap) checkoutPlugin(p *agent.Plugin) (*pluginCheckout, error) {
 
 	// Has it already been checked out?
 	if fileExists(pluginGitDirectory) {
-		b.shell.Commentf("Plugin \"%s\" already checked out", p.Label())
+		// It'd be nice to show the current commit of the plugin, so
+		// let's figure that out.
+		headCommit, err := b.shell.RunAndCapture("git", "--git-dir", pluginGitDirectory, "--work-tree", directory, "rev-parse", "--short=7", "HEAD")
```
lox: I'd be tempted to create a func in git.go for this

PR #686, opened 2018-03-14
Handle windows paths in all usage of shellwords parsing

lox:
This centralises our use of shellword parsing (e.g parsing something like `C:\Windows\System32\CMD.exe /S /C "bash.exe ./upload.sh"` into it's respective tokens).

It turns out the Windows is ok with paths with forward slashes, even for UNC paths, so this just rolls with that.

It also sneakily rolls back #683 in favour of just converting the path to the binary to one that uses backslashes.


agent/job_runner.go
```diff
@@ -79,9 +80,18 @@ func (r JobRunner) Create() (runner *JobRunner, err error) {
 		return nil, err
 	}
 
+	args, err := shell.Parse(r.AgentConfiguration.BootstrapScript)
```
lox: This moved up here because I moved it out of process.Process so that the error can be handled better.

Issue #674, opened 2018-03-13
Setting `no-command-eval` should also disable plugins


keithpitt:
Since plugins can essentially do "command-eval", we should turn off plugins when `no-command-eval` is enabled.

PR #690, opened 2018-03-15
Enabling `no-command-eval` will also disable use of plugins

keithpitt:
Fixes https://github.com/buildkite/agent/issues/674

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -231,7 +231,7 @@ var AgentStartCommand = cli.Command{
 		},
 		cli.BoolFlag{
 			Name:   "no-command-eval",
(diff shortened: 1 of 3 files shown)


clicommand/agent_start.go
```diff
@@ -231,7 +231,7 @@ var AgentStartCommand = cli.Command{
 		},
 		cli.BoolFlag{
 			Name:   "no-command-eval",
-			Usage:  "Don't allow this agent to run arbitrary console commands",
+			Usage:  "Don't allow this agent to run arbitrary console commands. Turning off command eval will also disallow this agent to use plugins.",
```
lox: Might `"Don't allow this agent to run arbitrary console commands, including plugins"` be terser?

keithpitt:
@lox yah yah.

Issue #689, opened 2018-03-14
Plugins with a `null` config shouldn't blow up


keithpitt:
We should probably just treat them as "plugins with 0 config". i.e this fails:

```yaml
steps:
  - plugins:
      detect-clowns#v2.0.0: ~
```

PR #691, opened 2018-03-15
Support plugins that have a `null` config

keithpitt:
Fixes https://github.com/buildkite/agent/issues/689

### DIFF
--- a/agent/plugin.go
+++ b/agent/plugin.go
@@ -86,18 +86,30 @@ func CreatePluginsFromJSON(j string) ([]*Plugin, error) {
 			plugins = append(plugins, plugin)
 		case map[string]interface{}:
 			for location, config := range vv {
-				// Ensure the config is a hash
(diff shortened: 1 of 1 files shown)


agent/plugin.go
```diff
@@ -86,18 +86,30 @@ func CreatePluginsFromJSON(j string) ([]*Plugin, error) {
 			plugins = append(plugins, plugin)
 		case map[string]interface{}:
 			for location, config := range vv {
-				// Ensure the config is a hash
-				config, ok := config.(map[string]interface{})
-				if !ok {
-					return nil, fmt.Errorf("Configuration for \"%s\" is not a hash", location)
+				// First check to make sure this plugin has config
```
lox: Took me a while to parse this, might it be clearer with a `if config == nil` block up the top for handling the special case then a continue rather than `if/else`?

keithpitt:
@lox like this?

PR #692, opened 2018-03-15
Write batch scripts for multi-line commands with CMD.EXE

lox:
In #648 we allowed for the default shell to be configured and we stopped writing temporary scripts for commands to be executed with in favour of directly invoking them with the configured shell command. This broke multi-line commands for CMD.EXE, as it seems to ignore everything after the first newline.

The future is likely powershell for Windows users, but this writes batch files if a CMD.EXE shell is detected to maintain backwards compatibility.

This also introduces a new library I wrote for parsing and quoting strings with specific shell semantics: buildkite/shellwords. This is used to provide a more robust solution to #683.


lox:
This is ready for code-review now @keithpitt! ✈️

bootstrap/bootstrap.go
```diff
@@ -890,38 +893,98 @@ func (b *Bootstrap) defaultCommandPhase() error {
...
 	}
 
-	var cmd []string
+	if len(shell) == 0 {
+		return fmt.Errorf("No shell set for bootstrap")
+	}
 
-	if commandIsScript {
+	// Windows CMD.EXE is horrible and can't handle newline delimited commands. We write
+	// a batch script so that it works, but we don't like it
```
sj26: :disappointed:

bootstrap/bootstrap.go
```diff
@@ -890,38 +893,98 @@ func (b *Bootstrap) defaultCommandPhase() error {
...
+
+	// The shell gets parsed based on the operating system
+	shell, err := shellwords.Split(b.Shell)
 	if err != nil {
-		return err
+		return fmt.Errorf("Failed to split shell (%q) into tokens: %v", b.Shell, err)
 	}
 
-	var cmd []string
+	if len(shell) == 0 {
```
keithpitt: What sort of shell values can be passed to cause it to get a `len` of 0?

clicommand/bootstrap.go
```diff
@@ -222,6 +222,7 @@ var BootstrapCommand = cli.Command{
 			Name:   "shell",
 			Usage:  "The shell to use to interpret build commands",
```
keithpitt: Should we mention "This defaults to our best guess at what makes sense based on the current operation system" or something? Just thinking about what we copy/paste into the docs for this `--help` output.

lox: Yeah, makes sense.

bootstrap/bootstrap.go
lox: An empty string :)

clicommand/bootstrap.go
lox: Oh, this is the bootstrap cli command. Did you mean here or the agent start command?

keithpitt: > Did you mean here or the agent start command?

Oh yeah, I meant that!

lox: "The shell to use to interpret build commands. Defaults to bash or cmd.exe depending on your operating system."

keithpitt: 👍

lox: Or maybe "The shell commands used to interpret build commands, e.g /bin/bash -e -c"

keithpitt: Yeah that works as well

Issue #695, opened 2018-03-19
Boolean pipeline.yml values coerced to strings


ticky:
beta.41 coerces boolean values to strings, which causes several them to fail validation in the pipeline uploader.

For instance, we specifically look for a real `true` or `false` in the `required` field of `block` steps, producing this error on a previously-valid pipeline:

```
POST http://agent.buildkite.localhost/v3/jobs/bb078d93-7cd0-4269-8aaf-a90c46da6292/pipelines: 422 `required` must be a boolean (`true` or `false`) (Attempt 1/5 Retrying in 1s)
```

Debugging revealed this was passed to this point as a string.

This issue was introduced in #668

PR #696, opened 2018-03-19
Preserve types in pipelines

lox:
In #668 we made some changes to the underlying YAML library which have caused a regression where bools, ints, floats and null get juggled to strings before being uploaded as JSON.

This passes through those types and panics if an unhandled type is encountered.

Closes https://github.com/buildkite/agent/issues/695.

### DIFF
--- a/agent/pipeline_parser.go
+++ b/agent/pipeline_parser.go
@@ -65,7 +65,7 @@ func (p PipelineParser) Parse() (interface{}, error) {
 
 	var result interface{}
 	if err := unmarshalAsStringMap(b, &result); err != nil {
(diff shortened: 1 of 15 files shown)


agent/pipeline_parser.go
```diff
@@ -293,9 +293,9 @@ func cleanupMapValue(v interface{}) interface{} {
 		return cleanupInterfaceArray(v)
 	case map[interface{}]interface{}:
 		return cleanupInterfaceMap(v)
-	case string:
+	case nil, bool, string, int, float64:
```
lox: This is the change in question.

PR #698, opened 2018-03-21
Expose no-git-submodules config and arg to start

lox:
The v2 bootstrap used to support `BUILDKITE_DISABLE_GIT_SUBMODULES` (via #168). 

This exposes that option (also `BUILDKITE_NO_GIT_SUBMODULES`).

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -31,6 +31,12 @@ Example:
 
    $ buildkite-agent start --token xxx`
 
+// Adding config requires changes in a few different spots
+// - The AgentStartConfig struct with a cli parameter
+// - As a flag in the AgentStartCommand (with matching env)
+// - Into an env to be passed to the bootstrap in agent/job_runner.go, createEnvironment()
+// - Into clicommand/bootstrap.go to read it from the env into the bootstrap config
+
 type AgentStartConfig struct {
 	Config                    string   `cli:"config"`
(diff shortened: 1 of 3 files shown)


clicommand/agent_start.go
```diff
@@ -240,6 +247,11 @@ var AgentStartCommand = cli.Command{
 			Usage:  "Don't allow this agent to load plugins",
 			EnvVar: "BUILDKITE_NO_PLUGINS",
 		},
+		cli.BoolFlag{
+			Name:   "no-git-submodules",
+			Usage:  "Don't automatically checkout git submodules",
+			EnvVar: "BUILDKITE_NO_GIT_SUBMODULES,BUILDKITE_DISABLE_GIT_SUBMODULES",
```
keithpitt: Wouldn't `BUILDKITE_DISABLE_GIT_SUBMODULES` flow through to the bootstrap process anyway?

lox: It would, but nothing there would read it, the bootstrap is looking for `BUILDKITE_GIT_SUBMODULES`. You make a good point though in that previously that setting was set in hooks and would be expected to be read in the checkout hook.

lox: I guess this just exposes a config setting for disabling git submodules, but it's probably a breaking change from the previous behaviour.

PR #699, opened 2018-03-21
Windows install script

lox:
This adds a powershell install script so that we can have users do windows installs with:

```powershell
$env:buildkiteAgentToken = "my_token"
$env:buildkiteAgentBeta = 'true'
Set-ExecutionPolicy Bypass -Scope Process -Force
iex ((New-Object System.Net.WebClient).DownloadString('https://raw.githubusercontent.com/buildkite/agent/master/install.ps1'))
```

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -1,10 +1,10 @@
 Param(
     [string]$installDir = "C:\buildkite",
     [string]$arch = "amd64",
-    [switch]$beta = $false,
+    [switch]$beta = ($env:buildkiteAgentBeta -eq 'true'),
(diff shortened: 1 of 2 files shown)


keithpitt:
Nice! Would we still bundle the service installer with the downloadable release?

lox:
Nah I don't think so, it's 3 lines and it's using a custom service manager. We can document that bit, this just makes the install really easy.

lox:
Needs some more niceties from https://chocolatey.org/install.ps1.

lox:
But it's a start!

PR #700, opened 2018-03-21
Add a configurable cancel-grace-period

lox:
This makes the time between `SIGTERM` and `SIGKILL` when a job is cancelled configurable. 

This allows people who have workloads with things like `packer` and `docker-compose` (which can often take 20-30 seconds to gracefully shutdown) to configure when they are forcefully killed. The alternative is often having to write cleanup scripts outside of the agent.

Ideally, this would be a setting which could be configured at a step/yaml level to and passed down to the job runner vs the blunt-force agent-wide config.

### DIFF
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -250,11 +250,11 @@ func (r *JobRunner) Kill() error {
 	defer r.killLock.Unlock()
 
 	if !r.cancelled {
-		logger.Info("Canceling job %s", r.Job.ID)
+		logger.Info("Canceling job %s with a grace period of %ds", r.Job.ID, r.AgentConfiguration.CancelGracePeriod)
 		r.cancelled = true
 
 		if r.process != nil {
-			r.process.Kill()
+			r.process.Kill(time.Second * time.Duration(r.AgentConfiguration.CancelGracePeriod))
 		} else {
 			logger.Error("No process to kill")
 		}
--- a/agent/agent_configuration.go
+++ b/agent/agent_configuration.go
@@ -18,5 +18,6 @@ type AgentConfiguration struct {
 	TimestampLines            bool
 	DisconnectAfterJob        bool
 	DisconnectAfterJobTimeout int
+	CancelGracePeriod         int
 	Shell                     string
 }
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -46,6 +46,7 @@ type AgentStartConfig struct {
 	DisconnectAfterJob        bool     `cli:"disconnect-after-job"`
 	DisconnectAfterJobTimeout int      `cli:"disconnect-after-job-timeout"`
 	BootstrapScript           string   `cli:"bootstrap-script" normalize:"commandpath"`
+	CancelGracePeriod         int      `cli:"cancel-grace-period"`
 	BuildPath                 string   `cli:"build-path" normalize:"filepath" validate:"required"`
 	HooksPath                 string   `cli:"hooks-path" normalize:"filepath"`
 	PluginsPath               string   `cli:"plugins-path" normalize:"filepath"`
@@ -162,6 +163,12 @@ var AgentStartCommand = cli.Command{
 			Value:  120,
 			Usage:  "When --disconnect-after-job is specified, the number of seconds to wait for a job before shutting down",
 			EnvVar: "BUILDKITE_AGENT_DISCONNECT_AFTER_JOB_TIMEOUT",
+		},
+		cli.IntFlag{
+			Name:   "cancel-grace-period",
+			Value:  10,
+			Usage:  "The number of seconds running processes are given to gracefully terminate before they are killed when a job is cancelled",
+			EnvVar: "BUILDKITE_CANCEL_GRACE_PERIOD",
 		},
 		cli.StringFlag{
 			Name:   "shell",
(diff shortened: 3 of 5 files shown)


ticky:
I suppose technically, because this is passed through the job runner, it could be overridden (or supplied) via the job runner pretty easily if we wanted.

agent/job_runner.go
```diff
@@ -173,11 +173,11 @@ func (r *JobRunner) Kill() error {
 	defer r.killLock.Unlock()
 
 	if !r.cancelled {
-		logger.Info("Canceling job %s", r.Job.ID)
+		logger.Info("Canceling job %s with a grace period of %ds", r.Job.ID, r.AgentConfiguration.CancelGracePeriod)
 		r.cancelled = true
 
 		if r.process != nil {
-			r.process.Kill()
+			r.process.Kill(time.Second * time.Duration(r.AgentConfiguration.CancelGracePeriod))
```
ticky: Does this need to be Seconds * Duration, or could it just be Second * int, like it was previously?

lox: time.Second is of type `time.Duration`, so the `time.Duration(...)` is just a cast to match the types.

keithpitt:
I'm afraid this is only half the change necessary, and the other half could get quite tricky...

We've got some code in the Buildkite backend that will forcefully remove a job from an agent if it's been in a "cancelling" state for more than 30 seconds. The backend assumes that job processes get killed after 10 seconds, so BK assumes that if a "cancelling" has been going on for 30 seconds or more, something is wrong is it tries to repair itself. It removes the agent from the job, and puts the agent back into the pool of agents that can perform work.

The additional 20 seconds are for fluctuations in our state checking background worker...

So I think we may need to bump this change until 3.1 and have a look at the whole thing again!

lox:
Ah that's a bummer. Agreed on pushing this one out to a later version.

ticky:
Hmm, makes sense!

eredi93:
@lox @keithpitt @ticky are you planning to get this out any time soon?
the lack of no control over the cancel graceful period is causing issues with our terraform pipeline where if someone cancels the build terraform won't release the lock.
should we wait for this to go out or get some other workaround together?

lox:
@eredi93 and @DazWorrall, myself and @keithpitt are investigating and we'll update soon.

lox:
@keithpitt has figured out the backend complications, which now means that the agent is in charge of cancellation.

keithpitt:
@lox lemme know when this is in a review state and I'll give it some fresh eyeballs!

lox:
Ok, good to go @keithpitt

keithpitt:
@DazWorrall I don't think that'd be too hard! (talking on behalf of @lox - it may be tricky, but I don't think so). Is it something you'd enable on a per-job basis? Or a global agent setting?

keithpitt:
Oh yeah, looking at the code I think that'd be easy to implement.

One thing I noticed with this as well is that the settings are for `cancel`, but the logic is for both "Cancel" and "Timing Out". I wonder if they should be different settings? Different things? Or maybe we can come up with a word that works for both? Or maybe just `cancel-*` is fine?

eredi93:
I endorse @DazWorrall proposal! i'd be really useful, but i wonder if we should address that in a separate PR. I'd love to get a configurable cancel-grace-period out first then then think about adding the `--cancel-signal`

lox:
Let's get this merged first and then I want to spend some time cleaning up (and consolidating) how we handle various forms of cancellation, agent shutdown and job timeout. Happy to add some custom signal settings if y'all still end up needing them. 

The other thing we will probably do is send some more information on the circumstances of the cancel or failures to buildkite in the `FinishJob` api call.

lox:
Actually, now that I think about it @DazWorrall, it sounds like what you might want is a way to signal the agent with `SIGUSR1` to shutdown the agent within 30 seconds or so? E.g SIGTERM, then SIGKILL after 30 seconds for the bootstrap? The cancel signal would be the signal that the Agent process sends its bootstrap sub-processes.

DazWorrall:
> it sounds like what you might want is a way to signal the agent with SIGUSR1 to shutdown the agent within 30 seconds or so? E.g SIGTERM, then SIGKILL after 30 seconds for the bootstrap?

For our particular environment, this is less useful for us as we're unable to send arbitrary signals through the kubernetes control plane - only a SIGTERM (pod deletion). To give some context, let me share some permutations of agent termination we need to handle. Off the top of my head:

* Node preemption. We have 30 seconds to play with here, and currently not a lot of control over what happens in that time, though we're working on it. The Buildkite agent could be signalled by a process supervisor, but also our own bootstrap could be signalled directly. We're working to get better control of this scenario ourselves, and aren't yet sure if we need anything extra from the agent on this - my suspicion is not. Once we have control here, our priority will be uploading artifacts - we can skip any cleanup because the node is about to be destroyed.
* Job cancelled by Buildkite (Buildkite agent signals our bootstrap, kills after a hardcoded (for now) 10 seconds). Once we can configure the timeout, we would definitely handle this scenario more gracefully - for example we will know the node is going to live on, and so we need to tidy up any job cruft. However currently we don't know that we're dealing with a cancellation if we get a `TERM`, for all we know it's because of preemption. A configurable cancel signal here would be useful, but only once the timeout is configurable.
* The current agent `TERM` handling - agent waits for any running job to complete, then exits - is still very important for us. We rely on it for a graceful scale down of our infrastructure, and rolling out updates to our bootstrap code.

PR #701, opened 2018-03-23
Build docker image as part of agent pipeline

lox:
Currently docker images are build via an async trigger to a pipeline that builds from https://github.com/buildkite/docker-buildkite-agent.

That repository builds images for all our versions (beta, stable and edge), regardless of what was actually built. 

This moves the docker packaging and building into the agent pipeline and only builds images for the current agent version. The following tags are created:

* For beta releases: 3.0-beta.42.2243 is tagged as beta, 3.0-beta.42, 3.0 and 3
* For edge releases: 3.0-beta.42.2243 is tagged as edge and edge-build-2243
* For stable release: 2.6.10 is tagged as stable, latest, 2.6.10, 2.6 and 2

Once this is in place, we can probably entirely remove the old https://github.com/buildkite/docker-buildkite-agent repo.

### DIFF
--- a/scripts/publish-docker-images.sh
+++ b/scripts/publish-docker-images.sh
@@ -52,8 +52,8 @@ if [[ "$DOCKER_PULL" =~ (true|1) ]] ; then
   docker pull "$PREBUILT_IMAGE"
 fi
 
-# variants of edge/unstable
-if [[ "$CODENAME" == "unstable" ]] ; then
+# variants of edge/experimental
+if [[ "$CODENAME" == "experimental" ]] ; then
   release_image "edge-build-${BUILDKITE_BUILD_NUMBER}"
   release_image "edge"
 fi
@@ -65,8 +65,8 @@ if [[ "$CODENAME" == "stable" ]] ; then
   done
 fi
 
-# variants of beta/experimental - e.g 3.0-beta.16
-if [[ "$CODENAME" == "experimental" ]] ; then
+# variants of beta/unstable - e.g 3.0-beta.16
+if [[ "$CODENAME" == "unstable" ]] ; then
   for tag in beta $(parse_version "$AGENT_VERSION") ; do
     if is_stable_version "$tag" ; then
       echo "--- :docker: Skipping tagging stable ${DOCKER_IMAGE}:${tag}"
--- a/scripts/publish-docker-images.sh
+++ b/scripts/publish-docker-images.sh
@@ -1,0 +1,77 @@
+#!/bin/bash
+set -euo pipefail
+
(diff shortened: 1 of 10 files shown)


lox:
@toolmantim I'll leave this one for a while in case you get a chance to eye ball it and provide feels. We can probably ship 3.0.0 without this change.

scripts/publish-docker-images.sh
```diff
@@ -0,0 +1,77 @@
...
+# variants of beta/experimental - e.g 3.0-beta.16
+if [[ "$CODENAME" == "experimental" ]] ; then
+  for tag in beta $(parse_version "$AGENT_VERSION") ; do
+    if is_stable_version "$tag" ; then
+      echo "--- :docker: Skipping tagging stable ${DOCKER_IMAGE}:${tag}"
+      continue
+    fi
+    release_image "$tag"
+  done
+fi
```
sj26: I thought beta == unstable, and edge == experimental?

sj26:
This is great, I wanted to do this while looking at agent packaging recently. 💯

toolmantim:
Don’t wait for me!

lox:
Such merge anxiety. 😱

scripts/publish-docker-images.sh
lox: Shoot, you are right. Good catch!

toolmantim:
What’s on the merge anxiety top list? Perhaps there’s some tests we can do to help.

lox:
Nah, more just I was nervous about creating more technical debt. This is the right way to go though, let's ship it.

scripts/publish-docker-images.sh
```diff
@@ -0,0 +1,77 @@
...
+
+if [[ "$DOCKER_PULL" =~ (true|1) ]] ; then
+  echo '--- :docker: Pulling prebuilt image'
+  docker pull "$PREBUILT_IMAGE"
+fi
+
+# variants of edge/experimental
+if [[ "$CODENAME" == "experimental" ]] ; then
+  release_image "edge-build-${BUILDKITE_BUILD_NUMBER}"
+  release_image "edge"
```
sj26: I don't think we want to overwrite the "edge" tag with a new build on the 2-6-stable branch. Can we limit this somehow?

sj26: Probably something like the is_stable_version protection below?

lox: But this code won't ever be on that branch, will it?

lox: But, I will anyway.

sj26: @lox good point!

PR #707, opened 2018-03-29
Add an agent config for no-local-hooks

lox:
This adds a `no-local-hooks` config for the agent. Global hooks can still set `BUILDKITE_NO_LOCAL_HOOKS` on a case-by-case basis.

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -60,6 +60,7 @@ type AgentStartConfig struct {
 	NoColor                   bool     `cli:"no-color"`
 	NoSSHKeyscan              bool     `cli:"no-ssh-keyscan"`
 	NoCommandEval             bool     `cli:"no-command-eval"`
+	NoLocalHooks              bool     `cli:"no-local-hooks"`
 	NoPlugins                 bool     `cli:"no-plugins"`
 	NoPTY                     bool     `cli:"no-pty"`
 	TimestampLines            bool     `cli:"timestamp-lines"`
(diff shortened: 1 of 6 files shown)


clicommand/agent_start.go
```diff
@@ -315,8 +321,8 @@ var AgentStartCommand = cli.Command{
 			cfg.BootstrapScript = fmt.Sprintf("%s bootstrap", shellwords.Quote(os.Args[0]))
 		}
 
-		// Turning off command eval will also turn off plugins.
-		if cfg.NoCommandEval {
+		// Turning off command eval or local hooks will also turn off plugins.
+		if cfg.NoCommandEval || cfg.NoLocalHooks {
```
keithpitt: Hrm...what's your thinking here?

lox: What's the point of disabling local hooks if you can use plugins? You'd just be able to modify pipeline.yml and achieve the same thing as local hooks otherwise.

keithpitt: Local hooks by themselves aren't dangerous because you need source code access to change them. Editing a local hook is the same as editing a build script.

The point behind `no-command-eval` is stopping Buildkite executing arbitrary commands on an agent. Since you can define plugins in the BK user interface, you could circumvent `no-command-eval` (which is why we turn off plugins if `no-command-eval` is on).

So I don't think we need to disable plugins if `no-local-hooks` is on? Unless I'm missing another attack vector?

lox: My understanding was that the point behind `no-local-hooks` is to prevent hooks in a repository being able to override the hooks set at the global level. They are for where you don't trust the source code, for instance third-party pull requests.

lox: Did you have a different understanding of their purpose?

keithpitt: Ok, spoke about on video with @lox. If you consider plugins to be "3rd party local hooks" then this makes total sense...!

PR #708, opened 2018-03-29
Prettier bootstrap output 💅🏻

lox:
Lots of cleanup on messages for debug and non-debug mode.


keithpitt:
@lox should we merge and put 1 final beta out?

PR #720, opened 2018-04-04
Added linux,ppc64le as valid build target.

derekmarcotte:
Cannot build on linux ppc64le, and wish to.  Seems to work fine with this.


lox:
Thanks, seems to make sense! What systems are `ppc64le`?

derekmarcotte:
Hrm, good question!  IBM p-series?  Mainly, it's the [native CI environment](https://github.com/derekmarcotte/kitefactory) I'm building for use by the [node_exporter](https://github.com/prometheus/node_exporter) project.  I'm not sure who is running node_exporter on Linux Power, and for what purposes though.

lox:
Kitefactory looks super cool!

lox:
Drop us an email at hello@buildkite.com if we can help @derekmarcotte!

PR #722, opened 2018-04-09
Upload github release and homebrew in beta/stable pipelines

lox:
Previously we were doing the github release and homebrew release only in the beta pipeline, which meant that when we released 3.0.0 it shipped even before the stable release pipeline was unlocked, which was very unexpected (for me).

This looks at the pipeline and skips the github and homebrew step depending on whether the version is a pre-release or not.

### DIFF
--- a/.buildkite/pipeline.release-stable.yml
+++ b/.buildkite/pipeline.release-stable.yml
@@ -13,6 +13,13 @@ steps:
     agents:
       queue: "deploy"
 
+  - name: ":octocat: :rocket:"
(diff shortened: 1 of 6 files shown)


keithpitt:
Wait, does this mean that there aren't GitHub releases for beta releases anymore?

keithpitt:
Ah right, I sees now (the PR description just confused me a little)

lox:
Yeah, sorry that wasn't super clear @keithpitt!

PR #494, opened 2017-06-22
Set the content encoding on artifacts uploaded to s3

airhorns:
This sucker isn't fully tested yet, wanted to get buy in on the approach before doing the leg work :)

We (Shopify) upload a whackload of artifacts to s3 using the buildkite agent, and have started to need to compress some of the bigger ones. We're working on a bunch of coverage stuff and other test time instrumentation who's artifacts can get really, really big at the highest fidelity, so we gzip them after we produce them and before they get uploaded. This is all well and good and Buildkite doesn't have to care, but, we also want to start doing useful stuff with them after they're uploaded! 

S3 serves them, and it gets the mime type right (or I guess as right as you can be for a `csv.gz` file), but it doesn't set a `Content-Encoding` header when it serves the file. Technically, it's fine to serve a gzipped stream without specifying that that is how the stream is compressed, but, it breaks consumers that aren't smart enough to know the thing is gzipped. We have one such consumer which is a codebase we don't really control and don't really want to couple to our file naming format (Neo4j). It uses the `Content-Encoding` to know how do decode the incoming stream and if it should decompress, and since S3 doesn't set the encoding unless you tell it to, it doesn't do any decoding, and we end up with compressed gobbledegook in the database.

So, I'd love it if the agent did the same kind of profiling that it does for the mime type for the content encoding, and set that in the s3 options for the put. That way, S3 would serve the blob with the right content encoding, and any downstream consumers (like neo4j or the browser for that matter) would consume knowing that it had to decompress.

What do y'all think of the idea? And if you like it, where should I put the code that does this profiling? I did a simple version to illustrate but I'm not the best go programmer so feel free to tell me how it can be better and how to test.

Thanks!

PR #732, opened 2018-04-17
Don't set Content-Encoding on s3 upload

lox:
This reverts #494, is it creates a confusing behaviour where when you uploading a gzipped artifact and then download it in a later step the file gets un-gzipped.


keithpitt:
@lox do we need to set `https://github.com/buildkite/agent/blob/master/agent/version.go#L9` to 3.0.1 first?

lox:
I just did that on `master` @keithpitt

keithpitt:
👍

PR #733, opened 2018-04-17
Support `--no-interpolation` option in `pipeline upload`

### LINKED ISSUE #715: Pipeline parsing attempts to interpolate invalid environment variables
Trying to upload a pipeline with `buildkite-agent pipeline upload` containing this snippet:

```
- trigger: some-pipeline
  build:
    message: Don't create remote checkouts for $0 checkouts
    commit: 1eda429965fd220878dd60d14cf1313988dcb319
    branch: master
```

Fails with:

```
DEBUG  Debug mode enabled
INFO   Reading pipeline config from "/tmp/buildkite-pipeline-20180403-1053-88yil1"
DEBUG  Parsing pipeline configuration as YAML
FATAL  Pipeline parsing of "buildkite-pipeline-20180403-1053-88yil1" failed (Expected identifier to start with a letter, got 0)
```

It's trying to interpolate the `$0` I guess. As a digit it not a valid variable name I feel like this shouldn't break the parser - it really shouldn't attempt to parse it - wdyt @lox?

lox:
This allows for interpolation to be disabled in `buildkite-agent pipeline upload`.

Closes #715.

### DIFF
--- a/agent/pipeline_parser.go
+++ b/agent/pipeline_parser.go
@@ -15,9 +15,10 @@ import (
 )
 
 type PipelineParser struct {
-	Env      *env.Environment
-	Filename string
-	Pipeline []byte
+	Env             *env.Environment
+	Filename        string
+	Pipeline        []byte
+	NoInterpolation bool
 }
 
 func (p PipelineParser) Parse() (interface{}, error) {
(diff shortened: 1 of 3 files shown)


agent/pipeline_parser.go
```diff
@@ -32,6 +33,15 @@ func (p PipelineParser) Parse() (interface{}, error) {
 		errPrefix = fmt.Sprintf("Failed to parse %s", p.Filename)
 	}
 
+	// If interpolation is disabled, just parse and return
+	if p.NoInterpolation {
+		var result interface{}
```
keithpitt: I think we may need to do the same logic as we do on line 50 (with the switch between a slice or a map?)

lox: We only do that so that we can `parseWithEnv` if it's a map. Without interpolation I don't think we need to do that, do we?

lox: Hmmmmm. We might end up with out-of-order env if we do that though.

lox: But we always had that, it only mattered once we started doing interpolation.

lox:
Let's get this one merged. I'm pretty sure that short-circuit is ok @keithpitt, as without any interpolation we don't need any of the other fancy parsing bits.

PR #737, opened 2018-04-19
Add a virtual provides to the RPM package

jnewbigin:
Allow package to be matched as `buildkite-agent-<major version>` ie
`yum install buildkite-agent-2`


lox:
Totally missed this, nice one, many thanks @jnewbigin!

lox:
The only problem is this will only apply to new releases!

jnewbigin:
Yep. It is in the packaging so unless you wanted to re-release an old version, only new build will have the new tag.

lox:
Yeah, I'm kind of in the same boat @keithpitt 😅

My only other question was whether provides for `buildkite-agent` is implicit @jnewbigin, any idea?

lox:
Let's merge this in an experiment a bit with it.

PR #740, opened 2018-04-23
Try to help people that forget to quote their upload globs

toolmantim:
I did this in a script, to upload some JUnit xml files:

```bash
buildkite-agent artifact upload *.xml
```

and it failed with this error message:

image

This had me very confused for a few hours, until I realised that there was a second argument to upload called "upload destination" — I had assumed that the error message on upload destination was that I had misconfigured my agent, or it was a backend problem.

What I needed to do was:

```bash
buildkite-agent artifact upload "*.xml"
```

This PR tries to provide a less confusing error message if people accidentally provide a second argument to the `artifact upload` command.


toolmantim:
Yay, thanks @lox! 🙌🏼

PR #744, opened 2018-04-27
Make BUILDKITE_AGENT the canonical prefix for env

lox:
Currently we have a mix of `BUILDKITE_` and `BUILDKITE_AGENT_` prefixes for env. This leads to confusion. 

This adds a canonical version with `BUILDKITE_AGENT_`, but still supports the older format. Perhaps once v4 lands, we can update our docs to always refer to the `BUILDKITE_AGENT_` version for much consistency.


lox:
Hmmm, this doesn't handles the env that the bootstrap searches for to update it's config: https://github.com/buildkite/agent/blob/e417b32ea188f37fba7e331e0856e1d2b6841f08/bootstrap/config.go#L64-L68

lox:
Yeah, honestly at this stage I'm not sure it's worth changing. Maybe we just make peace with inconsistency.

lox:
I'm closing this for now, I think it creates more complexity than it addresses.

PR #748, opened 2018-04-29
Add basic plugin definition parsing

lox:
This parses `plugin.yml` files like https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/blob/5e764199256d66e9064eae16d62e7a0a6c26e475/plugin.yml as part of the plugin checkout process in the bootstrap.

This will include: 

* [x] A non-fatal warning if a plugin definition doesn't exist
* [x] Check `requirements` (required command-line tools)
* [x] Check `configuration` json schema against provided plugin configuration from pipeline
* [x] Allow plugin validation to be disabled (`--no-plugin-validation`)
* [x] Allow `BUILDKITE_PLUGIN_VALIDATION` from pipeline env for per-pipeline rollout 

Failure looks like:

image

Suggestions welcomed on how to present errors, getting super user friendly JSON schema errors is hard.

### DIFF
--- a/agent/plugin/definition.go
+++ b/agent/plugin/definition.go
@@ -1,27 +1,54 @@
 package plugin
 
 import (
+	"encoding/json"
 	"errors"
+	"fmt"
 	"io/ioutil"
 	"os"
+	"os/exec"
 	"path/filepath"
 
-	"github.com/buildkite/yaml"
+	"github.com/buildkite/agent/yamltojson"
+	"github.com/qri-io/jsonschema"
 )
 
 var ErrDefinitionNotFound = errors.New("Definition file not found")
 
 // Definition defines the plugin.yml file that each plugin has
 type Definition struct {
-	Name         string   `yaml:"name"`
-	Requirements []string `yaml:"requirements"`
+	Name          string                 `json:"name"`
+	Requirements  []string               `json:"requirements"`
+	Configuration *jsonschema.RootSchema `json:"configuration"`
 }
 
 // ParseDefinition parses either yaml or json bytes into a Definition
 func ParseDefinition(b []byte) (*Definition, error) {
-	var def Definition
+	var parsed interface{}
 
-	if err := yaml.Unmarshal(b, &def); err != nil {
+	// Parse the definition as a json compatible string map, the plain yaml
+	// Unmarshal returns a structure that has map[interface{}]interface{}
+	// which causes anything that expects map[string]interface{} to break
+	if err := yamltojson.UnmarshalAsStringMap(b, &parsed); err != nil {
+		return nil, err
+	}
+
+	// Check we've got a map, vs a list or something else, this lets us
+	// give a useful error message
+	parsedMap, ok := parsed.(map[string]interface{})
+	if !ok {
+		return nil, fmt.Errorf("Failed to convert %T to map[string]interface{}", parsed)
+	}
+
+	// Marshal the whole lot back into json which will let the jsonschema library
+	// parse the schema into and object tree 💃🏼
+	jsonBytes, err := json.Marshal(parsedMap)
+	if err != nil {
+		return nil, err
+	}
+
+	var def Definition
+	if err := json.Unmarshal(jsonBytes, &def); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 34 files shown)


lox:
> One thing I realised for the configuration validation, is that plugins can be configured through the JSON, as well as environment variables.

😱

lox:
So @toolmantim, I reckon that plugin configuration via env is a bad idea. Beyond validation being hard I think it also opens up weirdness around how to handle things like users setting the non-array version, and then the plugin setting the array version of a variable (say `BUILDKITE_PLUGIN_DOCKER_COMPOSE_VOLUMES=/myvol` and then the plugin loads `BUILDKITE_PLUGIN_DOCKER_COMPOSE_VOLUMES_0`). 

I reckon we should think about an alternative and then explicitly not allow plugins to be configured via env. @keithpitt feels?

toolmantim:
Hrmm interesting. That weirdness around arrays sounds like a small thing, but maybe I’m missing context?

Any ideas for alternative run-time setting methods, in hooks that are executed before the plugin has been downloaded?

toolmantim:
Thinking through use cases, if that helps:

1. Setting stuff that you don’t want Buildkite to know about (secrets)
2. Setting defaults at an agent level (like in buildkite-plugins/docker-compose-buildkite-plugin#157)

If we did want to, how would we disallow plugins being configured via envs?

🤔

toolmantim:
For use case 1 (secrets). We really don’t want a documented property in plugin.yml, as it still encourages people to paste in the value into their pipeline.yml file. So currently, the pattern is to just have it undocumented, but follow the same plugin env key prefix.

If we had first class “secrets” maybe you’d have a “secretProperties” section of plugin.yml, and our auto-gen’d docs would show those in a separate table? I dunno.

lox:
@toolmantim yeah, perhaps I'm taking too hard a line on it, I just suspect that it's a bigger ask to have end-users grasp the env formatting of plugin config than it is plugin authors, and it opens up a world of pain with validating that users have provided valid plugin configuration.

In terms of what we'd offer instead, I'd suggest a convention like we established in https://github.com/buildkite-plugins/docker-login-buildkite-plugin/pull/20. Beyond that, I think we need a first class mechanism for secrets.

lox:
> Somewhat related, is there a way for people to opt into experiments on a per pipeline basis, without having to potentially break all their pipelines and/or migrate them in one big hit?

You could set `BUILDKITE_PIPELINE_VALIDATION=true` per-pipeline and it should work I believe. I will test.

bootstrap/bootstrap.go
```diff
@@ -431,11 +432,42 @@ func (b *Bootstrap) PluginPhase() error {
 		checkout, err := b.checkoutPlugin(p)
 		if err != nil {
 			return errors.Wrapf(err, "Failed to checkout plugin %s", p.Name())
-
+		}
+		if b.Config.PluginValidation && checkout.Definition == nil {
```
keithpitt: How would `checkout.Definition` (not) be nil here? Wouldn't it always be nil?

bootstrap/bootstrap.go
```diff
@@ -431,11 +432,42 @@ func (b *Bootstrap) PluginPhase() error {
...
+			if b.Debug {
+				b.shell.Commentf("Parsing plugin definition for %q from %s", p.Label(), checkout.CheckoutDir)
+			}
+			// parse the plugin definition from the plugin checkout dir
+			checkout.Definition, err = plugin.LoadDefinitionFromDir(checkout.CheckoutDir)
+			if err == plugin.ErrDefinitionNotFound {
+				b.shell.Warningf("Failed to find plugin definition")
```
keithpitt: Should this include the name of the plugin? Or is the name of the plugin already shown?

keithpitt: (Based on the error here: https://github.com/buildkite/agent/pull/748/files#diff-1d1158568615ba7f7c717c11183449f8L433)

bootstrap/bootstrap.go
```diff
@@ -431,11 +432,42 @@ func (b *Bootstrap) PluginPhase() error {
...
+				return err
+			}
 		}
 		b.plugins = append(b.plugins, checkout)
 	}
 
+	if b.Config.PluginValidation {
+		for _, checkout := range b.plugins {
+			if b.Debug {
+				json, _ := json.Marshal(checkout.Definition)
```
keithpitt: Will there be a definition if one couldn't be loaded? i.e. will `checkout.Definition` be nil and this code blow up?

clicommand/agent_start.go
```diff
@@ -327,6 +335,13 @@ var AgentStartCommand = cli.Command{
 			cfg.NoPlugins = true
 		}
 
+		// Initially, plugin validation must be enabled via an experiment
```
keithpitt: Does it need to be an experiment? Can it just be a config option with a `false` default value?

agent/plugin/definition.go
```diff
@@ -0,0 +1,154 @@
...
+
+	return result
+}
+
+type ValidateResult struct {
+	Valid  bool
+	Errors []string
+}
+
+func (vr ValidateResult) Error() string {
```
keithpitt: Is it worth making an `Valid()` method that checks for `len(Errors)` instead of managing a boolean? Does that make the code better/worse? I don't mind either way.

agent/plugin/definition.go
```diff
@@ -0,0 +1,154 @@
...
+	for _, filename := range possibleFilenames {
+		if _, err := os.Stat(filename); err == nil {
+			return filename, nil
+		}
+	}
+	return "", ErrDefinitionNotFound
+}
+
+type Validator struct {
+	CommandExists func(string) bool
```
keithpitt: Is this configurable to make testing easier?

agent/plugin/definition.go
```diff
@@ -0,0 +1,154 @@
...
+
+	var commandExistsFunc = v.CommandExists
+	if commandExistsFunc == nil {
+		commandExistsFunc = commandExists
+	}
+
+	// validate that the required commands exist
+	if def.Requirements != nil {
+		for _, command := range def.Requirements {
+			if !commandExistsFunc(command) {
```
keithpitt: Should this also check that it's readable / executable from the current user? Or is path checking enough?

bootstrap/bootstrap.go
lox: Fixed!

bootstrap/bootstrap.go
lox: Fixed!

bootstrap/bootstrap.go
lox: Yup, it will blow up. Good catch, added a check.

agent/plugin/definition.go
lox: Yeah, done.

agent/plugin/definition.go
lox: Yup, for testing purposes.

lox:
Ok, I've rolled back the experiment. It's disabled by default, but you can set `--no-plugin-validation=false` or set `BUILDKITE_PLUGIN_VALIDATION=true` at a pipeline or step level.

lox:
I'd like to ship this if you are happy with it @keithpitt.

agent/plugin/definition.go
lox: [LookPath](https://golang.org/pkg/os/exec/#LookPath) checks that the binary is in the path, readable and executable.

PR #749, opened 2018-04-29
Add ubuntu docker image

lox:
We still refer to an ubuntu docker image in https://buildkite.com/docs/agent/v3/docker. I think it makes sense to offer an ubuntu variant as well as alpine.

This will have a `-ubuntu` suffix, so `buildkite/agent:3.0-ubuntu` and will use the latest LTS.


toolmantim:
We deprecated the Ubuntu image quite a while ago… we just forgot to update the docs. I wonder if it's a good idea to bring it back? Then we'd add jessie too I guess? We initially deprecated it to reduce the support footprint.

We could instead have some instructions for building your own Ubuntu + Jessie images?

lox:
It seems like it's pretty standard to have a default alpine build but also an ubuntu or debian build for things @toolmantim. I'm ok with adding it to the build process, enough people have asked and the size of the ubuntu image has come right down over time. Feels?

toolmantim:
Cool cool, sounds worth it?

The only other thing maybe to consider is keeping our OS base up-to-date with security vulns too? Alpine base tends to have much less vulns, and at the moment we only update our images on agent release at the moment. 

https://hub.docker.com/r/buildkite/agent/tags/

toolmantim:
How do other people handle that I wonder? With Docker Hub auto-build you can have it trigger based on other dependent images. We could use a daily scheduled build? So package updates would be picked up max 24h after the OS packages are updated.

lox:
Yeah, it's a tricky problem. We basically need to monitor upstream images somehow and trigger agent builds, but we also need a way to notify us so we can cut new releases.

PR #751, opened 2018-05-02
Set TERM and PWD for commands that get executed in shell

lox:
The move to a golang bootstrap in 3.0 has meant that we miss out on a few things that the previous bash bootstrap was giving us that we didn't realize. 

In this PR I am setting `TERM` when a tty is created and `PWD`. The other minimum env that a shell gets is usually `PATH`, `HOME` and `HOSTNAME`, but these are up to `login` to set, which for us I think is in the systemd config or similar. 

I verified these were a thing with: 

```
docker run --rm alpine:3.3 env
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
HOSTNAME=978593e25f81
HOME=/root
```

```
docker run --rm -it alpine:3.3 env
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
HOSTNAME=032e49df3771
TERM=xterm
HOME=/root
```

```
docker run --rm -it alpine:3.3 sh -c env
HOSTNAME=dfd4924645b5
SHLVL=1
HOME=/root
TERM=xterm
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
PWD=/
```

### DIFF
--- a/bootstrap/shell/shell.go
+++ b/bootstrap/shell/shell.go
@@ -26,6 +26,10 @@ var (
 	lockRetryDuration = time.Second
 )
 
+const (
+	termType = `xterm-256color`
+)
+
 // Shell represents a virtual shell, handles logging, executing commands and
 // provides hooks for capturing output and exit conditions.
 //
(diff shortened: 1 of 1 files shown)


ticky:
Just curious, is the 3.0 agent _also_ not currently running any local shell configuration files? (i.e. .profile et al)

lox:
If it's running under systemd, it won't pull in any shell configuration. Previously it did sometimes depending on the init system, and you got some shell-like variables by virtue of the bootstrap being bash.

ticky:
Right, I’ve had a complaint about this in Support. I understand that it’s a breaking change in 3.0, but what’s the expectation here? Should things like `PATH` be set in a hook now?

PR #757, opened 2018-05-04
Expose the instance number as env-var

petemounce:
(Haven't tried this out yet).

It would be convenient to make the instance number available as an environment variable, since then I can export it to centralised logging as a field in its own right without needing to parse the `BUILDKITE_AGENT_NAME` one, which tacks it onto the end of the hostname.

I think this might also end up helping with autoscaling, since it could be exposed via `buildkite-metrics` and allow me to choose an un-busy node to terminate when scaling down?

### DIFF
--- a/packaging/linux/root/usr/share/buildkite-agent/systemd/buildkite-agent@.service
+++ b/packaging/linux/root/usr/share/buildkite-agent/systemd/buildkite-agent@.service
@@ -8,6 +8,7 @@ After=network.target
 Type=simple
 User=buildkite-agent
 Environment=HOME=/var/lib/buildkite-agent
+Environment=BUILDKITE_AGENT_INSTANCE_NUMBER=%i
 ExecStart=/usr/bin/buildkite-agent start
 RestartSec=5
 Restart=on-failure
--- a/packaging/linux/root/usr/share/buildkite-agent/systemd/buildkite-agent@.service
+++ b/packaging/linux/root/usr/share/buildkite-agent/systemd/buildkite-agent@.service
@@ -8,7 +8,7 @@ After=network.target
 Type=simple
 User=buildkite-agent
 Environment=HOME=/var/lib/buildkite-agent
-Environment=BUILDKITE_AGENT_INSTANCE_NUMBER=%i
+Environment=BUILDKITE_AGENT_SYSTEMD_INSTANCE_NUMBER=%i
 ExecStart=/usr/bin/buildkite-agent start
 RestartSec=5
 Restart=on-failure


lox:
🤔Does `systemd` interpolate the `%i` before passing it in?

petemounce:
I hope so, will try it tomorrow.

petemounce:
Looks like it does, but is off-by-one?!

systemd-unit
systemdunit-proof

keithpitt:
If this is something specific to `systemd`, would it make more sense to call it something like: `SYSTEMD_INSTANCE_NUMBER`? I wouldn't want to confuse people with "why is there an instance number when I run it through latest ubuntu vs. an old version of ubuntu"

sj26:
The %n is resolved server side on Buildkite.com to the first value which creates a unique agent name, so it may not always correspond to your systemd %i. But you could specify an agent name when starting the agent overriding the config file name format, if you can generate something known to be unique. I think in systemd this should turn into the hostname plus the systemd template instance name, which should be unique?

```
Environment=BUILDKITE_AGENT_NAME=“%H-%I”
```

sj26:
Also, %I isn’t always a number ;-)

lox:
> It would be convenient to make the instance number available as an environment variable, since then I can export it to centralised logging as a field in its own right without needing to parse the BUILDKITE_AGENT_NAME one, which tacks it onto the end of the hostname.

I reckon setting this at a systemd level is the wrong way to solve these problems. For your logging, I'd suggest using a bootstrap script and parsing `BUILDKITE_AGENT_NAME` and then setting a specific env for your jobs like `BUILDKITE_AGENT_LOG_PREFIX`. 

> I think this might also end up helping with autoscaling, since it could be exposed via buildkite-metrics and allow me to choose an un-busy node to terminate when scaling down?

If you wanted to accomplish this (I still think random with graceful shutdown is a is a better approach), I'd suggest setting `host=<unique hostname>` as an agent tag for each agent, that way you can aggregate by that tag in metrics and do your autoscaling based on that. Another alternative might be adding a `--add-host-tag` or similar flag to the agent so that it automatically does that. Happy to help with the latter.

petemounce:
> I reckon setting this at a systemd level is the wrong way to solve these problems. For your logging, I'd suggest using a bootstrap script and parsing BUILDKITE_AGENT_NAME and then setting a specific env for your jobs like BUILDKITE_AGENT_LOG_PREFIX.

If I went with parsing, that would break by surprise if the buildkite agent name format changed. I'd prefer to not parse the name at my end when it's available in isolation, therefore.

I'll go with the systemd approach and modify the unit file in preference, I think.

> (I still think random with graceful shutdown is a is a better approach)

This is sufficient for auto-scaling brutally, but I'd like to achieve something more graceful.

The reasons for that
* some of our compile/test/package type builds are on the order of 45 minutes for our pre-merge pipeline, so interrupting one is a high cost. 
    * We expect that bazel's remote caching will reduce this loss, but we're not online with that quite yet. 
* some of our soak testing is long running  (literally "start a world, wait for things to happen, make some assertions over the course of days", where "world" might mean tens of cloud VMs), those are even higher cost to interruption. 
    * We're again, working to make that style of test more of a managed service, but not there yet.

re: metrics on tags - I hadn't clocked that the metrics binary would label according to tags as well - thanks, I'll look into that.

lox:
Thoughts on whether the recent addition of the `machine-id` tag changes this @petemounce? Could we look at getting per-machine metrics based on that tag?

petemounce:
I don't think so - the `machine-id` does not vary with the build-agent process-number.

However, I think I can derive the agent process instance number via removing `hostname` from `agent-name` (the heading of each table cell) and trimming the remaining `-` prefix.

bk-agent-meta

lox:
@petemounce I've reread the thread, and I must admit I'm kind of lost as to what you are trying to accomplish here. Reckon you could talk me through the problem you are trying to solve here again?

lox:
Let me know if you want to talk this one through any more @petemounce!

lox:
Closing this one for now, feel free to raise in discourse or post here if this is still a pressing need.

PR #759, opened 2018-05-07
Pass jobs an authenticated unix socket rather than an access token 

lox:
Currently we pass jobs `BUILDKITE_AGENT_ACCESS_TOKEN`, which is a per-agent-connection session token that is derived from the Agent Registration Token. This token is then used by agent commands that run within jobs, like `buildkite-agent pipeline upload` and the `artifact` and `meta-data` sub-commands. 

The downside with this approach is that the access token can be easily stolen by untrusted jobs (see https://github.com/buildkite/feedback/issues/360). 

This PR implements a proxy that listens on a unix socket or a localhost http port and proxies through to the upstream https://agent.buildkite.com. This should radically improve the security profile, as if it's a unix socket then it's not possible to access externally, or exfiltrate. If it's http based, you need access to whatever it's bound to (localhost) and even then, it only lasts for the duration of the job.

### DIFF
--- a/agent/api_client.go
+++ b/agent/api_client.go
@@ -1,6 +1,8 @@
 package agent
 
 import (
+	"bufio"
+	"errors"
 	"net"
 	"net/http"
 	"net/url"
@@ -8,6 +10,7 @@ import (
 	"time"
 
 	"github.com/buildkite/agent/api"
+	"github.com/buildkite/agent/logger"
 	"golang.org/x/net/http2"
 )
 
@@ -23,6 +26,15 @@ func APIClientEnableHTTPDebug() {
 }
 
 func (a APIClient) Create() *api.Client {
+	u, err := url.Parse(a.Endpoint)
+	if err != nil {
+		logger.Warn("Failed to parse %q: %v", a.Endpoint, err)
+	}
+
+	if u != nil && u.Scheme == `unix` {
+		return a.createFromSocket(u.Path)
+	}
+
 	httpTransport := &http.Transport{
 		Proxy:              http.ProxyFromEnvironment,
 		DisableKeepAlives:  false,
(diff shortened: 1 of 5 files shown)


sj26:
This is so cool!

Does this validate that requests through the socket can't escape from the `$BUILDKITE_AGENT_ENDPOINT` URI? Can you issue a different host or something through the socket, or use Host to trick it?

lox:
@sj26 the host gets rewritten. You could use it to access any path on agent.buildkite.com. Next step was to whitelist what could be accessed down to certain paths, but I don't think it's critical for the security story.

lox:
Big question in my head is how we could merge this in without a major version bump. I think we probably need to support both and then perhaps have an option to `buildkite-agent start` to enable this mode.

sj26:
@lox nice!

I've wanted this for so long, thank you!

I notice you've removed support for using tokens and direct api clients from the subcommands completely. Should we still allow connections using the token for now, to give folks a chance to migrate with deprecation notices or something?

lox:
> I notice you've removed support for using tokens and direct api clients from the subcommands completely. Should we still allow connections using the token for now, to give folks a chance to migrate with deprecation notices or something?

Yeah, just adding them back in so that either approach will work.

sj26:
I think that solves the upgrade path. A minor bump with deprecation warnings targetting v4, I think. We shouldn't be afraid of major versions imho.

lox:
Ok, this is now optional and needs to be enabled with experiments:

```
buildkite-agent start --debug --experiment "agent-socket"
```

petemounce:
Looks great to me. Definitely also interested in the Windows Edition too.

lox:
@sj26 I rewrote this on the plane. No more custom env for the socket, it just gets passed in as `unix:///path/to/socket` in `BUILDKITE_AGENT_ENDPOINT`, this meant far fewer changes. It also uses a randomly generated token to secure access to the proxy for defense-in-depth. 

Thoughts?

sj26:
Love your work, @lox. Also fixes my concerns around hijacking the socket location.

Should we use the `http+unix` scheme, like docker etc, or just `unix`?

lox:
> Should we use the http+unix scheme, like docker etc, or just unix?

We actually do! The http+unix scheme is for the url `http+unix://buildkite-agent` that the api client uses as it's base host name

lox:
This is good to go after a review from @keithpitt.

lox:
Agree with that thinking @keithpitt!

PR #762, opened 2018-05-09
Run local pipelines with `buildkite-agent run`

lox:
This is an experiment to see how hard it would be to run pipelines locally for development / testing. 

```bash
$ cd ~/Projects/myproject
$ buildkite-agent run .buildkite/pipeline.yml
~~~ Running my build
....
```

See also https://github.com/toolmantim/bksr and https://github.com/SomeoneWeird/bkrun.

### TODO

 * [ ] Repository-less checkout (run builds in current dir)
 * [ ] Artifacts support
 * [ ] Meta-data support
 * [ ] All the other steps and command step params

### DIFF
--- a/clicommand/global.go
+++ b/clicommand/global.go
@@ -1,9 +1,17 @@
 package clicommand
 
 import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path"
+	"path/filepath"
+	"strings"
+
 	"github.com/buildkite/agent/agent"
 	"github.com/buildkite/agent/experiments"
 	"github.com/buildkite/agent/logger"
+	"github.com/buildkite/agent/stdin"
 	"github.com/oleiade/reflections"
 	"github.com/urfave/cli"
 )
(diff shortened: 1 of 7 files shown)


lox:
This was a neat experiment, but after some thought the agent is the wrong place for it. We'll find another better home!

petemounce:
Some feedback on the feature in general:
* I initially thought "don't allow artifacts to be written" but then thought a little harder and realised that wouldn't fly, since a step might depend on artifacts created earlier in a pipeline. So then I thought `tmpfs` (and equivalent on Windows)?
  * (I wouldn't want a local run to have the possibility of ever writing artifacts to the same place that my beautifully pristine build agents would compared to filthy developer workstations)
* Similarly metadata.
* Ansible has some really nice features here around slicing and dicing playbooks to run only particular steps, or start at a particular task. `--start-at-task` for example, and `--tags` (to run only those steps tagged with xyz)
  * Definitely a nice to have, but I'd definitely also not want to force interactive input, because being able to run CI locally is an _ideal_ way to run a `git bisect` to figure out when a problem was introduced to `master` when `master` isn't built on every merge (I know, right? Who would _not_ do such an eminently sensible thing...).

lox:
Good feedback, thanks @petemounce!

PR #764, opened 2018-05-10
Avoid pausing after job has finished

sj26:
When we run a job we start two background goroutines — one polling for new logs to upload, and the other polling the job status to see if a job is cancelled during execution. They sleep in between polling. The job status sleep interval is set by buildkite.com through the agent api, and starts at 5 seconds increasing to 30 seconds when there are many agents connected. So these goroutines may sleep for up to 30 seconds.

When a job finishes, we tell the Buildkite agent api that the job is finished, then we wait for these background goroutines to exit. In the meantime, the Buildkite agent api might notice we're no longer running a job — we finished it — and assign us new work to do, but we're still waiting on those goroutines to finish for up to 30 seconds.

This change creates a context to pass into the goroutines which can be cancelled, makes sure that when we sleep in the goroutines we are also listening for a cancellation, and cancels the context when the job is finished to wakeup the goroutines, avoiding the up-to-30-second wait.

### DIFF
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -155,9 +155,6 @@ func (r *JobRunner) Run() error {
 		logger.Warn("%d chunks failed to upload for this job", r.logStreamer.ChunksFailedCount)
 	}
 
-	// Finish the build in the Buildkite Agent API
-	r.finishJob(finishedAt, r.process.ExitStatus, int(r.logStreamer.ChunksFailedCount))
-
 	// Wait for the routines that we spun up to finish
 	logger.Debug("[JobRunner] Waiting for all other routines to finish")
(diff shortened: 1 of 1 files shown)


sj26:
@keithpitt one more adjustment — we can move the job finish down slightly to make sure that everything has finished before we.. finish.

keithpitt:
@sj26 is this minor-version-releasable?

keithpitt:
Actually, we're probably due for 3.2 with the other changes we've made!

sj26:
@keithpitt yeah, it's a fix, doesn't introduce new functionality or deprecate any current functionality, so I reckon it's a patch release. We've shipped a 3.1.2 beta.

PR #766, opened 2018-05-14
Don't trim whitespace from the annotation body

### LINKED ISSUE #391: Multiple commands error detection on Windows

petemounce:
... because this prevents bullet-lists (for example) from being created via appending annotations together.

Fixes #391

### DIFF
--- a/clicommand/annotate.go
+++ b/clicommand/annotate.go
@@ -121,9 +121,6 @@ var AnnotateCommand = cli.Command{
 			body = string(stdin[:])
 		}
 
-		// Trim any whitespace edges on the annotation body
-		body = strings.TrimSpace(body)
-
 		// Create the API client
 		client := agent.APIClient{
 			Endpoint: cfg.Endpoint,


petemounce:
Cool, thanks.

I wasn't able to run the tests, by the way, by 

```shell
git clone ... 
cd agent
export GOPATH=$(pwd)
go test
```

It would be convenient to be able to do that - what have I got wrong?

lox:
go test ./...

petemounce:
Ta!

PR #767, opened 2018-05-16
Unused import breaks build.

derekmarcotte:
Clean-up from #766


lox:
Thanks, I fixed this in master before I noticed this!

PR #768, opened 2018-05-16
Propagate exit code > 1 out of failing hooks

### LINKED ISSUE #754: `command` hook exit codes are squelched
It appears that [exit codes generated from `command` hooks](https://github.com/buildkite/agent/blob/fc99bcc46cb8bc94f0e15cac0a8020ad7d821bea/bootstrap/bootstrap.go#L147-L154) (such as in the docker-compose plugin) are swallowed up rather than propagating all the way out to the job.

lox:
Closes #754.


ticky:
Cool, so if I’m reading this right, this will propagate back up to `if phaseError != nil {` in `func (b *Bootstrap) Start() int` and thus expose the true exit code! Looks good to me :D

Are we able to add test coverage for this behaviour? :)

lox:
I spent about a day trying to get the tests to work. I'll have another shot.

ticky:
Nice! Good stuff 😃

PR #770, opened 2018-05-18
Remove DEPENDENCIES.txt, it's out of date anyway

lox:


keithpitt:
So I added this file ages ago because of some security requirement from a vendor. We had to list out all of the agents dependencies and their licences - I'm not 100% sure if we can remove it.. Can we keep it updated with some code perhaps?

PR #771, opened 2018-05-18
Experiments docs and experiment cleanup

lox:
This adds an `EXPERIMENTS.md` which lists agent experiments available.

### DIFF
--- a/EXPERIMENTS.md
+++ b/EXPERIMENTS.md
@@ -3,7 +3,7 @@
 We frequently introduce new experimental features to the agent. You can use the `--experiment` flag, or the `experiment` config value to opt-in to them and test them out.
 
 ```bash
-buildkite-agent start --experiment "experiment1,expertiment2"
+buildkite-agent start --experiment experiment1 --experiment expertiment2
 ```
 
 ## Available Experiments
--- a/EXPERIMENTS.md
+++ b/EXPERIMENTS.md
@@ -6,11 +6,15 @@ We frequently introduce new experimental features to the agent. You can use the 
 buildkite-agent start --experiment experiment1 --experiment expertiment2
 ```
 
-Or uou can set them in your [agent configuration file](https://buildkite.com/docs/agent/v3/configuration) and they will apply whenever your agent runs:
+Or you can set them in your [agent configuration file](https://buildkite.com/docs/agent/v3/configuration):
 
 ```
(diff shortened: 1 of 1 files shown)


lox:
Worth noting that we don't pass experiments on to the bootstrap. Will figure that out when we have an experiment in the bootstrap I guess!

EXPERIMENTS.md
```diff
@@ -0,0 +1,23 @@
+# Experiments
+
+We frequently introduce new experimental features to the agent. You can use the `--experiment` flag, or the `experiment` config value to opt-in to them and test them out.
+
+```bash
+buildkite-agent start --experiment experiment1 --experiment expertiment2
+```
```
toolmantim: Could we add a config file example too? Maybe even above this? Only because I think that's the most common way people configure their agent, and people might not be sure of the array/list syntax.

lox: Yup, can do!

lox:
Change made @toolmantim.

EXPERIMENTS.md
```diff
@@ -0,0 +1,29 @@
+# Experiments
+
+We frequently introduce new experimental features to the agent. You can use the `--experiment` flag to opt-in to them and test them out:
+
+```bash
+buildkite-agent start --experiment experiment1 --experiment expertiment2
+```
+
+Or uou can set them in your [agent configuration file](https://buildkite.com/docs/agent/v3/configuration) and they will apply whenever your agent runs:
```
toolmantim: "you" typo here! Probably don't need "and they will apply whenever your agent runs" I reckon.

toolmantim: Otherwise, looking ace! 🎉

EXPERIMENTS.md
```diff
@@ -0,0 +1,29 @@
+# Experiments
+
+We frequently introduce new experimental features to the agent. You can use the `--experiment` flag to opt-in to them and test them out:
```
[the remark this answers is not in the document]
keithpitt: > I dunno, I'm fine with the current behaviour. If it's there, it turns on, if not, it doesn't.

Cool cool, just wanted to make sure it didn't blow up with a "`experiment-123` does not exist" type thing..

lox:
@keithpitt I made the changes you requested, let me know what you think.

PR #772, opened 2018-05-18
Fix broken list parsing in cli arguments --tags and --experiments

lox:
Currently, `buildkite-agent start --tags "a=1,b=2"` sets a single tag of `a=1,b=2`. This is a shortcoming of the upstream urfave.cli library (see https://github.com/urfave/cli/issues/62) which expects `--tags "a=1" --tags "a=2"`.

This adds a normalization for --tags and --experiments that makes them behave as expected.

### DIFF
--- a/agent/agent_pool.go
+++ b/agent/agent_pool.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"runtime"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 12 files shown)


keithpitt:
Oh, I assume it's fine with `--tags "foo=bang,bar,bing"`

keithpitt:
(Where you specifically want `foo` to be a comma separated list..)

lox:
Nope! Won't work!

lox:
Wouldn't work in config either.

keithpitt:
Just chatted with @lox in real life, we've never supported commas in tag values before, so it's fine to ship as-is!

PR #765, opened 2018-05-13
Add a pipeline parse command

toolmantim:
For https://github.com/toolmantim/bksr I'd like to be able to support the env interpolation that the agent does, as well as inherit the pipeline config search paths, and a few other things that are built into the agent. This adds a new pipeline parse command to help other tools that might want to process pipelines:

```
$ go run main.go pipeline parse
Usage:

   buildkite-agent pipeline parse <file> [arguments...]

Description:

   Prints to STDOUT the parsed pipeline as JSON.

Example:

   $ buildkite-agent pipeline parse
   $ buildkite-agent pipeline parse my-custom-pipeline.yml
   $ ./script/dynamic_step_generator | buildkite-agent pipeline parse

Options:

   --no-color  Don't show colors in logging [$BUILDKITE_AGENT_NO_COLOR]
   --debug     Enable debug mode [$BUILDKITE_AGENT_DEBUG]
```

Example output, for the agent repo:

```
$ env BUILDKITE_COMMIT=commit123 BUILDKITE_BRANCH=branch123 BUILDKITE_MESSAGE=message123 go run main.go pipeline parse
2018-05-13 22:46:29 INFO   Searching for pipeline config...
2018-05-13 22:46:29 INFO   Found config file ".buildkite/pipeline.yml"
[...]

PR #781, opened 2018-06-06
Add a --dry-run to pipeline upload that dumps json

lox:
This is a follow on of https://github.com/buildkite/agent/pull/765. 

It adds a `buildkite-agent pipeline upload --dry-run` command that outputs the pretty-printed pipeline json rather than uploading it. This works locally with no configuration required.

### DIFF
--- a/clicommand/pipeline_upload.go
+++ b/clicommand/pipeline_upload.go
@@ -1,6 +1,7 @@
 package clicommand
 
 import (
+	"encoding/json"
 	"io/ioutil"
 	"os"
 	"path"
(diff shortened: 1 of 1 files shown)


clicommand/pipeline_upload.go
```diff
@@ -174,6 +181,30 @@ var PipelineUploadCommand = cli.Command{
...
+			if err := enc.Encode(parsed); err != nil {
+				logger.Fatal("%#v", err)
+			}
+
+			os.Exit(0)
+		}
+
+		// Check we have a job id set if not in dry run
+		if cfg.Job == "" {
+			logger.Fatal("Missing job parameter. Usually this is set in the environment for a Buildkite job via BUILDKITE_JOB_ID.")
```
lox: I'd love a way to make more informative error messages like this the standard vs the standard "required parameter" message.

lox:
@toolmantim sure! Would be good to verify it works with bksrsly.

toolmantim:
I've just done some standalone tests using various pipeline.yml files, and the env interpolation and JSON output seems to work really nicely! 🎉

I jumped into bksrsly to update it to so it works with it, but it's gonna take a bit more work than I thought, because sub-processes etc. I'll let you know when I've got something working 👌🏼

jgonera:
@lox Does this validate the schema of the pipeline too? Will it fail on invalid pipelines?

PR #790, opened 2018-07-05
Allow --no-plugins=false to force plugins on

lox:
By default plugins are disabled when no-command-eval or no-local-hooks is set. This allows a way around that.

I kind of hate the syntax of `--no-plugins=true`, and I hate even more the ternary nature of this, in that we only look at the `false` value if it's specifically set vs just the default value. Suggestions welcomed.


toolmantim:
👍🏼 Noice! Sounds sensible, given the way the args are set up.

Issue #773, opened 2018-05-21
Support `%hostname` in agent metadata


therealbnut:
We want to add hostname to the agent metadata, so we can target individual troublesome machines. We ideally don’t want to have to change the configuration on each machine to achieve this.

**User story**

We had a troublesome machine over the weekend, but didn’t have any way to target it directly (we have ssh disabled). This isn't the first time this has happened, but its the first time it has happened on the weekend and the people affected didn't know the hack workarounds.

In the end we had to run a script on all machines, and filter for the specific one with `BUILDKITE_AGENT_NAME`, and get it to update the agent's config file. We could not just turn off the agent because it has a launch daemon that keeps it running.

lox:
This is a great idea!

lox:
Now that we have #791, I'm going to close this one. Please let me know if that doesn't solve your problem @therealbnut.

lox:
It's in v3.3.0 @therealbnut! https://github.com/buildkite/agent/releases/tag/v3.3.0

PR #791, opened 2018-07-06
Allow tags from the host to be automatically added

lox:
We presently have flags to automatically add metadata from GCP and EC2 as tags, this adds `add-host-tags` which adds some tags from the underlying host, namely `hostname`, `os` and a new one `machine-id`. 

Often people want to target a specific machine, such as in https://github.com/buildkite/agent/issues/773. This gives them an easy way to enable a per-host tag in their agent metadata. 

Machine id is calculated with https://github.com/denisbrodbeck/machineid from underlying host metadata, generally UUID's. This means if you have multiple agents on the same host you will get the same `machine-id` tag. It's portable across all the operating systems we support.

The agent also reports the `machine-id` via the agent registration api so that we can use it to co-ordinate things like shutting down all agents on a machine in future, like in https://github.com/buildkite/agent/pull/590. 

image


lox:
I expected that to be much more contentious @keithpitt!

keithpitt:
@lox as in, I might have more feedback?

lox:
I figured there might be a reason why you didn't do this one in the first place. Something something "having host tags leads to builds that target the same agent as they previously ran on".

keithpitt:
Oh right! I always like having an answer to something, even if it's discouraged. What you've done here makes stuff possible that was (harder) before. Fine with me!

lox:
Agreed!

lox:
I'm just waiting on https://github.com/denisbrodbeck/machineid/pull/4, otherwise I will use my own fork.

petemounce:
Is there a docs PR to link to see how to use this?

Issue #794, opened 2018-07-12
Cancellation behavior on Windows and how to improve it?


amitsaha:
On Windows, job cancellation is:

```
# process\process.go
if runtime.GOOS == "windows" {
    // Sending Interrupt on Windows is not implemented.
    // https://golang.org/src/os/exec.go?s=3842:3884#L110
   err = exec.Command("CMD", "/C", "TASKKILL", "/F", "/PID", strconv.Itoa(p.Pid)).Run()
```

This essentially (due to no fault of buildkite agent) means that the spawned process doesn't get any opportunity to perform any cleanup tasks. 

What do we do then if we want to run cleanup tasks when a job is cancelled?

**Allow custom cancel command**

Instead of using `signals` as the only supported mechanisms for indicating to the spawned process, we can allow a custom command to be executed. An example of a custom command may be invoking a HTTP API endpoint which will be exposed by the spawned process which will trigger the shutdown process as part of which it will perform cleanup tasks.

__How do we implement it?__

For example (to minimize changes across the board), the agent can look for a environment variable
`CANCEL_COMMAND`:

```
steps:

    - label:  ':hammer: Build and Run All Tests'
      command: 'powershell.exe -File .buildkite/scripts/RunTests.ps1'
      env:
          CANCEL_COMMAND: 'Invoke-WebRequest http://127.0.0.1:8000/cancel'
```

And then:

```
# process\process.go
func (p *Process) Kill() error {
       if os.Getenv("CANCEL_COMMAND") {
           # Execute the custom command
       } else {
          # Current signalling behavior
           # ..
	
    }
```

Of course what I propose here is going to be usable on non-Windows platforms as well. However, this would be useful most on Windows platforms. 

What do you folks think of this?

lox:
Yeah, windows makes this really hard. If we had a `pre-cancel` hook, or had `pre-exit` reliably execute on windows, you could implement `CANCEL_COMMAND` yourself.

amitsaha:
> pre-exit reliably execute on windows,

Could we make the `pre-exit` hook execute even when it's a cancellation?

lox:
@amitsaha so currently we run `buildkite-agent bootstrap` as a sub-process to execute a job. It handles plugin setup, checkout and command execution, as well as all the hook.

We run `pre-exit` in that subprocess by catching the signal and running that hook before exiting, which is why it doesn't work in windows.

To make it work, we'd need to move `pre-exit` out of the bootstrap.

[...]

PR #795, opened 2018-07-13
Cancellation on Windows: Pass `/T` to `taskkill`

amitsaha:
`/T` will make `taskkill` kill all the child processes as well.  This should improve the cancellation behavior. (#794) 

Please let me know if I should file a separate issue.

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -260,7 +260,7 @@ func (p *Process) Kill() error {
 	if runtime.GOOS == "windows" {
 		// Sending Interrupt on Windows is not implemented.
 		// https://golang.org/src/os/exec.go?s=3842:3884#L110
-		err = exec.Command("CMD", "/C", "TASKKILL", "/F", "/PID", strconv.Itoa(p.Pid)).Run()
+		err = exec.Command("CMD", "/C", "TASKKILL", "/F", "/T", "/PID", strconv.Itoa(p.Pid)).Run()
 	} else {
 		// Send a sigterm
 		err = p.signal(syscall.SIGTERM)


lox:
This seems sensible. I'll do some testing on Windows.

lox:
Some interesting context and alternative approaches in https://github.com/golang/dep/issues/862.

amitsaha:
thanks for the merge @lox

amitsaha:
>Some interesting context and alternative approaches in golang/dep#862.

Very useful reference. Any objections to incorporating `TerminateTree()` for the cancellation as implemented [here](https://github.com/mattn/psutil/blob/master/psutil_windows.go#L53)?

lox:
@amitsaha I didn't see any huge upside in `TerminateTree()` over shelling out to `taskkill`, did you?

amitsaha:
@lox It [looks](https://github.com/mattn/psutil/blob/master/psutil_windows.go#L28) like  `TerminateTree` would kill the entire tree i.e. including processes spawned by child processes of the parent process? (whereas, it seems `taskkill /T` would kill only the immediate child processes)? (I haven't verified the `taskkill /T` behavior, so could be wrong)

lox:
Ah right, in that case I'd absolutely support `TerminateTree()` being used, I assumed the `/T` would have the same behaviour.

amitsaha:
I will have to look into this more, but it seems `taskkill /T` does also terminate child processes spawned by child processes.

amitsaha:
indeed, that's the case, so, we don't need `TerminateTree`. Some demo code: https://github.com/amitsaha/golang-powershell-task-kill-demo

PR #796, opened 2018-07-17
Warn in bootstrap when protected env are used

lox:
We frequently have people try and set things like `BUILDKITE_GIT_CLEAN_FLAGS` in their pipeline global environment or per-step environment and get very confused why it doesn't work.

The reason it doesn't work is that we override certain environment variables for each build in the job runner to ensure they are correct as they are security sensitive: https://github.com/buildkite/agent/blob/c9645249f076beef3cebce34dc9980fe8c54d502/agent/job_runner.go#L264-L276

We do allow certain environment variables set by hooks to override these values, but we completely ignore values from the pipeline environment. 

~~This change rewrites the original values so they are available in the bootstrap if the user wants them, for instance `BUILDKITE_GIT_CLEAN_FLAGS` becomes `BUILDKITE_X_GIT_CLEAN_FLAGS`~~. The job runner sets ignored keys via `BUILDKITE_IGNORED_ENV` and then the bootstrap then prints a warning:

20180717-6kpt6kf7osnyvzw

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -349,16 +349,16 @@ func (b *Bootstrap) setUp() error {
 
 	b.shell.Env.Set("BUILDKITE_BUILD_CHECKOUT_PATH", filepath.Join(b.BuildPath, dirForAgentName(b.AgentName), b.OrganizationSlug, b.PipelineSlug))
 
-	// The job runner rewrites any protected environment variables to a certain format
-	// This detects those environment vars and shows a warning to the user so they don't get confused
+	// The job runner sets BUILDKITE_IGNORED_ENV with any keys that were ignored
+	// or overwritten. This shows a warning to the user so they don't get confused
 	// when their environment changes don't seem to do anything
-	if ignored := b.ignoredEnv(); len(ignored) > 0 {
+	if ignored, exists := b.shell.Env.Get("BUILDKITE_IGNORED_ENV"); exists {
 		b.shell.Headerf("Detected protected environment variables")
 		b.shell.Commentf("Your pipeline environment has protected environment variables set. " +
 			"These can only be set via hooks, plugins or the agent configuration.")
 
-		for _, env := range ignored {
-			b.shell.Warningf("Ignoring %s", env)
+		for _, env := range strings.Split(ignored, ",") {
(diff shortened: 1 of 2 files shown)


bootstrap/bootstrap.go
```diff
@@ -349,6 +349,20 @@ func (b *Bootstrap) setUp() error {
 
 	b.shell.Env.Set("BUILDKITE_BUILD_CHECKOUT_PATH", filepath.Join(b.BuildPath, dirForAgentName(b.AgentName), b.OrganizationSlug, b.PipelineSlug))
 
+	// The job runner rewrites any protected environment variables to a certain format
+	// This detects those environment vars and shows a warning to the user so they don't get confused
+	// when their environment changes don't seem to do anything
+	if ignored := b.ignoredEnv(); len(ignored) > 0 {
+		b.shell.Headerf("Detected protected environment variables")
+		b.shell.Commentf("Your pipeline environment has protected environment variables set. These can only be set via hooks or via the agent configuration.")
```
toolmantim: Might want to mention plugins here also.

lox: Yup, good point!

lox:
> Looking at the error message makes me wonder though, should we let people configure BUILDKITE_GIT_CLEAN_FLAGS via pipeline env? Or is there security implications?

💯I believe we didn't in v2 because we were concerned about command injection, but we're tokenizing the arguments now. @keithpitt do you remember any context on that one?

lox:
There is also a complexity where we use negative environment vars like `BUILDKITE_NO_GIT_SUBMODULES` for the `buildkite-agent start` command to opt-out of functionality, but the bootstrap itself gets things in an opt-in format like `BUILDKITE_GIT_SUBMODULES`. 

I think provided we clearly document what can be set in ENV or hooks via https://buildkite.com/docs/builds/environment-variables then it should be ok, but it's complicated.

lox:
Ok @keithpitt, done. The runner sets `BUILDKITE_IGNORED_ENV=x,y,x` with the ignored env keys in there.

lox:
Thanks for the feedback @keithpitt! I like where it ended up.

keithpitt:
🥇

PR #798, opened 2018-07-17
Support --command-only and --checkout-only in bootstrap

lox:
This allows for only running the command or checkout phase of the bootstrap. This allows for things like https://github.com/wolfeidau/buildkite-serverless-agent to run checkout and command phases independently.

The neat thing about this split is that generally you need to expose ssh key material to the checkout phase, where as you don't want to expose it to the command phase. Suspect this will end up being useful for Public Build isolation in several ways. 

```bash
$ buildkite-agent bootstrap \
  --job "llamas" \
  --repository "$PWD" \
  --commit "HEAD" \
  --branch "master" \
  --agent "my-agent" \
  --organization "buildkite" \
  --pipeline "agent" \
  --pipeline-provider git \
  --build-path /usr/local/var/buildkite-agent/builds \
  --command-only 
  --command "echo hello world"

~~~ Running commands
$ echo hello world
hello world
```

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -29,6 +29,9 @@ type Bootstrap struct {
 	// Config provides the bootstrap configuration
 	Config
 
+	// Specific phases to execute
+	CheckoutOnly, CommandOnly bool
+
 	// Shell is the shell environment for the bootstrap
 	shell *shell.Shell
 
@@ -73,12 +76,16 @@ func (b *Bootstrap) Start() (exitCode int) {
 		return shell.GetExitCode(err)
 	}
 
-	// These are the "Phases of bootstrap execution". They are designed to be
-	// run independently at some later stage (think buildkite-agent bootstrap checkout)
-	var phases = []func() error{
-		b.PluginPhase,
-		b.CheckoutPhase,
-		b.CommandPhase,
+	// These are the "Phases of bootstrap execution".
+	var phases = []func() error{b.PluginPhase}
+
+	switch {
+	case b.CheckoutOnly:
+		phases = append(phases, b.CheckoutPhase)
+	case b.CommandOnly:
+		phases = append(phases, b.CommandPhase)
+	default:
+		phases = append(phases, b.CheckoutPhase, b.CommandPhase)
 	}
 
 	var phaseError error
--- a/clicommand/bootstrap.go
+++ b/clicommand/bootstrap.go
@@ -56,6 +56,8 @@ type BootstrapConfig struct {
 	PTY                          bool   `cli:"pty"`
 	Debug                        bool   `cli:"debug"`
 	Shell                        string `cli:"shell"`
+	CheckoutOnly                 bool   `cli:"checkout-only"`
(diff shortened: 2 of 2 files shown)


lox:
@wolfeidau thoughts?

keithpitt:
@lox I dig it! I'm wondering about the API, instead of `--checkout-only` or `--command-only`, what if it was just `--phase=checkout`, `--phase=command`?

That way you can avoid confusion between `buildkite-agnet bootstrap --checkout-only --command-only` being used at the same time?

If you also allow `--phase=plugin` that'd allow machines to actually pre-install plugins I think? Which could also be a cool side-effect?

lox:
I started out with dedicated sub-commands, e.g `buildkite-agent bootstrap checkout`, but it duplicated a ton of code. Next up I tried `--phases`, but I didn't like that it leaked an internal implementation detail. The user just wants to run checkout only or the command only, they don't want to have to worry about things like orders or phases, or whether the command phase needs the plugin phase before it. 

In terms of implementation, it also meant pushing awareness of what phases were allowed to the cli side of things to validate them, and it made the code messier. 

I ended up liking the `--checkout-only` approach better. I also like that if we add phases in future, we don't break integrations that have listed specific phases to execute. 

How about I add `--plugins-only` too?

keithpitt:
My only issue with the `--only` properties is that things get weird if you chain them to the bootstrap command, then it just reads a bit wrong since `buildkite-agnet bootstrap --checkout-only --command-only --plugins-only` didn't make much sense. We'd also need to keep expanding the footprint of the `--only` flags if we add another phase.

> In terms of implementation, it also meant pushing awareness of what phases were allowed to the cli side of things to validate them, and it made the code messier.

You could probably put a slice of valid phases as a constant and reference it?

> but I didn't like that it leaked an internal implementation detail

I'd say that phases are awesome! And we should totally document them and talk about them...

> __Bootstrap Phases__
>
> The `bootstrap` process is made up of 3 phases: `plugin`, `checkout` and `command`. Each phase represents a step within the bootstrap process required to run your jobs. You can choose to run only one of the phases using the --phase option. 
>
> If you choose to only run the `command` phase, you'll need to ensure you've setup your repository checkout's manually.
>
> You can use `--phase=plugin` to just run the `plugin` phase of the bootstrap process. If you know ahead which plugins your jobs use, you can use this to pre-install plugins on agents before jobs are run.

Having said all that, I'm cool with what ever you choose :) I'm like a 25% care factor on it, so what ever you think is best!

lox:
Yeah, cool, ok I'll try it again. Would you expect to be able to specify multiple phases?

keithpitt:
> Would you expect to be able to specify multiple phases?

I wouldn't no. I think `--only-phase` would be OK if you though that'd be better?

lox:
The tough bit is that you really need the `plugin` phase to be executed in the same process as `checkout` and/or `command`, because you need to get the values from the hooks. So really you end up with three different combos that will be common: `plugin` on it's own, `plugin+checkout` and `plugin+command`.

lox:
Closing in favour of #799.

PR #799, opened 2018-07-18
Allow specifying which phases bootstrap should execute

lox:
Another take on #798, allows specific phases of the bootstrap to be executed.

```
$ buildkite-agent bootstrap \
  --job "llamas" \
  --repository "$PWD" \
  --commit "HEAD" \
  --branch "master" \
  --agent "my-agent" \
  --organization "buildkite" \
  --pipeline "agent" \
  --pipeline-provider git \
  --build-path /usr/local/var/buildkite-agent/builds \
  --phases "plugin,command" \
  --command "echo hello world"

~~~ Running commands
$ echo hello world
hello world
```

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -88,18 +88,19 @@ func (b *Bootstrap) Start() (exitCode int) {
 		return false
 	}
 
-	//  Execute the bootstrap phases
-	var phaseError error
-	for phase, phaseFunc := range map[string]func() error{
-		`plugin`:   b.PluginPhase,
-		`checkout`: b.CheckoutPhase,
-		`command`:  b.CommandPhase,
-	} {
-		if includePhase(phase) {
-			if phaseError = phaseFunc(); phaseError != nil {
-				break
-			}
-		}
+	//  Execute the bootstrap phases in order
+	var phaseErr error
+
+	if includePhase(`plugin`) {
+		phaseErr = b.PluginPhase()
+	}
+
+	if phaseErr == nil && includePhase(`checkout`) {
+		phaseErr = b.CheckoutPhase()
+	}
+
+	if phaseErr == nil && includePhase(`command`) {
+		phaseErr = b.CommandPhase()
 	}
 
 	if err := b.uploadArtifacts(); err != nil {
@@ -109,9 +110,9 @@ func (b *Bootstrap) Start() (exitCode int) {
 
 	// Phase errors are where something of ours broke that merits a big red error
 	// this won't include command failures, as we view that as more in the user space
-	if phaseError != nil {
-		b.shell.Errorf("%v", phaseError)
-		return shell.GetExitCode(phaseError)
+	if phaseErr != nil {
+		b.shell.Errorf("%v", phaseErr)
+		return shell.GetExitCode(phaseErr)
 	}
 
 	// Use the exit code from the command phase
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -29,6 +29,9 @@ type Bootstrap struct {
 	// Config provides the bootstrap configuration
 	Config
 
+	// Phases to execute, defaults to all phases
+	Phases []string
+
 	// Shell is the shell environment for the bootstrap
 	shell *shell.Shell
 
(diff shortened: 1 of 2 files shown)


bootstrap/bootstrap.go
```diff
@@ -73,19 +76,29 @@ func (b *Bootstrap) Start() (exitCode int) {
...
-
-	for _, phase := range phases {
-		if phaseError = phase(); phaseError != nil {
-			break
+	for phase, phaseFunc := range map[string]func() error{
+		`plugin`:   b.PluginPhase,
+		`checkout`: b.CheckoutPhase,
+		`command`:  b.CommandPhase,
+	} {
+		if includePhase(phase) {
```
keithpitt: Cool cool, so it doesn't matter what order they're defined it, them just existing in the phases option just turns them on/off in the right order?

lox: Yeah, I can't decide if that's weird or not.

clicommand/bootstrap.go
```diff
@@ -230,6 +242,11 @@ var BootstrapCommand = cli.Command{
 			EnvVar: "BUILDKITE_SHELL",
 			Value:  DefaultShell(),
 		},
+		cli.StringSliceFlag{
+			Name:   "phases",
+			Usage:  "The specific phases to execute",
```
keithpitt: "The specific phases to execute. The order they're defined is is irrelevant."

toolmantim:
This is cool! So you can run just the `command` part now from your own command hook? 🙌🏼

keithpitt:
@toolmantim theoretically!

lox:
@toolmantim unfortunately that would create and infinite loop and the universe would collapse.

lox:
Because you'd run the `checkout` phase of the bootstrap, which would try and defer to the `checkout` hook that you were calling from. Perhaps we need a `--default-checkout` and `--default-command` or similar.

keithpitt:
@lox unless you did something sneaky like:

```
buildkite-agent bootstrap --phase command --hook-path /dev/null
```

Right? So it didn't run the `command` hook again?

lox:
That would just cover the agent hooks, it would still run repository hooks and plugin hooks. Worse than that, it would skip things like your environment hook, so you'd have no secrets.

keithpitt:
But at `command` hook the environment hook would have already run, right?

lox:
No. This is where it gets gnarly. Each invocation of the bootstrap will run the setUp hooks, e.g `environment` and then the `pre` and `post` for the phase (either `checkout` or `command`) and then the tearDown hooks, e.g `pre-exit` and artifact uploads.

lox:
We could theoretically add a `setup` and `teardown` phase (or similar), but it would change the semantics of phases to have the `teardown` phase run even when other phases fail (as it currently does).

toolmantim:
Chatting with @keithduncan, he reminded me that we have this option, and I realised that I don’t think we ever updated the [documentation on agent hooks](
https://buildkite.com/docs/agent/v3/hooks) to mention this? Anyone looking to optionally override the checkout hook, for example, would need to know how to call the original checkout phase in their `if/else` block.

lox:
@toolmantim yup, not documented. I'm not sure it should be yet, it's still really a bootstrap implementation detail. I'd want to document in more detail how and why you'd override the bootstrap first, at which point that would be a good thing to have in there.

There isn't a good mechanism for calling the original checkout phase at present.

lox:
@toolmantim I'd considered adding a `bootstrap default-checkout` command that did the default checkout, but hadn't ever implemented it. 

@keithpitt and I decided the best approach was to avoid that complexity and to allow the checkout to be skipped by a `pre-checkout` hook setting `BUILDKITE_REPO` to empty. Would that solve whatever you are trying to do?

toolmantim:
Oh right, sorry! I thought that was the use case for the phases argument, didn’t realize it was more of an implementation detail.

The use I had in mind was a checkout hook that did:

```bash
if [[ “$BUIDLKITE_PIPELINE_SLUG == “special-cased-pipeline” ]]; then
  # Do some optimized checkout
else
  buildkite-agent bootstrap --phases checkout
  # or: buildkite-agent default-hook checkout
fi
```

Or a command hook:

```bash
if [[ “$BUIDLKITE_PIPELINE_SLUG == “special-cased-pipeline” ]]; then
  # Do something special
else
  buildkite-agent bootstrap --phases command
  # or: buildkite-agent default-hook command
fi
```

I think they’re the only two hooks that override/disable the default behavior?

The repo thing works for checkout:

```bash
if [[ “$BUIDLKITE_PIPELINE_SLUG == “special-cased-pipeline” ]];
  # Disable default checkout behavior
  export BUILDKITE_REPO=“”
  # Do some optimized checkout
fi
```

I suppose that might work for the command hook to?

The only problem with it might be if you wanted to use the original value of `$BUILDKITE_REPO` elsewhere (say, in a plugin that was expecting it to be set, given it’s a standard env).

We could add a `BUILDKITE_{CHECKOUT,COMMAND}=false` or similar, might be another idea.

toolmantim:
I might remove the tag for now tho, sounds like this was the wrong PR!

lox:
The thing about this is that even `buildkite-agent bootstrap --phases command` will still run things like `environment, `pre-exit` and all the artifacts stuff, so it's not quite a fit for that usecase. It's more for where you want to skip checkout entirely in the bootstrap, including all the hooks.

toolmantim:
Gotcha! Sounds like phases is a completely different thing to what I was imagining, and shouldn’t be used in hooks at all. I’ll check back with @keithduncan what he thought it’d be useful for.

PR #812, opened 2018-08-13
Remove the checkout dir if the checkout phase fails

lox:
There are several ways a git repository checkout can become corrupted that are hard to detect (have tried). This manifests as failures from one of the several git commands we call, either the `git remote set-url`, or the `git clean` or the `git checkout`, etc. Previously we dealt with this problem by removing the checkout directory on certain failures, but we keep discovering new ones.

We retry the checkout process if it fails, so this removes the checkout after the first failure.

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -704,6 +704,13 @@ func (b *Bootstrap) CheckoutPhase() error {
 			err := b.defaultCheckoutPhase()
 			if err != nil {
 				b.shell.Warningf("Checkout failed! %s (%s)", err, s)
+
+				// Checkout can fail because of corrupted files in the checkout
+				// which can leave the agent in a state where it keeps failing
+				// This removes the checkout dir, which means the next checkout
+				// will be a lot slower (clone vs fetch), but hopefully will
+				// allow the agent to self-heal
+				_ = b.removeCheckoutDir()
 			}
 			return err
 		}, &retry.Config{Maximum: 3, Interval: 2 * time.Second})
@@ -765,14 +772,10 @@ func (b *Bootstrap) defaultCheckoutPhase() error {
 	if fileExists(existingGitDir) {
 		// Update the the origin of the repository so we can gracefully handle repository renames
 		if err := b.shell.Run("git", "remote", "set-url", "origin", b.Repository); err != nil {
-			// Remove the checkout as often this is due to a corrupt git repo
-			_ = b.removeCheckoutDir()
 			return err
 		}
 	} else {
 		if err := gitClone(b.shell, b.GitCloneFlags, b.Repository, "."); err != nil {
-			// Remove the checkout as often this is due to files left in the dir
-			_ = b.removeCheckoutDir()
 			return err
 		}
 	}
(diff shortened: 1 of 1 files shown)


bootstrap/bootstrap.go
```diff
@@ -704,6 +704,13 @@ func (b *Bootstrap) CheckoutPhase() error {
 			err := b.defaultCheckoutPhase()
 			if err != nil {
 				b.shell.Warningf("Checkout failed! %s (%s)", err, s)
+
+				// Checkout can fail because of corrupted files in the checkout
+				// which can leave the agent in a state where it keeps failing
+				// This removes the checkout dir, which means the next checkout
+				// will be a lot slower (clone vs fetch), but hopefully will
+				// allow the agent to self-heal
+				_ = b.removeCheckoutDir()
```
petemounce: This LGTM.

I'd love to see the agent also maintain a `bare` clone elsewhere as a local mirror, to optimise away the network trip?

lox: Me too, but the scope of that is a lot bigger I think.

jgavris: This would be *killer*.

keithpitt: 💯 totally agree. Would happily accept a PR if you'd like to see it sooner @petemounce @jgavris :) - but it's something we absolutely want to do in the future!

keithpitt:
Hrm..yeah I can see why this change is a bit tricky. I'm trying to think of a failure that's an "OK" failure... (one that just needs to be retried, and not a full re-clone). Maybe a transient network failure if it's a super large `git fetch`?

lox:
> Hrm..yeah I can see why this change is a bit tricky. I'm trying to think of a failure that's an "OK" failure... (one that just needs to be retried, and not a full re-clone). Maybe a transient network failure if it's a super large git fetch?

Yup, although that was already broken somewhat by destroying the working dir on git clone failing.

keithpitt:
But in a transient `git fetch` failure, the original `git clone` would have succeeded right? It'd just retry normally and then work the second time (or hopefully the third). Wouldn't this change mean that a regular repo that's working fine, be deleted on a future error that has nothing to do with the corruption of the repo?

lox:
@keithpitt maybe, yeah, but it's also very hard to know what operations might corrupt your git repo.

petemounce:
@lox that's a reassuring thing to need to say about a source control system :)

lox:
Distributed systems are hard ;)

lox:
A good example of a corrupted git checkout:

```
fatal: update_ref failed for ref 'HEAD': cannot lock ref 'HEAD': Unable to create '/var/lib/buildkite-agent/builds/buildkite-xxx-5/xxx/xxx/.git/HEAD.lock': File exists.
```

This lock file was left by a crashed git process.

petemounce:
Seen this again - any update on when this might get merged?

lox:
@keithpitt and I talked about this at some length. Our concerns with this were that deleting checkouts on error is potentially very expensive, as a fresh checkout requires a slow clone. Any ephemeral network error that occurs in checkout will cause deletion with this change.  

That said, we think that correctness and self-healing systems trumps fast checkouts. So, we're gonna merge this and then look at bare repositories and reference clones for checkout next. Thanks for your patience all ❤

petemounce:
Thanks! I think that's the right choice. Broken checkouts burn or agents to an unrecoverable state. I didn't think of a way to detect this automatically, in a way that could distinguish between _agent_ got into this state vs build script did.

The former is grounds for auto-burn-and-replace, the latter not, I thought.

PR #813, opened 2018-08-16
`annotate` takes body as an arg, or reads from a pipe

sj26:
The `buildkite-agent annotate --help` output isn't quite correct. The command takes an annotation body as an arg, or reads from a pipe. It does not accept file paths.

@bjeanes is this better?


bjeanes:
Yeah that's clearer, and would address one of the issues I raised in buildkite/docs#296

PR #816, opened 2018-08-22
Expose metrics via expvar interface

lox:
After talking a bit with @toolmantim about what metrics we could expose, I did some investigating. Turns out golang has a nice and minimal interface for exposing metrics. I prototyped it out and it works pretty well. 

To activate, you set `--metrics` on `buildkite-agent start` or set `metrics=true` in the config, which starts up a webserver as part of the buildkite-agent that you can view on the standard url `https://localhost:8080/debug/vars`.

See http://www.mikeperham.com/2014/12/17/expvar-metrics-for-golang/ for more details and details of how the datadog agent consumes these metrics at https://www.datadoghq.com/blog/instrument-go-apps-expvar-datadog/

The metrics I'm tracking initially are:

* [x] `uptime`
* [x] `heartbeats.Total`
* [x] `heartbeats.Success`
* [x] `heartbeats.Fail`
* [x] `pings.Total`
* [x] `pings.Success`
* [x] `pings.Fail`
* [ ] `jobs.Total`
* [ ] `jobs.Success`
* [ ] `jobs.Fail`

Open to suggestions!

### DIFF
--- a/agent/agent_pool.go
+++ b/agent/agent_pool.go
@@ -2,6 +2,7 @@ package agent
 
 import (
 	"errors"
+	"expvar"
 	"fmt"
 	"os"
 	"runtime"
@@ -61,6 +62,11 @@ func (r *AgentPool) Start() error {
 	logger.Debug("Ping interval: %ds", registered.PingInterval)
 	logger.Debug("Job status interval: %ds", registered.JobStatusInterval)
 	logger.Debug("Heartbeat interval: %ds", registered.HearbeatInterval)
+
+	var start = time.Now()
+	expvar.Publish("uptime", expvar.Func(func() interface{} {
+		return int64(time.Since(start) / time.Second)
+	}))
 
 	// Now that we have a registered agent, we can connect it to the API,
 	// and start running jobs.
--- a/agent/agent_worker.go
+++ b/agent/agent_worker.go
@@ -1,6 +1,7 @@
 package agent
 
 import (
+	"expvar"
 	"fmt"
 	"strings"
 	"sync"
@@ -46,6 +47,9 @@ type AgentWorker struct {
 
 	// Tracks the last successful heartbeat and ping
 	lastPing, lastHeartbeat int64
+
+	// Metrics that the worker exposes
+	heartbeatMetrics, pingMetrics *expvar.Map
 }
 
 // Creates the agent worker and initializes it's API Client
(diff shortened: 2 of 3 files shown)


toolmantim:
For DataDog I think a good first step is "helping people debug / operate their agent cluster" — so anything that could help analyse, alert, etc would be good.

toolmantim:
I reckon it's probably worth adding some metrics for parts of the bootstrap. See if there's some machines where checkout is performing badly, etc.

lox:
> I reckon it's probably worth adding some metrics for parts of the bootstrap. See if there's some machines where checkout is performing badly, etc.

Yeah, agreed. I suspect we'll need add support for tags/dimensions to get useful info out of those, which means direct integrations with Datadog, Prometheus and Cloudwatch.

For instance, whilst we could have `bootstrap.CheckoutTimeAvg`, it's probably not useful unless you can slice it by Organization and Pipeline, or by Repository.

lox:
https://github.com/go-kit/kit/tree/master/metrics looks like a good fit for this, but we will also need the following config:

```
# Send metrics to AWS CloudWatch
metrics-cloudwatch=true

# Send metrics to DogStatsD
metrics-datadog=true
metrics-datadog-host=localhost:8125

# Make metrics available for Prometheus over a local http server
metrics-prometheus=true
metrics-prometheus-bind=:8080
metrics-prometheus-path=/prometheus

# Make metrics available over the Golang expvar http server
metrics-expvar=true
```

At that point we can use tags on each metric so it can be sliced and diced. 

Does that seem too complicated @toolmantim?

toolmantim:
I'd opt for better integration over simpler configuration… only because config is a one time hit, and the better integration pays dividends for a long time.

toolmantim:
For job-level metrics, we can easily update the agent API JSON payload to include the time it spent queued / waiting for an available agent.

patsplat:
For most jobs `jobs.Success` and `jobs.Fail` will difficult to action without differentiating btw feature branches and master.

Some agents (i.e. PROD) are expected to fail, in which case a simple metric would be useful.

toolmantim:
If we went with the tagged version, then it’d work if we tagged the org, pipeline and branch yeah?

patsplat:
Maybe?

Am honestly not deep enough on datadog to tell exactly how tags aggregate.  The published examples group hosts (i.e. region:us-east-1).

patsplat:
Am in favor of giving it a shot and seeing how it works in practice.

lox:
Closing in favour of an initial more focused exploration of Datadog on https://github.com/buildkite/agent/pull/832

PR #818, opened 2018-09-04
Update aws sdk

sj26:
Update the aws-sdk-go that we use with:

```
govendor update github.com/aws/aws-sdk-go/...
```

Mainly looks like docs updates, and some sse additions to s3.

### DIFF
--- a/vendor/github.com/aws/aws-sdk-go/CHANGELOG.md
+++ b/vendor/github.com/aws/aws-sdk-go/CHANGELOG.md
@@ -1,0 +1,1736 @@
+Release v1.10.37 (2017-09-01)
+===
+
+### Service Client Updates
+* `service/MobileHub`: Adds new service
+* `service/gamelift`: Updates service API and documentation
+  * GameLift VPC resources can be peered with any other AWS VPC. R4 memory-optimized instances now available to deploy.
+* `service/ssm`: Updates service API and documentation
+  * Adding KMS encryption support to SSM Inventory Resource Data Sync. Exposes the ClientToken parameter on SSM StartAutomationExecution to provide idempotent execution requests.
+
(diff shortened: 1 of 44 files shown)


lox:
Updating the aws sdk with govendor is a real pain.

lox:
I often end up just doing this:

```
govendor remove github.com/aws/aws-sdk-go/...
govendor fetch github.com/aws/aws-sdk-go/^
```

Note that `govendor update` will update from your $GOPATH, which isn't always what you want.

lox:
I ended up with:

```
govendor remove github.com/aws/aws-sdk-go/...
govendor fetch github.com/aws/aws-sdk-go/...
```

lox:
Really looking forward to getting the new modules stuff in to replace govendor

lox:
What was the motivation btw @sj26?

sj26:
Oh, this did more updates? Neat!

Some customers noticed that uploading zero-size artifacts to a custom s3 bucket fails with "invalid headers". I think it's omitting the `Content-Length: 0` — I vaguely recall fixing this ages ago, but I think it was in the goamz library or something. I'm hoping an sdk update might fix it. I didn't see anything in my update, but thought I'd suggest the update anyway. But sounds like I didn't update hard enough.

lox:
Interesting, I thought I'd long since fixed that issue. I remember it too.

sj26:
It does look like the s3 manager uploader has had some tweaks:

https://github.com/buildkite/agent/pull/818/commits/91bad8eb3657640a51f22a96963f05cfd231c18a#diff-65c42f8618f85e0553b496e0baa2e107R501

Maybe that'll fix it.

sj26:
Maybe they're using an old version! I'll check.,

Issue #476, opened 2017-05-19
Hot-reloadable config, or graceful agent restart?


mallyvai:
Hi BK team!
We're currently evaluating BuildKite to see if it's a good fit for our build-deploy process long-term. Right now we're in the following situation we need to account for:

* We have an error or potential security issue with the buildkite-agent.cfg config file / environment config
* We want to reload the agents with the rolled-back or updated config
* We don't want to interrupt any builds that are in-flight

Right now our only option is to force-stop / start agents en masse, but this runs the risk of killing inflight jobs.

Is there a mechanism to either

Hot-reload the configuration (buildkite-agent.cfg in our case)
Be able to issue a command to live running agents to restart and pick up latest settings/environment/etc once the current build they're running has completed?

Thank you!

jeremy:
Nice fit for blue-green deploys if you're running your builds in a cluster of containers or VMs. Roll out the cfg change by building & deploying new containers and gracefully shutting down the running instances (allowing agents to complete their current jobs).

lox:
I'm not sure that we document it very well, but you can gracefully stop an agent and it will wait for builds to finish. That said, I think either live config reload or a `USR1`sig would make a lot of sense!

uri-canva:
Note that while the agent itself supports graceful stops, the init scripts that come with it do not: the timeout before giving up and sending sigkill to the agent is very low and the systemd one defaults to sending sigterm to the whole process group, not just the agent, which means all your build scripts would have to handle sigterm as well.

lox:
Hmmm, we should definitely look at that @uri-canva!

keithpitt:
Just had a sneaky chat with @lox about this, and I think the answer for us is to support a graceful restart of the agent, which would reload any config changes if you're using a config file. We've slated this for a potential 4.0 release.

MartinNowak:
The following [systemd service drop-in](https://www.freedesktop.org/software/systemd/man/systemd.unit.html) under `/etc/systemd/system/buildkite-agent.d/timeout.conf` works fine for me to get a graceful (but guaranteed) job cancellation.
```
[Service]
TimeoutStopSec=5min
KillMode=mixed
```

PR #819, opened 2018-09-11
Allow more time for systemd graceful stop

lox:
Prompted by https://github.com/buildkite/agent/issues/476#issuecomment-417940614, this makes two changes:

* Increase `TimeoutStopSec` to 5mins from 10secs. This is the time between the initial `SIGTERM` and the final fatal `SIGKILL`. This gives the agent time to finish it's job.
* Set `KillMode` to `mixed`, vs the default of `control-group`. This means that the initial `SIGTERM` is sent to `buildkite-agent` directly, but the final `SIGKILL` to the control group. The agent [already sends shutdown signals to it's control group](https://github.com/buildkite/agent/blob/master/bootstrap/shell/signal.go#L19), so it can handle it.


toolmantim:
Sounds like a very nice change 👌🏼

PR #820, opened 2018-09-11
 Adding 32 bits version of the buildkite-agent installer

lucaspbordignon:


lox:
Sorry for the slow response on this, can you tell us a bit more about the problem you ran into that needed a 32bit binary? Perhaps rather than duplicating the script we could detect the host system's needs?

lox:
We haven't heard back @lucaspbordignon, so I'm going to close this one out for the time being.

PR #824, opened 2018-09-20
Ensure pipeline parser maintains map order in output

### LINKED ISSUE #728: Add documentation for plugin ordering
Can you please add documentation on how to ensure plugins run in a specific order.

lox:
As reported in #728, the order of plugins in a pipeline don't respect source order in some cases. What was happening here was that the order was getting lost when converting to JSON. 

This PR implements a custom JSON serializer that reads the ordered YAML parse tree. It ends up being radically simpler and the code reads a bit better I think. 

Closes #728.

### DIFF
--- a/agent/pipeline_parser_test.go
+++ b/agent/pipeline_parser_test.go
@@ -126,14 +126,14 @@ func TestPipelineParserPreservesFloats(t *testing.T) {
 	result, err := PipelineParser{Pipeline: []byte("steps:\n  - trigger: hello\n    llamas: 3.142")}.Parse()
 	assert.Nil(t, err)
 	j, err := json.Marshal(result)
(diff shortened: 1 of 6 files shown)


toolmantim:
👏🏼👏🏼👏🏼

agent/pipeline_parser_test.go
```diff
@@ -241,3 +243,40 @@ func TestPipelineParserLoadsSystemEnvironment(t *testing.T) {
 		t.Fatalf("Unexpected: %q", decoded.Steps[0].Command)
 	}
 }
+
+func TestPipelineParserPreservesOrderOfPlugins(t *testing.T) {
+	var pipeline = `---
+steps:
+  - name: ":s3: xxx"
+    command: "script/buildkite/xxx.sh"
+    plugins:
```
zsims: Awesome! Love these tests <3

Issue #763, opened 2018-05-10
Plugin concurrency issue


tessereth:
I’ve run into a concurrency bug with plugins. I have two steps, run concurrently, both using the same plugin. We also run two agents per box. If the two steps are picked up by agents on the same box, the first sees the plugin doesn’t exist and clones it. The second sees the directory exists, tries to continue but fails to find the command script in the plugin. You probably need some kind of locking or something though I haven’t looked into it too hard. If you need any additional information, just let me know.

Agent version: v3.1.1

First step:

```
# Plugin "github.com/bugcrowd/test-summary-buildkite-plugin" will be checked out to "/etc/buildkite-agent/plugins/github-com-bugcrowd-test-summary-buildkite-plugin-9637cfd5e8383996c09f6e462c4f664d1a9e612a"
$ cd /etc/buildkite-agent/plugins/github-com-bugcrowd-test-summary-buildkite-plugin-9637cfd5e8383996c09f6e462c4f664d1a9e612a
# Switching to the plugin directory
$ git clone -v -- https://github.com/bugcrowd/test-summary-buildkite-plugin .
Cloning into '.'...
POST git-upload-pack (590 bytes)
remote: Counting objects: 650, done.
remote: Compressing objects: 100% (44/44), done.
remote: Total 650 (delta 16), reused 27 (delta 7), pack-reused 597
Receiving objects: 100% (650/650), 1.30 MiB | 0 bytes/s, done.
Resolving deltas: 100% (362/362), done.
Checking connectivity... done.
# Checking out `9637cfd5e8383996c09f6e462c4f664d1a9e612a`
$ git checkout -f 9637cfd5e8383996c09f6e462c4f664d1a9e612a
Note: checking out '9637cfd5e8383996c09f6e462c4f664d1a9e612a'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by performing another checkout.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -b with the checkout command again. Example:

 git checkout -b <new-branch-name>

HEAD is now at 9637cfd... Return zero exit status on error
$ cd /

Preparing working directory	0s
$ cd /var/lib/buildkite-agent/builds/[REDACTED]/test-summary-buildkite-plugin
# Host "github.com" already in list of known hosts at "/var/lib/buildkite-agent/.ssh/known_hosts"
[...]

PR #827, opened 2018-10-05
Add plugin locking before checkout

lox:
A race condition was reported in #763 where the concurrency lock for plugin checkout is only created after the plugin directory is created. 

This moves the lock earlier, which should prevent the problem.


lox:
Pretty sure this solves the issue, so going to sneaky merge.

PR #832, opened 2018-10-10
Metrics collection for Datadog

lox:
This adds a very minimal implementation of metrics collection for the Datadog agent. 

It's controlled by two config entries:

```
metrics-datadog = true|false
metrics-datadog-host = localhost:8125
```

Currently it streams metrics in real-time as they are collected. The following metrics are collected, in the format of `[metric type], [name], [tags]`:

* [x] TIMING `buildkite.jobs.duration.success` `[agent_name, org, pipeline, branch, exit_code]`
* [x] TIMING `buildkite.jobs.duration.failure` `[agent_name, org, pipeline, branch, exit_code]`
* [x] COUNT `buildkite.jobs.success` `[agent_name, org, pipeline, branch, exit_code]`
* [x] COUNT `buildkite.jobs.failure` `[agent_name, org, pipeline, branch, exit_code]`

Batch is done, we batch 10 metrics commands presently before sending them. 

I'd love feedback on what metrics to collect and how I'm collecting them!

### DIFF
--- a/agent/agent_worker.go
+++ b/agent/agent_worker.go
@@ -1,6 +1,7 @@
 package agent
 
 import (
+	"expvar"
 	"fmt"
 	"strings"
 	"sync"
@@ -52,6 +53,9 @@ type AgentWorker struct {
 	// When this worker runs a job, we'll store an instance of the
 	// JobRunner here
 	jobRunner *JobRunner
+
+	// Metrics that the worker exposes
+	heartbeatMetrics, pingMetrics *expvar.Map
 }
 
 // Creates the agent worker and initializes it's API Client
@@ -68,6 +72,10 @@ func (a AgentWorker) Create() AgentWorker {
 		Token:        a.Agent.AccessToken,
 		DisableHTTP2: a.DisableHTTP2,
 	}.Create()
+
+	// create counters for metrics
+	a.heartbeatMetrics = expvar.NewMap("heartbeats")
+	a.pingMetrics = expvar.NewMap("pings")
 
 	return a
 }
@@ -89,6 +97,9 @@ func (a *AgentWorker) Start() error {
 			if err != nil {
 				// Get the last heartbeat time to the nearest microsecond
 				lastHeartbeat := time.Unix(atomic.LoadInt64(&a.lastPing), 0)
+
+				// Track metrics
+				a.heartbeatMetrics.Add("Fail", 1)
 
 				logger.Error("Failed to heartbeat %s. Will try again in %s. (Last successful was %v ago)",
 					err, heartbeatInterval, time.Now().Sub(lastHeartbeat))
@@ -237,6 +248,10 @@ func (a *AgentWorker) Heartbeat() error {
 	// Track a timestamp for the successful heartbeat for better errors
 	atomic.StoreInt64(&a.lastHeartbeat, time.Now().Unix())
 
+	// Track metrics
+	a.heartbeatMetrics.Add("Total", 1)
+	a.heartbeatMetrics.Add("Success", 1)
+
 	logger.Debug("Heartbeat sent at %s and received at %s", beat.SentAt, beat.ReceivedAt)
 	return nil
 }
@@ -265,10 +280,17 @@ func (a *AgentWorker) Ping() {
 			logger.Debug("[DisconnectionTimer] Reset back to %d seconds because of ping failure...", a.AgentConfiguration.DisconnectAfterJobTimeout)
 		}
 
+		// Track metrics
+		a.pingMetrics.Add("Fail", 1)
+
 		return
 	} else {
 		// Track a timestamp for the successful ping for better errors
 		atomic.StoreInt64(&a.lastPing, time.Now().Unix())
+
+		// Track metrics
+		a.pingMetrics.Add("Total", 1)
+		a.pingMetrics.Add("Success", 1)
 	}
 
 	// Should we switch endpoints?
--- a/agent/agent_worker.go
+++ b/agent/agent_worker.go
@@ -70,20 +70,12 @@ func (a AgentWorker) Create() AgentWorker {
 		endpoint = a.Endpoint
 	}
 
-<<<<<<< HEAD
 	a.APIClient = APIClient{
 		Endpoint:     endpoint,
 		Token:        a.Agent.AccessToken,
 		DisableHTTP2: a.DisableHTTP2,
 	}.Create()
 
-	// create counters for metrics
-	a.heartbeatMetrics = expvar.NewMap("heartbeats")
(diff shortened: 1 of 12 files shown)


metrics/metrics.go
```diff
@@ -0,0 +1,101 @@
...
+		if k != "" && v != "" {
+			stringSlice = append(stringSlice, formatName(k)+":"+formatName(v))
+		}
+	}
+	sort.Strings(stringSlice)
+	return stringSlice
+}
+
+// Datadog allows '.', '_' and alphas only.
+// If we don't validate this here then the datadog error logs can fill up disk really quickly
```
toolmantim: Sounds like you learnt this the hard way? 😆

toolmantim:
This more minimal approach seems much better! And those metrics seem good to me? The real test is getting some into DataDog so we can figure out some useful dashboards for the official integration.

What's the difference between a ping and a heartbeat again?

metrics/metrics.go
lox: Thankfully I copied someone else's learnings :)

lox:
I could probably hide this behind an experiment so that we could get it into the agent?

lox:
> What's the difference between a ping and a heartbeat again?

Heartbeat is just "hey, I'm here still". It runs constantly.
Ping is "is there any work?". It stops during job execution.

toolmantim:
Are you thinking the main use case for the connection ones are for alerting on the `.error` counts? I'm trying to imagine how you'd use them as graphs on a dashboard, and it's breaking my brain.

For the jobs ones, perhaps a `buildkite.jobs.started`? It might be good to visualise that by host (and then by the extra tags) to correlate agent weirdness to number of jobs running, such as memory, CPU without having to do both success + failure.

Another metric we spoke about was to help people understand how their agent hooks are performing, and probably how bits of the bootstrap is performing (as some agent hooks are there to optimise things that the bootstrap run, like a pre-checkout cache for example). Too hard to try maybe?

lox:
The problem with an experiment flag is we don't really have the technology to exclude command-line options based on experiment.

I reckon just putting it behind `metrics-buildkite: <true|false>` should be adequate!

I wonder how we'd get datadog's agent into Elastic stack 🤔

toolmantim:
Can we make the command line arg hidden @lox?

plasticine:
Could these also be exposed for scraping via something like Prom?

toolmantim:
> I wonder how we'd get datadog's agent into Elastic stack 🤔

All done! (via a custom BootstrapScript)

lox:
I ditched the API metrics around heartbeats @toolmantim. If we want at some point we can add in generic metrics around all the API methods we call, but for now lets focus on job metrics that add operational value.

lox:
Given we can test this via the bootstrap script you created before merging, shall I skip anything with the `experiment` system?

toolmantim:
Sounds good! And yep, no need for experiments (yet) I reckon.

agent/agent_worker.go
```diff
@@ -61,12 +68,22 @@ func (a AgentWorker) Create() AgentWorker {
...
 
 	a.APIClient = APIClient{Endpoint: endpoint, Token: a.Agent.AccessToken}.Create()
-
 	return a
 }
 
 // Starts the agent worker
 func (a *AgentWorker) Start() error {
+	a.metrics = a.MetricsCollector.Scope(metrics.Tags{
+		"hostname":   a.Agent.Hostname,
```
DazWorrall: On datadog you (can) pay per metric, and tag cardinality can easily explode into many distinct metrics you're not careful. The datadog agent adds some [default tags](https://docs.datadoghq.com/tagging/) - including `host` - which are free. If this feature is specially aimed at DD users then this can safely be dropped, to save some users $$. Or adding these tags could be made configurable.

I'd also like to make the `agent_name` tag optional if possible, as for us it's the same as the kubernetes pod, which is also a free/reserved tag.

toolmantim: We’re going to have a DD Buildkite integration, which means that these tags will all be free. Given that, is it still a concern?

agent/job_runner.go
```diff
@@ -190,6 +196,15 @@ func (r *JobRunner) Run() error {
 		}
 	}
 
+	// Write some metrics about the job run
+	if r.process.ExitStatus == "0" {
+		r.Metrics.Timing(`jobs.duration.success`, finishedAt.Sub(startedAt))
+		r.Metrics.Count(`jobs.success`, 1)
+	} else {
+		r.Metrics.Timing(`jobs.duration.error`, finishedAt.Sub(startedAt))
+		r.Metrics.Count(`jobs.failed`, 1)
```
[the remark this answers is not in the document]
lox: Done!

agent/agent_worker.go
lox: I've removed `hostname` @DazWorrall, thanks for the heads up.

lox:
@mipearson we're keen on events too. Will follow up with another PR after we get this one landed and things moving with Datadog.

zsims:
We're super interested in this, namely using DD to count release metrics. This is something we've been using https://github.com/seek-oss/datadog-event-buildkite-plugin for.

It'd be awesome to extend the pipeline syntax someday to add support for events, e.g. so you could record a "deployment". It goes beyond @mipearson's request as I understand it :-).

agent/agent_worker.go
```diff
@@ -63,17 +70,35 @@ func (a AgentWorker) Create() AgentWorker {
...
 		DisableHTTP2: a.DisableHTTP2,
 	}.Create()
 
+	// create counters for metrics
+	a.heartbeatMetrics = expvar.NewMap("heartbeats")
+	a.pingMetrics = expvar.NewMap("pings")
+
+=======
+	a.APIClient = APIClient{Endpoint: endpoint, Token: a.Agent.AccessToken}.Create()
+>>>>>>> Initial implementation of datadog metrics
```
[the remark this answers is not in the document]
lox: I fail at merging, clearly.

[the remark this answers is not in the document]
lox: Fixed!

yob:
We're big users of both buildkite and datadog, so we're interested in this too :heart: 

Are you aiming to become an official datadog integration?

Is there anything datadog specific in the implementation, or could it be called a statsd feature?

lox:
@yob yup, the aim is to become an official datadog integration. We'll look at some more integrations later, but the idea is to get it working well with datadog first.

agent/agent_worker.go
DazWorrall: Thanks @lox 

> Given that, is it still a concern?

Nope, that helps with a bunch of things, thanks!

PR #844, opened 2018-10-25
Update kill timeout comment to 10 seconds

raylu:
The comment was always "3 seconds" in e73bcb9f6
The timeout was bumped from 5 to 10 in 231c783a1


lox:
Thanks!

PR #845, opened 2018-10-26
Fix bug where a new chunk is created

idledaemon:
The offset and size of a new chunk are incorrect.


lox:
Interesting, what was the error or symptom you were seeing @idledaemon?

Issue #846, opened 2018-10-26
Buildkite Agent 3.5.3 fails to start due to missing config file


skidder:
## Expected Behavior
Using the `buildkite/agent:3` Docker image tag should introduce critical bug fixes present in minor releases and not break compatibility with previous minor releases. Upgrading from 3.5.2 to 3.5.3 should not require any configuration changes.

## Actual Behavior
We had been using the `buildkite/agent:3` Docker image tag to run Buildkite Docker agents for many months. The following Buildkite environment variables were set:
```
      BUILDKITE_AGENT_TOKEN
      BUILDKITE_AGENT_META_DATA
      BUILDKITE_GIT_CLEAN_FLAGS
```

Our Buildkite agents are managed with Kubernetes. On Friday, October 26, 2018 we noticed there were zero agents running, and our builds had been backlogged for hours. The following output from `kubectl` shows that the agents were in a crashloop:

```shell
$ kubectl get pods -n development | grep buildkite-std
buildkite-std-33lrj    0/1       CrashLoopBackOff   277        23h
buildkite-std-df8rc    0/1       CrashLoopBackOff   277        23h
buildkite-std-p4v7m    0/1       CrashLoopBackOff   494        1d
```

Each agent reported the following log statement:
```shell
$ kubectl logs -n development buildkite-std-p4v7m
2018-10-26 16:27:44 FATAL  A configuration file could not be found at: "/buildkite/buildkite-agent.cfg"
```

## Root Cause
A breaking change was introduced with the 3.5.3 Buildkite agent release. [Pull-Request 834](https://github.com/buildkite/agent/pull/834) introduced the `BUILDKITE_AGENT_CONFIG` environment-variable pointing to a non-existent file. This leads to the agent terminating immediately. This change broke our build & deploy pipeline for 24 hours until we reverted to the 3.5.2 agent.

## Request
Don't introduce breaking changes in minor releases of the Buildkite agent. Requiring the Buildkite agent config file be supplied to Docker containers is a pretty major change.

lox:
I'm sorry we broke things for you and your team @skidder! We absolutely had not intended the change in #834 to require a config file, it was a fix for a regression where in containers the config file should be loaded from `/buildkite/buildkite-agent.cfg` if it existed. 

We're investigating why we didn't see CI failures on our side and we'll get a fix out soon.

PR #847, opened 2018-10-27
Add buildkite-agent.cfg to docker images

lox:
This adds `/buildkite/buildkite-agent.cfg` to our docker images, and also example hooks.

Closes https://github.com/buildkite/agent/issues/846.

### DIFF
--- a/.buildkite/steps/build-docker-image.sh
+++ b/.buildkite/steps/build-docker-image.sh
@@ -23,6 +23,7 @@ build_docker_image() {
   local packaging_dir="$2"
 
   echo "--- Building :docker: $image_tag"
+  cp -a packaging/linux/root/usr/share/buildkite-agent/hooks/ "${packaging_dir}/hooks/"
   cp pkg/buildkite-agent-linux-amd64 "${packaging_dir}/buildkite-agent"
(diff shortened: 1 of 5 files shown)


.buildkite/steps/build-docker-image.sh
```diff
@@ -23,6 +23,7 @@ build_docker_image() {
   local packaging_dir="$2"
 
   echo "--- Building :docker: $image_tag"
+  cp -a packaging/linux/root/usr/share/buildkite-agent/hooks/ "${packaging_dir}/hooks/"
```
toolmantim: I wonder if people look in the actual image for example hooks, or the Docker directories in GitHub? I wondered if the real problem was a lack of line to the standard example hooks from places like the Docker agent setup docs.

lox: Yeah, dunno, I've spoken to one human that did look for the example hooks there, so thought I'd do this. I can remove it if it's controversial.

toolmantim: Yep, I remembered. I just wasn’t sure where they exactly went looking.

PR #850, opened 2018-10-31
Adds a variable for tracking environment variable changes

zsims:
This is a rough attempt (I'd like to add many tests around this, but there's no examples in `bootstrap_test.go` :() to add an environment variable that tracks what environment variables have been updated.

This is useful for propagating hook variables into docker containers per buildkite-plugins/docker-buildkite-plugin#80

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -22,6 +22,10 @@ import (
 	"github.com/pkg/errors"
 )
 
+const (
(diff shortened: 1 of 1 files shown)


zsims:
Makes sense. Happy to percolate ☕️

pda:
I think this might have over-infused 😬
I'm going to close it due to inactivity; feel free to bring it up again if it's still of interest 🙏🏼

PR #851, opened 2018-11-01
Add a flag to disable http2 in the start command

lox:
This adds a flag that lets us optionally disable http2 for communicating with the Agent API. This is useful for proxies that don't support http2.


toolmantim:
We gotta update the docs too

PR #853, opened 2018-11-09
Better cancel signal handling

lox:
Currently we have a bug where if checkout is interrupted with a cancel, it gets retried. Cancellation sends a `SIGKILL` if the bootstrap doesn't finish in 10 seconds, which means that often this happens and results in confusing errors. 

This fixes that issue, as well as clearing up how cancellation is handled. Previously signal handling was done in the Shell abstraction. Prior to each subcommand being executed, signal handling was installed and then removed after. This PR moves signal handling to the top-level of the Bootstrap clicommand and passes cancellation down to the command-level with Golang's `context`'s. 

This also shows a clearer error around cancellation. Where it was previously: `The command exited with status -1`, it is now `The command was cancelled`.


sj26:
Amazing!

Does this correctly collect exit statuses when the command does exit within 10 seconds?

lox:
Good question, I will check.

toolmantim:
I might have sent us on a goose chase thinking this was just about cancelled jobs. I think the fix I was hoping for was to never print "Exited with status -1" in the log output anywhere — because `-1` is used for people to handle lost agents, and makes things confusing. But I think we might print `-1` in a whole bunch of scenarios, not related to canceled jobs?

lox:
Yup, I am gonna reconsider this one and make a simpler change @toolmantim.

PR #867, opened 2018-12-10
Add another search path for config file on Windows

petemounce:
... because I'm OCD enough that `BuildkiteAgent` struck me as inconsistent with `buildkite-agent` of most other usage sites. The user is kebab-case, the other search paths on other platforms are kebab-case, I couldn't think of a reason for Windows to be snowflake.


lox:
👏🏻

Issue #456, opened 2017-04-19
Using BUILDKITE_COMMIT in pipeline.yml files


toolmantim:
It's common to use `BUILDKITE_COMMIT` in a pipeline.yml file, but with builds that start with using `HEAD` this isn't possible if the `buildkite-agent pipeline upload` is the first step in the pipeline, because although the agent has calculated what the actual git commit is, it hasn't updated the `BUILDKITE_COMMIT` env before the command is run.

For example, this echo:

<img width="237" alt="echo" src="https://cloud.githubusercontent.com/assets/153/25158505/5fa878bc-24ec-11e7-91f9-17c704fdabba.png">

outputs:

<img width="817" alt="untitled 4" src="https://cloud.githubusercontent.com/assets/153/25158430/b27337c2-24eb-11e7-8586-7b623557f3ec.png">

Another example, a [this pipeline.yml example](https://gist.github.com/toolmantim/a5e31805a44e1a6381882658d461dbbb):

```yml
steps:
  - command: echo "$BUILDKITE_COMMIT"
```

<img width="827" alt="pipeline-example" src="https://cloud.githubusercontent.com/assets/153/25158481/12c3bb88-24ec-11e7-99b1-e84f4bc79662.png">

It be great if the buildkite-agent updated the `BUILDKITE_COMMIT` env, as soon as it became aware of the true value in the "Preparing build" step.

I had a quick look at implementing this a few months back, but it looks like it just sends the entire commit up to the agents API. Perhaps the Agents API could return the parsed commit info, and then the agent can update it's local env values?

lox:
Unfortunately this would require parsing the git commit locally in the agent, which ends up being pretty involved. Let's look at this for a future version!

reidab:
For the purposes of interpolating the current commit into pipeline.yml, I've found just changing the pipeline upload step to `GIT_COMMIT=$(git rev-parse HEAD) buildkite-agent pipeline upload` and then referencing `$GIT_COMMIT` in pipeline.yml seems to work fine. 

Overriding `BUILDKITE_COMMIT` in the same way seems like it would work as well.

nullobject:
This issue just bit me too. It would be awesome if `BUILDKITE_COMMIT` was what you expect it to be 😄

nullobject:
I totally forgot about this and it stung us again. Any ETA for a real fix?

lox:
Apologies @nullobject, no ETA yet. It's an annoying issue that we don't have an easy fix for.

[...]

PR #871, opened 2018-12-14
Resolve BUILDKITE_COMMIT before pipeline upload

lox:
This resolves `BUILDKITE_COMMIT` (which can be `HEAD` or `FETCH_HEAD`) to an actual commit before interpolation into a pipeline that is being uploaded.

Closes https://github.com/buildkite/agent/issues/456.

### DIFF
--- a/clicommand/pipeline_upload.go
+++ b/clicommand/pipeline_upload.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"io/ioutil"
 	"os"
+	"os/exec"
 	"path"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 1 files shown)


lox:
For historical context, this would be better placed at the bootstrap level so that anything in the job gets the resolved value for `$BUILDKITE_COMMIT`, but our concern was that might be too much of a breaking change. This hacky fix at least prevents people running into the issue and we can re-evaluate when we move to 4.0.0.

nullobject:
Ace, thank you. What a lovely xmas present 🎅❤️

keith:
It seems like this case may still exist if you're using a scheduled job that sets HEAD as the commit? In that case I still see BUILDKITE_COMMIT being set as HEAD

PR #877, opened 2018-12-17
Allow hooks to change git repo url

sj26:
Hooks should be able to export `$BUILDKITE_REPO` to change the git clone/fetch uri. The most compelling use case is for local mirrors of git repos, for caching purposes.

### DIFF
--- a/bootstrap/config.go
+++ b/bootstrap/config.go
@@ -23,7 +23,7 @@ type Config struct {
 	Debug bool
 
 	// The repository that needs to be cloned
-	Repository string
+	Repository string `env:"BUILDKITE_REPO"`
 
 	// The commit being built
 	Commit string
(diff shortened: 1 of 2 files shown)


lox:
My only concern would be the security implications. Would this allow BUILDKITE_REPO to be overridden from pipeline env? That would be worrying.

sj26:
I don't think so. `BUILDKITE_REPO` is always set to `project.repository` [during job creation](https://github.com/buildkite/buildkite/blob/master/app/models/job/environment.rb#L106) and cannot be overridden by user-supplied env in pipeline settings or yaml. It should only be modifiable in hooks, and only hooks earlier than checkout would have any effect.

sj26:
@lox ooh, that sounds interesting. Perhaps another PR?

harrietgrace:
@sj26  this is the docs we have at the moment for `BUILDKITE_GIT_REPO`:

image

Is there anything else we need to mention, or does the bit about exporting in hooks cover it?

sj26:
@harrietgrace that sounds right to me! :+1:

toolmantim:
@plaindocs and I just discovered that `BUILDKITE_CLEAN_CHECKOUT` isn't able to be changed in an env hook either, contrary to the docs 🤔

PR #878, opened 2018-12-18
Support vendored plugins

lox:
This change allows plugin paths to be relative to the checkout, so you can commit plugins along side code. The thinking is this provides a straight-forward way to lock-down what plugins are used and it has the nice upside of making it easier to iterate on plugins and code that are evolving. 

```yaml
steps:
  - label: "Test with vendored docker plugin"
    plugins:
      - "./.buildkite/plugins/docker":
          image: "hello-world"
```

Some more context on the background is at https://forum.buildkite.community/t/verifying-the-integrity-of-third-party-plugins/158/18. 

In terms of implementation, this adds an additional phase `vendored-plugin` that occurs in the bootstrap after `checkout`. This means that vendored plugins can't have `pre-checkout` or `checkout` hooks and that vendored plugins `environment` hook is called after checkout. I'm pretty ok with those constraints.

Future work could be a way to only allow vendored plugins (as a compromise on the stricter `--no-plugins`) setting we recommend. 🤔

### DIFF
--- a/agent/plugin/plugin.go
+++ b/agent/plugin/plugin.go
@@ -25,11 +25,17 @@ type Plugin struct {
 	// Any authentication attached to the repostiory
 	Authentication string
 
+	// Whether the plugin refers to a vendored path
+	Vendored bool
+
 	// Configuration for the plugin
 	Configuration map[string]interface{}
 }
 
-var locationSchemeRegex = regexp.MustCompile(`^[a-z\+]+://`)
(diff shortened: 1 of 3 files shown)


lox:
This would be really neat to pair with some `bk` commands for updating vendored plugins!

joscha:
fixes https://github.com/buildkite/feedback/issues/363

PR #879, opened 2018-12-20
Use windows console groups for process management

lox:
It turns out Windows has no concept of parent/child processes, beyond storing the PID of the process that created it. This means that trying to walk a process tree to cancel a job can leave sub-processes orphaned if a subprocess in the hierarchy dies, as you can't walk that tree past the dead process. For instance in:

`bash.exe (pid 1)` -> `bash.exe (pid 2)` -> `sleep.exe (pid 3)`

If pid 2 dies, pid 3 remains, but killing pid 1 won't be able to traverse down to pid 3.  

The best we can do is create processes inside a "console group" and then send break / ctrl-c events to that group. 

Context:
https://github.com/buildkite/agent/pull/795
https://github.com/golang/dep/issues/862
https://github.com/golang/go/issues/17608
https://github.com/golang/go/issues/6720
https://docs.microsoft.com/en-us/windows/console/generateconsolectrlevent
https://github.com/nodejs/node/issues/3617

### DIFF
--- a/process/process_test.go
+++ b/process/process_test.go
@@ -5,7 +5,6 @@ import (
 	"os"
 	"os/signal"
 	"reflect"
-	"runtime"
 	"strings"
 	"sync"
 	"sync/atomic"
@@ -93,10 +92,6 @@ func TestProcessCapturesOutputLineByLine(t *testing.T) {
 }
 
 func TestProcessInterrupts(t *testing.T) {
-	if runtime.GOOS == `windows` {
-		t.Skip("Not supported on windows")
-	}
-  
 	var lines []string
 	var mu sync.Mutex
 
--- a/process/process_test.go
+++ b/process/process_test.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"os/signal"
 	"reflect"
+	"runtime"
 	"strings"
 	"sync"
 	"sync/atomic"
@@ -92,6 +93,10 @@ func TestProcessCapturesOutputLineByLine(t *testing.T) {
 }
 
 func TestProcessInterrupts(t *testing.T) {
+	if runtime.GOOS == `windows` {
+		t.Skip("Works in windows, but not in docker")
+	}
+
 	var lines []string
 	var mu sync.Mutex
 
--- a/process/utils_windows.go
+++ b/process/utils_windows.go
@@ -19,7 +19,6 @@ import (
 
 var (
 	libkernel32                  = syscall.MustLoadDLL("kernel32")
-	procSetConsoleCtrlHandler    = libkernel32.MustFindProc("SetConsoleCtrlHandler")
 	procGenerateConsoleCtrlEvent = libkernel32.MustFindProc("GenerateConsoleCtrlEvent")
 )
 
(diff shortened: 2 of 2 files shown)


amitsaha:
Hi @lox - thanks for looping me in. This definitely sounds like an improvement. I can do some testing on this as well, however that will have to wait till early January. I am keen to see how it plays with docker containers  on Windows server on our buildkite agents.

lox:
Those are very good questions @amitsaha! I hadn't thought about docker interactions.

amitsaha:
Right now, the behavior that we see is that a job cancellation leaves the docker containers running and then we use a buildkite agent hook to kill any running containers before it starts a build (which spawns new containers).

I wonder if docker containers spawned from a process belong to the same console group - 'cause if that's the case, then we could improve the current situation by a lot!

lox:
@amitsaha I'd guess that is because under windows the bootstrap exits immediately on signal, so it doesn't have time to clean up. Whilst this PR doesn't do it, it would open the door of our bootstrap to catch the `ctrl_c` or `break` and run the `pre-exit` hooks. Exciting possibility if there aren't any major compatibility issues here.

lox:
Really good answer on how signal handling works in windows: https://stackoverflow.com/questions/35772001/how-to-handle-the-signal-in-python-on-windows-machine

lox:
So in Docker this results in an error for the Interrupt, but it will fall back to using `TASKKILL.EXE`. I think this is still a good improvement for folks that don't run the agent in windows docker. 

https://github.com/docker/for-win/issues/3173

amitsaha:
> So in Docker this results in an error for the Interrupt, but it will fall back to using `TASKKILL.EXE`. I think this is still a good improvement for folks that don't run the agent in windows docker.
> 
> [docker/for-win#3173](https://github.com/docker/for-win/issues/3173)

that is interesting. My current use-case is slightly different:

1. BK agent begins executing new job via PS script
2. PS Script spawns docker containers
3. ..


Job is manually cancelled. The spawned docker containers stay running. To clean up, we use a agent hook before the start of the next job to cleanup any running containers.

So, now after this change, I was hoping that since the new containers that are spawned by my PS script may be part of the same console group and hence the CTRL_BREAK_EVENT would also kill the containers. That however doesn't seem to be the case. I have a [demo](https://github.com/amitsaha/golang-windows-demo/tree/master/console-group-demo) here. Any thoughts?

lox:
@amitsaha unfortunately docker containers don't inherit a console group. That would be too sensible a thing for windows to do. This does pave the way for windows things to run pre-exit scripts on cancellation though, which would mean you could use pre-exit hooks to cleanup like we do in the [docker-compose](https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/blob/master/hooks/pre-exit) plugin.

amitsaha:
> That would be too sensible a thing for windows to do. 

Heh.

> This does pave the way for windows things to run pre-exit scripts on cancellation though, which would mean you could use pre-exit hooks

Can you please explain this a bit more? Is the process group rooted at the bootstrap process? And if that's the case, then we can set up a signal handler in the bootstrap process to run the pre-exit handlers. If that's the case, is the change already in or would need to be done?

mattn:
FYI, if you want to terminate process tree forcibly, you can use CreateJobObject/TerminateJobObject.

PR #880, opened 2018-12-21
Give all the things a logger instance

lox:
With the rise of our new `--spawn` param, I'd really like to have a prefix in the log output for the specific agent that console output is coming from. This is really, really hard to do with the global logger.

This moves to an instance that gets passed into anything that needs a logger. It's a lot of code and opens up a lot of spots for nil pointer errors as we rarely use constructors and tend to use direct struct instantiations. 

I'd love some feels on this approach. It's been hard work and I don't super like the result, but it's probably more correct?

### DIFF
--- a/agent/agent_pool.go
+++ b/agent/agent_pool.go
@@ -19,6 +19,7 @@ import (
 )
 
 type AgentPool struct {
+	Logger                logger.Logger
 	APIClient             *api.Client
 	Token                 string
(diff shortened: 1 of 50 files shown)


lox:
Yeah, I considered a lot of other options that would be more magical, but ultimately I think simply passing in the logger instance to places where it's needed is the most "go-like".

PR #882, opened 2018-12-29
Move graceful kill behaviour out of process.Process

lox:
Previously `process.Process.Kill()` handled our graceful shutdown behaviour, e.g first interrupt the sub-process, wait 10 seconds and then kill it. 

It turns out that windows support is a lot more complicated, so this moves that functionality to `agent.JobRunner.Cancel()`, which delegates to `process.Process.Interrupt()` and then `process.Process.Terminate()`.

This should be functionally identical to the current behaviour, just with some code moved around to better support future changes like #879.

### DIFF
--- a/process/utils.go
+++ b/process/utils.go
@@ -26,5 +26,7 @@ func terminateProcess(p *os.Process) error {
 
 func interruptProcess(p *os.Process) error {
 	logger.Debug("[Process] Sending signal SIGTERM to PID: %d", p.Pid)
+
+	// TODO: this should be SIGINT, but will be a breaking change
 	return p.Signal(syscall.SIGTERM)
 }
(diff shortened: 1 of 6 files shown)


process/utils.go
```diff
@@ -5,10 +5,26 @@ package process
...
+}
+
+func terminateProcess(p *os.Process) error {
+	logger.Debug("[Process] Sending signal SIGKILL to PID: %d", p.Pid)
+	return p.Signal(syscall.SIGKILL)
+}
+
+func interruptProcess(p *os.Process) error {
+	logger.Debug("[Process] Sending signal SIGTERM to PID: %d", p.Pid)
+	return p.Signal(syscall.SIGTERM)
```
lox: It's weird that this is `SIGTERM` and not `SIGKILL`, but it's how it is currently and I was reluctant to change it here.

Issue #866, opened 2018-12-06
Repeated `Warning: Checkout failed` errors while preparing working directory


filipesilva:
Heya,

We've recently setup a windows pipeline (https://buildkite.com/angular/angular) and have observed some odd errors (log [1](https://buildkite.com/angular/angular/builds/70#9fc8283b-7088-4aee-9582-53166c9a8d3d], [2](https://buildkite.com/angular/angular/builds/96#65a95d9f-7677-4d22-b794-d56b985463aa), [3](https://buildkite.com/angular/angular/builds/67#f8dc2fab-07e5-489e-8be4-f7a64d1099c9), [3](https://buildkite.com/angular/angular/builds/97#49ffb1e8-0930-453d-ae7a-439437201014)). The pipeline is public so anyone should be able to see the builds.

All of these fail in the `Preparing working directory` in a similar pattern: somehow the folder is no longer a git repository and the retries fail:
```
[2018-12-06T07:58:04Z] > cd C:\buildkite-agent\builds\gce-buildkite-windows-1-3\angular\angular
--
  | [2018-12-06T07:58:04Z] > git remote set-url origin https://github.com/angular/angular
  | [2018-12-06T07:58:04Z] fatal: not a git repository (or any of the parent directories): .git
  | [2018-12-06T07:58:04Z] # Removing C:\buildkite-agent\builds\gce-buildkite-windows-1-3\angular\angular
  | [2018-12-06T07:58:04Z] ⚠️ Warning: Checkout failed! Error running `C:\git\cmd\git.exe remote set-url origin https://github.com/angular/angular`: exit status 128 (Attempt 1/3 Retrying in 2s)
  | [2018-12-06T07:58:06Z] > git remote set-url origin https://github.com/angular/angular
  | [2018-12-06T07:58:06Z] fatal: not a git repository (or any of the parent directories): .git
  | [2018-12-06T07:58:06Z] # Removing C:\buildkite-agent\builds\gce-buildkite-windows-1-3\angular\angular
  | [2018-12-06T07:58:06Z] ⚠️ Warning: Checkout failed! Error running `C:\git\cmd\git.exe remote set-url origin https://github.com/angular/angular`: exit status 128 (Attempt 2/3 Retrying in 2s)
  | [2018-12-06T07:58:08Z] > git clone --config core.autocrlf=input --config core.eol=lf --config core.longpaths=true --config core.symlinks=true -- https://github.com/angular/angular .
[...]

PR #887, opened 2019-01-09
Acquire a lock on checkout dir

lox:
Sometimes when we cancel jobs under windows, processes are left over that might still be accessing the checkout dir. This leads to errors when the next job runs `git clean`, as I suspect we are seeing in #866. 

This PR adds a cross-process lock to the checkout dir to avoid concurrent access. Hopefully it will make for a clearer error than we are presently seeing.

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -886,6 +886,14 @@ func (b *Bootstrap) defaultCheckoutPhase() error {
 		return err
 	}
 
+	// Try and lock the directory in-case any processes remain using the directory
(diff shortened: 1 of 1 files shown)


keithpitt:
I think we’d be doing a similar thing when we switch to single repo agents? We’d be locking repos while one of the agents is doing work on it. Seems legit to me..? Maybe we just need some better logging, or we can be smarter about the lock. Can we see if the process owner of the lock is still around, and do something there?

lox:
Yeah, makes me nervous too. I'm gonna not do this for now.

PR #888, opened 2019-01-09
Fix for #475

### LINKED ISSUE #475: Raspberry Pi architecture is not detected correctly
Hi there,

I wanted to install a Buildkite agent on my Raspberry Pi. I thought I could just use the install script that you guys provide for Linux, but when I started the agent after that, it exited directly with the message `Illegal instruction`.
Since `uname -m` told me my architecture is `armv6l`, I tried manually downloading `buildkite-agent-linux-armhf-2.6.3.tar.gz` from the releases page on GitHub. Same thing happened: `Illegal instruction`.
Then I tried installing the `buildkite-agent-linux-arm-2.6.3.tar.gz` which, to my surprise, finally worked. My Buildkite agent is now happily "Waiting for work...".

TL;DR: Architecture `armv6l` should install the `arm` version instead of the `armhf` version.

ryankurte:
The RPi reports `uname -sm` as `Linux armv6l`, this PR adds a match for `armv6l` for checking prior to `armv6` to resolve arch to `arm` instead of `armhf` in `install.sh`.

I'm not sure this is the only way to fix it / there might be other things wrong / still needs to be tested.

Closes #475.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -39,6 +39,7 @@ else
     *x86_64*)  ARCH="amd64"   ;;
     *armv8*)   ARCH="arm64"   ;;
     *armv7*)   ARCH="armhf"   ;;
+    *armv6l*)  ARCH="arm"     ;;
     *armv6*)   ARCH="armhf"   ;;
     *arm*)     ARCH="arm"     ;;
     *ppc64le*) ARCH="ppc64le" ;;


lox:
Looks good to me! Thanks very much @ryankurte.

PR #889, opened 2019-01-10
Resolve ssh hostname aliases before running ssh-keyscan

ticky:
This is a PoC to improve compatibility with ssh aliases which don’t conform to our special-casing behaviour.

For any incoming ssh host, this asks `ssh` for what host it will connect to. This allows `ssh-keyscan` to connect to that host, rather than whatever the alias name is, which may or may not match.

It also supersedes the older behaviour wherein our special-case aliases were stripped, and on agents which are already correctly configured, should work the same.

~I haven’t been able to get the test suite working to confirm this works, but I figured putting the sketch up would be better than doing nothing! 😃~

I’ve updated this to pass, and add new, tests, and it’s looking good!

For reference, `ssh -G` appears to have been added in OpenSSH in 6.0 (released April 22, 2012, sadly no source link as this took gripping multiple tgz files to bisect), but it appears to be turned off in some OS’ distributions of 6.x (one of our virtual servers is running a version from 2014, but it’s missing) so the existence of the fallback remains potentially important.

### DIFF
--- a/bootstrap/git.go
+++ b/bootstrap/git.go
@@ -149,6 +149,32 @@ func parseGittableURL(ref string) (*url.URL, error) {
 // https://buildkite.com/docs/agent/ssh-keys#creating-multiple-ssh-keys
 var gitHostAliasRegexp = regexp.MustCompile(`-[a-z0-9\-]+$`)
 
-func stripAliasesFromGitHost(host string) string {
+func resolveGitHost(host string) string {
+	// ask SSH to print its configuration for this host, honouring .ssh/config
+	output, err := sh.RunAndCapture("ssh", "-G", host)
+
+	// if we got no error, let's process the output
+	if (err == nil) {
+		// split up the ssh -G output
+		lines := strings.Split(output, "\n")
+
+		// search the ssh -G output for a "hostname" line
+		for _, line := range lines {
+			tokens := strings.SplitN(line, " ", 2)
+
+			// skip any line which isn't a key-value pair
+			if len(tokens) != 2 {
+				break
+			}
+
+			if tokens[0] == "hostname" {
+				return tokens[1]
+			}
+		}
+	}
+
+	// if we got here, either the `-G` flag is unsupported, or ssh -G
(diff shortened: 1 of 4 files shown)


bootstrap/git.go
```diff
@@ -149,6 +149,32 @@ func parseGittableURL(ref string) (*url.URL, error) {
 // https://buildkite.com/docs/agent/ssh-keys#creating-multiple-ssh-keys
 var gitHostAliasRegexp = regexp.MustCompile(`-[a-z0-9\-]+$`)
 
-func stripAliasesFromGitHost(host string) string {
+func resolveGitHost(host string) string {
```
lox:
```suggestion
func resolveGitHost(sh *shell.Shell, host string) string {
```

bootstrap/knownhosts.go
```diff
@@ -142,7 +142,7 @@ func (kh *knownHosts) AddFromRepository(repository string) error {
 		return nil
 	}
 
-	host := stripAliasesFromGitHost(u.Host)
+	host := resolveGitHost(u.Host)
```
lox:
```suggestion
	host := resolveGitHost(kh.Shell, u.Host)
```

lox: I think that will work!

lox:
This looks awesome @ticky! I was wondering when that github alias stuff would bite us.

lox:
Looks good to go to me @ticky!

bootstrap/git.go
```diff
@@ -149,6 +149,54 @@ func parseGittableURL(ref string) (*url.URL, error) {
...
+		// if the port is the default, we can leave it off
+		if port == "22" {
+			return hostname
+		}
+
+		// otherwise, output it in hostname:port form
+		return fmt.Sprintf("%s:%s", hostname, port)
+	}
+
+	// if we got here, either the `-G` flag was unsupported, or ssh -G
```
ticky: I wonder if we should emit a warning here… 🧐

PR #890, opened 2019-01-10
Terminate bootstrap with unhandled signal after cancel

lox:
Currently the bootstrap will exit with whatever exit code it gets from the subprocesses it kills after cancellation. Following bash's example, it now removes its signal handler and SIGTERM's itself (arguably, it should use the signal it received, but this was easier initially). This shows up as consistently as a -1 in the UI. 

image

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -1,6 +1,8 @@
 package bootstrap
 
 import (
+	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -11,8 +13,6 @@ import (
 	"strconv"
 	"strings"
 	"time"
-	"encoding/json"
-	"context"
 
 	"github.com/buildkite/agent/agent/plugin"
 	"github.com/buildkite/agent/bootstrap/shell"
@@ -483,7 +483,7 @@ func (b *Bootstrap) loadPlugins() ([]*plugin.Plugin, error) {
 	return b.plugins, nil
 }
 
-func (b *Bootstrap) validatePluginCheckout(checkout *pluginCheckout)  error {
+func (b *Bootstrap) validatePluginCheckout(checkout *pluginCheckout) error {
 	if !b.Config.PluginValidation {
 		return nil
 	}
@@ -601,9 +601,9 @@ func (b *Bootstrap) VendoredPluginPhase() error {
 		}
 
 		checkout := &pluginCheckout{
-			Plugin: p,
+			Plugin:      p,
 			CheckoutDir: pluginLocation,
-			HooksDir: filepath.Join(pluginLocation, "hooks"),
+			HooksDir:    filepath.Join(pluginLocation, "hooks"),
 		}
 
 		// Also make sure that plugin is withing this repository
--- a/clicommand/bootstrap.go
+++ b/clicommand/bootstrap.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"os/signal"
 	"runtime"
+	"sync/atomic"
 	"syscall"
 
 	"github.com/buildkite/agent/bootstrap"
@@ -340,13 +341,32 @@ var BootstrapCommand = cli.Command{
 			syscall.SIGQUIT)
 		defer signal.Stop(signals)
 
+		var cancelled uint32
+
 		go func() {
 			for _ = range signals {
 				bootstrap.Cancel()
+				atomic.StoreUint32(&cancelled, 1)
 			}
 		}()
 
-		// Run the bootstrap and exit with whatever it returns
-		os.Exit(bootstrap.Start())
+		// Run the bootstrap and get the exit code
+		exitCode := bootstrap.Start()
+
+		// Cancelled with a non-zero should terminate ourselves so that
+		// the parent process gets a -1 exit code
+		if atomic.LoadUint32(&cancelled) == 1 {
+			p, err := os.FindProcess(os.Getpid())
+			if err != nil {
+				l.Error("Failed to find current process: %v", err)
+			}
+			l.Debug("Terminating bootstrap after cancellation")
+			err = p.Signal(syscall.SIGKILL)
+			if err != nil {
+				l.Error("Failed to signal self: %v", err)
+			}
+		}
+
+		os.Exit(exitCode)
 	},
 }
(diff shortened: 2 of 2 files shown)


toolmantim:
I might be mistaken, but do we want -1 in the UI? It’s currently reserved for “agent lost” situations?

lox:
If the agent gets signalled and shuts down, wouldn't that be "agent lost" @toolmantim? If not, what should the exit code be?

toolmantim:
Ah, I thought this was for when you cancel a build (e.g. somebody hits "cancel" in the web UI). If it's for spot termination signal handling situations, totally makes sense!

lox:
What do you think it should show for UI cancelled jobs @toolmantim? Presently I think it's `Exited with status 255`, which is basically a bug, IMO.

lox:
This is presently (not with this PR) what UI cancellation looks like:

image

It looks exactly the same when the host shuts down presently too. Been trying to figure out how to distinguish between them in terms of UI.

lox:
This PR just basically fixes the bug where the bootstrap -1 propagates to the UI as 255. It still leaves the problem of no way to differentiate between user cancellation and host shutdown, which is what I set out to solve initially! 😅

lox:
What does “agent lost” cover? For instance, it you cancel a job and it takes more than 10 seconds and we kill it, is that “agent lost”?

toolmantim:
I believe it’s “no comms at all from the agent after x minutes”. I don’t think it’s meant for “process didn’t respond” in any way.

matthewd:
> If you're authoring/debugging scripts that are meant to explicitly handle cancellation signals, I wonder what you'd need here? You'd probably want to see the signal that was sent to the bootstrap to trigger the cancellation?

I see that as the next step: once we have the bootstrap "properly" passing the true exit details up to the agent-proper, then we can teach the agent to pass it along in turn.

lox:
> I believe it’s “no comms at all from the agent after x minutes”. I don’t think it’s meant for “job/command didn’t respond” in any way.

What situations would that occur in then? Basically when the agent gets hard terminated when a machine is switched off, or a crash or something?

toolmantim:
You’re probably right! Or maybe we need a new -2 state? (“Command not responding” or similar)

Would it help if we had a table of the different states/scenarios, and what the agent/bootstrap should signal between each other, and what the UI should do?

Would the operating system shutdown eventually be a SIGKILL after a timeout waiting after a SIGTERM? SIGKILL’s would show as -1, but a successful SIGTERM and graceful shutdown (which would also run the pre-exit hook yeah) I imagine would just show a SIGTERM exit status? Or perhaps “-2 (Command not responding)”?

lox:
| Scenario | Exit Code reported |
| --- | --- | 
| Job runs and completes | Exit code of command |
| Job runs and completes, but a hook fails | Exit code of hook |
| User cancels a job and command terminates in the grace period with a non-zero exit code | ? |
| User cancels a job and command terminates in the grace period with zero exit code | 0 |
| User cancels a job and command takes too long and is killed with SIGKILL  | ? |
| Host running agent shuts down (SIGTERM), and command terminates inside grace period with a non-zero exit code | ? |
| Host running agent shuts down (SIGTERM), and command terminates inside grace period with a zero exit code | 0 |
| Host running agent shuts down (SIGTERM), and command doesn't terminates inside grace period and is killed with SIGKILL | ? |
| Agent is hard killed (SIGKILL) | -1 |
| Job runs, but exceeds agent-side timeout value and is killed with SIGKILL | ? |

lox:
I think we probably need more info than just the exit code in the UI. Something like a `reason` parameter, along with a `signal` code, so you'd get something like:

On ctrl-c of the agent: `exitCode: 1, reason: 'STOPPED', signal: SIGINT`

Some interesting ideas on this front in bosh's `${.sh.code}`: http://schillix.sourceforge.net/man/man1/bosh.1.html

toolmantim:
I have a quick question about this one:

> User cancels a job and command terminates with an exit code of -1 (unhandled signal)

I've never been able to create this scenario (but that could be because of a bash interpretation of `$?`). How does this situation happen?

```
 $ bash -ec 'exit -1'; echo "Exited with $?"
Exited with 255
$ bash -ec 'sleep 30'; echo "Exited with $?"
^C
Exited with 130
$ bash -ec 'while :; do printf "."; done'; echo "Exited with $?"
.............^C
Exited with 130
$ bash -ec 'exit -256'; echo "Exited with $?"
Exited with 0
$ node -e 'process.exit(-1)'; echo "Exited with $?"
Exited with 255
```

lox:
@toolmantim excellent question! It's not actually a thing. When bash has a child process exit via an unhandled signal it does the following:

> When a command terminates on a fatal signal whose number is N, Bash uses the value 128+N as the exit status.

See https://www.gnu.org/software/bash/manual/html_node/Exit-Status.html and https://www.tldp.org/LDP/abs/html/exitcodes.html. 

With that context, specific answers to each of your examples:

```
 $ bash -ec 'exit -1'; echo "Exited with $?"
Exited with 255
```

Exiting with -1 isn't a thing either, you are seeing 255 because that is what is used for `Exit status out of range`. Exit codes must be between 0-255 in bash.

```
$ bash -ec 'while :; do printf "."; done'; echo "Exited with $?"
.............^C
Exited with 130
```

When you ctrl-c, it sends SIGINT (2) to the process. It terminates because it has no signal handler and returns 128 + 2 (130). We could replicate this behaviour if we were so inclined, and I think it probably makes sense. 

```
$ bash -ec 'exit -256'; echo "Exited with $?"
Exited with 0
```

🤷🏼‍♂️ I'd have expected this to be 255, but perhaps the exit command itself failed and then it ran the echo successfully?"

```
$ node -e 'process.exit(-1)'; echo "Exited with $?"
Exited with 255
```

Same as the first example!

Also check out https://www.vidarholen.net/contents/blog/?p=34!

lox:
The reason why we see -1 is that our golang code uses -1 in the absence of an exit code. A better way to phrase it might be:

> User cancels a job and command is interrupted and terminates without an exit code (unhandled signal)

I reckon we should probably just follow bash's convention and exit with 128+signal.

harrietgrace:
The signal handling section mentions the SIGTERM on cancel, so I'm going to remove the docs update label from this. @lox if you wanted more docs about this, let me know! 

image

PR #894, opened 2019-01-11
Move struct initializations to constructors

lox:
This is a fairly large refactor to move from initializing structs manually to using constructors. The upside is the we get type-safety on required parameters. This also makes it clear what things are needed to create a struct vs internal state.


lox:
Ok this is done. My brain hurts.

PR #881, opened 2018-12-21
Add tests exposing a paths issue.

petemounce:
On windows, absolute paths start either with \ or with c: (as in, drive-letter then colon)

PR #895, opened 2019-01-13
Correct tests introduced in #881

lox:
@petemounce I renamed the test so that it would run as a test.

I wasn't sure what you intended with the `filepath.Join`, are you saying that `C:\programdata\buildkite-agent` should normalize to `C:programdata\buildkite-agent`? I know either works in windows, but I wasn't aware of a reason the latter should be preferred. 

See https://github.com/golang/go/issues/11551 for some context on the weird ways that golang handles file paths.

### DIFF
--- a/utils/path_windows_test.go
+++ b/utils/path_windows_test.go
@@ -13,10 +13,9 @@ import (
 func TestNormalizeWindowsDriveAbsolutePath(t *testing.T) {
 	t.Parallel()
 
-	fp, err := NormalizeFilePath("C:\\programdata\\buildkite-agent")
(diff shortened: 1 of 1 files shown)


petemounce:
Actually, `C:programdata\buildkite-agent` will not work on Windows - or if it does, I'll be shocked.

The issue is that the code in bk agent does not recognise a path that starts with a drive letter `x:` as an absolute path.

lox:
You are right, windows doesn’t dig it. 🤷🏼‍♂️ I believe the agent is doing the right thing though! I left some more context in the Discourse thread.

PR #896, opened 2019-01-14
Add a script to check the agent docs are up to date

lox:
This is to spawn some discussion on ways that we could check our docs were up to date in an automated way. 

The problem with actually failing builds is that often the docs come after a release, and it probably doesn't make sense to require them before. 

Thoughts @toolmantim and @harrietgrace?


lox:
Closing in favour of https://github.com/buildkite/docs/pull/360

PR #898, opened 2019-01-16
Retry forever to upload job chunks

keithpitt:
Currently we wait a max of 50 seconds on each chunk before trying to upload it. If chunks are failing to upload, the main reason would be a problem with Buildkite. Presumably if they're failing to upload, finishing a job and starting a new one will also have the same problem (at which point we're just throwing chunk data away, to get more, and probably throw them away too).

This change tweaks the retry rules so we don't give up on chunks. It does handle cases that BK returns a 422, which "is" a successful chunk upload (whether or not we actually stored it is another thing), but a 422 does mean communications were successful, and Buildkite did "the right thing"

### DIFF
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -560,7 +560,7 @@ func (r *JobRunner) onUploadChunk(chunk *LogStreamerChunk) error {
 	// from Buildkite that it's considered the chunk (a 422 will be
 	// returned if the chunk is invalid, and we shouldn't retry on that)
 	return retry.Do(func(s *retry.Stats) error {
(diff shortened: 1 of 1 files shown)


sj26:
image

We send 400 and 401 responses sometimes. I can't find any 422 responses.

400 Bad Request is sent for "Missing chunk" and "Job can't have chunks" at least. Both are unrecoverable and would halt an agent here.

401 is sent when the agent authorization fails, and is likely unrecoverable.

Perhaps 4XX codes should not infinitely retry, but 5XX should?

keithpitt:
@sj26 excellent point, changed!

ticky:
Good sleuthing, @sj26! I need to remember how ridiculously powerful Insights is!

PR #899, opened 2019-01-19
Track reason why bootstrap was cancelled

lox:
This is a follow-on to #890, and I believe the final missing piece before we have signal handling 👌🏻in the agent. 

To differentiate between the agent stopping and other types of cancellation, we now send two new parameters to `job.finish` in the agent API:

* `signal`. This tracks the operating system signal that the bootstrap was terminated with. This is useful for figuring out whether a failed command was due to it actually failing or whether it was terminated with a signal.
* `signal_reason`. This is `agent_stop` for agent stopping or `cancel` for agent cancellation.

I'd suggest that we add support for retrying on strings like `agent_lost` or `agent_stopped` as well as numeric exit codes. Given lots of previous agents erroneously returned an exit code of `255` we should probably consider that `agent_lost` for now. 

I'd also suggest that we surface the cancel reason in the UI, like we do `Exited with -1 (Agent Lost)`, we'd also see `Killed with signal SIGTERM (Agent stopping)`.

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -248,9 +248,14 @@ func (p *Process) Run() error {
 		}
 	}
 
+	exitSignal := "nil"
+	if p.status.Signaled() {
+		exitSignal = SignalString(p.status.Signal())
+	}
+
 	// Find the exit status of the script
-	p.logger.Info("Process with PID: %d finished with Exit Status: %d",
-		p.pid, p.status.ExitStatus())
+	p.logger.Info("Process with PID: %d finished with Exit Status: %d, Signal: %s",
+		p.pid, p.status.ExitStatus(), exitSignal)
 
 	// Sometimes (in docker containers) io.Copy never seems to finish. This is a mega
 	// hack around it. If it doesn't finish after 1 second, just continue.
--- a/process/process.go
+++ b/process/process.go
@@ -248,12 +248,11 @@ func (p *Process) Run() error {
 		}
 	}
 
+	// Find the exit status or terminating signal of the script
 	exitSignal := "nil"
 	if p.status.Signaled() {
 		exitSignal = SignalString(p.status.Signal())
 	}
-
-	// Find the exit status of the script
 	p.logger.Info("Process with PID: %d finished with Exit Status: %d, Signal: %s",
 		p.pid, p.status.ExitStatus(), exitSignal)
 
--- a/process/signal.go
+++ b/process/signal.go
@@ -3,6 +3,8 @@
 package process
 
 import (
+	"fmt"
+	"os"
 	"syscall"
 )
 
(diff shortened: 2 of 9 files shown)


lox:
Yeah, that's why I reckon we should smooth it over in the backend @toolmantim. The behaviour has been pretty variable in the agent across a few versions.

toolmantim:
For docs, once this is merged, I was thinking something like https://github.com/buildkite/docs/pull/363/commits/9c9733ce3ad5034945d5a35827650c798d680987 But is that enough to fully use this new feature, or do we need to add support for `cancel_reason` and/or `signal` to the pipeline.yml attributes also?

Any ideas for how to smooth over things, backend wise? I think most people are doing this, for spot instances:

```yml
steps:
 - command: test.sh
    retry:
      automatic:
        - exit_status: -1
           limit: 2
        - exit_status: 255
           limit: 2
```

Would we just automatically clone the `exit_status: 255` to an additional rule for `exit_status: 143`? 🤔 The UI would now show `exit_status: 143` and their builds would be automatically retrying, and they might like "Hrmm why is that retrying"

lox:
I think probably using the actual exit codes is a bad idea for retrying these situations, as it's impossible to distinguish between 143 when a host shuts down and something actually returning 143. Plus, the fact that it's 143 now might change in future if we move to SIGINT vs SIGTERM (which we probably will).

I think my preference would be to add support for aliases for exit codes like below:

```
steps:
 - command: test.sh
    retry:
      automatic:
        - exit_status: [agent_lost, agent_stopped]
           limit: 2
```

That would be backwards compatible with an exit code of -1 or 255, and then also look for cancel_reason from newer agents.

Thoughts @keithpitt?

lox:
Also, windows doesn't use the same convention.

The other option would be to have `agent_stopped` return a `-1` to buildkite, which is where I was leaning initially but @toolmantim seemed to object.

lox:
Ok, so I've updated it to send `signal=SIG....` and `signal_reason=agent` when the agent is stopping.

lox:
I'm going to carve some chunks out of this one into other PR's.

lox:
Rebased and moved irrelevant details to other PR's. This now sends a `signal=SIGTERM|SIGKILL` and `signal_reason=agent_stop` when the bootstrap is terminated as part of agent shutdown or `signal_reason=cancel` when the bootstrap is terminated because the job was cancelled.

lox:
Worth noting that exit_code=255 hasn't been a thing since Agent v3.11.0, nobody has noticed 🤔

DazWorrall:
\<stalking>

Did we ever get a configurable signal for job cancellation? We would still love to handle node shutdown and job cancellation differently in our scripts - these efforts here will help clarify what happened to in the UI, but the downstream code is still missing some context afaict. Being able to customise the signal sent on job cancellation is one way to address that.

lox:
@DazWorrall ah yes, that was one of the motivations for this extended rabbit hole 😅 Currently we send `SIGTERM` for cancellation, but IMO we should be sending `SIGINT` for cancel and `SIGTERM` for shutdown. Perhaps adding a `cancel-signal = SIGTERM` config is a good way of rolling out that change gradually 🤔

harrietgrace:
@lox since we've got configurable cancel signals now (https://github.com/buildkite/agent/pull/1041), is this PR still blocked?  

If it's not likely to move soon, I'll remove the docs label from https://github.com/buildkite/agent/pull/890 and will put the 'needs docs' label on this one.

lox:
Unfortunately it’s blocked on backend changes.

ticky:
Rebased this on current master, and added code to report the exit signal to the log. This seems neat! Also put together this little PoC to make this happen: buildkite/buildkite#4695

process/signal.go
```diff
@@ -36,3 +38,75 @@ func (p *Process) interruptProcessGroup() error {
...
+		return "SIGWINCH"
+	case 29:
+		return "SIGINFO"
+	case 30:
+		return "SIGUSR1"
+	case 31:
+		return "SIGUSR2"
+	}
+	return fmt.Sprintf("%d", int(s))
+}
```
pda: We could perhaps use https://godoc.org/golang.org/x/sys/unix#SignalName

pda: (pushed in c98dd9b2 with test in 94ea51cd)

ticky: I think this spec needs to be marked Linux-specific, as Windows' signal names are different, and failing the build: https://buildkite.com/buildkite/agent/builds/3231#5bba8d45-cfee-47ca-9541-0dff8c5ff0ad/138-209

Might be good to add a Windows-specific spec, too!

ticky:
OMG! It’s green!

Probably needs some sort of test case? I’m not sure what the best way to approach that is…

process/process.go
```diff
@@ -248,9 +248,13 @@ func (p *Process) Run() error {
 		}
 	}
 
-	// Find the exit status of the script
-	p.logger.Info("Process with PID: %d finished with Exit Status: %d",
-		p.pid, p.status.ExitStatus())
+	// Find the exit status or terminating signal of the script
+	exitSignal := "nil"
```
keithpitt: Perhaps the default could be "?" so folks know that we didn't know, as apposed to it just being blank?

ticky: It isn’t that we don’t know, it’s that there isn’t one.

ticky:
I’d love to pair with a golang person on getting this ready to merge! We should make this happen!

pda:
Yes! Tomorrow sometime after lunch?

ticky:
Let’s do it @pda :)

PR #905, opened 2019-02-01
Add an AWS variant of docker image

lox:
This adds an AWS specific variant of the docker image (e.g. `buildkite/agent:3.8-aws`)

The plan is to include the aws cli and also any aws related hooks. This image would be used in the future various Elastic Stacks.

### DIFF
--- a/.buildkite/steps/build-docker-image.sh
+++ b/.buildkite/steps/build-docker-image.sh
@@ -54,7 +54,7 @@ codename="${3:-}"
 version="${4:-}"
 push="${PUSH_IMAGE:-true}"
 
-if [[ ! "$variant" =~ ^(alpine|ubuntu)$ ]] ; then
+if [[ ! "$variant" =~ ^(alpine|ubuntu|aws)$ ]] ; then
   echo "Unknown docker variant $variant"
   exit 1
 fi


lox:
Basically that we'd want the `aws` variant of all the agent versions:

`buildkite/agent:3-aws`, `buildkite/agent:3.8-aws`, `buildkite/agent:3.8.5-aws`, `buildkite/agent:latest-aws`,  `buildkite/agent:beta-aws`

This seemed the simplest, most transparent and repeatable way to do that. Totally open to suggestions though!

toolmantim:
Yeah, tricky! Wouldn’t you want to iterate on that image outside the official agent release process tho?

lox:
I think you are probably right @toolmantim, it's a tough balancing act, do we want it to be easy to target a specific agent version + aws, or do we want to be able to iterate quickly on the aws agent. 

I'm probably inclined to go with the latter and just offer `edge`, `latest` and `beta` on the aws agent image and move it to a dedicated repository.

toolmantim:
I've had a ponder, and have some ideas, if it helps?

What if it was just an "ecs" agent rather than an AWS one? I realised https://github.com/buildkite/aws-agent-with-whitelist was a mistake — we're probably(?) going to end up doing things specific to this stack in there (as I did). It might be good as an example of an image built for AWS, but maybe let's bother trying to go for a generic AWS image? Perhaps just `buildkite/ecs-stack-agent`?

With versioning, what if all you started with was three tags:

* `buildkite/ecs-agent:v1-stable` (`FROM buildkite/agent:3-stable`)
* `buildkite/ecs-agent:v1-beta` (`FROM buildkite/agent:3-beta`)
* `buildkite/ecs-agent:v1-edge` (`FROM buildkite/agent:3-edge`)

Similar to how newer `docker` cli works with older Docker daemons, what if those ECS images will always be compatible with older versions of the ECS stack (within a major version). What this means would be that you could keep updating the images on those tags, whenever a new patch or point release of the ECS stack comes out, or whenever a new agent version is released, and stacks automatically pick it up.

You wouldn't get to pick whether you're using v3.2 or v3.3 etc of the buildkite-agent. We just trust our semver guarantees and quickly fix any regressions if we roll out a bad image that breaks new ECS agents spinning up.

And for a given major version of the ECS stack, it's pinned to a major version of the Buildkite Agent. So when v4 of the buildkite-agent comes out, for example, people might need to update to ECS Stack v8 or whatever.

I guess, if people wanted the ability to pin their stack to a non-moving Docker image tag, or customise the docker image name, you could have an extra tag that has the build number, and have public build pages too?

* `buildkite/ecs-agent:build-48-stable`
* `buildkite/ecs-agent:build-48-beta`
* `buildkite/ecs-agent:build-48-edge`

I also just realised we don't have a `buildkite/agent:3-stable` — which is very sad ;(

toolmantim:
Given we don't really do agent betas at moment, we could just start with `buildkite/ecs-agent:v0` for dev'ing?

PR #908, opened 2019-02-04
Filter env used for command config out of environment

lox:
We follow a linux process model where any environment that is available to `buildkite-agent start` gets passed down to `buildkite-agent bootstrap` and then onto the actual job environment.

This includes env used for configuring those commands, for instance things like `BUILDKITE_AGENT_TOKEN` and other more minor things like `BUILDKITE_AGENT_TAGS`. These filter down to the execution environment for jobs, which creates a really wide API that is hard to keep up to date in docs. 

This change filters out any environment used for command config. They are unset, which means they need to be explicitly set by the job runner that creates the environment for the job.

Note that things other than env used for config are still passed down from the parent environment.

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -362,6 +362,9 @@ var AgentStartCommand = cli.Command{
 		// Setup the any global configuration options
 		HandleGlobalFlags(l, cfg)
 
+		// Remove any config env from the environment to prevent them propagating to bootstrap
+		UnsetConfigFromEnvironment(c)
+
 		// Force some settings if on Windows (these aren't supported yet)
 		if runtime.GOOS == "windows" {
 			cfg.NoPTY = true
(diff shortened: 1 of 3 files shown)


toolmantim:
I think this is a good change! It really clears up "Set these envs to configure the agent" vs "These are envs available in your job’s environment".

What's the story with things such as `BUILDKITE_DISABLE_GIT_SUBMODULES`?

As for roll-out, perhaps we just disappear them in `buildkite-agent` v4? And we could just list them in the deprecated envs in the docs?

toolmantim:
Is there a point release version of this we could get out, that only filters the agent token?

lox:
Yeah, I was thinking exactly that @toolmantim.

lox:
> What's the story with things such as `BUILDKITE_DISABLE_GIT_SUBMODULES`?

As in, why does it exist vs `BUILDKITE_NO_GIT_SUBMODULES` @toolmantim? Basically just legacy.

toolmantim:
I don't even remember why I asked about `BUILDKITE_DISABLE_GIT_SUBMODULES` — sorry! 🤭

clicommand/agent_start.go
```diff
@@ -362,6 +362,9 @@ var AgentStartCommand = cli.Command{
 		// Setup the any global configuration options
 		HandleGlobalFlags(l, cfg)
 
+		// Remove any config env from the environment to prevent them propagating to bootstrap
+		UnsetConfigFromEnvironment(c)
```
toolmantim: I just realised this is going to filter much more than just the `BUILDKITE_AGENT_TOKEN` yeah? For example:

```
env "PATH=$PATH:/usr/local/bin" buildkite-agent start --token xxx
```

Okay for a point release?

toolmantim: I hadn't considered this for a point release, only filtering the agent token 🤔

lox: No, it will only filter out env used for config for the `start` command.

toolmantim: Okay cool! Having re-read that, it totally says it right on the tin.

PR #909, opened 2019-02-04
If BUILDKITE_REPO is empty, skip checkout

lox:
This allows hooks to unset or empty `BUILDKITE_REPO` to conditionally skip checkout entirely.

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -791,17 +791,17 @@ func (b *Bootstrap) createCheckoutDir() error {
 // CheckoutPhase creates the build directory and makes sure we're running the
 // build at the right commit.
 func (b *Bootstrap) CheckoutPhase() error {
-	if b.Config.Repository == "" {
-		b.shell.Headerf("Skipping checkout phase")
-		return nil
-	}
-
 	if err := b.executeGlobalHook("pre-checkout"); err != nil {
 		return err
 	}
 
 	if err := b.executePluginHook("pre-checkout", b.pluginCheckouts); err != nil {
 		return err
+	}
+
+	if b.Config.Repository == "" {
+		b.shell.Headerf("Skipping checkout phase")
+		return nil
 	}
 
 	// Remove the checkout directory if BUILDKITE_CLEAN_CHECKOUT is present
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -791,6 +791,11 @@ func (b *Bootstrap) createCheckoutDir() error {
 // CheckoutPhase creates the build directory and makes sure we're running the
 // build at the right commit.
 func (b *Bootstrap) CheckoutPhase() error {
+	if b.Config.Repository == "" {
+		b.shell.Headerf("Skipping checkout phase")
+		return nil
+	}
+
 	if err := b.executeGlobalHook("pre-checkout"); err != nil {
 		return err
 	}
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -48,6 +48,9 @@ type Bootstrap struct {
 
 	// Whether the checkout dir was created as part of checkout
 	createdCheckoutDir bool
+
+	// Directories to clean up at end of bootstrap
+	cleanupDirs []string
 }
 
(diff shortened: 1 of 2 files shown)


keithpitt:
You have no idea how happy this makes me...

lox:
This now creates a temp dir for `BUILDKITE_BUILD_CHECKOUT_PATH` and skips running the default checkout hook if `BUILDKITE_REPO` is empty.

If you have an agent checkout hook or a plugin checkout hook, they will still run.

bootstrap/bootstrap.go
```diff
@@ -814,6 +818,18 @@ func (b *Bootstrap) CheckoutPhase() error {
 
 	b.shell.Headerf("Preparing working directory")
 
+	// If we have a blank repository then use a temp dir for builds
+	if b.Config.Repository == "" {
+		buildDir, err := ioutil.TempDir("", "bootstrap-builds")
```
matthewd: If this is going to be semi-visible, maybe "buildkite" over "bootstrap"? (We could possibly even include the build id or something to allow better backtracking if one notices dirs hanging around, but I don't know whether that's worth the bother.)

bootstrap/bootstrap.go
```diff
@@ -48,6 +48,9 @@ type Bootstrap struct {
 
 	// Whether the checkout dir was created as part of checkout
 	createdCheckoutDir bool
+
+	// Directories to clean up at end of bootstrap
+	cleanupDirs []string
```
keithpitt: Wouldn't you want the gear to stick around so in the event you want to debug something, the files are still on the FS? Is there any reason we can't rely on the OS cleaning up tmp dirs?

keithpitt:
@lox the thing about letting the OS cleanup the temp dirs (I commented on the line, maybe GitHub is being weird)

bootstrap/bootstrap.go
lox: I dunno that the OS cleaning up tmp dirs is a thing! It's definitely not on Linux.

lox:
Ah-hah, I found it.

bootstrap/bootstrap.go
matthewd: Yeah, Ruby `Tempfile` instances delete themselves when they get GCed. Otherwise, your best hope is that `/tmp` might get cleaned out on boot (very likely these days, because it's probably a tmpfs... back in the day, it was something the boot script specifically arranged to do).

We _could_ arrange to keep around the working directories for the last N builds or something, but IMO that'd be a fair bit of bother for little practical utility: I'd rather be confident that someone can manually invoke to reproduce it during their investigation. (@lox along that line, do you think it'd be useful to have an option to not do the cleanup? I don't know how often people need to jump in and read the tea-leaves of a failed build. Either way, I suspect that's more of a question for a future where the directory is potentially full of interesting things; right now it seems unlikely to contain anything of note.)

keithpitt: > Yeah, Ruby Tempfile instances delete themselves when they get GCed

Ah right. Yeah I remember having to `unlink` temp files in Ruby land to have them persist shutting down the process.

> Either way, I suspect that's more of a question for a future where the directory is potentially full of interesting things; right now it seems unlikely to contain anything of note.

Yeah, totes agree. Repo-less builds is new territory, so who knows if anyone will want the files to be kept around. I have like 1/10 feels on it at the moment, so I don't mind that it's new each time! And thinking about it more, hopping in and reading the FS I think is less and less of a thing in a Dockerish world.

hehoo:
With this, can I skip the workspace checkout for a certain step, not for the entire pipeline?

toolmantim:
@hehoo there's a plugin you can use to do this already https://github.com/thedyrt/skip-checkout-buildkite-plugin but you could do the same with an agent hook that conditionally exports `BUILDKITE_REPO=""` for example:

```bash
#!/bin/bash
set -euo pipefail

if [[ "$BUILDKITE_STEP_KEY" == "list-docker-containers" ]]; then
  export BUILDKITE_REPO=""
fi
```

hehoo:
@toolmantim Many Thanks. I will give it a try.

PR #910, opened 2019-02-04
Add buildkite agent job start cli command

keithduncan:
This CLI command accepts the job given `jobId`, creates and runs a `JobRunner` for the job uploading the logs to Buildkite. Conceptually this sits between the “agent worker” and the bootstrap process, and allows the actual work for a job to be spawned on a separate host/platform to where the agent is registered.

This is a proof of concept for an experiment building a serverless, or on-demand agent set up, where each job is executed as a separate one off task. My goal was to reduce the overhead of persistently running agents where my workload is extremely sporadic, and have my CI infrastructure scale to absolutely nothing when I’m not using it. I have coupled this with the new `job.scheduled` webhook API to actually schedule the jobs in response to demand.

--

Some notes on my specific use of this so far, I’m using a Buildkite webhook notification configured to send `job.scheduled` events to an AWS API Gateway, the API Gateway resource proxies the request to a lambda, the lambda looks up the agent registration token for the assigned agent in Parameter Store, the lambda then issues an ECS `RunTask` action with the job ID from the webhook. I have multiple task definitions registered in ECS, each with their own IAM role and docker image, these are addressed using a `task_definition=` tag in the agent targeting rules. This allows the `job.scheduled` lambda to schedule the correct agent environment for a given job.

--

Moved from https://github.com/keithduncan/agent/pull/1


lox:
This is utterly awesome. I am 100% keen to make this happen, and then roll your approach above into http://github.com/buildkite/elastic-ci-stack-for-aws-ecs. It solves a lot of problems I have been having trying to get auto-scaling of an ECS Service of agents working nicely. Scheduling one-off tasks is 👌🏻

lox:
My primary concern here is what the impact on the server-side is, @keithpitt can you see any potential issues? Otherwise I'm pretty good to merge.

keithpitt:
I haven't looked at the backend state machine to see what'll happen.

@keithduncan does it work? :)

lox:
Given we figured out how to do this with `buildkite-agent start`, are you ok if we close this out @keithduncan?

keithduncan:
giphy

PR #912, opened 2019-02-07
Add a `--content-type` to `artifact upload` to allow specifying a content type

### LINKED ISSUE #911: artifacts uploaded with .tgz extension are downloaded (by Chrome 72, on OSX) as .gtar
Example: https://buildkite.com/organizations/batfish/pipelines/pybatfish-pre-commit/builds/409/jobs/6a5fa631-9130-4689-afb2-eb910a60062c/artifacts/fbd9edd9-f104-4dc2-a3de-a39a3d61ae1d

Will redirect to `https://s3.amazonaws.com/buildkiteartifacts.com/f3f58cf3-46c3-493c-93f2-26a4b657ce75/4e844c21-a0ec-442e-be17-2157937c997a/6b675dae-642d-49de-abbe-96886b365cc5/6a5fa631-9130-4689-afb2-eb910a60062c/workspace/questions.tgz\?response-content-type\=application%2Fx-gtar`...

And in Chrome 72 on OSX 10.14.2 it will be downloaded as questions.gtar

Note that this has the correct filename (`workspace/questions.tgz`) and [a correct content type (`application/x-gtar`)](https://en.wikipedia.org/wiki/List_of_archive_formats#Archiving_and_compression).

We (@ticky / @lox / I) talked about this quite a lot here: https://buildkite-community.slack.com/archives/C02T53V9H/p1549487000332100

the gist of which was:

* https://bugs.chromium.org/p/chromium/issues/detail?id=1503 and
* a lack of recognition for gtar in somewhere around here https://cs.chromium.org/chromium/src/net/base/mime_util.cc?dr=C&g=0&l=196

Just in case, I filed https://bugs.chromium.org/p/chromium/issues/detail?id=929409 too.

Summarizing output,

@lox says:

> My vote would be a) update the mime database to nginx's (it's terse and reasonable) and b) add a `--content-type` flag to `buildkite-agent artifact upload`.

and I say:

> Well, maybe my bug report will cause them to add gtar to their list... In the meantime, I can solve it via 10 other ways or just ignore it.

So this is already worked-around AFAI care.

lox:
It was pointed out to us that `.tgz` files upload with a `Content-Type` of `application/x-gtar` causes chrome to append a different extension. 

This PR does two things 1) adds a `--content-type` flag to allow an exact `Content-Type` to be specified, and 2) replaces the list of mime types we are using with the Nginx list, which is shorter and sweeter. 

To simplify internals, this PR also moves mime detection into the uploader, versus duplicating it in the s3 and gs uploader. 

Closes #911.


ticky:
How does this affect uploading multiple artifacts? Does it apply it to _all_ artifacts.

lox:
Yup, it would apply to all artifacts @ticky.

ticky:
This has me wondering if it’d be better to do an `ext=content-type` sort of thing, so you could map it for multiple files, without it breaking other things if you glob?

i.e.
```bash
buildkite-agent artifact upload \
  --content-type tgz=application/x-gtar \
  --content-type pem=application/x-pem-file \
  dist/**/*
```

sj26:
I like `--content-type` for specifying a content-type explicitly. Perhaps separately we could pursue allowing adding to the content type map like @ticky suggests but with a new name, like `--content-type-map tgz=application/x-gtar`, etc? Then folks could add extra, custom mappings to their config file, too.

toolmantim:
What content type are we wanting to serve for `.tgz`?

plasticine:
Awesome — this looks great. 👏🏼

toolmantim:
> What content type are we wanting to serve for .tgz?

@lox any idea? So I can do the backend fix too.

lox:
I'd go for `application/gzip` @toolmantim.

toolmantim:
Thanks @lox!

Does the agent send the calculated mime type to the backend? Because at the moment, the Buildkite Agent API doesn't take any param I can see, and it seems to use a separate mime-type database based on just the file path of the artifact.

The backend mime type calculation is then used in both the download URL (forced, via the [response_content_type](https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectGET.html)) and in the Artifacts tab in the UI:

image

I don't think this fixes #911 yet I'm afraid…

lox:
This just sets the content-type in S3. It looks like we overwrite that when we generate download links though 😓 That said, for folks with custom s3 buckets it will work though.

I wonder why we overwrite the content-type on the server side, @keithpitt do you remember?

toolmantim:
I think we just never posted the mime type from the agent to the backend, and we needed to show it in the UI, so we used the file name to calc it on upload, and store it in the DB.

Then we had quite a few people reporting some quirks downloading files from browsers, so we added the mime type override code to the download part too (we ignore what’s stored in the DB and shown in the UI, perhaps on purpose to make older artifacts work… so we could reply on support with “All fixed! Try downloading again”)

I’m worried now we’ve added this option, people might think it fixes these download quirks for them, but currently doesn’t.

We could post the mime type to the backend? And use that in the UI and download links. For older agents that don’t send them we fall back to our current auto-calc in the backend.

We check the new agent mime type DB handles the fixes we’ve encourentered and handled in the Ruby code. And eventually we should be able to axe the Ruby side? 

There’ll be no more easy fixes for customers experiencing bad mime types though, it’ll mean an agent release for us, and an agent update for them.

Alternatively, we could roll back the CLI option and make this fix just an updated mime type DB for S3?

lox:
> We could post the mime type to the backend? And use that in the UI and download links. For older agents that don’t send them we fall back to our current auto-calc in the backend.

Yeah, that sound sensible to me. 

> There’ll be no more easy fixes for customers experiencing bad mime types though, it’ll mean an agent release for us, and an agent update for them.

We could always maintain our own list of incorrect agent associations that override when we generate presigned download links?

keithpitt:
> I think we just never posted the mime type from the agent to the backend, and we needed to show it in the UI, so we used the file name to calc it on upload, and store it in the DB.
> 
> Then we had quite a few people reporting some quirks downloading files from browsers, so we added the mime type override code to the download part too (we ignore what’s stored in the DB and shown in the UI, perhaps on purpose to make older artifacts work… so we could reply on support with “All fixed! Try downloading again”)

Yeah - @toolmantim's spot on here with how we ended up in this situation.

I think this is a good change, but probably needs a few supporting things for it to be a totes legit.

1. We should scope it to single files only i.e. if you try and upload multiple files, it'll blow up with "can't force a content type to multiple files" This stops people from shooting themselves in the foot accidentally by trying to fix 1 content type in a bunch of files being uploaded
2. Buildkite should receive that content type and respect it in the UI just like @toolmantim suggested. 

I do like this change because it gives people a way to say "I really want this content type, I know it's wrong, but I have to because _reasons_" - which is totally a thing I remember being a thing (i.e. customer uploads to S3, they have this other program that reads files but only works with `text/plain` something something).

PR #914, opened 2019-02-12
Delete docker-cleanup.sh

toolmantim:
I don't think this is used any more? Our recommendations are here: https://buildkite.com/docs/tutorials/docker-containerized-builds#adding-a-cleanup-task


lox:
Yup, nice one.

PR #924, opened 2019-02-20
Artifactory artifact support

GaryPWhite:
This branch provides the ability to use `buildkite-agent artifact <upload|download>` with an Artifactory instance.

`arty_uploader.go` and `arty_downloader.go` are plugged into the `artifact_uploader` and `artifact_downloader` to connect to an instance configured through three environment variables:

```bash
# point to an artifactory instance like https://artifactory.com/artifactory 
# Note that when using the jfrog CLI, this would be expected to include the /artifactory postfix on URL.
# That same expectation follows here
BUILDKITE_ARTIFACTORY_URL
 # a username and password configured in artifactory
BUILDKITE_ARTIFACTORY_USER
BUILDKITE_ARTIFACTORY_PASSWORD
```

The uploader will store artifacts in the repository under a folder matching the Job ID that uploaded it. for instance, `README.md` uploaded during job id `abcdefgh-ijkl-mnop-qrst-uvwxyz123456` will be in `repo/abcdefgh-ijkl-mnop-qrst-uvwxyz123456` 

Willing to provide more context and documentation if needed!


keithpitt:
Ooooh!

agent/artifact_uploader.go
```diff
@@ -215,6 +215,11 @@ func (a *ArtifactUploader) upload(artifacts []*api.Artifact) error {
 				Destination: a.conf.Destination,
 				DebugHTTP:   a.apiClient.DebugHTTP,
 			})
+		} else if strings.HasPrefix(a.conf.Destination, "rt://") {
+			uploader, err = NewArtifactoryUploader(a.logger, ArtifactoryUploaderConfig{
+				Destination: a.conf.Destination,
+				DebugHTTP:   a.apiClient.DebugHTTP,
+			})
 		} else {
 			return errors.New(fmt.Sprintf("Invalid upload destination: '%v'. Only s3:// and gs:// upload destinations are allowed. Did you forget to surround your artifact upload pattern in double quotes?", a.conf.Destination))
```
lox:
```suggestion
			return errors.New(fmt.Sprintf("Invalid upload destination: '%v'. Only s3://, gs:// or rt:// upload destinations are allowed. Did you forget to surround your artifact upload pattern in double quotes?", a.conf.Destination))
```

agent/arty_uploader.go
```diff
@@ -0,0 +1,168 @@
...
+		return err
+	}
+
+	return nil
+}
+
+func (u *ArtifactoryUploader) artifactPath(artifact *api.Artifact) string {
+	jobID := os.Getenv("BUILDKITE_JOB_ID")
+	if jobID == "" {
+		jobID = "NO-JOB-ID"
```
lox: Might be worth resolving this in the constructor so you can return an error?

lox:
This looks really great @GaryPWhite! Super excited! I've got very little critical feedback, gave it a quick look at it looks generally excellent. I'll give it a more in-depth review this evening on the ✈

agent/arty_uploader.go
GaryPWhite: I figured the only time this would be true would be in a develpment space -- would say it's good to have all those variables in one place :) Can change if you don't agree!

GaryPWhite: updated with a warning separate from authentication warning for JOB ID -- want to avoid confusion for anyone using pipelines / using artifact upload for local development.

GaryPWhite:
Worth noting that any documentation should now also include `BUILDKITE_JOB_ID` as a necessary variable to set if working outside of a BK pipeline

lox:
Let's merge this!

harrietgrace:
Docs: https://buildkite.com/docs/integrations/artifactory
Changelog: https://buildkite.com/changelog/77-new-artifactory-support-and-documentation

PR #930, opened 2019-02-28
Add support for new config field: tag-from-gcp-labels

conorgil:
Adds support for a new configuration field `tags-from-gcp-labels` that pulls the GCP instance labels and turns them into agent tags. This is the same functionality as `tags-from-ec2-tags` in AWS.

I tested this manually and seems to work fine. However, this is my first time updating the agent, so please shout if there are additional tests that need writing, docs that need updating, etc. I am happy to update.

### DIFF
--- a/agent/agent_pool.go
+++ b/agent/agent_pool.go
@@ -274,8 +274,6 @@ func (r *AgentPool) CreateAgentTemplate() *api.Agent {
 		if err != nil {
 			r.logger.Error(fmt.Sprintf("Failed to find GCP instance labels: %s", err.Error()))
 		}
-	} else {
-		r.logger.Info("TagsFromGCPLabels was FALSE, so skipping the new code")
 	}
 
 	var err error
--- a/agent/agent_pool.go
+++ b/agent/agent_pool.go
@@ -19,29 +19,31 @@ import (
 )
 
 type AgentPoolConfig struct {
-	ConfigFilePath        string
-	Name                  string
-	Priority              string
-	Tags                  []string
-	TagsFromEC2           bool
-	TagsFromEC2Tags       bool
-	TagsFromGCP           bool
-	TagsFromHost          bool
-	WaitForEC2TagsTimeout time.Duration
-	Debug                 bool
-	DisableColors         bool
-	Spawn                 int
-	AgentConfiguration    *AgentConfiguration
-	APIClientConfig       APIClientConfig
+	ConfigFilePath          string
+	Name                    string
+	Priority                string
+	Tags                    []string
+	TagsFromEC2             bool
+	TagsFromEC2Tags         bool
+	TagsFromGCP             bool
+	TagsFromGCPLabels       bool
+	TagsFromHost            bool
+	WaitForEC2TagsTimeout   time.Duration
+	WaitForGCPLabelsTimeout time.Duration
+	Debug                   bool
+	DisableColors           bool
+	Spawn                   int
+	AgentConfiguration      *AgentConfiguration
+	APIClientConfig         APIClientConfig
 }
 
 type AgentPool struct {
-	conf                  AgentPoolConfig
-	logger                *logger.Logger
-	apiClient             *api.Client
-	metricsCollector      *metrics.Collector
-	interruptCount        int
-	signalLock            sync.Mutex
+	conf             AgentPoolConfig
+	logger           *logger.Logger
+	apiClient        *api.Client
+	metricsCollector *metrics.Collector
+	interruptCount   int
+	signalLock       sync.Mutex
 }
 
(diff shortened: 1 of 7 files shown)


petemounce:
At the moment, my molecule tests branch still doesn't work. Would you mind posting ansible debug log output of success and a screenshot of the agent page from buildkite showing the tags (redacting anything sensitive)?

petemounce:
Cc @efarem for what you're working on.

agent/gcp_labels.go
```diff
@@ -0,0 +1,46 @@
...
+	"context"
+
+	"golang.org/x/oauth2/google"
+	"google.golang.org/api/compute/v1"
+)
+
+type GCPLabels struct {
+}
+
+func (e GCPLabels) Get() (map[string]string, error) {
```
lox: Given there aren't any instance variables in the struct, I'd tend to just make this a function, perhaps `GetGCPLabels(ctx context.Context) (map[string]string, error)`?

agent/agent_pool.go
```diff
@@ -248,6 +250,34 @@ func (r *AgentPool) CreateAgentTemplate() *api.Agent {
...
+			}
+			return err
+		}, &retry.Config{Maximum: 5, Interval: r.conf.WaitForGCPLabelsTimeout / 5, Jitter: true})
+
+		// Don't blow up if we can't find them, just show a nasty error.
+		if err != nil {
+			r.logger.Error(fmt.Sprintf("Failed to find GCP instance labels: %s", err.Error()))
+		}
+	} else {
+		r.logger.Info("TagsFromGCPLabels was FALSE, so skipping the new code")
```
lox: Assume we can ditch this?

lox:
This is looking good @conorgil, thanks! Left a few comments.

agent/agent_pool.go
conorgil: derp. Good catch! Removing

agent/gcp_labels.go
conorgil: I thought the syntax was a little weird, but it is my first time working with Go so I followed the pattern established in `ec2_tags.go`: https://github.com/buildkite/agent/blob/d6dfcf8ded85579ca40f731c5479cb7c471cd7d0/agent/ec2_tags.go#L9 which is used in `agent_pool.go`: https://github.com/buildkite/agent/blob/d6dfcf8ded85579ca40f731c5479cb7c471cd7d0/agent/agent_pool.go#L214

Happy to change this to a function, but I think the EC2 tags and GCP labels should follow the same pattern. I can open a second PR to switch the EC2 tags to a function too if you'd like.

agent/agent_pool.go
conorgil: Removed.

conorgil:
@petemounce make sure that you have the correct permissions in the IAM Role granted to the service account assigned to the instance. Check out [the docs](https://cloud.google.com/compute/docs/reference/rest/v1/instances/get) for the API method I use to pull down the instance data.

Here is a picture of this working in the Buildkite dashboard:

screen shot 2019-02-28 at 9 23 13 am

lox:
Ok, this will hit edge soon. Could I trouble you folks to do some testing and then I'll roll it into a new release?

petemounce:
We turned this on and then had to turn it off because of an unexpected behaviour.

We have labels on our GCE instances like `capable-of-building:platform` and `environment:production`.

We _also_ edit the buildkite-agent.cfg so it has tags like `capable_of_building=platform` and `environment=production`.

When `tags-from-gcp-labels=true`, we saw on the agent description pages tags like

```
...
capable-of-building=platform
capable_of_building=platform
...
```

and on the job-environment tab

```
...
BUILDKITE_AGENT_META_DATA_ENVIRONMENT=production,production
BUILDKITE_AGENT_META_DATA_CAPABLE_OF_BUILDING=platform,platform
...
```

Our build-automation scripts didn't expect the double-feeding of values into environment variables, and so caused us breakages.

Our agent-targeting rules in some cases stopped working, because in our step-definitions we have
```
agents:
  - environment=production
```
and now our agents would have `environment=production,production` and so would not match up.

We can work around this by removing our `buildkite-agent.cfg` supplied tags (now that we have the tags-from-gcp-labels feature) - but it was unexpected that buildkite would double-up the values like this.

petemounce:
Screenshot 2019-03-20 at 18 12 16

petemounce:
Screenshot 2019-03-21 at 13 39 24
Screenshot 2019-03-21 at 13 40 39

conorgil:
@petemounce huh, that is unexpected. thanks for sharing the screenshots!

We are using `tag-from-gcp-labels` in production and it is working great, but we definitely do not share your use-case and tagging setup.

Sounds like a follow up PR to de-duplicate agent tags would make sense. I've moved on to other tasks and don't have the time to do that, so someone else will have to tackle it if they are motivated.

PR #931, opened 2019-02-28
Flush output immediately if timestamp-lines not on

### LINKED ISSUE #929: Regression in logs of commands which don't output newlines
If you have a long running command that doesn't output a new line, nothing gets output in the build log until the command ends:

```yml
steps:
   - command: |
       while true; do
         sleep 0.1
         printf "."
       done
```

image

```
$ buildkite-agent --version
buildkite-agent version 3.8.4, build 2783
```

It appears you can't sed/tr/awk your way around the problem either:
https://gist.github.com/toolmantim/f4801f3ed2321b156105ff5e97fdbfbb

I believe we used to do some type of flushing, so I did a binary search of tests from `buildkite/agent:2` to the current release, and it appears to have regressed in [v3.6.0](https://github.com/buildkite/agent/releases/tag/v3.6.0) as [v3.5.4](https://github.com/buildkite/agent/releases/tag/v3.5.4) worked AOK:

image

lox:
As pointed out in #929, we are buffering line output before flushing it to log chunks. This is a regression that impacts things like rspec that output lots of dots before they output a newline.

The regression was a result of me simplifying some code so that the same codepath was used whether `--timestamp-lines` was turned on or off. 

At this stage, it's just too hard to support `--timestamp-lines` and not buffer lines (as we need to only add timestamp prefixes to non-headers), so this PR brings back the two code paths and fixes #929 at the expense of more code and being a bit gross.

### DIFF
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -60,8 +60,8 @@ type JobRunner struct {
 	// The internal process of the job
 	process *process.Process
 
-	// The internal line buffer of the process output
-	lineBuffer *process.LineBuffer
+	// The internal buffer of the process output
+	output *process.Buffer
 
 	// The internal header time streamer
 	headerTimesStreamer *headerTimesStreamer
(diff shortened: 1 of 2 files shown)


tekumara:
Legend! Thank you.

PR #934, opened 2019-03-03
Use vendor dir in build

lox:
We spend a heap of build time downloading go modules. We still have a `vendor/` dir for pre-go 1.11 `go get` support, so let's keep using it for a while. It's much faster.


lox:
This halves our build times from 13 mins to 6 mins. 💪🏻

PR #936, opened 2019-03-03
Shared repositories (git mirrors) between checkouts

lox:
This is another take on #917 that should hopefully be less controversial. Rather than introducing a new phase, this creates a shared repository mirror that is then used for [reference cloning](https://randyfay.com/content/reference-cache-repositories-speed-clones-git-clone-reference) the subsequent checkout. This should radically reduce network usage for cloning big repositories. Otherwise the semantics are the same as without a reference clone, no new checkout dirs or other complexities. 

This introduces an experiment, `git-mirrors`, and several new configuration options:

* `git-mirrors-path`: a directory that contains the git mirrors shared by agents on a give machine, unique to a given repository address
* `git-clone-mirrors-flags`: the parameters passed to `git clone` for the mirroring portion of the checkout. Defaults to `-v --mirror`.
* `git-mirrors-lock-timeout`: the number of seconds to wait for a lock to expire on a checkout that has crashed. Needs to be greater than the maximum time for a git clone on the agent.

It also introduces the idea of running the integration tests with an experiment enabled, which whilst potentially could end up with an explosion of permutations, for the minute it allows us to test checkouts with and without the `git-mirrors` experiment.

To test this out, you can run your agent with:

```bash
$  buildkite-agent start --experiment git-mirrors --git-mirrors-path /tmp/buildkite-git-mirrors
```

My plan is to merge this and release it in `3.10.0`, with the intention of making it the default and removing the experiment wrapper in `4.0.0`.


bootstrap/bootstrap.go
```diff
@@ -1076,9 +1127,37 @@ func (b *Bootstrap) defaultCheckoutPhase() error {
...
+		} else {
+			for idx, repository := range submoduleRepos {
+				// submodules might need their fingerprints verified too
+				if b.SSHKeyscan {
+					addRepositoryHostToSSHKnownHosts(b.shell, repository)
+				}
+
+				// if we have a reference, add the submodule to it
+				if reference != "" {
+					name := fmt.Sprintf("submodule%d", idx+1)
```
lox: This might require some more thought when submodules change.

agent/job_runner.go
```diff
@@ -440,6 +441,7 @@ func (r *JobRunner) createEnvironment() ([]string, error) {
 	// Add options from the agent configuration
 	env["BUILDKITE_CONFIG_PATH"] = r.conf.AgentConfiguration.ConfigPath
 	env["BUILDKITE_BUILD_PATH"] = r.conf.AgentConfiguration.BuildPath
+	env["BUILDKITE_REPOS_PATH"] = r.conf.AgentConfiguration.ReposPath
```
DoomGerbil: nit: Maybe it would be a bit more unambiguous here to call it something like `BUILDKITE_REPO_CACHE_PATH` or the like?

Otherwise, I could see how one might not immediately see the functional difference between this var for setting where mirrored copies live and `BUILDKITE_BUILD_PATH` for setting where the working copies live.

bootstrap/bootstrap.go
```diff
@@ -946,9 +951,63 @@ func hasGitSubmodules(sh *shell.Shell) bool {
 	return fileExists(filepath.Join(sh.Getwd(), ".gitmodules"))
 }
 
+func (b *Bootstrap) referenceRepository() (string, error) {
+	path := os.Getenv(`BUILDKITE_REPO_MIRROR_PATH`)
```
DoomGerbil: Is this different from `BUILDKITE_REPOS_PATH` set at `agent/job_runner.go#444`?

bootstrap/bootstrap.go
```diff
@@ -946,9 +951,63 @@ func hasGitSubmodules(sh *shell.Shell) bool {
...
+		return "", err
+	}
+	defer repoDirLock.Unlock()
+
+	if !fileExists(path) {
+		b.shell.Commentf("Cloning a mirror of the repository to %s", path)
+		if err := gitCloneMirror(b.shell, b.GitCloneFlags, b.Repository, path); err != nil {
+			return "", err
+		}
+	} else {
```
DoomGerbil: Do you need to worry about the case where the directory already exists, but it's not actually a valid clone?  EG - if someone accidentally/maliciously sets the mirror dir env var to a value that makes the agent attempt to overwrite unrelated contents on the machine's drive?

bootstrap/integration/bootstrap_tester.go
```diff
@@ -78,6 +84,7 @@ func NewBootstrapTester() (*BootstrapTester, error) {
 			"HOME=" + homeDir,
 			"BUILDKITE_BIN_PATH=" + pathDir,
 			"BUILDKITE_BUILD_PATH=" + buildDir,
+			"BUILDKITE_REPOS_PATH=" + reposDir,
```
DoomGerbil: Same comment as above about `BUILDKITE_REPOS_PATH` vs `BUILDKITE_REPO_MIRROR_PATH`.

clicommand/agent_start.go
```diff
@@ -245,6 +246,12 @@ var AgentStartCommand = cli.Command{
 			Usage:  "Path to where the builds will run from",
 			EnvVar: "BUILDKITE_BUILD_PATH",
 		},
+		cli.StringFlag{
+			Name:   "repos-path",
+			Value:  "",
+			Usage:  "Path to where the repositories will be stored",
```
DoomGerbil: Maybe for explicitness' sake: 

```suggestion
			Usage:  "Path to where the mirrored copies of repositories will be stored",
```

agent/agent_configuration.go
```diff
@@ -4,6 +4,7 @@ type AgentConfiguration struct {
 	ConfigPath                string
 	BootstrapScript           string
 	BuildPath                 string
+	ReposPath                 string
```
petemounce: Maybe `ReposCachePath`?

agent/job_runner.go
petemounce: I agree (but `BUILDKITE_REPOS_CACHE_PATH` - pluralise).

bootstrap/bootstrap.go
```diff
@@ -946,9 +951,63 @@ func hasGitSubmodules(sh *shell.Shell) bool {
...
+	// Create the base dir if it doesn't exist
+	if baseDir := filepath.Dir(path); !fileExists(baseDir) {
+		b.shell.Commentf("Creating \"%s\"", baseDir)
+		if err := os.MkdirAll(baseDir, 0777); err != nil {
+			return "", err
+		}
+	}
+
+	// Try and lock the repository dir to prevent concurrent clones
+	repoDirLock, err := b.shell.LockFile(path+".lock", time.Minute*5)
```
petemounce: Could this timeout please be surfaced? We have git repositories that take more than 5 minutes to clone.

We're _expecting_ the perforce repositories that we _will_ have to be multi-TB, which will _definitely_ require more than 5 minutes.

bootstrap/bootstrap.go
```diff
@@ -946,9 +951,63 @@ func hasGitSubmodules(sh *shell.Shell) bool {
...
+	if err != nil {
+		return "", err
+	}
+	defer repoDirLock.Unlock()
+
+	if !fileExists(path) {
+		b.shell.Commentf("Cloning a mirror of the repository to %s", path)
```
petemounce: nit: `%q` to get a quoted path?

bootstrap/config.go
```diff
@@ -85,6 +85,9 @@ type Config struct {
 	// Path where the builds will be run
 	BuildPath string
 
+	// Path where the repository mirrors are stored
+	ReposPath string
```
petemounce: `ReposCachePath`?

clicommand/bootstrap.go
```diff
@@ -200,6 +201,12 @@ var BootstrapCommand = cli.Command{
 			Usage:  "Directory where builds will be created",
 			EnvVar: "BUILDKITE_BUILD_PATH",
 		},
+		cli.StringFlag{
+			Name:   "repos-path",
```
petemounce: `repos-cache-path`?

petemounce:
How does this interact with running multiple agent processes per host via
* separate processes (e.g. systemd template-unit based vs unit based)
* `--spawn n`

?

I would prefer that there be a single reference of each repo per _host_ as opposed to per _BK agent instance_, to save disk space.

DoomGerbil:
Another quick thing comes to mind - since it's probably not safe to assume that you can _guarantee_ that the mirrored repos never get any unexpected mutations, then it's probably safer to also include the [`--dissociate` flag](https://git-scm.com/docs/git-clone#Documentation/git-clone.txt---dissociate) to git clone as well as `--reference`.  Otherwise, the reference clone will mostly just hard-link to the mirror clone, which can cause issues if any mutations happen to the mirror clone.

DoomGerbil:
Also, though this doesn't necessarily reflect on everyone's experience (since this particular repo is _very_ large), I did some git benchmarking of clones of our worst-performing repo using various different sets of `git clone` flags.

```# Totally normal clone
time git clone git@github.com:improbableio/UnrealEngine.git \
std-clone
	Time: 9m21.282s
	Size: 3.3GB

# Mirror clone
time git clone git@github.com:improbableio/UnrealEngine.git \
--mirror \
std-clone
	Time: 9m1.757s
	Size: 2.1GB

# Standard clone with a local reference
time git clone git@github.com:improbableio/UnrealEngine.git \
--reference std-clone \
ref-clone
	Time: 0m30.103s
	Size: 1.2GB

# Shallow clone from remote, no reference
time git clone git@github.com:improbableio/UnrealEngine.git \
--depth 1 \
shallow-clone
	Time: 1m16.511s
	Size: 1.4GB

# Standard clone with a local reference, dissociated
time git clone git@github.com:improbableio/UnrealEngine.git \
--reference std-clone \
--dissociate \
ref-dis-clone
	Time: 1m1.195s
	Size: 3.3GB

# Clone with a local reference, dissociated, no post-clone checkout
time git clone git@github.com:improbableio/UnrealEngine.git \
--reference std-clone \
--dissociate \
--no-checkout \
ref-dis-nocheckout-clone
	Time: 0m26.485s
	Size: 2.1GB

# Shallow clone with a local reference, dissociated, no post-clone checkout
time git clone git@github.com:improbableio/UnrealEngine.git \
--reference std-clone \
--dissociate \
--no-checkout \
--depth 1 \
ref-dis-nocheckout-shallow-clone
	Time: 0m19.503s
	Size: 1.5GB

# Shallow clone with a local reference, dissociated
time git clone git@github.com:improbableio/UnrealEngine.git \
--reference std-clone \
--dissociate \
--depth 1 \
ref-dis-shallow-clone
	Time: 0m35.618s
	Size: 2.6GB

# Shallow clone with a local reference, no post-clone checkout
time git clone git@github.com:improbableio/UnrealEngine.git \
--reference std-clone \
--no-checkout \
--depth 1 \
ref-nocheckout-shallow-clone
	Time: 0m2.188s
	Size: 696K
```

Again, this is not intended to be a scientific study of the impacts of these flags, but in our case, this is the scale of change that we see, and highlights why we want this change so much.  :)

bootstrap/bootstrap.go
lox: This is the singular form, this is a full path to the reference repository.

bootstrap/bootstrap.go
lox: That's not really a threat model that we protect against, you could already do that by setting `BUILDKITE_BUILD_CHECKOUT_PATH` couldn't you?

lox:
That data is super helpful @DoomGerbil!

lox:
@DoomGerbil I think I agree that `--dissociate` is a sensible default.

lox:
> I would prefer that there be a single reference of each repo per host as opposed to per BK agent instance, to save disk space.

@petemounce, yup that's the design!

agent/job_runner.go
lox: I wonder if it is actually a "cache" though. I'm not entirely sure that word adds clarity that justifies the additional length. Will think it over.

bootstrap/bootstrap.go
lox: It's the fully realized path to the reference repository vs the dir that repositories are checked out to. Possibly badly named though.

lox:
After some discussion @keithpitt and I ended up on `BUILDKITE_GIT_MIRRORS_PATH`.

bootstrap/bootstrap.go
lox: This is configurable now!

lox:
@harrietgrace I might need some assistance from you with updating the docs for this at some point!

toolmantim:
Is it an opt-in config option at the moment? If so, why don’t we remove it from experiments?

lox:
@toolmantim we've done that in that past. I think the big downside is that we miss out on the really clear "this is super experimental" that having something behind an experiment guard misses out. This is likely to be quite flakey for a month or two whilst we work through the edge cases, so I feel like having to strongly opt-in to the experimental nature helps manage expectations. 

The alternative is releasing a beta agent, but I'd much prefer to keep iterating quickly on stable with new things behind experiment flags.

toolmantim:
Okay! Probably shouldn’t be changelogged or added to docs yet then?

lox:
@toolmantim any reason we can't mention an experimental feature in a changelog? It's something heaps of people are interested in and would be great to get feedback on. We mention other beta features there?

toolmantim:
Oh, it sounded this was “get out a sneaky build to test if this actually works as designed” and I was assuming you had enough people to test it. And then after we’re confident it works, turn off the experimental flag in a new point release and changelog. I thought a changelog announcement would be a good push for making sure we move it out of experimental.

But perhaps we just do a changelog for every new agent release? And we can list the experimental changes in there, alongside other fixes.

toolmantim:
Sorry, I should have made my question clearer!

> My plan is to merge this and release it in 3.10.0, with the intention of making it the default and removing the experiment wrapper in 4.0.0.

I was wondering why it couldn’t be removed from experiments in a point release once we’ve tested it, and then enabled by default in v4?

Related q: would we have a default calculated value for `--git-mirrors-path`, or it’d error if people didn’t update their config, or it’s just be disabled if people didn’t update their config?

lox:
> I was wondering why it couldn’t be removed from experiments in a point release once we’ve tested it, and then enabled by default in v4?

Sure, quite possibly! I guess it depends how it goes? 

> Related q: would we have a default calculated value for --git-mirrors-path, or it’d error if people didn’t update their config, or it’s just be disabled if people didn’t update their config?

Nope, I don't think we would calculate a default value. We don't for `plugins-path` for instance. In v4 we'll change the packaging to set a default in the configs and during installation.

lox:
@DoomGerbil I'm going to remove `--dissociate` from the defaults (it's presently accidentally on the mirror clone anyway). It's pretty recently added to git and it also causes problems for git LFS. It's easy to add it to your `git-clone-flags` if you want it!

DoomGerbil:
> @DoomGerbil I'm going to remove `--dissociate` from the defaults (it's presently accidentally on the mirror clone anyway). It's pretty recently added to git and it also causes problems for git LFS. It's easy to add it to your `git-clone-flags` if you want it!

https://github.blog/2015-02-06-git-2-3-has-been-released/ says it's been there for four years now, but the point about LFS is well taken.  We don't support LFS with the rest of our infrastructure, so we would probably add the flag to our configuration, since as you said, it's not hard to tack it on ourselves.

lox:
The bit I'm still very blurry on is how the submodule support will work in practice.

lox:
We've decided to drop reference clone support for submodules for the first pass.

silvamerica:
Just an FYI, we've been doing this for a little while with a custom `elastic_bootstrap` script to create the mirror and git clone flags to `--reference-if-able` and we had immediate issues with the cloned repo missing refs before adding `--disassociate`.

lox:
@silvamerica yeah, I suspect we might end up with it being a default, but it brings with it a quite recent git dependency, so going to see how it goes. It's only really an issue if you plan on deleting the git bare repositories, isn't it?

PR #937, opened 2019-03-04
Increase pipeline upload retries on 5xx errors

toolmantim:
In the case of server problems, we should give the Agent API a bit more time to recover, and to successfully accept a pipeline upload, before giving up and failing the pipeline upload job (and build) — which results on user intervention being required.

This increases the retry interval, and total number of retries, in the case of 5xx errors on pipeline uploads from 5 seconds to 5 minutes.

I think it's a good tradeoff to save user intervention and a failed build status, at the cost of the pipeline upload step taking much longer to run than usual in the case of server problems (such as downtime, maintenance windows, or unusually high load).

- [ ] Increase to more than 5 minutes?
- [ ] Is every 5 seconds a good number? It seems too short?
- [ ] Check the Agent API to see if there's any known 5xx responses from the backend which we'd want to fail more quickly

### DIFF
--- a/clicommand/pipeline_upload.go
+++ b/clicommand/pipeline_upload.go
@@ -247,7 +247,9 @@ var PipelineUploadCommand = cli.Command{
 			}
 
 			return err
-		}, &retry.Config{Maximum: 5, Interval: 1 * time.Second})
+			// On a server error, it means there is downtime or other problems, we
+			// need to retry. Let's retry every 5 seconds, for a total of 5 minutes.
+		}, &retry.Config{Maximum: 60, Interval: 5 * time.Second})
 		if err != nil {
 			l.Fatal("Failed to upload and process pipeline: %s", err)
 		}


toolmantim:
Here's the current error HTTP status codes that could be returned from the Agent API:

- **102 Processing**: The pipeline upload is being processed, and the agent should continue to retry the pipeline upload indefinitely until it receives a 2xx response.
- **422 Unprocessable Entity**: There's an error in the pipeline upload parameters, and the agent should not retry the request.
- **500 Internal Server Error**: That was a server error, and the agent should continue to retry the pipeline upload for a period of time, and then consider it failed. This could happen if a bug was deployed, or unusually high load is causing failures.

@lox we don't seem to treat any of these differently at the moment — they all just fall under the same retry block?

lox:
This seems like a sensible change @toolmantim. 

In terms of handling the status codes differently, the code just about your change fails hard on 422, but retries on 500's. I'm not sure about 102's, that predates my time I think.

keithpitt:
Eventually we'll make 500 a hard fail, and improve the handling of "Agent API having a bad time" to return a 102

toolmantim:
🎉

PR #938, opened 2019-03-05
Allow the Agent API to reject header times

sj26:
If an agent goes away and the job its running is reaped and we archive away the logs and header times, allow the agent api to reject header times from the agent which will never succeed.

This is the same logic from chunks upload.


sj26:
Thanks @lox!

PR #946, opened 2019-03-11
Agent start without registration, allows using an access token

lox:
In the schedulers we've been writing for Lambda, ECS and other places, we quite often want to register an agent in one spot and then schedule a "one-shot" agent elsewhere. 

This adds two new flags to the `buildkite-agent start` command: 

* `--register-only` - Registers an agent and prints the access token and then exits
* `--access-token` - Skips registration of the agent and makes use of the provided access-token

In practice, I didn't love this implementation. The `agent.register` command returns lots of parameters for configuring the agent worker, such as the interpolated agent name, how often it should ping, etc. These needed to be have place holder values used.

Beyond that, I didn't love the duplication or the kind of confusing `AgentPool` method `StartWithoutRegister`. 

That said, this gets it done and makes it possible to register an agent in one spot and then run a once-off in another spot without potentially disclosing the agent registration token.

Interested in thoughts.

### DIFF
--- a/agent/agent_pool.go
+++ b/agent/agent_pool.go
@@ -98,33 +98,23 @@ func (r *AgentPool) StartWithoutRegister(agentName, accessToken string) error {
 	// Show the welcome banner and config options used
 	r.showBanner()
 
-	// Create a prefixed logger for some context in concurrent output
-	l := r.logger.WithPrefix(agent.Name)
-
-	worker := NewAgentWorker(l, agent, r.metricsCollector, AgentWorkerConfig{
+	worker := NewAgentWorker(r.logger, agent, r.metricsCollector, AgentWorkerConfig{
 		Endpoint:           r.conf.APIClientConfig.Endpoint,
 		DisableHTTP2:       r.conf.APIClientConfig.DisableHTTP2,
 		Debug:              r.conf.Debug,
 		AgentConfiguration: r.conf.AgentConfiguration,
 	})
 
-	l.Info("Connecting to Buildkite...")
+	r.logger.Debug("Skipping agent registration as an access token was provided")
+	r.logger.Info("Connecting to Buildkite...")
 	if err := worker.Connect(); err != nil {
 		return err
 	}
 
-	if r.conf.AgentConfiguration.DisconnectAfterJob {
-		l.Info("Waiting for job to be assigned...")
-		l.Info("The agent will automatically disconnect after %d seconds if no job is assigned", r.conf.AgentConfiguration.DisconnectAfterJobTimeout)
-	} else if r.conf.AgentConfiguration.DisconnectAfterIdleTimeout > 0 {
-		l.Info("Waiting for job to be assigned...")
-		l.Info("The agent will automatically disconnect after %d seconds of inactivity", r.conf.AgentConfiguration.DisconnectAfterIdleTimeout)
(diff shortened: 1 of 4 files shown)


lox:
This was also largely inspired by #910, so I'd love to hear your thoughts @keithduncan!

agent/agent_worker.go
```diff
@@ -121,7 +121,7 @@ func (a *AgentWorker) Start() error {
 
 	// Create the intervals we'll be using
 	pingInterval := time.Second * time.Duration(a.agent.PingInterval)
-	heartbeatInterval := time.Second * time.Duration(a.agent.HearbeatInterval)
```
keithpitt: Lol at my typo

lox: I know I lolled too

keithpitt:
This is *very* exciting! Being able to split the `register` and `connect` parts of the start process opens up all sorts of new and interesting things.

The internal changes look fine to me, the API feels a little awkward (although I can totally see how within the context of the current agent commands it makes sense).

Another take on the API could be:

```
buildkite-agent agent register --name "Totes" | buildkite-agent agent connect
```

Maybe there's an awkwardly named `agent` subcommand we can introduce that could add the raw commands to?

```
buildkite-agent agent register --name "Totes" --count 5
```

Could perhaps register 5 agents at once, and returns the response.

Then conceptually, `buildkite-agent start` is a shortcut method that essentially the `register` and `connect` actions together?

Or, maybe we can have `buildkite-agent register` and `buildkite-agent connect` but don't document them, and have them as secret subcommands or something? Or would that be rude?

lox:
@keithpitt would `connect` JUST connect, or would it also then ping for jobs, accept, start, run and then disconnect?

agent/agent_pool.go
```diff
@@ -51,13 +51,88 @@ func NewAgentPool(l *logger.Logger, m *metrics.Collector, c AgentPoolConfig) *Ag
...
+		return err
+	}
+
+	r.logger.Info("Successfully registered agent \"%s\" with tags [%s]", registered.Name,
+		strings.Join(registered.Tags, ", "))
+
+	// Create a prefixed logger for some context in concurrent output
+	l := r.logger.WithPrefix(registered.Name)
+
+	l.Debug("Ping interval: %ds", registered.PingInterval)
```
keithduncan: What do you think about logging these as a serialised JSON object so they can be parsed with `jq` instead of a `grep` / `sed` / `cut` combo?

keithduncan:
This is pretty rad, I’m a fan of splitting agent registration and execution so that execution can be slightly less trusted.

My main thought looking at this is whether you’d want to expose an agent registration API? It might make consuming this in something like stock lambda easier with https/graphql vs building the agent binary.

lox:
> My main thought looking at this is whether you’d want to expose an agent registration API? It might make consuming this in something like stock lambda easier with https/graphql vs building the agent binary.

@keithduncan totally, that's the intention in future. The agent API is technically private, but quite accessible and callable!

agent/agent_pool.go
lox: That's a super interesting idea, I'd love to do that for all loggers and have a `--json` flag.

lox:
Going to split this into some other PR's.

PR #949, opened 2019-03-14
Add agent.Registrar and a `register` command

lox:
This adds a hidden `register` command that will be used with a forthcoming `connect` command for an advanced agent flow where you want to split the register from the agent execution, for instance when using a K8S or ECS scheduler. 

This involves a pretty heavy refactor of the `AgentPool`. I split out the registration related parts into a `Registrator` which let me take a lot of complexity out of `AgentPool`. I'm pleased with the result and think it will make future custom agents much easier to assemble out of components. 

This PR is 98% about the refactor and 2% about the register command, but interested in feels on both.

### DIFF
--- a/agent/agent_configuration.go
+++ b/agent/agent_configuration.go
@@ -1,5 +1,14 @@
 package agent
 
+import (
+	"fmt"
+	"os"
+
+	"github.com/buildkite/agent/logger"
+)
+
+// AgentConfiguration is the run-time configuration for an agent that
+// has been loaded from the config file and command-line params
 type AgentConfiguration struct {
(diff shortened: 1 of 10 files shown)


lox:
I could potentially split this into a few PR's if it's difficult to review.

lox:
Going to split this out.

plasticine:
This looks rad, I’m keen to see the split PRs! <3

lox:
The basic refactor that is needed to support this is in https://github.com/buildkite/agent/pull/956. Would love your feels on it @plasticine!

PR #950, opened 2019-03-14
Cleanup register and start clicommands

lox:


lox:
Gonna do this at the same time as a `connect` subcommand on-top of #954.

PR #951, opened 2019-03-14
Check the agent works with bk cli

lox:
This adds a final test of the agent with `bk` and a local pipeline that exercises all the main commands of the agent and checks an end-to-end test works.


lox:
Is this too meta @keithpitt?

PR #954, opened 2019-03-22
Refactor agent pool to externalize registration

lox:
This pulls a whole lot of dependencies out of the AgentPool and moves them to an AgentTemplate that gets passed to an `agent.Register` command. 

This makes agent registration and spawning workers much more compose-able for future changes.

This also moves some of the labels/metadata stuff for AWS and GCP to the system package, which becomes the spot for "information about the running host".

### DIFF
--- a/agent/agent_configuration.go
+++ b/agent/agent_configuration.go
@@ -1,5 +1,14 @@
 package agent
 
+import (
+	"fmt"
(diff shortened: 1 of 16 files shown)


lox:
I'll split this up into smaller pieces @keithpitt.

PR #956, opened 2019-03-23
Move AgentWorker construction and registration out of AgentPool

lox:
This is another take on #954 that reduces the diff somewhat. 

The aim here is to decouple the AgentPool from the creation of AgentWorkers and registration. This allows better composition and pushes the construction of dependencies up to the caller, which is the cli command level.

The only practical difference in how things work here is that there is a single signal handler for the whole AgentPool vs one for each AgentWorker. This shouldn't have anything other than aesthetic difference.

Registration and fetching of tags are now different things too, and the results are passed into things that need them.

### DIFF
--- a/agent/agent_configuration.go
+++ b/agent/agent_configuration.go
@@ -1,5 +1,7 @@
 package agent
 
+// AgentConfiguration is the run-time configuration for an agent that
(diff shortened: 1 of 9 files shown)


lox:
💪🏻

Issue #437, opened 2017-03-14
Support Shallow Clones


piotrb:
The main idea is that for larger repos a full fetch is not necessary and slower.

A few key points:

1. in shallow mode, repositories should not be cached between builds (shallow clones are finicky and should be made fresh each time)
2. the shallow clone should just attempt to fetch the specific sha1 the test needs

This should help not overload upstream providers with a lot of fetches and should speed up the initial fetch a lot.

I'm optimizing for the scenario where host machines are likely to have an entirely cold repo root cache and the download will take a significant amount of time. Especially in events when upsteam might be feeling a bit sluggish like during the github outage on Mar 14, 2017.

mallyvai:
@piotrb Was thinking of this use-case - Can't we already just be able to populate "git-clone-flags" with "--depth=1" to the agent configuration?  

https://buildkite.com/docs/agent/configuration

piotrb:
No .. at least last time I checked .. while the parameter is passed just fine, the checked out working copy ends up being broken and makes the agent basically useless since it can't find refs properly on the shallow clone (its been a while so I'm a bit vague about the specifics)

mallyvai:
Thanks Piotr. This is a great (and common) optimization in build systems. Going to +1 this. At a bare minimum it should be documented IMO! :)

lox:
I've avoided this because historically (pre git 1.9) shallow clones had lots of edge conditions. Perhaps we will look at this post v3.

Walther:
Hi! Just checking on this - I saw PR #957 getting merged to master, but no mention in the changelogs & wasn't sure if it has made it to a release, and if so, from which version up is it available? Thanks! <3

prathe:
Maybe one edge conditions of shallow clone and fetch is that it creates a lock file when fetching shallow.

With parallel agents running the `git fetch` on the same directory, I've had this error:

    git fetch -v --prune --depth=1 origin refs/pull/5190/head
    fatal: Unable to create '/var/lib/buildkite-agent/builds/buildkite-i-010121511371dd0c0-1/kimoby/kimoby-rails/.git/shallow.lock': File exists.

So `git-clone-flags` and `git-fetch-flags` allow shallow clone and fetch, but it does not seem compatible with parallel agents running the commands.

I don't know how to prevent this with the hosted solution. Any idea?

sj26:
Sorry, I pressed the wrong button!

sj26:
[...]

PR #957, opened 2019-03-23
Add git-fetch-flags

lox:
This adds a `git-fetch-flags` which is the missing piece for shallow clones.


lox:
Pretty sure this should be enough to close https://github.com/buildkite/agent/issues/437 along with some docs. 

We could possibly support `git-depth` if it wasn't clear enough that you need `git-clone-flags=--depth=1` and `git-fetch-flags=--depth=1`.

lox:
Would love your thoughts on a) whether this is a good idea @sj26 and b) whether `--prune` should be a default fetch param. I couldn't figure out why we didn't have `--prune` on the github pull request option.

sj26:
Seems reasonable given our previous approach to these flags. Turning on pruning by default does also sound sensible.

I wonder if we need to teach the agent more about git and allow options like "I want shallow clones" etc? Sounds like a longer term thing though.

lox:
@sj26 yeah, the whole point of this was to make shallow clones possible, I actually started out with a `git-clone-depth` param, and might go back there, but figured that adding `git-fetch-flags` makes it easier for folks to experiment, which seemed more inline with the philosophy of the other configuration params.

bootstrap/bootstrap.go
```diff
@@ -1120,7 +1122,7 @@ func (b *Bootstrap) defaultCheckoutPhase() error {
 		b.shell.Commentf("Fetch and checkout pull request head from GitHub")
 		refspec := fmt.Sprintf("refs/pull/%s/head", b.PullRequest)
 
-		if err := gitFetch(b.shell, "-v", "origin", refspec); err != nil {
```
lox: This is the only case we don't include `--prune` @sj26, do you happen to recall why?

sj26: I think we just never added it 🤷‍♂️

PR #961, opened 2019-03-29
Add checksums for artifactory uploaded artifacts

lox:
This adds checksums so that Artifactory doesn't complain in this screen:

Artifactory 2019-03-29 11-26-38


lox:
There was some question on Slack that perhaps the `SHA256` checksums actually caused problems with some versions of Artifactory.

PR #963, opened 2019-03-31
Install docker-compose with curl vs pip

lox:
Seeing random failures related to `python2` and `pip` in the Alpine docker build. This removes that dependency entirely.


lox:
Nope, that binary doesn't work with musl.

PR #964, opened 2019-03-31
Pin docker-compose to 1.23 to avoid extra deps

lox:
The new docker-compose version introduces a dependency on gcc when installing. To avoid this, we'll pin to the previous version for now.

https://github.com/docker/compose/issues/6617

### DIFF
--- a/packaging/docker/alpine-linux/Dockerfile
+++ b/packaging/docker/alpine-linux/Dockerfile
@@ -17,7 +17,7 @@ RUN apk add --no-cache \
       tzdata \
     && \
     pip install --upgrade pip && \
-    pip install docker-compose
+    pip install --quiet docker-compose~=1.23.0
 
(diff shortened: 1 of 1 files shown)


toolmantim:
Should we be looking to replace the pip line with a binary download instead?

```sh
curl -L https://github.com/docker/compose/releases/download/1.24.0/docker-compose-`uname -s`-`uname -m` -o /usr/local/bin/docker-compose
chmod +x /usr/local/bin/docker-compose
```

lox:
@toolmantim https://github.com/buildkite/agent/pull/963

toolmantim:
Ah, I see that in #963 — maybe we need a comment in the Dockerfile?

lox:
Yeah, I reckon @toolmantim, will revisit.

PR #965, opened 2019-03-31
Replace testify with gotesttools

lox:
Currently [testify](https://github.com/stretchr/testify) brings a lot of dependencies with it and also a lot of stuff we don't use. This replaces it with gotest.tools, which is a lighter and less "framework-y" set of assertions. 

Some details at https://vincent.demeester.fr/posts/2018-08-16-gotest-tools-assertions/.


lox:
Meh, given lots of our dependencies still use `testify`, this doesn't end up buying us much.

PR #977, opened 2019-04-04
Artifactory uploader patch

GaryPWhite:
Third time's the charm?

This  pr gets rid of our explicit reliance on `JOB_ID`.

In other installations like S3 and GCS, the usage of our `upload_destination` is:

`BUIDLKITE_UPLOAD_DESTINATION = (gs|s3)://some-repo/$BUIDLKITE_(BUILD|JOB)_ID`

Since this injection happens at the time we upload artifacts, the metadata we get _back_ during the download phase matches the exact URL we would expect. In our case, the downloader is _not_ provided with the right `UploadDestination`, because we augment it within our uploader with the appended `JOB_ID`.

This PR should solve that issue :)

Let me know if you have questions!


lox:
Apologies for not catching that in review, the uploader is still an area of the agent I'm a bit blurry on!

PR #978, opened 2019-04-04
Add a docker image to cache objects to speed things up

lox:


lox:
None of these things seemed to speed up builds at all 🤔

PR #990, opened 2019-04-17
Allow applying ec2 tags when config tags are empty

vanstee:
Looks like this might have been a bug introduced in [#970](https://github.com/buildkite/agent/pull/970/files#diff-64fc4f8002c51b533bf973bd314f0ed4R104). Not a huge deal since we can get around it by setting a custom tag in the config in the meantime.


lox:
Thanks @vanstee!

PR #994, opened 2019-04-20
Fix bk cli test in CI

lox:
Unfortunately if we'd paid attention to the soft-fail of the bk cli test we might have avoided the bug in #987.

This attempts to fix the failure I was ignoring.


lox:
Ah-hah. The issue was a regression where we output logs to stdout vs stderr. Fixing in https://github.com/buildkite/agent/pull/996.

PR #995, opened 2019-04-20
Build all binaries in parallel in a single step

lox:
Experiment with using `gox` to build all binaries in one step to see if it's faster.


lox:
Well, I can conclusively say this isn't any quicker.

PR #999, opened 2019-04-22
Only trigger idle disconnect if all workers are idle

### LINKED ISSUE #985: Agents with `--disconnect-after-idle-timeout` and `--spawn` don't terminate as expected
Using the new lambda scaler v0.4.0 with the new termination and graceful shutdown scripts.

For example, if I have 8 agents running on a single instance using the multiple agents per instance functionality, when the inactive or cooldown period elapses and only 1 of those agents actually did any work. 

7 of the inactive ones will shutdown, but the 1 that had done some work had its idle timer reset.

This leaves my instance in a state where I have a big box such as a t3.2xlarge or something like that with only a single agent running and is not a very good efficient use of resources or cost.

What we need to overcome this, is if I have multiple agents on a single instance for the instance to ONLY shutdown if ALL the agents have reached the inactive period. Its an all or nothing kind of deal.
Because now it looks like I only have 1 available agent and then the lambda function will scale my ASG up another instance, so I'd now have 9 total buildkite agents running but over 2 instances.

You could also get very unlucky and fill out the max_size on your autoscaling group with only a single agent per instance.

Hopefully that made sense.

lox:
Closes #985.

I don't love the addition of a monitor object into `AgentWorker.Start`, but it was the simplest approach that didn't require massive structural changes to how AgentPool and AgentWorker interact. I tried about 3 different ways of doing that, all of which led to pretty sweeping changes. 

Would welcome suggestions of alternate approaches!

### DIFF
--- a/agent/agent_worker.go
+++ b/agent/agent_worker.go
@@ -176,7 +176,7 @@ func (a *AgentWorker) Start(idle *IdleMonitor) error {
 				select {
 				case <-a.idleTimer.C:
 					// Mark this agent as idle in the shared idle monitor
-					a.idleMonitor.MarkIdle(a.agent.Name)
+					a.idleMonitor.MarkIdle(a.agent.UUID)
 
 					// Only terminate if all agents in the pool are idle, otherwise extend the timer
 					if a.idleMonitor.Idle() {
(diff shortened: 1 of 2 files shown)


agent/agent_worker.go
```diff
@@ -165,15 +166,34 @@ func (a *AgentWorker) Start() error {
...
 	if a.agentConfiguration.DisconnectAfterIdleTimeout > 0 {
 		a.idleTimer = time.NewTimer(time.Second * time.Duration(a.agentConfiguration.DisconnectAfterIdleTimeout))
 		go func() {
 			for {
 				select {
 				case <-a.idleTimer.C:
-					a.logger.Info("Agent has been idle for %d seconds", a.agentConfiguration.DisconnectAfterIdleTimeout)
-					a.stopIfIdle()
+					// Mark this agent as idle in the shared idle monitor
+					a.idleMonitor.MarkIdle(a.agent.Name)
```
keithpitt: Should we use it's UUID instead of it's name? Just in case multiple agents have the same one?

lox: Yeah, probably a good idea.

lox: Done ⚡️

PR #1000, opened 2019-04-23
Fix assumption that custom GCS hosts use the same path format as GCS.

DoomGerbil:
We store our BK-generated artifacts in a private GCS bucket, but want to expose those artifacts to internal users via HTTPS.  For this we use a GCS->HTTPS proxy that sits in the middle, checking user authorization, and only allowing our users through.

However, the buildkite-agent artifact URL generation func appears to be hard-coded to insert the GCS bucket name into the path of the URL it generates.  This is necessary when talking directly to GCS, since the URLs to use look like `https://storage.googleapis.com/$BUCKET_NAME/$FILE_PATH`.

However, our proxy only talks to one bucket, and thus the URLs used to talk to it do not include the unnecessary bucket name - it generates URLs of the format `https://$ARTIFACT_PROXY_HOST/$FILE_PATH`.  Thus, this hard-coded insertion into the URL prevents our proxy from working, unless we hack our proxy to strip out the bucket name for this specific case, which is less than ideal.

This PR preserves the current behavior unless `BUILDKITE_GCS_ACCESS_HOST` has been overridden.  In that case, since we're using a custom GCS access host, it doesn't make sense to hard-code in the same path modification done for talking directly to GCS - the path of the artifact, as passed into `URL()`, should properly point at the stored artifact.

FYI @petemounce @lox

### DIFF
--- a/agent/gs_uploader.go
+++ b/agent/gs_uploader.go
@@ -88,14 +88,18 @@ func newGoogleClient(scope string) (*http.Client, error) {
 
 func (u *GSUploader) URL(artifact *api.Artifact) string {
 	host := "storage.googleapis.com"
+	path := fmt.Sprintf("%s/%s", u.BucketName, u.artifactPath(artifact))
+
+	// When overriding the default GCS host to go through a proxy, we should use the path passed in directly - it's not safe to assume that a proxy's URL path follows the same path format as GCS.
 	if os.Getenv("BUILDKITE_GCS_ACCESS_HOST") != "" {
 		host = os.Getenv("BUILDKITE_GCS_ACCESS_HOST")
+		path = u.artifactPath(artifact)
 	}
 
 	var artifactURL = &url.URL{
 		Scheme: "https",
 		Host:   host,
-		Path:   u.BucketName + "/" + u.artifactPath(artifact),
+		Path:   path,
 	}
 	return artifactURL.String()
 }
--- a/agent/gs_uploader.go
+++ b/agent/gs_uploader.go
@@ -90,10 +90,14 @@ func (u *GSUploader) URL(artifact *api.Artifact) string {
 	host := "storage.googleapis.com"
 	path := fmt.Sprintf("%s/%s", u.BucketName, u.artifactPath(artifact))
 
-	// When overriding the default GCS host to go through a proxy, we should use the path passed in directly - it's not safe to assume that a proxy's URL path follows the same path format as GCS.
-	if os.Getenv("BUILDKITE_GCS_ACCESS_HOST") != "" {
-		host = os.Getenv("BUILDKITE_GCS_ACCESS_HOST")
-		path = u.artifactPath(artifact)
+	// Override the default host if required.
+	if envHost, ok := os.LookupEnv("BUILDKITE_GCS_ACCESS_HOST"); ok {
+		host = envHost
+	}
+
+	// If this is set, we trust the user to supply a valid path prefix, and don't use a default GCS path
+	if prefix, ok := os.LookupEnv("BUILDKITE_GCS_PATH_PREFIX"); ok {
+		path = fmt.Sprintf("%s%s", prefix, u.artifactPath(artifact))
 	}
 
 	var artifactURL = &url.URL{
--- a/agent/gs_uploader.go
+++ b/agent/gs_uploader.go
@@ -87,18 +87,23 @@ func newGoogleClient(scope string) (*http.Client, error) {
 }
 
 func (u *GSUploader) URL(artifact *api.Artifact) string {
+	// Set host and pathPrefix to default values
 	host := "storage.googleapis.com"
-	path := fmt.Sprintf("%s/%s", u.BucketName, u.artifactPath(artifact))
+	pathPrefix := u.BucketName
 
-	// Override the default host if required.
+	// Override default host if required.
 	if envHost, ok := os.LookupEnv("BUILDKITE_GCS_ACCESS_HOST"); ok {
 		host = envHost
 	}
 
-	// If this is set, we trust the user to supply a valid path prefix, and don't use a default GCS path
+	// If this is set, we trust the user to supply a valid path prefix, instead of using the default GCS path
 	if prefix, ok := os.LookupEnv("BUILDKITE_GCS_PATH_PREFIX"); ok {
-		path = fmt.Sprintf("%s%s", prefix, u.artifactPath(artifact))
+		pathPrefix = prefix
 	}
+
+	// Build the path from the prefix and the artifactPath
+	// Also ensure that we always have exactly one / between prefix and artifactPath
+	path := fmt.Sprintf("%s/%s", strings.TrimSuffix(pathPrefix, "/"), u.artifactPath(artifact))
 
 	var artifactURL = &url.URL{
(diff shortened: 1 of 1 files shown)


DoomGerbil:
> LGTM. Reasonable to add test coverage?

Not trivial to do - since this func is part of the `GSUploader` struct, it would require instantiating a `GSUploader`, which requires a GCS client, so we'd need to mock that bit out.

I don't see any existing GCS client mocks for other GCS-related functionality, and am unsure if there's an established process around that here, but am open to it.

lox:
@DazWorrall will this impact any of your usage of GCS artifact uploads?

lox:
Sorry for the slow response, saw this when it came in and got caught up with thinking through how it might affect other folks. I've reached out to a few and we'll aim to get it merged ASAP if there are no complications.

DazWorrall:
Hi all 👋 

We for sure rely on the current behaviour. To avoid running our own proxy (for the same reasons - to manage access to the bucket using Google perms) we set `BUILDKITE_GCS_ACCESS_HOST=storage.cloud.google.com`, which results in URLs like `https://storage.cloud.google.com/<bucket>/<path>` - these URLs transparently respect bucket permissions, and redirect users with access to a signed URL to access the content of `path`. Without the bucket component `storage.cloud.google.com` doesn't work.

I wonder if that would also work for your use case @DoomGerbil @petemounce ?

DoomGerbil:
Unfortunately it doesn't work for us.  As per the [GCS documentation](https://cloud.google.com/storage/docs/request-endpoints#cookieauth) cookie-auth-based GCS access is disabled if you have Audit Logging switched on for GCS buckets, which - due to the sensitive nature of some of the artifacts we store - our internal security policy requires that we use.  I've also tested it out manually to confirm, and sadly, even when I'm explicitly marked as owner on a GCS object, I cannot access it via the browser, confirming what the docs say.

I don't like to add special cases in code, but it seems that - in the case of URL generation, anyway - setting `BUILDKITE_GCS_ACCESS_HOST=storage.cloud.google.com` is functionally the same as not changing `BUILDKITE_GCS_ACCESS_HOST` at all.

I _would_ suggest doing this instead:

```
host := "storage.googleapis.com"
path := u.artifactPath(artifact)

	// When overriding the default GCS host to go through a proxy, we should use the path passed in directly - it's not safe to assume that a proxy's URL path follows the same path format as GCS.
	if os.Getenv("BUILDKITE_GCS_ACCESS_HOST") != "" {
		host = os.Getenv("BUILDKITE_GCS_ACCESS_HOST")
	}

	// These two URLs talk to GCS directly, which needs the bucket name injected into the URL path
	if host == "storage.googleapis.com" && host == "storage.cloud.google.com" {
		path = fmt.Sprintf("%s/%s", u.BucketName, u.artifactPath(artifact))
	}
```

Except that this still doesn't account for anyone [using a CNAME to talk to GCS directly](https://cloud.google.com/storage/docs/request-endpoints#cname) via a custom DNS entry with cookie auth.

Hmm.

One option could be to add a flag to the agent for this (ie `BUILDKITE_GCS_NONSTANDARD_PATH` or the like that bypasses default GCS path construction (which I'd also be happy to do, if that's the right answer).  WDYT @lox?

lox:
What if we added a new `BUILDKITE_GCS_PATH` var that could be used to override the path?

DoomGerbil:
> What if we added a new `BUILDKITE_GCS_PATH` var that could be used to override the path?

(Sorry, was OOO for a week).  

That makes sense to me.  I'll modify this PR and send it back for review.

My plan is:
1. Add a new CLI flag (`BUILDKITE_GCS_PATH`) that contains a path fragment to prefix to any GCS paths.
	1. By default, this will match the GCS standard prefix (ie `$GCS_BUCKET_NAME`)
1. Check this flag in the above func using [`os.LookupEnv`](https://golang.org/pkg/os/#LookupEnv) instead of [`os.GetEnv`](https://golang.org/pkg/os/#Getenv).
	1. If set, prepend the value to `artifactPath`.
	1. If not set, use the existing path construction logic.

This way anyone who doesn't set the `BUILDKITE_GCS_PATH` flag will continue to get the same behavior as today, and we'll be able to override it to get the behavior we need.

agent/gs_uploader.go
```diff
@@ -90,10 +90,14 @@ func (u *GSUploader) URL(artifact *api.Artifact) string {
...
-	if os.Getenv("BUILDKITE_GCS_ACCESS_HOST") != "" {
-		host = os.Getenv("BUILDKITE_GCS_ACCESS_HOST")
-		path = u.artifactPath(artifact)
+	// Override the default host if required.
+	if envHost, ok := os.LookupEnv("BUILDKITE_GCS_ACCESS_HOST"); ok {
+		host = envHost
+	}
+
+	// If this is set, we trust the user to supply a valid path prefix, and don't use a default GCS path
+	if prefix, ok := os.LookupEnv("BUILDKITE_GCS_PATH_PREFIX"); ok {
```
petemounce: Validate that it starts with a /?

lox:
Looks good to me, @DazWorrall any objections?

agent/gs_uploader.go
DoomGerbil: It shouldn't start with a `/` - https://golang.org/pkg/net/url/#URL specifies that `/` prefixed paths are treated differently than those without, but that both are valid.  The existing behavior does not have a leading `/`, so I'd rather not change that.

This does, however, raise the valid point that there will always need to be exactly one `/` between `prefix` and `u.artifactPath(artifact)`, so I'll change this to ensure that is always true.

DazWorrall:
> Looks good to me, @DazWorrall any objections?

Looks like the existing behaviour is preserved so we're happy with this change. Would be good to see some tests around this function to confirm we're getting the behaviour we expect.

DoomGerbil:
> > Looks good to me, @DazWorrall any objections?
> 
> Looks like the existing behaviour is preserved so we're happy with this change. Would be good to see some tests around this function to confirm we're getting the behaviour we expect.

Cool, that's fair - thanks for the review!  I'll refactor a bit for testability and add some tests around ensuring that paths are being set up properly, but won't be able to do proper integration tests without a mock GCS client.

agent/gs_uploader.go
```diff
@@ -87,15 +87,28 @@ func newGoogleClient(scope string) (*http.Client, error) {
...
+	}
+
+	// If this is set, we trust the user to supply a valid path prefix, instead of using the default GCS path
+	if prefix, ok := os.LookupEnv("BUILDKITE_GCS_PATH_PREFIX"); ok {
+		pathPrefix = prefix
 	}
 
+	// Build the path from the prefix and the artifactPath
+	// Also ensure that we always have exactly one / between prefix and artifactPath
+	path := fmt.Sprintf("%s/%s", strings.TrimSuffix(pathPrefix, "/"), u.artifactPath(artifact))
```
lox: [path.Join](https://golang.org/pkg/path/#Join) would clear this up

lox:
This is looking great @DoomGerbil, I'm going to do some work to try and get mocks into more places throughout the agent code over the next few months too.

lox:
Let's merge this now and then re-visit with some tests across the downloaders and uploaders.

PR #1001, opened 2019-05-07
Don't force set the executable bit on scripts to be set

kuroneko:
At present, buildkite sets the executable bit on any script it attempts to
execute as part of a job.

This is actually quite hazardous as there may be files that are valid
interpreter scripts, but are dangerous to run -- by having this behaviour, it
actually subverts any good that no-command-eval can achieve by leaving another
potentially gaping hole which we can't turn off!

It's also relatively unnecessary - git, by default, tries to store the state of
the execute bit into the repo.  We should not subvert this if the user doesn't
want us to.

This patch adds a flag that can disable this behaviour, so, if you're security
minded, and you really need to be able to ensure that arbitrary files in your
repo cannot be invoked, without use of a whitelist, you have the ability to do
so!

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -1370,6 +1370,28 @@ func (b *Bootstrap) defaultCommandPhase() error {
 
 		cmdToExec = batchScript
 	} else if commandIsScript {
+		// If we're running without CommandEval, the usual reason is we're
+		// trying to protect the agent from malicious activity from outside
+		// (including from the master).
+		//
+		// Because without this guard, we'll try to make the named file +x,
+		// and then attempt to run it, irrespective of any git attributes,
+		// should the queue source/master be compromised, this then becomes a
+		// vector through which a no-command-eval agent could potentially be
+		// made to run code not desired or vetted by the operator.
+		//
+		// Such undesired payloads could be delivered by hiding that payload in
+		// non-executable objects in the repo (such as through partial shell
+		// fragments, or other material not intended to be run on it's own),
+		// or by obfuscating binary executable code into other types of binaries.
+		//
+		// This also closes the risk factor with agents where you
+		// may have a dangerous script committed, but not executable (maybe
+		// because its part of a deployment process), but you don't want that
+		// script to ever be executed on the buildkite agent itself!  With
+		// command-eval agents, such risks are everpresent since the master
+		// can tell the agent to do anything anyway, but no-command-eval agents
+		// shouldn't be vulnerable to this!
 		if b.Config.CommandEval {
 			// Make script executable
 			if err = addExecutePermissionToFile(pathToCommand); err != nil {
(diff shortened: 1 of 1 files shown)


lox:
Thanks very much for this @kuroneko! 

> by having this behaviour, it actually subverts any good that no-command-eval can achieve by leaving another potentially gaping hole which we can't turn off!

Could you expand a bit on this? What would be an example of an attack on this?

> This patch adds a flag that can disable this behaviour

If we can, I'd love to remove the behaviour entirely if we think it's exploitable!

kuroneko:
The issue created is that now any file in your repository becomes a potential entrypoint for the agent to use, not just the ones you intended, subverting the general UNIX file permissions mechanism (and the level to which Git supports that)..

All attack vectors depend upon either the buildkite masters being compromised and/or impersonated, or attacks via the pipeline definition.  The pipeline definition attack is less of a concern for us since it's usually visible - Our concerns were more oriented towards the former.  The risks of this aren't treated here because they're outside of our control as the endusers of your service.

I assume this is the vector that no-command-eval was intended to protect against too.

The problem is, because the behaviour is one where any file can become set +x (or at least, that's what it looked like to me in my quick analysis of the behaviour in bootstrap), including scripts not intended to be executable, or binaries that just happen to have valid executable headers, it becomes possible for an attacker to try to introduce a payload they can exploit as a seemingly innocent part of another change, but doesn't look exploitable because it's a fragment of another part that's missing the execute bit, or because it's obfuscated into another type of binary, and then through any failure to control what is passed to a no-command-eval agent, invoke that payload.

lox:
> it becomes possible for an attacker to try to introduce a payload they can exploit as a seemingly innocent part of another change, but doesn't look exploitable because it's a fragment of another part that's missing the execute bit, or because it's obfuscated into another type of binary, and then through any failure to control what is passed to a no-command-eval agent, invoke that payload.

That makes a lot of sense, thanks! 

Will discuss with the team what approach we want to take. Appreciate the write up!

lox:
What would you think of disabling the setting of the executable bit when `no-command-eval` is set? That way we avoid another security config.

kuroneko:
> What would you think of disabling the setting of the executable bit when `no-command-eval` is set? That way we avoid another security config.

Sounds good to me - I only put in an extra option in the event that somebody relied on the combinational behaviour.

kuroneko:
New commit now only relies on the existing CommandEval flag, eliminating all the config change requirements.

bootstrap/bootstrap.go
```diff
@@ -1370,10 +1370,12 @@ func (b *Bootstrap) defaultCommandPhase() error {
 
 		cmdToExec = batchScript
 	} else if commandIsScript {
-		// Make script executable
-		if err = addExecutePermissionToFile(pathToCommand); err != nil {
-			b.shell.Warningf("Error marking script %q as executable: %v", pathToCommand, err)
-			return err
+		if b.Config.CommandEval {
```
lox: Could I be difficult and ask that you add a comment with some context around why we only do this in CommandEval and why it's a security risk? Will help us remember in 12 months time when we look in on it.

lox:
Thanks for the contribution and interesting discussion @kuroneko!

PR #1007, opened 2019-05-10
Refactor handling pings into main agent worker run loop

lox:
Currently the `AgentWorker.Ping()` method also handles the ping results and runs a job if one is included. This makes it hard to do things like add locking around ping, or even just find where the code is that runs a job. 

This PR makes the `Ping()` method simply return an `api.Ping`, which is then processed in the main loop in `AgentWorker.Start()`.


lox:
Gonna go a different way on this one I think, didn't end up helping after a re-read.

PR #1009, opened 2019-05-12
Deprecate disconnect-after-job-timeout

lox:
Originally the agent had two disconnect settings: `disconnect-after-job`, which caused the agent to disconnect after a job, and `disconnect-after-job-timeout` which was a timeout prior to receiving a job that if no job was received by would cause the agent to disconnect. 

We subsequently added `disconnect-after-idle-timeout` which disconnects the agent after it's been idle at any time, either before or after a job. 

This PR deprecates `disconnect-after-job-timeout` and suggests using `disconnect-after-idle-timeout` instead, as it simplifies a lot of code paths. Functionally, they are very similar.

### DIFF
--- a/agent/agent_configuration.go
+++ b/agent/agent_configuration.go
@@ -23,7 +23,6 @@ type AgentConfiguration struct {
 	RunInPty                   bool
 	TimestampLines             bool
 	DisconnectAfterJob         bool
-	DisconnectAfterJobTimeout  int
 	DisconnectAfterIdleTimeout int
 	CancelGracePeriod          int
 	Shell                      string
--- a/agent/agent_worker.go
+++ b/agent/agent_worker.go
@@ -57,12 +57,6 @@ type AgentWorker struct {
 	// Whether to enable debug
 	debug bool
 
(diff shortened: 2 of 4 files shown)


lox:
Thoughts on whether this ends up being a breaking change @toolmantim?

toolmantim:
When you say deprecated, does that mean it's marked as deprecated in `--help` / docs, and outputs a message, but still works for now? In which case, that's not a breaking change I think.

Perhaps then it'd be removed in the next major version, or the version after that?

lox:
@toolmantim yeah, marked as deprecated and shows a warning.

harrietgrace:
The `disconnect-after-job-timeout` Agent config option is listed as deprecated in the docs:

image

lox:
Thanks @harrietgrace 🙇🏼‍♂️

scalp42:
@harrietgrace the documentation does not mention `disconnect-after-idle-timeout` though

harrietgrace:
Thanks for pointing that out @scalp42! 👌🏻 I've PR'ed that addition just now.

PR #1013, opened 2019-05-15
update path parsing to support windows

GaryPWhite:
Path parsing for any Windows path was resulting in a %20 being injected into the URL, and would fail to upload in some cases to Artifactory.

### DIFF
--- a/agent/arty_uploader.go
+++ b/agent/arty_uploader.go
@@ -87,7 +87,7 @@ func (u *ArtifactoryUploader) URL(artifact *api.Artifact) string {
 	// ensure proper URL formatting for upload
 	url.Path = strings.Join([]string{
 		strings.Trim(url.Path, "/"),
-		u.artifactPath(artifact),
+		strings.Replace(u.artifactPath(artifact), "\\", "/", -1),
 	}, "/")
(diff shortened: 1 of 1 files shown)


lox:
Hmmm, this is a tricky one. I suspect we need to be handling this slightly differently and letting golang do some of the heavy lifting on path normalization. 

Reading the code again, it seems like `artifactPath()` is only called in `URL()`, let's just move the logic in there to make it easier to figure out what is going on?

```go
func (u *ArtifactoryUploader) URL(artifact *api.Artifact) string {
	url := *u.iURL

	// update the path to point to the repository, path and artifact path
	url.Path = path.Join(u.Repository, u.Path, filepath.ToSlash(artifact.Path))

	return url.String()
}
```

lox:
(Note that in golang `path` is for generic forward slash paths, where as `filepath` depends on the OS)

GaryPWhite:
Updated with proposed heavy lifting changes, didn't insert u.Repository because URL would then be having repository inserted twice per our implementation.

lox:
Makes sense, in retrospect my code snippet didn't make much sense, but you got the good bits form it  😅

PR #1014, opened 2019-05-16
update path parsing to support windows artifact download 

MartinRudzki:
This compliments PR(https://github.com/buildkite/agent/pull/1013) for windows upload, both need to be merged for artifactory to work properly.

### DIFF
--- a/agent/arty_downloader.go
+++ b/agent/arty_downloader.go
@@ -6,10 +6,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
-<<<<<<< HEAD
 	"path"
-=======
->>>>>>> 45ad5d0c7b4032e08718813b97b812d1dba5b3ba
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


agent/arty_downloader.go
```diff
@@ -78,7 +79,7 @@ func (d ArtifactoryDownloader) Start() error {
 
 func (d ArtifactoryDownloader) RepositoryFileLocation() string {
 	if d.RepositoryPath() != "" {
-		return strings.TrimSuffix(d.RepositoryPath(), "/") + "/" + strings.TrimPrefix(d.conf.Path, "/")
+		return strings.TrimSuffix(d.RepositoryPath(), "/") + "/" + strings.TrimPrefix(filepath.ToSlash(d.conf.Path), "/")
```
lox: Can we use `path.Join` here too?

lox:
Looks like a rogue .DS_Store is included here, is it possible to remove it?

agent/arty_downloader.go
MartinRudzki: Sure will get it out soon!

lox:
Thanks very much @MartinRudzki!

PR #1020, opened 2019-05-21
Refactor api package to an interface

lox:
Previously the api client used different sub-structs for different "services", which ended up being very hard to test. This folds it down into methods on the `api.Client` struct, which is less modular, but easier to mock in tests (and understand).

Additionally, the api client construction lived in the `agent` package. This is moved into the `api` package so everything lives together.

A basic test was added for the api client.


lox:
@sj26 could I walk you through this one?

sj26:
Yup @lox, grab me when you'd like to go through it. Maybe tomorrow, if we'll both be in league?

Issue #422, opened 2017-01-09
Agent should update to the latest version in some cases


lox:
Currently the agent will download a plugin for something like `blah-blah-plugin` from the master branch the first time it needs it, but it won't update it ever again.

This means that go get your changes as you are iterating, you end up updating your pipeline to reference specific commits like `blah-blah-plugin#{commit-hash}`. 

The meta-effect of this is that it's very hard to get your plugin users hotfixes on master. They need to delete the plugin or change their pipeline.

It would be great to fetch the latest version on each job if the plugin isn't a ref or a symbolic ref /cc @sj26.

lox:
For now the conclusion in discussion with @keithpitt is that we should encourage people to use tags vs branches!

vektah:
I agree!

For public plugins tags are :ok_hand:, you don't want unexpected changes to external repos breaking builds.

For internal plugins you really want this instant cascade, maybe following master, maybe following a go style branch for breaking changes (v1, v2). In this case it should always pull the latest ref for that branch.

mipearson:
Assuming that this runs on every step, how much time will checking that the tag/brach is up to date over multiple plugins take?

Github can be .. slow, sometimes.

lox:
Good question @mipearson! That was why we didn't implement it this way initially.

mipearson:
... if it's done inside agent (ie, in Go), could be easily done in parallel.

(and I wonder if a similar thing can be done with the s3 secrets plugin which is another slows-down-every-step point)

lox:
Yeah, most likely. Will experiment.

In terms of the s3 secrets plugin, I just think that s3 is not a good spot for secrets. It's very hard to make efficient. SSM or Secrets Manager is the future there IMO.

tigris:
FWIW you can _hack_ this yourself by doing a `git fetch v1` (or whatever "stable" branch you're using, perhaps you can grok the branch from vars) as early as possible in the hook chain. E.g. the environment hook.

The downside is you now can't use the environment hook for anything else.

Plus the obvious existing downsides of using a mutable branch, including but not limited to non-repeatable builds, slower build steps, etc.

But the upside of instant release. Great for things you deem internally to be sufficient for the risks, e.g. vulnerability scanning. If you trust the source code chain of access, then I can see this being quite beneficial.

evandam:
:+1: for this, not sure if there have been any recent developments.

[...]

PR #1025, opened 2019-06-09
Support plugin SHA-1 digests

toolmantim:
This adds the ability to specify a plugin git commit SHA-1 (ala `git show-ref` / `git rev-parse`), preventing any modification to the plugin after it has been added to the pipeline.yml file:

```yml
steps:
  - command: |
      node -e 'console.log("Hello world")'
    plugins:
      docker#v3.2.0@git-sha1:f199bc812a7d7c770139fe8f67203010e103d0f8:
        image: node
```

If the digest matches then this is the output, and everything continues hunky dory:

image

If the SHA-1 of the plugin checkout doesn't match the SHA-1 specified in the pipeline.yml, you get

image

Some design notes:

* The intention is that this is the new default for how people specify plugins in their pipeline.yml files. Digests should be baked into the default usage for everyone, helping increase the security of the entire ecosystem, just like it is bundler/npm/yarn/etc. Rather than something that can be specified for those who are worried.
* I can easily add support for the commit SHA updating into https://github.com/toolmantim/boomper, which we're using to help keep the plugin readme version numbers up-to-date. This means we retain the same API for usage (copying+pasting from readme markdown file examples), and no extra maintenance burden for plugin authoring.
* The `git-sha1` syntax was chosen over just specifying the SHA1, so we're explicitly stating what/where the SHA1 comes from. i.e. `docker#v3.2.0@git-sha1:f199bc812a7d7c770139fe8f67203010e103d0f8` is assumed to be clearer, and more self-documenting, than `docker#v3.2.0@f199bc812a7d7c770139fe8f67203010e103d0f8`
* There's nothing to change on the backend side, it just passes the plugin string all the way back through to the agent to handle. I've given it a test, and works well.
* Running this new digest plugin step on an older agent will fail (e.g. `error: pathspec 'v3.2.0@git-sha1:f199fc812a7d7c770139fe8f67203010e103d0f8' did not match any file(s) known to git.`). I believe this behaviour is okay (even desired) because if you're specifying digests then you want to make sure they're being enforced and not just ignored. (We can auto-detect the error in the build output, same as we do SSH key problems).
* Providing a digest here should avoid the need to vendor plugins for the sake of ensuring they're not tampered with (e.g. a tag is force-pushed, or the repo taken over)

- [x] Works on backend (no changes needed)
- [x] Initial implementation
- [x] Feedback on overall design
- [x] Error on incorrect digest format (e.g. `docker#v3.2.0@invalid:f199bc812a7d7c770139fe8f67203010e103d0f8:`). At the moment invalidly formatted digests are silently ignored, which is bad because they might appear correct to the eye but are silently ignored.
- [ ] Tests
- [ ] Code review

### DIFF
--- a/agent/agent_configuration.go
+++ b/agent/agent_configuration.go
@@ -19,6 +19,7 @@ type AgentConfiguration struct {
 	CommandEval                bool
 	PluginsEnabled             bool
 	PluginValidation           bool
+	RequirePluginDigests       bool
 	LocalHooksEnabled          bool
 	RunInPty                   bool
 	TimestampLines             bool
--- a/agent/plugin/plugin.go
+++ b/agent/plugin/plugin.go
@@ -19,8 +19,11 @@ type Plugin struct {
 	// The version of the plugin that should be running
 	Version string
 
-	// The digest of the plugin that should match the checkout
-	Digest string
+	// The digest type, e.g. "git-sha1"
+	DigestType string
+
+	// The digest value of the plugin
+	DigestValue string
 
 	// The clone method
 	Scheme string
@@ -37,7 +40,7 @@ type Plugin struct {
 
 var (
 	vendoredRegex = regexp.MustCompile(`^\.`)
-	versionRegex  = regexp.MustCompile(`([^@]*)(?:@git-sha1:(.*))?`)
+	versionRegex  = regexp.MustCompile(`([^@]*)(?:@(.*):(.*))?`)
 )
 
 func CreatePlugin(location string, config map[string]interface{}) (*Plugin, error) {
@@ -55,14 +58,20 @@ func CreatePlugin(location string, config map[string]interface{}) (*Plugin, erro
 	versionGroups := versionRegex.FindStringSubmatch(u.Fragment)
 	if versionGroups != nil {
 		plugin.Version = versionGroups[1]
-		plugin.Digest = versionGroups[2]
+		plugin.DigestType = versionGroups[2]
+		plugin.DigestValue = versionGroups[3]
 	} else {
 		plugin.Version = ""
-		plugin.Digest = ""
+		plugin.DigestType = ""
+		plugin.DigestValue = ""
 	}
 
 	if plugin.Version != "" && strings.Count(plugin.Version, "#") > 0 {
 		return nil, fmt.Errorf("Too many #'s in \"%s\"", location)
+	}
+
+	if plugin.DigestType != "" && plugin.DigestType != "git-sha1" {
+		return nil, fmt.Errorf("Unsupported plugin digest type \"%s\" in \"%s\"", plugin.DigestType, u.Fragment)
 	}
 
 	if u.User != nil {
--- a/agent/plugin/plugin.go
+++ b/agent/plugin/plugin.go
@@ -19,6 +19,9 @@ type Plugin struct {
 	// The version of the plugin that should be running
 	Version string
 
+	// The digest of the plugin that should match the checkout
+	Digest string
+
 	// The clone method
 	Scheme string
 
@@ -34,6 +37,7 @@ type Plugin struct {
 
 var (
 	vendoredRegex = regexp.MustCompile(`^\.`)
+	versionRegex  = regexp.MustCompile(`([^@]*)(?:@git-sha1:(.*))?`)
 )
 
 func CreatePlugin(location string, config map[string]interface{}) (*Plugin, error) {
@@ -46,8 +50,16 @@ func CreatePlugin(location string, config map[string]interface{}) (*Plugin, erro
 
 	plugin.Scheme = u.Scheme
 	plugin.Location = u.Host + u.Path
-	plugin.Version = u.Fragment
 	plugin.Vendored = vendoredRegex.MatchString(plugin.Location)
+
+	versionGroups := versionRegex.FindStringSubmatch(u.Fragment)
+	if versionGroups != nil {
+		plugin.Version = versionGroups[1]
+		plugin.Digest = versionGroups[2]
+	} else {
+		plugin.Version = ""
+		plugin.Digest = ""
+	}
 
 	if plugin.Version != "" && strings.Count(plugin.Version, "#") > 0 {
 		return nil, fmt.Errorf("Too many #'s in \"%s\"", location)
--- a/agent/plugin/plugin.go
+++ b/agent/plugin/plugin.go
@@ -33,8 +33,7 @@ type Plugin struct {
 }
 
 var (
-	locationSchemeRegex = regexp.MustCompile(`^[a-z\+]+://`)
-	vendoredRegex       = regexp.MustCompile(`^\.`)
+	vendoredRegex = regexp.MustCompile(`^\.`)
 )
 
 func CreatePlugin(location string, config map[string]interface{}) (*Plugin, error) {
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -723,14 +723,14 @@ func (b *Bootstrap) checkoutPlugin(p *plugin.Plugin) (*pluginCheckout, error) {
 	headCommit, err := gitRevParseInWorkingDirectory(b.shell, directory, "HEAD")
 
 	// Verify plugin digest
-	if p.Digest != "" {
-		if headCommit != p.Digest {
+	if p.DigestType == "git-sha1" {
+		if headCommit != p.DigestValue {
 			if p.Version != "" {
-				return nil, fmt.Errorf("Plugin digest mismatch. Expected git tag %s to have SHA1 of %s, found %s", p.Version, p.Digest, headCommit)
+				return nil, fmt.Errorf("Plugin digest mismatch. Expected git tag %s to have SHA1 of %s, found %s", p.Version, p.DigestValue, headCommit)
 			}
-			return nil, fmt.Errorf("Plugin digest mismatch. Expected git SHA1 of %s, found %s", p.Digest, headCommit)
+			return nil, fmt.Errorf("Plugin digest mismatch. Expected git SHA1 of %s, found %s", p.DigestValue, headCommit)
 		}
-		b.shell.Commentf("Plugin digest matches %s", p.Digest)
+		b.shell.Commentf("Plugin digest matches %s", p.DigestValue)
 	}
 
 	// Has it already been checked out?
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -718,17 +718,29 @@ func (b *Bootstrap) checkoutPlugin(p *plugin.Plugin) (*pluginCheckout, error) {
 		HooksDir:    filepath.Join(directory, "hooks"),
 	}
 
+	// It'd be nice to show the current commit of the plugin, so
+	// let's figure that out.
+	headCommit, err := gitRevParseInWorkingDirectory(b.shell, directory, "HEAD")
+
+	// Verify plugin digest
+	if p.Digest != "" {
+		b.shell.Commentf("Verifying plugin digest equals %s", p.Digest)
+		if headCommit != p.Digest {
+			if p.Version != "" {
+				return nil, fmt.Errorf("Plugin digest mismatch. Expected git tag %s to have SHA1 of %s, found %s", p.Version, p.Digest, headCommit)
+			} else {
+				return nil, fmt.Errorf("Plugin digest mismatch. Expected git SHA1 of %s, found %s", p.Digest, headCommit)
+			}		
+		}
+	}	
+
 	// Has it already been checked out?
 	if fileExists(pluginGitDirectory) {
-		// It'd be nice to show the current commit of the plugin, so
-		// let's figure that out.
-		headCommit, err := gitRevParseInWorkingDirectory(b.shell, directory, "--short=7", "HEAD")
 		if err != nil {
 			b.shell.Commentf("Plugin %q already checked out (can't `git rev-parse HEAD` plugin git directory)", p.Label())
 		} else {
 			b.shell.Commentf("Plugin %q already checked out (%s)", p.Label(), strings.TrimSpace(headCommit))
 		}
-
 		return checkout, nil
 	}
 
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -722,6 +722,10 @@ func (b *Bootstrap) checkoutPlugin(p *plugin.Plugin) (*pluginCheckout, error) {
 	// let's figure that out.
 	headCommit, err := gitRevParseInWorkingDirectory(b.shell, directory, "HEAD")
 
+	if b.RequirePluginDigests && (p.DigestType == "" || p.DigestValue == "") {
+		return nil, fmt.Errorf("Agent configuration requires plugin digests, but no digest found for %s", p.Name())
+	}
+
 	// Verify plugin digest
 	if p.DigestType == "git-sha1" {
 		if headCommit != p.DigestValue {
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -724,15 +724,14 @@ func (b *Bootstrap) checkoutPlugin(p *plugin.Plugin) (*pluginCheckout, error) {
 
 	// Verify plugin digest
 	if p.Digest != "" {
-		b.shell.Commentf("Verifying plugin digest equals %s", p.Digest)
 		if headCommit != p.Digest {
 			if p.Version != "" {
 				return nil, fmt.Errorf("Plugin digest mismatch. Expected git tag %s to have SHA1 of %s, found %s", p.Version, p.Digest, headCommit)
-			} else {
-				return nil, fmt.Errorf("Plugin digest mismatch. Expected git SHA1 of %s, found %s", p.Digest, headCommit)
-			}		
+			}
+			return nil, fmt.Errorf("Plugin digest mismatch. Expected git SHA1 of %s, found %s", p.Digest, headCommit)
 		}
-	}	
+		b.shell.Commentf("Plugin digest matches %s", p.Digest)
+	}
 
 	// Has it already been checked out?
 	if fileExists(pluginGitDirectory) {
--- a/bootstrap/config.go
+++ b/bootstrap/config.go
@@ -85,6 +85,9 @@ type Config struct {
 	// Whether to validate plugin configuration
 	PluginValidation bool
 
+	// Whether there must be a valid digest to use a plugin
+	RequirePluginDigests bool
+
 	// Are local hooks enabled?
 	LocalHooksEnabled bool
 
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -75,6 +75,7 @@ type AgentStartConfig struct {
 	NoPlugins                  bool     `cli:"no-plugins"`
 	NoPluginValidation         bool     `cli:"no-plugin-validation"`
 	NoPTY                      bool     `cli:"no-pty"`
+	RequirePluginDigests       bool     `cli:"require-plugin-digests"`
 	TimestampLines             bool     `cli:"timestamp-lines"`
 	MetricsDatadog             bool     `cli:"metrics-datadog"`
 	MetricsDatadogHost         string   `cli:"metrics-datadog-host"`
@@ -330,6 +331,11 @@ var AgentStartCommand = cli.Command{
 			EnvVar: "BUILDKITE_NO_GIT_SUBMODULES,BUILDKITE_DISABLE_GIT_SUBMODULES",
 		},
 		cli.BoolFlag{
+			Name:   "require-plugin-digests",
+			Usage:  "Require all plugins to be pinned to a version and digest",
+			EnvVar: "BUILDKITE_REQUIRE_PLUGIN_DIGESTS",
+		},
+		cli.BoolFlag{
 			Name:   "metrics-datadog",
 			Usage:  "Send metrics to DogStatsD for Datadog",
 			EnvVar: "BUILDKITE_METRICS_DATADOG",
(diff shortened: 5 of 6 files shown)


lox:
This is looking awesome!

toolmantim:
One concern I've thought of, is that if the readmes of the plugins are updated to use this new syntax, people might copy+paste them and try to run it on an older agents that don't support it. In this case, they'll get the error `pathspec 'v3.2.0@git-sha1:f199fc812a7d7c770139fe8f67203010e103d0f8' did not match any file(s) known to git.`. We can auto-detect this error message in the build output in the web UI, and show a useful helpful message with links to the docs, same as we do for ssh key permissions problems.

We should also ensure we've released a new Elastic Stack with support, in advance of changing the readmes.

lox:
@toolmantim could we check for the agent version on the server side and trim off the digest for agents that don't support it? We do pre-processing of the plugins currently. 

We could show an error on the build to indicate that the agent didn't support the digest pinning.

toolmantim:
Yeah we could, I thought of that too. I reckon you'd want the job to fail still though yeah? So either you fail the job on the backend with some fake build log output somehow or extra UI logic, or you just let it fail on the old agent but also show a nicer error message in the web UI like we do with SSH problems.

lox:
> or you just let it fail on the old agent but also show a nicer error message in the web UI like we do with SSH problems.

I reckon that is the way to go!

lox:
Do we want to allow partial sha1 matches? E.g let folks provide 7-10 digits of the sha1 prefix?

toolmantim:
> Do we want to allow partial sha1 matches? E.g let folks provide 7-10 digits of the sha1 prefix?

For the sake of security, I was thinking not. I wasn't sure what guarantees you get with the truncated string. Any idea?

I just did a quick search on the crypto guarantees Git gives you, and found this disheartening advice:
https://en.wikipedia.org/wiki/SHA-1#Data_integrity

Which links to: https://marc.info/?l=git&m=115678778717621&w=2

But, it looks like Git is transitioning to SHA256 since that, and that the SHA-1 implementation is hardened against SHAttered:
https://github.com/git/git/blob/master/Documentation/technical/hash-function-transition.txt

We could go with the SHA-1 for now, and support a `git-sha256:xxx` digest in the future? Not sure if this meets people's needs and expectations? I would say it might, but I don't want to presume.

The alternative would be finding another way to specify and calculate the digests, but it's mighty less convenient than being able to just use Git's full SHA-1 values:

```
$ curl https://api.github.com/repos/buildkite-plugins/docker-compose-buildkite-plugin/git/refs/tags/v3.0.3
{
  "ref": "refs/tags/v3.0.3",
  "node_id": "MDM6UmVmNDUwMTc0NDA6djMuMC4z",
  "url": "https://api.github.com/repos/buildkite-plugins/docker-compose-buildkite-plugin/git/refs/tags/v3.0.3",
  "object": {
    "sha": "f7f2d362fe37c51f7f42f485385314b412416e17",
    "type": "commit",
    "url": "https://api.github.com/repos/buildkite-plugins/docker-compose-buildkite-plugin/git/commits/f7f2d362fe37c51f7f42f485385314b412416e17"
  }
}
```

toolmantim:
I added a few things, but I couldn't quite get the new `--require-plugin-digests` to work — it seems nil/false in the bootstrap 🤔

Would love a hand getting this finished up @lox!

lox:
I'll help get it across the line this week @toolmantim!

lox:
@toolmantim interesting thinking on "--require*" I was imagining that we'd go for `--no-plugins --allow-plugins-with-digests --allow-vendored-plugins`. 

Thoughts on which direction you prefer?

lox:
How would you only allow vendored plugins with the `--require*` syntax I wonder. It seems to not lend itself very well to allowing multiple different types of "secure" plugin methods.

lox:
The allow syntax also opens up things like `--no-plugins --allow-plugins-with-org=buildkite --allow-plugins-with-digests --allow-vendored-plugins`.

lox:
Although that is ambiguous as to whether digest plugins are required to be `buildkite` org 🤔

lox:
`--plugin-condition "(plugin.organization == 'buildkite' && plugin.digest != nil) || plugin.vendored"`? 😅

lox:
(We could have that very easily)

lox:
Shall we merge this without the flag @toolmantim and look at something like https://github.com/buildkite/agent/pull/1034 next?

toolmantim:
> Shall we merge this without the flag @toolmantim and look at something like #1034 next?

Yeah, you're totally right! We should just add this support for it first!

toolmantim:
> `--no-plugins --allow-plugins-with-org=buildkite --allow-plugins-with-digests --allow-vendored-plugins`

I really like this one I think! 👌🏼

toolmantim:
> > `--no-plugins --allow-plugins-with-org=buildkite --allow-plugins-with-digests --allow-vendored-plugins`
> I really like this one I think! 👌🏼

Actually, I was hoping that `--require-plugin-digests` would become the new default in a future version of the agent, with a warning perhaps for non-vendored plugins without digests. It's sort of separate to the idea of "allowing".

So if we wanted to recommend a new secured agent setup, it might not use `--no-plugins` at all, but perhaps:
* `--require-plugin-digests --allowed-plugin-hosts=https://github.com/my-org`
* `--require-plugin-digests --allowed-plugin=https://github.com/my-org/my-plugin`
* `--require-plugin-digests --allowed-plugin-hosts=https://github.com/my-org --allowed-plugin=https://github.com/other-org/other-plugin`

And in future agent versions we could drop the `--allowed-plugin-hosts` part altogether, so a common secure agent setup might be just `--allowed-plugin-hosts=https://github.com/my-org` — which would require plugin digests, only allow your own orgs plugins, and also allow vendored plugins.

lox:
I reckon it’s going to get too complicated to do with flags @toolmamtim. I’m still also a bit unsure about whether requiring everyone use digests should be a thing, but keen to see how it plays out.

keithpitt:
I'd like to hear more about this! Specifically the approached taken in this PR vs:

> Providing a digest here should avoid the need to vendor plugins for the sake of ensuring they're not tampered with (e.g. a tag is force-pushed, or the repo taken over)

Asking folks to figure out the magic to make `docker#v3.2.0@git-sha1:f199bc812a7d7c770139fe8f67203010e103d0f8` work I think sounds a little complicated?

toolmantim:
> I reckon it’s going to get too complicated to do with flags

Fair call!

> I’m still also a bit unsure about whether requiring everyone use digests should be a thing

The intention of the PR wasn't just "Add the ability to specify a SHA in the pipeline.yml if you're worried about security" but more "Figure out a way forward where SHA digests are the default".

Perhaps I'll put down some of the thinking / context, because I'm not entirely sure of the overall future strategy for plugin installation and usage, so I'm not sure if 

I can understand the tension from the point of view that it's a human authored file (rather than a `yarn.lock` for example). And if you're using a plugin from your own private GHE install for example, or your own public GitHub organization, then it might feel like overkill (although if it's your own open-source GitHub plugin, we still want the README examples to include pinning for other peeps).

I guess our situation is similar to a Dockerfile, in that it's human authored, and there's no additional lockfile that's used to pin things down (unlike most other dep systems, like RubyGems/NPM/Yarn). Few people seem to use Docker SHA pinning, which means that the whole ecosystem is insecure-by-default. Remembering to add a SHA, or deciding if you should, is left as an exercise to the user every time they do it—some teams add Dockerfile linters to require it.

Ours is different to the Dockerfile declaration though, because Dockerfile tags are also used for sem-ver like fuzzy version matching (e.g. `FROM node` or `FROM node:12`) and the image potentially changing depending on what machine you're running it on. With Docker it's common to push an existing a tag with entirely new content too.   With `pipeline.yml` we recommend always using a full tag it's the same plugin version on all jobs, and that it doesn't just break without you changing anything. This means we recommend people bump plugin version numbers manually, but with the help of tools like Renovate/Dependabot to make it easier, and potentially via BK CLI too.

We don't support sem-ver like deps, and the agent doesn't even pull down the latest master if you refer to just `docker-compose` rather than `docker-compose#v3.2.0`. So I feel the format/design of the system already lends itself to full pinning, with the goal/assumption that the plugin code should never change without you touching the the pipeline.yml.

Another thing I'm trying to think though is that our install method is copy+pasting from GitHub readmes into the pipeline.yml. Because the readmes aren't in control, it's been up to the authors to put the full version numbers in there. Luckily it looks like most plugin authors are using our linter 🙌🏼, and that ensures that usage examples include an up-to-date version number, but they're not using https://github.com/toolmantim/boomper to automatically update the readme. Having to manually bump `#v3.0.3:git-sha1:f7f2d362fe37c51f7f42f485385314b412416e17` is harder than just `#v3.0.3` (though at least the GitHub UI shows the commit SHA for the release pretty easily). I think that the digest I've proposed here is still fairly accessible to plugin authors, and if we were to publish a `buildkite-plugin-toolkit` GitHub App/Action people can add to their plugin repos we could make it dead simple (it'd be a small rename and reworking of the existing Boomer app). There's a few other things that plugin authors would love to see in that app already too, like having it perform the linter checks for you too.

To sum up the assumptions with this PR so far:

- For now we're sticking with recommending pinned plugin versions in pipeline.yml files
- For now we're staying with a readme based install method
- We want people to include a digest for any third-party plugin use — which means that the copy+paste plugins from GitHub will need to include a digest — so we're secure by default rather than 'secure if you could be bothered'
- We have a good chance of moving plugin authors to including digests in their readme, through linter updates and a potentially new plugin toolkit GitHub Action/app

I reckon if we change enough of these assumptions, we should look at a potential new solution too. There's probably some alternative approaches to the same problem if we're thinking of changing direction.

toolmantim:
> I'd like to hear more about this! Specifically the approached taken in this PR vs:
>> Providing a digest here should avoid the need to vendor plugins for the sake of ensuring they're not tampered with (e.g. a tag is force-pushed, or the repo taken over)
> Asking folks to figure out the magic to make `docker#v3.2.0@git-sha1:f199bc812a7d7c770139fe8f67203010e103d0f8 work` I think sounds a little complicated?

Yeah, for sure! So vendored plugins solves it for those willing to do that. And there's other things you can do to make things secure too, like: https://github.com/buildkite/buildkite-allowed-plugins-hook-example. They all require a lot of work though, and only a small % of people will do it.

This was an attempt to try to make the default secure, for 99% of uses of plugins, in the same way you don't worry about RubyGems / NPM package security either (they have digests built into their lockfiles, so no-one can just force-push a new git tag to GitHub and own your production builds).

If vendoring plugins were the default way plugins are used, then it'd change everything. But I don't think that's the vision for plugins really, just an option for some.

toolmantim:
How about we start by adding this digest support to the pipeline.yml (as per the PR)? It's optional, and people can use agent hooks to enforce their existence if they want, same as they can for limiting the plugin repo hosts. And we can mention it in the securing guide.

I was already keen to build a "Buildkite Plugin Toolkit" GitHub Action that let's plugin authors keep their readmes up-to-date and SHA'd. At the moment we're using (and recommending) my personal Boomper GitHub app on all our plugins, which is bad. A GitHub Action means we don't have to run our own servers, and we don't have access to their repo. Plugin authors don't want to manually bump version numbers in their readme as it is, so it's pretty easy to do version + digest in there too.

I think these two things gets us 99% of the way to "secure by default" without this being too complicated. And I'm keen to work on the second one anyway, so I can retire my own GitHub App.

keithpitt:
I'm cool with the approach in general (as mentioned on the call).

I'd like to float a different syntax:

```
steps:
  - command: "go build -o dist/my-app ."
    artifact_paths: "./dist/my-app"
    plugins:
      - docker#v3.2.0+sha.f199bc8:
          image: "golang:1.11"
```

The things I like about this are:

1. It's got the semver tick of approval (so potentially the version can be used in other contexts)
2. It's shorter than the full blown SHA

I'm not sure however of how long a SHA needs to be for it to be "safe"

toolmantim:
@keithpitt 👍🏼

I was all excited and checked out the [semver spec](https://semver.org/#spec-item-10):

> Build metadata MAY be denoted by appending a plus sign and a series of dot separated identifiers immediately following the patch or pre-release version. Identifiers MUST comprise only ASCII alphanumerics and hyphen [0-9A-Za-z-]. Identifiers MUST NOT be empty. Build metadata SHOULD be ignored when determining version precedence. Thus two versions that differ only in the build metadata, have the same precedence. Examples: 1.0.0-alpha+001, 1.0.0+20130313144700, 1.0.0-beta+exp.sha.5114f85.

Looks like `sha.xxx` is only an example, and it's mostly about the `+` sign — and we can do anything after the `+` sign?

But I just realised, that trying to make it part of the "version" has the problem that it clashes with the plugin's actual version. i.e. If someone was to do a release of their plugin with the tag `v3.2.0+something` that would be totally valid semver, and would work in our plugin directory as is now (because we use semver checking), but wouldn't work with this syntax:

```yml
steps:
  - command: "go build -o dist/my-app ."
    artifact_paths: "./dist/my-app"
    plugins:
      - docker#v3.2.0+something+sha.f199bc8:
          image: "golang:1.11"
```

I think having our own clear delimiter between "this is the version part" and "this is the digest part" might be best?

lox:
Personally I really liked @toolmantim's use of the `@` like docker uses. I like at least using `sha1` as it allows us to move to other hash formats over time. 

I view it as independent of the version number FWIW. It's an extra annotation on there.

lox:
IMO this would be just as legit (excluding the branch/tag/version) in favor of the content addressable version. 

`docker@git-sha1:f199bc812a7d7c770139fe8f67203010e103d0f8`

lox:
Some context on how Docker does it: https://engineering.remind.com/docker-image-digests/

I like `docker@sha1:f199bc812a7d7c770139fe8f67203010e103d0f8`

matthewd:
> I think having our own clear delimiter between "this is the version part" and "this is the digest part" might be best?

> I like `docker@sha1:f199bc812a7d7c770139fe8f67203010e103d0f8`

:+1:

To me it's important to note that while the LHS is version-ish, it's technically a git refname, so we want to avoid a separator character that's likely to be used for that purpose. (To that end, I'd slightly favour a second `#` over `@`, because the latter [has existing meaning](https://www.git-scm.com/docs/gitrevisions/2.21.0#Documentation/gitrevisions.txt-emltrefnamegtltdategtemegemmasteryesterdayememHEAD5minutesagoem), but both feel more like punctuation than something like `+`.)

toolmantim:
Thanks for all the feedback! What about this?

```yml
steps:
  - command: node -v
    plugins:
      docker#v3.2.0@sha1:f199bc812a7d7c770139fe8f67203010e103d0f8:
        image: node
```

We could avoid the double-colon weirdness if we wanted, with something like a `-` character instead:

```yml
steps:
  - command: node -v
    plugins:
      docker#v3.2.0@sha1-f199bc812a7d7c770139fe8f67203010e103d0f8:
        image: node
```

I've also had a look around and I don't see anyone using short SHAs for security/digest related things, so I'm a bit nervous about anything other than the full SHA for what the goals of this is (i.e. it's not to identify the commit in the most friendly way, it's to verify the digest of the contents)

toolmantim:
(I updated the above comment with a bit of a clearer example, that shows a potential route for DRY'ing the plugin cruft)

sj26:
My vote would be adopting the Docker syntax because it will look familiar and has the same meaning (content-addressed version), and it seems sensible:

```
steps:
  - command: "dotnet publish -c Release -o published"
    plugins:
      - docker#v3.2.0@sha1:f199bc812a7d7c770139fe8f67203010e103d0f8:
          image: "microsoft/dotnet@sha1:7203010e103d0f8f199bc812a7d7c770139fe8f6"
```

I don't think the double colon is a problem.

sj26:
My vote would be adopting the Docker syntax because it will look familiar and has the same meaning (content-addressed version), and it seems sensible:

```
steps:
  - command: "dotnet publish -c Release -o published"
    plugins:
      - docker#v3.2.0@sha1:f199bc812a7d7c770139fe8f67203010e103d0f8:
          image: "microsoft/dotnet@sha1:7203010e103d0f8f199bc812a7d7c770139fe8f6"
```

I don't think the double colon is a problem.

toolmantim:
Aight, I reckon let's go with this!

```yml
steps:
  - command: "dotnet publish -c Release -o published"
    plugins:
      - docker#v3.2.0@sha1:f199bc812a7d7c770139fe8f67203010e103d0f8:
          image: "microsoft/dotnet:latest"
```

I'll get the PR fixed up.

I'd love to add a DRY version of it too, like this, but let's discuss it elsewhere:

```yml
steps:
  - command: "dotnet publish -c Release -o published"
    plugins:
      - docker:
          image: "microsoft/dotnet:latest"

plugins:
  docker: "docker#v3.2.0@sha1:f199bc812a7d7c770139fe8f67203010e103d0f8"
```

keithpitt:
I think if you use the double colon, this will older hash syntax will break:

```yaml
steps:
  - command: "dotnet publish -c Release -o published"
    plugins:
      docker#v3.2.0@sha1:f199bc812a7d7c770139fe8f67203010e103d0f8:
          image: "microsoft/dotnet:latest"
```

toolmantim:
According to https://yaml-online-parser.appspot.com that parses okay?

```json
{
  "steps": [
    {
      "command": "dotnet publish -c Release -o published", 
      "plugins": {
        "docker#v3.2.0@sha1:f199bc812a7d7c770139fe8f67203010e103d0f8": {
          "image": "microsoft/dotnet:latest"
        }
      }
    }
  ]
}
```

keithpitt:
Oh cool, I thought YAML wouldn't like it because it uses the `:` as a key.

If it works, 🚀 

Just to be clear, would this change (and associated community tool changes) start pushing the full SHA syntax into plugin definitions?

toolmantim:
> Just to be clear, would this change (and associated community tool changes) start pushing the full SHA syntax into plugin definitions?

What do you mean?

keithpitt:
As in, would `docker#v3.2.0@sha1:f199bc812a7d7c770139fe8f67203010e103d0f8` be the new best practice? Or just an extra security thing folks can do if they want?

toolmantim:
The idea at the moment is that it'd be the new best practice, and the default. I'd love to use short SHAs, but I don't think anyone's done the work to show that it's secure enough for what we need, so I've been defaulting to full SHA1s because Maths I Don't Understand™

lox:
I'll help out with some tests too. 

IMO let's roll with full SHA1 hashes. For cryptographic digests it's generally less of a thing to use prefixes. The risk isn't a collision, it's a complete bypass of the security that the hash is supposed to provide.

We're still a long way from practical chosen prefix attacks against SHA1, but it's not far off:

* https://www.zdnet.com/article/sha-1-collision-attacks-are-now-actually-practical-and-a-looming-danger/

lox:
@toolmantim so that we can get this shipped, how would you feel about ditching the parameter to require plugin digests?

I reckon https://github.com/buildkite/agent/pull/1034 is the right answer to handling that.

toolmantim:
Yeah, let’s remove the config argument.

toolmantim:
I'm working on a securing plugins doc with @harrietgrace — when this is merged we'll update that to recommend SHA pinning.

toolmantim:
@pda and I had a big discussion about this a while back, and came to the conclusion that if the key vector that we're wanting to protect from is a malicious force push of a third party plugin tag, perhaps a better approach would be to provide a global "known-SHA" value for a given plugin's tag, via the BK agent API. We can have the agent report to BK what the tag was after checking out each plugin, and have the BK API error if it doesn't match a previous value.

That way, we can keep the pipeline ergonomics of a friendly tag reference, not require all the readmes of all the plugins to change, and increase the default security of the entire plugin platform.

I really like this suggestion as a way of making things more secure by default, as opposed to having to explicitly opt to do things in a more secure way (use a CLI tool, remember to add digests manually, etc etc)

yob:
> key vector that we're wanting to protect from is a malicious force push of a third party plugin tag, perhaps a better approach would be to provide a global "known-SHA" value for a given plugin's tag, via the BK agent API. We can have the agent report to BK what the tag was after checking out each plugin, and have the BK API error if it doesn't match a previous value

sort of like a centralised (BK-administered) SSH known-hosts file?

How would plugin sha's get into that list - would we scrape the plugin repos on github and record the sha the first time we find a new branch/tag?

toolmantim:
> sort of like a centralised (BK-administered) SSH known-hosts file?

Yeah!

> How would plugin sha's get into that list - would we scrape the plugin repos on github and record the sha the first time we find a new branch/tag?

The thought was that the agents would send back the digest, via the agent API, after plugin checkouts and if we hadn't seen that plugin+tag we'd store it and return okay, or if exists then check if it matches and return okay or error.

lox:
How would folks force update tags? 🤔

toolmantim:
> How would folks force update tags? 🤔

There’s a few options perhaps, but what’s the use case for force updating a plugin tag?

lox:
> There’s a few options perhaps, but what’s the use case for force updating a plugin tag?

Mistakes. I've done it several times and I guarantee that folks will too, so worth discussing. Git tags aren't immutable, so building a system that acts like they are seems like it will go badly.

toolmantim:
Most package managers don’t let you push an updated version with the same tag, but instead get you to create a new release and delete the old one. Would that be acceptable you think?

toolmantim:
Given its a central plugin commit registry, we could provide ways for plugin authors to update/delete a known bad tag commit. But I was hoping to avoid that if a delete/pulling a release + re-release of a plugin would be acceptable.

toolmantim:
I forgot to mention in my initial comment above, if we go with this new approach we should discuss if we want to automatically verify SHAs on all sources of plugin for repos, both public on GitHub and private plugin repos.

lox:
Let me start off with saying that I think figuring out a mechanism to make plugin versions immutable and also to prevent malicious plugin tampering attacks is a really good goal. 

Perhaps the approach as stated is right, it would certainly work well for public pipelines. I think I get a bit twitchy about a world where things are "kinda" git, but there are proprietary tags that get used in some situations. I feel like it would be hard to know what was going on and when the guarantees would be in place. For instance, a public registry would be hard to make work with private plugin registries because they URLs aren't guaranteed to be unique. e.g `ssh://plugins.local/llamas-buildkite-plugin`.

lox:
I'm only about 3/10 on those objections though. If y'all think it's the right way to go, I'm keen to see where it goes.

sj26:
We had some good discussion about tags vs branches and mutability over here, if it helps:
https://github.com/buildkite/agent/issues/422

PR #1030, opened 2019-06-12
Escape pipe in windows test

lox:
We sometimes see `TestIsStdinIsReadableWithAPipe` failing under windows, this tries to fix the issue.


lox:
Nope, looks like I was wrong.

PR #1032, opened 2019-06-13
Add profile option to all cli commands

lox:
This makes it easy to get profiling traces for different agent commands. 

```bash
buildkite-agent start --profile memory
go tool pprof /var/folders/kk/2wmv_tq14m907m49pr66t1hh0000gn/T/profile511809002/mem.pprof
top5
```


lox:
Golang tends to work in one type of profile at a time @plasticine (that is my understanding anyway)

plasticine:
@lox That would totally make sense — was just a random thought 👍

PR #1034, opened 2019-06-16
Allow plugins to be restricted by an expression

lox:
Up until this point, we've recommended that folks concerned with security in the agent disable plugins because they could allow a variety of attacks, specifically:

* Users could reference plugins that execute arbitrary code which might by-pass the `--no-command-eval` option set on the agent
* Upstream plugin could be compromised and changed, which would result in a way to execute arbitrary code

We've recommended that people either disable them, or filter then with tools like https://github.com/buildkite/buildkite-allowed-plugins-hook-example.

This aims to use the [Conditional](http://github.com/buildkite/conditional) language to allow folks to restrict which agents an agent will allow. 

For example:

```
buildkite-agent start \
  --allow-plugin-if 'plugin.vendored == true || plugin.location == "buildkite/llamas")'
```

### DIFF
--- a/agent/agent_configuration.go
+++ b/agent/agent_configuration.go
@@ -19,6 +19,7 @@ type AgentConfiguration struct {
 	CommandEval                bool
 	PluginsEnabled             bool
 	PluginValidation           bool
+	PluginCondition            string
 	LocalHooksEnabled          bool
 	RunInPty                   bool
 	TimestampLines             bool
--- a/agent/agent_configuration.go
+++ b/agent/agent_configuration.go
@@ -19,7 +19,7 @@ type AgentConfiguration struct {
 	CommandEval                bool
 	PluginsEnabled             bool
 	PluginValidation           bool
-	PluginCondition            string
+	AllowPluginIf              string
 	LocalHooksEnabled          bool
 	RunInPty                   bool
 	TimestampLines             bool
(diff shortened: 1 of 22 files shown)


lox:
Still todo: 

* [ ] Conditionally allow plugins if `--plugin-condition` is set along side `--no-command-eval` (unless `--no-plugins` is also set)

toolmantim:
This is a cool use of conditionals. The plugin JSON env var blob is a total pain to deal with in hooks (https://github.com/buildkite/buildkite-allowed-plugins-hook-example). This makes it way easier to create a list of allowed plugins yeah? Wonder what that looks like in this syntax?

All other security logic is currently done via hooks (e.g. https://buildkite.com/docs/agent/v3/securing) and boolean configs. Any thoughts on making some of that stuff able to be done via conditional logic too? Stuff like repos, teams, plugins, etc? I.e. are there more configs we should add with conditional style logic? Or you thinking everything else would stay as-is?

I think there’s another issue I was hoping to address, that wasn’t related to locked down agents: helping people move to a world where digests are required for github.com repos by default. That was probably by I was shooting for just a boolean config for it.

petemounce:
> @lox feels on renaming to `--allow-plugin-if` ?

+1 - it reads more like authorization, then.

lox:
Closing this out as a very old stale PR.

If we still want this capability, it should be re-proposed cleanly from current `main` rather than dragging this branch forward.

PR #1036, opened 2019-06-19
Fix bug where file upload hangs and add a test

lox:
Turns out my fix in #1033 had a race condition in it where the main go routine would race to write form fields with the go routine that write the files. 

This fixes it, and adds a test like I should have done in the first place.

### DIFF
--- a/agent/form_uploader.go
+++ b/agent/form_uploader.go
@@ -92,7 +92,8 @@ func (u *FormUploader) Upload(artifact *api.Artifact) error {
 
 // Creates a new file upload http request with optional extra params
 func createUploadRequest(l logger.Logger, artifact *api.Artifact) (*http.Request, error) {
-	file, err := os.Open(artifact.AbsolutePath)
+	// Check the file exists
+	_, err := os.Stat(artifact.AbsolutePath)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


lox:
Darn this is failing on actual s3:

```

2019-06-19 14:08:00 DEBUG  POST https://s3.amazonaws.com/buildkiteartifacts.com
--
  | 2019-06-19 14:08:02 WARN   <?xml version="1.0" encoding="UTF-8"?>
  | <Error><Code>MissingContentLength</Code><Message>You must provide the Content-Length HTTP header.</Message><RequestId>xxx</RequestId><HostId>xxx</HostId></Error> (411) (Attempt 1/10 Retrying in 5s)


```

keithpitt:
@lox order somehow change?

lox:
That was my first thought, but it's much worse than that. Because it's now a stream there isn't a Content-Length, so we need to provide one. It's a multi-part stream though, so figuring out the Content-Length is dark magic.

lox:
Code has changed completely, mind looking again @keithpitt?

lox:
I think there might be a bug where the file handle isn't closed properly. Going to fix.

PR #1039, opened 2019-06-28
Add meta-data keys command

lox:
Implement a `buildkite-agent meta-data keys` command for listing meta-data keys:

E.g:

```bash
buildkite-agent meta-data keys # default delimited by newlines
buildkite-agent meta-data keys --delimiter ';'
```

You can use it to enumerate meta-data like:

```bash
#!/bin/bash
set -euo pipefail

buildkite-agent meta-data keys | while read -r key; do
  echo "$key: $(buildkite-agent meta-data get "$key")"
done
```

/cc @toolmantim

### DIFF
--- a/api/meta_data.go
+++ b/api/meta_data.go
@@ -64,3 +64,20 @@ func (c *Client) ExistsMetaData(jobId string, key string) (*MetaDataExists, *Res
 
 	return e, resp, err
 }
+
+func (c *Client) MetaDataKeys(jobId string) ([]string, *Response, error) {
+	u := fmt.Sprintf("jobs/%s/data/keys", jobId)
+
+	req, err := c.newRequest("POST", u, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	keys := []string{}
+	resp, err := c.doRequest(req, &keys)
+	if err != nil {
+		return nil, resp, err
+	}
+
+	return keys, resp, err
+}
--- a/clicommand/meta_data_keys.go
+++ b/clicommand/meta_data_keys.go
@@ -104,8 +104,13 @@ var MetaDataKeysCommand = cli.Command{
 			l.Fatal("Failed to find meta-data keys: %s", err)
 		}
 
-		for _, key := range keys {
-			fmt.Printf("%s%s", cfg.Delimiter, key)
+		last := len(keys) - 1
+
+		for idx, key := range keys {
+			fmt.Printf("%s", key)
+			if idx != last {
+				fmt.Printf("%s", cfg.Delimiter)
+			}
 		}
 	},
 }
--- a/clicommand/meta_data_keys.go
+++ b/clicommand/meta_data_keys.go
@@ -1,0 +1,111 @@
+package clicommand
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/buildkite/agent/api"
+	"github.com/buildkite/agent/cliconfig"
+	"github.com/buildkite/agent/retry"
+	"github.com/urfave/cli"
+)
+
+var MetaDataKeysHelpDescription = `Usage:
+
+   buildkite-agent meta-data keys [arguments...]
+
+Description:
+
+   Lists all meta-data keys that have been previously set, delimited by a newline.
+
+Example:
+
+   $ buildkite-agent meta-data keys`
+
+type MetaDataKeysConfig struct {
+	Job       string `cli:"job" validate:"required"`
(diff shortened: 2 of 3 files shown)


toolmantim:
Amazing! Thanks @lox. I went to give it a test, but I don't think the `meta-data keys` subcommand exists yet yeah?

lox:
Ok @toolmantim this should be good to go.

toolmantim:
❤️👏🏼 So excited. Can’t wait to give this a test!

toolmantim:
Woo, almost works a treat.

I tested it with this:

```yml
steps:
  - label: "Set the keys"
    commands: |
      cd ~/Codez/buildkite/agent
      go run main.go meta-data set foo1 bar1
      go run main.go meta-data set foo2 bar2
      go run main.go meta-data set foo3 bar3
  - wait
  - label: "Get the keys"
    commands: |
      cd ~/Codez/buildkite/agent
      echo "+++ List the keys"
      go run main.go meta-data keys
      echo "+++ Read each key"
      go run main.go meta-data keys | while read -r key; do
        echo "$$key: $$(go run main.go meta-data get "$$key")"
      done
```

The only thing I found is that I think we'll need to output the delimiter after the last item too, as the above outputs:

```

```

lox:
@toolmantim I dunno, we don't output a trailing \n at the end of `meta-data get`, why would we for `meta-data keys`? My take is that a delimiter should occur between records, not between and after. 

I think the use-case of printing to stdout is probably pretty rare, generally it's going to be consumed by a tool, for which the trailing delimited might actually cause issues. Thoughts?

lox:
I'm going to merge this and we can discuss in master.

lox:
@sj26 suggested a nice thing would be to detect if it was a TTY and maybe output a trialing newline?

toolmantim:
Did you spot that this didn't actually output all the keys though?

```
go run main.go meta-data keys | while read -r key; do
  echo "$$key: $$(go run main.go meta-data get "$$key")"
done
```

lox:
That is unexpected 🤔

lox:
```bash
while IFS= read -r key; do
  echo "$$key: $$(go run main.go meta-data get "$$key")"
done <(go run main.go meta-data keys)
```

Should work!

toolmantim:
Given this is designed for bash, I figured you'd want it to work with `read`? (like in the PR description?)

i.e. the following only prints "foo1" and "foo2":

```shell
printf "foo1\nfoo2\nfoo3" | while read -r key; do
  echo "$key"
done
# foo1
# foo2
# foo3
```

This shows the same:

```shell
while IFS= read -r key; do
  echo "$key"
done < <(printf "foo1\nfoo2\nfoo3")
# foo1
# foo2
```

Echo works, because it echoes the trailing newline:

```shell
echo -e "foo1\nfoo2\nfoo3" | while read -r key; do
  echo "$key"
done
# foo1
# foo2
# foo3
```

Maybe `--delimiter` is the wrong name for the argument? 🤷🏼‍♂️ But that's why I brought it up…

toolmantim:
Or… you could remove `--delimiter` altogether and make sure to put a newline at the end too, so it works for both `read` and just reading it in a terminal? And if people have meta-data keys containing newlines… 🤷🏼‍♂️

lox:
I’m inclined to agree and add a trailing newline @toolmantim

lox:
I’ll PR it

sj26:
Yeah I think the `--delimiter` may have warped my impression, it makes sense to make it newline only, and make sure there's a trailing newline. 🙌

lox:
This is a tricky one 🤔 I didn't understand the nuance of `read` where it returns a false value if it encounters an EOF before a newline. 

The correct way to do this with read is:

```bash
while IFS= read -r key || [[ -n "$key" ]] ; do 
  echo "$key"
done < <(printf "foo1\nfoo2\nfoo3")
```

This is because even though read returns false, the read still happens and `$key` is populated.

toolmantim:
The extra newline still seems like a more human API to me? 🤷🏼‍♂️

lox:
That is what I did @toolmantim!

lox:
Was just trying to understand the bash-ness.

lox:
See #1040.

toolmantim:
Ah, I see! 👍🏼 We can always add a `format` json/newlines argument for machine reading?

lox:
Yeah, I reckon.

PR #1041, opened 2019-07-15
Allow signal agent sends to bootstrap on cancel to be customized

lox:
Currently the agent sends a `SIGTERM` to the bootstrap on user cancellation of a job. 

This adds `cancel-signal` which accepts signals in the form of `SIGTERM`, `SIGHUP`, etc.

The default will remain `SIGTERM` until the next major version and then it will be `SIGINT`. 

/cc @DazWorrall


DazWorrall:
Thanks @lox, glad to see this coming. Did we add configurable timeout duration already or is that still TBD?

lox:
@DazWorrall `cancel-grace-period` exists, but it's agent-wide.

PR #1049, opened 2019-07-21
Show plugin validation errors as a warning

lox:
So that we can gradually transition to plugin validation warnings, this shows plugin validation errors as warnings:

image

This can be disabled with `--no-plugin-validation` still.

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -526,10 +526,6 @@ func (b *Bootstrap) preparePlugins() error {
 }
 
 func (b *Bootstrap) validatePluginCheckout(checkout *pluginCheckout) error {
-	if !b.Config.PluginValidation {
-		return nil
-	}
-
 	if checkout.Definition == nil {
 		if b.Debug {
 			b.shell.Commentf("Parsing plugin definition for %s from %s", checkout.Plugin.Name(), checkout.CheckoutDir)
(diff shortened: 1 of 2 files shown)


lox:
Would love feedback on the errors and messages!

toolmantim:
🎉🎉🎉

How much control of the validation errors do we have? It’s a shame we can’t give a line number of the pipeline.yml in the error.
 
Typo’d config keys, missing required keys, and invalid types for a key would be my guess on the biggest three mistakes people will make as plugin users.

Here’s one idea, for that example in the screenshot:

```
There are 2 configuration errors for plugin `docker-compose`:
* `alpacas` key not allowed at / (must be one of `build` or `run`)
* `llamas` key not allowed at / (must be one of `build` or `run`)
```

lox:
@toolmantim we don't really have much control over the error messages at all. Because we are validating the YAML as JSON, we also don't have the original line numbers. 

The library in use is https://github.com/qri-io/jsonschema.

bootstrap/bootstrap.go
```diff
@@ -526,10 +526,6 @@ func (b *Bootstrap) preparePlugins() error {
 }
 
 func (b *Bootstrap) validatePluginCheckout(checkout *pluginCheckout) error {
-	if !b.Config.PluginValidation {
```
keithpitt: Moving this further down in the function seems like a change in behaviour since the code below it could return an `err` and break the bootstrap, specially this bit:

image

Was that intentional? If so, totally fine. Just wanted to double check!

lox: You are totally right @keithpitt, will fix.

toolmantim:
@lox no worries! What feedback did you want on the errors/messages exactly?

lox:
Probably the kind you gave, I've been trying to figure out how to make them nicer :(

toolmantim:
Okay cool! It's sad that https://github.com/qri-io/jsonschema doesn't give better error messages, or much control over them.

lox:
Closing this out as a very old stale PR.

If this idea still matters, it should come back as a fresh change against the current agent codebase rather than trying to revive this branch.

PR #1059, opened 2019-08-01
Fix git submodule command

mattvenz:


lox:
Thanks @mattvenz, sorry you ran into this!

This was fixed in https://github.com/buildkite/agent/pull/1029, which is released in v3.13.2.

PR #1064, opened 2019-08-06
Separate the fetch and checkout decisions

### LINKED ISSUE #1063: Setting BUILDKITE_REFSPEC when BUILDKITE_COMMIT=HEAD breaks "git checkout"
When setting `BUILDKITE_REFSPEC`, the bootstrap command will fetch the given and will checkout `BUILDKITE_COMMIT`. Manually creating a build and leaving the default `BUILDKITE_COMMIT` value of `HEAD`, will result in a `git checkout HEAD` execution. Since `git fetch` doesn't touch `HEAD`, the checkout command has no affect.

https://github.com/buildkite/agent/blob/be5ea8b1c9575a598bea450fcc7fa5e22c7e6318/bootstrap/bootstrap.go#L1105-L1115

This appears to be handled properly when `BUILDKITE_REFSPEC` is unset as seen in https://github.com/buildkite/agent/blob/be5ea8b1c9575a598bea450fcc7fa5e22c7e6318/bootstrap/bootstrap.go#L1136-L1146

Given the effects of `BUILDKITE_REFSPEC` would be within the user's control, it would seem that `FETCH_HEAD` should be used to provide a consistent experience. Alternatively, `BUILDKITE_COMMIT`'s value could be checked for `HEAD` as in `Fetch and checkout remote branch HEAD commit` case.

matthewd:
Regardless of how we decide to fetch it, the choice of what to checkout is always the same: `FETCH_HEAD` if `b.Commit` is `HEAD`, and `b.Commit` otherwise.

Fixes #1063 -- "explicit `b.RefSpec` supplied" is the only branch whose behaviour is changing here, and only when `b.Commit` is `HEAD`.

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -1110,10 +1110,6 @@ func (b *Bootstrap) defaultCheckoutPhase() error {
 			return err
 		}
 
-		if err := b.shell.Run("git", "checkout", "-f", b.Commit); err != nil {
-			return err
-		}
-
 		// GitHub has a special ref which lets us fetch a pull request head, whether
 		// or not there is a current head in this repository or another which
 		// references the commit. We presume a commit sha is provided. See:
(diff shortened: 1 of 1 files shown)


lox:
Why didn't tests run on this? 🤔

lox:
Oh! It's a @matthewd fork.

PR #1065, opened 2019-08-06
Correctly log last successful heartbeat time.

philwo:
The agent mistakenly logged the last successful ping time instead, when an error occurred.


lox:
🤦🏼‍♂️ Good catch @philwo

PR #1066, opened 2019-08-06
Let the agent serve a status page via HTTP.

philwo:
This feature can optionally be enabled via the `--health-check-addr addr:port` flag or the equivalent `BUILDKITE_HEALTH_CHECK_ADDR` environment variable.

When enabled, the agent starts an HTTP server on the given `addr:port` that can be used for pull-based health checks, like [Google Cloud's instance group autohealing](https://cloud.google.com/compute/docs/instance-groups/creating-groups-of-managed-instances#monitoring_groups). This allows the instance group to automatically detect when a VM isn't running a healthy Buildkite agent and recreate it.

It exports two endpoints:
- `/`: This is an overall status page for the agent itself.
- `/agent/<spawn_id>`: This is the status page of the agent worker with the numeric index `spawn_id`, starting from 1. If you use `--spawn=N`, there will be one status page for each running agent. This one actually takes the last heartbeat into account and if the heartbeat fails, this will return a failed HTTP response, signaling that the agent isn't healthy.

I'll still do some polish and testing of this, but am sending it already for review to get your feedback on this idea. :)

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -2,6 +2,7 @@ package clicommand
 
 import (
 	"fmt"
+	"net/http"
 	"os"
 	"os/signal"
 	"path/filepath"
@@ -77,6 +78,7 @@ type AgentStartConfig struct {
 	NoPluginValidation         bool     `cli:"no-plugin-validation"`
 	NoPTY                      bool     `cli:"no-pty"`
 	TimestampLines             bool     `cli:"timestamp-lines"`
+	HealthCheckAddr            string   `cli:"health-check-addr"`
 	MetricsDatadog             bool     `cli:"metrics-datadog"`
 	MetricsDatadogHost         string   `cli:"metrics-datadog-host"`
 	Spawn                      int      `cli:"spawn"`
@@ -296,6 +298,11 @@ var AgentStartCommand = cli.Command{
 			Name:   "timestamp-lines",
 			Usage:  "Prepend timestamps on each line of output.",
 			EnvVar: "BUILDKITE_TIMESTAMP_LINES",
+		},
+		cli.StringFlag{
+			Name:   "health-check-addr",
+			Usage:  "Start an HTTP server on this addr:port that returns whether the agent is healthy, disabled by default",
+			EnvVar: "BUILDKITE_AGENT_HEALTH_CHECK_ADDR",
 		},
 		cli.BoolFlag{
 			Name:   "no-pty",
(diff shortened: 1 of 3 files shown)


clicommand/agent_start.go
```diff
@@ -297,6 +299,11 @@ var AgentStartCommand = cli.Command{
 			Usage:  "Prepend timestamps on each line of output.",
 			EnvVar: "BUILDKITE_TIMESTAMP_LINES",
 		},
+		cli.StringFlag{
+			Name:   "health-check-addr",
+			Usage:  "Start an HTTP server on this addr:port that returns whether the agent is healthy, disabled by default",
+			EnvVar: "BUILDKITE_HEALTH_CHECK_ADDR",
```
lox: For modern params we use `BUILDKITE_AGENT_` as a prefix, it's long, but let's go with `BUILDKITE_AGENT_HEALTH_CHECK_ADDR`

lox:
This looks awesome! Very keen on the idea. 

My biggest hesitation is the use of the default http server. It took me a while to click that a handler was being added in the agent package. I know this is how some of the standard library stuff works, but it's always seemed like a bad use of global scope to me. 

What are your thoughts on that @philwo? I'd be happy to accept this implementation and we can refactor the next time we need to touch it, but figured it was worth discussing.

lox:
Oh, my only other very minor observation was there might be a small window of time where the `/` route replies with `OK` before the agent worker has actually spun up and registered.

philwo:
@lox Thanks for the review and good feedback! I'm on vacation this week, will follow up next week :)

philwo:
> Oh, my only other very minor observation was there might be a small window of time where the / route replies with OK before the agent worker has actually spun up and registered.

I think this case is handled OK'ish by the two endpoints that you can query:
- If you just care that the agent master process itself is running, you can `GET /`. This doesn't necessarily mean that the agent workers are correctly registered, but is still easy to use when using multiple agent workers per instance.
- If you want to check that the agent worker is correctly registered and doing regular heartbeats, you can `GET /agent/1` (or `/2`, ...). This is a more reliable check, but problematic if you use BUILDKITE_AGENTS_PER_INSTANCE to launch multiple workers, because I think in that case it's legitimate if one worker already shuts down (due to idleness), so handling the health check becomes more complex (e.g. do you consider the VM still healthy if just one agent worker is still online).

I'll go and try this out on Bazel's CI and will report back how it works for us. :)

philwo:
I've been running our CI agents with this patch and it's been working fine.

The VMs' health is now verified correctly by the GCE instance group managers and VMs that don't correctly start their agent due to some flaky boot issue are recreated quickly. 😊

Let me know if you want any changes to the code, otherwise I think it's OK for now.

clicommand/agent_start.go
lox: Turns out, I'm totally wrong :D

PR #1067, opened 2019-08-06
Allow fetching arbitrary tag=suffix pairs from GCP/EC2 meta-data

plasticine:
This patch enhances the integration of the agent with EC2 and GCP meta-data endpoints by allowing for arbitrary fetching of tag name & meta-data path suffix pairs.

This new functionality is intended to be additive only and can be used via two new flags, on for AWS and one for GCP; `--tags-from-ec2-meta-data-paths`, `--tags-from-gcp-meta-data-paths`.

#### Example

```
--tags-from-ec2-meta-data-paths "omg=spot/termination-time"
```

This patch also deprecates the `--tags-from-ec2` & `--tags-from-gcp` flags in favour of `--tags-from-ec2-meta-data` && `--tags-from-gcp-meta-data` respectively for consistency. The deprecated flags are not removed yet, but would be in v4.

### DIFF
--- a/agent/ec2_meta_data_test.go
+++ b/agent/ec2_meta_data_test.go
@@ -1,28 +1,8 @@
 package agent
 
 import (
-	"fmt"
-	// "github.com/stretchr/testify/assert"
-	"net/http"
-	"net/http/httptest"
-	"net/url"
-	// "os"
 	"testing"
 )
 
 func TestEC2MetaDataGetSuffixes(t *testing.T) {
-	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		w.WriteHeader(http.StatusOK)
-
-		switch path := r.URL.EscapedPath(); path {
-		default:
-			t.Fatalf("Error %q", path)
-		}
-	}))
-	defer ts.Close()
-
-	url, err := url.Parse(ts.URL)
-	if err != nil {
-		t.Fatalf("Error %q", err)
-	}
 }
--- a/agent/ec2_meta_data_test.go
+++ b/agent/ec2_meta_data_test.go
@@ -1,0 +1,28 @@
+package agent
+
+import (
+	"fmt"
+	// "github.com/stretchr/testify/assert"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	// "os"
+	"testing"
+)
+
+func TestEC2MetaDataGetSuffixes(t *testing.T) {
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+
+		switch path := r.URL.EscapedPath(); path {
+		default:
+			t.Fatalf("Error %q", path)
+		}
+	}))
+	defer ts.Close()
+
+	url, err := url.Parse(ts.URL)
+	if err != nil {
+		t.Fatalf("Error %q", err)
+	}
+}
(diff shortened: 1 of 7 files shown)


agent/ec2_meta_data_test.go
```diff
@@ -0,0 +1,28 @@
+package agent
```
plasticine: Open to ideas for how to test this — ping @lox

agent/tags.go
```diff
@@ -143,3 +205,38 @@ func FetchTags(l *logger.Logger, conf FetchTagsConfig) []string {
 
 	return tags
 }
+
+// TODO: Seems maybe el crapola, dunno will ask lox
```
plasticine: Well, is it? 😅

agent/ec2_meta_data_test.go
lox: I'll 🧠 today.

plasticine: @lox I’m sure it’s pretty easy with a little test http server but I ran out of gojuice to work it out

agent/tags.go
```diff
@@ -95,9 +108,28 @@ func (t *tagFetcher) Fetch(l logger.Logger, conf FetchTagsConfig) []string {
 		}
 	}
 
+	// Attempt to add the EC2 meta-data fetched from user-default path suffixes
+	if len(conf.TagsFromEC2MetaDataPaths) > 0 {
+		paths, err := parseTagValuePathPairs(conf.TagsFromEC2MetaDataPaths)
+		if err != nil {
+			l.Error(fmt.Sprintf("Error parsing meta-data tag and path pairs: %s", err.Error()))
```
lox: This will show an error but continue, is that the intention?

agent/tags.go
```diff
@@ -123,9 +155,27 @@ func (t *tagFetcher) Fetch(l logger.Logger, conf FetchTagsConfig) []string {
...
+		}
+	}
+
 	// Attempt to add the Google Cloud meta-data
-	if conf.TagsFromGCP {
-		gcpTags, err := t.gcpMetadata()
+	if len(conf.TagsFromGCPMetaDataPaths) > 0 {
+		paths, err := parseTagValuePathPairs(conf.TagsFromGCPMetaDataPaths)
+		if err != nil {
+			l.Error(fmt.Sprintf("Error parsing meta-data tag and path pairs: %s", err.Error()))
```
lox: Same as the other one.

agent/tags.go
lox: You could probably just use `strings.SplitN` to force it to split into two chunks and then check the len is 2.

agent/tags.go
```diff
@@ -164,3 +214,29 @@ func (t *tagFetcher) Fetch(l logger.Logger, conf FetchTagsConfig) []string {
...
+	for _, pair := range paths {
+		// Sanity check the format of each pair to ensure that it’s parseable
+		index := strings.LastIndex(pair, "=")
+		if index == -1 || index == 0 || index == len(pair)-1 {
+			return result, errors.New(fmt.Sprintf("`%s` cannot be parsed, format should be `tag=metadata/path`", pair))
+		}
+
+		x := strings.Split(pair, "=")
+		// TODO: Should we just let people have stupid keys? Probably?
+		key := strings.ToLower(strings.Trim(x[0], " "))
```
lox: strings.TrimSpace ftw

lox:
This is looking good @plasticine. Mind just grepping for the config keys that you've deprecated and make sure the configs we ship for packages are updated?

agent/tags.go
plasticine: Good catch — nope, this should ‼️💥

plasticine: Actually — now I think about it a little more I’m actually not sure about this. I think it’s fine to throw an error but continue and it’s how it already works...

PR #1071, opened 2019-08-12
Reword the cancel-grace-timeout config option

toolmantim:


lox:
Technically this is around the time the bootstrap is given to finish executing, but that is probably not a meaningful distinction. The agent also has no concept of "timed out", it's just canceled it, but also probably a technical vs useful distinction.

toolmantim:
Without mentioning "timed out", wouldn't someone looking for config options to tweak for timed out jobs overlook this config option, thinking it only affects jobs that are cancelled? Nowhere does the job show as "canceled" in the web UI.

I guess they'd have to know how the `Agent API > Agent > bootstrap` works, to understand what config option to change in the case of trying to improve graceful exits of timed out jobs?

lox:
Yup, I agree, was just working through the layers of abstraction in my head.

toolmantim:
Cool cool!

This came about trying to help someone give their agent more time to upload artifacts after setting the `timeout_in_minutes` to 1h in their pipeline.yml. And all our docs, REST API and GraphQL API have two distinct job states: `canceled` and `timed_out`. But maybe this change isn't quite right.

Having a look [at the docs for timeout_in_minutes](https://buildkite.com/docs/pipelines/command-step#command-step-attributes) though, it does say "The number of minutes a job created from this step is allowed to run. If the job does not finish within this limit, it will be automatically canceled and the build will fail." — though maybe we just need to update that description.

In this particular case, it showed the job as this:

image

With this at the bottom of the job when the bootstrap failed to exit in the timeframe:

image

toolmantim:
(agent version v3.13.2)

toolmantim:
Should we just merge this for now?

PR #1075, opened 2019-08-15
Allow buildkite agent to finish jobs before being killed

james-portman-contino:
Set TimeoutStopSec to "infinity" per the systemd documentation
I am assuming when someone set this to 0, they were thinking that 0 meant disable/infinity so that the agent would have chance to finish running any current builds

See:
https://www.freedesktop.org/software/systemd/man/systemd.service.html#TimeoutStopSec=

### DIFF
--- a/packaging/linux/root/usr/share/buildkite-agent/systemd/buildkite-agent.service
+++ b/packaging/linux/root/usr/share/buildkite-agent/systemd/buildkite-agent.service
@@ -13,7 +13,7 @@ RestartSec=5
 Restart=on-failure
 RestartForceExitStatus=SIGPIPE
 TimeoutStartSec=10
-TimeoutStopSec=0
+TimeoutStopSec=infinity
 KillMode=process
 
 [Install]


lox:
Thanks @james-portman-contino, `systemd` is super tricky in that lots of things aren't supported in the older versions on Amazon Linux. 

Are you sure that  `TimeoutStopSec=0` doesn't disable the timeout logic? I'd bet that `TimeoutStopSec=infinity` doesn't work on older systemd versions.

lox:
I'll try making this change in Elastic Stack and see how it goes.

lox:
I definitely agree it's the correct change, just nervous about backwards compat.

lox:
Could I trouble you to also make this change to https://github.com/buildkite/agent/blob/master/packaging/linux/root/usr/share/buildkite-agent/systemd/buildkite-agent%40.service @james-portman-contino?

james-portman-contino:
Sure - just created #1080

jeremiahsnapp:
I run buildkite-agent on a number of systemd platforms so I figured I'd test `TimeoutStopSec=0` and `TimeoutStopSec=infinity` on each of them and share my results here.

I tested by using `TimeoutStopSec=0` and `TimeoutStopSec=infinity` in each platform's cron service file and then ran `systemctl daemon-reload` and `systemctl show crond.service` to determine if it changed the parameter correctly.

All of the following platforms were compatible with both `TimeoutStopSec=0` and `TimeoutStopSec=infinity`.

Amazon Linux 2
Debian 9
RHEL 7
RHEL 8
SLES 12
SLES 15
Ubuntu 16.04
Ubuntu 18.04

Debian 8 was the only platform that couldn't handle `TimeoutStopSec=infinity`. It could only handle `TimeoutStopSec=0`.

For what it's worth, I'm going to continue using `TimeoutStopSec=0` since we're supporting Debian 8 until it reaches EOL next Summer.

james-portman-contino:
> I run buildkite-agent on a number of systemd platforms so I figured I'd test `TimeoutStopSec=0` and `TimeoutStopSec=infinity` on each of them and share my results here.
> 
> I tested by using `TimeoutStopSec=0` and `TimeoutStopSec=infinity` in each platform's cron service file and then ran `systemctl daemon-reload` and `systemctl show crond.service` to determine if it changed the parameter correctly.
> 
> All of the following platforms were compatible with both `TimeoutStopSec=0` and `TimeoutStopSec=infinity`.
> 
> Amazon Linux 2
> Debian 9
> RHEL 7
> RHEL 8
> SLES 12
> SLES 15
> Ubuntu 16.04
> Ubuntu 18.04
> 
> Debian 8 was the only platform that couldn't handle `TimeoutStopSec=infinity`. It could only handle `TimeoutStopSec=0`.
> 
> For what it's worth, I'm going to continue using `TimeoutStopSec=0` since we're supporting Debian 8 until it reaches EOL next Summer.

Just to check with you, do you realise that using `TimeoutStopSec=0` for systems that support `infinity` might mean they you are terminating builds abruptly - not waiting for them finish?
It seems to be interpreted as 0 seconds (don't wait).
It wasn't just a matter of changing something for the sake of it, it's actually wrong for most OSs now

jeremiahsnapp:
>Just to check with you, do you realise that using TimeoutStopSec=0 for systems that support infinity might mean they you are terminating builds abruptly - not waiting for them finish?
It seems to be interpreted as 0 seconds (don't wait).
It wasn't just a matter of changing something for the sake of it, it's actually wrong for most OSs now

Prior to systemd v229 you would disable the timeout logic by setting `TimeoutStopSec=0` as you can see in this section from `man systemd.service` from an Amazon Linux 2 system which is using systemd v219.

```
       TimeoutStopSec=
           Configures the time to wait for stop. If a service is asked to stop, but does not terminate in the specified time, it will be terminated forcibly via SIGTERM, and after another timeout of equal duration with SIGKILL (see KillMode= in systemd.kill(5)).
           Takes a unit-less value in seconds, or a time span value such as "5min 20s". Pass "0" to disable the timeout logic. Defaults to DefaultTimeoutStopSec= from the manager configuration file (see systemd-system.conf(5)).
```

Ever since systemd v229 you can set `TimeoutStopSec=infinity` to disable the timeout logic but `TimeoutStopSec=0` is still acceptable to maintain backwards compatibility as you can see in the [section of v219 changelog](https://github.com/systemd/systemd/blob/master/NEWS#L4111-L4115) below.

```
        * Most configurable timeouts in systemd now expect an argument of
          "infinity" to turn them off, instead of "0" as before. The semantics
          from now on is that a timeout of "0" means "now", and "infinity"
          means "never". To maintain backwards compatibility, "0" continues to
          turn off previously existing timeout settings.
```

So setting `TimeoutStopSec=infinity` actually doesn't provide anything different from `TimeoutStopSec=0` on systems running systemd v229 or better.

However, setting `TimeoutStopSec=infinity` actually breaks things on systems that still use versions of systemd older than v229, such as Amazon Linux 2 which uses systemd v219, as you can see in this issue I just opened.

https://github.com/buildkite/elastic-ci-stack-for-aws/issues/629

Since systemd v229 and later explicitly allows `0` to continue to be used to disable timeout logic for timeout settings that existed prior to systemd v229 (such as `TimeoutStopSec`) for backwards compatibility I think the safest thing to do in this case is to continue using `0`.

lox:
Interesting, that is what I recall being the case when I chose the `0`, but it sounds like you are observing something different @james-portman-contino?

Thanks for the detailed research @jeremiahsnapp, I missed the changelog detail when I went looking.

lox:
> It wasn't just a matter of changing something for the sake of it, it's actually wrong for most OSs now

@james-portman-contino what systemd version were you testing against? I'll do some testing too.

PR #1081, opened 2019-08-20
Allow for exponential backoff between retries for artifact uploads

srmocher:
We occasionally get rate-limited from GCS (where we upload artifacts to) due to high load on the bucket. The suggested retry policy in such cases is to use exponential backoff between retries. https://cloud.google.com/storage/docs/exponential-backoff. We add a cli flag to optionally specify backoff strategy.

Since the agent doesn't do this currently, we're having to workaround by using `gsutil` which already does this. However, we'd like to do this through the agent so that we can see the artifacts in UI and also allow sharing of artifacts between jobs (which we've lost with using `gsutil`). 

Tested locally as follows with dummy values:
```
C02Y231UJHD2:agent sridhar.mocherla$ buildkite-agent artifact upload --job testjob --backoff-strategy exponential --endpoint localhost test.txt
2019-08-19 23:07:47 INFO   Found 1 files that match "test.txt"
2019-08-19 23:07:47 INFO   Creating (0-1)/1 artifacts
2019-08-19 23:07:47 WARN   Post localhost/jobs/testjob/artifacts: unsupported protocol scheme "" (Attempt 1/10 Retrying in 2s)
2019-08-19 23:07:49 WARN   Post localhost/jobs/testjob/artifacts: unsupported protocol scheme "" (Attempt 2/10 Retrying in 4s)
2019-08-19 23:07:53 WARN   Post localhost/jobs/testjob/artifacts: unsupported protocol scheme "" (Attempt 3/10 Retrying in 8s)
2019-08-19 23:08:01 WARN   Post localhost/jobs/testjob/artifacts: unsupported protocol scheme "" (Attempt 4/10 Retrying in 16s)
```

### DIFF
--- a/agent/artifact_batch_creator.go
+++ b/agent/artifact_batch_creator.go
@@ -17,6 +17,9 @@ type ArtifactBatchCreatorConfig struct {
 
 	// Where the artifacts are being uploaded to on the command line
 	UploadDestination string
+
+	// backoff strategy between retries
+	BackoffStrategy string
 }
 
 type ArtifactBatchCreator struct {
@@ -76,7 +79,7 @@ func (a *ArtifactBatchCreator) Create() ([]*api.Artifact, error) {
 			}
 
 			return err
-		}, &retry.Config{Maximum: 10, Interval: 5 * time.Second})
+		}, &retry.Config{Maximum: 10, Interval: 5 * time.Second, BackoffStrategy: a.conf.BackoffStrategy})
 
 		// Did the batch creation eventually fail?
 		if err != nil {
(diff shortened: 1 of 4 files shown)


lox:
Thanks for this! We'll review and discuss.

lox:
We'd prefer to not have a different config for this, how about we just look for specific exit codes and use a backoff strategy in those cases?

srmocher:
@lox Sure, I'll update the PR accordingly.

moskyb:
Closing this as it's largely (though not entirely) superseded by https://github.com/buildkite/agent/pull/1588.

PR #1086, opened 2019-09-11
allow gs_downloader to use GS_APPLICATION_CREDENTIALS

GaryPWhite:
Behind our firewall, the default google client will explode when performing some operations. This PR is to address that issue by allow us to authenticate during the download process with the client. By not using the default client, we're able to download protected artifacts and bypass the firewall restriction.

Let me know if we have any problems/questions, and thanks!

### DIFF
--- a/agent/gs_downloader.go
+++ b/agent/gs_downloader.go
@@ -1,13 +1,11 @@
 package agent
 
 import (
-	"context"
 	"errors"
 	"fmt"
 	"strings"
 
 	"github.com/buildkite/agent/logger"
-	"golang.org/x/oauth2/google"
 	storage "google.golang.org/api/storage/v1"
 )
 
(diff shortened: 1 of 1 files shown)


lox:
Is `newGoogleClient` defined somewhere else?

GaryPWhite:
Bumping in case y'all have more questions or if I need to contribute elsewhere to make this work 😁

lox:
For future generations, the function this change references is at https://github.com/buildkite/agent/blob/6be67b354dcb95c7600caa09e69077f7ab3e6390/agent/gs_uploader.go#L74

I reckon in future we should avoid `new*` on constructors that have extra behaviour like reading env.

lox:
Of course I wrote that code. 🤦🏼‍♂️

lox:
Sorry that took me so long @GaryPWhite. Gotta get us some more Golang developers.

GaryPWhite:
Hiring? ;)

harrietgrace:
Docs updated: https://github.com/buildkite/docs/pull/556

PR #1093, opened 2019-09-26
Allow for raw injection of JSON through BUILDKITE_GS_APPLICATION_CRED…

### LINKED ISSUE #1085: GS_APPLICATION_CREDENTIALS as an environment variable, not a file
In some cases we've seen now, there is a call for us to not store `BUILDKITE_GS_APPLICATION_CREDENTIALS` as a file on disk. It seems like a relatively easy task to accept either a file or a `.json` string. I would like feedback on that thought / how we ought to go about that. Should we have a `BUILDKITE_GS_APPLICATION_CREDENTIALS_(RAW|JSON)` that accepts an env var, and supersedes the existing file? What do y'all think?

GaryPWhite:
Resolves #1085 

will allow for injection of BUILDKITE_GS_APPLICATION_CREDENTIALS as raw JSON, instead of through a file. We've seen some appetite for this in our organization, and would love to hear thoughts.

### DIFF
--- a/agent/gs_uploader.go
+++ b/agent/gs_uploader.go
@@ -71,17 +71,24 @@ func ParseGSDestination(destination string) (name string, path string) {
 	return
 }
 
+func clientFromJSON(data []byte, scope string) (*http.Client, error) {
+	conf, err := google.JWTConfigFromJSON(data, scope)
+	if err != nil {
+		return nil, err
+	}
+	return conf.Client(oauth2.NoContext), nil
+}
+
 func newGoogleClient(scope string) (*http.Client, error) {
-	if os.Getenv("BUILDKITE_GS_APPLICATION_CREDENTIALS") != "" {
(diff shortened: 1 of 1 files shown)


lox:
My main objection here is that it encourages credentials in environment vs in a file 🤔

I'd love to hear a bit more context about why it's easier in env!

GaryPWhite:
easier in env so we don't have to hold the credentials "in plain text" on the agents.

For example, if I wanted to encrypt them in a file then only read the file at runtime there's not an easy way to do that, where if I wanted to scope only `buildkite-agent` to be able to run `decrypt` as a script/executable to lock down permissions -- I could have something like 
```
BUILDKITE_GS_APPLICATION_CREDENTIALS_JSON=$(decrypt /path/to/encrypted/gcs.json)
```
...if that makes sense :)

lox:
That makes sense. Ok, I'm cool!

PR #1100, opened 2019-10-07
Add a fetch-only option for the git mirror experiment

eberle1080:
The git mirror experiment currently calls `git remote update` as the primary update command. When the repository has a large number of branches (n > 10,000) this operation significantly slows down the server (especially with many parallel builds).

This PR adds an option which will optionally (must opt-in) call `git fetch origin refspec` instead of `git remote update`. This is essentially the same call except that `git fetch` only operates on one branch. This makes much less work for the server in the case of many branches.

And in case anyone asks, yes this is actually a concern for my team on the project we work with. We're running in to these limitations daily.

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -1045,8 +1045,15 @@ func (b *Bootstrap) updateGitMirror() (string, error) {
 	}
 
 	// Update our mirror
-	if err := b.shell.Run("git", "--git-dir", mirrorDir, "remote", "update", "--prune"); err != nil {
-		return "", err
+	if b.GitMirrorsFetchOnly {
+		gitFetchFlags := fmt.Sprintf("--git-dir '%s' %s", mirrorDir, b.GitFetchFlags)
+		if err := gitFetch(b.shell, gitFetchFlags, "origin", b.Branch); err != nil {
+			return "", err
+		}
+	} else {
+		if err := b.shell.Run("git", "--git-dir", mirrorDir, "remote", "update", "--prune"); err != nil {
+			return "", err
+		}
 	}
 
 	return mirrorDir, nil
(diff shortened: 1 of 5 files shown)


lox:
Thanks very much for the PR! 

I wonder if we could make this behaviour the default? Looking at the docs for `git remote update`, it seems like `git fetch --prune origin refspec` would be totally adequate.

pda:
I like the idea of making it default and having fewer config options.
Do we anticipate downsides to always using `git fetch`?

adamhowardprice:
Any updates on this?

adamhowardprice:
We would like to use this option.

bootstrap/bootstrap.go
```diff
@@ -1045,8 +1045,15 @@ func (b *Bootstrap) updateGitMirror() (string, error) {
 	}
 
 	// Update our mirror
-	if err := b.shell.Run("git", "--git-dir", mirrorDir, "remote", "update", "--prune"); err != nil {
-		return "", err
+	if b.GitMirrorsFetchOnly {
+		gitFetchFlags := fmt.Sprintf("--git-dir '%s' %s", mirrorDir, b.GitFetchFlags)
+		if err := gitFetch(b.shell, gitFetchFlags, "origin", b.Branch); err != nil {
```
pda, line 1050: I wonder if this has been tested (i.e. executed at all).
AFAIK `--git-dir` must come before the git subcommand like `fetch`, otherwise it's silently ignored.
Testing with a random git repo I have locally:

```
pda@paulbookpro ~/tmp ❯ git fetch --git-dir ~/code/example.git
fatal: not a git repository (or any of the parent directories): .git
```

```
pda@paulbookpro ~/tmp ❯ git --git-dir ~/code/example.git fetch
remote: Enumerating objects: 81, done.
remote: Counting objects: 100% (81/81), done.
remote: Compressing objects: 100% (32/32), done.
remote: Total 167 (delta 49), reused 62 (delta 40), pack-reused 86
Receiving objects: 100% (167/167), 50.87 KiB | 5.65 MiB/s, done.
Resolving deltas: 100% (81/81), completed with 16 local objects.
…
```

(I'm looking into merging #1112 which is basically a version of this PR without conditionals around the new behaviour)

PR #1101, opened 2019-10-08
Add --include-retried-jobs to artifact download/shasum

toolmantim:
By default, jobs that are retried are excluded from `buildkite-agent artifact download` searches. This adds support for searching retried jobs when downloading/shasumming with a new `--include-retried-jobs` CLI argument.


toolmantim:
Thanks @lox!! ❤️

PR #1103, opened 2019-10-10
Add ansi timestamp output experiment

lox:
Currently, we calculate timing information in the Buildkite UI by sending a stream of timestamps from when headers are output. This can lead to some weird bugs with time skew and at best shows time elapsed between headers.

This change uses invisible ANSI control codes at the start of every line to send more granular timing information to Buildkite, which allows for magically toggling in the UI:

This also disables sending header timing information. 

The aim is for this to be the default in 4.0.

### DIFF
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -140,23 +140,14 @@ func NewJobRunner(l logger.Logger, scope *metrics.Scope, ag *api.AgentRegisterRe
 
 	if experiments.IsEnabled(`ansi-timestamps`) {
 		// If we have ansi-timestamps, we can skip line timestamps AND header times
-
-		go func() {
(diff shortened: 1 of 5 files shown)


agent/job_runner.go
```diff
@@ -138,8 +138,25 @@ func NewJobRunner(l logger.Logger, scope *metrics.Scope, ag *api.AgentRegisterRe
 
 	pr, pw := io.Pipe()
 
-	// If we have timestamp lines on, we have to buffer lines before we flush them
-	if conf.AgentConfiguration.TimestampLines {
+	if experiments.IsEnabled(`ansi-timestamps`) {
```
keithpitt: Caaaaaan it be at the start of the line?

Issue #982, opened 2019-04-08
Filter secrets from build logs


azz:
We add some secrets to our builds using the `environment` hook.

Currently, a pipeline containing

```yaml
steps:
  - command: env
```

will list out all of the secrets in the build log. Not great!

We're going to look at maybe using the `command` hook to replace the secrets with `******` to prevent leaks, but thought it was worth raising an issue here to see if it's a more general issue that could be solved?

GreyKn:
Why run `env` in the first place as a command?

If you need to read values from it for some reason, you could just pipe into grep with -v to skip secrets

azz:
It's not that `env` is used as a command, it can be used to exfiltrate secrets from the environment, either intentionally or accidentally. 

One use case would be turning on `set -x` in bash, a command that looked like this:

```
set -x
curl "https://${FOO_USER}:${FOO_PASSWORD}/"
```

would leak the secrets to the build log.

lox:
Filtering secrets from logs is unfortunately very difficult to do comprehensively, especially if your aim is to prevent exfiltration. Attackers can generally work around any mechanism you have in place, and in some cases it can make it easier to enumerate secrets for them. 

Filtering out accidentally disclosed secrets could be something we could look at, either by filtering out values that are exported from an `environment` hook, or perhaps detecting high-entropy strings. If we get a chance we'll prototype something up, otherwise we'd love a PR!

azz:
What about a simple `BUILDKITE_SECRETS` variable that gets sets during the `environment` hook to a list of environment variables whose values should be considered "secrets".

e.g., `environment`:

```sh
GITHUB_ACCESS_TOKEN=$(get_secret 'github-access-token')
export GITHUB_ACCESS_TOKEN

export BUILDKITE_SECRETS='GITHUB_ACCESS_TOKEN'
```

This way the agent can filter out any secret values from `stdout`/`stderr`. 

> in some cases it can make it easier to enumerate secrets for them.

This approach _would_ do that, but they're already in the environment, so if an agent is compromised you can just serialise the whole `env` and post it off to some URL anyway.

azz:
Fixed in #1109

PR #1109, opened 2019-10-16
Add filter to remove passwords, etc from job output

dbaggerman:
I've been taking a look at #982, and wanted to put up what I've got so far for initial feedback. I haven't had a chance to actually build an agent with the change to confirm it does what I expect.

I think `Bootstrap` is the right place to put this. I was initially looking at putting it under `JobRunner`, but that won't have access to secrets that get pulled by `environment`/`pre-command` hooks which seems like a common use-case. I'm thinking it should probably be added to `executeHook` as well.

This implementation has ended up being more complex than I'd expected. I considered other implementations but found the trade off seems to be that while this is complex, it's also well encapsulated in that it's just an `io.Writer` with an extra `Sync` method.

Using something like `bufio.Scanner`/`strings.Replace` requires an `io.Reader`, so it ends up requiring `io.Pipe` and running asynchronously in a goroutine. That all ends up making it a bit messier to fit into the current code.

I also tried building an `io.Writer` that just uses `bytes.Replace`, but that won't handle tokens that happen to fall across `Write()` boundaries.

I can go back and take another look at alternative methods, but I wanted to get some feedback since I'm being too indecisive.


lox:
This is super interesting! Thank you for the PR! 

In terms of implementation of a bootstrap output filter, this looks like it's heading in a great direction. I'll dive into the code with some more feedback.

My main concerns around redaction have always been how we get the secrets to redact into that filter. Looks like this implementation has env keys defined in `BUILDKITE_SENSITIVE_VARS` and those then get redacted? Interested what the trade-offs there are, I'd be worried that folks would forget to opt-in. 

Another approach we considered a way back was to add a `secrets` hook after the `environment` hook that would automatically redact anything exported. The reason we didn't was that we thought that we thought that encouraging folks to export secrets into plain text environments long-term wasn't a good thing to do. Secrets are best when they are short-lived, ideally individual command invocations would call something like `SHORT_LIVED_ENV=$(buildkite-agent secrets get xxxxx) command to exec`. 

It's tricky to balance though, can definately see how this approach makes for a pragmatic defense-in-depth approach with the current approach folks use to secrets management.

dbaggerman:
> My main concerns around redaction have always been how we get the secrets to redact into that filter. Looks like this implementation has env keys defined in `BUILDKITE_SENSITIVE_VARS` and those then get redacted? Interested what the trade-offs there are, I'd be worried that folks would forget to opt-in.

Yeah, that's the idea. if `BUILDKITE_SENSITIVE_VARS=FOO_PASSWORD,BAR_API_TOKEN`, then it will look for whatever those two environment variables are set to before running the command, and strip those from the output. I thought allowing wildcards like `*PASSWORD*,*API_TOKEN*` might be an option to simplify opt-in.

I agree that it's better when secrets are shorter-lived and this approach won't cover that if the value isn't visible to `buildkite-agent`. But, as you say, it does cover many common use-cases. We have around 10 secrets that we populate via a global hook, and sticking `SECRET_A=$(get-secret ...) SECRET_B=$(get-secret ...) SECRET_C=$(get-secret ...)`, etc on commands would get a bit cumbersome. Not that every job needs every secret and it would be nice to be more fine-grained, but it's a hassle to manage that many secrets individually on hundreds of builds.

lox:
Btw, this is a lovely implementation! Some very nice go code. Using Boyer-Moore skip distances is a nice way to not buffer too much.

lox:
I'd be tempted to specifically call it a `Redactor` vs `LogFilter`. That way it's clear that it's not general purpose.

lox:
Then on that theme, I'd say `BUILDKITE_REDACTED_VARS`.

dbaggerman:
> Btw, this is a lovely implementation! Some very nice go code. Using Boyer-Moore skip distances is a nice way to not buffer too much.

Thanks! Incidentally, this is actually the fastest implementation I played around with. `strings.Replace`/`bytes.Replace` require scanning the input once per string being redacted, and also allocate a lot of garbage since they return copies of the input.

> I'd be tempted to specifically call it a Redactor vs LogFilter. That way it's clear that it's not general purpose.

Funny you say that, since that's what I had it called originally. I can't remember why I changed it, but I'm happy to change it back and change the variable name to match.

lox:
I've posted a thread internally, we'll discuss and provide some more feedback.

keithpitt:
(I'm totally on board with this btw)

keithpitt:
@lox I wonder if we should surround the `[REDACTED]` text with an ANSI code so we can do some nice highlighting in the UI?

image

Imagine if you hovered over it and it gave you a popup like "This was removed, click here for more info" or something?

(Maybe I"m overcomplicating it, feel free to ignore me...)

lox:
Yeah, I reckon we start simple and get this in and then look at improving things like tooltips @keithpitt?

bootstrap/bootstrap.go
```diff
@@ -1437,6 +1437,23 @@ func (b *Bootstrap) defaultCommandPhase() error {
 		return runDeprecatedDockerIntegration(b.shell, []string{cmdToExec})
 	}
 
+	if sensitiveVarNames, ok := b.shell.Env.Get("BUILDKITE_REDACTED_VARS"); ok {
```
lox: Because this is looking at the current shell environment, it means it can be overridden at a build or pipeline level. I'd say that this is probably not desirable.

This probably needs to be an AgentConfiguration that can only be set in the agent configuration or in a hook. I can help with this bit if you need @dbaggerman.

keithpitt:
@lox yeah that's cool, ignore me!

lox:
@keithpitt suggested a default of `BUILDKITE_REDACTED_VARS=*_SECRET;*_TOKEN;*_PASSWORD`.

lox:
>  I'm thinking it should probably be added to executeHook as well.

Yup, reckon we need this. The only other way I can think of to leak strings would be in error messages 🤔

bootstrap/bootstrap.go
dbaggerman: If you can point me in the right direction for this, that would be great.

dbaggerman:
I've moved `BUILDKITE_REDACTED_VARS` to AgentConfiguration and added the glob matching. Matching is courtesy of `path.Match` so it technically has a couple of options other than just `*`, but it's standard UNIX shell globbing so I think it's fine that way.

I've done some testing with `buildkite-agent bootstrap` outside of a running agent and it seems to be working as expected.

I noticed that the buffering was user-visible when content is being written line by line, so I added some code to not hold line endings in the buffer between `Write`s. So long as nobody needs to redact passwords that have newlines in them, it should still handle the `Write` boundaries properly.

That does still leave an edge-case of progress output without newlines though -- like printing a `.` every second until a task completes. These can be held in the buffer and not written out consistently. A workaround to handle this might be to just push through single character writes unfiltered. This means that a malicious actor could exfiltrate secrets by writing them one character at a time, but there are plenty of ways a malicious actor can exfiltrate secrets - I think preventing accidental disclosure is a more reasonable target.

dbaggerman:
@lox, @keithpitt, let me know how you feel about the current state of this PR.

Cheers.

lox:
I'll do some manual testing next week @dbaggerman and then we'll see if we can get it merged!

toolmantim:
From a plugin perspective, will it be possible for a plugin like https://github.com/seek-oss/aws-sm-buildkite-plugin to mark it's exported secrets as redacted via `BUILDKITE_REDACTED_VARS`?

dbaggerman:
@toolmantim, that's the intention. `aws-sm-buildkite-plugin` exports variables from an `environment` hook, which means they're visible to `buildkite-agent`. That means they'll get redacted as long as the exported variable name matches one of the patterns in `BUILDKITE_REDACTED_VARS`.

We use a global `environment` hook to export a set of common secrets, but otherwise it's the same  as my use case.

What won't get redacted is variables populated inside the command itself (I expect this extends to `command` hooks), which won't be visible to `buildkite-agent` - at least until after the job has run.

toolmantim:
Nice one @dbaggerman! So does that mean for something like this:

```yml
steps:
  - commands: "..."
    plugins:
      - seek-oss/aws-sm#v2.0.0:
          env:
            SOME_SECRET: the-secret-id
```

…that the plugin could implement some code like the following, that ensures that any secrets it exports get added to the redacted list?

```bash
BUILDKITE_REDACTED_VARS="${BUILDKITE_REDACTED_VARS:-},SOME_SECRET"
```

dbaggerman:
@toolmantim not exactly. `BUILDKITE_REDACTED_VARS` is set in the agent configuration and hooks can't overwrite/clobber it, but it's now a glob pattern rather than a list of specific variable names. The default is `*_PASSWORD,*_SECRET,*_TOKEN`. In your example, `SOME_SECRET` would be redacted because it matches `*_SECRET`.

Allowing plugins like `aws-sm-buildkite-plugin` to append to (but not clobber) `BUILDKITE_REDACTED_VARS` seems like an enhancement that could be added later.

toolmantim:
Okay cool, yep that makes total sense @dbaggerman. Thanks for clarifying! 🙏🏼🙏🏼🙏🏼

lox:
I'm tempted to allow hooks (which includes plugins) to update or override `BUILDKITE_REDACTED_VARS`. Otherwise it won't be possible to configure `BUILDKITE_REDACTED_VARS` in `environment` hooks, or to update them in plugins.

The downside is that you can use a plugin or a repository hook to unredact things, but if we are considering that attack vector there are a million other ways to exfiltrate the secrets.

lox:
One question @keithpitt had was whether we'd put a minimum length on the redacted values. For instance, what if someone accidentally redacts `1` or `e`?

bootstrap/redactor.go
```diff
@@ -0,0 +1,197 @@
...
+	// We can offset the next Write processing by how far cursor is ahead of
+	// the end of this input segment
+	redactor.offset = cursor - len(input)
+
+	return len(input), err
+}
+
+// Flush should be called after the final Write. This will Write() anything
+// retained in case of a partial match and reset the output buffer.
+func (redactor Redactor) Sync() error {
```
lox: I tend to think `Flush` was the better method name, did the semantics change to make `Sync` more appropriate?

keithpitt:
> I'm tempted to allow hooks (which includes plugins) to update or override BUILDKITE_REDACTED_VARS. Otherwise it won't be possible to configure BUILDKITE_REDACTED_VARS in environment hooks, or to update them in plugins.

I can't think of a use case where overriding would be useful, upgrading maybe, but globbing kinda makes this not really a problem. Unless you name your secret env's something like `OMG_PASSWORD_PARTY`, then for the most part, shouldn't it all be fine?

keithpitt:
> One question @keithpitt had was whether we'd put a minimum length on the redacted values. For instance, what if someone accidentally redacts 1 or e?

Maybe minimum length of a redaction is 6?

image

toolmantim:
> globbing kinda makes this not really a problem

What about when the people that set up and configure the agents aren’t necessarily the same people that setup and configure pipelines and secrets, and you have something like `GITHUB_API_KEY` which happens to not be in the standard list? This is common for the elastic CI stack, where reconfiguring agents (e.g. changing the elastic CI stack parameters) might require rolling the whole agent fleet, which might be running jobs for hundreds of other pipelines. At the moment they just have to change a pipeline env file in an S3 bucket… and they might not even know how the agent has been configured.

It’d be nice if we could put the secrets/redacting power in the hands of non-admins. It works well for secrets managers too, because often you set the secret there, and don’t touch agents or hooks, but instead use a plugin to load it / expose it on the step that needs it.

Thinking about the docs (https://buildkite.com/docs/pipelines/secrets) for the current PR we’d probably add a new section, talking about the agent config. But if you could add to the redacted list, maybe we’d make all the examples that are like this:

```bash
#!/bin/bash
set -euo pipefail

if [[ "$BUILDKITE_STEP_KEY" == "trigger-github-deploy" ]]; then
  export GITHUB_API_KEY="bd0fa963610b..."
fi
```

be like this:

```bash
#!/bin/bash
set -euo pipefail

if [[ "$BUILDKITE_STEP_KEY" == "trigger-github-deploy" ]]; then
  export GITHUB_API_KEY="bd0fa963610b..."
  export BUILDKITE_REDACTED_VARS="${BUILDKITE_REDACTED_VARS:-}, GITHUB_API_KEY"
fi
```

Sounds like @dbaggerman has ideas for how it can be done as a stage II though, after this catch-all/blanket glob approach?

lox:
Yeah, I agree @toolmantim, also it's nice to be able to configure it from your `environment` hook for things like the Elastic Stack. 

IMO if we allow things like git flags to be configured in hooks allowing redacted vars makes sense too.

keithpitt:
```
export BUILDKITE_REDACTED_VARS=""
```

Seems like a bad time if you can do that from a plugin. And why not just make the default:

```
BUILDKITE_REDACTED_VARS=*_SECRET;*_TOKEN;*_PASSWORD;*_API_KEY;
```

It's easy to add support for overriding from a hook later on, harder to take that feature away.

I don't want this PR to go on forever (I really appreciate the work you've done @dbaggerman, so I'm happy to step out in the spirit of getting this over the line).

lox:
Yeah, IMO let's get this merged pretty much as is.

lox:
Should we put this behind an experiment?

lox:
IMO, yes, it should be behind an experiment initially.  

So the last things that need to happen pre-merge in this are:

* [ ] Only enable the Redactor if experiment is set (e.g `if experiments.IsEnabled("output-redactor")`)
* [ ] Change Redactor.Sync to Redactor.Flush (unless there is good reason not too)
* [ ] Ignore (and possibly warn about) any secrets shorter than 5 characters?

bootstrap/redactor.go
dbaggerman: Because naming things is hard. I was thinking of the `sync`/`fsync` system calls in linux for triggering an immediate write of the file system buffers. However, I just noticed there's also an `fflush` libc function for doing almost the same thing (write the libc buffer to the operating system).

I'm happy to change it back.

lox:
Thanks so much for the excellent PR and being patient with our feedback @dbaggerman!

blakestoddard:
Any chance this comes to agent "input" values in build logs, too? Say I have two steps in my pipeline: #1: that runs a pipeline generation script (that generates the YAML for step #2), and #2: a Docker build step (not using a plugin, it's just a straight `docker build . . .`. That `docker build` command passes $GITHUB_TOKEN to Docker for private gems, etc., and it is redacted properly from the Agent YAML output that the pipeline generate script spits out in step #1, but it is not redacted from the Agent input in step #2 where the output is like so:
```
Running commands
# Using output-redactor experiment 🧪
$ docker build --build-arg github_token=abc123imnotredactedbutishouldbe . . .
<docker build output is here>
```

dbaggerman:
@blakestoddard, the redactor applies to output captured from the command. I don't think the current implementation applies to output generated by the agent itself.

I would avoid injecting passwords into the pipeline itself, as that means it will be uploaded to Buildkite's servers and will be visible from the uploaded YAML on the timeline tab regardless of redaction. If possible, try moving the password lookup to an `environment` or `pre-command` hook.

blakestoddard:
Totally, that’s a work in progress because it requires deeper changes in some of our pipelines than something I can do in one fell swoop.

PR #1112, opened 2019-10-21
Fetch only branch vs remote update

lox:
This changes the default for the `git-mirrors` experiment to fetch only the branch vs a full `git remote update`. As reported in #1100, this can be a hugely expensive operation in large repositories.

---

_@pda update:_ Closes #1100 — this is the same idea, but without being user configurable.

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -1078,8 +1078,8 @@ func (b *Bootstrap) updateGitMirror() (string, error) {
 		return "", err
 	}
 
-	// Update our mirror
-	if err := b.shell.Run("git", "--git-dir", mirrorDir, "remote", "update", "--prune"); err != nil {
+	// Fetch the build branch from the upstream repository into the mirror.
+	if err := b.shell.Run("git", "--git-dir", mirrorDir, "fetch", "origin", b.Branch); err != nil {
 		return "", err
 	}
 


bootstrap/bootstrap.go
```diff
@@ -1058,8 +1058,10 @@ func (b *Bootstrap) updateGitMirror() (string, error) {
 		return "", err
 	}
 
-	// Update our mirror
-	if err := b.shell.Run("git", "--git-dir", mirrorDir, "remote", "update", "--prune"); err != nil {
+	// Update our mirror, but just the branch requested. Otherwise on huge repositories this can be
+	// exceedingly slows to run
+	gitFetchFlags := fmt.Sprintf("--git-dir %q %s", mirrorDir, b.GitFetchFlags)
+	if err := gitFetch(b.shell, gitFetchFlags, "origin", b.Branch); err != nil {
```
pda, line 1064: I wonder if we actually want to pass `b.GitFetchFlags` to git-mirror operations.

It was added in https://github.com/buildkite/agent/pull/957 where it seems the use-case was for `--depth=1` to maintain shallow clones. Do we want shallow cloned mirrors? I wonder what else people might be using GitFetchFlags for, and whether they should or shouldn't apply to mirrors.

We have separate `GitCloneFlags` and `GitCloneMirrorFlags` so that config isn't shared. I think if anything we'd need to add `GitFetchMirrorFlags`. But maybe we can get away without user-configurable fetch flags for the mirror?

bootstrap/bootstrap.go
```diff
@@ -1058,8 +1058,10 @@ func (b *Bootstrap) updateGitMirror() (string, error) {
 		return "", err
 	}
 
-	// Update our mirror
-	if err := b.shell.Run("git", "--git-dir", mirrorDir, "remote", "update", "--prune"); err != nil {
+	// Update our mirror, but just the branch requested. Otherwise on huge repositories this can be
+	// exceedingly slows to run
+	gitFetchFlags := fmt.Sprintf("--git-dir %q %s", mirrorDir, b.GitFetchFlags)
```
pda, line 1063: See also https://github.com/buildkite/agent/pull/1100/files#r442633344 — I don't think it's valid for `--git-dir` to go after the `fetch` subcommand.

pda:
I've changed this to basically run:

```sh
git --git-dir $mirrorDir fetch origin $branch
```

It's important that `--git-dir` goes between `git` and `fetch`.

We're not passing `GitFetchFlags` because those are for the main repo, not the mirror. We can invent `GitFetchMirrorFlags` in future if somebody needs it.

I investigated `fetch`'s `--prune` option, but I think in the context of pulling a specific branch it has no effect.

I've tested this locally, both in create-mirror and update-mirror mode.

Any pre-merge thoughts from @lox or @eberle1080 or anybody else?

frankh:
@pda this seems to have broken git-mirrors for PRs off forks, i now get 

```
git --git-dir /usr/local/var/buildkite-agent/repos/git-github-com-my-repo-git fetch origin frankh/my-branch
--
  | fatal: couldn't find remote ref frankh
```

PR #1113, opened 2019-10-23
Convert windows paths to unix ones on artifact download

### LINKED ISSUE #354: Normalise upload/download paths with Windows
It shouldn't matter what sort of `/` you use on Windows `foo/bar.zip`or `foo\bar.zip` between the Agent API and the Agent itself it should be able to find the right artifact for you.

lox:
Currently, a file uploaded on a windows agent named `my\folder\blah.txt` will be downloaded on a UNIX agent as the literal file of the same name vs the folders and file structure implied. 

This is the minimal change to fix that. It preserves the file separator that the file is uploaded with and just normalizes it on non-windows hosts on download. 

Closes #354.

### DIFF
--- a/agent/artifact_downloader.go
+++ b/agent/artifact_downloader.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	"github.com/buildkite/agent/logger"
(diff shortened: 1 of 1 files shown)


petemounce:
Will there be a follow-on to fix the _upload_ to normalise paths to be `/`-separated? With this change by itself, each thing that accesses artifacts _without_ doing so via `buildkite-agent artifact download` still needs to fix this behaviour.

cc @jamiebrynes7 as I think you'll still need your plugin fork.

lox:
@petemounce the thinking was that we'd make either windows or linux glob formats work for downloading artifacts vs normalizing the paths on upload 🤔

lox:
> With this change by itself, each thing that accesses artifacts without doing so via buildkite-agent artifact download still needs to fix this behaviour.

Do you have any examples of the sorts of things you mean?

petemounce:
Totally missed the followups, sorry!

`gsutil cp` / `aws s3 cp` / etc

My thinking was that since artifacts are accessed over http, and http path separators are /, normalise on upload provides the least surprising cross-platform UX.

DuBistKomisch:
:+1: to fixing the download path slash matching as well, you still need to know ahead of time whether to use / or \

there's also a related annoying thing when downloading artifacts via the website where the attachment filename includes the path with _ in place of the backslashes, instead of stripping the the path completely like it would with forward slashes

this PR is still a huge improvement though

PR #1116, opened 2019-11-04
Propagate signals in intermediate bash shells

lox:
If you execute multiple commands, e.g:

```yaml
steps:
  -label: "My commands" 
   commands:
     - ./process1
     - ./process2
```

then we join them together in a single bash shell as `/bin/bash -e -c './process1\b./process2`, which has the side effect of creating an intermediate shell in between the bootstrap process and your commands. By default, most shells won't automatically propagate signals to subprocesses, so when the job is cancelled the signal gets lost in this intermediate shell. 

This could be manually fixed by adding a trap as the first command:

```yaml
steps:
  -label: "My commands" 
   commands:
     - trap "kill -- \$\$" TERM QUIT INT
     - ./process1
     - ./process2
```

That ends up kind of noisy and for those not versed in shell voodoo, confusing. 

This PR automatically adds this into non-script commands.

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -1444,10 +1444,12 @@ func (b *Bootstrap) defaultCommandPhase() error {
 			b.shell.Commentf("Detected deprecated docker environment variables")
 		}
 		return runDeprecatedDockerIntegration(b.shell, []string{cmdToExec})
-	} else if !commandIsScript {
+	}
 
-		// If we aren't running a script and we aren't running the command in our deprecated docker integration
-		// this adds a trap to ensure that an intermediate shell doesn't swallow signals from cancellation
+	// If we aren't running a script, try and detect if we are using a posix shell
+	// and if so add a trap so that the intermediate shell doesn't swallow signals
+	// from cancellation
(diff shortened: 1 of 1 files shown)


toolmantim:
Noice fix! 👌🏼

bootstrap/bootstrap.go
```diff
@@ -1463,6 +1470,22 @@ func (b *Bootstrap) defaultCommandPhase() error {
...
+// isPosixShell attempts to detect posix shells (e.g bash, sh, zsh )
+func isPosixShell(shell []string) bool {
+	bin := filepath.Base(shell[0])
+
+	if filepath.Base(shell[0]) == `env` {
+		bin = filepath.Base(shell[1])
+	}
+
+	switch bin {
+	case `bash`, `sh`, `zsh`, `ksh`:
```
matthewd: \+`dash` ?

lox:
Past @lox, just as an FYI, I find trap command in all build output really confusing. The incantation of `trap "kill -- \$\$" TERM QUIT INT` could probably be hidden in non-debug mode.

PR #1117, opened 2019-11-07
Remove vendor in favour of go modules

lox:
We introduced `go.mod` and `go.sum` a long while back, but have been sticking with committed `vendor/` whilst support for go modules progressed in Go. 

It's quite stable now, so I think we can rely on it.


yob:
I'm less across the golang ecosystem than you two, but this seems like a reasonable move to me.

I had a crack at updating the docker-compose environment so go does the modules thing automatically.

yob:
I rebased onto master to resolve a conflict and force pushed - the build is still green

PR #1121, opened 2019-11-19
Download absolute-path artifacts to correct path

### LINKED ISSUE #1035: Artifact is downloaded to incorrect path
We have a build step that runs `buildkite-agent artifact upload /tmp/some_file.json`. In a following step, we run:

```
buildkite-agent artifact download tmp/some_file.json /tmp
```

The `INFO` logs from the command indicate that the file was downloaded successfully to `/tmp`. However, in reality, it ends up in `./tmp`. That is to say, instead of being downloaded to `/tmp`, it is downloaded to a path under the working directory.

Here is an example of this occurring, with debug output enabled:

```
2019-06-18 14:31:52 INFO   Searching for artifacts: "tmp/parse-dev-plan-c8a3072894b6dde220e112d501c91d249b4e1748.json"
--
  | 2019-06-18 14:31:52 DEBUG  GET https://agent.buildkite.com/v3/builds/dddf4bf8-cd29-4603-9f9a-dbdcd6275b38/artifacts/search?query=tmp%2Fparse-dev-plan-c8a3072894b6dde220e112d501c91d249b4e1748.json
  | 2019-06-18 14:31:52 DEBUG  ↳ GET https://agent.buildkite.com/v3/builds/dddf4bf8-cd29-4603-9f9a-dbdcd6275b38/artifacts/search?query=tmp%2Fparse-dev-plan-c8a3072894b6dde220e112d501c91d249b4e1748.json (HTTP/2.0 200 OK 87.342012ms)
  | 2019-06-18 14:31:52 INFO   Found 1 artifacts. Starting to download to: /tmp
  | 2019-06-18 14:31:52 DEBUG  Downloading https://s3.amazonaws.com/buildkiteartifacts.com/<REDACTED> to tmp/parse-dev-plan-c8a3072894b6dde220e112d501c91d249b4e1748.json
  | 2019-06-18 14:31:52 INFO   Successfully downloaded "tmp/parse-dev-plan-c8a3072894b6dde220e112d501c91d249b4e1748.json" 3 bytes
```

The artifact is downloaded successfully, but to a new `tmp` directory that is created within the working directory.

Not sure if this is related to #694.

Helcaraxan:
The handling of artifacts uploaded with absolute paths is currently
undefined and resulted in it being handled as a relative path instead.

We add explicit handling of absolute paths. The download configuration
of such artifacts still requires a download path to be specified as the
nature of paths of artifacts that will end up being downloaded can only
be determined at runtime.

We also add a new test to prevent regression of the newly implemented
behaviour as well explicit documentation in the CLI help.

Fixes #1035.

### DIFF
--- a/agent/artifact_downloader_test.go
+++ b/agent/artifact_downloader_test.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/buildkite/agent/v3/api"
 	"github.com/buildkite/agent/v3/logger"
+	"github.com/stretchr/testify/assert"
 )
 
 func TestArtifactDownloaderConnectsToEndpoint(t *testing.T) {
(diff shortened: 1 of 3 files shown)


pda:
Thanks very much for the PR @Helcaraxan.

We do have some concerns about changing the agent so that it will write outside the current working directory based on payload from the artifacts API. Especially if it's ignoring a relative destination argument.

 I would compare it to the behaviour of `tar`:

```
$ echo hello > /tmp/foo

$ tar -cf test.tar /tmp/foo
tar: Removing leading `/' from member names

$ tar -tf test.tar
tmp/foo

$ tar -xvf test.tar
tmp/foo
```

For `tar` and for `buildkite-agent artifact download`, the _caller_ should be able to opt into writing relative to the root of the filesystem explicitly;

```
$ rm /tmp/foo

$ tar -C / -xvf test.tar
tmp/foo

$ cat /tmp/foo
hello
```

I think the equivalent would be;

```
buildkite-agent artifact download foo /
```

What do you think?

Helcaraxan:
Hello @pda. I understand the concern. The proposed solution looks good to me. I do would propose to actually make this clearer in the documentation. Do you want me to alter this PR to take care of this?

pda:
Hey @Helcaraxan - I think the documentation deficiency is largely fixed here:
https://github.com/buildkite/agent/pull/1183

I'll make sure these changes land on https://buildkite.com/docs/agent/v3/cli-artifact#downloading-artifacts soon.

Thanks for pointing this out.

PR #1122, opened 2019-11-20
add powershell support for hooks

pdemirdjian:
This adds the ability to write native powershell hooks for windows pipelines so that executing a batch file to call a powershell script is unnecessary.  The hook still creates a batch file to handle any new environment variables and pass the exit status properly.

### DIFF
--- a/bootstrap/hook.go
+++ b/bootstrap/hook.go
@@ -51,6 +51,7 @@ func newHookScriptWrapper(hookPath string) (*hookScriptWrapper, error) {
 	var scriptFileName string = `buildkite-agent-bootstrap-hook-runner`
 	var isBashHook bool
 	var isPwshHook bool
+	var isWindows = runtime.GOOS == "windows"
 
 	// we use bash hooks for scripts with no extension, otherwise on windows
 	// we probably need a .bat extension
@@ -100,7 +101,7 @@ func newHookScriptWrapper(hookPath string) (*hookScriptWrapper, error) {
 
 	// Create the hook runner code
 	var script string
-	if runtime.GOOS == "windows" && !isBashHook && !isPwshHook {
+	if isWindows && !isBashHook && !isPwshHook {
 		script = "@echo off\n" +
 			"SETLOCAL ENABLEDELAYEDEXPANSION\n" +
 			"SET > \"" + h.beforeEnvFile.Name() + "\"\n" +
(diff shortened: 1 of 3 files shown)


bootstrap/hook.go
```diff
@@ -50,10 +50,14 @@ func newHookScriptWrapper(hookPath string) (*hookScriptWrapper, error) {
...
 	var scriptFileName string = `buildkite-agent-bootstrap-hook-runner`
 	var isBashHook bool
+	var isPwshHook bool
 
 	// we use bash hooks for scripts with no extension, otherwise on windows
 	// we probably need a .bat extension
-	if filepath.Ext(hookPath) == "" {
+	if filepath.Ext(hookPath) == ".ps1" {
+		isPwshHook = true
+		scriptFileName += ".bat"
```
GaryPWhite: wait -- is this adding a .bat to powershell?

pdemirdjian: if you look at line 51 above it’s actually the name of the file generated on the fly to execute the hook wrapped by some code to capture exit code and env variable changes. so basically the flow is:
- determine hook file type (bash, batch file, powershell)
- if batch or powershell, add .bat to the hook wrapper
- if powershell, change from CALL to a powershell.exe call on the file

bootstrap/hook.go
```diff
@@ -105,6 +109,15 @@ func newHookScriptWrapper(hookPath string) (*hookScriptWrapper, error) {
...
 			"EXIT %" + hookExitStatusEnv + "%"
+	} else if runtime.GOOS == "windows" && isPwshHook {
+		script = "@echo off\n" +
+			"SETLOCAL ENABLEDELAYEDEXPANSION\n" +
+			"SET > \"" + h.beforeEnvFile.Name() + "\"\n" +
+			"powershell.exe -ExecutionPolicy Bypass -NoProfile -NonInteractive -File \"" + absolutePathToHook + "\"\n" +
+			"SET " + hookExitStatusEnv + "=!ERRORLEVEL!\n" +
+			"SET " + hookWorkingDirEnv + "=%CD%\n" +
+			"SET > \"" + h.afterEnvFile.Name() + "\"\n" +
+			"EXIT %" + hookExitStatusEnv + "%"
```
sj26: The purpose of these wrappers is to execute the provided command script in the same context and capture any different to the environment files so we can fold those changes into the parent env and consecutive hooks and scripts — is there a way to do this in powershell, too?

sj26:
This is looking good! But I think we'll have to figure out how to make a powershell wrapper which captures env differences, if we can.

bootstrap/hook.go
```diff
@@ -105,6 +109,14 @@ func newHookScriptWrapper(hookPath string) (*hookScriptWrapper, error) {
...
 			"SET > \"" + h.afterEnvFile.Name() + "\"\n" +
 			"EXIT %" + hookExitStatusEnv + "%"
+	} else if runtime.GOOS == "windows" && isPwshHook {
+		script = "$ErrorActionPreference = \"STOP\"\n" +
+			"Get-ChildItem Env: | Foreach-Object {\"$($_.Name)=$($_.Value)\"} | Set-Content \"" + h.beforeEnvFile.Name() + "\"\n" +
+			absolutePathToHook + "\n" +
+			"if ($LASTEXITCODE -eq $null) {$Env:" + hookExitStatusEnv + " = 0} else {$Env:" + hookExitStatusEnv + " = $LASTEXITCODE}\n" +
+			"$Env:" + hookWorkingDirEnv + " = $PWD | Select-Object -ExpandProperty Path\n" +
+			"Get-ChildItem Env: | Foreach-Object {\"$($_.Name)=$($_.Value)\"} | Set-Content \"" + h.afterEnvFile.Name() + "\"\n" +
+			"exit $Env:" + hookExitStatusEnv
```
pda: Would Go's backtick-strings work here to (a) reduce escaping and (b) help distinguish between outer-Go-strings and inner-PS-strings? I think it would make it quite a bit more readable. Like this:

```diff
-		script = "$ErrorActionPreference = \"STOP\"\n" +
-			"Get-ChildItem Env: | Foreach-Object {\"$($_.Name)=$($_.Value)\"} | Set-Content \"" + h.beforeEnvFile.Name() + "\"\n" +
-			absolutePathToHook + "\n" +
-			"if ($LASTEXITCODE -eq $null) {$Env:" + hookExitStatusEnv + " = 0} else {$Env:" + hookExitStatusEnv + " = $LASTEXITCODE}\n" +
-			"$Env:" + hookWorkingDirEnv + " = $PWD | Select-Object -ExpandProperty Path\n" +
-			"Get-ChildItem Env: | Foreach-Object {\"$($_.Name)=$($_.Value)\"} | Set-Content \"" + h.afterEnvFile.Name() + "\"\n" +
-			"exit $Env:" + hookExitStatusEnv
+		script = `$ErrorActionPreference = "STOP"\n` +
+			`Get-ChildItem Env: | Foreach-Object {"$($_.Name)=$($_.Value)"} | Set-Content "` + h.beforeEnvFile.Name() + `"\n` +
+			absolutePathToHook + `\n` +
+			`if ($LASTEXITCODE -eq $null) {$Env:` + hookExitStatusEnv + ` = 0} else {$Env:` + hookExitStatusEnv + ` = $LASTEXITCODE}\n` +
+			`$Env:` + hookWorkingDirEnv + ` = $PWD | Select-Object -ExpandProperty Path\n` +
+			`Get-ChildItem Env: | Foreach-Object {"$($_.Name)=$($_.Value)"} | Set-Content "` + h.afterEnvFile.Name() + `"\n` +
+			`exit $Env:` + hookExitStatusEnv
```

bootstrap/hook.go
```diff
@@ -50,10 +50,14 @@ func newHookScriptWrapper(hookPath string) (*hookScriptWrapper, error) {
 	var err error
 	var scriptFileName string = `buildkite-agent-bootstrap-hook-runner`
 	var isBashHook bool
+	var isPwshHook bool
```
pda: I wonder if this should change to an enum (or some Go equivalent) rather than two booleans that could contradict each other if both are true.

pda: Also maybe worth copying the `var isWindows = runtime.GOOS == "windows"` from `shell.go` to make the conditionals below a bit simpler to read.

bootstrap/shell/shell.go
```diff
@@ -267,10 +268,19 @@ func (s *Shell) RunScript(path string, extra *env.Environment) error {
...
+		}
+		command = "powershell.exe"
+		args = []string{"-file", path}
+
+
 	case !isWindows && isBash:
 		command = "/bin/bash"
 		args = []string{"-c", path}
 
+
```
pda: Looks like unintended blank line, but no big deal.

bootstrap/hook.go
pdemirdjian: thought about how this could be accomplished in Go but haven't fully fledged out the implementation yet. I can revisit this in the future

pda:
Thanks @pdemirdjian :+1:

PR #1124, opened 2019-11-21
Improve OS and arch compatibility

oxplot:
Buildkite agent should be buildable across all platforms that Go supports, but it has some platforms specific code scattered around, both in its own code base and some in deps, that get in the way.

This PR disables PTY and machine fingerprinting functionality on platforms that are not supported by the relevant modules, as these are not crucial to the operation of the agent (as evident by the error checking around where they're used).


vendor/github.com/denisbrodbeck/machineid/id_unsupported.go
```diff
@@ -0,0 +1,12 @@
+// +build !freebsd,!netbsd,!openbsd,!dragonfly,!windows,!linux,!darwin
```
pda: I don't think we can directly patch vendor code without it leading to problems later. There's a bit about our module/vendoring here: https://github.com/buildkite/agent#dependency-management

A pull request to https://github.com/denisbrodbeck/machineid might be a better option.

process/pty.go
```diff
@@ -1,4 +1,4 @@
-// +build !windows
+// +build linux,netbsd,freebsd,dragonfly darwin,386 darwin,amd64 openbsd,386 openbsd,amd64
```
pda: Do you have any links/reference/anecdotes explaining why this is the right list of architectures? Same question for `process/pty_unsupported.go`. It would probably be helpful for future generations to understand, and give people confidence to change it if/when required.

vendor/github.com/denisbrodbeck/machineid/id_unsupported.go
oxplot: Ye, I thought about that first but seeing the how stale the upstream is, I took a shot to have it patched here. I will submit this patch to upstream as well.

process/pty.go
oxplot: Thanks, I added some comments.

vendor/github.com/denisbrodbeck/machineid/id_unsupported.go
oxplot: Actually, came to think of it — the right approach is probably to abstract away the machineid lib and use build tags in this project.

PR #1128, opened 2019-11-26
Detect ansi clear lines and add ansi timestamps

lox:
Currently the ANSI timestamps don't show up certain output (like docker) that uses creative ANSI control characters for cursor movements.

This detects things like `\x1b[K` which clears the current line and includes the ANSI timestamp technology.

### DIFF
--- a/process/prefixer.go
+++ b/process/prefixer.go
@@ -3,6 +3,8 @@ package process
 import (
 	"bytes"
 	"io"
+	"log"
+	"unicode/utf8"
 )
 
 type Prefixer struct {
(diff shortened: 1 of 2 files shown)


keithpitt:
<3<3<3

keithpitt:
For further context, currently Buildkite can't show line timestamps like here:

image

This change will allow us to show timestamps on each of those lines.

process/prefixer.go
```diff
@@ -32,19 +34,40 @@ func (p *Prefixer) Write(data []byte) (n int, err error) {
 	offset := 0
 	out := make([]byte, 0, len(data))
 
-	// loop through newlines
-	for offset < len(data) {
-		next := bytes.IndexRune(data[offset:], '\n')
+	// loop through line breaks and add prefix
+	for l := len(data); offset < l; {
+		// find either a newline or an escape char
+		next := bytes.IndexAny(data[offset:], "\n\x1b")
```
pda: It would be nice to standardise on hex `\x1b` or octal `\033` for escape chars in the codebase/docs/etc.
Currently there’s a bit of both in our codebase. Octal is a ridiculous radix. So this `\x1b` is good. Basically ignore this comment.

process/prefixer.go
```diff
@@ -32,19 +34,40 @@ func (p *Prefixer) Write(data []byte) (n int, err error) {
 	offset := 0
 	out := make([]byte, 0, len(data))
 
-	// loop through newlines
-	for offset < len(data) {
-		next := bytes.IndexRune(data[offset:], '\n')
+	// loop through line breaks and add prefix
+	for l := len(data); offset < l; {
```
pda: Would a [`bufio.Scanner`](https://golang.org/pkg/bufio/#Scanner) be appropriate here, with a custom split function detecting new line or escape?

lox:
It needs some more testing, I suspect there are some more escape codes that we need to handle. There are some edge cases that I don't handle too.

process/prefixer.go
lox: > Programs that need more control over error handling or large tokens, or must run sequential scans on a reader, should use bufio.Reader instead.

lox: bufio.Scanner fails on long lines, it's a pain.

keithpitt:
Can we merge and see how it goes?

lox:
I’ll test it now.

keithpitt:
I fixed the last remaining bug in the UI for this, so I'm keen to get this merged in.

PR #1135, opened 2019-12-09
Ignore multiple checkout plugin hooks

toolmantim:
The agent [ignores multiple checkout and command hooks](https://buildkite.com/docs/agent/v3/hooks#available-hooks), and in #1055 we extended that to plugins as well for consistency (the previous behaviour was undefined).

This extends the previous PR to add the same check for the `checkout` hook, and adds a warning message for any people that might have been using the now-unsupported behaviour.

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -666,7 +666,21 @@ func (b *Bootstrap) VendoredPluginPhase() error {
 
 // Executes a named hook on plugins that have it
 func (b *Bootstrap) executePluginHook(name string, checkouts []*pluginCheckout) error {
-	for _, p := range checkouts {
+	// Command and checkout hooks are a little different, in that we only execute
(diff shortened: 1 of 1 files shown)


toolmantim:
The pipeline:

```yml
steps:
  - command: pwd
    plugins:
      - thedyrt/skip-checkout#v0.1.1: ~
      - thedyrt/skip-checkout#v0.1.1: ~
      - thedyrt/skip-checkout#v0.1.1: ~
```

Yields:

image


The pipeline:

```yml
  - command: pwd
    plugins:
      - docker#v3.3.0:
          image: "golang:1.11"
          mount-checkout: false
      - docker#v3.3.0:
          image: "golang:1.11"
          mount-checkout: false
      - docker#v3.3.0:
          image: "golang:1.11"
          mount-checkout: false
```

Yields:

image

toolmantim:
Should I just merge, even though the Windows tests are broken?

PR #1138, opened 2019-12-10
Job Acquisition Technology™️

keithpitt:
image

This change adds a new `--acquire-job` option to `buildkite-agent start`. You use it like this:

```sh
buildkite-agent start --token "xx" --acquire-job "f6f1a0c9-3431-451d-942d-846dfd1fb623"
```

When the agent boots, instead of doing the normal "ping until you find work" routine, it'll specifically ask Buildkite if it can acquire the job provided (acquire is the word we use to describe an agent self-assigning and accepting a job). Once the job has been acquired, it'll run the job and exit when it's finished.

If it fails to acquire the job, it'll just bail:

image

In an original cut of this code I just had it as `buildkite-agent start --job`, but I didn't really mean much in context of other stuff. For example, we've got `disconnect-after-job` which is a verbose argument, so I figured `--acquire-job` is pretty clear as it what it's doing.

### DIFF
--- a/agent/agent_configuration.go
+++ b/agent/agent_configuration.go
@@ -29,4 +29,5 @@ type AgentConfiguration struct {
 	Shell                      string
 	Profile                    string
 	RedactedVars               []string
+	AcquireJob                 string
 }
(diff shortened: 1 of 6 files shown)


keithpitt:
Ah, turns out we @lox had already suggested it be `--acquire-job` - here's proof:

image

PR #1146, opened 2020-01-14
Fix AgentPool to disconnect if AgentWorker.Start fails 

keithduncan:
When using `--acquire-job` to start a one-shot agent to pull a specific job, there exists a race condition when the job can be cancelled on Buildkite before the scheduled agent actually boots up. When this happens, `apiClient.AcquireJob(jobId)` returns a 422 which isn't retried, but returning an error from `AgentWorker.Start` means that `AgentPool.runWorker` skips the disconnect step.

I dug into this because I had some ghost one-shot agents that I was worried were using capacity. I checked the logs and saw that they were exiting immediately because they don’t `POST /v3/disconnect` to the Buildkite API they hang around in the agents list until they are pruned out for inactivity. I don’t think this causes any problems for scheduling because `AgentRegisterRequest.IgnoreInDispatches` is set, but it does make it appear there are more agents running than there are in practice.

@keithpitt what do you think?

### DIFF
--- a/agent/agent_pool.go
+++ b/agent/agent_pool.go
@@ -51,20 +51,11 @@ func (r *AgentPool) runWorker(worker *AgentWorker, im *IdleMonitor) error {
 	if err := worker.Connect(); err != nil {
 		return err
 	}
+	// Ensure the worker is disconnected at the end of this function.
+	defer worker.Disconnect();
 
-	// Starts the agent worker and wait for it to finish
-	err := worker.Start(im)
-
-	// Now that the agent has finished, we can disconnect it
-	disconnect := worker.Disconnect()
-
-	// If start exited due to an error return that
-	if err != nil {
-		return err
-	}
-
-	// Otherwise return the disconnect result
-	return disconnect
+	// Starts the agent worker and wait for it to finish.
+	return worker.Start(im)
 }
 
 func (r *AgentPool) Stop(graceful bool) {
--- a/agent/agent_pool.go
+++ b/agent/agent_pool.go
@@ -53,16 +53,18 @@ func (r *AgentPool) runWorker(worker *AgentWorker, im *IdleMonitor) error {
 	}
 
 	// Starts the agent worker and wait for it to finish
-	if err := worker.Start(im); err != nil {
+	err := worker.Start(im)
+
+	// Now that the agent has finished, we can disconnect it
+	disconnect := worker.Disconnect()
+
+	// If start exited due to an error return that
+	if err != nil {
 		return err
 	}
 
-	// Now that the agent has stopped, we can disconnect it
-	if err := worker.Disconnect(); err != nil {
-		return err
-	}
-
-	return nil
+	// Otherwise return the disconnect result
+	return disconnect
 }
 
 func (r *AgentPool) Stop(graceful bool) {
--- a/agent/agent_worker.go
+++ b/agent/agent_worker.go
@@ -412,9 +412,7 @@ func (a *AgentWorker) AcquireAndRunJob(jobId string) error {
 
 	// If `acquiredJob` is nil, then the job was never acquired
 	if acquiredJob == nil {
-		a.logger.Warn("Failed to acquire job: %v", err)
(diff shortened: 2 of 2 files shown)


agent/agent_worker.go
```diff
@@ -412,7 +412,9 @@ func (a *AgentWorker) AcquireAndRunJob(jobId string) error {
 
 	// If `acquiredJob` is nil, then the job was never acquired
 	if acquiredJob == nil {
-		return fmt.Errorf("Failed to acquire job: %v", err)
+		a.logger.Warn("Failed to acquire job: %v", err)
+		// Return nil so that the AgentPool disconnnects this one-shot agent
+		return nil
```
keithpitt: Oh this is a good find! I need to double check the code, but should we be returning `err` here perhaps?

keithduncan: I think returning the `err` here will cause the same behaviour as before.

Where we call `s.Break()` above :point_up:, the case we want to disconnect the agent under, we know `err` is non-nil. That function returns the non-nil `err` to `retry.Do` which then returns it to us. If we return it from here the `AgentPool` won't disconnect us and leaves the agent hanging until it is pruned.

Maybe this is actually a problem with the `AgentPool` and we want it to try disconnecting the agent when `Agent.Start` returns error or no?

keithpitt: > Maybe this is actually a problem with the AgentPool and we want it to try disconnecting the agent when Agent.Start returns error or no?

I reckon that might be the winner. Are you ok to make that change? (I appreciate you bug finding and fixing, so I don't want to push my luck with code change requests :P)

keithduncan:
I’ve changed the `AgentPool` to always disconnect, let me know what you think 😄

agent/agent_pool.go
```diff
@@ -53,16 +53,18 @@ func (r *AgentPool) runWorker(worker *AgentWorker, im *IdleMonitor) error {
 	}
 
 	// Starts the agent worker and wait for it to finish
-	if err := worker.Start(im); err != nil {
-		return err
-	}
+	err := worker.Start(im)
```
keithpitt: Perhaps a comment here ala "we need to ensure the agent disconnects successfully if a worker process fails to start, so this error will be checked later on"

keithduncan: I also thought about using a `defer` after `Connect`, would that be more obvious without a comment?

keithpitt: Oh yeah, good idea.

keithduncan:
This time with defer™ technology but this version would mask a disconnect error. We could defer a closure with a named `err` parameter so that `Disconnect` could overwrite it but… that seems gnarly compared to the non-defer version. What do you think?

When looking at what happens to the errors I noticed that `AgentPool.Start` only pulls a single error from the `errs` channel, should this be a `for err := range errs` to grab them all until all the workers have finished and `close(errs)` runs?

agent/agent_pool.go
```diff
@@ -51,18 +51,11 @@ func (r *AgentPool) runWorker(worker *AgentWorker, im *IdleMonitor) error {
 	if err := worker.Connect(); err != nil {
 		return err
 	}
+	// Ensure the worker is disconnected at the end of this function.
+	defer worker.Disconnect();
```
pda: I think the `defer worker.Disconnect()` approach is a good trade-off for discarding the error that `worker.Disconnect()` might return, because;

* `agent.AgentWorker.Disconnect()` has [already logged the disconnect error,](https://github.com/buildkite/agent/blob/67c82ef910518f1718f2c65045a8e70691885da3/agent/agent_worker.go#L493-L496)
* it's questionable whether the caller of `runWorker` would want to consider it an error when both Connect & Start were successful, 
* as you mention, `AgentPool.Start` currently only reports one of N errors anyway
    * this is probably a bug, but it reinforces that this error scenario isn't critical

pda:
This is merged in 17e5216b38b192271c9906873341ed8896fb2aee, GitHub just doesn't know it.

PR #1149, opened 2020-01-21
agent/form_uploader: Implement HTTP Request debug

ticky:
This completes the implementation of the `DebugHTTP` option, which we describe as "dumps all request and response bodies to the log".

It previously did not dump request bodies made by the form_uploader, only their responses.

Implementation borrowed from `api/client`, which already handled this similarly.

### DIFF
--- a/agent/form_uploader.go
+++ b/agent/form_uploader.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"net/http/httputil"
 	"regexp"
+	"strings"
 
 	// "net/http/httputil"
 	"errors"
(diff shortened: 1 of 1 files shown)


agent/form_uploader.go
```diff
@@ -54,6 +55,21 @@ func (u *FormUploader) Upload(artifact *api.Artifact) error {
...
+		// file contents into the debug log (especially if it's been
+		// gzipped)
+		var requestDump []byte
+		if strings.Contains(request.Header.Get("Content-Type"), "multipart/form-data") {
+			requestDump, err = httputil.DumpRequestOut(request, false)
+		} else {
+			requestDump, err = httputil.DumpRequestOut(request, true)
+		}
+
+		u.logger.Debug("\nERR: %s\n%s", err, string(requestDump))
```
keithpitt: Is this log output in the right order? Should we be showing the request dump first then the error? Or is this how the other places do it?

ticky: It’s the same order as the other instance in this file:

https://github.com/buildkite/agent/blob/f99832974503f0d4dd4bff8145d86e2b0f00f2c4/agent/form_uploader.go#L88-L91

PR #1154, opened 2020-01-28
The great "it's" purge

pda:
Fix "it's" that should be "its", and a few "its" that should be "it's", plus one that should be neither.

These were mostly in internal code comments, but also some user-facing docs (CHANGELOG and CLI usage output).

FYI the low-tech way I did this:

```
rg --glob '!vendor' "it's" -l | xargs sed -i '' -e "s/it's/its/g"
git add -p     # manually decide
git restore .  # restore the ones I chose not to add
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -765,7 +765,7 @@ No changes
 
 ## 3.0-beta.19 - 2017-03-29
 ### Added
-- `buildkite-agent start --disconnect-after-job` will run the agent, and automatically disconnect after running it's first job. This has sometimes been referred to as "one shot" mode and is useful when you spin up an environment per-job and want the agent to automatically disconnect once it's finished it's job
+- `buildkite-agent start --disconnect-after-job` will run the agent, and automatically disconnect after running its first job. This has sometimes been referred to as "one shot" mode and is useful when you spin up an environment per-job and want the agent to automatically disconnect once its finished its job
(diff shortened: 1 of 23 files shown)


CHANGELOG.md
```diff
@@ -765,7 +765,7 @@ No changes
 
 ## 3.0-beta.19 - 2017-03-29
 ### Added
-- `buildkite-agent start --disconnect-after-job` will run the agent, and automatically disconnect after running it's first job. This has sometimes been referred to as "one shot" mode and is useful when you spin up an environment per-job and want the agent to automatically disconnect once it's finished it's job
+- `buildkite-agent start --disconnect-after-job` will run the agent, and automatically disconnect after running its first job. This has sometimes been referred to as "one shot" mode and is useful when you spin up an environment per-job and want the agent to automatically disconnect once its finished its job
```
matthewd: Middle one

agent/job_runner.go
```diff
@@ -572,7 +572,7 @@ func (r *JobRunner) onProcessStartCallback() {
 			r.logger.Debug("[JobRunner] Routine that refreshes the job has finished")
 		}()
 		for {
-			// Re-get the job and check it's status to see if it's been
+			// Re-get the job and check its status to see if its been
```
matthewd: Last one

pda:
Nice spotting @matthewd. Semantic satiation indeed :confused:

PR #1156, opened 2020-01-29
Fix plugin execution being skipped with duplicate hook warning

### LINKED ISSUE #1152: Plugin command hook not executed when plugin is not the first
The following pipeline will only execute the artifact download plugin, but not the docker command in agent version 3.18:

```yaml
  - label: my-step
    command: my-command.sh
    plugins:
    - artifacts#v1.2.0:
        download:
        ...
    - docker#v3.5.0:
        ...
```

Will result in no command being executed at all with this warning: 
`⚠️ Warning: Ignoring additional command hook (docker plugin, position 2)`

The cultprint appears to be https://github.com/buildkite/agent/pull/1135. It checks only the first plugin in the list, and then sets hookTypeSeen = true even though it didn't actually run anything.

toolmantim:
In #1135 (and released in v3.18.0) a bug was introduced where plugins hooks were being incorrectly skipped. This fixes the bug by ensuring that the hook exists in the plugin, before checking it for duplicate execution.

This now executes correctly and uploads the artifact:

```yml
steps:
- command: "echo 'foo' > bar.txt"
  plugins:
    - docker#v3.5.0:
        image: bash
    - artifacts#v1.2.0:
        upload: bar.txt
```

The following pipeline was also tested, and outputs the expected warning on execution of the second `docker` plugin:

```yml
steps:
- command: "echo 'foo' > bar.txt"
  plugins:
    - docker#v3.5.0:
        image: bash
    - docker#v3.5.0:
        image: bash
    - artifacts#v1.2.0:
        upload: bar.txt
```

Fixes #1152 

---

_Edit by @pda_: I didn't quite understand the bug, so I dug into the code; here's my summary:

* If any plugin for a step has a checkout hook (same applies to command hooks), only the first plugin gets the opportunity to run its checkout hook;
    * subsequent plugins show the "Ignoring ..." warning message regardless of whether they implement a checkout hook.
* But if none of the plugins implement a checkout hook, there are no warning messages.
* If the first plugin implements a checkout hook, it runs as normal. If it doesn't, nothing happens, as normal.
    * If subsequent steps implement a checkout hook, they are ignored with a warning.
    * If subsequent steps _don't_ implement a checkout hook, the warning appears anyway.
* all of this applies to command hooks as well as checkout hooks.

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -674,8 +674,10 @@ func (b *Bootstrap) executePluginHook(name string, checkouts []*pluginCheckout) 
 	for i, p := range checkouts {
 		// First we verify the plugin actually implements the hook
 		hookPath, err := b.findHookFile(p.HooksDir, name)
-		if err != nil {
+		if os.IsNotExist(err) {
 			continue
+		} else if err != nil {
+			return err
 		}
 
 		// Disallow plugins executing the command or checkout hook more than once
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -290,6 +290,8 @@ func (b *Bootstrap) findHookFile(hookDir string, name string) (string, error) {
 	if p := filepath.Join(hookDir, name); fileExists(p) {
 		return p, nil
 	}
+	// don't wrap os.ErrNotExist without checking callers handle it.
+	// e.g. os.IfNotExist(err) does not handle wrapped errors.
 	return "", os.ErrNotExist
 }
 
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -674,6 +674,10 @@ func (b *Bootstrap) executePluginHook(name string, checkouts []*pluginCheckout) 
 	for i, p := range checkouts {
 		// First we verify the plugin actually implements the hook
 		hookPath, err := b.findHookFile(p.HooksDir, name)
+		// os.IsNotExist() doesn't unwrap wrapped errors (as at Go 1.13).
+		// agent is still go pre-1.13 compatible (I think) so we're avoiding errors.Is().
+		// In future somebody should check if os.IsNotExist() has added support for
+		// error unwrapping, or change this code to errors.Is(err, os.ErrNotExist)
 		if os.IsNotExist(err) {
 			continue
 		} else if err != nil {
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -672,6 +672,13 @@ func (b *Bootstrap) executePluginHook(name string, checkouts []*pluginCheckout) 
 	hookTypeSeen := make(map[string]bool)
 
 	for i, p := range checkouts {
+		// First we verify the plugin actually implements the hook
+		hookPath, err := b.findHookFile(p.HooksDir, name)
+		if err != nil {
+			continue
+		}
+
+		// Disallow plugins executing the command or checkout hook more than once
 		if name == "command" || name == "checkout" {
 			if hookTypeSeen[name] {
 				b.shell.Warningf("Ignoring additional %s hook (%s plugin, position %d)", name, p.Plugin.Name(), i+1)
@@ -679,11 +686,6 @@ func (b *Bootstrap) executePluginHook(name string, checkouts []*pluginCheckout) 
 			} else {
 				hookTypeSeen[name] = true
 			}
-		}
-
-		hookPath, err := b.findHookFile(p.HooksDir, name)
-		if err != nil {
-			continue
 		}
 
 		env, _ := p.ConfigurationToEnvironment()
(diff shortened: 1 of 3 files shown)


bootstrap/bootstrap.go
```diff
@@ -672,6 +672,13 @@ func (b *Bootstrap) executePluginHook(name string, checkouts []*pluginCheckout)
 	hookTypeSeen := make(map[string]bool)
 
 	for i, p := range checkouts {
+		// First we verify the plugin actually implements the hook
+		hookPath, err := b.findHookFile(p.HooksDir, name)
+		if err != nil {
```
pda: I think we should check for the `os.ErrNotExist` that `findHookFile` returns, rather than assuming _any_ error is that scenario. That'll stop this same issue from being triggered by some future possible error condition other than the hook file not existing.

toolmantim: Yep, good idea. So perhaps:

```go
		// First we verify the plugin actually implements the hook
		hookPath, err := b.findHookFile(p.HooksDir, name)
		if err == os.ErrNotExist {
			continue
		} else if err != nil {
			return err
		}
```

pda: I was worried about the future compatibility of that; if `findHookFile` wraps the error then that equality check will no longer be true. I think this will do the trick: https://godoc.org/os#IsNotExist

Pushed as d2f3c9056531426553cd13669750533ce313d070

pda: Unfortunately `os.IsNotExist` doesn't actually handle unwrapping:
https://play.golang.org/p/AaGYo-6pFoV

```
	// Note that this function is not errors.Is:
	// underlyingError only unwraps the specific error-wrapping types
	// that it historically did, not all errors.Wrapper implementations.
```
-- https://golang.org/src/os/error.go?s=2847:2878#L82

I guess there's hope that in future it will do unwrapping, though?
And until then, we're only returning that one error type directly, so it'll probably be fine?

Using proper `errors.Is()` unwrapping here would make the agent incompatible with Go prior to v1.13, which I think we'd want to avoid for a bit longer.

pda: I pushed some code comments to keep track of this.

bootstrap/bootstrap.go
```diff
@@ -672,6 +674,19 @@ func (b *Bootstrap) executePluginHook(name string, checkouts []*pluginCheckout)
 	hookTypeSeen := make(map[string]bool)
 
 	for i, p := range checkouts {
+		// First we verify the plugin actually implements the hook
+		hookPath, err := b.findHookFile(p.HooksDir, name)
+		// os.IsNotExist() doesn't unwrap wrapped errors (as at Go 1.13).
+		// agent is still go pre-1.13 compatible (I think) so we're avoiding errors.Is().
```
sj26: I'd say we just bump to latest and use the nice things. But that's separate from this PR.

ahilsend:
Running the docker plugin 2nd is a good test case: 

```yaml
steps:
- command: "echo 'foo' > bar.txt"
  plugins:
    - docker#v3.5.0:
        image: bash
    - artifacts#v1.2.0:
        upload: bar.txt

- command: "cat bar.txt"
  plugins:
    - artifacts#v1.2.0:
        download: bar.txt
    - docker#v3.5.0:
        image: bash
```

pda:
I'm looking into testability.

pda:
I added a more detailed bug summary to this PR description;

> * If any plugin for a step has a checkout hook, only the first plugin gets the opportunity to run its checkout hook;
>     * subsequent plugins show the "Ignoring ..." warning message regardless of whether they implement a checkout hook.
> * But if none of the plugins implement a checkout hook, there are no warning messages.
> * If the first plugin implements a checkout hook, it runs as normal. If it doesn't, nothing happens, as normal.
>     * If subsequent steps implement a checkout hook, they are ignored with a warning.
>     * If subsequent steps _don't_ implement a checkout hook, the warning appears anyway.

pda:
> Is there any way to write a test so this doesn't regress?

Well that was an adventure. I've pushed a test. Here's the result on master / v3.18.0:

```
$ go test ./bootstrap/integration -run TestOnceOnlyHooks
--- FAIL: TestOnceOnlyHooks (0.60s)
    expectation.go:166: Expected [plugin-b "checkout"] to be called at least 1 times, got 0
    mock.go:239: Not all expectations were met (1 out of 2)
    expectation.go:166: Expected [plugin-c "command"] to be called at least 1 times, got 0
    mock.go:239: Not all expectations were met (1 out of 2)
    plugin_integration_test.go:208: Message: 'Ignoring additional checkout hook' wanted count: 1, got: 3
    plugin_integration_test.go:208: Message: 'Ignoring additional command hook' wanted count: 1, got: 3
FAIL
FAIL	github.com/buildkite/agent/v3/bootstrap/integration	0.600s
FAIL
```
(note the "at least 1 times" really means "exactly once"; [the implementation](https://github.com/buildkite/bintest/blob/820c89d3b3a066515bbad3ced13b12fe72c6626a/expectation.go#L165-L175) checks the lower bound and then the upper bound, only reporting the first failure)


Here's the result on this branch:

```
$ go test ./bootstrap/integration -run TestOnceOnlyHooks
ok  	github.com/buildkite/agent/v3/bootstrap/integration	0.594s
```

sj26:
Nice one, @pda!

pda:
Released in https://github.com/buildkite/agent/releases/tag/v3.19.0

PR #1161, opened 2020-02-06
RFC: rollback changes preventing multiple checkout/command plugin hooks

pda:
Summary: This PR rolls back some changes released over the past few months regarding whether multiple checkout and command plugin hooks will run or be skipped. As a result of this PR, if multiple plugins had checkout or command hooks, they would all run. But as always, if any plugin has one of those hook types, the corresponding repo-local / agent-global and bootstrap-builtin hook would not run.

----

Hooks are documented here: https://buildkite.com/docs/agent/v3/hooks#available-hooks

Historically (since the beginning? maybe some adjustments to ordering) the agent has partially avoided running multiple `checkout` and `command` hooks by doing the following;

- If there are _any_ plugins with a `checkout` hook, run all plugin `checkout` hooks,
- else, if there's a global (agent) `checkout` hook, run that,
- else, run the default bootstrap checkout phase.

https://github.com/buildkite/agent/blob/8feefa9869002b7a35b4bb37770f1204a066e21c/bootstrap/bootstrap.go#L889-L902

And,

- If there are _any_ plugins with a `command` hook, run all plugin `command` hooks,
- else, if there's a local repository `command` hook, run that,
- else, if there's a global (agent) `command` hook, run that,
- else, run the default bootstrap command phase.

https://github.com/buildkite/agent/blob/8feefa9869002b7a35b4bb37770f1204a066e21c/bootstrap/bootstrap.go#L1293-L1303

Around May 16, 2018 this was documented as:

> Unlike other hooks, if multiple command hooks are found only the first of Plugin, Pipeline or Agent hook type will be run.
-- https://github.com/buildkite/docs/commit/d6335220cd5f47980394d964cf851f8794962d60#diff-3d3907dc1c723c2368dd7c612733900fR29

And on May 22, 2018 is was updated to:

> If multiple checkout hooks are found, only the first will be run.
> If multiple command hooks are found, only the first will be run.
-- https://github.com/buildkite/docs/commit/0aea7830d3f9aa416a583df2aa7cdb5ec2dc291c#diff-3d3907dc1c723c2368dd7c612733900fR26-R29

Neither of these quite captured the fact that jobs with multiple plugins may execute multiple checkout and command steps respectively.

It could definitely be argued that if plugin/local/global/bootstrap are mutually exclusive for checkout and command, then it's inconsistent that multiple plugin hooks would run. But it's also how it's worked for many years.

---

This discrepancy was then pointed out in an issue and subsequent pull request, which led to some changes, bugs and head scratching:

#1052 (Jul 23, 2019)
An issue noting that [our docs](https://buildkite.com/docs/agent/v3/hooks) say if only the first of multiple command hooks will run, but actually multiple plugin command hooks will run.

#1055 (merged Sep 17, 2019, shipped in agent v3.15.0 on Sep 25, 2019)
A patch to prevent multiple plugin command hooks from running, to match a certain interpretation of the docs.
After it shipped, several people reported in the #1055 comment thread that the change in behavior had broken their builds.

#1135 (merged Dec 12, 2019, shipped in agent v3.18.0 on Jan 29, 2020)
A patch to also prevent multiple plugin checkout hooks (as per docs), and also add warnings when subsequent command or checkout hooks are being skipped.
This contained a bug which resulting in many hooks being skipped.

#1156 (merged Jan 30, 2019, shipped in agent v3.19.0 on Jan 30, 2019).
A patch to fix the bug introduced above.
Also introduces an integration test for multiple plugins with the same hooks as each other.

---

Current state is:

- Since agent v3.15.0 on Sep 25, 2019 only one plugin command hook runs, the rest are silently skipped.
- Since agent v3.18.0 on Jan 29, 2020 only one plugin checkout hook runs, skipped steps show warnings.
- These changes broke some builds set up between 2015 and 2019.
- Changing them back (merging this PR) may break some builds set up in the past four months.

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -668,32 +668,16 @@ func (b *Bootstrap) VendoredPluginPhase() error {
 
 // Executes a named hook on plugins that have it
 func (b *Bootstrap) executePluginHook(name string, checkouts []*pluginCheckout) error {
-	// Command and checkout hooks are a little different, in that we only execute
-	// the first one we see. We run the first one, and output a warning for all
-	// the subsequent ones.
-	hookTypeSeen := make(map[string]bool)
-
-	for i, p := range checkouts {
-		// First we verify the plugin actually implements the hook
+	for _, p := range checkouts {
 		hookPath, err := b.findHookFile(p.HooksDir, name)
 		// os.IsNotExist() doesn't unwrap wrapped errors (as at Go 1.13).
 		// agent is still go pre-1.13 compatible (I think) so we're avoiding errors.Is().
 		// In future somebody should check if os.IsNotExist() has added support for
 		// error unwrapping, or change this code to errors.Is(err, os.ErrNotExist)
 		if os.IsNotExist(err) {
-			continue
+			continue // this plugin does not implement this hook
 		} else if err != nil {
 			return err
-		}
-
-		// Disallow plugins executing the command or checkout hook more than once
-		if name == "command" || name == "checkout" {
-			if hookTypeSeen[name] {
-				b.shell.Warningf("Ignoring additional %s hook (%s plugin, position %d)", name, p.Plugin.Name(), i+1)
-				continue
-			} else {
-				hookTypeSeen[name] = true
-			}
 		}
 
 		env, _ := p.ConfigurationToEnvironment()
--- a/bootstrap/integration/plugin_integration_test.go
+++ b/bootstrap/integration/plugin_integration_test.go
@@ -144,7 +144,11 @@ func TestMalformedPluginNamesDontCrashBootstrap(t *testing.T) {
 	tester.CheckMocks(t)
 }
 
-func TestOnceOnlyHooks(t *testing.T) {
+// A job may have multiple plugins that provide multiple hooks of a given type.
+// For a while (late 2019 / early 2020) we disallowed duplicate checkout and
+// command hooks from plugins; only the first would execute.  We since decided
+// to roll that back and permit e.g. multiple checkout plugin hooks.
+func TestOverlappingPluginHooks(t *testing.T) {
 	t.Parallel()
 
 	tester, err := NewBootstrapTester()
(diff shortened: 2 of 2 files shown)


pda:
FYI ping to @philwo and to @tduffield @ecv @elruwen @elliott-davis as you all ran into problems when the skip-multiple-plugin-hooks change was introduced.

vanstee:
I work with @elliott-davis. This would let us back out some recent changes we've had to make to an internal plugin that we use in combination with the docker-compose plugin to conditionally include a command hook. Definitely a +1 from our team.

philwo:
I don’t mind either way - if it helps others, I’m fine with rolling back to the old behavior of running multiple command hooks. It would just be good if the documentation matched the behavior. :)

I’m trusting the Buildkite people to do the right thing here. What they built has always perfectly made sense to me, so I’m sure we can make it work for us.

tduffield:
So despite being hit by this, I stand by the Buildkite team's decision to remove this functionality. From my recollection there was a clear reason that they made this decision, and I trust that they did it fully knowing the impact it would have.

lox:
@tduffield we made it because we thought it was undefined behaviour that nobody was relying on. Turns out it's super common to have multiple instances of plugins that execute things. I think that we need to revert this as it is a breaking change in retrospect.

In future versions we should figure out ways to let plugins opt-in to overriding the default checkout vs executing something themselves.

lox:
FWIW @tduffield, @pda is a new agent maintainer and generally a better programmer than I am. Hopefully he'll help us get back some of the momentum we had on Agent changes!

elruwen:
I am still in favour of allowing multiple plugins to hook into the lifecycle wherever they want. This allows reuse and composition. I know I can just have multiple steps, but then I need to carry the local state across. For compile-languages, I either have to compile everything again or pass around the compiled artifacts, which again slows down the build.

pda:
Thanks @vanstee @philwo @tduffield @elruwen - it's great to have your input.

In trying to round up all the context of what led to this, I forgot to also give the context that I'm working at Buildkite and having some back-channel conversations with @lox @toolmantim etc :)

I'll merge this and work on getting a release out today / tomorrow.

elruwen:
Amazing! Thanks! Luckily I didn't start to migrate all of my pipelines ;)

tduffield:
Awesome, welcome @pda! I have to applaud you on your wonderful PR description. Thanks!

ecv:
Late to the party but excited to see a new maintainer w/ the same high-quality thought going into the product. We adapted our plugins to dodge the 3.15 change, they're all pre- or post-command hooks now; and we should require no work to run on 3.20. 🎉

PR #1165, opened 2020-02-11
Alter debian package build step to work on agents without ruby

yob:
This step requires an environment with ruby (to use the fpm gem). Historically we've used the "queue=deploy" targeting to ensure it runs on an agent that happens to have ruby.

However, we'd like to be able to run this step on a standard elastic stack environment, which only has the bk agent binary, awscli, and docker.

This changes the step to launch a docker container with ruby for
the parts that need it.

I considered using the docker-compose plugin for the whole step like other parts of the pipeline, however there's a reasonable amount of existing shell code used here and I wasn't sure how it would interact with the plugin. In the end, I optimised for the least disruption to what was already there.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -233,8 +233,6 @@ steps:
       - set-metadata
     command: ".buildkite/steps/build-debian-packages.sh"
     artifact_paths: "deb/**/*"
-    agents:
-      queue: "deploy"
 
   - name: ":redhat: build"
     key: build-rpm-packages
--- a/.buildkite/steps/build-debian-packages.sh
+++ b/.buildkite/steps/build-debian-packages.sh
@@ -19,9 +19,6 @@ dry_run() {
   fi
 }
 
-echo "--- Installing dependencies"
-bundle
-
 # Make sure we have a clean deb folder
 rm -rf deb
 
@@ -39,5 +36,5 @@ for ARCH in "amd64" "386" "arm" "armhf" "arm64"; do
   chmod +x "$BINARY"
 
   # Build the debian package using the architectre and binary, they are saved to deb/
-  ./scripts/build-debian-package.sh "$ARCH" "$BINARY" "$AGENT_VERSION" "$BUILD_VERSION"
+  .buildkite/steps/ruby-env ./scripts/build-debian-package.sh "$ARCH" "$BINARY" "$AGENT_VERSION" "$BUILD_VERSION"
 done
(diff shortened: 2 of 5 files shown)


yob:
Once we agree on the how to to achieve this change, I'll open a follow up PR that  makes the same change to the rpm building step

.buildkite/steps/build-debian-packages.sh
```diff
@@ -39,5 +36,5 @@ for ARCH in "amd64" "386" "arm" "armhf" "arm64"; do
   chmod +x "$BINARY"
 
   # Build the debian package using the architectre and binary, they are saved to deb/
-  ./scripts/build-debian-package.sh "$ARCH" "$BINARY" "$AGENT_VERSION" "$BUILD_VERSION"
+  .buildkite/steps/ruby-env ./scripts/build-debian-package.sh "$ARCH" "$BINARY" "$AGENT_VERSION" "$BUILD_VERSION"
```
yob: leverage the new `.buildkite/steps/ruby-env` (also added in this PR) to launch a process inside a container that has ruby. 

This script will come in handy as other steps in the build (like building the rpm) are also converted to be runnable on elastic stacks

sj26:
This is a nice idea!

It looks like ruby-env runs a command inside a docker container from the docker compose. Could we use the docker-compose plugin instead? a la:

```yaml
    command: .buildkite/steps/build-debian-packages.sh
    plugins:
      docker-compose#v3.0.0:
        config: .buildkite/docker-compose.yml
        run: ruby
```

.buildkite/pipeline.yml
```diff
@@ -233,8 +233,6 @@ steps:
       - set-metadata
     command: ".buildkite/steps/build-debian-packages.sh"
     artifact_paths: "deb/**/*"
-    agents:
-      queue: "deploy"
```
pda: Do we not need to target some other agent rule?

yob:
I considered using the docker-compose plugin for the whole step like other parts of the pipeline, however there's a reasonable amount of existing shell code used here and I wasn't sure how it would interact with the plugin. In the end, I optimised for the least disruption to what was already there.

sj26:
Sorry @yob, I didn't read the description carefully enough 😅

The plugin pretty much just does what you've scripted, I don't think there would be more disruption.

sj26:
But, if you're more comfortable shipping this as is, then please do! We can consider compose later.

yob:
would I have to do any shenanigans to get the artifact and metadata downloads working from inside the container (all the stuff in `.buildkite/steps/build-debian-packages.sh`)?

sj26:
Oh, right. Hmm.

In the past we've done something like
mount:
- /usr/bin/buildkite-agent:/usr/bin/buildkite-agent

yob:
Ok, given the extra complexity I might go with this:

> But, if you're more comfortable shipping this as is, then please do! We can consider compose later.

.buildkite/pipeline.yml
yob: the elastic-runners stack processes the default queue

PR #1169, opened 2020-02-12
Lock our windows builds to a specific image digest

yob:
Windows builds of the agent have started failing, and all the failed jobs use the following digest of the windows/servercore image:

    sha256:47bb7fe42f88823f3e54c03cefa2f20e3d7cff8a1a051878af2f45f6b6afef65

The last green windows build used this digest:

    sha256:23e7bb473261f12307e82d0cd52a1bcbad4bc64c00e889cee344401d0c3de3d3

According to [docker hub](https://hub.docker.com/_/microsoft-windows-servercore), microsoft updated this image at:

    02/11/2020 18:12:05

The timezone of that stamp is unclear but it's with 24 hours of when our builds started failing. Maybe locking us back to the older digest will help?

### DIFF
--- a/.buildkite/Dockerfile-windows
+++ b/.buildkite/Dockerfile-windows
@@ -1,5 +1,5 @@
 ## Development docker image for buildkite-agent on windows
-FROM mcr.microsoft.com/windows/servercore:1809
+FROM mcr.microsoft.com/windows/servercore:1809@sha256:23e7bb473261f12307e82d0cd52a1bcbad4bc64c00e889cee344401d0c3de3d3
 
 ENV ChocolateyUseWindowsCompression false
 


pda:
Nice. Perhaps this will also give us a warm image cache more often? Actually probably not.

Is there a risk that we now drift behind on security issues etc? Should we unpin it and see if a newer image is fixed?

yob:
> Is there a risk that we now drift behind on security issues etc? Should we unpin it and see if a newer image is fixed?

[dfresh](https://github.com/realestate-com-au/dfresh) suggests there is a newer version of the tag:

```
$ dfresh check .buildkite/Dockerfile-windows 
.buildkite/Dockerfile-windows:2: mcr.microsoft.com/windows/servercore:1809
  old sha256:23e7bb473261f12307e82d0cd52a1bcbad4bc64c00e889cee344401d0c3de3d3
  new sha256:a14c8e7a925f09bdd6eb5cc42e798349c4345be08d89514383e9731a900964e1
```

However, I just noticed this on [docker hub](https://hub.docker.com/_/microsoft-windows-servercore):

> Important: You might encounter issues when using Windows Server containers with the February 11, 2020 security update release
>
> You might encounter issues using the Windows Server Container Image for the February 11, 2020 security update when used on a Windows Server container host that is not updated to the same security update. For further information and guidance, visit the Microsoft Support Article for this issue.

The [support article](https://support.microsoft.com/en-us/help/4542617/you-might-encounter-issues-when-using-windows-server-containers-with-t) says:

> You might encounter issues using Windows Server containers, unless both the Windows container host and Windows Server containers are matched with the February 11, 2020 security update. Compatibility was affected by a security change that must be addressed in both the container host and the container image, requiring them to match. 

I haven't looked into this further, but the timing is suspicious.

Do the hosts running our windows agents have auto-updates for security patches enabled?

PR #1170, opened 2020-02-14
run "publish to s3" steps on the new deploy-production agents

yob:
The "deploy" queue currently runs on the bastion, and we're trying to move things off that so it can be shut down.

publishing to s3 requires the buildkite-agent (to fetch metadata) and awscli (to push to s3), both of which are available on the elastic stack agents that process the "deploy-production" queue.

There's a good chance the agents won't have IAM permissions for pushing to the bucket, but we can address that if it happens.

### DIFF
--- a/.buildkite/pipeline.release-experimental.yml
+++ b/.buildkite/pipeline.release-experimental.yml
@@ -6,7 +6,7 @@ steps:
     env:
       CODENAME: "experimental"
     agents:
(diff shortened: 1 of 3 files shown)


yob:
This is difficult to test ahead of time, so I'm going to just merge and see how it goes.

I've set the release pipelines to `DRY_RUN: true`, so nothing will actually be deployed.

PR #1177, opened 2020-02-26
Migrate publishing of deb package to the new "deploy" queue

yob:
The `deploy-legacy` queue is deprecated, and we're slowly moving steps off it.

The new `deploy` queue is processed by agents in an elastic stack, which has a few implications:

* most of the tools we need (like gnupg and ruby) aren't available by default, so we have to use a docker container with the tools we need
* sensitive values (like gpg keys) aren't stored on the new agents, so we need to fetch them from AWS SSM

Some bonus highlights:

* I pushed the `dry_run` logic to later in the scripts, to after the point where we configure gnupg. This lets us verify the gnupg setup in dry runs
* The gnupg keys are never written to disk - they only exist in a docker tmpfs volume

### DIFF
--- a/.buildkite/pipeline.release-stable.yml
+++ b/.buildkite/pipeline.release-stable.yml
@@ -38,8 +38,18 @@ steps:
     command: ".buildkite/steps/publish-debian-package.sh"
     env:
       CODENAME: "stable"
+      DEB_S3_BUCKET: "apt.buildkite.com/buildkite-agent"
     agents:
-      queue: "deploy-legacy"
+      queue: "deploy"
+    plugins:
+      - ecr#v2.0.0:
+          login: true
+          account-ids: "032379705303"
+      - docker#v3.5.0:
+          image: "032379705303.dkr.ecr.us-east-1.amazonaws.com/deploytools:2020.03"
+          propagate-environment: true
+          tmpfs:
+            - "/root/.gnupg"
 
   - name: ":docker:"
     command: ".buildkite/steps/publish-docker-images.sh"
--- a/.buildkite/steps/publish-debian-package.sh
+++ b/.buildkite/steps/publish-debian-package.sh
@@ -1,12 +1,35 @@
 #!/bin/bash
-set -e
+set -euo pipefail
 
 artifacts_build=$(buildkite-agent meta-data get "agent-artifacts-build" )
+
+# /root/,gnupg is a tmpfs volume, so we can safely store key data there and know
+# it won't be written to a disk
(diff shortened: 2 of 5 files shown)


.buildkite/pipeline.release-stable.yml
```diff
@@ -31,8 +31,18 @@ steps:
...
-      queue: "deploy-legacy"
+      queue: "deploy"
+    plugins:
+      - ecr#v2.0.0:
+          login: true
+          account-ids: "032379705303"
+      - docker#v3.5.0:
+          image: "032379705303.dkr.ecr.us-east-1.amazonaws.com/deploytools:2020.03"
+          propagate-environment: true
+          tmpfs:
```
pda: :+1: 

I wonder if there's any way to validate that the GPG dir is not persisting to disk prior to downloading the secret key. Otherwise it feels like a slightly fragile security model, with this tmpfs specification so far away from the key being placed there. 

```
pda@x1pda ~ $ docker run --rm -it --tmpfs /hello archlinux mount | grep hello
tmpfs on /hello type tmpfs (rw,nosuid,nodev,noexec,relatime)
```

.buildkite/steps/publish-debian-package.sh
```diff
@@ -1,21 +1,36 @@
...
-  if [[ "${DRY_RUN:-}" == "false" ]] ; then
-    "$@"
-  else
-    echo "[dry-run] $*"
-  fi
-}
+# /root/,gnupg is a tmpfs volume, so we can safely store key data there and know
+# it won't be written to a disk
+secret_key_path="/root/.gnupg/gpg-secret.gpg"
+public_key_path="/root/.gnupg/gpg-public.gpg"
```
pda: Maybe we could assert tmpfs here so it can't accidentally write to disk if the docker invocation changes?

yob: Do you have a recommended approach for the assertion? Check the results of `mount | grep gnupg`?

pda: As we discovered while pairing:

```shell
apk add --update findmnt

findmnt --source tmpfs --target /root/.gnupg
```

Which will show a nice summary on success, and exit non-zero on failure.

```
TARGET SOURCE FSTYPE OPTIONS
/hello tmpfs  tmpfs  rw,nosuid,nodev,noexec,relatime
```

yob:
I'm getting a 500 when trying to merge this one :thinking:

PR #1178, opened 2020-02-26
use deploytools to publish to s3

yob:
In PR #1170 (and then c6de240) I jumped through some hoops to get the publishing to S3 to work without the "deploy-legacy" queue.

It worked, but the result was pretty messy.

This changes that step to launch within a docker container with all the tools we need, so the scripts can assume the tools are on the path.


yob:
This seems to work well, so I'll merge

PR #1180, opened 2020-02-26
migrate publishing of rpm packages to the new "deploy" queue

yob:
The "deploy-legacy" queue is deprecated, and we're slowly moving steps off it.

The new "deploy" queue is processed by agents in an elastic stack, which means most of the tools we need (like createrepo_c) aren't available by default. We can use a docker container with the tools we need.

Some highlights to call out here:

1. createrepo_c is available in the alpine package repositories (lucky!), so we install it each time
2. createrepo_c requires a full local copy of the rpm repository. The elastic stack agents processing the "deploy" queue don't have persistent local disk, so each time this step runs we have to re-download the full archive. It's currently ~9Gb and takes ~90 seconds to download. It's inefficient, but the time is probably tolerable given this doesn't run *that* often
3. Downloading the full archive for (2) outputs a lot to stdout, so I added the --only-show-errors switch
4. The existing script was running bundler to download gems, however we weren't using them
5. The agents processing the "deploy-legacy" queue have createrepo_c v0.10, but the new agents will have v0.15.6

### DIFF
--- a/.buildkite/steps/publish-rpm-package.sh
+++ b/.buildkite/steps/publish-rpm-package.sh
@@ -27,7 +27,7 @@ echo '--- Installing dependencies'
 apk --no-cache add createrepo_c --repository=http://dl-cdn.alpinelinux.org/alpine/edge/testing
 
 # createrepo_c requires some exotic flags on the cp, which aren't available on the busybox version
-apk add coreutils
+apk --no-cache add coreutils
 
 # Make sure we have a local copy of the yum repo
 echo "--- Syncing s3://$RPM_S3_BUCKET to `hostname`"
--- a/.buildkite/steps/publish-rpm-package.sh
+++ b/.buildkite/steps/publish-rpm-package.sh
@@ -24,12 +24,15 @@ mkdir -p rpm
 buildkite-agent artifact download --build "$artifacts_build" "rpm/*.rpm" rpm/
 
 echo '--- Installing dependencies'
-bundle
+apk --no-cache add createrepo_c --repository=http://dl-cdn.alpinelinux.org/alpine/edge/testing
+
+# createrepo_c requires some exotic flags on the cp, which aren't available on the busybox version
+apk add coreutils
 
 # Make sure we have a local copy of the yum repo
 echo "--- Syncing s3://$RPM_S3_BUCKET to `hostname`"
 mkdir -p $YUM_PATH
-aws --region us-east-1 s3 sync --delete "s3://$RPM_S3_BUCKET" "$YUM_PATH"
+aws --region us-east-1 s3 sync --delete --only-show-errors "s3://$RPM_S3_BUCKET" "$YUM_PATH"
 
 # Add the rpms and update meta-data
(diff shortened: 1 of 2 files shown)


pda:
We could look into attaching an EFS volume (basically an NFS file share) to these agents, if we wanted a more local cache. It's still pulling data over the network, but maybe in a faster/cheaper way than from S3?

.buildkite/steps/publish-rpm-package.sh
```diff
@@ -24,12 +24,15 @@ mkdir -p rpm
...
+apk --no-cache add createrepo_c --repository=http://dl-cdn.alpinelinux.org/alpine/edge/testing
+
+# createrepo_c requires some exotic flags on the cp, which aren't available on the busybox version
+apk add coreutils
 
 # Make sure we have a local copy of the yum repo
 echo "--- Syncing s3://$RPM_S3_BUCKET to `hostname`"
 mkdir -p $YUM_PATH
-aws --region us-east-1 s3 sync --delete "s3://$RPM_S3_BUCKET" "$YUM_PATH"
+aws --region us-east-1 s3 sync --delete --only-show-errors "s3://$RPM_S3_BUCKET" "$YUM_PATH"
```
pda: `--only-show-errors` - nice :+1:

.buildkite/steps/publish-rpm-package.sh
```diff
@@ -24,12 +24,15 @@ mkdir -p rpm
 buildkite-agent artifact download --build "$artifacts_build" "rpm/*.rpm" rpm/
 
 echo '--- Installing dependencies'
-bundle
+apk --no-cache add createrepo_c --repository=http://dl-cdn.alpinelinux.org/alpine/edge/testing
+
+# createrepo_c requires some exotic flags on the cp, which aren't available on the busybox version
+apk add coreutils
```
pda: Is it safe to assume we have a working APK index here? i.e. `apk update` or whatever has already run?

yob:
> We could look into attaching an EFS volume (basically an NFS file share) to these agents, if we wanted a more local cache. It's still pulling data over the network, but maybe in a faster/cheaper way than from S3?

That could be worthwile. Should we consider it an optimisation that can come later if we need it?

.buildkite/steps/publish-rpm-package.sh
yob: I can it it works in the dry run builds, but I guess that could change at some point. Docker isn't as consistent as I'd like it to be.

Do you think it's worth running `apk update` as a precaution?

yob:
> Looks good to my eyeballs. I haven't dug deeper or run the code

It's super hard to be 100% confident this will work. The impact of it not working is low (customers shouldn't be impacted), so I think we'll have to build our confidence to a certain level and then merge and see what happens.

There's always revert.

.buildkite/steps/publish-rpm-package.sh
pda: Probably worth adding, in case the `apk update` in the `docker build` is optimised away in future.

PR #1182, opened 2020-02-27
No fake exits

ticky:
follows on from #899 to add a feature where the golang implementation detail-derived `-1` exit status is never passed to Buildkite, meaning that it will henceforth only be returned from Buildkite itself, and thus have a single meaning

please see EXPERIMENTS.md where I’ve put together a description of what this aims to achieve

### DIFF
--- a/EXPERIMENTS.md
+++ b/EXPERIMENTS.md
@@ -18,6 +18,12 @@ If an experiment doesn't exist, no error will be raised.
 
 ## Available Experiments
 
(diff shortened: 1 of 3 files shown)


chloeruka:
@ticky what's left to get this into a ready for review state?

ticky:
Backend support, unfortunately; Buildkite/Buildkite#4802

ticky:
It'd still be a great idea for this to be resurrected!

PR #1197, opened 2020-03-12
bump Go to 1.14 for tests and compilation

yob:
1.14 was released a few weeks ago, let's see if the agent pipelines likes it

### DIFF
--- a/.buildkite/docker-compose.yml
+++ b/.buildkite/docker-compose.yml
@@ -2,7 +2,7 @@ version: '3.5'
 
 services:
   agent:
-    image: golang:1.13
+    image: golang:1.14.0-buster
     volumes:
       - ../:/work:cached
     working_dir: /work
--- a/.buildkite/docker-compose.yml
+++ b/.buildkite/docker-compose.yml
@@ -2,7 +2,7 @@ version: '3.5'
 
 services:
   agent:
-    image: golang:1.14.0-buster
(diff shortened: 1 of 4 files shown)


.buildkite/docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: '3.5'
 
 services:
   agent:
-    image: golang:1.13
+    image: golang:1.14.0-buster
```
pda: What's the thinking around pinning to buster? Just trying to be specific/deterministic?
Also I'd be pretty happy to stick with auto-patch-bumps via the `1.14` tags.
What do you think?

yob: Yep. Nothing special about buster, but I lean towards determinism and `1.14` currently points to the same image as `1.14.0-buster`

yob: I guess we could stick with `1.14`, but add the digest? At least then tools like renovate bit can help us get patch updates semi-automatically?

pda: Sounds good 👍🏼

yob: dfresh from REA is *super* useful:

```
$ dfresh resolve golang:1.14
golang:1.14@sha256:d27017d27f9c9a58b361aa36126a29587ffd3b1b274af0d583fe4954365a4a59
```

PR #1206, opened 2020-04-08
Fix PowerShell hook support

nickgw:
Thanks to @pdemirdjian for adding the catches for PowerShell hooks, but with all the quoting and escaping a couple quotes got missed.

This fixes the block and reverts to using `"` for quoting and `\` for esecaping.


keithduncan:
Thanks for reporting this @nickgw and I’m so sorry this has taken this long to get to. I should have this fixed with #1497 and have added Windows PowerShell hooks to my testing matrix done on releases. I’d also like to come back and add some unit tests for these scriptwrappers on the respective hosts 🤔  This fix will release as part of Buildkite Agent v3.32.3 and Elastic CI Stack for AWS v5.6.0 :bow:

PR #1207, opened 2020-04-17
Thread GitCloneMirrorFlags through from env vars

KevinGrandon:
We would like to be able to specify custom mirror flags as environment variables. This currently works for the other fetch and clone flags, but does not yet work for mirror flags.


KevinGrandon:
Note: I haven't tested this locally yet, and this is just a guess as this was the only config missing the env: statement.

DrJosh9000:
This was completed in #2056.

PR #1210, opened 2020-05-01
Add a retry for errors during artifact search

lox:
This will have the agent retry 502 errors during artifact search, which should make it more resilient to upstream faults.

### DIFF
--- a/agent/artifact_searcher.go
+++ b/agent/artifact_searcher.go
@@ -1,8 +1,11 @@
 package agent
 
 import (
+	"time"
+
 	"github.com/buildkite/agent/v3/api"
 	"github.com/buildkite/agent/v3/logger"
+	"github.com/buildkite/agent/v3/retry"
 )
 
 type ArtifactSearcher struct {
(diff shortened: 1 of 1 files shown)


jerrymarino:
I've been noticing an uptick in 502s on buildkite's search API over the last couple weeks e.g.:
```
2020-05-04 11:35:17 FATAL  Failed to download artifacts: GET https://agent.buildkite.com/v3/builds/407d5881-6ac6-45a3-a4f5-909a0e409e33/artifacts/search?query=%2A.dSYM-37170.zip: 502
```

I think this will improve things this situation a bit, kudos for this PR 👍 !

That being said, it'd be great if the agent was more resilient to buildkite.com errors. Perhaps baking in retries for all requests to buildkite.com at the `apiClient` could be a longer term solution?

lox:
> That being said, it'd be great if the agent was more resilient to buildkite.com errors. Perhaps baking in retries for all requests to buildkite.com at the apiClient could be a more global solution?

Yeah, I considered that when implementing @jerrymarino, definitely something we will consider.

PR #1211, opened 2020-05-01
Normalise upload path to Unix/URI form on Windows

### LINKED ISSUE #354: Normalise upload/download paths with Windows

### LINKED ISSUE #1159: Buildkite Agent v3.19.0 fails to download files from GCS with backslashes in the name?
Hi,

I'm debugging a mysterious pipeline failure that seems to occur since we upgraded our Buildkite Agents from v3.14.0 to v3.19.0.

We have some build artifacts that are in a sub-folder on Windows, e.g. `bazel-bin\bin\bazel.exe`. When Buildkite Agent uploads them, they get stored like this on GCS: https://storage.googleapis.com/bazel-trusted-buildkite-artifacts/148f94cc-c9bd-4600-b73c-7b41d5bdc8e6/bazel-bin%5Csrc%5Cbazel.exe (the backslashes remain and aren't converted into forward slashes).

Screenshot 2020-02-04 at 21 23 40

This is also visible in the GCS console:

Screenshot 2020-02-04 at 21 33 56

Since v3.19.0, Buildkite Agent fails to download these artifacts, at least when it's running on Linux (haven't tested what happens when the downloading agent runs on Windows).

With v3.14.0:
Screenshot 2020-02-04 at 21 19 13

With v3.19.0:
Screenshot 2020-02-04 at 21 18 56

The issue seems to be that the agent now converts the backslashes to forward slashes when it builds the URL for the download. This is the one it tries to download from:

https://www.googleapis.com/storage/v1/b/bazel-trusted-buildkite-artifacts/o/edf99f72-5065-4b49-815e-afb715d8fad3%2Fbazel-bin%2Fsrc%2Fbazel.exe?alt=media

This would be the correct one (note that %2F vs. %5C):

https://www.googleapis.com/storage/v1/b/bazel-trusted-buildkite-artifacts/o/edf99f72-5065-4b49-815e-afb715d8fad3%2Fbazel-bin%5Csrc%5Cbazel.exe?alt=media

Our pipeline is now failing. I can see a workaround here, of course - we can upload the binaries without the path in their name, but this seems like a bug? 🤔

ticky:
This is in an effort to address failures to download on Google Cloud due to the mix of path separators being unsupported.

I have updated the related specs to always expect Unix/URI style paths for the relative file path (Path field) on all platforms.

Given that the uploaders all use the `AbsolutePath`, which is left unchanged here, this should allow uploads to continue to function, but with more correct URLs on upload. Currently, this will also change the path shown to the user in the console output, API and Buildkite UI, which may or may not be desirable. Further discussion of the implications here is likely needed, and in either case this may be considered a breaking change!

This is now behind an experiment flag, as it would be a breaking change for some people, but likely will become the default in the next major release.

potentially fixes #354, fixes #1159, follows #638 (which did almost exactly the same thing!) and #1113.

See also [this post in the forum](https://forum.buildkite.community/t/enhancement-artifact-uploads-into-gcs-from-windows-use-backslashes-in-their-keys/390/6)


ticky:
Given the change in paths used for display, perhaps this change requires a new normalised-path field, which is used for computing the upload URL, but not the URL shown during upload or in the UI. I think this would require the API to accept this field, or some way to filter it out, because the `api.Agent` struct used to pass around this information is built around that purpose. Alternately, we could shift the path normalisation to the steps which compute the URL, but that requires changes to every upload strategy we implement, whereas this nips it in the bud early.

Happy to discuss, though!

ticky:
Additionally, should this bit of Artifact Downloader logic also do the opposite, and be using [filepath.FromSlash / filepath.ToSlash](https://golang.org/pkg/path/filepath/#FromSlash)?
https://github.com/buildkite/agent/blob/57f90454e6689e7a5c18e995b893c7f606a9ba7b/agent/artifact_downloader.go#L93-L97

petemounce:
This looks good to me. The behaviour that causes difficulty for us is uploading from Windows to then download on not-Windows. It makes sense to me to defend against that at the edge, which here is the upload and download (your follow up comment for on Windows). cc @jamiebrynes7

sj26:
This looks really good, and I think is a great direction.

I'm going to remove the review request because it looks like @pda is assigned.

jayco:
This is awesome @ticky 👌

PR #1216, opened 2020-05-13
Add in-process synchronisation to git-mirror-related actions

denbeigh2000:
This change adds a simple in-process lock on top of the existing lockfile dependency and uses it when locking git mirror repositories for cloning actions.

With this, the buildkite-agent should be able to use mirrored repositories when running >1 agents/process; right now it cannot dependably do this, because the lockfile is only isolated by PID, leading to one git action to sometimes fail due to not being able to acquire `.git/index.lock`

### DIFF
--- a/bootstrap/shell/shell.go
+++ b/bootstrap/shell/shell.go
@@ -16,10 +16,10 @@ import (
 	"time"
 
 	"github.com/buildkite/agent/v3/env"
+	"github.com/buildkite/agent/v3/lockfile"
 	"github.com/buildkite/agent/v3/logger"
 	"github.com/buildkite/agent/v3/process"
 	"github.com/buildkite/shellwords"
-	"github.com/nightlyone/lockfile"
 	"github.com/pkg/errors"
 )
 
@@ -186,7 +186,7 @@ func (s *Shell) LockFile(path string, timeout time.Duration) (LockFile, error) {
 		}
 	}
 
-	return &lock, err
+	return lock, err
 }
 
 // Run runs a command, write stdout and stderr to the logger and return an error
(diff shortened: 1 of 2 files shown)


keithpitt:
@denbeigh2000 thanks for the PR! I'll take a look with @lox and @pda

aashishkapur:
@keithpitt Wondering if you had any updates on this PR

lox:
Thanks for this @denbeigh2000, sorry a reply has taken a while. 

I'm confused, shouldn't there be a different pid file for each invocation of `buildkite-agent bootstrap`? 

I totally agree we need a lock around git actions that need a lock on `.git/index.lock`, but I'm not sure that this will do it!

denbeigh2000:
Hi @lox,

This aims to handle the opposite case, running multiple workers with `buildkite-agent --spawn N`.

As far as I can tell, when this is set the agent runs multiple workers in a pool within the same process, and there is no protection when grabbing the file-lock in this case.

lox:
@denbeigh2000 the architecture of the agent here that is relevant is that with `--spawn` there is a single process for the [Agent Pool](https://github.com/buildkite/agent/blob/master/agent/agent_pool.go) and the [Agent Workers](https://github.com/buildkite/agent/blob/master/agent/agent_worker.go) (1 per agent spawned), but each job still spawns a new sub-process of `buildkite-agent bootstrap` to process the job. All the git interactions happen in the `bootstrap` process, so should have unique pid's! 

I think the problem here is that a lock is needed across various git operations vs a lock around checkout and a different lock around mirror update.

denbeigh2000:
Ah yeah it seems I've gone barking up the wrong tree there, I somehow missed where the JobRunner forks off into its' own process - I'll close this PR.

We're very interested in making better use of  this feature someday!

lox:
Sorry about that @denbeigh2000! We're keen on fixing it too. We'll brain on what locks are needed to prevent the errors you are seeing. As always, we deeply appreciate the PR and your time and will try to give feedback quicker next time!

PR #1218, opened 2020-05-26
Add sidecar Dockerfile and build steps

keithduncan:
This replaces https://github.com/buildkite/buildkite-sidecar and the keithduncan/buildkite-sidecar Docker Hub repository by tagging `buildkite/agent:X-sidecar` directly at the source.

Something I haven’t yet included is the [environment hook](https://github.com/buildkite/buildkite-sidecar/blob/master/agent/hooks/environment) that I used to print `BUILDKITE_ABOUT_*` variables in the build log. I’ve found that useful for my builds but I’m not sure whether to include for all sidecars?

cc @chloeruka @pda

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -286,5 +286,6 @@ steps:
       - build-docker-alpine
       - build-docker-ubuntu
       - build-docker-centos
+      - build-docker-sidecar
       - build-github-release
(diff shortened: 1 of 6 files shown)


yob:
Hmm, this has broken our release pipeline: https://buildkite.com/buildkite/agent-release-edge/builds/294#8d50db3e-781c-4690-9378-f1ddf83988be

I'll look into why

PR #1219, opened 2020-05-27
Provide example dogstatsd integration options

GaryPWhite:
Noticed as I was filling out my config file for Datadog that there descriptions with most buildkite-agent options.

Since this might be an option few know about or some people may need to use, took a swing at filling in the blanks here, and provided a link to read on.

Let me know what you think!


pda:
Thanks @GaryPWhite — looks like a good addition to the sample config / docs.

PR #1220, opened 2020-05-28
Accept `--experiment` flags in all buildkite-agent subcommands

ticky:
While the upload path normalisation logic added in #1211 worked in unit tests, it turns out the subcommand wasn't actually accepting the experiment flag, only `buildkite-agent start` was.

This updates all other subcommands to read the experiment values, even if they don't themselves currently implement an experiment. This permits experiments such as that added in #1211 to exist within in-job commands such as `artifact upload`.

Given that there currently are no end-to-end tests of the subcommand mechanism, should some be added? What shape would that take?

Tested by looking at the output of this (fake, expected to error out) command:

```bash
BUILDKITE_AGENT_ACCESS_TOKEN=fake \
BUILDKITE_JOB_ID=fake-uuid \
BUILDKITE_AGENT_EXPERIMENT=nomalise-upload-paths \
  buildkite-agent artifact upload \
    --debug ./install.sh
```

Output should contain "DEBUG  Enabled experiment `nomalise-upload-paths`,” but without this patch it does not.

### DIFF
--- a/clicommand/annotate.go
+++ b/clicommand/annotate.go
@@ -55,9 +55,10 @@ type AnnotateConfig struct {
 	Job     string `cli:"job" validate:"required"`
 
 	// Global flags
-	Debug   bool   `cli:"debug"`
-	NoColor bool   `cli:"no-color"`
-	Profile string `cli:"profile"`
+	Debug   bool         `cli:"debug"`
+	NoColor bool         `cli:"no-color"`
+	Experiments []string `cli:"experiment" normalize:"list"`
+	Profile string       `cli:"profile"`
 
 	// API config
 	DebugHTTP        bool   `cli:"debug-http"`
@@ -102,6 +103,7 @@ var AnnotateCommand = cli.Command{
 		// Global flags
 		NoColorFlag,
 		DebugFlag,
+		ExperimentsFlag,
 		ProfileFlag,
 	},
 	Action: func(c *cli.Context) {
--- a/clicommand/pipeline_upload.go
+++ b/clicommand/pipeline_upload.go
@@ -56,13 +56,14 @@ type PipelineUploadConfig struct {
 	NoInterpolation bool   `cli:"no-interpolation"`
 
 	// Global flags
-	Debug   bool   `cli:"debug"`
-	NoColor bool   `cli:"no-color"`
-	Profile string `cli:"profile"`
+	Debug   bool         `cli:"debug"`
+	NoColor bool         `cli:"no-color"`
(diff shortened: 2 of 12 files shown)


lox:
This seems sensible, tests would too. It was something I'd been meaning to do. 

They get some coverage via the integration tests and the testing via the `bk` cli, but it's not adequate.

ticky:
> Do we have the test infrastructure in place to make that feasible (or easy)?

I think we might, but I don’t understand how it works _at all_! There’s this, but I can’t make heads nor tails of it: https://github.com/buildkite/agent/blob/master/agent/integration/job_runner_integration_test.go

pda:
Does the agent pass through experiments by setting `BUILDKITE_AGENT_EXPERIMENT` in the environment of jobs that it runs?

ticky:
> Does the agent pass through experiments by setting `BUILDKITE_AGENT_EXPERIMENT` in the environment of jobs that it runs?

It does! It fills that in with a list of whatever values are given. I tested this by printing that value from within a job.

clicommand/pipeline_upload.go
```diff
@@ -56,13 +56,14 @@ type PipelineUploadConfig struct {
...
-	Profile string `cli:"profile"`
+	Debug   bool         `cli:"debug"`
+	NoColor bool         `cli:"no-color"`
+	Experiments []string `cli:"experiment" normalize:"list"`
+	Profile string       `cli:"profile"`
 
 	// API config
 	DebugHTTP        bool   `cli:"debug-http"`
-	AgentAccessToken string `cli:"agent-access-token"`
+	AgentAccessToken string `cli:"agent-access-token" validate:"required"`
```
pda, line 66: I see this brings it consistent with all the others:

image

clicommand/annotate.go
```diff
@@ -55,9 +55,10 @@ type AnnotateConfig struct {
 	Job     string `cli:"job" validate:"required"`
 
 	// Global flags
-	Debug   bool   `cli:"debug"`
-	NoColor bool   `cli:"no-color"`
-	Profile string `cli:"profile"`
+	Debug   bool         `cli:"debug"`
+	NoColor bool         `cli:"no-color"`
+	Experiments []string `cli:"experiment" normalize:"list"`
```
pda, line 60: I see the global flags, including `Experiments`, are handled by HandleGlobalFlags() here:

https://github.com/buildkite/agent/blob/09bd9dbb76b23e986342685707751808e1c4da76/clicommand/global.go#L137-L147

PR #1223, opened 2020-06-04
Fix checkout when there is a file called HEAD in the repository root

zhenyavinogradov:
Problem: `git show HEAD` fails when there is a file called 'HEAD' in the
repository root, complaining about ambiguous reference

Solution: call `git show` without 'HEAD' argument, which will always show
the HEAD commit


zhenyavinogradov:
Thanks for replying, I've updated it to use `--`

PR #1230, opened 2020-06-15
Wrap calls for GCP metadata in a retry

jayco:
Problem:
The Buildkite agent is trying to get GCP metadata before the network is ready on the VM

Solution:
This wraps the call (TagsFromGCPMetaData) in a retry, giving it the same treatment as we do for [TagsFromEC2MetaData](https://github.com/buildkite/agent/blob/master/agent/tags.go#L87) 

Still to do:
  - Tests [ ]

### DIFF
--- a/agent/tags.go
+++ b/agent/tags.go
@@ -163,7 +163,7 @@ func (t *tagFetcher) Fetch(l logger.Logger, conf FetchTagsConfig) []string {
 			gcpTags, err := t.gcpMetaDataDefault()
 			if err != nil {
 				// Don't blow up if we can't find them, just show a nasty error.
-				l.Error(fmt.Sprintf("Failed to fetch Google Cloud meta-data: %s", err.Error()))
+				l.Error(fmt.Sprintf("Failed to fetch GCP meta-data: %s", err.Error()))
 			} else {
 				for tag, value := range gcpTags {
 					tags = append(tags, fmt.Sprintf("%s=%s", tag, value))
--- a/agent/tags.go
+++ b/agent/tags.go
@@ -163,13 +163,15 @@ func (t *tagFetcher) Fetch(l logger.Logger, conf FetchTagsConfig) []string {
 			gcpTags, err := t.gcpMetaDataDefault()
 			if err != nil {
 				// Don't blow up if we can't find them, just show a nasty error.
-				l.Error(fmt.Sprintf("Failed to fetch GCP meta-data: %s", err.Error()))
-			} else {
-				for tag, value := range gcpTags {
-					tags = append(tags, fmt.Sprintf("%s=%s", tag, value))
-				}
+				l.Error(fmt.Sprintf("Failed to fetch Google Cloud meta-data: %s", err.Error()))
+				return err
 			}
-			return err
+
+			for tag, value := range gcpTags {
+				tags = append(tags, fmt.Sprintf("%s=%s", tag, value))
+			}
+
+			return nil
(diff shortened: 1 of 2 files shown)


agent/tags.go
```diff
@@ -157,14 +157,24 @@ func (t *tagFetcher) Fetch(l logger.Logger, conf FetchTagsConfig) []string {
...
-		} else {
-			for tag, value := range gcpTags {
-				tags = append(tags, fmt.Sprintf("%s=%s", tag, value))
+		l.Info("Fetching GCP meta-data...")
+
+		err := retry.Do(func(s *retry.Stats) error {
+			gcpTags, err := t.gcpMetaDataDefault()
+			if err != nil {
+				// Don't blow up if we can't find them, just show a nasty error.
+				l.Error(fmt.Sprintf("Failed to fetch Google Cloud meta-data: %s", err.Error()))
```
lox, line 166: Nitpick, but if you used GCP above, use it here too.
```suggestion
				l.Error(fmt.Sprintf("Failed to fetch GCP meta-data: %s", err.Error()))
```

lox:
This is looking good @jayco, no substantial feedback.

agent/tags.go
```diff
@@ -157,14 +157,24 @@ func (t *tagFetcher) Fetch(l logger.Logger, conf FetchTagsConfig) []string {
...
-		} else {
-			for tag, value := range gcpTags {
-				tags = append(tags, fmt.Sprintf("%s=%s", tag, value))
+		l.Info("Fetching GCP meta-data...")
+
+		err := retry.Do(func(s *retry.Stats) error {
+			gcpTags, err := t.gcpMetaDataDefault()
+			if err != nil {
+				// Don't blow up if we can't find them, just show a nasty error.
+				l.Error(fmt.Sprintf("Failed to fetch GCP meta-data: %s", err.Error()))
```
pda, line 166: I think this can be turned into an early return, and the `else` code de-indented.

agent/tags.go
```diff
@@ -157,14 +157,24 @@ func (t *tagFetcher) Fetch(l logger.Logger, conf FetchTagsConfig) []string {
...
+				for tag, value := range gcpTags {
+					tags = append(tags, fmt.Sprintf("%s=%s", tag, value))
+				}
 			}
+			return err
+		}, &retry.Config{Maximum: 5, Interval: 1 * time.Second, Jitter: true})
+
+		// Don't blow up if we can't find them, just show a nasty error.
+		if err != nil {
+			l.Error(fmt.Sprintf("Failed to fetch EC2 meta-data: %s", err.Error()))
```
pda, line 177:
```suggestion
			l.Error(fmt.Sprintf("Failed to fetch GCP meta-data: %s", err.Error()))
```

agent/tags.go
```diff
@@ -157,14 +157,24 @@ func (t *tagFetcher) Fetch(l logger.Logger, conf FetchTagsConfig) []string {
...
+			if err != nil {
+				// Don't blow up if we can't find them, just show a nasty error.
+				l.Error(fmt.Sprintf("Failed to fetch GCP meta-data: %s", err.Error()))
+			} else {
+				for tag, value := range gcpTags {
+					tags = append(tags, fmt.Sprintf("%s=%s", tag, value))
+				}
 			}
+			return err
+		}, &retry.Config{Maximum: 5, Interval: 1 * time.Second, Jitter: true})
```
pda, line 173: I wonder if five seconds will always be enough for the described case of network not being online yet.
I haven't looked into the retry code, but do we have a back-off option? Or maybe just crank up the retry count?

PR #1233, opened 2020-06-26
update github.com/urfave/cli to v2

yob:
`github.com/urfave/cli` has started using go modules, and has released v2 with some breaking changes to the API. Nothing major, mostly just cosmetic changes.

I don't think there's any change in behavior that affects us.

I followed the [migration guide](https://github.com/urfave/cli/blob/master/docs/migrate-v1-to-v2.md), then fixed the remaining compiler errors.

I recommend reviewing the diff with whitespace changes hidden.


ticky:
I think this actually changed a bunch of how the CLI worked, with this heading in particular being dangerous for us: https://github.com/urfave/cli/blob/master/docs/migrate-v1-to-v2.md#flags-before-args

This means that, for instance, this command:

```bash
buildkite-agent meta-data get my-key --default fallback
```

doesn’t do what the user expects (the “default” flag in this is just _completely ignored_), and as of this patch, it is required to be written like this:

```bash
buildkite-agent meta-data get --default fallback my-key
```

which is a _major_, breaking change.

yob:
yer, what a nasty change :cry: 

There's some discussion here too https://github.com/buildkite/agent/issues/1285

ticky:
There’s some discussion of making it less stringent in urfave/cli#1113, but it doesn’t look like it’s moving too fast.

PR #1234, opened 2020-06-26
update aws-sdk-go to the latest version

yob:
We were using a version published nearly 2 years ago. We really only use the s3 support and I can't see any critical updates or bugfixes in that package, but no harm in trying the update anyway.

The CHANGELOG lists a number of fixes for rare bugs in session handling and retry logic that **might** be nice to have, but I have no evidence our customers are directly effected by any of the bugs that've been fixed.


yob:
I was a bit nervous about the large jump in an important dependency, but we only use a very small surface area of aws-sdk-go (the s3 bits) so the risk seems pretty low.

After merge we can try an edge build in our own stacks to confirm everything still works.

PR #1237, opened 2020-07-07
Fetch amd64 binaries on Apple Silicon Macs

ticky:
Upcoming macOS computers [will use the `arm64` architecture](https://developer.apple.com/documentation/xcode/building_a_universal_macos_binary). On such machines, our install script will currently fall through to attempting to fetch a `darwin-arm` binary, which does not exist. This future version of macOS allows `arm64` Macs to run `amd64` binaries through an emulation layer called Rosetta 2, and thus `darwin-amd64` is the most appropriate binary type for these machines until golang supports native `darwin-arm64` (see golang/go#38485).

This updates our install script to special-case the arm64 architecture when the platform is darwin, emit a message to the user, and proceed more or less as expected.

image

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -37,13 +37,23 @@ else
   case $UNAME in
     *amd64*)   ARCH="amd64"   ;;
     *x86_64*)  ARCH="amd64"   ;;
+    *arm64*)
+      ARCH="arm64"
+      if [[ "$PLATFORM" == "darwin" ]]; then
+        ARCH="amd64"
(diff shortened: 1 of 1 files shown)


ticky:
Confirmed that this updated install script behaves correctly on an Apple Silicon Mac, and that the buildkite-agent-darwin-amd64 binary runs (enough to emit usage, which seems like plenty in this case) 👍🏼

PR #1238, opened 2020-07-08
Stop building i386 for darwin after 14 years of amd64 Mac hardware

pda:
Apple switched from Intel Core Duo (32-bit) to Core 2 Duo (64-bit) around 2006. Nothing running darwin since then needs a 32-bit i386 build. Our usage stats show that nobody is using darwin-386 builds of buildkite-agent v3.x. AFAIK the latest macOS (10.15 Catalina) won't run 32-bit binaries at all. And upcoming Macs are shifting to arm64.

There's no big gain from dropping this, but no real reason to keep it either.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -112,16 +112,6 @@ steps:
         config: .buildkite/docker-compose.yml
         run: agent
 
-  - name: ":mac: 386"
-    command: ".buildkite/steps/build-binary.sh darwin 386"
-    key: build-binary-darwin-386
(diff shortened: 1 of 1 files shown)


ticky:
The biggest gain is just that it makes it less likely for someone to download the first binary which happens to say “Darwin” (which happens to be i386) and have it not work, given the current macOS version cannot run it, and future arm64 Macs definitely also cannot!

If people absolutely need a binary for this they can follow the build instructions! ✌🏼

pda:
That is a good point @ticky

PR #1239, opened 2020-07-08
submit basic OS info when registering from a BSD system

yob:
On macos, windows and linux the agent registration request includes some basic info on the operating system. Version, distribution, etc. This data is used only for metrics on what environments the agent is used in.

We publish official binaries for some BSD systems (FreeBSD, OpenBSD, NetBSD and dragonfly), however we haven't been submitting any OS versions so we have a poor understanding of how often these operating systems are used.

This change includes the output of `uname -a` in agent registration requests on those operating systems.

I checked to see if the BSDs use `/etc/os-release` or similar, but it seems there's no consensus. FreeBSD look like [they  might be adding it in the next release](https://reviews.freebsd.org/D22271), the others seem not to have it (based on googling).

We don't run tests against these GOOS so I haven't added any, but here's a screenshot of running an agent on a freebsd VM with this patch, and `--debug-http` enabled:

Screenshot from 2020-07-09 00-53-10

### DIFF
--- a/system/version_dump.go
+++ b/system/version_dump.go
@@ -15,6 +15,8 @@ func VersionDump(l logger.Logger) (string, error) {
 		return process.Run(l, "sw_vers")
 	} else if runtime.GOOS == "linux" {
 		return process.Cat("/etc/*-release")
+	} else if runtime.GOOS == "freebsd" || runtime.GOOS == "openbsd" || runtime.GOOS == "netbsd" || runtime.GOOS == "dragonfly" {
+		return process.Run(l, "uname", "-a")
 	}
 
 	return "", nil
--- a/system/version_dump.go
+++ b/system/version_dump.go
@@ -16,7 +16,7 @@ func VersionDump(l logger.Logger) (string, error) {
 	} else if runtime.GOOS == "linux" {
 		return process.Cat("/etc/*-release")
 	} else if runtime.GOOS == "freebsd" || runtime.GOOS == "openbsd" || runtime.GOOS == "netbsd" || runtime.GOOS == "dragonfly" {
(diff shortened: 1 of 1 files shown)


yob:
offline, @matthewd wondered if it was worth falling back to the `uname` strategy for any systemwhere we can find `uname`. 

We could do that, but I feel like difference is minimal. I believe the set of conditionals here captures all the remain systems we publish binaries for that will have uname.

yob:
> we could trim that further to something like `uname -sr` (`FreeBSD 12.1-RELEASE-p3` on FreeBSD, `Linux 4.19.97-v7+` on a Raspberry Pi, `Darwin 19.5.0` on macOS

Nice idea, I'll go with `-sr`

system/version_dump.go
```diff
@@ -15,6 +15,8 @@ func VersionDump(l logger.Logger) (string, error) {
 		return process.Run(l, "sw_vers")
 	} else if runtime.GOOS == "linux" {
 		return process.Cat("/etc/*-release")
+	} else if runtime.GOOS == "freebsd" || runtime.GOOS == "openbsd" || runtime.GOOS == "netbsd" || runtime.GOOS == "dragonfly" {
+		return process.Run(l, "uname", "-sr")
```
pda, line 19: The repetition of `runtime.GOOS` might show it's time for a `switch`:

```go
	switch runtime.GOOS {
	case "darwin":
		return process.Run(l, "sw_vers")
	case "linux":
		return process.Cat("/etc/*-release")
	case "freebsd", "openbsd", "netbsd", "dragonfly":
		return process.Run(l, "uname", "-sr")
	default:
		return "", nil
	}
```

Issue #1162, opened 2020-02-07
Can we get an rpm for aarch64


jnewbigin:
Hi,

We are looking at running some of our build agents on ARM.

Our current process does a yum install but we had to do manual setup for the ARM boxes. An rpm for aarch64 would streamline our agent setup.

Looking at https://github.com/buildkite/agent/blob/aa5791c912c04ec261191a5e3561b8605a248e1f/.buildkite/steps/build-rpm-packages.sh#L30 it might be straightforward.

I am happy to test anything if that helps.
Thanks

jnewbigin:
I am happy to test a package for you. You can dm me on buildkite-community.slack.com if the file is not public.

chloeruka:
Hey @jnewbigin, I've published an edge release if you want to give this one a try with your infrastructure/ARM builds. Edge release is at `yum.buildkite.com/buildkite-agent/experimental/aarch64/buildkite-agent_3.23-beta.1-3452_aarch64.rpm`

So to install via yum:
1. Install our repo: 
```sh
sudo sh -c 'echo -e "[buildkite-agent]\nname = Buildkite Pty Ltd\nbaseurl = https://yum.buildkite.com/buildkite-agent/experimental/aarch64/\nenabled=1\ngpgcheck=0\npriority=1" > /etc/yum.repos.d/buildkite-agent.repo'
```

2. Configure and start the agent:
```sh
sudo yum -y install buildkite-agent
sudo sed -i "s/xxx/INSERT-YOUR-AGENT-TOKEN-HERE/g" /etc/buildkite-agent/buildkite-agent.cfg
sudo systemctl enable buildkite-agent && sudo systemctl start buildkite-agent
sudo tail -f /var/log/messages
```

3. If everything went OK, you'll see your agent in the console:
<img width="585" alt="image" src="https://user-images.githubusercontent.com/30171259/87495249-3d0dde80-c694-11ea-93e4-2acc7ab90459.png">

jnewbigin:
Hi
Sorry it took so long to get round to this.
I can confirm that I can yum install and use the aarch64 rpm agent on CentOS aarch64.
Thanks
John

jayco:
@jnewbigin thanks for letting us know

PR #1241, opened 2020-07-10
Experimental: Build RPM package for ARM64 architecture (aarch64)

chloeruka:
To address #1162 I'm tossing together a naive package builder codepath for RPM using our linux `arm64` binary to upload with the architecture `aarch64`. From googling, this appears to be the canonical architecture nomenclature for Red Hat based systems to refer to modern ARM processors. 

Please let me know if there's more that might need to be done, of particular concern is `scripts/build-rpm-package.sh` which possibly might need other adjustments to support ARM directly. I'd like to merge this ASAP to test it on an agent. I also think @jnewbigin has offered to do some testing if they see this 😄

### DIFF
--- a/.buildkite/steps/build-rpm-packages.sh
+++ b/.buildkite/steps/build-rpm-packages.sh
@@ -24,7 +24,7 @@ rm -rf rpm
 
 # Build the packages into rpm/
 PLATFORM="linux"
-for ARCH in "amd64" "386"; do
+for ARCH in "amd64" "386" "arm64"; do
   echo "--- Building rpm package ${PLATFORM}/${ARCH}"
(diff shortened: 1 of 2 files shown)


pda:
> I'd like to merge this ASAP to test it on an agent.

@chloeruka: does [this (non-public) build step](https://buildkite.com/buildkite/agent/builds/3443#a1f7e5dd-1624-4a70-b450-83a611a3973d) give you any useful feedback? It has the `rpm/buildkite-agent_3.22.1-3443_aarch64.rpm` as an artifact.

chloeruka:
Merging to create an edge release for further testing

PR #1242, opened 2020-07-13
allow CleanCheckout to be set via hooks

niceking:
Add functionality to support `BUILDKITE_CLEAN_CHECKOUT` being set via hooks, and some fancy reflections to handle the conversion of the string value of this environment variable to a boolean.

### DIFF
--- a/bootstrap/config.go
+++ b/bootstrap/config.go
@@ -59,7 +59,7 @@ type Config struct {
 	AgentName string
 
 	// Should the bootstrap remove an existing checkout before running the job
-	CleanCheckout bool
+	CleanCheckout bool `env:"BUILDKITE_CLEAN_CHECKOUT"`
 
 	// Flags to pass to "git clone" command
 	GitCloneFlags string `env:"BUILDKITE_GIT_CLONE_FLAGS"`
--- a/bootstrap/config.go
+++ b/bootstrap/config.go
@@ -2,6 +2,9 @@ package bootstrap
 
 import (
 	"reflect"
+	"strconv"
+
+	"log"
 
 	"github.com/buildkite/agent/v3/env"
 )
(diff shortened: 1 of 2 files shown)


lox:
Nice one @niceking!  

The reason we didn't allow this initially is we were concerned about the possibility of command-line injection attacks via ENV at a pipeline level. I'm probably less concerned about this after some time has played out, but that is the historical context.

lox:
Oh! I think I am thinking of `BUILDKITE_GIT_CLEAN_FLAGS` ignore me 😅

jayco:
@niceking I think the issue is here https://github.com/buildkite/agent/blob/30a64c6166f7d82843f138dae62fbef16611e338/bootstrap/config.go#L148, so you could perhaps change that code that does the setting in the ReadFromEnvironment method to:
```
			// We only care if the value has changed
			if newValue != value.String() {
				switch value.Kind() {
				case reflect.Bool:
					b, _ := strconv.ParseBool(newValue)
					value.SetBool(b)
				default:
					value.SetString(newValue)
				}
				changed[tag] = newValue
			}
```

I am not sure if we want to have explicit checks in there, but if we are going to allow other possible env vars to be passed through in the future then we may have to. What are you thoughts @pda @lox

bootstrap/config.go
```diff
@@ -59,7 +59,7 @@ type Config struct {
 	AgentName string
 
 	// Should the bootstrap remove an existing checkout before running the job
-	CleanCheckout bool
+	CleanCheckout bool `env:"BUILDKITE_CLEAN_CHECKOUT"`
```
pda, line 62: I initially didn't understand how adding this made sense.

The hooks changing `BUILDKITE_CLEAN_CHECKOUT` were being run _inside_ the bootstrap process, not prior to it running, so how does passing it to the bootstrap as environment help?

Then I realised this gets called after each hook: https://github.com/buildkite/agent/blob/75c55cb5c99d8635d69ed6456b6643de861a0a95/bootstrap/bootstrap.go#L253

So we're not really expecting it to be passed to the bootstrap as environment, but we do want the internal config `ReadFromEnvironment()` to understand it in the fake environment that we build after running hooks.

👌🏼

pda:
Nice work @niceking 👌🏼
Sorry you landed in a gnarly pit of [Go reflection](https://blog.golang.org/laws-of-reflection) 😬

And yep that's roughly the solution @jayco — just with a bit more ceremony to check if the bool value has changed… `value.String()` on a reflected bool is always `"<bool Value>"`.

I've pushed a commit that refactors `ReadFromEnvironment` a bit, and adds support for `bool` config values.

bootstrap/config.go
```diff
@@ -131,22 +134,38 @@ func (c *Config) ReadFromEnvironment(environ *env.Environment) map[string]string
 	changed := map[string]string{}
 
 	// Use reflection for the type and values
-	t := reflect.TypeOf(*c)
-	v := reflect.ValueOf(c).Elem()
+	fields := reflect.TypeOf(*c)
+	values := reflect.ValueOf(c).Elem()
```
pda, line 138: We had short-named var `v` outside the loop, and long-named var `value := v.Field(i)` inside, which felt backwards, and tripped me up. So I've given them sensible long names out here, and use `f` and `v` in the loop.

bootstrap/config.go
```diff
@@ -131,22 +134,38 @@ func (c *Config) ReadFromEnvironment(environ *env.Environment) map[string]string
...
+					log.Printf("warning: cannot parse %s=%s as bool, ignoring", tag, newStr)
+					break
+				}
+				if newBool == v.Bool() {
+					break
+				}
+				v.SetBool(newBool)
+				changed[tag] = newStr
+			default:
+				log.Printf("warning: bootstrap.Config.ReadFromEnvironment does not support %v for %s", v.Kind(), tag)
```
pda, line 168: The only other type in the Config struct is `int`, but it doesn't have an `env` tag so I don't think it's relevant here yet. If it becomes relevant, we can add a `case reflect.Int` with int-parsing etc.

I felt like a warning was better than an error… maybe 🤔

pda:
We now get "Cleaning pipeline checkout" if a hook sets `BUILDKITE_CLEAN_CHECKOUT` 👌🏼

image

PR #1249, opened 2020-07-18
Log HTTP requests to S3 if the global DebugHTTP flag is on

yob:
The DebugHTTP flag is a helpful tool for debugging what's happening when something goes wrong on customer-managed build servers.

This change ensures that flag applies to artifact uploads and downloads from S3. Hopefully this will help us understand what's happening when artifact uploads fail in interesting ways.

### DIFF
--- a/agent/s3.go
+++ b/agent/s3.go
@@ -93,7 +93,7 @@ func awsS3Session(region string) (*session.Session, error) {
 	return sess, nil
 }
 
-func newS3Client(l logger.Logger, bucket string) (*s3.S3, error) {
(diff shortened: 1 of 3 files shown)


yob:
This turns out to not be a thing I need right now.

I'm aiming to debug uploads to the buildkite owned S3 bucket, and it turns out they're done by the `FormUploader`. `S3Uploader` is used for uploads to customer owned buckets.

Interesting that `S3Uploader` accepts a `DebugHTTP` option but doesn't use it. We should fix that one day - maybe we'll want to debug uploads to a customer bucket?

PR #1250, opened 2020-07-20
Fix unsetting of agent config env to work with ufave/cli/v2

yob:
This method dynamically fetches a list of all the environment variables that can be used to configure the agent, and removes them from the environment so they're not passed down to child processes (like the bootstrap used to run a job).

It's implemented using reflection on the flag structs from the `ufave/cli` package we use for CLI structure. It checks the EnvVar field on the flag struct, which works just fine with `ufave/cli/v1`. However, that field doesn't exist any more in `ufave/cli/v2` and this method has started silently failing since we upgraded to v2 in #1233.

In commit 1, I changed the method to fail loudly if the flag structs don't have the EnvVar field we expect. This would've made the issue obvious when #1233 was first proposed.

In commit 2, I update the method to work with `ufave/cli/v2`. In `ufave/cli/v1` each flag supported a single environment variable key, stored in the EnvVar field. In `ufave/cli/v2` each flag can have multiple environment variable keys and they're stored in a slice of Strings called EnvVars (note the `S`).

I've also added some safety checks that will fail loudly if `ufave/cli/v3` decides to change/rename things again.

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -489,7 +489,11 @@ var AgentStartCommand = cli.Command{
 		defer done()
 
 		// Remove any config env from the environment to prevent them propagating to bootstrap
-		UnsetConfigFromEnvironment(c)
+		err = UnsetConfigFromEnvironment(c)
+		if err != nil {
+			fmt.Printf("%s", err)
+			os.Exit(1)
+		}
 
 		// Check if git-mirrors are enabled
 		if experiments.IsEnabled(`git-mirrors`) {
--- a/clicommand/global.go
+++ b/clicommand/global.go
@@ -1,6 +1,7 @@
 package clicommand
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"reflect"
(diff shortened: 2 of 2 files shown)


yob:
This is my first attempt at golang reflection, so feedback welcome!

clicommand/agent_start.go
```diff
@@ -489,7 +489,11 @@ var AgentStartCommand = cli.Command{
 		defer done()
 
 		// Remove any config env from the environment to prevent them propagating to bootstrap
-		UnsetConfigFromEnvironment(c)
+		err = UnsetConfigFromEnvironment(c)
+		if err != nil {
+			fmt.Printf("%s", err)
+			os.Exit(1)
+		}
```
yob, line 496: exit feels a bit harsh, but we do it earlier in this method when a thing that should never happen ... happens. I followed the pattern.

clicommand/global.go
```diff
@@ -150,19 +150,26 @@ func HandleGlobalFlags(l logger.Logger, cfg interface{}) func() {
...
-			for _, env := range strings.Split(envVars, ",") {
-				os.Unsetenv(env)
-			}
+		f := reflect.Indirect(r).FieldByName(`EnvVars`)
+		if !f.IsValid() {
+			return errors.New("EnvVars field not found on flag")
+		}
+		if f.Kind() != reflect.Slice {
+			return errors.New("EnvVars not a slice")
+		}
```
yob, line 164: fail loudly if `ufave/cli/v3` decides to change/rename things again.

yob:
master is a bit broken without this, so I'm going to merge.

@lox @pda if you have any retrospective feedback (particularly on the reflection code), I'm happy to adjust in a follow up PR :pray:

ticky:
I think this actually changed a bunch of how the CLI worked, with this heading in particular being dangerous for us: https://github.com/urfave/cli/blob/master/docs/migrate-v1-to-v2.md#flags-before-args

This means that, for instance, this command:

```bash
buildkite-agent meta-data get my-key --default fallback
```

doesn’t do what the user expects (the “default” flag in this is just _completely ignored_), and as of this patch, it is required to be written like this:

```bash
buildkite-agent meta-data get --default fallback my-key
```

which is a _major_, breaking change.

PR #1251, opened 2020-07-20
Enable BUILDKITE_AGENT_DEBUG_HTTP in jobs if it's enabled in the agent process

yob:
Depends on #1250.

`BUILDKITE_AGENT_DEBUG` is already passed down to the job environment, and this expands that behaviour to also include `BUILDKITE_AGENT_DEBUG_HTTP`.

This is desirable because the build-agent binary is often used with jobs (eg. for artifact uploading/downloading) and from time to time it can be helpful to see precisely what those HTTP requests look like.

### DIFF
--- a/agent/agent_worker.go
+++ b/agent/agent_worker.go
@@ -19,6 +19,9 @@ type AgentWorkerConfig struct {
 	// Whether to set debug in the job
 	Debug bool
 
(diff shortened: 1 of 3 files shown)


yob:
I rebased onto master and force pushed, no change  to the code

PR #1255, opened 2020-07-23
log the S3 IP used during artifact uploads

yob:
We get occasional reports of slow and/or failed S3 uploads. One possibility is that there's networking issues between the agent and us-east-1 (where the default buildkite-managed artifacts bucket lives).

The IP address for s3.amazonaws.com rotates every 5 seconds, so from upload to upload it's impossible for us to guess the server IP that was used after the fact. This will output a log line for each upload (when `BUILDKITE_AGENT_DEBUG_HTTP=true`) with the exact server IP address that was used.

If we see upload issues in the future that sound like a networking issue, at least we'll an address to debug with.

the log output looks like this:

```
2020-07-23 21:21:56 DEBUG  artifact 8d67fb1c-83c6-4006-8a3b-17c438918197 uploading to: 52.216.237.165:443
```

### DIFF
--- a/agent/form_uploader.go
+++ b/agent/form_uploader.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"mime/multipart"
 	"net/http"
+	"net/http/httptrace"
 	"net/http/httputil"
 	"regexp"
 	"strings"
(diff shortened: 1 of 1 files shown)


yob:
I'm hoping this will help with debugging an active customer issue, so I'm going to go ahead and merge.

Ping @lox @pda - I can adjust after-the-fact f you have feedback!

pda:
👍🏼

Issue #1188, opened 2020-03-02
Resolve BUILDKITE_COMMIT after checkout?


sj26:
When `$BUILDKITE_COMMIT` is `HEAD`, would it make sense to update `$BUILDKITE_COMMIT` to the resolved commit after checkout? We already do this for pipeline upload jobs:

https://github.com/buildkite/agent/pull/871

but not for pipelines which don't include a pipeline upload step. This means a pipeline with a default build pipeline configured from the dashboard and a scheduled build at `master` and `HEAD` will have a `$BUILDKITE_COMMIT` value of `HEAD` even after checkout, and may fail if the first job includes logic which presumes that `$BUILDKITE_COMMIT` holds an actual commit.

Perhaps this could be introduced as an agent bootstrap option `BUILDKITE_AGENT_RESOLVE_COMMIT_AFTER_CHECKOUT` or something which will default to true after v4?

toolmantim:
Could you share more about why were you worried about making this the default behavior immediately? At a glance, it seems more like a bug that needs fixing, vs a breaking change?

sj26:
True, the value is sent back to the buildkite api and becomes the BUILDKITE_COMMIT for all future jobs, maybe committing (lol) the change locally as well makes sense and should be considered a bug. I was just worried about plugins or pipelines which might be relying on this behaviour somehow, but I can't think of a good reason, and haven't seen an actual case, so I'm probably being overly cautious. 😅

keithduncan:
I just ran in to this today 😄

I knew that `BUILDKITE_COMMIT` was a commit sha in some of my pipelines, but wasn’t sure why my newer ones didn’t have this. Searching I found https://github.com/buildkite/agent/pull/871 and realised that since I’ve switched to using the YAML steps editor instead of pipeline upload step my pipelines don’t get a resolved sha ☹️

ailisp:
I was using BUILDKITE_COMMIT to tag build docker images xxx:$BUILDKITE_BRANCH-$BUILDKITE_COMMIT, but this results image names to be xxx:master-HEAD.  To work around this i have to do:
```
      if [[ ${BUILDKITE_COMMIT} == "HEAD" ]]; then
        export BUILDKITE_COMMIT=$(git rev-parse HEAD)
      fi
```
[...]

PR #1256, opened 2020-08-05
Commit after checkout experiment

jayco:
Adds the same gear in https://github.com/buildkite/agent/pull/871 into the bootstrap under an experiment to address issue https://github.com/buildkite/agent/issues/1188

### DIFF
--- a/EXPERIMENTS.md
+++ b/EXPERIMENTS.md
@@ -44,3 +44,12 @@ Take `buildkite-agent artifact upload coverage\report.xml` as an example:
 - With this experiment enabled, it would be `s3://example/coverage/report.xml`.
 
 **Status**: a major improvement for Windows compatibility, we'd like this to be the standard behaviour in 4.0. 👍👍
+
+### `resolve-commit-after-checkout`
+
+When $BUILDKITE_COMMIT is HEAD, this updates $BUILDKITE_COMMIT to the resolved commit after checkout.
+
+This fixes the issue where a pipeline with a default build pipeline configured from the dashboard and a scheduled build at master and HEAD will have a $BUILDKITE_COMMIT value of HEAD even after checkout.
+The result is that the build may fail if the first job includes logic which presumes that $BUILDKITE_COMMIT holds an actual commit.
+
+**Status**: broadly useful, we'd like this to be the standard behaviour in 4.0. 👍👍
--- a/EXPERIMENTS.md
+++ b/EXPERIMENTS.md
@@ -47,8 +47,6 @@ Take `buildkite-agent artifact upload coverage\report.xml` as an example:
 
 ### `resolve-commit-after-checkout`
 
-When BUILDKITE_COMMIT is HEAD, this updates BUILDKITE_COMMIT to the resolved commit after checkout.
-
-Fixes unexpected behaviour where a pipeline with builds at master and HEAD will have a BUILDKITE_COMMIT value of HEAD even after checkout.
+After repository checkout, resolve `BUILDKITE_COMMIT` to a commit hash. This makes `BUILDKITE_COMMIT` useful for builds triggered against non-commit-hash refs such as `HEAD`.
 
 **Status**: broadly useful, we'd like this to be the standard behaviour in 4.0. 👍👍
--- a/EXPERIMENTS.md
+++ b/EXPERIMENTS.md
@@ -47,9 +47,8 @@ Take `buildkite-agent artifact upload coverage\report.xml` as an example:
 
 ### `resolve-commit-after-checkout`
 
-When $BUILDKITE_COMMIT is HEAD, this updates $BUILDKITE_COMMIT to the resolved commit after checkout.
+When BUILDKITE_COMMIT is HEAD, this updates BUILDKITE_COMMIT to the resolved commit after checkout.
 
-This fixes the issue where a pipeline with a default build pipeline configured from the dashboard and a scheduled build at master and HEAD will have a $BUILDKITE_COMMIT value of HEAD even after checkout.
-The result is that the build may fail if the first job includes logic which presumes that $BUILDKITE_COMMIT holds an actual commit.
+Fixes unexpected behaviour where a pipeline with builds at master and HEAD will have a BUILDKITE_COMMIT value of HEAD even after checkout.
 
 **Status**: broadly useful, we'd like this to be the standard behaviour in 4.0. 👍👍
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -1304,8 +1304,10 @@ func (b *Bootstrap) resolveCommit() {
 		return
 	}
 	trimmedCmdOut := strings.TrimSpace(string(cmdOut))
-	b.shell.Commentf("Updating BUILDKITE_COMMIT from %q to %q", commitRef, trimmedCmdOut)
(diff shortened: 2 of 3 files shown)


toolmantim:
This is nice! Any reason we can't release it as a bugfix, rather than having it under an experimental flag?

lox:
This is great!

jayco:
@toolmantim there was some concern was that might be too much of a breaking change mentioned here https://github.com/buildkite/agent/pull/871, I discussed with @pda and we thought an experiment might be the best solution here?

toolmantim:
👍🏼

bootstrap/bootstrap.go
```diff
@@ -1265,6 +1265,21 @@ func (b *Bootstrap) defaultCheckoutPhase() error {
...
+			if err != nil {
+				b.shell.Warningf("Error running git rev-parse %q: %v", commitRef, err)
+			} else {
+				trimmedCmdOut := strings.TrimSpace(string(cmdOut))
+				b.shell.Commentf("Updating BUILDKITE_COMMIT to %q", trimmedCmdOut)
+				b.shell.Env.Set(`BUILDKITE_COMMIT`, trimmedCmdOut)
+			}
+		}
+	}
+
```
pda, line 1282: @jayco What do you think about splitting this out into a method, so that we can use idiomatic early-returns instead of incrementally-indented blocks? e.g.

```diff
diff --git a/bootstrap/bootstrap.go b/bootstrap/bootstrap.go
index 82021c6b..ae064bd4 100644
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -1268,16 +1268,7 @@ func (b *Bootstrap) defaultCheckoutPhase() error {
 	// resolve BUILDKITE_COMMIT based on the local git repo
 	if experiments.IsEnabled(`resolve-commit-after-checkout`) {
 		b.shell.Commentf("Using resolve-commit-after-checkout experiment 🧪")
-		if commitRef, ok := b.shell.Env.Get(`BUILDKITE_COMMIT`); ok {
-			cmdOut, err := b.shell.RunAndCapture(`git`, `rev-parse`, commitRef)
-			if err != nil {
-				b.shell.Warningf("Error running git rev-parse %q: %v", commitRef, err)
-			} else {
-				trimmedCmdOut := strings.TrimSpace(string(cmdOut))
-				b.shell.Commentf("Updating BUILDKITE_COMMIT to %q", trimmedCmdOut)
-				b.shell.Env.Set(`BUILDKITE_COMMIT`, trimmedCmdOut)
-			}
-		}
+		b.resolveCommit()
 	}
 
 	// Grab author and commit information and send
@@ -1301,6 +1292,22 @@ func (b *Bootstrap) defaultCheckoutPhase() error {
 	return nil
 }
 
+func (b *Bootstrap) resolveCommit() {
+	commitRef, _ := b.shell.Env.Get("BUILDKITE_COMMIT")
+	if commitRef == "" {
+		b.shell.Warningf("BUILDKITE_COMMIT was empty")
+		return
+	}
+	cmdOut, err := b.shell.RunAndCapture(`git`, `rev-parse`, commitRef)
+	if err != nil {
+		b.shell.Warningf("Error running git rev-parse %q: %v", commitRef, err)
+		return
+	}
+	trimmedCmdOut := strings.TrimSpace(string(cmdOut))
+	b.shell.Commentf("Updating BUILDKITE_COMMIT from %q to %q", commitRef, trimmedCmdOut)
+	b.shell.Env.Set(`BUILDKITE_COMMIT`, trimmedCmdOut)
+}
+
 // CommandPhase determines how to run the build, and then runs it
 func (b *Bootstrap) CommandPhase() error {
 	if err := b.executeGlobalHook("pre-command"); err != nil {
```

pda:
@jayco I was a bit nervous about the test potentially leaving the experiment enabled if it returns early, so I played with a nice(?) way of deferring the undo of enabling the experiment in f36b16cfad5a6043dfef56f7c3e60e7d1f8bfa64. It's maybe a bit unclear though, that `defer foo()()` calls foo immediately and defers calling the function returned by foo 🤔

bootstrap/bootstrap.go
jayco: Makes sense 👍

pda: I committed it as 15c642d3.

EXPERIMENTS.md
```diff
@@ -44,3 +44,11 @@ Take `buildkite-agent artifact upload coverage\report.xml` as an example:
 - With this experiment enabled, it would be `s3://example/coverage/report.xml`.
 
 **Status**: a major improvement for Windows compatibility, we'd like this to be the standard behaviour in 4.0. 👍👍
+
+### `resolve-commit-after-checkout`
+
+When BUILDKITE_COMMIT is HEAD, this updates BUILDKITE_COMMIT to the resolved commit after checkout.
```
pda, line 50: Looks like the behaviour actually resolves any `BUILDKITE_COMMIT`, not just `HEAD`. That means it'll work for things like `HEAD^` and `HEAD~2` and I guess `master` and things like that. I think it'll also resolve commit hashes, presumably to themselves, but it would expand short commit hashes to full ones?

Do you think `resolveCommit()` should be a no-op if `BUILDKITE_COMMIT` already looks like a commit SHA? It might be tricky to know for sure. Or maybe it's harmless (and safer) always running it. 🤷‍♂️

Either way, I think this line of doc should say something like:

> Resolve `BUILDKITE_COMMIT` (which supports refs like `HEAD`) to a specific commit hash after checkout.

pda: I did some testing:

> Using resolve-commit-after-checkout experiment 🧪
> Updating BUILDKITE_COMMIT from "HEAD" to "5051b52eb78ca152edaeeed10743ed582f3d176a"

So far so good.

---

> Using resolve-commit-after-checkout experiment 🧪
> Updating BUILDKITE_COMMIT from "HEAD~1" to "d1be6bc8485bec8475f445a43ecd772efe3c14be"

I guess it's nice that it works on more interesting refs…

---

> Using resolve-commit-after-checkout experiment 🧪
> Updating BUILDKITE_COMMIT from "master" to "d1be6bc8485bec8475f445a43ecd772efe3c14be"

… including branches.

---

> Using resolve-commit-after-checkout experiment 🧪
> Updating BUILDKITE_COMMIT from "d1be6b" to "d1be6bc8485bec8475f445a43ecd772efe3c14be"

It does indeed expand short commit hashes, which is probably good?

---

> Using resolve-commit-after-checkout experiment 🧪
> Updating BUILDKITE_COMMIT from "d1be6bc8485bec8475f445a43ecd772efe3c14be" to "d1be6bc8485bec8475f445a43ecd772efe3c14be"

But that ends up looking weird when the `BUILDKITE_COMMIT` is already a commit SHA.

pda: > > Using resolve-commit-after-checkout experiment 🧪
> > Updating BUILDKITE_COMMIT from "d1be6bc8485bec8475f445a43ecd772efe3c14be" to "d1be6bc8485bec8475f445a43ecd772efe3c14be"
> 
> But that ends up looking weird when the BUILDKITE_COMMIT is already a commit SHA.

Fixed this one in ffb6ac48

pda (approved):
👌🏼

PR #1261, opened 2020-08-07
Job should fail when process.Run() -> err

sj26:
In rare cases (like a server without much memory left), it possible starting the bootstrap child process can return an error:

```go
if err := r.process.Run(); err != nil {
```

We handle the error, however we've had reports that the job may be reported as passed rather than failed. Here's two screenshots demonstrating what it can look like:

Screenshot from 2020-08-10 10-05-45

Screenshot from 2020-08-11 11-00-24

This part is suspicious:

> Exited with status 0

In this failure case the process never started, so it doesn't have an exit code. This PR synthesizes a `-1` exit status to ensure the job is marked as failed.

 `-1` is an overloaded exit status though, so we also send along a "reason" string. That's not displayed in the web UI yet, but at least we have a permanent record of what causes the `-1` and job failure.

### DIFF
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -246,13 +246,28 @@ func (r *JobRunner) Run() error {
 		return err
 	}
 
+	// Default exit status is no exit status
+	exitStatus := ""
+	signal := ""
+	signalReason := ""
+
 	// Run the process. This will block until it finishes.
 	if err := r.process.Run(); err != nil {
 		// Send the error as output
 		r.logStreamer.Process(fmt.Sprintf("%s", err))
+
+		// The process did not run at all, so make sure it fails
+		exitStatus = "-1"
+		signalReason = "process_run_error"
 	} else {
 		// Add the final output to the streamer
 		r.logStreamer.Process(r.output.String())
+
+		// Collect the finished process' exit status
+		exitStatus = fmt.Sprintf("%d", r.process.WaitStatus().ExitStatus())
+		if ws := r.process.WaitStatus(); ws.Signaled() {
+			signal = process.SignalString(ws.Signal())
+		}
 	}
 
 	// Store the finished at time
@@ -284,12 +299,6 @@ func (r *JobRunner) Run() error {
 		r.logger.Debug("[JobRunner] Deleted env file: %s", r.envFile.Name())
 	}
 
-	exitStatus := fmt.Sprintf("%d", r.process.WaitStatus().ExitStatus())
-	signal := ""
-	if ws := r.process.WaitStatus(); ws.Signaled() {
-		signal = process.SignalString(ws.Signal())
-	}
-
 	// Write some metrics about the job run
 	jobMetrics := r.metrics.With(metrics.Tags{
 		"exit_code": exitStatus,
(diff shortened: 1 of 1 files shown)


pda:
Looking good 👌🏼

agent/job_runner.go
```diff
@@ -517,18 +526,23 @@ func (r *JobRunner) startJob(startedAt time.Time) error {
...
-	if r.stopped {
+	if signalReason != "" {
+		r.job.SignalReason = signalReason
+	} else if r.stopped {
+		// If the agent has been stopped, send a signal reason of
+		// `agent_stop` to distinguish between user-generated cancels
+		// and those due to the agent getting an operating system
+		// signal.
 		r.job.SignalReason = `agent_stop`
 	} else if r.cancelled {
```
yob, line 543: It's not really a problem with this PR, but can this branch ever be true? I can't see where `r.cancelled` is set to True

sj26: No idea :-)

sj26: Looks like it should probably be set to true somewhere in `func (r *JobRunner) Cancel() error`

yob: yup, PR incoming

yob: looks like we stopped setting `r.cancelled` in 6275d45e, as part of #882

sj26: It looks like `r.cancelled = true` was removed in 6275d45e4eda1111b326c71668ab76120fbd6ff3, I think perhaps by mistake

yob: I think setting it in `func (r *JobRunner) Cancel() error` will work, but then when `func (r *JobRunner) CancelAndStop()` is called as part of graceful agent shutdown we'll end up with both `stopped` AND `cancelled` set to `True` :thinking:

sj26: I pushed up a quick commit which I think will fix.

sj26: I'd be tempted to move the signalReason calculation up into `Run` instead of having it within finishJob, tbqh.

yob: Agreed

sj26: Oh sorry @yob, I didn't see "yup, PR incoming" until I re-read this from a notification. I think GitHub wasn't rendering it in page until I did a proper refresh.

Feel free to pop that commit off here and do it separately.

agent/job_runner.go
```diff
@@ -299,11 +308,22 @@ func (r *JobRunner) Run() error {
...
+	if signalReason == "" {
+		if r.stopped {
+			// The agent is being gracefully stopped, and we signalled the job to end. Often due
+			// to pending host shutdown or EC2 spot instance termination
+			signalReason = `agent_stop`
+		} else if r.cancelled {
+			// The job was signalled because it was cancelled via the buildkite web UI
+			signalReason = `cancel`
+		}
+	}
```
sj26, line 320: Perhaps this logic could go where we set `exitStatus` from the process?

https://github.com/buildkite/agent/blob/be13cdb015da221f61ddec772846a292f1d70e0c/agent/job_runner.go#L260-L267

yob: nice idea. I've put it inside the conditional that sets `signal` - it feels weird to leave a possibility that `signalReason` is set but `signal` is blank...

sj26: I guess `signalReason == ""` is synonymous for "unknown", i.e. the job might have been sent a signal by something outside the agent's control?

sj26:
@lox it made sense when the finish job message was being crafted from shared state, but I think it's better in the caller now that it's calculated up there. :+1:

yob:
Cool, let's merge and see if this helps. We'll have to reach out to some beta testers to try it, and we can watch the API to see if any jobs receive `signal_reason=process_run_error`

sj26:
@yob maybe we could have the backend throw `process_run_error` into Bugsnag for visibility, or something?

pda:
Nice work @sj26 & @yob 👌🏼

PR #1263, opened 2020-08-09
pass sidecar docker image to release pipeline

yob:
In #1218 we added a new docker image to the build and release, but we forgot to pass the meta_data through to the release pipeline. This resulted in the release pipeline failing - eg:
https://buildkite.com/buildkite/agent-release-edge/builds/294#8d50db3e-781c-4690-9378-f1ddf83988be


yob:
The failure this hopes to fix is in the release pipelines, so the only way to test it is to merge and see if it helps.

YOLO.

yob:
it worked :raised_hands:

PR #1273, opened 2020-08-24
Add support for tracing

goodspark:
This adds support for tracing via Datadog (wrapped in Opentracing so in the future people can integrate their own tracing services if needed).

### Why?

As we build out larger CICD agent pools/queues, we want to be able to see failures and metrics at a holistic level while also being able to zoom into specific jobs to see causes of failures or performance regressions (eg. Where was the failure/latency? Was it the specific agent? Something about the hooks?). The current metrics don't give us this kind of granularity.

### Changes

This is for the most part broken out self-contained commits, with the first one being a quick POC we did to prove out the idea.

- Tracing is reporting if the `BUILDKITE_TRACING_DATADOG_ADDR` env var or the `tracing-datadog-addr` config is set.
- If neither are, the tracing code still runs, but uses a no-op tracer so that things aren't reported.
- This will also ensure distributed tracing works between any dynamic jobs that might be created by encoding tracing context data into an env var and propagating that to the new jobs.
- Includes tags to traces such as agent, queue, pipeline IDs, etc. This will help make aggregation in reporting easier.
- Spans are wrapped around the major phases of a job and its hooks (pre-commit, etc).
- Errors propagate back up to the root span so the trace itself will be marked erroneous if something failed.
- Service name is ` buildkite-agent`, operation name is `job.run`, resource names follow an `org / pipeline / job label` pattern

### Examples

Screenshot from 2020-08-24 16-12-42
This shows one main job spawning off five more jobs, which are all linked to the first.

Screenshot from 2020-08-24 16-18-02
Zoom into a specific failing job (in this case the failing coming from a command)


lox:
This is amazing! 😳 Thanks so much for your time and effort @goodspark, we'll review and come back to you with feedback.

bootstrap/shell/shell.go
```diff
@@ -240,10 +250,22 @@ func (s *Shell) RunAndCapture(command string, arg ...string) (string, error) {
...
+// injectTraceCtx adds tracing information to the given env vars to support
+// distributed tracing across jobs/builds.
+func (s *Shell) injectTraceCtx(ctx context.Context, env *env.Environment) {
+	span := opentracing.SpanFromContext(ctx)
+	// Not all shell runs will have tracing (nor do they really need to).
+	if span == nil {
+		return
+	}
+	// Intentionally ignoring error. Logging will spam agent logs.
+	_ = opentracing.GlobalTracer().Inject(span.Context(), opentracing.TextMap, env.ToMap())
```
petemounce, line 262: Could this be ignored _unless_ the buildkite-agent has had the `--debug` flag specified, please?

clicommand/agent_start.go
```diff
@@ -84,6 +84,7 @@ type AgentStartConfig struct {
 	HealthCheckAddr            string   `cli:"health-check-addr"`
 	MetricsDatadog             bool     `cli:"metrics-datadog"`
 	MetricsDatadogHost         string   `cli:"metrics-datadog-host"`
+	TracingDatadogAddr         string   `cli:"tracing-datadog-addr"`
```
petemounce, line 87: Is an addr different from a host, above? It sounds like it might be...? It would be great to use consistent terms with consistent content expectations.

petemounce (commented):
This, frankly, looks awesome in intent. Eagerly hoping for a review from someone at BK!

petemounce:
This looks really useful. I have a few questions:
* You mentioned it uses OpenTracing - would you mind elaborating on how that can be supplied over to other products like Jaeger?
    * If so, and if all it needs is an address, could the tracing datadog addr be named something less vendor-specific?
* How're you noticing when hooks start/stop? That was a shortcoming of our own implementation that we used agent-level hooks to work around changing the agent itself.
* Does the trace context propagate over
    * `trigger`'d pipelines? It _looks_ like it...?
    * parallel jobs that are sharded across agents via `command.parallelism`?

clicommand/agent_start.go
goodspark: It could be but chances are it'll be the same. We could just have one but that'd break backwards compatibility with configurations. Or we reuse the metrics Datadog host config, which then is 'incorrect' when read in the tracing code. But neither of these are strictly terrible issues. Open to either or other ideas.

samschlegel: Addr includes a port, where I believe the host shouldn't, though it seems like MetricsDatadogHost does support the port, which is indeed a bit confusing.

Also as far as sharing host between metrics and tracing, it can be different. Most DataDog libraries support specifying these separately, and there are some products that require them to be different, such as Lightstep's DataDog ingest support.

samschlegel, line 87: As a separate comment, I think this should be split up into a `bool` and a `string`, where having it enabled but with the address set to the empty string allows us to fall back on DataDog's default environment variables like DD_AGENT_HOST etc. This means that things will "just" work in environments that are already configured to use those, such as things running in a K8s environment

tracetools/propagate.go
```diff
@@ -0,0 +1,96 @@
...
+// so that it can used as an environment variable.
+//
+// Why the carrier data into a single string? Since carrier data is just key-value
+// pairs, they could be directly added as env vars. However, when propagating the
+// env vars to new processes, users will have to 1) know the list of carrier keys
+// and 2) iterate and push them through. Since carrier keys are a bit of an
+// implementation detail, they can change easily. So containing it all in one well
+// known env var makes things simpler.
+//
+// TODO: fallback to BUILD_ID/JOB_ID?
```
samschlegel, line 28: This TODO made sense when we were using ddtrace in python, but in Go the trace_id and span_id have to be int64s, which isn't compatible with `BUILD_ID/JOB_ID` since they're 128-bit UUIDs. Should be fine either way as long as people have searching by tags available to them in their tracing solution.

bootstrap/bootstrap.go
```diff
@@ -173,8 +192,89 @@ func (b *Bootstrap) Cancel() error {
...
+
+	buildID, _ := b.shell.Env.Get("BUILDKITE_BUILD_ID")
+	source, _ := b.shell.Env.Get("BUILDKITE_SOURCE")
+	label, hasLabel := b.shell.Env.Get("BUILDKITE_LABEL")
+	if !hasLabel {
+		label = "job"
+	}
+
+	// Set specific tracing library here. Everything else should be using opentracing.
+	// Use a constant sampler - CI runs aren't high traffic.
```
samschlegel, line 229: It's also worth mentioning that because of how the datadog agent works, this still won't guarantee all spans are sent to APM. Datadog uses a priority-based sampler that rate limits based on the response of the agent after it pushes a batch in case the agent is struggling to keep up.

If you actually want to make sure that it it is kept you need to add
```go
tracer.WithGlobalTag(ext.SamplingPriority, ext.PriorityUserKeep)
```

bootstrap/shell/shell.go
petemounce: Oh - yeah, log it, definitely. Sorry, could have been clearer!

petemounce:
> * Opentracing: So to plug in other tracing libraries, people would need to make code changes. Specifically in
>   
>   1. Bootstrap.startTracing (pass opentracing.SetGlobalTracer an object that meets the opentracing.Tracer interface) and
>   2. ~(if you want distributed tracing working) define a propagator that's compatible with the TextMap format, accepts a `map[string]string` as the carrier, and configure your specific tracer to use that propagator (as seen in the Datadog-specific setup code in Bootstrap.startTracing).~ This bit is no longer necessary!
>   
>   In fact, if you look at the first commit, you can see that's how we started! But we're moving to use Datadog more so the implementation was changed.
>   We can definitely make the tracing host less vendor-specific, but we'll need another flag to indicate the tracing implementation so that the specific tracing library can be pulled. Very similar to the metrics Datadog env vars I suppose.

I'd very much like to be able to plug in opentracing to publish into jaeger. We currently (still!) do this by writing out JSON blobs when we see particular patterns in the build log, which we put there via agent-level hooks (to capture the coarse-grained step and hook-level spans fleet-wide) and via build-automation (the user-level finer-grained spans). We then pair those start/end blobs up at the end of the step and write them into elasticsearch, which jaeger-UI is configured to be reading from (we don't currently run the rest of jaeger, since this approach seemed simpler and we weren't sure people were going to use it at the time - they definitely do though, especially people with long-running or unstable builds!).

> * Hook start/stop: This is caught directly in the agent code itself. To make the code clearer, I pulled out each hook phase's lookup and execution code into a function. Then it's easier to setup spans and error tracking. This is the code that runs the global, agent, and (if applicable) plugin hooks for a phase.

Ah, nice :)

> * Trace propagation: Yup! This is why we built the custom EnvVarPropagator (which unfortunately is Datadog specific at the moment) and configured Datadog's tracer to use that instead of its defaults. ~I have not tested with `command.parallelism` but I can do that now. Will update this comment on results.~ Yup this also works with `parallelism` steps! However, there's no data in the spans or tags that indicate which parallel job run a span belongs to.

Is there no data in the spans/tags because the `BUILDKITE_PARALLEL_JOB` isn't noticed, or some other reason? We don't at the moment, but will soon do the same thing with our tracing as we do with our logging - add pretty much all the `BUILDKITE_*` environment variables onto the trace context key/value bucket.

That makes me think - it would be nice to see the step's `depends_on` reflected into there. The tracing UI that jaeger offers is a standard time-based waterfall, not a DAG, and so one has to flip between the BK /dag and jaeger to trace the critical path; that's a bit of a rough edge. If the dependency information can be reflected into tracing and then used, that would be really nice.

clicommand/agent_start.go
goodspark: I turned this into `tracing-backend`. By default an empty string will mean no tracing. Setting it to 'datadog' will enable it. This lets people in the future add support for other tracing libs, like jaeger. And specific tracing lib configuration will be done through the common env vars like you mentioned. Verified this works on my local setup.

goodspark: I think this is largely resolved with the new approach (no BK-specific configuration env vars - just a backend env var and then use the tracing libs' common config env vars like DD_AGENT_HOST).

petemounce: Thanks very much!

goodspark:
Hi Buildkite team,

Is there anything I can do to help get this reviewed?

Thanks!

jayco:
@goodspark this is amazing, thank you for all the work you have put into this! I have shared this with the team cc/ @chloeruka and @pda for a review! 👌

keithpitt:
This is very cool. I'm gonna aim to get this merged in next week! 👍🏻

chloeruka:
@goodspark first of all, awesome work! Thanks so much, we're really excited to test this and roll it out. I did have a question you might be able to answer; how hard would it be to migrate this to [OpenTelemetry](http://opentelemetry.io)? Is it something that can be done soon or is there still some missing piece that we'd be waiting on? It sounds like OpenTelemetry is still pretty new, and I'm wondering how the APIs are coming along on the go/Datadog side.

goodspark:
I can definitely look into it. @samschlegel and I were talking about that a while ago too. I'll spend a day on it and if it's too much, I'll report back here.

samschlegel:
AFAICT there's still no DataDog Trace exporter, which is what blocked us from using it in the first place :(

goodspark:
Yeah I've been looking around for code or examples. Looks like [Datadog only has Ruby, Python, and JS support](https://opentelemetry.io/registry/?s=datadog). But once they have Golang support, will be happy to help code/review the move to OpenTelemetry.

chloeruka:
> AFAICT there's still no DataDog Trace exporter, which is what blocked us from using it in the first place :(

Bummer! But that makes me feel way better about advancing with the OpenTracing side for now.

> Yeah I've been looking around for code or examples. Looks like [Datadog only has Ruby, Python, and JS support](https://opentelemetry.io/registry/?s=datadog). But once they have Golang support, will be happy to help code/review the move to OpenTelemetry.

Thanks! We'll keep an eye on things as that develops.

chrisfowles:
Hey @goodspark  - I was chatting about this PR in the #agent-dev channel on the buildkite slack with @chloeruka and it was mentioned that it'd be great if the merge conflicts could be fixed, as they should be able to start testing this out.

let me know if you want a hand with this as I'm pretty eager for this

goodspark:
Sure. Done.
@chrisfowles @chloeruka 
Let me know if there's any way I can help the review process.

goodspark:
I also added a commit to add a URL to each job in the tags, since I found that would've been helpful when debugged some pipelines with several jobs in the past.

bootstrap/bootstrap_test.go
```diff
@@ -41,3 +44,36 @@ func TestGetValuesToRedact(t *testing.T) {
...
+func TestStartTracing(t *testing.T) {
+	oriCtx := context.Background()
+	var err error
+
+	// When there's no Datadog tracing address, the tracer should be a no-op.
+	cfg := Config{}
+	b := New(Config{})
+	b.shell, err = shell.NewWithContext(oriCtx)
+	if err != nil {
+		assert.FailNow(t, "Unexpected error while createing shell: %v", err)
```
chloeruka, line 57:
```suggestion
		assert.FailNow(t, "Unexpected error while creating shell: %v", err)
```

chloeruka, line 57:
```suggestion
		assert.FailNow(t, "Unexpected error while creating shell: %v", err)
```

bootstrap/bootstrap_test.go
```diff
@@ -41,3 +44,36 @@ func TestGetValuesToRedact(t *testing.T) {
...
+	assert.Equal(t, span, opentracing.SpanFromContext(ctx))
+
+	// With the Datadog tracing backend, the global tracer should be from Datadog.
+	cfg = Config{
+		TracingBackend: "datadog",
+	}
+	b = New(cfg)
+	b.shell, err = shell.NewWithContext(oriCtx)
+	if err != nil {
+		assert.FailNow(t, "Unexpected error while createing shell: %v", err)
```
chloeruka, line 72:
```suggestion
		assert.FailNow(t, "Unexpected error while creating shell: %v", err)
```

go.mod
```diff
@@ -4,7 +4,7 @@ go 1.14
 
 require (
 	cloud.google.com/go v0.0.0-20170217213217-65216237311a
-	github.com/DataDog/datadog-go v0.0.0-20180822151419-281ae9f2d895
+	github.com/DataDog/datadog-go v3.7.2+incompatible
```
chloeruka, line 7: I'd love to fix this, but it looks like we'll have to wait for https://github.com/DataDog/datadog-go/pull/150 so it isn't a blocker.

chloeruka (approved):
This is looking good! I like your approach and appreciate the additional tests. I've tried to be thorough, but it is a large PR. I'm happy to merge this and promote it to our edge release for further testing where we'll dogfood it internally for a while. For watchers/interested folks, I'll circle back with how to opt in to this via the elastic stack when this is available on edge.

clicommand/bootstrap.go
```diff
@@ -148,6 +150,12 @@ var BootstrapCommand = cli.Command{
 			Usage:  "The name of the agent running the job",
 			EnvVar: "BUILDKITE_AGENT_NAME",
 		},
+		cli.StringFlag{
+			Name:   "queue",
+			Value:  "",
+			Usage:  "The name of the queue the agent belongs to, if tagged",
+			EnvVar: "BUILDKITE_AGENT_META_DATA_QUEUE",
```
chloeruka, line 157: FYI to other watchers that this is configured through meta data [passed to the agent at runtime](https://buildkite.com/docs/pipelines/environment-variables#bk-env-vars-buildkite-agent-meta-data-star). GitHub seems to have gobbled my previous comment, but the gist is that this is typically set on the agent using `BUILDKITE_AGENT_TAGS`. It might be more reliable to pull from that value instead as it doesn't rely on a running job. It's worth being aware that you can configure multiple queues, I don't remember exactly how it would look but I suspect it's a comma separated value. Given that; it might also be interesting to try and record a value to capture what the job agent queue query was, but it would require a change on the Buildkite API side that I can organise.

go.mod
chrisfowles: As a customer, I've pinged the maintainers of this on their slack to get some movement on this PR.

chloeruka:
> For watchers/interested folks, I'll circle back with how to opt in to this via the elastic stack when this is available on edge.

Sorry about the delay; had to shave a few yaks. This is now available as an experimental edge release, you can opt into using it by setting the `BuildkiteAgentRelease` in the [Elastic CI stack for AWS](https://github.com/buildkite/elastic-ci-stack-for-aws) to "edge"

OR manually install by downloading the v3.27.0 experimental release:
- https://download.buildkite.com/agent/experimental/3.27.0.3715/buildkite-agent-linux-386
- https://download.buildkite.com/agent/experimental/3.27.0.3715/buildkite-agent-linux-amd64
- https://download.buildkite.com/agent/experimental/3.27.0.3715/buildkite-agent-linux-arm
- https://download.buildkite.com/agent/experimental/3.27.0.3715/buildkite-agent-linux-arm64

bungfut:
#1491 
Bung

bootstrap/bootstrap_test.go
bungfut: [firebase-gtk-web-start-rclyjo.zip](https://github.com/buildkite/agent/files/7075796/firebase-gtk-web-start-rclyjo.zip)

bootstrap/bootstrap_test.go
bungfut: #1273 
[usage_2021-06-17_2021-08-16.csv](https://github.com/buildkite/agent/files/7075803/usage_2021-06-17_2021-08-16.csv)
_

PR #1057, opened 2019-07-31
Heartbeat hooks

sj26:
This is a terrible implementation of an interesting idea for discussion — what if we let folks run hook scripts on certain agent-level events, like heartbeats?

Often users want to advertise an agent's health to an outer layer, like a host, or a monitoring platform. At the moment they can see that the agent process is alive, but that's about it. This sometimes isn't enough — the agent process is resilient and continues if it can't contact the API until it reconnects.

What if they could also tell when an agent is heartbeating — that it can contact the Buildkite API and can run jobs? With a heartbeat hook script they can radiate health events and notice if those events stop firing and investigate or take automated action if needed.

I'd also love to see an "accept" hook which allows a hook script to reject the job instead, for example.

PR #1275, opened 2020-08-27
Add an agent-wide shutdown hook

goodspark:
This will be useful for ensuring graceful shutdowns and cleanup when the
agent is integrated with other systems, like k8s.

This only runs once per agent, regardless of the number of workers. It looks for a `shutdown` script in the hooks directory on the agent.

Question:
- Does anything special need to be done for Windows?

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/buildkite/agent/v3/agent"
 	"github.com/buildkite/agent/v3/api"
+	"github.com/buildkite/agent/v3/bootstrap/shell"
 	"github.com/buildkite/agent/v3/cliconfig"
 	"github.com/buildkite/agent/v3/experiments"
 	"github.com/buildkite/agent/v3/logger"
@@ -720,6 +721,9 @@ var AgentStartCommand = cli.Command{
 		// Setup the agent pool that spawns agent workers
 		pool := agent.NewAgentPool(workers)
 
+		// Agent-wide shutdown hook. Once per agent, for all workers on the agent.
+		defer func() { shutdownHook(l, cfg.HooksPath) }()
+
 		// Handle process signals
 		signals := handlePoolSignals(l, pool)
 		defer close(signals)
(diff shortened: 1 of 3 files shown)


lox:
This is neat, thanks @goodspark. My only reservation with this idea is that it's the first hook we have where the output won't be user visible. What happens if there is a fatal error in the script for instance I wonder? Where does stdout/stderr go?

goodspark:
Since this is an agent-level thing happening during shutdown, I assumed logs are probably more of a concern for the people who own and manage the Buildkite agents, not the CICD users (if that's what you mean by user visible). Because by then, there are no more jobs running (except if the agent is being force killed because it took too long to stop after a SIGTERM).

As for where stdout/err go - the hook uses shell.RunScript, which pipes stderr to stdout and sends stdout to os.Stdout.

Usually the people who manage the Buildkite agents will have existing logging solutions that collect the output from the agent process (be it on-box log collection, sidecars in a container orchestration system, etc).

lox:
Yup, I agree @goodspark. I have no objections to merging this. @pda?

pda:
Verifying locally and merging now.

pda:
The output logging vibe is a bit strange:

image

But, at least it's visible.

I guess we could pipe it into the agent's main logger, so it gets some sort of context prefix and timestamps etc? I'm not sure if that's better or not.

clicommand/agent_start.go
```diff
@@ -720,6 +721,9 @@ var AgentStartCommand = cli.Command{
 		// Setup the agent pool that spawns agent workers
 		pool := agent.NewAgentPool(workers)
 
+		// Agent-wide shutdown hook. Once per agent, for all workers on the agent.
+		defer func() { shutdownHook(l, cfg.HooksPath) }()
```
pda, line 725: Could we drop the `func` wrapper here?
I don't think the logger nor the HooksPath could change between `defer` and the call.

```suggestion
		defer shutdownHook(l, cfg.HooksPath)
```

pda (approved):

sj26:
This is cool!

If there is precedent for agent hooks now, I'd love to get #1057 in as well!

pda:
> Question:
> Does anything special need to be done for Windows?

Hmm… good question that I forgot about until merging :)

Most importantly — as long as `filepath.Join(hooksPath, "shutdown")` doesn't exist, it's not going to do any harm.

But when that file does exist, it doesn't seem to work in Windows… here's a test with a `C:\buildkite-agent\hooks\shutdown` file present containing `echo hello`:
image

pda:
I'm looking into extracting a new `hook` package from `bootstrap` so it can be used here too.

pda:
I wonder if this (these?) hooks should have a prefix, like `agent-shutdown` and potentially a future `agent-heartbeat` to distinguish them from job hooks…?

goodspark:
An agent prefix sounds like a nice self-documenting standard for the hook filenames.

Hooking the hook into standard logging is nice too.

pda:
I'll open a PR for an `agent-` prefix on the hook name.

https://github.com/buildkite/agent/pull/1326 is ready for feedback/review for piping the output to the logger.

PR #1276, opened 2020-08-28
Add optional brackets to file arg

harrietgrace:
As noted in https://github.com/buildkite/docs/pull/806 - the `file` arg is optional, and so should be wrapped in square brackets in the help text.

### DIFF
--- a/clicommand/pipeline_upload.go
+++ b/clicommand/pipeline_upload.go
@@ -21,7 +21,7 @@ import (
 
 var PipelineUploadHelpDescription = `Usage:
 
-   buildkite-agent pipeline upload [<file>] [arguments...]
+   buildkite-agent pipeline upload [file] [arguments...]
 
 Description:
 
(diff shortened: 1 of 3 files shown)


clicommand/pipeline_upload.go
```diff
@@ -21,7 +21,7 @@ import (
 
 var PipelineUploadHelpDescription = `Usage:
 
-   buildkite-agent pipeline upload <file> [arguments...]
+   buildkite-agent pipeline upload [<file>] [arguments...]
```
pda, line 24: I think the square brackets should replace the pointy ones? Rather than being additional.

```suggestion
   buildkite-agent pipeline upload [file] [arguments...]
```

pda: There's a couple of others with both which I guess should also be fixed:

```
pda@paulbookpro ~/bk/agent ❯ rg '\[<'
clicommand/meta_data_set.go
16:   buildkite-agent meta-data set <key> [<value>] [arguments...]

clicommand/annotate.go
18:   buildkite-agent annotate [<body>] [arguments...]
```

harrietgrace: Yeah I was wondering about those brackets! Cool, I'll update all those places.

pda (approved):
👌🏼

PR #1277, opened 2020-08-31
Fix: typo, resolve composite literal warn

chloeruka:
PR to address some small errors I noticed while spelunking the artifacts commands:

- Typo corrected `theseArtiacts` -> `theseArtifacts`
- Fix `[go-vet] api.ArtifactBatch composite literal uses unkeyed fields` by keying all the struct fields in use.


sj26 (approved):
Thanks for keeping the campsite clean!

Looks like unrelated broken windows tests?

chloeruka:
> Looks like unrelated broken windows tests?

_Seems_ like a transient error, using a different windows agent resolved it. If I find time I'll look into why.

PR #1288, opened 2020-09-09
artifact_uploader_test fixed for Windows.

pda:
There was a bug in the artifact_uploader_test  test code only, which caused it to fail in some windows scenarios.

The test was deriving the codebase root directory in a way that used OS-native path separators (back-slashes on Windows), and then forward-slashed components were added to the end.

This would not have been noticed if the codebase was in a top-level directory like `C:\buildkite-agent`.

I imagine it crept in during #1211 /cc @ticky 

Previously:

    C:\pda\agent>go test ./agent -run TestCollect
    --- FAIL: TestCollect (0.01s)
        artifact_uploader_test.go:31: C:\pda\agent\agent
        --- FAIL: TestCollect/Smile.gif#01 (0.00s)
            artifact_uploader_test.go:140:
                    Error Trace:    artifact_uploader_test.go:140
                    Error:          Not equal:
                                    expected: "pda\\agent/test/fixtures/artifacts/gifs/Smile.gif"
                                    actual  : "pda/agent/test/fixtures/artifacts/gifs/Smile.gif"

                                    Diff:
                                    --- Expected
                                    +++ Actual
                                    @@ -1 +1 @@
                                    -pda\agent/test/fixtures/artifacts/gifs/Smile.gif
                                    +pda/agent/test/fixtures/artifacts/gifs/Smile.gif
                    Test:           TestCollect/Smile.gif#01
    FAIL
    FAIL    github.com/buildkite/agent/v3/agent     0.457s
    FAIL

### DIFF
--- a/agent/artifact_uploader_test.go
+++ b/agent/artifact_uploader_test.go
@@ -3,7 +3,6 @@ package agent
 import (
 	"fmt"
 	"os"
-	"path"
 	"path/filepath"
 	"strings"
 	"testing"
@@ -112,7 +111,8 @@ func TestCollect(t *testing.T) {
 
 	experiments.Disable(`normalised-upload-paths`)
 
-	// These test cases use filepath.Join, which is the behaviour without normalised-upload-paths
+	// These test cases use filepath.Join, which uses per-OS path separators;
+	// this is the behaviour without normalised-upload-paths.
 	for _, tc := range testCases {
 		t.Run(tc.Name, func(t *testing.T) {
 			a := findArtifact(artifactsWithoutExperimentEnabled, tc.Name)
@@ -128,7 +128,8 @@ func TestCollect(t *testing.T) {
 		})
 	}
 
-	// These test cases use path.Join, which is the behaviour without normalised-upload-paths
+	// These test cases uses filepath.ToSlash(), which always emits forward-slashes.
+	// this is the behaviour with normalised-upload-paths.
 	for _, tc := range testCases {
 		t.Run(tc.Name, func(t *testing.T) {
 			a := findArtifact(artifactsWithExperimentEnabled, tc.Name)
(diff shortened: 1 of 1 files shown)


yob:
why was this failing intermittently and not 100% of the time?

pda:
I don't _think_ this one was intermittent. I think it was passing when the codebase was a top-level directory, and failing when it wasn't. (And only on Windows).

(There's another intermittent Windows test failure lurking in there still… maybe Windows & Linux but more frequent on Windows? Not sure. This one just came up as a tangent to chasing that one.)

ticky:
Ahh, darn, I thought I’d gotten all these gotchas ironed out in my testing. Nice catch!

yob:
> I don't think this one was intermittent. I think it was passing when the codebase was a top-level directory, and failing when it wasn't. (And only on Windows).

Oh, so it was passing 100% of the time in CI (because it used a top-level directory) and failed 100% of the time when you started testing on windows (because you used a nested directory)?

pda:
> so it was passing 100% of the time in CI (because it used a top-level directory)

Maybe? Although it looks like maybe it's not running in a top-level directory:
https://github.com/buildkite/agent/blob/bdbc54645c2ecab070a39c0f7950b273be9b3f1d/.buildkite/Dockerfile-windows#L18

I have no idea why it's been passing in CI 🤷‍♂️

> failed 100% of the time when you started testing on windows (because you used a nested directory)?

Yep! `git clone` into `C:\pda\agent` and then `go test ./...` was failing on this 100% of the time.

pda:
Actually maybe it is in a top-level dir:
https://github.com/buildkite/agent/blob/bdbc54645c2ecab070a39c0f7950b273be9b3f1d/.buildkite/docker-compose.windows.yml#L10-L12

yob (approved):
I haven't run this change myself on windows (really must setup a VM), but this is well explained an makes sense :+1:

PR #1289, opened 2020-09-09
Improve Apple Silicon Mac support

ticky:
Following on from #1237, with new discoveries!

Turns out that in many cases, Apple Silicon Macs will in fact run Bash in amd64 mode, which in turn means the reported architecture is amd64, even though it's an Apple Silicon Mac. Homebrew has worked around this in Homebrew/brew#7995, and we more or less follow their lead here.

This makes it so that we _always_ show our little notice on Apple Silicon Macs, whether their shell is in amd64 or arm64 mode.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -48,7 +48,7 @@ function buildkite-cpu-arm64 {
 function buildkite-apple-silicon-check {
   if buildkite-cpu-arm64; then
     ARCH="amd64"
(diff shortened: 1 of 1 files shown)


ticky:
Confirmed that this works correctly in both modes on Apple Silicon Macs, and that it still runs normally on amd64 Macs.

sj26 (approved):
Nice! 🙌

yob (approved):
I don't have a mac with apple silicon to test this on, but it makes sense to me :+1:

pda:
Wow, nice work @ticky 🕵️‍♀️

PR #1290, opened 2020-09-09
Add a Homebrew-based installation method for macOS hosts

### LINKED ISSUE #1284: Token not entered into config when using brew install
The docs say that running the following should enter the token into the Buildkite config file when installing.

```
brew install --token='INSERT-YOUR-AGENT-TOKEN-HERE' buildkite-agent
```

However, the token in my config file is empty and so I have to update it manually.
```
# The token from your Buildkite "Agents" page
token=""
```

I'm using Homebrew 2.4.16 on MacOS 10.15.6.

ticky:
Given that Homebrew no longer supports formula flags with values in them (see Homebrew/brew#7638, buildkite/homebrew-buildkite#18, buildkite/homebrew-buildkite#15), this updates the Unix `install.sh` to allow the use of the Homebrew install method, while also providing one-line token configuration.

To do so, I've divided up the install process into a few more functions, and shared as much of the logic as possible between them.

The original `curl`/`wget`-based installation process is kept for serving other platforms, and is also used if `BETA` is set to `true`, if Homebrew is not found, or if a `DESTINATION` was explicitly specified.

This also currently means that if Homebrew is present and used, our releases endpoint is currently not fetched, so we don't get any statistics. Perhaps we would want this anyway? I guess that's a decision we'd have to make.

I have tested this on macOS, with Homebrew removed from the path, with `DESTINATION` overridden, and with `BETA` set.

This also builds on #1289, as both rearrange a bit of infrastructure. Along with buildkite/homebrew-buildkite#18, will fix #1284. Closes buildkite/buildkite#5523.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -18,8 +18,6 @@ echo -e "\033[33m
                                                 __/ |
                                                |___/\033[0m"
 
-echo -e "Finding latest release..."
-
 SYSTEM=$(uname -s | awk '{print tolower($0)}')
 MACHINE=$(uname -m | awk '{print tolower($0)}')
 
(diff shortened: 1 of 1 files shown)


sj26 (approved):
This is great! 🙌

ticky:
Moving this back to draft as we figure out what we want to replace the existing one-liner with. ⏸

PR #1293, opened 2020-09-10
Expand the CLI test step

ticky:
This adds more virtual Buildkite steps to the "bk cli" test. Initially intended for testing the bk CLI against the Agent, this is also a handy way for us to check we haven't broken any Agent CLI arguments.

So, this adds a test which uses the meta-data subcommand with a default set, and a suite of tests which exercise the annotate subcommand.

It also renames it so it’s a little more visible in the timeline, as I completely missed that this step even existed! 😅

### DIFF
--- a/.buildkite/pipeline.bk-test.yml
+++ b/.buildkite/pipeline.bk-test.yml
@@ -15,6 +15,25 @@ steps:
       [[ "\${LOCAL_ENV?}" == "alpacas" ]]
       [[ "\${ENV_ORDER?}" == "local" ]]
 
+  - label: "add annotation"
+    command: |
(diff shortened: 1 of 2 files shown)


lox:
This seems like a pragmatic way to extend some end-to-end testing coverage between the agent and the API it's talking to. I can't decide if it's a bad idea that it's relying on `bk` for that API. I wonder if we might be better off using https://golang.org/pkg/net/http/httptest/ to mock the agent API for end-to-end tests. I guess that gives us yet another thing we need to update when server-side things happen.

Overall, 👍🏻

lox (approved):

PR #1295, opened 2020-09-11
Fix flaky tests: experiments and parallel tests don't mix

pda:
Some tests have been mysteriously flaky (particularly on Windows, but that may be a red herring), failing without an error message. Somehow, https://github.com/buildkite/agent/pull/1294 has revealed the error messages:

```
=== Failed
  | === FAIL: bootstrap/integration TestCheckingOutLocalGitProject (7.00s)
  | mock.go:248: Unexpected call to git "rev-parse", "HEAD"
  | mock.go:255: More invocations than expected (1 vs 7)
  |  
  | === FAIL: bootstrap/integration TestCheckingOutShallowCloneOfLocalGitProject (7.69s)
  | mock.go:248: Unexpected call to git "rev-parse", "HEAD"
  | mock.go:255: More invocations than expected (1 vs 7)
```

I believe this is being caused by `TestWithResolvingCommitExperiment` in `bootstrap/integration/checkout_integration_test.go` running with `t.Parallel()` and then temporarily enabling the `resolve-commit-after-checkout` experiment. That experiment causes an extra `git rev-parse $BUILDKITE_COMMIT` command to run, which messes up the mock expectations of tests running in parallel, which are written to assume that experiment is turned off.

Also, `TestCollect` in `agent/artifact_uploader_test.go` was also enabling an experiment, but ~it was not running in parallel~, and it was disabling the experiment afterwards, so I think it was not causing any problems. But, this PR makes the experiment revert slightly more robust using `defer` and avoiding the (currently safe) assumption that the experiment should be disabled by default. _(edit: it was running in parallel. but it wasn't causing the test failures.)_

### DIFF
--- a/agent/artifact_uploader_test.go
+++ b/agent/artifact_uploader_test.go
@@ -95,6 +95,15 @@ func TestCollect(t *testing.T) {
 	// path.Join function instead (which uses Unix/URI-style path separators,
 	// regardless of platform)
 
+	experimentKey := "normalised-upload-paths"
+	experimentPrev := experiments.IsEnabled(experimentKey)
+	defer func() {
+		if experimentPrev {
+			experiments.Enable(experimentKey)
+		} else {
+			experiments.Disable(experimentKey)
+		}
+	}()
 	experiments.Disable(`normalised-upload-paths`)
 	artifactsWithoutExperimentEnabled, err := uploader.Collect()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


ticky (approved):
Ohhh, good catch!! Completely makes sense.

yob:
neat, thanks for getting to the bottom of this!

PR #1298, opened 2020-09-12
Remove unused dependency checksum in go.sum

JuanitoFatas:
The result of ran: `$ go mod tidy`

We are using urfave/cli v1.22.4

[go mod tidy document](https://golang.org/cmd/go/#hdr-Add_missing_and_remove_unused_modules)


pda:
`go.sum` is a funny thing. The key is to not think of it like a `Gemfile.lock`.

> In part because go.sum is not a lock file, it retains cryptographic checksums for module versions even after you stop using a module or particular module version. This allows validation of the checksums if you later resume using something, which provides additional safety.

— https://github.com/golang/go/wiki/Modules#is-gosum-a-lock-file-why-does-gosum-include-information-for-module-versions-i-am-no-longer-using

There's generally no reason to `go mod tidy` unless things get weird.

PR #1300, opened 2020-09-14
Fix build script

goodspark:
The `BUILD_VERSION` arg was not being properly set because the
package/import path changed with the v3 agent.

Without:

```
scripts/build-binary.sh linux amd64 1
pkg/buildkite-agent-linux-amd64 --version
buildkite-agent version 3.23.1, build x
```

With:

```
scripts/build-binary.sh linux amd64 1
pkg/buildkite-agent-linux-amd64 --version
buildkite-agent version 3.23.1, build 1
```

### DIFF
--- a/scripts/build-binary.sh
+++ b/scripts/build-binary.sh
@@ -40,7 +40,7 @@ fi
 export CGO_ENABLED=0
 
 mkdir -p $BUILD_PATH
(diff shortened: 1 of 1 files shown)


pda (approved):

pda:
Thanks @goodspark — I noticed that `x` floating around, nice to see it found & fixed.

Issue #1269, opened 2020-08-19
Git checkout failed at: argument list too long


alexzhaobin:
When we squash merge a large branch in github, it will generate a huge commit message like this with 2000 lines:

```
Commit message (#2297)

* Bugfix/3007 .............................

* Bugfix/1210 .............................

* Bugfix/2783 .............................

* Bugfix/2118 .............................
```

This causes buildkite agent to fail at git checkout with error: fork/exec /usr/local/bin/buildkite-agent: argument list too long

Full error message:
Error: Error running `/usr/local/bin/buildkite-agent meta-data set buildkite:git:commit "commit ..."`: fork/exec /usr/local/bin/buildkite-agent: argument list too long

sj26:
Oh no, sorry that's probably just not a case we've seen very often — but a bug!

We can modify this code to pipe the commit message into the meta-data command instead. I'll see if we can pop up a quick fix.

pda:
Thanks again for the report @alexzhaobin — we have one fix merged, another ready for review, and it should land in a release within the next week. /cc @jayco

pda:
Hey @alexzhaobin — [buildkite/agent v3.24.0](https://github.com/buildkite/agent/releases/tag/v3.24.0) contains #1302 and #1307 which should fix up the handling of very large commit messages. Let us know how you go with it 🙏🏼

PR #1302, opened 2020-09-16
buildkite:git:commit meta-data via stdin; avoid Argument List Too Long

pda:
This builds on @sj26's [great investigation](https://github.com/buildkite/agent/pull/1270) identifying that `buildkite-agent meta-data set buildkite:git:commit` should take the commit message via stdin, not argv. Otherwise a very large commit message may exceed the system argv limit,
resulting in: `fork/exec /usr/local/bin/buildkite-agent: argument list too long`.

The key to this approach is `func (s *Shell) WithStdin(r io.Reader) *Shell` — it avoids threading our new `stdin` through multiple layers and variants of `Run()` etc.

The initial “package process supports Stdin” and “bootstrap/shell.Shell.WithStdin() to pass stdin to commands.” commits in this PR are self-contained with a test each, which I think demonstrates there's no blocking/closing issue with passing the `io.Reader` to our `process` package.

I haven't tested this on a real build yet, and there's heaps of bintest mock expectations in other tests that this breaks.

Closes #1270

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -1278,13 +1278,12 @@ func (b *Bootstrap) defaultCheckoutPhase() error {
 	b.shell.Commentf("Checking to see if Git data needs to be sent to Buildkite")
 	if err := b.shell.Run("buildkite-agent", "meta-data", "exists", "buildkite:git:commit"); err != nil {
 		b.shell.Commentf("Sending Git commit information back to Buildkite")
-
-		gitCommitOutput, err := b.shell.RunAndCapture("git", "--no-pager", "show", "HEAD", "-s", "--format=fuller", "--no-color", "--")
+		out, err := b.shell.RunAndCapture("git", "--no-pager", "show", "HEAD", "-s", "--format=fuller", "--no-color", "--")
 		if err != nil {
 			return err
 		}
-
-		if err = b.shell.Run("buildkite-agent", "meta-data", "set", "buildkite:git:commit", gitCommitOutput); err != nil {
+		stdin := strings.NewReader(out)
+		if err := b.shell.WithStdin(stdin).Run("buildkite-agent", "meta-data", "set", "buildkite:git:commit"); err != nil {
 			return err
 		}
 	}
--- a/bootstrap/shell/shell.go
+++ b/bootstrap/shell/shell.go
@@ -40,6 +40,12 @@ type Shell struct {
 	// Whether the shell is a PTY
 	PTY bool
 
+	// stdin is an optional input stream used by Run() and friends.
+	// It remains unexported on the assumption that it's not useful except via
+	// WithStdin() to get a shell-copy prepared for a single command that needs
+	// input.
+	stdin io.Reader
+
 	// Where stdout is written, defaults to os.Stdout
 	Writer io.Writer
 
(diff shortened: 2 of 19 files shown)


pda:
I think https://godoc.org/github.com/buildkite/bintest#Expectation doesn't currently support expectations on stdin… I'm hoping to get that added today… might be optimistic though. It's required to fix all the failing integration tests without discarding the assertion completely.

sj26:
This looks great, @pda! Let me check it out and try it locally to confirm the fix.

pda:
So it does solve the _stated_ problem in local testing…

Before:
image

After:
image

But, as a result, uploaded steps end up with enormous `BUILDKITE_MESSAGE` in their env, which triggers this in the subsequent jobs:
image

> fork/exec ./buildkite-agent: argument list too long

I was stumped by the “argument list” part, but apparently that includes ENV.

So, I think we'll have to truncate the env somewhere.

sj26:
Yeah we can probably truncate the environment variable value in buildkite.

sj26 (approved):
This is great, thanks for taking my hacky attempt and doing it properly @pda! 🙇

process/process.go
```diff
@@ -64,6 +64,7 @@ type Config struct {
 	Path            string
 	Args            []string
 	Env             []string
+	Stdin           io.Reader
```
pda, line 67: Here's the lowest layer learning to take Stdin — our `process.Config`.
The zero-value of io.Reader is nil, so when it's not passed it's the same as the old `p.command.Stdin = nil` code.

bootstrap/shell/shell.go
```diff
@@ -84,6 +90,24 @@ func NewWithContext(ctx context.Context) (*Shell, error) {
...
+	// Can't copy struct like `newsh := *s` because sync.Mutex can't be copied.
+	return &Shell{
+		Logger: s.Logger,
+		Env:    s.Env,
+		stdin:  r, // our new stdin
+		Writer: s.Writer,
+		wd:     s.wd,
+		ctx:    s.ctx,
+	}
+}
```
pda, line 109: Here's the second-lowest layer learning to take Stdin — our Shell implementation.
The `stdin` eventually gets passed to the `process.Config`.

bootstrap/bootstrap.go
```diff
@@ -1278,13 +1278,12 @@ func (b *Bootstrap) defaultCheckoutPhase() error {
...
-
-		gitCommitOutput, err := b.shell.RunAndCapture("git", "--no-pager", "show", "HEAD", "-s", "--format=fuller", "--no-color", "--")
+		out, err := b.shell.RunAndCapture("git", "--no-pager", "show", "HEAD", "-s", "--format=fuller", "--no-color", "--")
 		if err != nil {
 			return err
 		}
-
-		if err = b.shell.Run("buildkite-agent", "meta-data", "set", "buildkite:git:commit", gitCommitOutput); err != nil {
+		stdin := strings.NewReader(out)
+		if err := b.shell.WithStdin(stdin).Run("buildkite-agent", "meta-data", "set", "buildkite:git:commit"); err != nil {
```
pda, line 1286: Here's the actual change built on the two layers beneath it.
The output of `git show …` is passed as stdin instead of an argument to `buildkite-agent meta-data set`.

pda (commented):
(just some PR annotation to make the core of this patch clear)

pda:
This fixed #1269 but uncovered another large commit message bottleneck [mentioned above](https://github.com/buildkite/agent/pull/1302#issuecomment-694732003) and addressed in #1307

PR #1308, opened 2020-09-24
Handle pull requests with git-mirrors

frankh:
This fixes a regression that was introduced in https://github.com/buildkite/agent/pull/1112 which means git-mirrors doesn't work with PRs from forks

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -1082,9 +1082,17 @@ func (b *Bootstrap) updateGitMirror() (string, error) {
 		return "", err
 	}
 
-	// Fetch the build branch from the upstream repository into the mirror.
-	if err := b.shell.Run("git", "--git-dir", mirrorDir, "fetch", "origin", b.Branch); err != nil {
-		return "", err
+	if b.PullRequest != "false" && strings.Contains(b.PipelineProvider, "github") {
+		refspec := fmt.Sprintf("refs/pull/%s/head", b.PullRequest)
+		// Fetch the PR head from the upstream repository into the mirror.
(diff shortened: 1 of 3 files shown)


bootstrap/bootstrap.go
```diff
@@ -1049,44 +1049,6 @@ func (b *Bootstrap) updateGitMirror() (string, error) {
 		return mirrorDir, nil
 	}
 
-	// If it exists, immediately release the clone lock
-	mirrorCloneLock.Unlock()
```
frankh, line 1053: no need to immediately unlock as the defer will handle this fine

ticky (approved):
This looks good to me, going to get some more eyes on it, too :)

sj26 (commented):
This looks great!

Can we add some comment output, like the other fetch, to make it obvious which code path is being taken?

pda (approved):

ticky:
@sj26 Added some comment output to the GitHub PR pathway! Feel free to merge if this is good

sj26:
Perfect!

Do the tests pass? Is there a way to add a test for this easily? (Doesn't look they were run here, perhaps turned off for third party builds?)

ticky:
`go test` is weird and not trustworthy for me locally so I’ve forced the issue by duplicating this in #1318

sj26:
The existing tests are passing, nice.

We either need to manually try this out and verify it works, or add an automated test.

ticky:
Took some doing, but I’ve added a spec in bc418f792194bacf3eb18d4f21691afdbfb9b2ef, which I hope covers enough of it to be useful. The rest of the GitHub-specific code is missing coverage, also.

sj26 (approved):
Wonderful, thanks @frankh and @ticky!

PR #1312, opened 2020-09-28
Upgrade docker ubuntu to 20.04 focal

sj26:
Upgrade the ubuntu release we use when packaging the agent docker image. We're currently using the previous LTS release (18.04) and this bumps it to the latest LTS release from earlier this year (20.04).

I built and ran the image locally and it worked fine.

### DIFF
--- a/packaging/docker/ubuntu-linux/Dockerfile
+++ b/packaging/docker/ubuntu-linux/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:18.04
+FROM ubuntu:20.04
 
 ENV DEBIAN_FRONTEND=noninteractive
 


yob:
This is a good idea, however there is a risk that we'll break some folks agent fleets when supporting tools jump in version when they next upgrade the agent (the new libc, jq, awscli, etc that come in 20.04).

I guess there's no avoiding that though, without changing our docker tagging strategy to have an Ubuntu tag per LTS release. That doesn't seem sustainable for our small though. Hopefully most folks are using tags like `3.23-ubuntu` instead of plain `ubuntu` (or `3-ubuntu`), so any surprises will only happen when they intentionally update the tag to the next agent release.

I propose merging this after the upcoming 3.25.0 release. Our alpine docker images are running a fairly old version too - I might open a PR to bump that at the same time.

yob (approved):

PR #1314, opened 2020-10-05
gitFetch() terminates options with -- before repo/refspecs

pda:
When constructing the `git fetch` command, user-specified branch names (refspecs) should never be considered flags/options by `git fetch`. This patch inserts `--` after the valid flags so that subsequent arguments can only be treated as a repository name and a list of refspecs.


lox:
Good catch! I'd expect that this breaks all the expectations in the integration tests 🤔

sj26 (approved):
Nice!

yob:
neat :+1:

PR #1315, opened 2020-10-05
gitCheckout() validates branch, plus unit tests

pda:
When constructing the `git checkout` command, the sometimes-user-specified branch name should never be considered flags/options by `git`. It's not possible to terminate arg parsing with `--` because that has special meaning to `git checkout` — it indicates the following names are files, not refs. So, the ref must be validated. There's a [`git check-ref-format`](https://git-scm.com/docs/git-check-ref-format) subcommand which roughly does that. Rather than shelling out to git, this patch introduces a simple `gitCheckRefFormat(ref)` implementation to verify the ref doesn't start with a hyphen. This could be replaced with shell-out or a more thorough implementation if that feels worth doing.

With branch set to `--foo`…
* Before: `git checkout -f --foo` → `error: unknown option 'foo'`.
* After: `git checkout -f -- --foo` → `error: pathspec '--foo' did not match any file(s) known to git`

Related: #1314

---

This pull request also introduces unit tests (with a `mockShellRunner`) for `gitCheckout()`, `gitClone()`, `gitClean()`, `gitCleanSubmodules()` and `gitFetch()`.

```
$ go test ./bootstrap -run TestGit -v
=== RUN   TestGitCheckRefFormat
=== RUN   TestGitCheckRefFormat/hello
=== RUN   TestGitCheckRefFormat/--world
--- PASS: TestGitCheckRefFormat (0.00s)
    --- PASS: TestGitCheckRefFormat/hello (0.00s)
    --- PASS: TestGitCheckRefFormat/--world (0.00s)
=== RUN   TestGitCheckoutValidatesRef
--- PASS: TestGitCheckoutValidatesRef (0.00s)
=== RUN   TestGitCheckoutSomething
--- PASS: TestGitCheckoutSomething (0.00s)
=== RUN   TestGitClone
--- PASS: TestGitClone (0.00s)
=== RUN   TestGitClean
--- PASS: TestGitClean (0.00s)
=== RUN   TestGitCleanSubmodules
--- PASS: TestGitCleanSubmodules (0.00s)
=== RUN   TestGitFetch
--- PASS: TestGitFetch (0.00s)
PASS
```

### DIFF
--- a/bootstrap/git.go
+++ b/bootstrap/git.go
@@ -8,7 +8,6 @@ import (
 	"path/filepath"
 	"regexp"
 	"strings"
-	"unicode"
 
 	"github.com/buildkite/agent/v3/bootstrap/shell"
 	"github.com/buildkite/shellwords"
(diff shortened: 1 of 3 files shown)


bootstrap/git.go
```diff
@@ -27,11 +27,18 @@ type gitError struct {
...
+	Run(string, ...string) error
+}
+
+func gitCheckout(sh shellRunner, gitCheckoutFlags, reference string) error {
 	individualCheckoutFlags, err := shellwords.Split(gitCheckoutFlags)
 	if err != nil {
 		return err
 	}
+	if !gitCheckRefFormat(reference) {
+		return fmt.Errorf("%q is not a valid git ref format", reference)
```
pda, line 40: This should be the only functional change, the rest is testability refactoring.

sj26 (dismissed):
Seems legit 👍

bootstrap/git.go
```diff
@@ -248,9 +249,63 @@ func resolveGitHost(sh *shell.Shell, host string) string {
...
+			}
+		}
+		if i == 0 {
+			// (disallow refs that would be interpreted as command options/flags)
+			for _, deny := range `-` {
+				if c == deny {
+					return false
+				}
+			}
+		}
```
sj26, line 300: This seems like a convoluted way to do `strings.HasPrefix(ref, "-")`?

bootstrap/git.go
```diff
@@ -248,9 +249,63 @@ func resolveGitHost(sh *shell.Shell, host string) string {
...
+		if i == len(ref)-1 {
+			// 7. They cannot end with a dot .
+			for _, deny := range `.` {
+				if c == deny {
+					return false
+				}
+			}
+		}
```
sj26, line 308: This seems like a convoluted way to do `strings.HasSuffix(ref, ".")`?

bootstrap/git.go
pda: Yeah… I went through various implementations as I got lured by unit testing into implementing more and more of the spec. I was hoping to implement the whole function as a single pass over `ref`, rather than lots of `strings.Whatever` calls that each process the whole string. Especially as I was thinking there would be various disallowed prefix/suffix characters… turns out at this point there's only one of each. I should probably simplify it back down.

pda: I should probably just rewrite the whole thing as one regexp.

pda: b3cb5f97f7482c8d247e520d794466b6f0e94251

sj26 (approved):
Nice

PR #1316, opened 2020-10-05
Update the way we detect windows versions

yob:
While running the agent on Windows 10 I noticed we were submitting the version as "Windows version 6.2", which [maps to windows 8](https://en.wikipedia.org/wiki/List_of_Microsoft_Windows_versions).

According to MSDN, the GetVersion call we're using [always reports windows 8 for compatibility reasons](https://docs.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getversion):

> GetVersion may be altered or unavailable for releases after Windows 8.1.
> Instead, use the Version Helper functions. For Windows 10 apps, please see
> [Targeting your applications for
> Windows](/windows/win32/sysinfo/targeting-your-application-at-windows-8-1).
>
> With the release of Windows 8.1, the behavior of the GetVersion API has changed
> in the value it will return for the operating system version. The value
> returned by the GetVersion function now depends on how the application is
> manifested.
>
> Applications not manifested for Windows 8.1 or Windows 10 will return the
> Windows 8 OS version value (6.2).

Adding a manifest seems to involve embedding XML in the executable, which doesn't sound like fun.

Lucky for us there's an alternative syscall that doesn't lie about the version: [RtlGetVersion](https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-rtlgetversion).

I've tested this on:

1. Windows 10 (1909) and it reports version "10.0".
2. Windows 7 and it reports "6.1"

MSDN says RtlGetVersion is available from Windows 2000.

### DIFF
--- a/system/version_dump_windows.go
+++ b/system/version_dump_windows.go
@@ -2,15 +2,12 @@ package system
 
 import (
 	"fmt"
-	"syscall"
-
 	"github.com/buildkite/agent/v3/logger"
+	"golang.org/x/sys/windows"
 )
 
 func VersionDump(_ logger.Logger) (string, error) {
-	dll := syscall.MustLoadDLL("kernel32.dll")
-	p := dll.MustFindProc("GetVersion")
-	v, _, _ := p.Call()
+	info := windows.RtlGetVersion()
 
-	return fmt.Sprintf("Windows version %d.%d (Build %d)\n", byte(v), uint8(v>>8), uint16(v>>16)), nil
+	return fmt.Sprintf("Windows version %d.%d (Build %d)\n", info.MajorVersion, info.MinorVersion, info.BuildNumber), nil
 }


yob:
One thing this doesn't address is detecting the difference between server and non-server versions of windows.

[According to wikipedia](https://en.wikipedia.org/wiki/List_of_Microsoft_Windows_versions), a major.minor of "10.0" can mean windows 10, but it can also mean:

1. Windows Server 2016
1. Windows Server 2019
1. Windows Server, version 1903
1. Windows Server, version 1909
1. Windows Server, version 2004

:shrug:

chloeruka (approved):
👍 I feel way better about using golang sys/windows for this, everything about it is nicer, especially that we're no longer misreporting agent OS' as Windows 8. 

From what I've seen, in the brave new world of _Windows 10 everywhere_, the language has shifted from talking about major/minor in vague terms (e.g. 7, 8.0, 8.1) to some sort of double-barrel version numbering system such as "Windows 10 version 2004"<sup>1,2</sup>, which refers to a specific released build number (in this case [19041.423](https://support.microsoft.com/en-us/help/4568831/windows-10-update-kb4568831)) of the desktop/server product. It'd be nice but not essential to also have that version, but my brief googling doesn't suggest an easy way of getting it. We should still be able to relate build number to version so we can tease out specific compatibility issues.

<sup>1</sup> also known as the "Windows 10 May 2020 Update"
<sup>2</sup> ...also also known as "update 20H1"

pda:
I just noticed last night that Buildkite was reporting my Windows 10 agent as Windows 8, and was going to look into it.
Thanks!

pda:
Seems good on my Windows 10 Pro machine…

Before:
Windows 8

After:
Windows 10

Issue #1292, opened 2020-09-10
Job and it's processes are unable to be stopped/canceled/terminated on windows


filipVisko:
We're seeing some odd behaviour where`buildkite-agent` is unable to stop spawned processes.
This occurs during game engine builds (which run ugly batch or other things out of our control) and when either:
- The job being executed times out. ie: reaches its `BUILDKITE_TIMEOUT` limit.
- A user visiting buildkite.com clicks the cancel button for the job.

Expected outcome: Job cancels immediately! The job state in the buildkite.com UI is now red.
Actual outcome: Job is in a `canceling` state forever. The job state in the UI is yellow. (The only reliable way to terminate the process is to use the OS service manager)

https://github.com/buildkite/agent/blob/master/agent/job_runner.go#L357
Inside of `Cancel()`, it logs with `Canceling job %s with a grace period of %d` and then with `Job %s hasn't stopped in time, terminating`.
This happens because `Cancel()` is called by `CancelAndStop()` which is called from within `Stop()`, but `Stop()` never checks the return value of `CancelAndStop()`.

I suspect that `CancelAndStop()` is returning an error which needs to be handled. `go-lint` also flags this line as an unhandled error.
https://github.com/buildkite/agent/blob/master/agent/agent_worker.go#L271

sj26:
Thanks for the report @filipVisko, this does sound like a problem. We've got a few folks looking at agent stuff soon, I'll see if we can get their eyes on this as well.

meteorcloudy:
We encountered the same problem on Bazel CI:
https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/633#637566fd-07bd-4ed7-9571-6a1d8f9c77fd

filipVisko:
The added symptom of this issue is that the Buildkite http health endpoint still returns `200`. Which means that our autoscaling groups never kills these instances and wastes our compute time (until we notice this via alerts for long lived instances).

yob:
Thanks for reporting this @filipVisko. We've had another customer report the same symptoms and it's definitely something we'd like to fix.

In #1317 we added some debug logging for the unhandled errors you identified here, and with the other customer we confirmed that `r.Cancel()` is returning an error: 

```
2020-10-09 09:29:18 INFO   host-1 Canceling job 818efd1d-d0e7-4c91-8696-6ce3ec150a0c with a grace period of 10s 
[...]

PR #1321, opened 2020-10-11
Add sampling logger for job-related processes on windows

yob:
We've had reports from multiple customers that the windows agent can sometimes get "stuck" (eg #1292)

- from the perspective of our servers the agent is running the stuck job for days
- the main agent process considers the job to still be running, so it doesn't mark the job as finished with our API
- the main agent process sees no further output from the bootstrap process, so no new log chunks are uploaded
- when the user tries to cancel the job on buildkite.com, the main agent process attempts to terminate the bootstrap process and fails

This PR doesn't attempt ro fix the issue - we're still looking for a root cause that we can reliably reproduce.

This adds a sampling logger that prints the full process tree of the bootstrap and job every 5 seconds.

I'm told that the process hierarchy in windows can't be trusted too much. Each process maintains a pointer to its parent, but it's common for the parent process to end and leave orphan processes that can't be found by walking the parent/child tree from the bootstrap.

I've attempted to mitigate that by adding the bootstrap process to a Job Object. All child processes of the bootstrap are automatically added to the same Job, and we can use the QueryInformationJobObject windows API call to retrieve all processes in the Job.

I'm interested to see how the process tree evolves over the life of one of these stuck jobs. Hopefully we can get the clue we need to reproduce the issue and come up with a reliable fix.

The result looks like this (I considered unicode box drawing characters for a better tree-look..... but it's late). 

Screenshot from 2020-10-12 00-42-24

Note: this is not production ready code! I haven't put much thought into the abstractions I've modified. To enable the logging, start the agent like this:

    buidlkite-agent.exe start -token=foo -debug -experiments=debug-processes

Some useful reading on Job Objects:

1. https://docs.microsoft.com/en-us/windows/win32/procthread/job-objects?redirectedfrom=MSDN
2. https://docs.microsoft.com/en-us/windows/win32/api/jobapi2/
3. https://nikhilism.com/post/2017/windows-job-objects-process-tree-management/
4. https://gist.github.com/hallazzang/76f3970bfc949831808bbebc8ca15209

### DIFF
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -696,8 +696,12 @@ func (r *JobRunner) onProcessStartCallback() {
 				r.logger.Debug("[ProcessLogger] Routine that logs the job process tree has finished")
 			}()
 			for {
-				tree := r.process.Processtree()
-				r.logger.Warn("Bootstrap Processtree:\n%s", tree.String())
+				tree, err := r.process.Processtree()
+				if err != nil {
+					r.logger.Error("Fetching process tree failed: %v", err)
+				} else {
+					r.logger.Warn("Bootstrap Processtree:\n%s", tree.String())
+				}
 				// Sleep for a bit, or until the job is finished
 				select {
 				case <-time.After(5 * time.Second):
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -683,6 +683,32 @@ func (r *JobRunner) onProcessStartCallback() {
 			}
 		}
 	}()
+
+	// Start a routine that will constantly log out the process tree created by the job
+	// Windows only, because that's where I'm debugging things atm
+	if experiments.IsEnabled(`debug-processes`) {
+		r.routineWaitGroup.Add(1)
+		go func() {
+			defer func() {
+				// Mark this routine as done in the wait group
+				r.routineWaitGroup.Done()
+
+				r.logger.Debug("[ProcessLogger] Routine that logs the job process tree has finished")
+			}()
+			for {
+				tree := r.process.Processtree()
+				r.logger.Warn("Bootstrap Processtree:\n%s", tree.String())
+				// Sleep for a bit, or until the job is finished
+				select {
+				case <-time.After(5 * time.Second):
+				case <-r.context.Done():
+					return
+				case <-r.process.Done():
+					return
+				}
+			}
+		}()
+	}
 }
 
 func (r *JobRunner) onUploadHeaderTime(cursor int, total int, times map[string]string) {
--- a/process/process.go
+++ b/process/process.go
@@ -99,12 +99,12 @@ func (p *Process) Pid() int {
 }
 
 // Return a tree collection of process and its descendent processes in a tree structure
-func (p *Process) Processtree() ProcessTreeRoot {
+func (p *Process) Processtree() (ProcessTreeRoot, error) {
 	tree, err := p.processGroup.processTree()
 	if err != nil {
-		p.logger.Error("error fetching process tree: %v", err)
+		return tree, err
 	}
-	return tree
+	return tree, nil
 }
 
 // WaitResult returns the raw error returned by Wait()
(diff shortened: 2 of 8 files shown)


agent/job_runner.go
```diff
@@ -683,6 +684,32 @@ func (r *JobRunner) onProcessStartCallback() {
 			}
 		}
 	}()
+
+	// Start a routine that will constantly log out the process tree created by the job
+	// Windows only, because that's where I'm debugging things atm
+	if runtime.GOOS == `windows` {
```
yob, line 690: :see_no_evil:

process/process.go
```diff
@@ -97,6 +100,34 @@ func (p *Process) Pid() int {
...
+func (p *Process) Processlist() []ps.Process {
+	if runtime.GOOS == `windows` {
+		processes, err := p.winExitGroup.listProcesses()
+		if err != nil {
+			p.logger.Error("error fetching windows pids: %v", err)
+		}
+		return processes
+	} else {
+		result := make([]ps.Process,0)
+		return result
```
yob, line 113: :see_no_evil:

process/process.go
```diff
@@ -97,6 +100,34 @@ func (p *Process) Pid() int {
...
+func (p *Process) Processtree() ProcessTreeRoot {
+	if runtime.GOOS == `windows` {
+		tree, err := p.winExitGroup.processTree()
+		if err != nil {
+			p.logger.Error("error fetching windows process tree: %v", err)
+		}
+		return tree
+	} else {
+		tree := ProcessTreeRoot{}
+		return tree
```
yob, line 127: :see_no_evil:

process/process_exit_group.go
```diff
@@ -0,0 +1,156 @@
...
+	//		LimitFlags: windows.JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE,
+	//	},
+	//}
+	//if _, err := windows.SetInformationJobObject(
+	//	handle,
+	//	windows.JobObjectExtendedLimitInformation,
+	//	uintptr(unsafe.Pointer(&info)),
+	//	uint32(unsafe.Sizeof(info))); err != nil {
+	//	return 0, err
+	//}
```
yob, line 79: Job objects are super interesting, and it **seems* like they might provide a way for us to reliably terminate **all** processes that run as part of a job. I'm nervous that enabling it now might make it harder to observe the root issue at play in #1292 though, so haven't included them in this PR for now.

yob:
Oops, this is crash-y in its current for:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x28 pc=0xbc0d08]

goroutine 32 [running]:
github.com/buildkite/agent/v3/process.processGroup.processTree(0x424, 0xc000197f50, 0xc0000331b8, 0xc000197f68, 0x91a1eb, 0xc00050a0f8)
        /home/jh/git/bk/agent/process/process_group_windows.go:145 +0x128
github.com/buildkite/agent/v3/process.(*Process).Processtree(0xc000580200, 0xc000313efc, 0x3, 0x0)
        /home/jh/git/bk/agent/process/process.go:103 +0x3d
github.com/buildkite/agent/v3/agent.(*JobRunner).onProcessStartCallback.func3(0xc0000f6e00)
        /home/jh/git/bk/agent/agent/job_runner.go:699 +0x8f
created by github.com/buildkite/agent/v3/agent.(*JobRunner).onProcessStartCallback
        /home/jh/git/bk/agent/agent/job_runner.go:691 +0x112
```

yob:
> Oops, this is sometimes crash-y in its current form:

fixed, I think. Mapping PIDs to processes can sometimes fail if the process has finished in the interim.

yob:
I'm going to close this without merging for now. It was useful for debugging the agent behaviour on windows for a while, but I'm not yet convinced it's worth carrying the ongoing maintenance cost on master.

PR #1322, opened 2020-10-12
shutdown hook on Windows supports Batch & PowerShell

pda:
## Background

#1275 introduced a `shutdown` hook, which runs as the agent is shutting down. This is the first hook we've added which doesn't run as part of the `bootstrap` process for a job.

The current implementation looks for a `$HOOKS/shutdown` file, and runs it directly as a shell command, which assumes it's a bash script due to the lack of a file extension. Windows users are more likely to specify a shutdown hook as a batch or PowerShell file.

## Solution

I've done some refactoring in #1324, #1325 and to extract some hook code from package `bootstrap` into a new package `hook`.

This pull request uses the new `hook.Find(dir, name)` function to give `shutdownHook()` the same search logic as bootstrap hooks; that means on Windows `shutdown.bat`, `shutdown.cmd` and `shutdown.ps1` files are considered.

`TestShutdownHook()` now uses `shutdown.bat` on Windows, which appears to be more likely to work than a `shutdown` bash script. The test also verifies no errors are logged — previously the test was logging failure messages on Windows but they were being ignored.

## Local verification

In my local Windows environment, this works for `shutdown.bat`:
```
> C:\buildkite-agent\hooks\shutdown.bat
C:\Users\pda\bk\agent>echo hello
hello
```

I get an error for `shutdown.ps1`, but it looks like it's being invoked fine, and my machine isn't configured to allow the script to run, so that's fine:
```
> C:\buildkite-agent\hooks\shutdown.ps1
File C:\buildkite-agent\hooks\shutdown.ps1 cannot be loaded because running scripts is disabled on this system. For
more information, see about_Execution_Policies at https:/go.microsoft.com/fwlink/?LinkID=135170.
    + CategoryInfo          : SecurityError: (:) [], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : UnauthorizedAccess
2020-10-12 21:25:42 ERROR  Shutdown hook error: exit status 1 hook=shutdown
```

### Caveat

I still get an error during shutdown when specifying a `shutdown` bash hook:
```
/bin/bash: C:/buildkite-agent/hooks/shutdown: No such file or directory
2020-10-12 21:16:39 ERROR  Shutdown hook error: exit status 127
```
I'm not sure why that's not working, or how/whether Windows bash scripts work in our existing bootstrap hooks. I think we can figure that out in a future pull request — it's not a regression in this pull request.

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -15,6 +15,7 @@ import (
 	"github.com/buildkite/agent/v3/bootstrap/shell"
 	"github.com/buildkite/agent/v3/cliconfig"
 	"github.com/buildkite/agent/v3/experiments"
+	"github.com/buildkite/agent/v3/hook"
 	"github.com/buildkite/agent/v3/logger"
 	"github.com/buildkite/agent/v3/metrics"
 	"github.com/buildkite/agent/v3/process"
@@ -798,8 +799,11 @@ func handlePoolSignals(l logger.Logger, pool *agent.AgentPool) chan os.Signal {
 // if it's available. This is very similar to the bootstrap hook code. Just run at
 // the agent level.
 func shutdownHook(log logger.Logger, hooksPath string) {
-	p := filepath.Join(hooksPath, "shutdown")
-	if _, err := os.Stat(p); os.IsNotExist(err) {
+	p, err := hook.Find(hooksPath, "shutdown")
+	if err != nil {
+		if !os.IsNotExist(err) {
+			log.Error("Error finding shutdown hook: %v", err)
+		}
 		return
 	}
 	sh, err := shell.New()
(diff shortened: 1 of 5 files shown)


clicommand/agent_start.go
```diff
@@ -798,15 +799,19 @@ func handlePoolSignals(l logger.Logger, pool *agent.AgentPool) chan os.Signal {
 // if it's available. This is very similar to the bootstrap hook code. Just run at
 // the agent level.
 func shutdownHook(log logger.Logger, hooksPath string) {
-	p := filepath.Join(hooksPath, "shutdown")
-	if _, err := os.Stat(p); os.IsNotExist(err) {
+	p, err := hook.Find(hooksPath, "shutdown")
```
pda, line 802: Here's the crux of the change — `hook.Find` searches `.BAT;.CMD;.PS1` using [the Windows version of `shell.LookPath()`](https://github.com/buildkite/agent/blob/99d709436d9b2894f5e94727d261600831aa1269/bootstrap/shell/lookpath_windows.go#L59).

keithpitt (approved):

PR #1323, opened 2020-10-12
use golang.org/x/sys/windows instead of syscall, where we can

yob:
The syscall package in the stdlib has a frozen interface, however `golang.org/x/sys` continues to evolve. Some of the windows APIs that we're manually accessing via syscall now have go-ish wrappers in sys/windows that are a bit nicer to work with.

There should be no functional difference from this change.

This is extracted from #1321, which I expect to take some time to mature. This little bit is totally mergable though!


pda (approved):
Looks much nicer.
✅ … assuming it's been verified on a Windows machine :)

yob:
> assuming it's been verified on a Windows machine :)

I just retested how the agent responds to ctl-c on both master and this branch, to make sure I didn't end up with egg on my face. 

In both cases the `GenerateConsoleCtrlEvent` call works, the process tree is sent the interrupt, and the job finishes non-zero.

PR #1325, opened 2020-10-13
Refactor: extract bootstrap.hookScriptWrapper → hook.ScriptWrapper

pda:
_This is a refactor extracted from #1322, to reduce the size/scope of that pull request._

#1324 began refactoring elements of `bootstrap` into a new `hook` package.
This pull request extracts `bootstrap.hookScriptWrapper` → `hook.ScriptWrapper`:

```diff
-// package bootstrap
-
-type hookScriptWrapper struct
-
-func newHookScriptWrapper(hookPath string) (*hookScriptWrapper, error)

+// package hook
+
+type ScriptWrapper struct
+
+// CreateScriptWrapper creates and configures a ScriptWrapper.
+// Writes temporary files to the filesystem.
+func CreateScriptWrapper(hookPath string) (*ScriptWrapper, error)
```

This allows code outside package `bootstrap` to run hook scripts with env/pwd change tracking. It also reduces the size of package `bootstrap`.

To support this refactor,  `bootstrap.newTestShell()` was extracted from `bootstrap/hook_test.go` (from which it was being called from various other test files) to `shell.NewTestShell()` in a new `bootstrap/shell/test.go` file:

```diff
-// package bootstrap
-func newTestShell(t *testing.T) *shell.Shell
+// package shell
+// NewTestShell creates a minimal shell suitable for tests.
+func NewTestShell(t *testing.T) *Shell
```

Also, `bootstrap.addExecutePermissionToFile()` is extracted to `utils.ChmodExecutable()`:

```diff
-// package bootstrap
-// Makes sure a file is executable
-func addExecutePermissionToFile(filename string) error
+// package utils
+// ChmodExecutable sets the executable mode/flag on a file, if not already.
+func ChmodExecutable(filename string) error

### DIFF
--- a/utils/file.go
+++ b/utils/file.go
@@ -1,0 +1,21 @@
+package utils
+
+import (
+	"fmt"
+	"os"
+)
+
+// ChmodExecutable sets the executable mode/flag on a file, if not already.
+func ChmodExecutable(filename string) error {
+	s, err := os.Stat(filename)
+	if err != nil {
+		return fmt.Errorf("Failed to retrieve file information of \"%s\" (%s)", filename, err)
+	}
+	if s.Mode()&0100 == 0 {
+		err = os.Chmod(filename, s.Mode()|0100)
+		if err != nil {
+			return fmt.Errorf("Failed to mark \"%s\" as executable (%s)", filename, err)
+		}
+	}
+	return nil
+}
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: utils/file.go [hunk 1]
```diff
@@ -0,0 +1,21 @@
+package utils
+
+import (
+	"fmt"
+	"os"
+)
+
+// ChmodExecutable sets the executable mode/flag on a file, if not already.
+func ChmodExecutable(filename string) error {
```
keithpitt, line 9: Feels on renaming this to `SetChmodExecutable`? Reading the code that uses it, I can either interpret it as a "does this file have chmod executable" or "set this as chmod executable"

pda: My thinking was “chmod” is short for “change mode”, so it implies the verb.

keithpitt: Oh yeah, doh! Good point.

keithpitt (dismissed):

PR #1326, opened 2020-10-13
shutdown hook: pipe output lines to main logger

pda:
This pull request pipes the output of the new shutdown script (#1275) into the main logger.

## Before

The shutdown hook output is written directly to `os.Stdout`:

image

(The :skull: is my shell warning that the program terminated without a trailing newline)

## After

The shutdown hook output is pipes to the logger, with a prefix of `shutdown` to identify where it came from:

image

## And with `--no-color`

I had some issues with color creeping in despite `--no-color` being passed. All fixed:

image

## Test hook

Here's the `shutdown` hook I was using:

```
$ cat tmp/hooks/shutdown
#!/bin/bash

echo "hello stdout from shutdown hook"
echo >&2 "hello stderr from shutdown hook"
echo -n "no line ending here"
```

## Notes

It would be nice to refactor the `io.Writer`→`logger.Logger` pipe gear out of `shutdownHook()` so it can be used elsewhere. That'll need a bit of finesse to decouple it without concurrency issues, so I figure we can do it later when there's a second use-case.

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -1,12 +1,15 @@
 package clicommand
 
 import (
+	"bufio"
 	"fmt"
+	"io"
 	"net/http"
 	"os"
 	"os/signal"
 	"path/filepath"
 	"runtime"
+	"sync"
 	"syscall"
 	"time"
 
@@ -723,7 +726,7 @@ var AgentStartCommand = cli.Command{
 		pool := agent.NewAgentPool(workers)
 
 		// Agent-wide shutdown hook. Once per agent, for all workers on the agent.
-		defer func() { shutdownHook(l, cfg.HooksPath) }()
+		defer shutdownHook(l, cfg)
 
 		// Handle process signals
 		signals := handlePoolSignals(l, pool)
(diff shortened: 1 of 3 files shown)


goodspark (commented):
I don't know if approvals from non-Buildkite org members mean anything, but this looks good to me! Nice and consistent, clean, colored log output.

goodspark (approved):
Oops forgot the approval

clicommand/agent_start.go
```diff
@@ -811,8 +815,29 @@ func shutdownHook(log logger.Logger, hooksPath string) {
...
+	sh.Logger = &shell.WriterLogger{Writer: w, Ansi: !cfg.NoColor} // for Promptf
+	sh.Writer = w                                                  // for stdout+stderr
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		scan := bufio.NewScanner(r) // log each line separately
+		log = log.WithFields(logger.StringField("hook", "shutdown"))
+		for scan.Scan() {
+			log.Info(scan.Text())
```
pda, line 830: Unfortunately there's no way to differentiate stdout vs stderr without substantial changes to our `shell`.

clicommand/agent_start.go
```diff
@@ -723,7 +726,7 @@ var AgentStartCommand = cli.Command{
 		pool := agent.NewAgentPool(workers)
 
 		// Agent-wide shutdown hook. Once per agent, for all workers on the agent.
-		defer func() { shutdownHook(l, cfg.HooksPath) }()
+		defer shutdownHook(l, cfg)
```
pda, line 729: I'm pretty sure this is a harmless change, and neater.

clicommand/agent_start.go
```diff
@@ -798,8 +801,9 @@ func handlePoolSignals(l logger.Logger, pool *agent.AgentPool) chan os.Signal {
 // shutdownHook looks for a shutdown hook script in the hooks path and executes it
 // if it's available. This is very similar to the bootstrap hook code. Just run at
 // the agent level.
-func shutdownHook(log logger.Logger, hooksPath string) {
-	p, err := hook.Find(hooksPath, "shutdown")
+func shutdownHook(log logger.Logger, cfg AgentStartConfig) {
```
pda, line 804: We now need `cfg.NoColor` in addition to `cfg.HooksPath`, so receive the entire `cfg AgentStartConfig` instead of just `hooksPath string`

sj26 (approved):
Seems sensible, and looks well implemented to me 👍

PR #1328, opened 2020-10-14
Add --mirror flag by default for mirror clones

### LINKED ISSUE #1319: Documentation for git-clone-mirror-flags is confusing
Hey folks, the [documentation](https://buildkite.com/docs/agent/v3/configuration) around the `git-clone-mirror-flags` is a little confusing. It mentions an example "-v --mirror" and no default. I set it to `--config pack.threads=35` to speed up our checkouts. However, that meant that the `-v` and `--mirror` flags were lost, making the mirror clone a full checkout clone. I think the best solution here is to update the mirror clone to always add the `--mirror` flag (like what happens for `--reference`) and make the `-v` flag a default.

(I didn't notice this previously because I'm pre-cloning mirrors for a few select repos onto machines before the agent starts up and I'm doing my own clone there).

chrislloyd:
The mirror clone can't operate without it as subsequent steps will fail if they are performed on a full checkout. I think I mentioned this in the original issue, but I think this is a better default.

Fixes #1319

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -1001,7 +1001,11 @@ func (b *Bootstrap) updateGitMirror() (string, error) {
 	// If we don't have a mirror, we need to clone it
 	if !utils.FileExists(mirrorDir) {
 		b.shell.Commentf("Cloning a mirror of the repository to %q", mirrorDir)
-		if err := gitClone(b.shell, b.GitCloneMirrorFlags, b.Repository, mirrorDir); err != nil {
+
+		gitCloneMirrorFlags := b.GitCloneMirrorFlags
+		gitCloneMirrorFlags += "--mirror"
+
+		if err := gitClone(b.shell, gitCloneMirrorFlags, b.Repository, mirrorDir); err != nil {
 			return "", err
 		}
 
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -1001,11 +1001,8 @@ func (b *Bootstrap) updateGitMirror() (string, error) {
 	// If we don't have a mirror, we need to clone it
 	if !utils.FileExists(mirrorDir) {
 		b.shell.Commentf("Cloning a mirror of the repository to %q", mirrorDir)
-
-		gitCloneMirrorFlags := b.GitCloneMirrorFlags
-		gitCloneMirrorFlags += "--mirror"
-
-		if err := gitClone(b.shell, gitCloneMirrorFlags, b.Repository, mirrorDir); err != nil {
+		flags := "--mirror " + b.GitCloneMirrorFlags
+		if err := gitClone(b.shell, flags, b.Repository, mirrorDir); err != nil {
 			return "", err
 		}
 
--- a/bootstrap/integration/checkout_integration_test.go
+++ b/bootstrap/integration/checkout_integration_test.go
@@ -116,7 +116,7 @@ func TestCheckingOutLocalGitProject(t *testing.T) {
 	// But assert which ones are called
 	if experiments.IsEnabled(`git-mirrors`) {
 		git.ExpectAll([][]interface{}{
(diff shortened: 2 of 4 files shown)


bootstrap/bootstrap.go
```diff
@@ -1001,7 +1001,11 @@ func (b *Bootstrap) updateGitMirror() (string, error) {
 	// If we don't have a mirror, we need to clone it
 	if !utils.FileExists(mirrorDir) {
 		b.shell.Commentf("Cloning a mirror of the repository to %q", mirrorDir)
-		if err := gitClone(b.shell, b.GitCloneMirrorFlags, b.Repository, mirrorDir); err != nil {
+
+		gitCloneMirrorFlags := b.GitCloneMirrorFlags
+		gitCloneMirrorFlags += "--mirror"
```
pda, line 1006: If `b.GitCloneMirrorFlags` is `"-v"` will this result in `"-v--mirror"` without a space?
It's a shame we're passing this around as a string, but that would be a bigger refactor.
We could `shellwords.Split()`, `append`, and then `strings.Join()` it here, which seems pretty clunky but will give the correct result.

Or, just…

```suggestion
		gitCloneMirrorFlags += " --mirror"
```

pda: I've pushed a commit which addresses this.
(First time I've ever pushed a commit to a third-party PR branch like that 🎉)

bootstrap/bootstrap.go
```diff
@@ -1001,7 +1001,8 @@ func (b *Bootstrap) updateGitMirror() (string, error) {
 	// If we don't have a mirror, we need to clone it
 	if !utils.FileExists(mirrorDir) {
 		b.shell.Commentf("Cloning a mirror of the repository to %q", mirrorDir)
-		if err := gitClone(b.shell, b.GitCloneMirrorFlags, b.Repository, mirrorDir); err != nil {
+		flags := "--mirror " + b.GitCloneMirrorFlags
```
pda, line 1004: This should do the trick, with a harmless trailing space if there's no `b.GitCloneMirrorFlags` configured.

bootstrap/integration/checkout_integration_test.go
```diff
@@ -63,7 +63,7 @@ func TestWithResolvingCommitExperiment(t *testing.T) {
 	// But assert which ones are called
 	if experiments.IsEnabled(`git-mirrors`) {
 		git.ExpectAll([][]interface{}{
-			{"clone", "--bare", "--", tester.Repo.Path, matchSubDir(tester.GitMirrorsDir)},
+			{"clone", "--mirror", "--bare", "--", tester.Repo.Path, matchSubDir(tester.GitMirrorsDir)},
```
pda, line 66: I had to flip all the test expectations, but I think it makes more sense this way, with `--mirror` first`.

bootstrap/integration/checkout_integration_test.go
```diff
@@ -103,7 +103,7 @@ func TestCheckingOutLocalGitProject(t *testing.T) {
 
 	env := []string{
 		"BUILDKITE_GIT_CLONE_FLAGS=-v",
-		"BUILDKITE_GIT_CLONE_MIRROR_FLAGS=--bare",
+		"BUILDKITE_GIT_CLONE_MIRROR_FLAGS=--config pack.threads=35",
```
pda, line 106: Let's get your use-case in the test suite 😁

pda (approved):
Thanks @chrislloyd 

Are you happy with the changes I've made on your PR?

bootstrap/integration/checkout_integration_test.go
chrislloyd: Ha ha ha, awesome!

chrislloyd:
💯 thank you! LGTM too.

BTW, youse should turn on public builds for buildkite/agent! Was kind of flying blind with the tests :D

pda:
> public builds for buildkite/agent!

Agreed! I was looking into that, we'd need to do some security due diligence and perhaps agent isolation. I hope we can find time to do that soon.

PR #1329, opened 2020-10-15
Use a win32 Job Object to manage job termination on windows

### LINKED ISSUE #1292: Job and it's processes are unable to be stopped/canceled/terminated on windows

yob:
Imagine a windows process tree like this:

    1000 buildkite-agent.exe start
    └─ 1001 buildkite-agent.exe bootstrap
      └─ 1002 cmd.exe
        └─ 1003 python.exe
          └─ 1004 python.exe

It's reasonably common in windows for a process to end and orphan its children, leaving a tree like this:

    1000 buildkite-agent.exe start
    └─ 1001 buildkite-agent.exe bootstrap
      └─ 1002 cmd.exe
    └─ 1004 python.exe

Then the bootstrap might end, leaving the orphan process hanging around after the job has completed. Oops.

On UNIX systems we avoid this by creating a progress group, so any process that starts during job execution is tidied up when the bootstrap process finishes.

Here I've introduced win32 Job Object to achieve a similar effect on Windows. This has two noteworthy impacts:

1. The bootstrap process created to run a Buildkite Job is added to a new win32 Job Object. Any child process started by descendant processes will be added to the same Job Object, and they'll be automatically terminated when the bootstrap process finishes
2. A request to hard terminate a running Buildkite Job now does so by terminating the win32 Job (which in turn terminates all processes within it). This should be more reliable than TASKKILL.EXE, which was unable to find and destroy orphans and would error completely if the bootstrap process had completed

HT to [this gist](https://gist.github.com/hallazzang/76f3970bfc949831808bbebc8ca15209#gistcomment-2948162) that pointed me in the right direction [1]. The [MSDN docs on Job Objects](https://docs.microsoft.com/en-us/windows/win32/procthread/job-objects?redirectedfrom=MSDN) were reasonably useful too.

Fixes #1292

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -204,8 +204,7 @@ func (p *Process) Run() error {
 		if err != nil {
 			return err
 		}
-		err = p.postStart()
-		if err != nil {
+		if err := p.postStart(); err != nil {
 			p.logger.Error("[Process] postStart failed: %v", err)
 		}
 		p.pid = p.command.Process.Pid
--- a/process/process.go
+++ b/process/process.go
@@ -82,6 +82,8 @@ type Process struct {
 	command       *exec.Cmd
 	mu            sync.Mutex
 	started, done chan struct{}
+
+	winJobHandle uintptr
 }
 
 // New returns a new instance of Process
@@ -202,7 +204,10 @@ func (p *Process) Run() error {
 		if err != nil {
 			return err
 		}
-
+		err = p.postStart()
+		if err != nil {
+			p.logger.Error("[Process] postStart failed: %v", err)
+		}
 		p.pid = p.command.Process.Pid
 
 		// Signal waiting consumers in Started() by closing the started channel
--- a/process/signal.go
+++ b/process/signal.go
@@ -19,6 +19,11 @@ func (p *Process) setupProcessGroup() {
 	}
 }
 
+func (p *Process) postStart() error {
+	// a no-op on non-windows
+	return nil
+}
+
 func (p *Process) terminateProcessGroup() error {
 	p.logger.Debug("[Process] Sending signal SIGKILL to PGID: %d", p.pid)
 	return syscall.Kill(-p.pid, syscall.SIGKILL)
--- a/process/signal_test.go
+++ b/process/signal_test.go
@@ -1,9 +1,9 @@
 package process_test
 
 import (
+	"runtime"
 	"syscall"
 	"testing"
-	"runtime"
 
 	"github.com/buildkite/agent/v3/process"
 	"github.com/stretchr/testify/assert"
--- a/process/signal_windows.go
+++ b/process/signal_windows.go
@@ -5,9 +5,8 @@ package process
 import (
 	"errors"
 	"fmt"
-	"os/exec"
-	"strconv"
 	"syscall"
+	"unsafe"
 
 	"golang.org/x/sys/windows"
 )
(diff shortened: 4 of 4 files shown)


yob:
It'd be really nice to have a regression test for this, I'm not across our integration testing options (particularly on windows).

My testing was all manual, using a windows 10 VM and a [custom-built go program that intentionally created orphaned processes](https://github.com/yob/win-process-tree). The repo for that program has a Buildkite pipeline that compiles and runs it, which generates orphan processes that the master-branch agents will leave running.

The failure mode is actually worse than just leaking processes. We use `os/cmd.Wait()` in the main agent process to block until the bootstrap process finishes. However, that `Wait()` call never returns because there's a known [golang issue](https://github.com/golang/go/issues/23019) where `Wait()` **also** waits for all child processes to terminate before returning. If the orphaned processes never finish, then the agent never marks the Job finished on buildkite.com. Ooof.

That known issue happens on Unix and Windows, however it's never caused us trouble on Unix because we're careful to clean up all child processes and not leave orphans when the bootstrap completes. This change means the windows agent also side-steps that issue.

chloeruka:
Yesss, you got to the bottom of it! Awesome work and thorough write up, really keen to see this roll out.

process/process.go
```diff
@@ -202,7 +204,10 @@ func (p *Process) Run() error {
 		if err != nil {
 			return err
 		}
-
+		err = p.postStart()
+		if err != nil {
+			p.logger.Error("[Process] postStart failed: %v", err)
+		}
```
pda, line 210: Minor style thing, if you're not keeping any return values, I believe this is the preferred form:

```suggestion
		if err := p.postStart(); err != nil {
			p.logger.Error("[Process] postStart failed: %v", err)
		}
```

It's less sensitive/leaky to existing/future values of `err`.

pda (dismissed):

pda (approved):

yob:
Oh. I meant to include a note about compatibility too. Here's the win32 functions this adds:

* [CreateJobObject](https://docs.microsoft.com/en-us/windows/win32/api/jobapi2/nf-jobapi2-createjobobjectw) - Added WindowsXP/Server 2003
* [SetInformationJobObject](https://docs.microsoft.com/en-us/windows/win32/api/jobapi2/nf-jobapi2-setinformationjobobject) - Added WindowsXP/Server 2003
* [AssignProcessToJobObject](https://docs.microsoft.com/en-us/windows/win32/api/jobapi2/nf-jobapi2-assignprocesstojobobject)  - Added WindowsXP/Server 2003
* [OpenProcess](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-openprocess) - Added WindowsXP/Server 2003
* [CloseHandle](https://docs.microsoft.com/en-us/windows/win32/api/handleapi/nf-handleapi-closehandle) - Added Windows 2000/Server 2000


I've checked our agent telemetry and the oldest version we've seen in the past month is Windows 7, so I think this is fine

sj26:
This is really cool @yob, well done!

philwo:
This will help us so much! Thanks for adding this feature to the agent!

PR #1330, opened 2020-10-16
add productType to the OS telemetry we send from windows

yob:
This value is required to disambiguate between desktop windows (like windows 7, 8 and 10) and server windows (like server 2012, 2016 and 2019).

Ref: https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/ns-wdm-_osversioninfoexw#remarks

### DIFF
--- a/system/version_dump_windows.go
+++ b/system/version_dump_windows.go
@@ -6,8 +6,27 @@ import (
 	"golang.org/x/sys/windows"
 )
 
+var (
+	VER_NT_WORKSTATION       = byte(0x0000001)
+	VER_NT_DOMAIN_CONTROLLER = byte(0x0000002)
+	VER_NT_SERVER            = byte(0x0000003)
+)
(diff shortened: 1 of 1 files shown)


keithpitt (approved):
Seems legit to me

system/version_dump_windows.go
```diff
@@ -6,8 +6,27 @@ import (
 	"golang.org/x/sys/windows"
 )
 
+var (
+	VER_NT_WORKSTATION       = byte(0x0000001)
+	VER_NT_DOMAIN_CONTROLLER = byte(0x0000002)
+	VER_NT_SERVER            = byte(0x0000003)
+)
```
yob, line 13: surprisingly, these constants aren't in `golang.org/x/sys/windows` yet

ticky (approved):

PR #1332, opened 2020-10-20
Rename (not yet released) shutdown hook to agent-shutdown.

pda:
#1275 introduced an agent-wide `shutdown` hook, with iterations in #1322 and #1326. Those pull requests are merged, but not yet released.

This pull request renamed the hook from `shutdown` to `agent-shutdown`, making it clear that it's different to existing job/bootstrap hooks, and setting a precedent for any future agent-level hooks (e.g. #1057).

/cc @goodspark

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -726,7 +726,7 @@ var AgentStartCommand = cli.Command{
 		pool := agent.NewAgentPool(workers)
 
 		// Agent-wide shutdown hook. Once per agent, for all workers on the agent.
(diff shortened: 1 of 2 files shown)


yob (approved):
I haven't manually tested this change, but I agree with adding the prefix and the diff looks unsurprising :+1:

goodspark (approved):
LGTM!

yob:
I've manually tested this on linux and confirmed a hook named `agent-shutdown` was executed as expected :+1:

PR #1334, opened 2020-11-01
upgrade from golang 1.14 to 1.15

yob:
There's no driving reason to upgrade, but 1.15 has been out for a couple of months and has a few bugfix releases. Maybe it's ready for us to try?


yob:
merged on the assumption this in uncontroversial. On our way to being ready to use 1.16 when it's released (early 2012) to build a macos/arm agent.

PR #1339, opened 2020-11-18
Use gotip (pre 1.16) to compile a native executable for Apple Silicon

yob:
This adds a new build step that compiles a native agent for Apple Silicon. I think. I don't have one to test on.

Cross compiling for Apple Silicon requires `GOARCH=arm64` and `GOOS=darwin`, which is a valid combination on unreleased version of go and will be valid in the upcoming 1.16 release.

We currently compile using go 1.15, so this new step uses gotip to install the latest unreleased golang. That probably means the darwin-arm64 binary it spits out should be considered experimental.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -116,6 +116,16 @@ steps:
         config: .buildkite/docker-compose.yml
         run: agent
 
+  - name: ":mac: arm64"
+    command: ".buildkite/steps/build-apple-silicon-binary.sh darwin arm64" # this can go back to the standard build-binary.sh once go 1.16 is released
+    key: build-binary-darwin-arm64
+    depends_on: test-linux # don't wait for slower windows tests
+    artifact_paths: "pkg/*"
+    plugins:
+      docker-compose#v3.0.0:
+        config: .buildkite/docker-compose.yml
+        run: agent
+
   - name: ":freebsd: amd64"
     command: ".buildkite/steps/build-binary.sh freebsd amd64"
     key: build-binary-freebsd-amd64
@@ -246,6 +256,7 @@ steps:
       - build-binary-linux-arm64
       - build-binary-linux-ppc64le
       - build-binary-darwin-amd64
+      - build-binary-darwin-arm64
       - build-binary-freebsd-amd64
       - build-binary-freebsd-386
       - build-binary-openbsd-amd64
--- a/.buildkite/steps/build-apple-silicon-binary.sh
+++ b/.buildkite/steps/build-apple-silicon-binary.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+set -euo pipefail
+
+echo "--- :${1}: Building ${1}/${2}"
+
+rm -rf pkg
+
+go get golang.org/dl/gotip
+
+gotip download
+
+gotip version
+
+./scripts/build-binary.sh "${1}" "${2}" "${BUILDKITE_BUILD_NUMBER}"
--- a/.buildkite/steps/build-apple-silicon-binary.sh
+++ b/.buildkite/steps/build-apple-silicon-binary.sh
@@ -8,7 +8,7 @@ rm -rf pkg
 
 go get golang.org/dl/gotip
 
-gotip download 272258
+gotip download
 
 gotip version
 
--- a/.buildkite/steps/build-apple-silicon-binary.sh
+++ b/.buildkite/steps/build-apple-silicon-binary.sh
@@ -8,7 +8,7 @@ rm -rf pkg
 
 go get golang.org/dl/gotip
 
-gotip download
+gotip download 272258
 
 gotip version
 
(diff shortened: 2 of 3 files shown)


JuanitoFatas (dismissed):
👍🏼

ticky:
Looks like we’re missing ad-hoc code signing, which is a hard requirement on Apple Silicon.

On macOS (of either flavour) it’s possible to apply an ad-hoc signature using `codesign --sign - --force --preserve-metadata=entitlements,requirements,flags,runtime [executable]` but I’m not sure if it’s yet possible from Linux. Hmmm…

yob:
oooohhh. Interesting 🤔

ticky:
Looks like there’s a golang issue open at golang/go#42684, with the interesting tidbit of https://github.com/isignpy/isign#isign-command-line-arguments existing, which might have an adhoc mode and be applicable on non-Mac platforms

ticky:
See also: there’s a python3 port https://github.com/appknox/isign

yob:
The latest comments on https://github.com/golang/go/issues/42684 are interesting.

It sounds like it doesn't matter what key the executables are signed with, they just need to be signed. I guess they're keen to verify binaries haven't been modified or corrupted, even if they can't verify the identity of the signer/publisher?

isign looks like it can sign with real apple developer keys. I wonder if it will gain the ability to sign with non-apple developer keys? Or maybe a different tool will be released that can do that?

ticky:
The `--adhoc` option appears to do what we need for Apple Silicon (i.e. just create a signature, without any associated signing keys), though I’m yet to be able to verify its output

yob:
I assume `codesign` is shipped with Big Sur, and to do that automatically in the pipeline we'd need a step that runs on a Big Sur machine?

ticky:
`codesign` has been around a while, but I think Big Sur is needed for the `arm64` compatibility, yes. I believe we have or had a MacStadium machine from the [buildkite-mac project](https://github.com/buildkite/buildkite-mac), but it’s been a long time and I’m not sure what the status is of that. @lox might have more recent context on that one?

yob:
It seems highly likely that *many* folks will run into this issue, and there's heaps of activity on the golang issue. Given our amd64 agent runs fine on Rosetta2, I reckon we could sit on our hands for a week or two and see how the landscape looks then.

It might save a lot of pipeline engineering that's ultimately unnecessary.

ticky:
That’s a fair point! I’ll look at getting some testing done sooner than later, in any case :)

ticky:
Okay! With Misty’s help I’ve put this agent through its paces running [the bash-example](https://github.com/buildkite/bash-example); it works!
image

yob:
great success!

Good to know that we're good to go once the codesigning issue is solved. Out of interest, what does the agent show page on buildkite.com look like? Presumably it reports OS of Big Sur and the user agent reports an arm arch?

ticky:
`buildkite-agent/3.25.0.3658 (darwin; arm64)`, with OS `Unknown`… huh, I didn’t expect _that_

ticky:
Oh huh, `os_id ` is `osx_11.1` on that agent, as there’s an 11.1 beta now, which I guess indicates that future major macOS updates will, quite against the grain, in fact be major updates!

yob:
off we go to update the OS ID mappings again (and again, and again) :upside_down_face:

ticky:
PR for the OS mappings is up :)

dbast:
Other go-pipelines enabled signing on macOS in the past with codesign, maybe e.g. this helps https://github.com/git-lfs/git-lfs/pull/4143

ticky:
@dbast yep, something like that’s what we’re looking at, though we currently cross-compile from Linux in all cases. Depending on how tooling settles we may have to change that!

TobiColada (dismissed):

yob:
More activity on https://github.com/golang/go/issues/42684 overnight. It looks like there's a pending changeset that will get go doing the adhoc signing automatically on compilation.

yob:
The upstream issue has been resolved, and a fresh build using gotip results in a binary that runs on M1 :tada: 

I reckon this is close to mergable. I wonder if it's possible to pin gotip to build the current golang master commit consistently, rather than fetch the latest golang master for each build? Consistency would be nice.

yob:
Bummer, I can't see any way to `gotip download <git sha>`

ticky:
what if you [`gotip download 272258`](https://go-review.googlesource.com/c/go/+/272258)?

ticky:
Niiiice ⚡️

ticky:
Oh no 🙈

yob:
using the CL syntax result's in an interactive prompt:

> This will download and execute code from golang.org/cl/272258, continue? [y/n]

yob:
Neat. @keithpitt has confirmed that the binary this spits out runs natively on his new M1 mac :tada: 

I'm going to merge, and we can discuss how and when to ship this to customers elsewhere.

It's unfortunate that the darwin/arm64 builds will be non-deterministic (thanks to always using the latest golang master), but it won't be that way for long. A beta release of 1.16 will be out soon-ish, and we  should be able to switch from gotip to a docker image with the beta.

PR #1340, opened 2020-11-26
install: use the arm64 binary for aarch64 machines

gc-plp:
Enable the install script to use the arm64 binary when uname reports that the machine is aarch64.


yob (approved):
cheers!

PR #1341, opened 2020-11-27
Clarify --include-retries docs and fixes

plaindocs:
- switch options in examples to match usage
- clarify --include-retries and add example

CC: @jayco

### DIFF
--- a/clicommand/artifact_download.go
+++ b/clicommand/artifact_download.go
@@ -22,7 +22,7 @@ Description:
 
 Example:
 
-   $ buildkite-agent artifact download "pkg/*.tar.gz" . --build xxx
+   $ buildkite-agent artifact download --build xxx "pkg/*.tar.gz" .
 
    This will search across all the artifacts for the build with files that match that part.
    The first argument is the search query, and the second argument is the download destination.
@@ -30,9 +30,14 @@ Example:
    If you're trying to download a specific file, and there are multiple artifacts from different
    jobs, you can target the particular job you want to download the artifact from:
 
-   $ buildkite-agent artifact download "pkg/*.tar.gz" . --step "tests" --build xxx
+   $ buildkite-agent artifact download --step "tests" --build xxx "pkg/*.tar.gz" .
 
-   You can also use the step's jobs id (provided by the environment variable $BUILDKITE_JOB_ID)`
+   You can also use the step's jobs id (provided by the environment variable $BUILDKITE_JOB_ID)
+
+   By default, only artifacts from the most recent job in a chain of retried jobs are downloaded.
+   To include artifacts from previous retried jobs, use the "--include-retried-jobs" flag:
+
+   $ buildkite-agent artifact download --include-retried-jobs "pkg/*.tar.gz" .`
 
 type ArtifactDownloadConfig struct {
 	Query              string `cli:"arg:0" label:"artifact search query" validate:"required"`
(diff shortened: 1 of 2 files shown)


clicommand/artifact_download.go
```diff
@@ -22,17 +22,22 @@ Description:
...
    The first argument is the search query, and the second argument is the download destination.
 
    If you're trying to download a specific file, and there are multiple artifacts from different
    jobs, you can target the particular job you want to download the artifact from:
 
-   $ buildkite-agent artifact download "pkg/*.tar.gz" . --step "tests" --build xxx
+   $ buildkite-agent artifact download --step "tests" --build xxx "pkg/*.tar.gz" .
 
-   You can also use the step's jobs id (provided by the environment variable $BUILDKITE_JOB_ID)`
+   You can also use the step's jobs id (provided by the environment variable $BUILDKITE_JOB_ID)
```
ticky, line 35: This feels like a typo?

```suggestion
   You can also use the step's job id (provided by the environment variable $BUILDKITE_JOB_ID)
```

clicommand/artifact_download.go
```diff
@@ -22,17 +22,22 @@ Description:
...
+
+   By default, only artifacts from the most recent job in a chain of retried jobs are downloaded.
+   To include artifacts from previous retried jobs, use the "--include-retried-jobs" flag:
```
ticky, line 38: I wonder if this makes it clearer? “Retried” really feels like it does double duty here:

```suggestion
   By default, only artifacts from the last job in a chain of retried jobs are downloaded.
   To include artifacts from superseded jobs, use the "--include-retried-jobs" flag:
```

ticky (approved):
Liking this change! Got a couple of little bits of feedback on the words (which are mostly my own from the other day 😇)

clicommand/artifact_download.go
plaindocs: I'm less keen on superseded, because if I read that I'd be wondering if it meant something specific in this context. We could just use "from previous jobs"


Huh, that was unintentional:
image

ticky: I’m just not sure “previous jobs” conveys “previous tries of this step” though, it seems ambiguous whether this means “prior jobs in the build” or something else

clicommand/artifact_download.go
```diff
@@ -22,17 +22,22 @@ Description:
...
    jobs, you can target the particular job you want to download the artifact from:
 
-   $ buildkite-agent artifact download "pkg/*.tar.gz" . --step "tests" --build xxx
+   $ buildkite-agent artifact download --step "tests" --build xxx "pkg/*.tar.gz" .
 
-   You can also use the step's jobs id (provided by the environment variable $BUILDKITE_JOB_ID)`
+   You can also use the step's job id (provided by the environment variable $BUILDKITE_JOB_ID)
+
+   By default, only artifacts from the most recent job in a chain of retried jobs are downloaded.
+   To include artifacts from previous retried jobs, use the "--include-retried-jobs" flag:
```
plaindocs, line 38: Yeah, I agree. How about

```suggestion
   To include artifacts also from all retried jobs belonging to this step, use the "--include-retried-jobs" flag:
```

"Also", to make clear that it is as well as the last one. What do you reckon?

plaindocs: CC @ticky

ticky: I’m not sure this solves for the ambiguity in “retried,” and do we otherwise expose the step/job dichotomy to users?

plaindocs: Good point. I'll ponder and reword tomorrow.

plaindocs: We expose step just a couple of lines above, on line 35. But if we're not specifying a job ID, we're just using the artifact path, right, so this flag is just saying "get all the stuff, not just the most recent stuff", so mentioning the step isn't helpful. This might be better:

```suggestion
   To include matching artifacts also from previous retried jobs as well as the most recent, use the "--include-retried-jobs" flag:
```

PR #1351, opened 2020-12-09
Use Docker CLI packages, update Docker Compose, and update centos to 8.x

RemcodM:
This basically replaces #1214 as it seems to be abandoned and we would really like to see the Docker Compose version upgraded.

This PR upgrades de `tini` and `docker-compose` binaries to their newest versions on Ubuntu & CentOS. It also replaces the deprecated `docker.io` package with the installation method described on https://docs.docker.com/engine/install/ubuntu/ for Ubuntu.

The big difference with this PR in comparison to #1214 is the bare inclusion of Docker's CLI packages. No longer is the whole Docker engine included, as I suspect most people will just bind mount the socket inside the Docker container. Please let me know if this is undesirable somehow, then I will add the engine back.

Centos is also bumped from 7.x to 8.x, the current stable release.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,7 +11,7 @@
 /.agent.*.conf
 /.buildkite-agent.cfg
 /.vagrant
-packaging/docker/alpine-linux/buildkite-agent
-packaging/docker/ubuntu-linux/buildkite-agent
+packaging/docker/*/buildkite-agent
+packaging/docker/*/hooks/
 
 .idea
--- a/packaging/docker/alpine-linux/Dockerfile
+++ b/packaging/docker/alpine-linux/Dockerfile
@@ -2,20 +2,20 @@ FROM alpine:3.12
 
 RUN apk add --no-cache \
       bash \
-      git \
-      perl \
-      rsync \
-      openssh-client \
       curl \
-      docker \
+      docker-cli \
       docker-compose \
+      git \
       jq \
-      su-exec \
-      py-pip \
       libc6-compat \
+      perl \
+      py-pip \
+      rsync \
       run-parts \
+      su-exec \
       tini \
-      tzdata
+      tzdata \
+      openssh-client
 
 ENV BUILDKITE_AGENT_CONFIG=/buildkite/buildkite-agent.cfg
 
--- a/packaging/docker/alpine-linux/Dockerfile
+++ b/packaging/docker/alpine-linux/Dockerfile
@@ -8,14 +8,14 @@ RUN apk add --no-cache \
       git \
       jq \
       libc6-compat \
+      openssh-client \
       perl \
       py-pip \
       rsync \
       run-parts \
       su-exec \
       tini \
-      tzdata \
-      openssh-client
+      tzdata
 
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -11,7 +11,7 @@
 /.agent.*.conf
 /.buildkite-agent.cfg
 /.vagrant
-packaging/docker/alpine-linux/buildkite-agent
-packaging/docker/ubuntu-linux/buildkite-agent
+packaging/docker/*/buildkite-agent
+packaging/docker/*/hooks/
```
yob, line 15: :ok_hand:

packaging/docker/alpine-linux/Dockerfile
```diff
@@ -2,20 +2,20 @@ FROM alpine:3.12
...
       libc6-compat \
+      perl \
+      py-pip \
+      rsync \
       run-parts \
+      su-exec \
       tini \
-      tzdata
+      tzdata \
+      openssh-client
```
yob, line 18: it's not hugely important, but if these packages have been rearranged to be alphabetical is openssh-client in the right spot?

packaging/docker/centos-linux/Dockerfile
```diff
@@ -1,24 +1,28 @@
-FROM centos:7.7.1908
+FROM centos:8.3.2011
```
yob, line 1: I'm not very familiar with centos, but is jumping from 7.x to 8.x significant?

Assuming 8.x is stable then I don't see a problem with it (we recently bumped alpine and  ubuntu by major versions as well), but it might be worth calling out in the PR description to raise the visibility of the change.

yob (commented):
Thanks!

I left a few small comments, but no major blockers from my perspective.

yob:
> The big difference with this PR in comparison to #1214 is the bare inclusion of Docker's CLI packages. No longer is the whole Docker engine included, as I suspect most people will just bind mount the socket inside the Docker container. Please let me know if this is undesirable somehow, then I will add the engine back.

I think this is probably fine? I know of **many** folks who mount the host docker socket and/or use docker-in-docker via a sidecar, but I haven't come across folk running the docker daemon within the agent container. I might chat to folks internally to double check my assumptions here.

Also worth noting, but probably not a problem: this seems to upgrade the docker cli from 19.03 to 20.10.

packaging/docker/alpine-linux/Dockerfile
RemcodM: Fixed in new commit.

packaging/docker/centos-linux/Dockerfile
RemcodM: CentOS 8.x is their current stable release: I see no reason not to bump it. I found no issues with it either.

yob (approved):
Lovely!

We're still discussing the docker vs docker-cli question internally, but this PR is a nett win regardless of that discussion so I'll merge.

If we decide the docker daemon should stay, I'll restore it in a follow up PR prior to the next release.

PR #1352, opened 2020-12-21
compile the darwin/arm64 binary using 1.16beta1

yob:
In PR #1339 we started building a native binary for Apple M1 systems, and we used gotip to compile with the latest master version of go.

A 1.16beta1 release is now available, and we can switch to compiling the darwin/arm64 binary with that so we get consistent output.

Eventually, when 1.16 final is released we can use that for all compilation and the special case for darwin/arm64 can go away.


.buildkite/steps/publish-debian-package.sh
```diff
@@ -3,7 +3,7 @@ set -euo pipefail
 
 artifacts_build=$(buildkite-agent meta-data get "agent-artifacts-build" )
 
-# /root/,gnupg is a tmpfs volume, so we can safely store key data there and know
```
yob, line 6: a sneaky unrelated typo fix

yob:
I've downloaded the darwin/arm64 binary from a build of this branch and manually confirmed it boots on an m1 macbook air :rocket:

PR #1354, opened 2021-01-05
Add support for the /opt/homebrew prefix

### LINKED ISSUE #1353: Homebrew release does not use /opt/homebrew/etc/buildkite-agent/buildkite-agent.cfg path
The agent is currently hard-coded to only try to read a config from `/usr/local/etc/buildkite-agent`, but this is not where it is installed via Homebrew on ARM Macs: https://github.com/buildkite/agent/blob/f3fb271317c3f130d3c22d02e7898fc113f623bb/clicommand/agent_start.go#L143

ticky:
Homebrew for ARM Macs will support installation into the /opt/homebrew prefix, instead of /usr/local.

This patch updates our path code to accept both on Apple silicon, but only the former in amd64 builds, allowing for two installations on the same machine if desired.

Fixes #1353

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -130,7 +130,7 @@ func DefaultShell() string {
 }
 
 func DefaultConfigFilePaths() (paths []string) {
-	// Toggle beetwen windows an *nix paths
+	// Toggle beetwen windows and *nix paths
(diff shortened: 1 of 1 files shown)


sj26 (approved):
Nice!

PR #1364, opened 2021-01-22
Add a command to allow removal of existing annotations

ticky:


ticky:
This… somehow breaks the BSD build? 😔

sj26:
It feels odd to me to add this to the existing command. Could we perhaps add a new command?

There's good precedence for `buildlite-agent noun verb`. Perhaps `buildkite-agent annotation remove --context=blah`                                                                                     would be a good choice?

clicommand/annotation_remove.go
```diff
@@ -0,0 +1,121 @@
...
+    cli.StringFlag{
+      Name:   "context",
+      Value:  "default",
+      Usage:  "The context of the annotation used to differentiate this annotation from others",
+      EnvVar: "BUILDKITE_ANNOTATION_CONTEXT",
+    },
+    cli.StringFlag{
+      Name:   "job",
+      Value:  "",
+      Usage:  "Which job should the annotation come from",
```
sj26, line 59:
```suggestion
      Usage:  "Which job is removing the annotation",
```

sj26 (dismissed):
Great! I gave this a go against the corresponding agent API changes and it works a treat. :+1:

clicommand/annotation_remove.go
```diff
@@ -0,0 +1,121 @@
...
+var AnnotationRemoveHelpDescription = `Usage:
+
+   buildkite-agent annotation remove [arguments...]
+
+Description:
+
+   Remove an existing annotation which was previously published using the
+   buildkite-agent annotate command.
+
+   Or if you leave context blank, it will use the default context.
```
ticky, line 21:
```suggestion
   If you leave context blank, it will use the default context.
```

ticky:
Using my privileges to merge this because the code changes since the review were not meaningful.

PR #1365, opened 2021-01-23
Use dd-trace-go >= v1.27.1 to fix multiplatform build

chloeruka:
Will fix build regression introduced in #1273 due to https://github.com/DataDog/dd-trace-go/issues/753


ticky (approved):
Nice! Thank you :)

PR #1366, opened 2021-01-27
Use core.symlinks to attempt symlink testing on Windows

chloeruka:
_This is not a code change since this bootstrapper is managed manually (for now)._ 

The build is currently broken because (my hypothesis) the symlink changes introduced in 029676f131e776a2f580d38cd6ccf94a815067c8 were expected to work on Windows. We need to enable `core.symlinks` to correctly checkout the symlinks on the Windows platform ([here's a StackOverflow](https://stackoverflow.com/questions/5917249/git-symlinks-in-windows/59761201), here be dragons). This _does_ expect that Windows checkouts are run as Administrator, _unless_ [developer mode](https://blogs.windows.com/windowsdeveloper/2016/12/02/symlinks-windows-10/) is already activated on Windows Server Core AMIs (probably not?)


yob (approved):
I'm not really across how symlinks and windows work, but this fixes the build so :+1: from me

chloeruka:
Once I configured it with --system so the Buildkite user account runs with it, it started working (the bootstrapper must run on a different user account). Our test agent stack for Windows already had WindowsAdministrator configured to true.

PR #1369, opened 2021-02-01
Use golang 1.16rc1 for darwin/ARM builds

chloeruka:
Bump us to the latest release candidate just for darwin/ARM64 builds for Apple silicon. We're still waiting on a stable 1.16 release, but this is probably very close.

### DIFF
--- a/.buildkite/docker-compose.yml
+++ b/.buildkite/docker-compose.yml
@@ -15,7 +15,7 @@ services:
   # a temporary service for building a mac+arm binary. We can drop this once the above service
   # is using 1.16+
   agent1-16:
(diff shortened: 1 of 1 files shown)


chloeruka:
[597d85d darwin/arm64 image](https://buildkite.com/organizations/buildkite/pipelines/agent/builds/3712/jobs/338f92e3-9731-42ac-8c92-fe0a9f97cf91/artifacts/c236762e-4415-40f0-b946-c6706b264419)

yob (approved):
I've downloaded the binary and confirmed to runs great on an m1 air :+1:

PR #1371, opened 2021-02-01
systemctl disable via template and symlinks

### LINKED ISSUE #1370: Ubuntu uninstall systemd glob error
I noticed this `Glob pattern passed to enable, but globs are not supported for this.
Invalid unit name "buildkite-agent@*" escaped as "buildkite-agent@\x2a".`

buildkite-agent version 3.26 3674
Ubuntu 20.04 in a VM

```shell
$ sudo apt-get remove buildkite-agent
Reading package lists... Done
Building dependency tree       
Reading state information... Done
The following packages will be REMOVED:
  buildkite-agent
0 upgraded, 0 newly installed, 1 to remove and 3 not upgraded.
After this operation, 36,6 MB disk space will be freed.
Do you want to continue? [Y/n] 
(Reading database ... 160410 files and directories currently installed.)
Removing buildkite-agent (3.26.0-3674) ...
Stopping buildkite-agent systemd service
Glob pattern passed to enable, but globs are not supported for this.
Invalid unit name "buildkite-agent@*" escaped as "buildkite-agent@\x2a".
```

sj26:
`systemctl disable` doesn't accept wildcards (unlike start/stop/etc) but:

> Note that this removes all symlinks to matching unit files, including manually created symlinks, and not just those actually created by enable or link.

https://www.freedesktop.org/software/systemd/man/systemctl.html#disable%20UNIT…

so we can disable the template and that should remove all symlinked instances.

Fixes #1370.

### DIFF
--- a/packaging/linux/scripts/before-remove.sh
+++ b/packaging/linux/scripts/before-remove.sh
@@ -17,7 +17,7 @@ if command -v systemctl > /dev/null; then
   systemctl --no-reload disable buildkite-agent || :
   systemctl stop buildkite-agent || :
 
-  systemctl --no-reload disable "buildkite-agent@*" || :
+  systemctl --no-reload disable "buildkite-agent@" || :
   systemctl stop "buildkite-agent@*" || :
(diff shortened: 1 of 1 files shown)


yob (approved):
:+1:

PR #1374, opened 2021-02-03
collect instance-life-cycle as a default tag on EC2 instances

yob:
When the agent is started with the --tags-from-ec2-meta-data flag it collects a default set of agent tags by querying the instance metadata service.

This adds a new tag to the default set - aws:instance-life-cycle. The tag value will be either "on-demand" or "spot", which can be useful to debugging."on-demand" or "spot", which can be useful to debugging.

The instance-life-cycle metadata key was [added in IMDSv2](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html). I'm not 100% sure what will happen if the agent runs on an instance with only IMDSv1. I added the new code at the end of the method, so I suspect the new tag will just be missing on those agents.

Here's the output starting an agent built with this branch:

    $ ./buildkite-agent start --build-path=~/foo --token=xxx --tags-from-ec2-meta-data=true

                                                     __/ |

    2021-02-03 11:57:51 NOTICE Starting buildkite-agent v3.27.0 with PID: 2175
    2021-02-03 11:57:51 NOTICE The agent source code can be found here: https://github.com/buildkite/agent/v3
    2021-02-03 11:57:51 NOTICE For questions and support, email us at: hello@buildkite.com
    2021-02-03 11:57:51 INFO   Fetching EC2 meta-data...
    2021-02-03 11:57:51 INFO   Successfully fetched EC2 meta-data
    2021-02-03 11:57:51 INFO   Registering agent with Buildkite...
    2021-02-03 11:57:52 INFO   Successfully registered agent "ip-172-31-87-129" with tags [aws:instance-type=t3.medium, aws:ami-id=ami-0885b1f6bd170450c, aws:instance-life-cycle=on-demand, aws:instance-id=i-0d55d5962c54f30b3]
    2021-02-03 11:57:52 INFO   Starting 1 Agent(s)
    2021-02-03 11:57:52 INFO   You can press Ctrl-C to stop the agents
    2021-02-03 11:57:52 INFO   ip-172-31-87-129 Connecting to Buildkite...
    2021-02-03 11:57:52 INFO   ip-172-31-87-129 Waiting for work...

### DIFF
--- a/agent/ec2_meta_data.go
+++ b/agent/ec2_meta_data.go
@@ -40,7 +40,6 @@ func (e EC2MetaData) Get() (map[string]string, error) {
 	if err != nil {
 		return metaData, err
 	}
-
 	metaData["aws:instance-id"] = string(instanceId)
 
 	instanceType, err := c.GetMetadata("instance-type")
@@ -54,6 +53,11 @@ func (e EC2MetaData) Get() (map[string]string, error) {
 		return metaData, err
 	}
 	metaData["aws:ami-id"] = string(amiId)
+
+	instanceLifeCycle, err := c.GetMetadata("instance-life-cycle")
+	if err == nil {
+		metaData["aws:instance-life-cycle"] = string(instanceLifeCycle)
+	}
 
 	return metaData, nil
 }
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -239,7 +239,7 @@ var AgentStartCommand = cli.Command{
 		cli.StringSliceFlag{
 			Name:   "tags-from-ec2-meta-data",
 			Value:  &cli.StringSlice{},
-			Usage:  "Include the default set of host EC2 meta-data as tags (instance-id, instance-type, and ami-id)",
+			Usage:  "Include the default set of host EC2 meta-data as tags (instance-id, instance-type, ami-id, and instance-life-cycle)",
 			EnvVar: "BUILDKITE_AGENT_TAGS_FROM_EC2_META_DATA",
 		},
 		cli.StringSliceFlag{


yob:
Thanks to @vgrigoruk for pointing out this was an option in https://github.com/buildkite/elastic-ci-stack-for-aws/pull/710#issuecomment-772390196

ticky (approved):

yob:
Hmm, there's a test failing on windows:

```

=== Failed
--
  | === FAIL: agent TestPipelineParserSupportsYamlMergesAndAnchors (0.00s)
  | pipeline_parser_test.go:61:
  | Error Trace:	pipeline_parser_test.go:61
  | Error:      	Not equal:
  | expected: "{\"base_step\":{\"type\":\"script\",\"agent_query_rules\":[\"queue=default\"]},\"steps\":[{\"type\":\"script\",\"agent_query_rules\":[\"queue=default\"],\"name\":\":docker: building image\",\"command\":\"docker build .\",\"agents\":{\"queue\":\"default\"}}]}"
  | actual  : "{\"base_step\":{\"type\":\"script\",\"agent_query_rules\":[\"queue=default\"]},\"steps\":[{\"type\":\"script\",\"agent_query_rules\":[\"queue=default\"],\"name\":\":docker: building image\",\"command\":\"docker build .\",\"agents\":{\"queue\":\"default\"}}],\"env\":{\"BUILDKITE_TRACE_CONTEXT\":\"Hv-BBAEBDlRleHRNYXBDYXJyaWVyAf-CAAEMAQwAAAT_ggAA\"}}"
  |  
  | Diff:
  | --- Expected
  | +++ Actual
  | @@ -1 +1 @@
  | -{"base_step":{"type":"script","agent_query_rules":["queue=default"]},"steps":[{"type":"script","agent_query_rules":["queue=default"],"name":":docker: building image","command":"docker build .","agents":{"queue":"default"}}]}
  | +{"base_step":{"type":"script","agent_query_rules":["queue=default"]},"steps":[{"type":"script","agent_query_rules":["queue=default"],"name":":docker: building image","command":"docker build .","agents":{"queue":"default"}}],"env":{"BUILDKITE_TRACE_CONTEXT":"Hv-BBAEBDlRleHRNYXBDYXJyaWVyAf-CAAEMAQwAAAT_ggAA"}}
  | Test:       	TestPipelineParserSupportsYamlMergesAndAnchors
  |  
  | DONE 1245 tests, 13 skipped, 1 failure in 80.502s
```

@chloeruka could this be because the windows agents are now running a 3.27.0 pre-release?

pda:
> Hmm, there's a test failing on windows:

That should be fixed by #1378 … I'll rebase this PR

pda (approved):

pda:
Hmm… I wonder about tag noise/pollution.
What if we set `aws:instance-life-cycle=spot` when it's spot, but leave it unset when it's `on-demand`?

Also:

> The instance-life-cycle metadata key was added in IMDSv2. I'm not 100% sure what will happen if the agent runs on an instance with only IMDSv1. I added the new code at the end of the method, so I suspect the new tag will just be missing on those agents.

I think we should get some test coverage or even just manual testing with IMDSv1… and e.g. what happens if we add _another_ tag after this one, and this one fails to load.

yob:
Good points, I'll manually confirm how this behaves on a IMDSv1-only instance and report back.

> What if we set aws:instance-life-cycle=spot when it's spot, but leave it unset when it's on-demand?

yob:
> Good points, I'll manually confirm how this behaves on a IMDSv1-only instance and report back.

Oh. I don't think IMDSv1-only instances are a thing? At least, not on new instances. Maybe folks might run the agent on an old instance that launched before IMDSv2?

Here's the launch wizard for a new instance:

Screenshot from 2021-02-15 00-14-13

I've manually started a build of the agent from this branch on both a IMDSv1+2 and IMDSv2 instance. In both cases the new tag was added as expected. I guess we're using a version of the go-aws sdk that always uses IMDSv2?

> I think we should get some test coverage or even just manual testing with IMDSv1… and e.g. what happens if we add another tag after this one, and this one fails to load.

Is it possible it won't fail (other than for catastrophic IMDS failure, which can happen but this shouldn't make the agent behave any worse than it probably already does in that scenario)?

sj26:
This seems reasonable to me, although if we're unsure if this piece of meta-data is consistently available then perhaps the error case should be softly handled (i.e. ignore the attribute and move on) instead of returning the error?

yob:
>  if we're unsure if this piece of meta-data is consistently available then perhaps the error case should be softly handled (i.e. ignore the attribute and move on) instead of returning the error

I like it. I'd blindly copied the existing error handling pattern for consistency, but a little caution makes sense. I've pushed your suggested change.

yob:
@sj26 @pda I reckon this is ready for another look

sj26 (approved):
Nice

Issue #1375, opened 2021-02-04
inconsistent usage of `arguments` vs `options` in `usage` strings


nickatsegment:
(From https://github.com/buildkite/docs/issues/916)

Some commands say `[arguments...]`:

```
$ ag 'arguments...' clicommand
clicommand/meta_data_keys.go
15:   buildkite-agent meta-data keys [arguments...]

clicommand/artifact_shasum.go
14:   buildkite-agent artifact shasum [arguments...]

clicommand/meta_data_get.go
15:   buildkite-agent meta-data get <key> [arguments...]

...
```

Others use `[options]`:

```
$ ag '\[options\]' clicommand
clicommand/artifact_upload.go
12:   buildkite-agent artifact upload [options] <pattern> [destination]

clicommand/artifact_download.go
12:   buildkite-agent artifact download [options] <query> <destination>

clicommand/artifact_search.go
17:   buildkite-agent artifact search [options] <query>
```

I found the `[arguments...]` style kind of confusing. To me it means a variable number of _positional_ arguments, not optional arguments aka flags. The fact that it is listed at the end reinforced the idea that they were positional args.

At the very least, it should be one or the other. And my vote is for `[options]` (or `[flags]` :)

nickatsegment:
Thanks!

PR #1376, opened 2021-02-05
Use options for consistency

jayco:
As mentioned by one of our users there is a bit of a mix-match between using arguments and options through the agent.

Closes https://github.com/buildkite/agent/issues/1375


pda (approved):
Tests were failing unrelated to this PR… that's fixed by #1378 … I've rebase this PR.

PR #1377, opened 2021-02-05
Allow using the spawn index in the agent name

ticky:
This allows using the special token `%spawn` in the agent name, which will locally substitute the spawned worker index.

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -10,6 +10,8 @@ import (
 	"os/signal"
 	"path/filepath"
 	"runtime"
+	"strconv"
+	"strings"
 	"sync"
 	"syscall"
 	"time"
@@ -730,6 +732,9 @@ var AgentStartCommand = cli.Command{
 			} else {
 				l.Info("Registering agent %d of %d with Buildkite...", i, cfg.Spawn)
 			}
+
+			// Handle per-spawn name interpolation, replacing %spawn with the spawn index
+			registerReq.Name = strings.ReplaceAll(cfg.Name, "%spawn", strconv.Itoa(i))
 
 			// Register the agent with the buildkite API
 			ag, err := agent.Register(l, client, registerReq)


chloeruka (dismissed):

pda (approved):

sj26 (dismissed):
Wonderful!

The build failures seem unrelated, about windows+tracing.

ticky:
> The build failures seem unrelated, about windows+tracing.

Ah, I think I was on an old master branch, fixing!

pda:
@ticky I was looking into the windows/tracing failures now

ticky:
Oh, you're on it? I'll let you get it sorted then @pda

pda:
This should fix the (unrelated-to-this-PR) test failure: https://github.com/buildkite/agent/pull/1378

pda:
Rebased after #1378 merged in the hopes of a green test suite.

clicommand/agent_start.go
```diff
@@ -731,6 +733,9 @@ var AgentStartCommand = cli.Command{
 				l.Info("Registering agent %d of %d with Buildkite...", i, cfg.Spawn)
 			}
 
+			// Handle per-spawn name interpolation, replacing %spawn with the spawn index
+			registerReq.Name = strings.ReplaceAll(cfg.Name, "%spawn", strconv.Itoa(i))
```
pda, line 737: Interesting… I guess this mutates the shared `api.AgentRegisterRequest` on each `for` iteration.
We could jump through some hoops to copy the request, but there's probably no good reason, unless if/when we decide to do the registrations in parallel, at which point I'm sure we'll scrutinise for things like this.
So… carry on :)

pda (approved):

clicommand/agent_start.go
ticky: That's right, yeah. I didn't see a reason to avoid this right now, but it wouldn't be too hard to swap it to create an instance on each iteration either.

I agree that if we do parallel registrations we will probably instead be passing the config to each worker instance and having it be responsible for the registration, which will avoid the reuse.

PR #1379, opened 2021-02-05
support mips64le architecture target.

houfangdong:
I am going to submit mips64le architecture，The main reasons are:
1、The mips64le architecture is also the mainstream cpu architecture (amd64, arm64), which is used by many users;
2、Golang supports cross compilation, and mips64le is officially supported;
Therefore, I hope that the release package also supports the mips64le architecture, which is convenient for more users.
Build result：
buildkit-agent

Signed-off-by: houfangdong houfangdong@loongson.com

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -116,6 +116,16 @@ steps:
         config: .buildkite/docker-compose.yml
         run: agent
 
+  - name: ":linux: mips64le"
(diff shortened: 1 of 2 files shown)


houfangdong:
@chloeruka，PTAL，thanks.

chloeruka (approved):
Yep, can confirm it's an officially supported [binary target in Go](https://gist.github.com/asukakenji/f15ba7e588ac42795f421b48b8aede63). Thanks!

PR #1382, opened 2021-02-17
Export current plugin name and Configuration

moensch:
Experimental - see https://forum.buildkite.community/t/native-support-for-golang-plugins/776/5

tl;dr:
Exposes two new environment variables to plugins:

* `BUILDKITE_PLUGIN_NAME` - The "env key compatible" version of the current plugin name - that's what's being used to construct the `BUILDKITE_PLUGIN_<name>_<param>` variables
* `BUILDKITE_PLUGIN_CONFIGURATION` - JSON string holding the current plugin's configuration (as opposed to all the plugin configs as found in `BUILDKITE_PLUGINS`

### DIFF
--- a/agent/plugin/plugin.go
+++ b/agent/plugin/plugin.go
@@ -258,6 +258,16 @@ func (p *Plugin) ConfigurationToEnvironment() (*env.Environment, error) {
 	// Sort them into a consistent order
 	sort.Strings(envSlice)
 
+	// Append current plugin name
+	envSlice = append(envSlice, fmt.Sprintf("BUILDKITE_PLUGIN_NAME=%s", formatEnvKey(p.Name())))
+
+	// Append current plugin configuration as JSON
+	configJson, err := json.Marshal(p.Configuration)
+	if err != nil {
+		return nil, err
+	}
+	envSlice = append(envSlice, fmt.Sprintf("BUILDKITE_PLUGIN_CONFIGURATION=%s", configJson))
+
 	return env.FromSlice(envSlice), nil
 }
 
--- a/agent/plugin/plugin_test.go
+++ b/agent/plugin/plugin_test.go
@@ -286,61 +286,83 @@ func TestConfigurationToEnvironment(t *testing.T) {
 
 	envMap, err = pluginEnvFromConfig(t, `{ "config-key": 42 }`)
 	assert.NoError(t, err)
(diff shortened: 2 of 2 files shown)


moensch:
This PR is not tackling the underlying question in the community post regarding native Golang plugins as first class citizens.

In a multi-OS build environment, I would imagine a Golang plugin to be wrapped in something like this:

```bash
# file: pre-command
function get_ostype() {
  if [[ "$OSTYPE" == "linux"* ]]; then
    echo "linux"
  elif [[ "$OSTYPE" == "darwin"* ]]; then
    echo "darwin"
  else
    >&2 echo -e "${RED}ERROR: Cannot determine os type from '$OSTYPE'${NC}"
    return 1
  fi
}

DIR="$(cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd)"

# call <hook>-ostype binary and pass requisite values
BUILDKITE_PLUGIN_CONFIGURATION=$BUILDKITE_PLUGIN_CONFIGURATION BUILDKITE_PLUGIN_NAME=$BUILDKITE_PLUGIN_NAME ./pre-command-$(get_ostype)
```

pda:
I'm just copying a useful bit of info from the [forum reply](https://forum.buildkite.community/t/native-support-for-golang-plugins/776/5) into here:

> The reason we need the current plugin’s configuration in `BUILDKITE_PLUGIN_CONFIGURATION_<name>` and cannot just rely on parsing the full `BUILDKITE_PLUGINS` json is that within the context of the plugin, I don’t actually know which entry in the plugin list is the appropriate one to look at, **especially if the same plugin is called more than once in the same step** (which is technically possible today I think).

yob:
Thanks @moensch, we like this idea and are keen to hear how it goes for you in the real world :sparkles: 

One caveat to be aware of is that plugins implemented in bash also have magic that ensures exported environment variables in the plugin become available in subsequent hooks and the command. This magic obviously won't apply within the golang binary, so if your plugin needs to influence the environment for later hooks or the command you'll need a way to pass them back to the bash shim and re-export them.

yob (approved):

yob:
Oh  yer, I should note that I built a version of the agent from this branch and ran a [silly debug plugin](https://github.com/yob/envprinter-buildkite-plugin) to double-check it works the way it said it would on the tin :+1:

PR #1385, opened 2021-02-22
bump agent version to include -pre

yob:
Bumping this to be higher than the current released version will mean our build/release pipeline starts releasing each main build as an experimental release that's available to customers.


yob:
Motivation: to make #1382 available to @moensch

PR #1386, opened 2021-02-23
Fix the current beta version

ticky:
The current version string isn't valid and doesn't parse on the Buildkite end, so `3.28.0.pre` agents cannot register

This fixes the version string to be valid, and adds a note about valid version strings to version.go


yob (dismissed):
oops, thanks

yob:
ping @moensch

chloeruka (approved):

PR #1388, opened 2021-02-23
Update agent.APIClient interface

### LINKED ISSUE #1387: [chore] agent.APIClient interface is outdated
The `agent.APIClient` interface is behind `github.com/buildkite/agent/v3/api.Client`. It is missing

```go
MetaDataKeys(string) ([]string, *api.Response, error)
StepExport(string, *api.StepExportRequest) (*api.StepExportResponse, *api.Response, error)
```

moensch:
fixes #1387

* added `tools.go` to avoid `github.com/rjeczalik/interfaces` from being removed by `go mod tidy`
* ran `go mod tidy`
* ran `go generate api/client.go`


yob:
Thanks!

This seems reasonable, however it's failing to build on CI. That might be a Ci issue though?

Here's where the build happens: https://github.com/buildkite/agent/blob/ed94b48d575f40558b3b3eb454894f8e0fec38e5/scripts/build-binary.sh#L43

... and here's what I get when I emulate the build locally:

```bash
# this works
$ go build -o buildkite-agent .

# this doesn't
$ go build -o buildkite-agent *.go
go: downloading github.com/rjeczalik/interfaces v0.1.1
go: downloading golang.org/x/tools v0.0.0-20190501045030-23463209683d
found packages main (main.go) and tools (tools.go) in /home/jh/git/bk/agent
tools.go:6:2: import "github.com/rjeczalik/interfaces/cmd/interfacer" is a program, not an importable package
```

moensch:
I believe explicitly referencing .go files no longer honors the build tag I put into `tools.go`.
I'd be ok removing this file from the change. But I'd think it would be "more right" to keep the file and instead adjust the build script.

I am fine either way though and can adjust as needed.

pda:
Yeah I don't think it's normal / desirable to specify `*.go` in `go build`.
I can't remember if I know of a specific reason why we're doing that.

pda:
> added `tools.go` to avoid `github.com/rjeczalik/interfaces` from being removed by `go mod tidy`

Is this a pattern/idiom used widely?
It feels like fighting against tooling, and I'm not 100% clear on the value of keeping the interfacer tool tracked as a build dependency, but if it's used widely enough to be idiomatic then maybe that's fine.

moensch:
> Is this a pattern/idiom used widely?

I'll preface this by stating that I'm totally ok removing it. It's a pattern I follow based on guidance I got from:

* https://marcofranssen.nl/manage-go-tools-via-go-modules/
* https://github.com/golang/go/issues/25922#issuecomment-402918061
* https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module

The rationale behind it is:
There is an interest in ensuring every developer on their workstation uses the same version of a given tool. Hence tracking the version in `go.mod` is desired. If we do _not_ add a `tools.go` file, a `go mod tidy` would always end up removing it.

It is a workaround, in absence of native support in the go tool chain to manage tools.

yob:
Neat, thanks for the extra context.

Let's stick with tools.go and update scripts/build-binary.sh to `go build <stuff> -o $BUILD_PATH/$BINARY_FILENAME .`

moensch:
🙇  - I updated the build script

yob (approved):
thanks!

pda:
Thanks for the `tools.go` background @moensch 👌🏼

PR #1389, opened 2021-02-26
Use new client-side %spawn not server-side %n in default configs

pda:
`%spawn` was introduced in 8f6ee24646d4e1f028845fab6fa987d4cebdc68e / https://github.com/buildkite/agent/pull/1377 and is resolved client-side via `buildkite-agent start --spawn <n>`. 

`%n` is resolved server-side and is considered soft-deprecated.

This PR changes the default configuration files to use `%spawn` rather than `%n`.

This has some potential for mischief if people are running multiple agents with the same name as separate processes, rather than using `--spawn`. For example `name=%hostname-%spawn` with multiple agent processes per machine. Or, if `%hostname` is replaced with a fixed value like `name=acme-%spawn` across multiple machines. In those cases, `%n` would resolve to a different number for each process, while `%spawn` would resolve to `1` in each process.

In the case of multiple agent processes named `%hostname-%spawn` running on a single host as the same user, there's potential for the different agent processes to compete for a checkout directory, overwriting files out from under each other. This is not a recommended configuration, and we believe most/all existing installs will not pick up these new configuration files anyway.

With those risks in mind, this PR is a proposal / discussion, which I'm happy to close without merging if that's the outcome.

### DIFF
--- a/packaging/docker/alpine-linux/buildkite-agent.cfg
+++ b/packaging/docker/alpine-linux/buildkite-agent.cfg
@@ -2,7 +2,7 @@
 #token="xxx"
 
 # The name of the agent
-name="%hostname-%n"
+name="%hostname-%spawn"
 
 # The number of agents to spawn in parallel (default is "1")
 # spawn=1
--- a/packaging/docker/centos-linux/buildkite-agent.cfg
+++ b/packaging/docker/centos-linux/buildkite-agent.cfg
@@ -2,7 +2,7 @@
 #token="xxx"
 
 # The name of the agent
-name="%hostname-%n"
+name="%hostname-%spawn"
 
 # The number of agents to spawn in parallel (default is "1")
 # spawn=1
--- a/packaging/docker/sidecar/buildkite-agent.cfg
+++ b/packaging/docker/sidecar/buildkite-agent.cfg
@@ -2,7 +2,7 @@
 #token="xxx"
 
 # The name of the agent
-name="%hostname-%n"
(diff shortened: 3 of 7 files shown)


ticky:
My gut feel is that this is reasonable to merge, but maybe we want to look at adding locker type behaviour? Make a `.buildkite-agent.lock` PIDfile when a job is running within a checkout directory, bail if it's present and the PID points to a current `buildkite-agent` process, and always clean it up afterwards?

yob:
The config changes for docker in particular should be super safe and uncontroversial, right?

Changing the config for deb/rpm packages should be safe, if we confirm 100% that the config is only used for new installs and not package upgrades.

Windows, mac and manual installs using `install.sh` I'm less familiar with and therefore less confident about whether this will impact them.

pda:
> should be safe, if we confirm 100% that the config is only used for new installs

What about people with automated infrastructure regularly pulling the (latest) package onto new machines? If those people are running multiple processes per host, their stack that worked before could go bad.

yob:
good point

pda:
(That said, it doesn't break expected/intended usage of `--spawn`… we should push out the change somehow… but it might warrant waiting for a major/breaking release number bump, and/or a bit more messaging around the possible gotcha.)

yob (approved):
@sj26 @chloeruka and I just took another look at this and decided to merge so it's included in tomorrow's planned release.

We're confident that the default config in rpm's and deb's will only be used on new deploys.

PR #1390, opened 2021-03-03
Allow signal bootstrap sends to commands on cancel to be customized

brentleyjones:
Currently bootstrap sends a `SIGTERM` to commands when it, itself, is interrupted.

This adds `cancel-signal` which accepts signals in the form of SIGTERM, SIGHUP, etc.

This is the same config that `start` accepts, and is based on the PR that added that feature: https://github.com/buildkite/agent/pull/1041

Finally, `start` passes its `cancel-signal` to `bootstrap`, to prevent having to customize the `bootstrap-script`.

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -536,7 +536,11 @@ var AgentStartCommand = cli.Command{
 
 		// Set a useful default for the bootstrap script
 		if cfg.BootstrapScript == "" {
-			cfg.BootstrapScript = fmt.Sprintf("%s bootstrap", shellwords.Quote(os.Args[0]))
+			cfg.BootstrapScript = fmt.Sprintf(
+				"%s bootstrap --cancel-signal %s",
+				shellwords.Quote(os.Args[0]),
+				cfg.CancelSignal,
+			)
 		}
 
 		// Show a warning if plugins are enabled by no-command-eval or no-local-hooks is set
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -536,11 +536,7 @@ var AgentStartCommand = cli.Command{
 
 		// Set a useful default for the bootstrap script
 		if cfg.BootstrapScript == "" {
-			cfg.BootstrapScript = fmt.Sprintf(
-				"%s bootstrap --cancel-signal %s",
-				shellwords.Quote(os.Args[0]),
-				cfg.CancelSignal,
-			)
+			cfg.BootstrapScript = fmt.Sprintf("%s bootstrap", shellwords.Quote(os.Args[0]))
 		}
 
 		// Show a warning if plugins are enabled by no-command-eval or no-local-hooks is set
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -509,6 +509,11 @@ func (r *JobRunner) createEnvironment() ([]string, error) {
 	env["BUILDKITE_AGENT_EXPERIMENT"] = strings.Join(experiments.Enabled(), ",")
 	env["BUILDKITE_REDACTED_VARS"] = strings.Join(r.conf.AgentConfiguration.RedactedVars, ",")
 
+	// propagate CancelSignal to bootstrap, unless it's the default SIGTERM
+	if r.conf.CancelSignal != process.SIGTERM {
+		env["BUILDKITE_CANCEL_SIGNAL"] = r.conf.CancelSignal.String()
+	}
+
 	// Whether to enable profiling in the bootstrap
 	if r.conf.AgentConfiguration.Profile != "" {
 		env["BUILDKITE_AGENT_PROFILE"] = r.conf.AgentConfiguration.Profile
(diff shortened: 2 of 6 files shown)


chloeruka (approved):
Thanks for your contribution @brentleyjones 🙏🏻 

This is a straightforward implementation and safe improvement to current behaviour. It might be worth adding a test around to confirm the configuration signal reaches the shell configuration, but otherwise these are all well tested corners of the code.

We'll carry out our intention to transition the defaults to `SIGINT` in the future, but for now, it makes sense to add the configuration option and allow us to experiment with the finer details of how the change will affect users. I'm very appreciative of your well-reasoned need for using this to safely dispose of process groups, this is helpful for communicating why we will make this change in the next major release.

pda:
_Excuse me while I think this through “out loud”…_

The existing `buildkite-agent start` command has `--cancel-signal=SIGTERM` which can also be set via `BUILDKITE_CANCEL_SIGNAL` environment instead. This propagates into AgentWorker → JobRunner → process (where it's called InterruptSignal) and is used when `process.Interrupt` is called by JobRunner.Cancel. It sends the configured signal to the _process group_ of the bootstrap process. My limited mental model of process groups is that generally all processes in the group would receive the signal — i.e. the bootstrap and all the commands it's running. Or is it only the process leader? Or “it depends”?

JobRunner TL;DR: on job cancellation, the job runner sends the bootstrap process (and its child processes?) the configured signal.

This PR introduces to `buildkite-agent bootstrap` a new `--cancel-signal=SIGTERM` which can also be set via `BUILDKITE_CANCEL_SIGNAL` environment when the bootstrap is started. This PR changes `buildkite-agent start` to pass its configured cancel signal through to the boostrap. The bootstrap uses this new config to set `b.shell.InterruptSignal`, which passes it to shell's `process.InterruptSignal`, which is used during `Process.Interrupt`, which is called by `Shell.Interrupt`. That's called when the `cancelCh` channel is written to (which logs “Received cancellation signal, interrupting”). That channel is written to when `Bootstrap.Cancel()` is called, which [happens in the bootstrap signal handler](https://github.com/brentleyjones/agent/blob/e020fb9ea5f8dec32bb32e9c89395125bdfadcb3/clicommand/bootstrap.go#L437). That signal handler listens to `SIGHUP`, `SIGTERM`, `SIGINT`, `SIGQUIT`.

Bootstrap TL;DR: bootstrap signal handler calls `bootstrap.Cancel()`  → `b.shell.Interrupt()` → `Process.Interrupt()` which uses this new bootstrap `--cancel-signal` / `BUILDKITE_CANCEL_SIGNAL` instead of being hard-coded to `SIGTERM`.

So I think the intention of this PR is to make `buildkite-agent start --cancel-signal=SIGINT` actually send SIGINT to the build commands being run within the bootstrap. That sounds like it was the intention of the original option, and the bootstrap layer was unintentionally getting in the way. It does that by adding the same `--cancel-signal` option to bootstrap and passing the option through.

So … 👍🏼

My only feedback is that I think we should pass `BUILDKITE_CANCEL_SIGNAL` from JobRunner to Bootstrap via environment (which might already be happening implicitly?) rather than injecting `--cancel-signal` into the bootstrap command ARGV. Thoughts?

clicommand/agent_start.go
```diff
@@ -536,7 +536,11 @@ var AgentStartCommand = cli.Command{
 
 		// Set a useful default for the bootstrap script
 		if cfg.BootstrapScript == "" {
-			cfg.BootstrapScript = fmt.Sprintf("%s bootstrap", shellwords.Quote(os.Args[0]))
+			cfg.BootstrapScript = fmt.Sprintf(
+				"%s bootstrap --cancel-signal %s",
+				shellwords.Quote(os.Args[0]),
+				cfg.CancelSignal,
+			)
```
pda, line 543: This is the part that I think should be propagated via environment rather than CLI args. I think we have other options passing through via environment, perhaps most recently added by @ticky?

pda:
Something else we should think about is either test coverage, or at least a before/after proof copied/pasted into this comment thread showing that this results in the configured signal being sent to the actual commands being run within the bootstrap.

brentleyjones:
>My only feedback is that I think we should pass BUILDKITE_CANCEL_SIGNAL from JobRunner to Bootstrap via environment (which might already be happening implicitly?) rather than injecting --cancel-signal into the bootstrap command ARGV. Thoughts?

That was what I tried first, but it wasn't being passed through. I would be fine with it working that way, the direct argument call was the easiest for me to get it to work.

brentleyjones:
>Something else we should think about is either test coverage, or at least a before/after proof copied/pasted into this comment thread showing that this results in the configured signal being sent to the actual commands being run within the bootstrap.

I wasn't sure how to do the tests, but I would be supportive of that. I can instead create a little example using traps that shows which signals are being sent (it was how I noticed the issue to begin with). Just lmk.

pda:
Sorry for the silence @brentleyjones…

I've pushed a commit to this PR which propagates the option to bootstrap via `BUILDKITE_CANCEL_SIGNAL` env rather than an argv flag; there's precedent for this in `JobRunner.createEnvironment()`.

I was also unsure how to write tests for this, so I did some local testing. I built a simple binary that reports which signal it was killed by. I ran `buildkite-agent start --cancel-signal SIGINT`, and cancelled a build running on the main branch (top) and on this PR branch (bottom):

image

`got signal: interrupt` looks good 👍🏼

pda (approved):

yob:
thanks @brentleyjones :heart:

brentleyjones:
Thanks!

PR #1394, opened 2021-03-16
Propagate --no-pty to the bootstrap --pty flag

pda:
`buildkite-agent start` has a `--no-pty` flag which prevents the bootstrap from being run as a PTY process.

image

Seperately, `buildkite-agent bootstrap` has a default-true `--pty` flag which determines whether commands run _within_ the bootstrap are run as PTY processes.

Presumably, `buildkite-agent start --no-pty` should propagate to `buildkite-agent bootstrap --pty=false`, but that was not happening. That meant that builds running under `buildkite-agent start --no-pty` invoked a bootstrap as a non-PTY process, but the bootstrap invoked commands in PTY mode. On systems with no PTY capability (e.g. AWS Lambda) that was causing severe errors:

image

This pull request propagates the flag into the bootstrap via the internal `BUILDKITE_PTY` environment variable.

Here's that previously failing build step:

image

### DIFF
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -514,6 +514,12 @@ func (r *JobRunner) createEnvironment() ([]string, error) {
 		env["BUILDKITE_AGENT_PROFILE"] = r.conf.AgentConfiguration.Profile
 	}
 
+	// PTY-mode is enabled by default in `start` and `bootstrap`, so we only need
+	// to propagate it if it's explicitly disabled.
+	if r.conf.AgentConfiguration.RunInPty == false {
+		env["BUILDKITE_PTY"] = "false"
+	}
+
 	enablePluginValidation := r.conf.AgentConfiguration.PluginValidation
 	// Allow BUILDKITE_PLUGIN_VALIDATION to be enabled from env for easier
 	// per-pipeline testing


agent/job_runner.go
```diff
@@ -514,6 +514,12 @@ func (r *JobRunner) createEnvironment() ([]string, error) {
 		env["BUILDKITE_AGENT_PROFILE"] = r.conf.AgentConfiguration.Profile
 	}
 
+	// PTY-mode is enabled by default in `start` and `bootstrap`, so we only need
+	// to propagate it if it's explicitly disabled.
+	if r.conf.AgentConfiguration.RunInPty == false {
+		env["BUILDKITE_PTY"] = "false"
+	}
```
pda, line 521: We could unconditionally set `env["BUILDKITE_PTY"]` to stringified `AgentConfiguration.RunInPty`, but I prefered to avoid introducing a new `BUILDKITE_PTY=true` environment variable to all the builds/jobs that haven't explicitly opted out of PTY mode.

keithpitt (approved):

chloeruka (approved):
Nice find!

yob:
Thanks @pda  :heart:

PR #1395, opened 2021-03-16
Configure dependabot to propose update to the golang version we build with

yob:
This configures dependabot to propose updates to the golang version we use, but at most once per week.

The nice thing about this is we can continue to provide precise docker image digests so builds are deterministic (no failures when a docker tag is mutated with an unexpected change), but we'll also be prompted to start using golang security updates.

Unfortunately dependabot doesn't support docker compose files, so I had to add some indirection and move the base image details into a Dockerfile.


yob:
The build is still green so I haven't broken anything, but we can only confirm the dependabot config is correct by merging to main.

I'll do so, and then seek feedback from folks that this is a thing we want :pray:

PR #1401, opened 2021-03-18
output-redactor: redact shell logger, including changed env vars

pda:
Currently the `output-redactor` only applies to `Shell.Writer`, so the output of programs run via the shell are redacted. However `Shell` also has a `Logger` which is used to show the command being run, and — importantly — to show environment variable changes after hooks run. Those environment variable changes may contain secrets, e.g. if `BUILDKITE_REPO` is changed by a hook to include a secret token in the URL.

This PR applies the redactor to `Shell.Logger` too (assuming it's a `shell.WriterLogger`, which is true except in tests). This actually requires setting up a second redactor instance, because shell's default logger outputs separately to `os.Stderr`. To manage two redactors instead of one, a `type RedactorMux []*Redactor` is introduced to multiplex `Flush()` and `Reset()` commands to both. As a bonus, this obviates some `nil` checks, because a zero-length `RedactorMux` can have methods called.

This PR also resets the redactors when environment variables change, so that new potentially-secret values are added as strings to redact.

There's non test coverage for the complete integration, but a new `TestRedactorResetMidStream` demonstrates that a redactor can indeed be flushed and reset with new strings during output streaming.

Thanks @JuanitoFatas for the concept in #1399

Here's a before/after of the bootstrap output:

image-3

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -335,9 +335,8 @@ func (b *Bootstrap) executeHook(ctx context.Context, scope string, name string, 
 
 	b.shell.Headerf("Running %s hook", name)
 
-	if redactor := b.setupRedactor(); redactor != nil {
-		defer redactor.Flush()
-	}
+	redactors := b.setupRedactor()
+	defer redactors.Flush()
 
 	// We need a script to wrap the hook script so that we can snaffle the changed
 	// environment variables
@@ -391,11 +390,11 @@ func (b *Bootstrap) executeHook(ctx context.Context, scope string, name string, 
 	}
 
 	// Finally, apply changes to the current shell and config
-	b.applyEnvironmentChanges(changes.Env, changes.Dir)
+	b.applyEnvironmentChanges(changes.Env, changes.Dir, redactors)
 	return nil
 }
 
-func (b *Bootstrap) applyEnvironmentChanges(environ *env.Environment, dir string) {
+func (b *Bootstrap) applyEnvironmentChanges(environ *env.Environment, dir string, redactors RedactorMux) {
 	if dir != b.shell.Getwd() {
 		_ = b.shell.Chdir(dir)
 	}
@@ -404,6 +403,12 @@ func (b *Bootstrap) applyEnvironmentChanges(environ *env.Environment, dir string
 	if environ == nil || environ.Length() == 0 {
 		return
 	}
+
+	mergedEnv := b.shell.Env.Merge(environ)
+
+	// reset output redactors based on new environment variable values
+	redactors.Flush()
+	redactors.Reset(getValuesToRedact(b.shell, b.Config.RedactedVars, mergedEnv.ToMap()))
 
 	// First, let see any of the environment variables are supposed
 	// to change the bootstrap configuration at run time.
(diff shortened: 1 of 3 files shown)


bootstrap/bootstrap.go
```diff
@@ -335,9 +335,8 @@ func (b *Bootstrap) executeHook(ctx context.Context, scope string, name string,
 
 	b.shell.Headerf("Running %s hook", name)
 
-	if redactor := b.setupRedactor(); redactor != nil {
-		defer redactor.Flush()
-	}
+	redactors := b.setupRedactor()
+	defer redactors.Flush()
```
pda, line 339: `redactors` is a `type RedactorMux []*Redactor` which might be `nil`, but can still safely call `Flush()`.

bootstrap/bootstrap.go
```diff
@@ -405,6 +404,12 @@ func (b *Bootstrap) applyEnvironmentChanges(environ *env.Environment, dir string
 		return
 	}
 
+	mergedEnv := b.shell.Env.Merge(environ)
+
+	// reset output redactors based on new environment variable values
+	redactors.Flush()
```
pda, line 410: It's necessary to `Flush()` the redactor before resetting it, otherwise we'll lose/confuse any partial output that hadn't been terminated with a newline etc. It would be nice if the redactor implementation dealt with this detail, but I wasn't willing to go that deep right now.

bootstrap/bootstrap.go
```diff
@@ -1792,31 +1796,54 @@ func (b *Bootstrap) ignoredEnv() []string {
...
+	// If the shell.Logger is already a redacted WriterLogger, reset the values to redact.
+	// (maybe there's a better way to do two levels of type assertion? ...
+	// shell.Logger may be a WriterLogger, and its Writer may be a Redactor)
+	var shellWriterLogger *shell.WriterLogger
+	var shellLoggerRedactor *Redactor
+	if logger, ok := b.shell.Logger.(*shell.WriterLogger); ok {
+		shellWriterLogger = logger
+		if redactor, ok := logger.Writer.(*Redactor); ok {
+			shellLoggerRedactor = redactor
+		}
```
pda, line 1833: 👆🏼 I didn't like writing this code, and I don't like reading it. It's the closest and simplest I could come to the equivalent of the previous code, but with two different redactors, and with one of them only applicable to a certain concrete type of the shell logger.

It determines whether `b.shell.Logger` is a `shell.WriterLogger` (which it is by default outside of tests), because that's the implementation of shell's `type Logger interface { … }` that has an `io.Writer` which can be wrapped in a redactor.

And if it _is_ a `shell.WriterLogger`, it checks whether it's already wrapped in a redactor, in which case the subsequent code `Reset()`s it instead of wrapping it (just like the previous code did with the original output redactor).

bootstrap/bootstrap.go
```diff
@@ -1792,31 +1796,54 @@ func (b *Bootstrap) ignoredEnv() []string {
...
 
-// Check the redaction config and create a redactor if necessary - may return
-// nil if there's nothing to redact.
-// The redactor is returned so the caller can `defer redactor.Flush()`
-func (b *Bootstrap) setupRedactor() *Redactor {
+// setupRedactor wraps shell output and logging in Redactor if any redaction
+// is necessary based on RedactedVars configuration and the existence of
+// matching environment vars.
+// RedactorMux (possibly empty) is returned so the caller can `defer redactor.Flush()`
+func (b *Bootstrap) setupRedactor() RedactorMux {
```
keithpitt, line 1803: Feels?

```suggestion
func (b *Bootstrap) setupRedactors() RedactorMux {
```

pda: Yeah probably a slight improvement 👍🏼

bootstrap/bootstrap.go
yob: :exploding_head:

yob (approved):
lovely :+1:

PR #1402, opened 2021-03-22
Explicitly set a working directory when we start the bootstrap

yob:
Currently we start the bootstrap process without an explicit working directory. Presumably the operating system assigns it one, maybe the same as the main agent process?

We've seen an example of a customer build where the default working directory for the bootstrap didn't exist. This results in a fatal exception and the job fails to run: https://github.com/buildkite/agent/blob/bed05dda2d7bfb888d80c621af827ee2bce39144/bootstrap/shell/shell.go#L62-L65

Our process abstraction supports setting an explicit working directory for the child process, so we may as well use it: https://github.com/buildkite/agent/blob/bed05dda2d7bfb888d80c621af827ee2bce39144/process/process.go#L127-L132

I've opted to use the agent build path, which should be a directory that exists. There's no real reason for the bootstrap process to start with that as the working directory, other than the fact that it should exist.

Not long after it starts, the bootstrap process will find or create the full checkout path where the source code will be cloned to, and then change its own working directory to that: https://github.com/buildkite/agent/blob/bed05dda2d7bfb888d80c621af827ee2bce39144/bootstrap/bootstrap.go#L842-L846

Our process abstraction wraps [cmd.Exec](https://golang.org/pkg/os/exec/#Cmd), which has this so say about setting the working directory:

```go
// Dir specifies the working directory of the command.
// If Dir is the empty string, Run runs the command in the
// calling process's current directory.
```

### DIFF
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -219,6 +219,7 @@ func NewJobRunner(l logger.Logger, scope *metrics.Scope, ag *api.AgentRegisterRe
 	runner.process = process.New(l, process.Config{
 		Path:            cmd[0],
 		Args:            cmd[1:],
+		Dir:             conf.AgentConfiguration.BuildPath,
 		Env:             processEnv,
 		PTY:             conf.AgentConfiguration.RunInPty,
 		Stdout:          processWriter,
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -25,6 +25,7 @@ import (
 	"github.com/buildkite/agent/v3/logger"
 	"github.com/buildkite/agent/v3/metrics"
 	"github.com/buildkite/agent/v3/process"
+	"github.com/buildkite/agent/v3/utils"
 	"github.com/buildkite/shellwords"
 	"github.com/urfave/cli"
 )
@@ -690,6 +691,15 @@ var AgentStartCommand = cli.Command{
 		cancelSig, err := process.ParseSignal(cfg.CancelSignal)
 		if err != nil {
 			l.Fatal("Failed to parse cancel-signal: %v", err)
+		}
+
+		// confirm the BuildPath is exists. The bootstrap is going to write to it when a job executes,
+		// so we may as well check that'll work now and fail early if it's a problem
+		if !utils.FileExists(agentConf.BuildPath) {
+			l.Info("Build Path doesn't exist, creating it (%s)", agentConf.BuildPath)
+			if err := os.MkdirAll(agentConf.BuildPath, 0777); err != nil {
+				l.Fatal("Failed to create builds path: %v", err)
+			}
 		}
 
 		// Create the API client


yob:
When the bootstrap process fails to start due to an invalid working directory, it looks like this:

Screenshot from 2021-03-23 00-30-31

> Error creating shell: Failed to find current working directory: getwd: no such file or directory

clicommand/agent_start.go
```diff
@@ -692,6 +693,15 @@ var AgentStartCommand = cli.Command{
...
 		}
 
+		// confirm the BuildPath is exists. The bootstrap is going to write to it when a job executes,
+		// so we may as well check that'll work now and fail early if it's a problem
+		if !utils.FileExists(agentConf.BuildPath) {
+			l.Info("Build Path doesn't exist, creating it (%s)", agentConf.BuildPath)
+			if err := os.MkdirAll(agentConf.BuildPath, 0777); err != nil {
+				l.Fatal("Failed to create builds path: %v", err)
+			}
+		}
```
yob, line 703: Is this a controversial thing to do? The bootstrap process will implicitly create the builds directory later on, but it might be worth doing this explicitly in the main agent process and failing early?

Here's where it's done in the bootstrap: https://github.com/buildkite/agent/blob/fce406bd64a0d5a6ba2075931bbfd99cc2aeca65/bootstrap/bootstrap.go#L956-L961

This is include in what was going to be a 1-line PR because the bk cli integration specs specify a BuildPath that doesn't exist, and we're unable to start the bootstrap processes with a working directory that doesn't exist.

agent/job_runner.go
```diff
@@ -219,6 +219,7 @@ func NewJobRunner(l logger.Logger, scope *metrics.Scope, ag *api.AgentRegisterRe
 	runner.process = process.New(l, process.Config{
 		Path:            cmd[0],
 		Args:            cmd[1:],
+		Dir:             conf.AgentConfiguration.BuildPath,
```
yob, line 222: This is the main part of this PR - setting an explicit working directory for the bootstrap process. Everything else in the PR is a change that's required to make this line possible.

yob:
The customer who ran into this error was using the deb package on Ubuntu, and I've been unable to reproduce the error in that context.

However, here's a contrived example using a 3.28.1 agent binary.

Start with this directory structure:
```
$ tree .                                                                                                                   
.                                                                                                                                                                                                                                                                                           
├── buildkite-agent                                                                                                                           
├── builds                                                                                                                                    
├── foo                                                                                                                                                                                                                                                                                     
└── plugins
```

cd into the `foo` directory and start the agent:

```
$ cd foo                                                                                                                   
$ ../buildkite-agent start --build-path=/home/jh/git/bk/agent/tmp/builds --plugins-path=/home/jh/git/bk/agent/tmp/plugins --debug                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                            
                                                 __/ |                                                                                        
                                                                                                                                              
2021-03-24 08:57:19 NOTICE Starting buildkite-agent v3.28.1 with PID: 2555692                                                                 
2021-03-24 08:57:19 NOTICE The agent source code can be found here: https://github.com/buildkite/agent                                        
2021-03-24 08:57:19 NOTICE For questions and support, email us at: hello@buildkite.com                                                        
2021-03-24 08:57:19 INFO   Configuration loaded path=/etc/buildkite-agent/buildkite-agent.cfg                                                 
2021-03-24 08:57:19 DEBUG  Bootstrap command: ../buildkite-agent bootstrap                                                                    
2021-03-24 08:57:19 DEBUG  Build path: /home/jh/git/bk/agent/tmp/builds                                                                                                                                                                                                                     
2021-03-24 08:57:19 DEBUG  Hooks directory: /etc/buildkite-agent/hooks                                                                        
2021-03-24 08:57:19 DEBUG  Plugins directory: /home/jh/git/bk/agent/tmp/plugins                                                               
2021-03-24 08:57:19 INFO   Registering agent with Buildkite...
```

The working directory of the main agent process will be `foo`, and that working directory will be passed on to the bootstrap processes. Let's break it by deleting `foo` (from another terminal) and then starting a new job:

```
$ rm -rf foo
```

The bootstrap process will early exit with a fatal error, and the user will see this:

![Screenshot from 2021-03-24 09-13-33](https://user-images.githubusercontent.com/8132/112225677-4159b500-8c81-11eb-854f-c031d39f7fad.png)

With a build of the agent from this branch, the job will execute just fine even if the `foo` directory is deleted.

shevaun (commented):
Tentative +1 because I have no background with this code, but on the face of it it seems to be a reasonable change

clicommand/agent_start.go
pda: Is `os.MkdirAll(checkoutPath, 0777)` legit on Windows/NTFS?
I imagine so? Is this the same code we would have run later on in the bootstrap anyway? Or different?

pda (approved):
I'm puzzled what edge-case is triggering this, but seems good 👍🏼

clicommand/agent_start.go
ticky: Looks like the shape of the code is the same no matter the OS:

https://golang.org/src/os/file.go?s=7592:7636#L245

yob: > Is this the same code we would have run later on in the bootstrap anyway? Or different?

It's the same code we run in the bootstrap later to create the checkout directory, which implicitly creates the builds directory as well (if it doesn't exist).

Issue #1200, opened 2020-03-16
Hooks always use /bin/bash


keithduncan:
I’ve been running the buildkite agent in some reduced environments like alpine which only has `/bin/sh` (linked to the busybox implementation) but found that there’s a [hard dependency](https://github.com/buildkite/agent/blob/bed05dda2d7bfb888d80c621af827ee2bce39144/bootstrap/shell/shell.go#L279-L281) on `/bin/bash` to invoke hooks.

I tried setting `BUILDKITE_SHELL` that that isn’t used for hooks 😢 

I wasn’t sure if bash was a hard dependency for the before/after environment magic, or if these could use the `BUILDKITE_SHELL` if given?

sj26:
> I wasn’t sure if bash was a hard dependency for the before/after environment magic

This one, I think. :-( We do some `set` commands and some env/`export -p` magic which I think relies on bash-isms. I'm hoping in the future that plugins will support more than bash, but not for the moment sorry.

I think it's reasonable to widen from `/bin/bash` to searching the `$PATH`, though, like [Windows already does](https://github.com/buildkite/agent/blob/bed05dda2d7bfb888d80c621af827ee2bce39144/bootstrap/shell/shell.go#L262-L267).

keithduncan:
Closing my own issue as a duplicate of #1136 😄

PR #1404, opened 2021-03-24
search the path for bash when the bootstrap wants to run a script

yob:
The bootstrap process runs a lot of shell scripts - all hooks and plugins are implemented as shell. Previously the bash path on non-windows was hard coded to /usr/bin, while on windows we search the path for it.

This unifies the behaviour and searches the path on not-windows (macos, linux, etc) as well.

This could be desirable in a number of situations, but one particular use case is on macos where the system bash (/bin/bash) is 3.x and newer versions can be installed elsewhere on the path by homebrew.

Note that the agent does have a `--shell` flag, which can also be set by `BUILDKITE_SHELL` env var. This controls the shell executable that's used to execute the command phase of a job. I considered looking for a way to use that config option in the plugin/hook execution as well. However, that would be complicated. We use some bash-isms in hook and plugin execution so we must use bash, yet the `--shell` flag is documented as being useful when the command phase might need to execute within a
different shell (zsh, busybox, etc).

Maybe one day we'd like to support any posix shell for hook/plugin execution (see #1200), but in the interim this change is a useful one to have.

### DIFF
--- a/bootstrap/shell/shell.go
+++ b/bootstrap/shell/shell.go
@@ -337,7 +337,11 @@ func (s *Shell) RunScript(ctx context.Context, path string, extra *env.Environme
 		args = []string{"-file", path}
 
 	case !isWindows && isBash:
-		command = "/bin/bash"
+		bashPath, err := s.AbsolutePath("bash")
+		if err != nil {
+			return fmt.Errorf("Error finding bash, needed to run scripts: %v.", err)
+		}
+		command = bashPath
 		args = []string{"-c", path}
 
 	default:
--- a/bootstrap/shell/test.go
+++ b/bootstrap/shell/test.go
@@ -36,7 +36,9 @@ func NewTestShell(t *testing.T) *Shell {
 			"ProgramData=" + os.Getenv("ProgramData"),
 		})
 	} else {
-		sh.Env = env.New()
+		sh.Env = env.FromSlice([]string{
+			"PATH=" + os.Getenv("PATH"),
+		})
 	}
 
 	return sh


ticky (approved):

bootstrap/shell/shell.go
```diff
@@ -337,7 +337,14 @@ func (s *Shell) RunScript(ctx context.Context, path string, extra *env.Environme
...
 
 	case !isWindows && isBash:
-		command = "/bin/bash"
+		bashPath, err := s.AbsolutePath("bash")
+		if err != nil {
+			return fmt.Errorf("Error finding bash, needed to run scripts: %v.", err)
+		}
+		if s.Debug {
+			s.Commentf("Found bash at %s", bashPath)
+		}
```
pda, line 346: I wonder if this is too verbose/trivial even for debug logging?
I'm unsure, no strong feels, happy to leave it if it's useful / consistent with other debug logging.

pda (approved):

bootstrap/shell/shell.go
yob: I added it because it mirrors the debug output we have a few lines up for windows. We could remove it for both though?

yob:
d'oh, test failures

bootstrap/shell/shell.go
sj26: I think the windows ones advertise bash or powershell because they diverge. I don't think we need to report finding bash in here.

sj26 (approved):
Looks good! I wouldn't add the debug log of the bash location, but I'm also not strongly against it, your call.

yob:
I've pushed a fix for the test failure and removed the debug logging - is anyone up for a quick check on whether it's an OK approach?

ticky (approved):

sj26 (commented):
Nice! Retroactive :+1:

PR #1405, opened 2021-03-24
Add *_ACCESS_KEY & *_SECRET_KEY to default redactor-var

pda:
`AWS_SECRET_ACCESS_KEY` is a commonly used environment variable which holds credentials that should be redacted, but wasn't matched by the existing defaults `*_PASSWORD,*_SECRET,*_TOKEN`.

Likewise, for custom S3 artifact destinations, `BUILDKITE_S3_SECRET_KEY` is used.

I don't think we should try to match everything under the sun in the default config, but AWS is a big one.


yob (approved):

JuanitoFatas (approved):
Awesome!

PR #1407, opened 2021-03-27
bump version to 3.29.0-beta.1

yob:
We've merged a few bug fixes since 3.28.1, and I'm keen to offer a pre-release version to a few customers so they can confirm the fixes ahead of a formal 3.29.0 release.

I've carefully followed the pre-release version format we used for 3.28.1-beta.1 to avoid a repeat of the build issues I caused a few weeks ago (see PR #1385)


shevaun (approved):
does what it says on the tin!

PR #1421, opened 2021-04-25
add s390x as a target architecture

yob:
A customer is keen to run the agent on IBM Z systems. It's an officially supported architecture for go, so in theory this should just work. Only one way to find out!


yob:
I'm going to merge and then bump the agent version to a 3.30 prerelease. That'll publish the s390x binary to S3 and I can ask the customer to confirm it runs on their system.

PR #1422, opened 2021-04-25
bump version to 3.30.0-beta.1

yob:
The build pipeline will recognise this as a pre-release version and start publishing it through our experimental channels (apt, yum,
binaries on s3).

I've carefully confirmed this version format matches the requirements!


yob:
the primary driver for this is making #1421 available for customer testing

Issue #1420, opened 2021-04-24
[aws mac1.metal] buildkite agents not getting ec2 metadata tags


OliverKoo:
**Environment**

- using buildkite-agent v.3.29
- have `tags-from-ec2-tags=true` and `tags-from-ec2` both set to true and `wait-for-ec2-tags-timeout=300s` in config.
- launching the agent via launchctl on mac1.metal ec2 instance (Big Sur).

**Issue** 

buildkite-agent is set to launch when machine boot. it however does not get the ec2 metadata tags (aws:instance-id, aws:instance-type...etc). Looking at the log I see that it did indeed wait for ec2-tags (the tag I set) but not the metadata tags. fetch retried 5 times then gave up 

```
2021-04-24 00:37:23 WARN   RequestError: send request failed
caused by: Get "http://169.254.169.254/latest/meta-data/instance-id": dial tcp 169.254.169.254:80: connect: network is unreachable (Attempt 5/5 Retrying in 1.465s)
2021-04-24 00:37:24 ERROR  Failed to fetch EC2 meta-data: RequestError: send request failed
```

If I unload the daemon and reload it (restart the agent) then I get all the tags. I am suspecting ec2-metadata tags was not ready yet during fetch)

**Ask**

Is there an equivalent of  `wait-for-ec2-tags-timeout` for metadata tags? seems like the timeout only applies to ec2 tags

sj26:
Sorry @OliverKoo, it looks like we only use that timeout for the resource tags:
https://github.com/buildkite/agent/blob/693b319b4b892e84e2e66daece8a84acd4d1c053/agent/tags.go#L150

I think it'd be totally reasonable to add an equivalent setting for the ec2 meta-data too. If you're keen, you could follow the same pattern and apply it here and submit a pull request:
https://github.com/buildkite/agent/blob/693b319b4b892e84e2e66daece8a84acd4d1c053/agent/tags.go#L103

Otherwise we'll get to it when we can. :-)

OliverKoo:
Here ya go - https://github.com/buildkite/agent/pull/1425

PR #1425, opened 2021-04-26
Add wait ec2 metadata tag timeout config

OliverKoo:
ref https://github.com/buildkite/agent/issues/1420

Currently we have `wait-for-ec2-tags-timeout` config that waits for an EC2 instance's resource tags when buildkite-agent starts, but we don't have an equivalent config that waits for EC2's meta-data tags.

Add a config `wait-for-ec2-meta-data-tags-timeout` to set timeout waiting for ec2 meta-data tags when starting agents.

### DIFF
--- a/agent/tags.go
+++ b/agent/tags.go
@@ -25,6 +25,7 @@ type FetchTagsConfig struct {
 	TagsFromGCPLabels        bool
 	TagsFromHost             bool
 	WaitForEC2TagsTimeout    time.Duration
+	WaitForEC2MetaDataTagsTimeout time.Duration
 	WaitForGCPLabelsTimeout  time.Duration
 }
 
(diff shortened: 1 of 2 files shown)


sj26:
Thanks @OliverKoo, that was quick!

I think we need to call this `wait-for-ec2-meta-data-timeout` (without the tags) because it's about waiting for the ec2 meta-data which is then turned into agent tags. `wait-for-ec2-tags-timeout` is literally about ec2 resource tags which are then also turned into agent tags. I'll push a little commit to rename, and get the indents aligned (you can do this in future with `go fmt`).

sj26:
Great, I'll see if somebody can review this for us, and then merge it so it'll be in the next agent release.

OliverKoo:
Thanks a lot for the review and adjustment!

yob (approved):
Looks good to me :+1: 

This is not a criticism of the PR in any way, but I was surprised at how long it took me to remember the difference between `--tags-from-ec2-meta-data`, `--tags-from-ec2-meta-data-paths` and `--tags-from-ec2-tags`. Once I had that straight, it's pretty clear consistent timeout behavior across them is sensible.

Thanks @OliverKoo!

Issue #1145, opened 2020-01-09
trailing slash in BUILDKITE_ARTIFACT_UPLOAD_DESTINATION duplicates


jufemaiz:
Hi team, love your work!

I've just noticed on my builds that if I set the env `BUILDKITE_ARTIFACT_UPLOAD_DESTINATION` with a trailing slash, it duplicates the slash when joining to an artifact path.

e.g.
```
export BUILDKITE_ARTIFACT_S3_BUCKET="sweet-buildkite-artifacts"
export BUILDKITE_ARTIFACT_S3_PATH="$BUILDKITE_PIPELINE_SLUG/$BUILDKITE_BRANCH/$BUILDKITE_BUILD_NUMBER/"
export BUILDKITE_ARTIFACT_UPLOAD_DESTINATION="s3://$BUILDKITE_ARTIFACT_S3_BUCKET/$BUILDKITE_ARTIFACT_S3_PATH"
```

yields a file `test.txt` uploaded in `.artifacts` to be found at:

```
https://sweet-buildkite-artifacts.s3.amazonaws.com/repository-name/branch/name/build-number//.artifacts/test.txt
```

S3 however removes the duplicate slash (unsure of other destination services).

shevaun:
Thanks @jufemaiz for the report and sorry it's taken us so long to address it! This has been fixed now and will be released in the next version of the agent :)

PR #1427, opened 2021-05-06
Trim trailing slash from the bucketPath

shevaun:
If a user passes a destination with a trailing slash, the artifact url ends up with a double slash in it (between the bucketPath and artifact path).

This trims the trailing slash before setting the bucketPath, so the rest of the code doesn't have to account for whether it has a trailing slash or not.

See also: https://github.com/buildkite/agent/issues/1145

### DIFF
--- a/agent/s3_uploader_test.go
+++ b/agent/s3_uploader_test.go
@@ -13,6 +13,7 @@ func TestParseS3DestinationBucketPath(t *testing.T) {
 	}{
 		{"s3://my-bucket-name/foo/bar", "foo/bar"},
 		{"s3://starts-with-an-s/and-this-is-its/folder", "and-this-is-its/folder"},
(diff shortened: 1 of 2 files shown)


agent/s3_uploader_test.go
```diff
@@ -13,6 +13,7 @@ func TestParseS3DestinationBucketPath(t *testing.T) {
 	}{
 		{"s3://my-bucket-name/foo/bar", "foo/bar"},
 		{"s3://starts-with-an-s/and-this-is-its/folder", "and-this-is-its/folder"},
+		{"s3://custom-s3-domain/folder/ends-with-a-slash/", "folder/ends-with-a-slash"},
```
yob, line 16: noice :+1:

yob (approved):

PR #1430, opened 2021-05-12
Use /usr/pkg/bin/bash as default shell on NetBSD

ibuclaw:
I'm [in the process of adding](https://github.com/creack/pty/pull/117) pty support on NetBSD to the upstream dependency - which I note has also changed location.

Meanwhile, this is a tweak to the default shell location that allows the agent to run with no configuration (other than token).

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -128,8 +128,10 @@ func DefaultShell() string {
 	switch runtime.GOOS {
 	case "windows":
 		return `C:\Windows\System32\CMD.exe /S /C`
-	case "freebsd", "openbsd", "netbsd":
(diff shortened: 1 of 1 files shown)


yob (approved):
neat, thanks!

I'm not familiar with netBSD, but I assume this will work great.

> pty support on NetBSD to the upstream dependency - which I note has also changed location.

oh, interesting. I think we'd happily accept a PR that updates us to use the new location.

ibuclaw:
> I'm not familiar with netBSD, but I assume this will work great.

As per [documentation](http://www.netbsd.org/docs/pkgsrc/using.html), binary installed packages are put into `/usr/pkg`.

PR #1431, opened 2021-05-16
Replace kr/pty with creack/pty

ibuclaw:
This Go module has been moved to creak/pty, the latest release of which adds support for NetBSD (following on from #1430).


yob (approved):

yob:
thanks!

yob:
I was going to link you to a pre-release binary for netbsd to test with, but I just noticed netbsd isn't part of our build matrix. Would it be useful to add it or are you happy building your own version?

ibuclaw:
> I was going to link you to a pre-release binary for netbsd to test with, but I just noticed netbsd isn't part of our build matrix. Would it be useful to add it or are you happy building your own version?

If you are interested in expanding your binaries I'm glad to test. I assume you make these using a cross-compiler?

Currently I'm building my own as I'm overriding the hostname with a hardcoded one - there's no reason why I can't set it in the OS though.

PR #1433, opened 2021-05-17
Option to send Datadog Metrics as Distribution

amukherjeetwilio:
Datadog has added support for the ["Distribution"](https://docs.datadoghq.com/metrics/distributions/) type of metric.

As a summary, the existing "timing" types in Datadog calculate things
like average, min, max, p95 on the client itself.  Therefore, it only
has visibility in what the client has access to. In the context of a
build job, this means that, in most cases, average == min == max == p95,
because they'll only be one job per flush interval.

A Distribution, instead, calculates all of these metrics server side at
the Datadog level itself. This makes it possible to aggregate metrics
across a wide variety of clients or sources. For example, you could have
a hundred Buildkite agents submit their job duration as raw data to
Datadog. Datadog will take the responsibility of calculating all the
rollup values based on the raw data.

please note, first real PR for this repo, so am not fully sure if I'm missing any obvious requirements (tests, etc)

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -54,51 +54,52 @@ Example:
 // - Into clicommand/bootstrap.go to read it from the env into the bootstrap config
 
 type AgentStartConfig struct {
-	Config                     string   `cli:"config"`
-	Name                       string   `cli:"name"`
-	Priority                   string   `cli:"priority"`
-	AcquireJob                 string   `cli:"acquire-job"`
-	DisconnectAfterJob         bool     `cli:"disconnect-after-job"`
-	DisconnectAfterIdleTimeout int      `cli:"disconnect-after-idle-timeout"`
-	BootstrapScript            string   `cli:"bootstrap-script" normalize:"commandpath"`
-	CancelGracePeriod          int      `cli:"cancel-grace-period"`
-	BuildPath                  string   `cli:"build-path" normalize:"filepath" validate:"required"`
-	HooksPath                  string   `cli:"hooks-path" normalize:"filepath"`
-	PluginsPath                string   `cli:"plugins-path" normalize:"filepath"`
-	Shell                      string   `cli:"shell"`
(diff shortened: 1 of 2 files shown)


clicommand/agent_start.go
```diff
@@ -408,6 +409,11 @@ var AgentStartCommand = cli.Command{
 			EnvVar: "BUILDKITE_METRICS_DATADOG_HOST",
 			Value:  "127.0.0.1:8125",
 		},
+		cli.BoolFlag{
+			Name:   "metrics-datadog-distributions",
+			Usage:  "Use Datadog Distributions for Timing metrics",
+			EnvVar: "BUILDKITE_METRICS_DATADOG_DISTRIBUTIONS",
```
amukherjeetwilio, line 415: not sure on the naming here, if this is appropriate or not

metrics/metrics.go
```diff
@@ -87,7 +89,19 @@ func (s *Scope) Timing(name string, value time.Duration, tags ...Tags) {
...
+		// Datadog recommends that, as distributions are a new distinct metric,
+		// they belong to a new metric name. We handle this by just slamming
+		// .distribution to end of all metrics that we submit this way
+		if !strings.HasSuffix(name, ".distribution") {
+			name = name + ".distribution"
+		}
+		err = s.c.client.Distribution(name, float64(value.Milliseconds()), mergedTags, 1)
+	} else {
+		err = s.c.client.Timing(name, value, mergedTags, 1)
+	}
```
amukherjeetwilio, line 103: also just worth pointing out that I don't really golang in my day job -- not sure if there's a more appropriate idiomatic way to do the error handling in this block

chloeruka: this is pretty canonical from what I understand

chloeruka (approved):
Feels like a good low-touch improvement! Thanks for your contribution @amukherjeetwilio

PR #1434, opened 2021-05-17
Send queue metrics to Datadog when job received

amukherjeetwilio:
We would like to also send how long a job was queued for when a job is
acquired by the agent to Datadog. This will empower teams to validate
that their clusters are not causing jobs to queue for "too long" (for
the users definition of that) on the assumption that all jobs will
eventually execute on an agent. Note that this :see_no_evil: that jobs
get stuck indefinitely.

This is not a complete Pull Request as we are waiting for our friends in
the Buildkite team to add the actual "scheduled_at" time to the acquire
job API, which will enable this pull request to actually calculate how
long the job was sitting in the queue for.

### DIFF
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -250,6 +250,17 @@ func (r *JobRunner) Run() error {
 
 	startedAt := time.Now()
 
+	// Publish metric for how long this job was in the queue for, if we can
+	// calculate that
+	if r.job.RunnableAt != "" {
+		runnableAt, err := time.Parse(time.RFC3339Nano, r.job.RunnableAt)
+		if err != nil {
+			r.logger.Error("Metric submission failed to parse %s", r.job.RunnableAt)
+		} else {
+			r.metrics.Timing("queue.duration", startedAt.Sub(runnableAt))
+		}
+	}
+
 	// Start the build in the Buildkite Agent API. This is the first thing
 	// we do so if it fails, we don't have to worry about cleaning things
 	// up like started log streamer workers, etc.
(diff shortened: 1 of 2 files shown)


agent/job_runner.go
```diff
@@ -248,7 +249,9 @@ func NewJobRunner(l logger.Logger, scope *metrics.Scope, ag *api.AgentRegisterRe
 func (r *JobRunner) Run() error {
 	r.logger.Info("Starting job %s", r.job.ID)
 
+	scheduledAt := time.Now().Add(time.Duration(-1*rand.Intn(6400)) * time.Millisecond)
```
amukherjeetwilio, line 252: this is just a placeholder while we wait for the API, and is only here for demonstration purposes

agent/job_runner.go
```diff
@@ -250,6 +250,17 @@ func (r *JobRunner) Run() error {
...
+	// calculate that
+	if r.job.RunnableAt != "" {
+		runnableAt, err := time.Parse(time.RFC3339Nano, r.job.RunnableAt)
+		if err != nil {
+			r.logger.Error("Metric submission failed to parse %s", r.job.RunnableAt)
+		} else {
+			r.metrics.Timing("queue.duration", startedAt.Sub(runnableAt))
+		}
+	}
+
```
amukherjeetwilio, line 263: Does it feel appropriate to put this just before we make the "start job" request? Should it be after we "start job" on the Buildkite API?

keithduncan: Good thinking! I’ve been looking at this and agree that the metric should be submitted after `startJob` returns success.

A case I thought of where startJob could fail, is when multiple agents are started with `--acquire-job` for a given job’s UUID. In the event an agent fails to grab the job because another agent has already grabbed it, the latter agent’s start time isn’t the one that matters.

I’ll make that change now and merge this so it’s included in tomorrow’s release. 🎉

PR #1440, opened 2021-05-19
Resolve the absolute path to the buildkite-agent binary for bootstrap

### LINKED ISSUE #1439: Path used to run buildkite-agent binary is reused without resolving the absolute path
It is possible, particularly when using [the generic Linux/Unix installation method](https://buildkite.com/docs/agent/v3/linux), to call the `buildkite-agent` with a command line which, due to bootstrap scripts running from a build-specific directory, are no longer available, resulting in a "fork/exec: no such file or directory" error when running a job. One such command line which causes this problem is running the agent using `.buildkite-agent/bin/buildkite-agent` instead of `~/.buildkite-agent/bin/buildkite-agent`; the former path is only valid when your current working directory is `~`, while the latter will be valid from any working directory as long as the user is the same.

This is due to this piece of code which pulls in `os.Args[0]` (the name of the binary as used on the command line) and uses that as the path to the bootstrap script:

https://github.com/buildkite/agent/blob/a4a804fd3ff2ca614b414c33452eb8caf0c07977/clicommand/agent_start.go#L547-L550

It seems as though it ought to follow the lead of this other piece of code, which resolves the absolute path in order to grab the configuration file:

https://github.com/buildkite/agent/blob/a4a804fd3ff2ca614b414c33452eb8caf0c07977/clicommand/agent_start.go#L161-L167

ticky:
Previously, the built-in bootstrap subcommand was called using the raw value of `os.Args[0]`, which meant it could fail if the path was relative to the current working directory, as jobs run in a build-specific directory.

This patch updates it to use `filepath.Abs` to resolve an absolute version of that path, so, for example, calling `.buildkite-agent/bin/buildkite-agent` from the user's home directory would no longer fail to run jobs.

It also treats an error resolving this path as fatal, though looking at the code for it this should only really happen if the current working directory is deleted, but given this happens at agent start that is extremely unlikely.

Fixes #1439

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -546,7 +546,14 @@ var AgentStartCommand = cli.Command{
 
 		// Set a useful default for the bootstrap script
 		if cfg.BootstrapScript == "" {
-			cfg.BootstrapScript = fmt.Sprintf("%s bootstrap", shellwords.Quote(os.Args[0]))
+			pathToBinary, err := filepath.Abs(os.Args[0])
+			if err != nil {
+				// This error *likely* means the working directory got deleted,
+				// which is possible, but given this runs at startup, unlikely
+				// https://github.com/golang/go/blob/6c1c055d1ea417d050503efe92c1eead0da68cef/src/path/filepath/path.go#L238-L256
+				l.Fatal("Could not determine absolute path to buildkite-agent binary")
+			}
+			cfg.BootstrapScript = fmt.Sprintf("%s bootstrap", shellwords.Quote(pathToBinary))
 		}
 
 		// Show a warning if plugins are enabled by no-command-eval or no-local-hooks is set


yob (approved):

yob:
I think this might've broken the agent on my local system. I have the agent installed via the deb package, with the binary at `/usr/bin/buildkite-agent`:

```
$ which buildkite-agent 
/usr/bin/buildkite-agent
```

Because it's my local laptop I typically don't have the agent running, but I do start it manually from time to time to test things out, often just from whatever rando directory my terminal is in.

Since this merged, if I start the agent from my home directory like this:

```
$ pwd
/home/jh

$ buildkite-agent start
2021-05-28 09:23:20 DEBUG  Enabled experiment `output-redactor`

                                                 __/ |
```

... then the bootstrap process fails to start and I see this in the job log:

Screenshot from 2021-05-28 09-24-12

If I downgrade from edge to 3.29.0 then the same steps result in the bootstrap successfully starting.

ticky:
Hmmmm, might be missing a path lookup, in that case. I could've sworn I tested that case

ticky:
my gut feel is that we need a https://golang.org/pkg/os/exec/#LookPath call before we do the path joining, and if LookPath has a value we use that instead

yob:
#1444 is our new attempt at this

PR #1441, opened 2021-05-26
Re-add Ubuntu 18.04 support

keithduncan:
This renames the `ubuntu` packaging environment to `ubuntu-20.04` and re-adds an 18.04 environment called `ubuntu-18.04`. The `ubuntu` image tag lives on and will point to the `ubuntu-20.04` variant images.


.buildkite/steps/build-docker-image.sh
```diff
@@ -54,7 +54,7 @@ codename="${3:-}"
 version="${4:-}"
 push="${PUSH_IMAGE:-true}"
 
-if [[ ! "$variant" =~ ^(alpine|ubuntu|centos|sidecar)$ ]] ; then
+if [[ ! "$variant" =~ ^(alpine|ubuntu-18\.04|ubuntu-20\.04|centos|sidecar)$ ]] ; then
```
keithduncan, line 57: I debated calling the variants and folders "ubuntu-bionic" and "ubuntu-focal" to have fewer escaping concerns where the names are used, but decided against it as the variant name is used on in the publishing script which I wanted to avoid changing too much.

.buildkite/steps/publish-docker-image.sh
```diff
@@ -61,6 +61,14 @@ if [[ "$codename" == "stable" ]] ; then
   done
   release_image "${variant}"
 
+  # publish bare 'ubuntu' only from ubuntu-20.04
+  if [[ "$variant" == "ubuntu-20.04" ]] ; then
```
keithduncan, line 65: I reckon duplicating the the loop above here makes more sense than embedding an `if` statement in the existing loop and again where we do the 'bare' variant release. This way there is only a single branch statement to update when we want to update the `ubuntu` tag series to point to a 22.04 release.

.buildkite/steps/upload-release-steps.sh
```diff
@@ -28,7 +28,8 @@ trigger_step() {
         agent-version-build: "${build_version}"
         agent-version-full:  "${full_agent_version}"
         agent-docker-image-alpine: "${agent_docker_image_alpine}"
-        agent-docker-image-ubuntu: "${agent_docker_image_ubuntu}"
+        'agent-docker-image-ubuntu-18.04': "${agent_docker_image_ubuntu_bionic}"
```
keithduncan, line 31: These YAML keys shouldn’t need wrapping in quotes, but yam never know

shevaun (approved):
Seems reasonable to me!

yob:
This is [working well for edge releases at least](https://hub.docker.com/r/buildkite/agent/tags?page=1&ordering=last_updated&name=ubuntu), which bodes well for stable releases!

Screenshot from 2021-05-27 22-43-11

PR #1443, opened 2021-05-28
Revert "Merge pull request #1440 from buildkite/resolve-absolute-agent-path

yob:
This reverts PR #1440. As discussed in the comments of that PR we found it introduced a new issue and we'd like some more time to evaluate the best response without holding up releasing 3.30.0


chloeruka (approved):
@pda @yob @keithduncan and I have reviewed this and decided that [os.executable](https://golang.org/src/os/executable_procfs.go) might be the better path for fixing the underlying issue being reverted here. For now, we're going to revert this so we can unblock the 3.30.0 release and create a patch later.

PR #1444, opened 2021-05-28
Use os.Executable() to detect path to the agents own binary

yob:
We need the path to start the bootstrap as a child process. We've previously detected it using os.Args[0], however there are some edge cases where that doesn't work (eg #1439).

We think os.Executable() might be a more reliable way to detect the path.

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -552,7 +552,11 @@ var AgentStartCommand = cli.Command{
 
 		// Set a useful default for the bootstrap script
 		if cfg.BootstrapScript == "" {
-			cfg.BootstrapScript = fmt.Sprintf("%s bootstrap", shellwords.Quote(os.Args[0]))
+			exePath, err := os.Executable()
+			if err != nil {
+				l.Fatal("Unable to find executable path for bootstrap")
+			}
+			cfg.BootstrapScript = fmt.Sprintf("%s bootstrap", shellwords.Quote(exePath))
 		}
 
 		// Show a warning if plugins are enabled by no-command-eval or no-local-hooks is set


yob:
I haven't fully tested this yet, but I have confirmed the following works and can run jobs on linux:

1. `./buildkite-agent start`
2. Copy the binary to `/usr/local/bin/buildkite-agent` and running `buildkite-agent` with my home directory as PWD
3. ./tmp/buildkite-agent start`

keithduncan (approved):
I was just starting a branch to do this and saw it come down in a git fetch 🥳

clicommand/agent_start.go
```diff
@@ -552,7 +552,11 @@ var AgentStartCommand = cli.Command{
 
 		// Set a useful default for the bootstrap script
 		if cfg.BootstrapScript == "" {
-			cfg.BootstrapScript = fmt.Sprintf("%s bootstrap", shellwords.Quote(os.Args[0]))
+			exePath, err := os.Executable()
+			if err != nil {
+				l.Fatal("Unable to find executable path for bootstrap")
```
yob, line 557: I'm not  sure what kind of error might happen :man_shrugging:

pda: Yeah, it would probably be unusual / edge-case / exotic errors, given it's implemented differently for plan9, freebsd, solaris, darwin, linux/procfs, and windows. Some of them mention potential buffer size problems, I guess the procfs one depends on procfs not being broken and/or having the right permissions to read `/proc/self/exe`, etc.

pda: e.g. running buildkite-agent in a poorly-configured chroot on my Kindle, I encountered all sorts of errors that wouldn't normally happen, e.g. when `/dev/random` doesn't exist or isn't accessible, and I may have hit some procfs issues too.

Issue #1438, opened 2021-05-18
Documentation should describe download destination argument


yonran:
`buildkite-agent artifact download <pattern> <destination>` has a pretty counterintuitive [algorithm](https://github.com/buildkite/agent/blob/v3.29.0/agent/download.go#L71-L96) for determining the destination path. Probably the [documentation](https://buildkite.com/docs/agent/v3/cli-artifact#downloading-artifacts) (same in [v2](https://buildkite.com/docs/agent/v2/cli-artifact#downloading-artifacts)) should be expanded more fully so people can use it.

In particular, the rules appear to be:
* If `destination` ends in /, or if the last part of `destination` does not match the first part of `path`, then `destination` is joined with `path`
* If the last part of `destination` matches the first part of `path`, then the last part of destination is removed before joining with `path`
* If the the last _n_ parts (counting from 0) of `destination`, reversed, match the first 100 × _n_ parts of `source`, then those last _n_ parts of `destination` are removed before joining with `path`

I created a test case to verify these rules:

a_test.go:

```go
package a
import (
	"testing"
	"github.com/stretchr/testify/assert"
)
func TestGetTargetPath(t *testing.T) {
	// If the destination ends in /,
	// then no part of destination is stripped
	// before joining with path
	assert.Equal(t, "a/a", GetTargetPath("a", "a/"))
	assert.Equal(t, "c/a/a/b", GetTargetPath("a/b", "c/a/"))
	// if the last part of destination
	// does not match the first part of path,
	// then just join path to destination
	assert.Equal(t, "a", GetTargetPath("a", "."))
	assert.Equal(t, "a/b/c", GetTargetPath("a/b/c", "."))
	assert.Equal(t, "a/b/a/b", GetTargetPath("a/b", "a/b"))
	// If the last part of destination
	// matches the first part of path,
	// then remove the last part of destination
	// before joining with path
	assert.Equal(t, "a", GetTargetPath("a", "a"))
	assert.Equal(t, "c/a/b", GetTargetPath("a/b", "c/a"))
	assert.Equal(t, "lambda.zip", GetTargetPath("lambda.zip", "lambda.zip"))
	// when the path[0], path[100] matches
	// destination[-1], destination[-2]
	// then the last 2 characters of destination
	// are removed before joining path
	// This is a tricky one to use.
	assert.Equal(t,
[...]

PR #1446, opened 2021-06-01
Document artifact download behaviours and add tests for the variants

keithduncan:
In #1438 @yonran highlighted a surprising agent download behaviour where path/pattern and destination match.

Single deep these would collapse as expected and the file would be downloaded to the expected location. Appending a trailing slash would nest the path/pattern in the destination regardless of matching name. The surprising case is when path/pattern and destination are more than single depth and equal, as the first part of path didn’t match the last part of destination they would be nested.

I discussed this with a couple of folks internally and several of us have run into this gotcha. I think that’s a good sign that this isn’t a "least surprise" behaviour, and it isn’t aping a particular implementation that I can find. I traced this behaviour back to 2014 in [First cut at downloading](https://github.com/buildkite/agent/commit/512fed97cd932d1a0db3305871c1eb2bfa185f89).

However, this change is not backwards compatible. The specific change is highlighted in the tests diff in https://github.com/buildkite/agent/commit/5565d2100d2a64518682d2b1745b38f06e68edb3. Currently, `buildkite-agent artifact download foo/bar foo/bar` will create a file `$PWD/foo/bar/foo/bar`. With this change those arguments would instead create a file `$PWD/foo/bar` (indeed as this change intends, though not without potential consequences).

The more I’ve noodled on this the more I’ve started to renege on my own change 🙈 perhaps a tar style `--strip-components`  argument would be easier for folks to adopt, allowing you to select across the whole set of artifacts, then drop some number of leading path components, and write the remaining set to a given destination directory.

--

Edit: I’ve dropped the behaviour change and have gone back to just docs + tests.


yob:
> The more I’ve noodled on this the more I’ve started to renege on my own change see_no_evil perhaps a tar style --strip-components argument would be easier for folks to adopt, allowing you to select across the whole set of artifacts, then drop some number of leading path components, and write the remaining set to a given destination directory.

I agree that a breaking change in the behaviour is risky - it's bound to bite someone.

I don't think we have an official list of breaking changes we'd like to make in a 4.x release - maybe we could start one and add this? We could potentially make the breaking change behind an `--experiment` flag in the meantime and folks can opt-in that way? `--strip-components` or similar could be fine too.

I guess the other thing we can do that isn't a breaking change is expand the cli docs for the subcommand. It seems like explaining the complexities clearly might be a challenge though :grimacing:

clicommand/artifact_download.go
```diff
@@ -13,12 +13,27 @@ var DownloadHelpDescription = `Usage:
...
 
    Note: You need to ensure that your search query is surrounded by quotes if
-   using a wild card as the built-in shell path globbing will provide files,
-   which will break the download.
+   using a wild card as the built-in shell path globbing will expand the wild
+   card and break the query.
+
+   If the last path component of <destination> matches the first path component
+   of your <query>, the last component of <destination> is dropped from the
+   final path e.g. a query of 'app/logs/*' with a destination of 'foo/app' will
```
harrietgrace, line 25: This looks great! The only thing I'd change is to make this a full stop and start the example with For example instead of e.g. here and in the next para. But that's personal preference, feel free to ignore 😊

keithduncan: Done!

keithduncan:
I’ve removed the behaviour change and gone with documenting the existing behaviour along with adding the tests from #1438, this is ready for a final review

yob (approved):

yob:
oh no, looks like the new tests fail on windows?

keithduncan:
All working! I just excluded the test file because the paths are all hard coded unix style.

PR #1452, opened 2021-06-15
Remove the experiment flag from output redactor

keithduncan:
This will remove the `output-redactor` experiment flag and enable it all the time.

I have read through the redactor source code and added a couple more comments to help me understand it. It appears sound to me. I identified one issue which I added a test for to verify where if two needles share a prefix the shortest one will be redacted while the tail of the longer will be printed. This seems unlikely to occur in practice.

I have also ensured support for an empty list of redacted vars throughout so that (once the flag is changed) a lack of vars to redact will act as the switch to disable the `Redactor` being added to the log stream processing. This gives us an escape hatch should there be any issues.

Updated docs are in https://github.com/buildkite/docs/pull/1073

### DIFF
--- a/bootstrap/redactor_test.go
+++ b/bootstrap/redactor_test.go
@@ -65,3 +65,24 @@ func TestRedactorResetMidStream(t *testing.T) {
 		t.Errorf("Redaction failed: %s", buf.String())
 	}
 }
+
+func TestRedactorSlowLoris(t *testing.T) {
+	var buf bytes.Buffer
+	redactor := NewRedactor(&buf, "[REDACTED]", []string{"secret1111"})
+
+	redactor.Write([]byte("s"))
+	redactor.Write([]byte("e"))
+	redactor.Write([]byte("c"))
+	redactor.Write([]byte("r"))
+	redactor.Write([]byte("e"))
+	redactor.Write([]byte("t"))
+	redactor.Write([]byte("1"))
+	redactor.Write([]byte("1"))
+	redactor.Write([]byte("1"))
+	redactor.Write([]byte("1"))
+	redactor.Flush()
+
+	if buf.String() != "[REDACTED]" {
+		t.Errorf("Redaction failed: %s", buf.String())
+	}
+}
--- a/bootstrap/redactor_test.go
+++ b/bootstrap/redactor_test.go
@@ -86,3 +86,29 @@ func TestRedactorSlowLoris(t *testing.T) {
 		t.Errorf("Redaction failed: %s", buf.String())
 	}
 }
+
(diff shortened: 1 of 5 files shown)


keithduncan:
The remaining change to make here is in bootstrap.go `setupRedactors` to change how we enable / disable the redacting to be based on whether there are any variables to redact. I’m hoping to pair with @eleanorakh on that tomorrow.

cliconfig/loader.go
```diff
@@ -392,18 +392,22 @@ func (l Loader) normalizeField(fieldName string, normalization string) error {
...
 
-		// Normalize the field to be a command
+		// Normalize the field to be a string
 		if valueAsSlice, ok := value.([]string); ok {
 			normalizedSlice := []string{}
 
 			for _, value := range valueAsSlice {
 				// Split values with commas into fields
 				for _, normalized := range strings.Split(value, ",") {
+					if normalized == "" {
```
keithduncan, line 407: This change has broader implications beyond the redaction change as it affects the other list type config parameters. 

Those fields are:
- Tags
- TagsFromEC2MetaDataPaths
- TagsFromGCPMetaDataPaths
- Experiments

None of those convey semantic meaning to an empty string so I don’t have concerns that this will result in a change in practice.

bootstrap/redactor_test.go
```diff
@@ -65,3 +65,50 @@ func TestRedactorResetMidStream(t *testing.T) {
 		t.Errorf("Redaction failed: %s", buf.String())
 	}
 }
+
+func TestRedactorSlowLoris(t *testing.T) {
```
yob, line 69: 👏

keithduncan:
I’ve added the last piece to this which is enabling the redactor code path based on whether there are any vars to redact. I’ve verified that it isn’t enabled when `--redacted-vars=` is provided on the command line :tada:

yob (commented):
So great to see an experimental feature graduate to general availability 😍

I gave this a go, and confirmed the redaction was enabled without any experiment flags:

Screenshot from 2021-06-19 10-15-19

Do we definitely want to output the comment into every build log? I guess it might be nice to offer the visible reassurance that redaction is enabled?

keithduncan:
> Do we definitely want to output the comment into every build log? I guess it might be nice to offer the visible reassurance that redaction is enabled?

Good point, I don’t think we should print our own stuff during the command step. I’ve conditionalised this logging on whether `--debug` is being used, I also had it print the env var name patterns it is matching:

Screen Shot 2021-06-21 at 11 13 23

yob (approved):

PR #1453, opened 2021-06-22
WIP Experimental support for sending trace data to honeycomb

yob:
Support for sending tracing data to DataDog was added in PR #1273, using the OpenTracing API through the codebase, and DataDog's OpenTracing compatible library.

In theory we can add support for additional tracing vendors with only minimal changes. However, that requires the vendor to provide an OpenTracing compatible library. It sounds like OpenTracing is going out of fashion, and many/most vendors are adopting OpenTelemetry instead. This commit explores wrapping OpenTelemetry libraries in an OpenTracing shim so we can expand our vendor support without drastic changes across the agent codebase.

If OpenTracing continues to drop in popularity and OpenTelemetry is where vendors go, we could probably remove the shim and change our tracing code throughout the agent codebase to use  OpenTelemetry natively.

I hacked and slashed my way to something that sends data to honeycomb, but there's lots of work to do before this is mergable. In particular, the changes to tracetools/propagate.go aren't ideal and I think they might've broken linking of job traces into a single trace per build. They were required because the [otel Bridge implementation doesn't support the opentracing.TextMap type](https://github.com/open-telemetry/opentelemetry-go/blob/v0.20.0/bridge/opentracing/bridge.go).

[The honeycomb+OpenTelemetry+golang docs are here](https://docs.honeycomb.io/getting-data-in/opentelemetry/). I couldn't find any nice docs for the OpenTelemetry+OpenTracing bridge, so that was reverse engineered by reading source+docs.


bootstrap/bootstrap.go
```diff
@@ -281,6 +289,52 @@ func (b *Bootstrap) startTracing(ctx context.Context) (opentracing.Span, context
...
+				otlpgrpc.WithEndpoint("api.honeycomb.io:443"),
+				otlpgrpc.WithHeaders(map[string]string{
+					"x-honeycomb-team":    apikey,
+					"x-honeycomb-dataset": dataset,
+				}),
+			),
+		)
+		if err != nil {
+			fmt.Printf("ERROR: %v\n", err)
+			os.Exit(1)
```
yob, line 315: lol

tracetools/propagate.go
```diff
@@ -15,15 +15,16 @@ const EnvVarTraceContextKey = "BUILDKITE_TRACE_CONTEXT"
 
 // EncodeTraceContext will serialize and encode tracing data into a string and place
 // it into the given env vars map.
```
yob, line 17: I'd really prefer to not be changing this file at all. Unfortunately the otel provided OpenTracing shim doesn't support the `opentracing.TextMap` format

yob:
In it's current state you can start an agent like this:

```
$ HONEYCOMB_API_KEY=foo HONEYCOMB_DATASET=whatever ./buildkite-agent start --tracing-backend=honeycomb
```

.. and get job traces that look like this:

Screenshot from 2021-06-23 08-50-51

keithduncan:
This looks rad!

yob:
We're not actively working on this atm so I'll close  the PR but keep the branch around so we can mine it for ideas in the future

yob:
@keithduncan this branch might be useful! :heart: 

I'm about to reference it in a discussion about Open Tel.

PR #1454, opened 2021-06-24
Update architecture and OS support policy

keithduncan:
This updates and documents our OS and architecture promises to be current in terms of end-of-lifed operating systems. It also expands on the Docker section under installing to discuss the tagging system and lists the Docker base operating systems we build on top of.


yob (approved):
Love it :tada:

PR #1456, opened 2021-06-28
Add pre-bootstrap hook which can green light the Job env before it is loaded into a process env

keithduncan:
This provides a place to examine the proposed job environment before using as a process environment.

Notes on the todos I wrote up for this below.

### DIFF
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -617,9 +617,6 @@ func (r *JobRunner)executePreBootstrapHook(hook string) (bool, error) {
 		return false, err
 	}
 
-	// TODO stream logging up to buildkite?
-	// TODO if capturing and printing output of this script do we expect the redactors to run?
-
 	// TODO do we expect this script to be executed by bash, or sourced by bash
 	// - agent-shutdown is executed
 	// - all other current hooks are sourced by the hook/scriptwrapper.go
@@ -632,7 +629,7 @@ func (r *JobRunner)executePreBootstrapHook(hook string) (bool, error) {
 	sh.Env.Set("BUILDKITE_ENV_FILE", r.envFile.Name())
 
 	if output, err := sh.RunAndCapture(hook); err != nil {
-		r.logStreamer.Process(output)
+		// TODO print to agent output
 		return false, err
 	}
 
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -616,25 +616,37 @@ func truncateEnv(l logger.Logger, env map[string]string, key string, max int) er
 	return nil
 }
 
+type LogWriter struct {
+	l logger.Logger
+}
+
+func (w LogWriter) Write(bytes []byte) (int, error) {
+	w.l.Info("%s", bytes)
+	return len(bytes), nil
+}
+
 func (r *JobRunner)executePreBootstrapHook(hook string) (bool, error) {
+	r.logger.Info("Running pre-bootstrap hook %q", hook)
+
 	sh, err := shell.New()
 	if err != nil {
 		return false, err
 	}
 
-	sh.Debug = true
-
 	// This (plus inherited) is the only ENV that should be exposed
 	// to the pre-bootstrap hook.
 	sh.Env.Set("BUILDKITE_ENV_FILE", r.envFile.Name())
 
-	output, err := sh.RunAndCapture(hook)
-	if err != nil {
-		r.logger.Error("pre-bootstrap hook %q rejected job: %s", hook, output)
+	sh.Writer = LogWriter {
+		l: r.logger,
+	}
+
+	if err := sh.RunWithoutPrompt(hook); err != nil {
+		r.logger.Error("Finished pre-bootstrap hook %q: job rejected", hook)
 		return false, err
 	}
 
-	r.logger.Info("pre-bootstrap hook %q accepted job: %s", hook, output)
+	r.logger.Info("Finished pre-bootstrap hook %q: job accepted", hook)
 	return true, nil
 }
 
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -624,6 +624,11 @@ func (r *JobRunner)executePreBootstrapHook(hook string) (bool, error) {
 	// TODO pass line logging up to buildkite?
 	// TODO if capturing and printing output of this script do we expect the redactors to run
 
+	// TODO do we expect this script to be executed by bash, or sourced by bash
+	// - agent-shutdown is executed
+	// - all other current hooks are sourced by the hook/scriptwrapper.go
+	// - proposed hearbeat hook expects an executable https://github.com/buildkite/agent/pull/1057
+
 	sh.Promptf("%s", hook)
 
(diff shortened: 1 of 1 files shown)


agent/job_runner.go
```diff
@@ -283,30 +285,55 @@ func (r *JobRunner) Run() error {
...
-	// Run the process. This will block until it finishes.
-	if err := r.process.Run(); err != nil {
-		// Send the error as output
-		r.logStreamer.Process(fmt.Sprintf("%s", err))
+	// Before executing the bootstrap process with the received Job env,
+	// execute the pre-bootstrap hook (if present) for it to tell us
+	// whether it is happy to proceed.
+	environmentCommandOkay := true
+
+	// FIXME(KD) is looking on disk suitable for this use case, do we want an
```
keithduncan, line 293: Discussing this with @pda I think this is fine without a flag and up to the environment surrounding the agent to provide security.

The directory for the hooks should owned by another unix user that does not provide w or chmod permissions to the user running the buildkite-agent.

agent/job_runner.go
```diff
@@ -587,6 +621,34 @@ func truncateEnv(l logger.Logger, env map[string]string, key string, max int) er
 	return nil
 }
 
+func (r *JobRunner)executePreBootstrapHook(hook string) (bool, error) {
+	sh, err := shell.New()
+	if err != nil {
+		return false, err
+	}
+
+	// TODO stream logging up to buildkite?
```
keithduncan, line 630: I think we should not stream the logs up to buildkite.com, and possibly not even send up the process output at all. Instead just sending a static allowed/blocked message based on the exit code.

agent/job_runner.go
```diff
@@ -587,6 +621,34 @@ func truncateEnv(l logger.Logger, env map[string]string, key string, max int) er
...
+	// TODO if capturing and printing output of this script do we expect the redactors to run?
+
+	// TODO do we expect this script to be executed by bash, or sourced by bash
```
keithduncan, line 633: I think we should introduce a new terminology to describe these new 'hooks' that aren’t part of the bootstrap process and are exec'd instead of sourced.

Lifecycle Hooks?

keithduncan: Another benefit of these 'Agent Lifecycle Hooks' being exec'd and not sourced by Bash is that they can be written in a higher level language. I think it’s reasonable to expect this one in particular to be written in something like golang, or Rust and placed in the right spot for the buildkite-agent to fork/exec.

agent/job_runner.go
keithpitt: This is an interesting idea. Is this to perhaps guard against someone removing the pre-bootstrap and thus remove all the security checks?

agent/job_runner.go
keithpitt: What's the thinking here, security measure?

keithpitt: Thinking about this more, I think I'd be OK with not sending the output to BK, but instead just sending it to the agents STDOUT. That way if we introduce other "agent lifecycle hooks" like this "startup", "shutdown", etc, there can be a consistent message "Bootstrap hooks are sent to BK, agent hooks are sent to STDOUT" or something.

agent/job_runner.go
keithpitt: I like the idea of exec-ing them over sourcing them for the reasons you describe! Also, as per my comment about STDOUT and sending the log back to BK, I think it's OK for agent lifecycle hooks to have different behaviours. I tried to think of another word than "hook" though... We already have plugins... 🤔

agent/job_runner.go
keithduncan: That was my original idea, though as @pda pointed out if they can remove the file they can also replace it with a symlink to `/bin/true` so I don’t think this would actually provide any security benefit :+1:

keithpitt: Yeah, I was going to say the same thing :P

agent/job_runner.go
```diff
@@ -498,7 +525,14 @@ func (r *JobRunner) createEnvironment() ([]string, error) {
...
 	// currently running file (there is only 1 binary)
-	dir, _ := filepath.Abs(filepath.Dir(os.Args[0]))
+	exePath, err := os.Executable()
+	if err != nil {
+		return nil, err
+	}
+	dir, err := filepath.Abs(filepath.Dir(exePath))
+	if err != nil {
+		return nil, err
+	}
```
yob, line 535: nice find. I assume this is a clean up similar to #1444? Is it required for this PR, or just a thing you noticed while changing this part of the code?

clicommand/agent_start.go
```diff
@@ -161,10 +161,13 @@ func DefaultConfigFilePaths() (paths []string) {
 
 	// Also check to see if there's a buildkite-agent.cfg in the folder
 	// that the binary is running in.
-	pathToBinary, err := filepath.Abs(filepath.Dir(os.Args[0]))
+	exePath, err := os.Executable()
```
yob, line 164: :+1:

agent/job_runner.go
keithduncan: Just a thing I noticed while here and I debated about splitting it out into is own change set.

This one in particular could introduce a new error case. Previously `BUILDKITE_BIN_PATH` may have been set to an empty string but the function succeeded, now it would error. What do you think?

yob: I don't have a strong view about splitting it out or not, but I lean towards splitting it only because I assume the review intensity and feedback cycle count will be very different for adding the new hook vs. improving the binary path generation. I assume the latter will land very easily and Just Be Done.

> This one in particular could introduce a new error case. Previously BUILDKITE_BIN_PATH may have been set to an empty string but the function succeeded, now it would error. What do you think?

I assume the empty string case would be rare? If so, this seems like a behaviour change that we can accept?

keithduncan:
How this looks in the agent logs:

Screenshot 2021-07-02 at 13 10 04

How this looks in the build logs:

Screenshot 2021-07-02 at 13 10 12

agent/job_runner.go
```diff
@@ -283,30 +285,56 @@ func (r *JobRunner) Run() error {
...
-		r.logStreamer.Process(r.output.String())
+	if hook, _ := hook.Find(r.conf.AgentConfiguration.HooksPath, "pre-bootstrap"); hook != "" {
+		// Once we have a hook any failure to run it MUST be fatal to the job to guarantee a true
+		// positive result from the hook
+		okay, err := r.executePreBootstrapHook(hook)
+		if !okay {
+			environmentCommandOkay = false
+
+			// Ensure the Job UI knows why this job resulted in failure
+			r.logStreamer.Process("pre-bootstrap hook rejected this job, see the buildkite-agent logs for more details")
```
keithduncan, line 301: Made a tweak to this error message since the screenshot to point folks towards the agent jobs

keithduncan:
This is as-is ready for review, support for the specific signal_reason needs to land in buildkite/buildkite before it can be adopted.

PR #1457, opened 2021-06-29
Fix two more instances of os.Args[0]

keithduncan:
Swap two more instances of `os.Args[0]` with `os.Executable()` for a more reliable way to locate the binary.

Follow up to https://github.com/buildkite/agent/pull/1444

### DIFF
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -498,7 +498,14 @@ func (r *JobRunner) createEnvironment() ([]string, error) {
 
 	// We know the BUILDKITE_BIN_PATH dir, because it's the path to the
 	// currently running file (there is only 1 binary)
-	dir, _ := filepath.Abs(filepath.Dir(os.Args[0]))
+	exePath, err := os.Executable()
+	if err != nil {
+		return nil, err
+	}
+	dir, err := filepath.Abs(filepath.Dir(exePath))
+	if err != nil {
+		return nil, err
+	}
 	env["BUILDKITE_BIN_PATH"] = dir
 
 	// Add options from the agent configuration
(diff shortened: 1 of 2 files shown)


agent/job_runner.go
```diff
@@ -498,7 +498,14 @@ func (r *JobRunner) createEnvironment() ([]string, error) {
...
 	// We know the BUILDKITE_BIN_PATH dir, because it's the path to the
 	// currently running file (there is only 1 binary)
-	dir, _ := filepath.Abs(filepath.Dir(os.Args[0]))
+	exePath, err := os.Executable()
+	if err != nil {
+		return nil, err
+	}
+	dir, err := filepath.Abs(filepath.Dir(exePath))
+	if err != nil {
+		return nil, err
```
keithduncan, line 507: Follow on from https://github.com/buildkite/agent/pull/1456#discussion_r660229792

> I assume the empty string case would be rare? If so, this seems like a behaviour change that we can accept?

I would assume so. I’ve been analysing this and my concern might actually be moot 😅

The two sources of error here are `os.Executable()` and `filepath.Abs()`.

The first one doesn’t matter, because we use `os.Executable()` to find the binary for the bootstrap script setting. If that can’t find the binary, we aren’t going to have a bootstrap script to execute _anyway_ and won’t be this far in to the program. (Ignoring that someone might set it to something else, not setting it and having it self discover the path is the normal path).

Dir can’t fail, [Abs](https://golang.org/src/path/filepath/path.go?s=7025:7062#L233) might fail if it can’t [Getwd](https://golang.org/src/os/getwd.go?s=468:504#L12) which itself seems unlikely.

So I reckon this is fine!

yob (approved):

PR #1465, opened 2021-07-12
Add BUILDKITE_BIN_PATH to end, not start, of PATH

### LINKED ISSUE #1464: BUILDKITE_BIN_PATH is added to the front of PATH which effects other tools
I'm managing some Arm Linux bots for Linaro, for the https://buildkite.com/llvm-project/libcxx-ci project and recently they started failing. This was because our bot images replace `/usr/local/bin/cc` and `/usr/local/bin/c++` with a copy of clang-12. Since a certain point, it started picking up /usr/bin/cc which is the system gcc.

Before:
```
[437/523] /usr/local/bin/c++  -DNDEBUG -D_FILE_OFFSET_BITS=64 -D_GNU_SOURCE...
```
After:
```
[412/521] /usr/bin/c++  -DNDEBUG -D_FILE_OFFSET_BITS=64 -D_GNU_SOURCE ...
```

One thing that happened around this time is we rebuilt our docker images and picked up v3.31.0 of the agent. Unfortunately I don't have a copy or a way to recreate the previous docker image and your apt repo appears to only have that one version. So this is based on reading the commits.

What I think has happened is that when https://github.com/buildkite/agent/commit/35d96bf062982777ad72cab6e7241a9e9362e420 improved the method of finding `BUILDKITE_BIN_PATH` it started returning an actual path.

We run the agent with:
`sudo -i -u <build-user> buildkite-agent start`

Since Argv[0] isn't always a path but here just a name, it was either failing to get anything or adding some bogus entry on the front of `PATH` which got ignored. Now the improved code finds `buildkite-agent` in `/usr/bin` and puts that first on the path.

Which means that cmake now finds `/usr/bin/cc` before `/usr/local/bin/cc`. This is supported by some manual logging in the current bots:
```
+ echo /usr/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin
--
  | /usr/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin
```
Note how `/usr/bin` is repeated, I assume this is from `BUILDKITE_BIN_PATH`.

I found this issue https://github.com/buildkite/agent/pull/267 which talks about the same thing with the old shell script version. The original conversion commit was 8af44d2e0cb13996509aa01e804a6e8bd0a4ab6d and it did not include that fix unfortunately.

https://github.com/buildkite/agent/blob/eade03a3ac0ddaa624ebb6787fcdb3fcac01a6fa/bootstrap/bootstrap.go#L539-L543

Any reason not to apply the same fix again?

DavidSpickett:
Otherwise that folder becomes the highest priority path
and hides others futher down.

For example buildkite-agent might be in /usr/bin but you
have /usr/local/bin/cc that you expect to find.
If we added to the front of PATH, we'd hide that cc.

Fixes #1464.

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -539,7 +539,8 @@ func (b *Bootstrap) setUp(ctx context.Context) error {
 	// Add the $BUILDKITE_BIN_PATH to the $PATH if we've been given one
 	if b.BinPath != "" {
 		path, _ := b.shell.Env.Get("PATH")
-		b.shell.Env.Set("PATH", fmt.Sprintf("%s%s%s", b.BinPath, string(os.PathListSeparator), path))
+		// BinPath goes last so we don't disturb other tools
+		b.shell.Env.Set("PATH", fmt.Sprintf("%s%s%s", path, string(os.PathListSeparator), b.BinPath))
 	}
 
 	// Set a BUILDKITE_BUILD_CHECKOUT_PATH unless one exists already. We do this here


keithduncan (approved):

keithduncan:
Thank you for fixing this! And sorry for the regression 🙈 we’ll get this released shortly

DavidSpickett:
Np, always good when new code works too well.

PR #1474, opened 2021-07-26
More architecture packaging

### LINKED ISSUE #1393: RPM package for PPC64 and PPC64LE
Hi Buildkite

I am running 2 agents on Linux PPC64 (Big Endian) and PPCLE (Little Endian) respectively by compiling / creating custom systemd service files (which are based on the x86_64). I would like to ask if Buildkite could also provide the RPM for these two which would help streamline my setup.

If your devs need access to a proper PowerPC box, the [KernelCafe](https://github.com/KernelCafe/welcome) does offer some free nodes.

I am happy to contribute code changes if it helps.

Many thanks in advance

keithduncan:
Update DEB and RPM packaging to include ppc64(le). These architectures are community supported but we already build them so including them in our packaging is an easy win.

Fixes #1393

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -289,6 +289,7 @@ steps:
     depends_on:
       - build-binary-linux-amd64
       - build-binary-linux-386
+      - build-binary-linux-arm64
       - set-metadata
     command: ".buildkite/steps/build-rpm-packages.sh"
     artifact_paths: "rpm/**/*"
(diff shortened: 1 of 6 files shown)


.buildkite/pipeline.yml
```diff
@@ -288,6 +289,9 @@ steps:
     depends_on:
       - build-binary-linux-amd64
       - build-binary-linux-386
+      - build-binary-linux-arm64
```
keithduncan, line 292: This was already being used but was missing the dep 😅

PR #1488, opened 2021-08-17
Change the environment diff algorithm to support removed keys

### LINKED ISSUE #1471: `unset VARIABLE` should be merged into the `buildkite-agent bootstrap` environment
Currently, only changed variables are propagated out of hooks. If a variable needs to be `unset` rather than set to an empty string this isn’t currently possible.

[Environment.Diff](https://github.com/buildkite/agent/blob/main/env/environment.go#L80-L90) should instead of returning another environment object, return three sets of added, changed, and removed keys. The diff is only consumed in the scriptwrapper whose Changes() function is only consumed by the bootstrap file.

I think this should be a nicely isolated change.

keithduncan:
A work in progress to support `unset` in hooks and propagate those to subsequent hooks and phases. The primary use case for this being AWS credentials stored in environment variables.

Fixes #1471

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -390,21 +390,21 @@ func (b *Bootstrap) executeHook(ctx context.Context, scope string, name string, 
 	}
 
 	// Finally, apply changes to the current shell and config
-	b.applyEnvironmentChanges(changes.Env, changes.Dir, redactors)
+	b.applyEnvironmentChanges(changes, redactors)
 	return nil
 }
 
-func (b *Bootstrap) applyEnvironmentChanges(environ *env.Environment, dir string, redactors RedactorMux) {
-	if dir != b.shell.Getwd() {
-		_ = b.shell.Chdir(dir)
+func (b *Bootstrap) applyEnvironmentChanges(changes hook.HookScriptChanges, redactors RedactorMux) {
+	if changes.Dir != b.shell.Getwd() {
+		_ = b.shell.Chdir(changes.Dir)
 	}
 
 	// Do we even have any environment variables to change?
-	if environ == nil || environ.Length() == 0 {
+	if changes.Diff.Empty() {
 		return
 	}
 
-	mergedEnv := b.shell.Env.Merge(environ)
+	mergedEnv := b.shell.Env.Merge(changes.Diff)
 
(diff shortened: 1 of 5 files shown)


bootstrap/bootstrap.go
```diff
@@ -390,29 +390,29 @@ func (b *Bootstrap) executeHook(ctx context.Context, scope string, name string,
...
+	mergedEnv := b.shell.Env.Apply(changes.Diff)
 
 	// reset output redactors based on new environment variable values
 	redactors.Flush()
 	redactors.Reset(getValuesToRedact(b.shell, b.Config.RedactedVars, mergedEnv.ToMap()))
 
 	// First, let see any of the environment variables are supposed
 	// to change the bootstrap configuration at run time.
-	bootstrapConfigEnvChanges := b.Config.ReadFromEnvironment(environ)
+	bootstrapConfigEnvChanges := b.Config.ReadFromEnvironment(mergedEnv)
```
keithduncan, line 415: We want to double check this part, previously we passed `environ` which was just the _diff_ not the merged environment. Will this print report the config variables as being changed now?

keithduncan:
Tested this with a local agent:

Screenshot 2021-09-21 at 13 05 27

pda:
I felt like I had already done this, or something like it. Turns out I was thinking of https://github.com/buildkite/agent/pull/1232 which allowed hooks to _introduce_ empty environment variables. A different facet of the same category of problems, I guess.

env/environment.go
```diff
@@ -12,6 +12,27 @@ type Environment struct {
 	env map[string]string
 }
 
+type Pair struct {
+	Old string
+	New string
+}
```
pda, line 18: Minor point, but I'd suggest a different name for this type.

In the spatial context of being near `type Diff` it makes sense, but in the `package env` context, I think I'd expect `env.Pair` to be `{Key string, Value string}` or something.

Maybe it should be `DiffPair` or `StringDelta` or `StringChange` or something?

pda (approved):
Nice

env/environment.go
keithduncan: Changed in https://github.com/buildkite/agent/pull/1488/commits/b4038de920e11cf17b1329b1b2bf6414a6bb22c6

PR #1501, opened 2021-09-03
Run linux tests on amd64 and arm64

### LINKED ISSUE #1451: Add Linux arm64 tests step
Currently we run our tests on Linux x86_64 and Windows x86_64. Our elastic ci stack repository also runs an integration test on Linux arm64 so it’s worth knowing in this repo if there are any breakages before progressing a release up to the elastic ci stack.

keithduncan:
Fixes #1451


yob:
😍😍😍

PR #1502, opened 2021-09-03
Use buildx to build multi arch images

keithduncan:
Depends on docker-compose v2 because they don’t build arm64 executables prior to that release. It’s in rc phase so hopefully releases soon.

TODO:

- [x] Parallelise the multi-platform `test_docker_image` step, once these per-build images are `--output="type=registry"` we can run the per-architecture tests on separate machines and aggregate the results using Buildkite. This should be much faster than waiting for the serial results of each platform arch/variant to download and run through the tests.

### DIFF
--- a/.buildkite/pipeline.test-image.yml
+++ b/.buildkite/pipeline.test-image.yml
@@ -1,0 +1,6 @@
+---
+steps:
+  - name: ":docker: $IMAGE $PLATFORM"
+    command: .buildkite/steps/test-image.sh $IMAGE $PLATFORM
+    agents:
+      queue: $QUEUE
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -311,7 +311,7 @@ steps:
       - build-binary-linux-armhf
       - build-binary-linux-arm64
       - set-metadata
-    command: ".buildkite/steps/build-docker-image.sh alpine linux/arm/v7,linux/arm64/v8,linux/amd64"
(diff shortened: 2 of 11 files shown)


keithduncan:
Depends on a stack with https://github.com/buildkite/elastic-ci-stack-for-aws/pull/903

yob:
> Depends on docker-compose v2 because they don’t build arm64 executables prior to that release

oh, this is good news! 

We (mostly me) committed some horrible crimes to get a working docker compose in the elastic stack arm images:

https://github.com/buildkite/elastic-ci-stack-for-aws/blob/4c26d00a7347e84b6e26d9da1c48588835b5c571/packer/linux/scripts/install-docker.sh#L40-L53

keithduncan:
> oh, this is good news!
> 
> We (mostly me) committed some horrible crimes to get a working docker compose in the elastic stack arm images:

🙈

I don’t know what the backwards compatibility story is like between docker-compose v1 and v2, we can review merging this and adopting it in these images once the RC goes final :smile:

keithduncan:
The `test_docker_image` passes on Alpine but fails on CentOS and Ubuntu {18.04,20.04}. I wonder if running an emulated architecture docker CLI in a container connecting out to the host Docker socket is a supported use case or not.

sophiabits:
Hey just wondering when this might get merged? Docker compose v2 has been GA for a couple months now 😁 

Happy to help with any issues that might be blocking this :)

PR #1504, opened 2021-09-07
ArtifactDownloader: retain destination trailing slash

jonathan-brand:
[Documentation](https://buildkite.com/docs/agent/v3/cli-artifact#downloading-artifacts-description) for `buildkite-agent artifact download` states:
> If the last path component of matches the first path component of your , the last component of is dropped from the final path. For example, a query of 'app/logs/*' with a destination of 'foo/app' will write any matched artifact files to 'foo/app/logs/', relative to the current working directory. 

> To avoid this behaviour, use a argument with a trailing slash. For example, a query of 'app/logs/*' and a destination of 'foo/app/' will write the matched artifact files to 'foo/app/app/logs/', relative to the current working directory.

`Download.getTargetPath` does behave in this way [here](https://github.com/buildkite/agent/blob/4d7e6819c11a1844283600b31fab1bca677d1073/agent/download.go#L66). However, any trailing slash provided by the user in the destination string is stripped by `filepath.Abs` in `ArtifactDownloader.Download` [here](https://github.com/buildkite/agent/blob/4d7e6819c11a1844283600b31fab1bca677d1073/agent/artifact_downloader.go#L57) prior to `Download.getTargetPath` being invoked.

I am proposing that we retain any tailing slash after calling `filepath.Abs` in `ArtifactDownloader.Download`.

### DIFF
--- a/agent/artifact_downloader.go
+++ b/agent/artifact_downloader.go
@@ -57,8 +57,8 @@ func getDownloadDestination(destination string) string {
 	// trailing forward slash. We will then retain any trailing forward slash
 	// as it may indicate path merging behavior for download.getTargetPath.
 	downloadDestination, _ := filepath.Abs(destination)
-	if strings.HasSuffix(destination, "/") && len(destination) > 1 {
-		downloadDestination += "/"
+	if strings.HasSuffix(destination, string(os.PathSeparator)) && len(destination) > 1 {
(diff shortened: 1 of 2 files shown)


keithduncan (approved):
Thank you for opening this pull request @jonathan-brand! This will be incorporated into the next agent release.

PR #1530, opened 2021-10-29
Combine --spawn and --priority flags

sema:
When running multiple agents on the same host, it is recommended to use the --priority flag if you want to force load to be balanced evenly across the hosts (https://buildkite.com/docs/agent/v3/prioritization).

The --spawn flag allows for easily registering multiple agents on the same host, greatly simplifying the operational overhead of running multiple agents.

However, --spawn and --priority can't be mixed.

This pull request proposes a general approach to combine the two flags to allow for the ease-of-use of --spawn and also solve the load balancing problem for which --priority is being suggested for. The solution is to automatically assign a --priority to each spawned agent based on the internal agent index.

I'm happy to contribute further to this (tests, etc.) as soon as there is a general agreement on the direction.

### DIFF
--- a/.buildkite/steps/build-docker-image.sh
+++ b/.buildkite/steps/build-docker-image.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 set -euo pipefail
-
+set -x
 ## This script can be run locally like this:
 ##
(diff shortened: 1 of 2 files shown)


eleanorakh:
This is great, thanks so much @sema! I've just popped a commit on to use `with` instead of `splay` in the naming. The implementation looks 👍🏼

PR #1534, opened 2021-11-16
bash shouldn't be assumed to be in /bin for portability

jgedarovich:


pda:
Howdy @jgedarovich 👋🏼
Sorry for taking a long time to reply.

Could you describe the problem you're encountering?

The bash scripts this PR changes can be categorised into:

- `.buildkite/*` — CI build & release process.
- `*_test.go` — some integration tests that write and execute temporary shell scripts.
- `install.sh`— bash install script (designed to be piped into `bash -c`?)
- `packaging/docker/*/entrypoint.sh` — docker entrypoints for specific base images
- `packaging/linux/root/usr/share/buildkite-agent/hooks/*.sample` — sample hooks, could run on any system
- `scripts/*.sh` — I believe these are primarily used in the CI build & release process

Of those, I can imagine these might have portability issues:

- `*_test.go` — for running integration tests on machines without `/bin/bash`
- `…/hooks/*.sample` — when installed on machines without `/bin/bash`, and then customised by the user without fixing the shebang.

While `/bin/bash` is less flexible/portable, the tradeoff of `/usr/bin/env bash` is susceptibility to `$PATH` manipulation; where an attacker could trick the agent or its operator into running a malicious `bash` command elsewhere on the system. Admittedly this seems like a far-fetched scenario, and control over `$PATH` would be bad anyway.

I'd be interested to hear if you're running into one of the described scenarios, or something I didn't think of.

v1gnesh:
Hello @pda,

Here's a use case - non-standard platforms!


```
2025-02-09 04:13:27 DEBUG  Loaded config command=bootstrap agent_version=3.91.0+x..dirty
# Enabling output redaction for values from environment variables matching: [*_PASSWORD *_SECRET *_TOKEN *_PRIVATE_KEY *_ACCESS_KEY *_SECRET_KEY *_CONNECTION_STRING]
> Job API
> Buildkite environment variables
> Preparing working directory
Running commands
$ /bin/bash -e -c /data/v1gnesh/build-the-kite.sh
$ /bin/bash -e -c /data/v1gnesh/build-the-kite.sh
2025-02-09 04:13:28 DEBUG  [Process] Running without a PTY agent_version=3.91.0+x..dirty
# ↳ Command completed in 40.748ms
🚨 Error: error running "/bin/bash -e -c /data/v1gnesh/build-the-kite.sh": error starting command: fork/exec /bin/bash: EDC5129I No such file or directory.
# Local hook post-command doesn't exist: file does not exist, skipping
# Local hook pre-exit doesn't exist: file does not exist, skipping
# Successfully shut down Job API server
```

Please allow me to override or specify where to find bash shell.
The above log is from a platform that is entirely new - z/OS (a mainframe operating system).

I don't remember this failing when [this port](https://github.com/zopencommunity/buildkiteport) was initially created.
It has been a while since I tried to run BK in z/OS; so, trying again now, and I see the above.

pda:
Possibly also related / relevant:
https://github.com/buildkite/agent/pull/1995

v1gnesh:
Okay, I'm not sure how I'm supposed to use this... ?

v1gnesh:
@pda  ... ?

pda:
Ah I'm not sure #1995 is directly helpful, it just looked maybe-relevant.

I reckon if this PR was rebased onto latest `main` there's a strong chance it would be merged.

(I'm not working on the Buildkite Agent current, but as far as I know, nobody on that team is attached to hard-coding `/bin/bash`, and would be open to standardising on `/usr/bin/env bash`)

v1gnesh:
Tagging @jgedarovich in case this PR can be updated :)
Else will try to do this before month end.

jgedarovich:
Sorry I missed the convo, ill look into rebasing this week

jgedarovich:
@pda I've rebased against main, is there any magic words that need to be encanted to trigger the build? I see other PRs have had status reported automatically but it's been 4 hours since I rebased and still seems to say it's waiting.

DrJosh9000:
> magic words that need to be encanted to trigger the build? 

For the time being I've kicked off a build manually.

DrJosh9000 (approved):
LGTM

Thanks so much for this!

pda:
Thanks @DrJosh9000 🙏🏼

v1gnesh:
@pda or @DrJosh9000 - pls do a release when possible :)

PR #1535, opened 2021-11-22
Allow BUILDKITE_S3_DEFAULT_REGION to be used for unconditional bucket region

keithduncan:
If present this environment variable is used unconditionally and must be set to the region of the S3 bucket. Dynamic region discovery is bypassed completely owing to unreliability of the API.

This matches the secrets bucket algorithm in https://github.com/buildkite/elastic-ci-stack-s3-secrets-hooks/pull/54.

cc https://github.com/buildkite/elastic-ci-stack-for-aws/issues/937

### DIFF
--- a/agent/s3.go
+++ b/agent/s3.go
@@ -10,7 +10,6 @@ import (
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
 	"github.com/aws/aws-sdk-go/aws/defaults"
-	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/aws/aws-sdk-go/service/s3/s3manager"
@@ -52,30 +51,6 @@ func (e *credentialsProvider) IsExpired() bool {
 	return !e.retrieved
 }
 
-func awsS3RegionFromEnv() (region string, err error) {
-	regionName := os.Getenv("BUILDKITE_S3_DEFAULT_REGION")
-	if regionName == "" {
-		regionName, err = awsRegion()
-		if err != nil {
-			return "", err
-		}
-	}
-
-	// Check to make sure the region exists.
-	resolver := endpoints.DefaultResolver()
-	partitions := resolver.(endpoints.EnumPartitions).Partitions()
-
-	for _, p := range partitions {
-		for id := range p.Regions() {
-			if id == regionName {
-				return regionName, nil
-			}
-		}
-	}
-
-	return "", fmt.Errorf("Unknown AWS S3 Region %q", regionName)
-}
-
 func awsS3Session(region string) (*session.Session, error) {
 	// Chicken and egg... but this is kinda how they do it in the sdk
 	sess, err := session.NewSession()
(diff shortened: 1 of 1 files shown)


agent/s3.go
```diff
@@ -107,35 +82,51 @@ func webIdentityRoleProvider(sess *session.Session) *stscreds.WebIdentityRolePro
...
+
+		// Using the guess region, construct a session and ask that region where the
+		// bucket lives
+		session, err := awsS3Session(region)
+		if err != nil {
+			return nil, err
+		}
+		bucketRegion, err := s3manager.GetBucketRegion(aws.BackgroundContext(), sess, bucket, region)
+		if err != nil {
+			return nil, err
```
keithduncan, line 112: Make this assume current region as a worst case

agent/s3.go
```diff
@@ -107,35 +84,51 @@ func webIdentityRoleProvider(sess *session.Session) *stscreds.WebIdentityRolePro
...
+		// bucket lives
+		session, err := awsS3Session(region)
+		if err != nil {
+			return nil, fmt.Errorf("Could not load the AWS SDK config (%v)", err)
+		}
+
+		bucketRegion, bucketRegionErr := s3manager.GetBucketRegion(aws.BackgroundContext(), sess, bucket, region)
+		if bucketRegionErr == nil && bucketRegion != "" {
+			l.Debug("Discovered %q bucket region as %q\n", bucket, bucketRegion)
+			session.Config.Region = &bucketRegion
```
keithduncan, line 118: Needs testing to verify that this can be set after the session is constructed

keithduncan: Verified using a local agent that this does correctly re-configure the session and subsequent client to use the discovered region.

agent/s3.go
```diff
@@ -10,14 +10,15 @@ import (
...
 	"github.com/aws/aws-sdk-go/aws/defaults"
-	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/aws/aws-sdk-go/service/s3/s3manager"
 	"github.com/aws/aws-sdk-go/service/sts"
 	"github.com/buildkite/agent/v3/logger"
 )
 
+var regionHintEnvVar = "BUILDKITE_S3_DEFAULT_REGION"
```
matthewborden, line 20: Could this be a constant instead of a variable?

agent/s3.go
```diff
@@ -107,35 +84,51 @@ func webIdentityRoleProvider(sess *session.Session) *stscreds.WebIdentityRolePro
...
-	if err != nil {
-		return nil, err
-	}
-	bucketRegion, err := s3manager.GetBucketRegion(aws.BackgroundContext(), sess, bucket, region)
-	if err != nil {
-		return nil, err
-	}
+	regionHint := os.Getenv(regionHintEnvVar)
+	if regionHint != "" {
+		// If there is a region hint provided, we use it unconditionally
```
matthewborden, line 91:
```suggestion
        l.Debug("Using bucket region %q from environment variable %q", regionHint, regionHintEnvVar)
		// If there is a region hint provided, we use it unconditionally
```

matthewborden:
Might be worth mentioning in CHANGELOG.md about the subtle change of preferring the BUILDKITE_S3_DEFAULT_REGION  environment variable.

PR #1537, opened 2021-11-24
Fetch git tags by default

lox:
Historically this wasn't added due to --tags not being supported in earlier git versions (#338), but that was 2016. By now it's well supported and optimized so it's quite fast.

I'd advocate for this change being the least surprising, and adding it to defaults shouldn't break any existing configurations.

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -310,7 +310,7 @@ var AgentStartCommand = cli.Command{
 		},
 		cli.StringFlag{
 			Name:   "git-fetch-flags",
-			Value:  "-v --prune",
+			Value:  "-v --prune --tags",
 			Usage:  "Flags to pass to \"git fetch\" command",
 			EnvVar: "BUILDKITE_GIT_FETCH_FLAGS",
 		},


keithduncan:
It would be good to codify the minimum required git version for the agent.

Is there a way to feature detect whether the git CLI or remote remote supports this and degrade gracefully?

jradtilbrook:
I know some people operate by tagging every commit to the default branch (not sure the reasoning but I know it happens). So I think this could cause performance issues for that sort of thing. Would it be safer to default to off? Though if that's the case, there's almost no point in this change I suppose.  
Also, I wonder what the coverage of this change would be - are enough people building on or doing things with tags to make it useful or will it hurt more than it'll help?

lox:
I'd argue this is the sensible default and that folks with weird workflows should opt-out of it.

lox:
Version detection with git sounds like a good thing to have @keithduncan.

lox:
> So I think this could cause performance issues for that sort of thing. 

@jradtilbrook given the commit object is already being pulled down, is there reason to think adding a tag ref would incur substantial performance penalties? Looking at the git changelog, there has been lots of work done to make it quick.

jradtilbrook:
@lox I honestly don't know. I was just contributing a thought I had. It's quite possible that it won't - especially since the default is not to perform clean checkouts (unless its a new agent/instance) so pulling a few more commits to what is already there along with their tags is probably no big deal

lox:
@jradtilbrook yeah honestly I’m not sure either!

lox:
Without this behaviour in place, I've experienced users frequently running into weird race conditions where the latest reachable tags are based on which build on the host has fetched tags for that repo. It creates weirdly unreproducible bugs for anything that uses tags (for instance `goreleaser` or any manner of versioning tool), which is a bummer. The answer is always "Oh! Add a git fetch --tags". 

It's definitely possible this could break things, but would be awesome to consider for a v4.0.0 release as I think it's a good default.

lox:
After living with this change for a year or so, I will say it makes build output very verbose with lots of tags fetched each time 😬

PR #1543, opened 2021-12-07
Support multiple commands on Windows

### LINKED ISSUE #1538: Incorrect behaviour when running multiple batch/cmd files in a single command step
Agent version 3.33.3

We had a scenario where we needed to run two cmd files, `Setup.cmd` and `BuildProject.cmd`, in a single command step on Windows. The BK agent writes these into a single batch file as follows:
image
However, this generated batch file is incorrect. When calling batch/cmd files from other batch/cmd files, the `call` keyword should be used, otherwise multiple files are not called correctly and only the first file is called. In the above example `Setup.cmd` is run and then the command completes; `BuildProject.cmd` is never run. The correct batch file should be:
```
@echo off
call Setup.cmd
if %errorlevel% neq 0 exit /b %errorlevel%
call BuildProject.cmd
if %errorlevel% neq 0 exit /b %errorlevel%
```
The workaround here is to wrap the calls to multiple files in a single batch/cmd file and only run that single file in the command step.

(cc @PaulSonOfLars @elizabethking2)

keithduncan:
Insert a leading `'call '` for commands on Windows in the generated batch script per #1538 

I’ve also added a work in progress test for this functionality that I expect to fail pending finishing it off.

Fixes #1538

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -1747,6 +1747,36 @@ func isPosixShell(shell []string) bool {
 	}
 }
 
+/*
+	If line is another batch script, it should be prefixed with `call ` so that
+	the second batch script doesn’t early exit our calling script.
+
+	See https://www.robvanderwoude.com/call.php
+*/
+func shouldCallBatchLine(line string) bool {
+	// "  	gubiwargiub.bat /S  /e -e foo"
+	// "    "
+
+	/*
+		1. Trim leading whitespace characters
+		2. Split on whitespace into an array
+		3. Take the first element
+		4. If element ends in .bat or .cmd (case insensitive), the line should be prefixed, else not.
+	*/
+
+	trim := strings.TrimSpace(line) // string
+
+	elements := strings.Fields(trim) // []string
+
+	if len(elements) < 1 {
+		return false
+	}
+
+	first := strings.ToLower(elements[0]) // string
+
+	return (strings.HasSuffix(first, ".bat") || strings.HasSuffix(first, ".cmd"))
+}
+
 func (b *Bootstrap) writeBatchScript(cmd string) (string, error) {
 	scriptFile, err := shell.TempFileWithExtension(
 		`buildkite-script.bat`,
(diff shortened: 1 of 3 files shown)


bootstrap/bootstrap.go
```diff
@@ -1756,15 +1756,16 @@ func (b *Bootstrap) writeBatchScript(cmd string) (string, error) {
 	}
 	defer scriptFile.Close()
 
-	var scriptContents = "@echo off\n"
+	var scriptContents = []string{"@echo off"}
 
 	for _, line := range strings.Split(cmd, "\n") {
 		if line != "" {
-			scriptContents += line + "\n" + "if %errorlevel% neq 0 exit /b %errorlevel%\n"
+			scriptContents = append(scriptContents, "call " + line)
```
keithduncan, line 1763: @aknarayan I’m not very familiar with the Batch shell, are there occasions where this would not be correct? Should this be looking for lines that have a "command" like prelude e.g. "thing.ext"?

aknarayan: that's correct, but if we can safely assume that `line` will be the name of another script intended to be called in a command step then the use of `call` here should be fine. i don't know how the agent parses this, but for example, if a command step has an attribute that looks like this:
```
commands:
	- a.bat
	- b.bat
```
and when parsed, `line` will only ever take the values `a.bat` and `b.bat`, then the use of `call` here is correct.

aknarayan: it might also be better to leave the `errorlevel` check in there as well; batch sometimes behaves weirdly with error levels (more detail [here](https://waynes-world-it.blogspot.com/2008/08/difference-between-bat-and-cmd.html)) and i'm not too familiar with it either, but i don't think it can hurt to leave the `errorlevel` check in.

keithduncan: > if we can safely assume that `line` will be the name of another script intended to be called in a command step

That’s what I’m wondering, is there other Batch shell valid syntax that you could include in the "command" field of a step that would have worked before, like variable assignment perhaps?

aknarayan: yeah, potentially - in that case it would be safer to use the `call` keyword iff `line` is of the format `thing.ext`.

bootstrap/integration/bootstrap_tester.go
```diff
@@ -303,10 +303,13 @@ func (b *BootstrapTester) ReadEnvFromOutput(key string) (string, bool) {
 
 // Run the bootstrap and then check the mocks
 func (b *BootstrapTester) RunAndCheck(t *testing.T, env ...string) {
-	if err := b.Run(t, env...); err != nil {
-		t.Logf("Bootstrap output:\n%s", b.Output)
+	err := b.Run(t, env...)
+	t.Logf("Bootstrap output:\n%s", b.Output)
```
keithduncan, line 307: This doesn’t pollute the test logs because the `gosum` test runner only shows an individual test’s output if the test ultimately fails. Changing this means that _any_ test failure will include the bootstrap’s logs, not just an exec failure on the bootstrap.

PR #1556, opened 2022-01-18
Add issue templates

eleanorakh:
Add issue templates for bug reports and feature suggestions. Thanks here to the GitHub Wizard 😂

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,8 +1,8 @@
 blank_issues_enabled: true
- contact_links:
-   - name: Buildkite Support
-     url: https://buildkite.com/support
-     about: Ask questions not specifically related to the Buildkite Agent.
-   - name: Buildkite Community Slack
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,8 @@
+blank_issues_enabled: true
+ contact_links:
```
mmolina, line 2: I think there is an extra space before `contact_links:`. It looks like every line below is shifted one extra space right, and I fear that'll invalidate the YAML file

mmolina (commented):
I think the `config.yml` file is not valid YAML as it is because of an extra space at the beginning of a line. Unsure if there are any side effects 🤷‍♂️

.github/ISSUE_TEMPLATE/config.yml
eleanorakh: Good catch! Fixed.

mmolina (approved):
Looks good to me! 😃

PR #1562, opened 2022-02-09
Lock down file permissions on windows

tessereth:
Given the buildkite-agent runs arbitrary code as administrator, we should lock down access to the agent's files. I've opted to only change the acls if we're creating the installation directory for the first time and just warn if the directory already exists and is world accessible. If I was being fancy, I'd probably only warn for world write permissions (and allow read permissions) but I'm not an expert on windows acls or powershell so I'm not sure I can figure that out how to do that in a reasonable length of time.

### Testing

I have tested that a clean install with this script blocks read access from an unprivileged user but an administrator can still run the agent.

Installing with this script over the top of a previous install from `main` gives:

```
PS C:\Users\Administrator> iex ((New-Object System.Net.WebClient).DownloadString('https://raw.githubusercontent.com/buildkite/agent/4d566ab4184e23f5850de08d9ca75b9efdcd0342/install.ps1'))

 | '_ \| | | | | |/ _\ | |/ / | __/ _ \  / _\ |/ _\ |/ _ \ '_ \| __|
                                                __/ |
                                               |___/
Finding latest release
Downloading https://github.com/buildkite/agent/releases/download/v3.33.3/buildkite-agent-windows-amd64-3.33.3.zip
Preparing installation directory C:\buildkite-agent

WARNING: All users have the following access to the installation directory 'C:\buildkite-agent':
WARNING:   ReadAndExecute, Synchronize AppendData CreateFiles
WARNING: Consider only allowing administrators access to this directory.

Expanding buildkite-agent.zip
Expanding buildkite-agent.exe into bin
Updating PATH
buildkite-agent version 3.33.3, build 4013
Updating configuration in C:\buildkite-agent\buildkite-agent.cfg
Successfully installed to C:\buildkite-agent

You can now start the agent!

  C:\buildkite-agent\bin\buildkite-agent.exe start

For docs, help and support:

  https://buildkite.com/docs/agent/v3

Happy building! <3
```

(The extra comma in the list of permissions is because "ReadAndExecute, Synchronize" is a single row in the control list.)

Installing this version of the script twice does not show the warning.

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -49,6 +49,32 @@ foreach ($line in $resp.Content.Split("`n")) {
 Write-Host "Downloading $($releaseInfo.url)"
 Invoke-WebRequest -Uri $releaseInfo.url -OutFile 'buildkite-agent.zip'
 
+Write-Host "Preparing installation directory $installDir"
+if (Test-Path -Path $installDir) {
+    $permissions = (Get-Acl $installDir).Access |
+        where {$_.IdentityReference -match 'BUILTIN\\Users' -and `
+            $_.AccessControlType -eq [System.Security.AccessControl.AccessControlType]::Allow}
+    if ($permissions) {
+        Write-Host "
+WARNING: All users have the following access to the installation directory '$installDir':
+WARNING:   $($permissions.FileSystemRights)
+WARNING: Consider only allowing administrators access to this directory.
+        "
+    }
+} else {
+    New-Item -ItemType "directory" -Path $installDir | Out-Null
+    $acl = Get-Acl $installDir
+    # Disable ACL inheritance and remove existing inherited rules
+    $acl.SetAccessRuleProtection($true,$false)
+    # Allow System and Administrators full access
+    $acl.AddAccessRule($rule)
+    $rule = New-Object System.Security.AccessControl.FileSystemAccessRule("NT AUTHORITY\SYSTEM","FullControl","Allow")
+    $acl.AddAccessRule($rule)
+    $rule = New-Object System.Security.AccessControl.FileSystemAccessRule("BUILTIN\Administrators","FullControl","Allow")
+    $acl.AddAccessRule($rule)
+    $acl | Set-Acl $installDir
+}
+
 Write-Host 'Expanding buildkite-agent.zip'
 Expand-Archive -Force -Path buildkite-agent.zip -DestinationPath $installDir
 Remove-Item buildkite-agent.zip -Force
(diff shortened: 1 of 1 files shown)


sj26:
Yeah, very nice. 🙌

Could we add some output to the install script to say that it has locked down the installation path? Otherwise folks might miss that it happened and wonder why they can't do something, or why permissions are weird, etc.

I'd love @pda to have a look at this, too.

yob:
a very naive question from someone who's last real experience of windows was on XP: our linux packages install the buildkite-agent as a non-superuser (on Debian at least, we create a dedicated user called `buildkite-agent`) so by default other users of the system can't change agent files.

Is something similar possible on windows? Or does it Just Not Work Like That?

tessereth:
> Could we add some output to the install script to say that it has locked down the installation path?

Good call. It now says `Restricting installation directory access to administrators` if it goes down that branch.

@yob I wouldn't assume my windows is any better than yours. I'm not sure service accounts are really a thing? But I really don't know.

tessereth:
Some googling later, service accounts are totally a thing (see https://docs.microsoft.com/en-us/windows/security/identity-protection/access-control/service-accounts). But we aren't currently setting up the agent as a proper service in this script. We're just unzipping the package and telling people to run `buildkite-agent start`. I think if we wanted to make a separate user for this we'd have to figure out setting up a service and potentially how that interacts with active directory. That's probably all out of scope for the time being.

sj26:
Yeah I love that direction, but I reckon it's more than we should bite off here.

I'd love to add a windows installer or whatever is the right way to do that sort of thing which could create an appropriate service account and directories and all that jazz, if folks want that option. But until we know what folks would find palatable and useful I don't want to prescribe it. And CI tends to be a slightly atypical environment.

I feel like this is a good level of improvement for an individual giving the buildkite agent a go on their windows box. Beyond that I think it's up to delivery teams to craft a windows stack which scales and has a security posture etc which works for their particular needs.

install.ps1
```diff
@@ -49,6 +49,31 @@ foreach ($line in $resp.Content.Split("`n")) {
...
+            $_.AccessControlType -eq [System.Security.AccessControl.AccessControlType]::Allow}
+    if ($permissions) {
+        Write-Host "
+WARNING: All users have the following access to the installation directory '$installDir':
+WARNING:   $($permissions.FileSystemRights)
+WARNING: Consider only allowing administrators access to this directory.
+        "
+    }
+} else {
+    Write-Host "Restricting installation directory access to administrators"
```
pda, line 64: Is there enough context around this message output so that it's obvious what directory it's talking about?
Or should the message include `$installDir`?

Is it my imagination or does Windows generally capitalise **A**dministrators?

```suggestion
    Write-Host "Restricting installation directory access to Administrators: '$installDir'"
```

tessereth: Good point, I like it. Change applied.

pda (approved):

PR #1564, opened 2022-02-24
add option to enable job log file

albertywu:
Adds a new agent config option `--enable-job-log-file`.

When set, all logs for a job are written to a temporary file and exposed to the job via an environment variable `BUILDKITE_JOB_LOG`.

This log file can be used by post-command, pre-exit, and exit hooks to analyze the logs without having to fetch them via API (which is slower, and subject to race conditions). One use-case we have for this is log failure analysis of the command-step to enable dynamic retries.

This temporary file is deleted when the job finishes.

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -62,6 +62,7 @@ type AgentStartConfig struct {
 	DisconnectAfterIdleTimeout  int      `cli:"disconnect-after-idle-timeout"`
 	BootstrapScript             string   `cli:"bootstrap-script" normalize:"commandpath"`
 	CancelGracePeriod           int      `cli:"cancel-grace-period"`
+	EnableJobLogFile            bool     `cli:"enable-job-log-file"`
 	BuildPath                   string   `cli:"build-path" normalize:"filepath" validate:"required"`
 	HooksPath                   string   `cli:"hooks-path" normalize:"filepath"`
 	PluginsPath                 string   `cli:"plugins-path" normalize:"filepath"`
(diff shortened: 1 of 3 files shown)


clicommand/agent_start.go
```diff
@@ -62,6 +62,7 @@ type AgentStartConfig struct {
 	DisconnectAfterIdleTimeout  int      `cli:"disconnect-after-idle-timeout"`
 	BootstrapScript             string   `cli:"bootstrap-script" normalize:"commandpath"`
 	CancelGracePeriod           int      `cli:"cancel-grace-period"`
+	EnableJobLogFile            bool     `cli:"enable-job-log-file"`
```
moskyb, line 65: Could we change this (and all its dependent bits and bobs, the flag and envar etc) to something like
```suggestion
	JobLogLocalCache            bool     `cli:"job-log-local-cache"`
```
i'm very much open to suggestions on naming here, but i want to make it really clear to casual readers of the docs that the log file that this flag enables is ephemeral, and won't stick around after the job is complete - i'm keen to have a word like "cache" or "ephemeral" or "temporary" or something in there.

moskyb (changes requested):
thanks for your contribution @albertywu! i'm happy with the implementation here but i'm keen to iterate a little bit on the flag/envar names

clicommand/agent_start.go
pda: Yeah perhaps `--enable-job-log-tmpfile` or similar?

albertywu: Sounds good, renaming to `--enable-job-log-tmpfile` to make the naming more explicit

albertywu:
@moskyb @pda 

I've updated the PR to rename to use `EnableJobLogTmpfile`. Could you take another look?

(btw, not sure why I'm unable to resolve the request changes)

thanks!

moskyb:
@albertywu works on my machine ¯\\\_(ツ)_/¯
Screen Shot 2022-04-12 at 9 46 32 AM

moskyb (approved):
@albertywu LGTM 🎉

i'll roll this into the next minor version update we do

PR #1566, opened 2022-02-24
Use build matrix feature to simplify pipeline [PIP-200]

ticky:
This updates the pipeline configuration to use [the `matrix` feature](https://buildkite.com/blog/announcing-build-matrix), which reduces repetition significantly.

One negative effect is that this may make builds take slightly longer, as in-build dependencies are now on the entire build matrix and not individual steps.


tessereth (approved):
🎉

PR #1567, opened 2022-02-25
Ping immediately after completing a job

extemporalgenome:
This should minimize job "waiting for agent" time without
changing overall load on Buildkite servers.

Related: https://forum.buildkite.community/t/decrease-wait-time/1922

### DIFF
--- a/agent/agent_worker.go
+++ b/agent/agent_worker.go
@@ -205,6 +205,18 @@ func (a *AgentWorker) startPingLoop(idleMonitor *IdleMonitor) error {
 						return nil
 					}
 					lastActionTime = time.Now()
+
+					// Observation: jobs are rarely the last within a pipeline,
+					// thus if this worker just completed a job,
+					// there is likely another immediately available.
+					// Skip waiting for the ping interval until
+					// a ping without a job has occurred,
+					// but in exchange, ensure the next ping must wait a full
+					// pingInterval to avoid too much server load.
+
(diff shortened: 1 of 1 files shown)


agent/agent_worker.go
```diff
@@ -205,6 +205,18 @@ func (a *AgentWorker) startPingLoop(idleMonitor *IdleMonitor) error {
...
+					// thus if this worker just completed a job,
+					// there is likely another immediately available.
+					// Skip waiting for the ping interval until
+					// a ping without a job has occurred,
+					// but in exchange, ensure the next ping must wait a full
+					// pingInterval to avoid too much server load.
+
+					pingTicker.Reset(pingInterval)
+
+					continue
```
pda, line 219: So the purpose of this `continue` is to skip the part at the end of the loop that waits for the next tick:

```go
		select {
		case <-pingTicker.C:
			continue
		case <-a.stop:
			return nil
		}
```

And because we're effectively forcing a ping, `pingTicker.Reset(pingInterval)` sets the clock back to a full ping duration, so that the time since the last ping doesn't come out of the next iteration of the loop.

Makes sense 👍🏼

extemporalgenome: That's the intent!

pda (approved):

pda:
Thanks @extemporalgenome 👍🏼

PR #1568, opened 2022-02-27
Retry 500 responses when batch creating artifacts

### LINKED ISSUE #1565: Incorrect error message when artifact uploading returns 500
If an artifact upload fails with a 500 response code, an error message is printed showing it will retry 10 times when in fact the code is written to not retry for this error code. So it's a bit confusing to users.

<details>
<summary>Example error message in job logs:</summary>

Screen Shot 2022-02-24 at 14 18 59

</details>

Responsible code: https://github.com/buildkite/agent/blob/61a06503fe330d93f00415c99d841215d82f8b4f/agent/artifact_batch_creator.go#L75

I think this should print an accurate error message that it will not be retrying.

It could also retry as it states 🤷

moskyb:
Previously, when attempting to batch-create artifacts, if the agent API returned a 500 error, we would just sorta say 'okay' and never try again, emitting a somewhat confusing log message of:

```
POST https://agent.buildkite.com/v3/jobs/<some job id>/artifacts: 500 (Attempt 1/10 retrying in 10s)
``` 
... and then a failure message with no retries.

This PR makes it so that when we run across a 500 when trying to batch-create some artifacts, we retry the operation, as the logs imply we should. also, 500s can be (though are not always) transient, so retrying them is a good idea regardless.

Closes #1565
Closes PIP-219

### DIFF
--- a/agent/artifact_batch_creator.go
+++ b/agent/artifact_batch_creator.go
@@ -72,7 +72,7 @@ func (a *ArtifactBatchCreator) Create() ([]*api.Artifact, error) {
 		// Retry the batch upload a couple of times
 		err = retry.Do(func(s *retry.Stats) error {
 			creation, resp, err = a.apiClient.CreateArtifacts(a.conf.JobID, batch)
-			if resp != nil && (resp.StatusCode == 401 || resp.StatusCode == 404 || resp.StatusCode == 500) {
+			if resp != nil && (resp.StatusCode == 401 || resp.StatusCode == 404) {
 				s.Break()
 			}
 			if err != nil {


sj26 (approved):
It seems sensible to me to retry on 500. :+1:

I don't know why 500s were explicitly excluded from retry. It looks like that rule was added back when we added the new batch-based upload process in ee417c3754ddcd426ac9291e6f728aff4e55ca7d. Perhaps it was to guard against potential backend bugs, so we didn't create inordinate load after that change? That time of risk has passed, if so.

I wish we distinguished between transient and non-transient errors on the backend, but we don't, so this might lead to many retries of operations which will never succeed. But they are bugs we should fix.

I'll approve, but maybe wait and get @pda to pass eyes over this as well.

moskyb:
> Perhaps it was to guard against potential backend bugs, so we didn't create inordinate load after that change?

@sj26 yeah, i wondered about that too - i thought about changing the retry backoff from constant to exponential for this purpose, but that felt like scope creeping a bit too far. i might add it to our backlog somewhere though

pda (approved):

PR #1571, opened 2022-03-11
Directory artifact paths

### LINKED ISSUE #1570: Support directories in `artifact_paths`
Suppose I want to upload all contents of a directory `foo/bar`. I would love to specify artifact paths like this:

```yaml
artifact_paths:
  - foo/bar
```

I would expect this to be treated the same as `foo/bar/*`. But actually no artifacts are uploaded, even if the directory exists and is non-empty.  This is further complicated by the name of the variable/key.  The problem is that`artifact_paths` are not paths, they are _patterns_.  It is even more confusing that they do not support directories which _are_ paths!

Suppose in the further I have a command step that wants to create one directory from `$BUILDKITE_ARTIFACT_PATHS` if it doesn't exist. How would I go about it without first-class directory support?

Well, let's try the common shell script and assume we have the following glob pattern configured:

```yaml
artifact_paths:
  - foo/**/*
```

In our command:

```ShellSession
# Try to create directory foo
mkdir $BUILDKITE_ARTIFACTS_PATHS
```

... oops, now we created a directory called `foo/**/*`, what I really wanted was to create a directory called `foo` and put all my artifacts in that directory.

This is even _more_ confusing as a user because of the existence of `nullglob`, `dotglob` and the other variety of shell options.  Maybe we should all stop using shell scripts (I don't disagree). But I think for now, it would be better to support directories in `artifact_paths`.

If a path `foo` is specified in `artifact_paths` and it is a directory that exists, it should be treated like `foo/*` during the upload step.

jsoo1:
Closes #1570 

Previously, only glob patterns were allowed in `artifact_paths`. The name was confusing and also made certain patterns impossible.  In particular, in a common bash script, it is impossible to unambiguously "unglob" a string. For instance, in a shell script with `$BUILDKITE_ARTIFACT_PATHS=foo/**/*` and `$PWD` in an empty directory, the following would create a directory named literally `foo/**/*` (assuming no globbing shell options were set):

```ShellSession
mkdir $BUILDKITE_ARTIFACT_PATHS
```

Even if you were to try one of the globbing related shell options, there are none that would accomplish the desired effect.

Of course, the workaround would be to "just" `mkdir foo` in the shell script. However this breaks the decoupling between buildkite and the script as the buildkite configuration must "know" about the shell script and vice versa.

This change also makes new things possible.

One example is the following in a shell script `command` step:

```ShellSession
make prefix=$BUILDKITE_ARTIFACT_PATHS
```

Which is a common idiom in many build systems.

### DIFF
--- a/agent/artifact_uploader.go
+++ b/agent/artifact_uploader.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -117,7 +118,86 @@ func (a *ArtifactUploader) Collect(ctx context.Context) (artifacts []*api.Artifa
 	// file paths are deduplicated after resolving globs etc
 	seenPaths := make(map[string]bool)
 
-	for _, globPath := range strings.Split(a.conf.Paths, ArtifactPathDelimiter) {
+	prospects := strings.Split(a.conf.Paths, ArtifactPathDelimiter)
+
+	// Handle directories first
+	for _, literalPath := range prospects {
+		// I think this is a bug; '  name.ext  ' is a valid filepath.
+		// Keeping it because it was handled in the globbing case.
+		literalPath = strings.TrimSpace(literalPath)
+		if literalPath == "" {
+			continue
+		}
+
+		a.logger.Debug("Searching for %s", literalPath)
+
+		absolutePath, err := filepath.Abs(literalPath)
+		if err != nil {
+			return nil, err
+		}
+
+		if _, ok := seenPaths[absolutePath]; ok {
+			a.logger.Debug("Skipping duplicate path %s", literalPath)
+			continue
+		}
+
+		if isDir(absolutePath) {
+			a.logger.Debug("Searching for %s", literalPath)
+
+			err := filepath.WalkDir(absolutePath, func(p string, d fs.DirEntry, error error) error {
+				if error != nil {
+					return error
+				}
+
+				if _, ok := seenPaths[p]; ok {
+					a.logger.Debug("Skipping duplicate path %s", p);
+					return nil
+				}
+
+				if d.IsDir() {
+					return nil
+				}
+
+				path := p
+
+				if a.conf.GlobResolveFollowSymlinks {
+					resolvedLink, err := filepath.EvalSymlinks(path)
+					if err != nil {
+						return err
+					}
+					path = resolvedLink
+				}
+
+				seenPaths[p] = true
+
+				path, err := filepath.Rel(wd, path)
+				if err != nil {
+					return err
+				}
+
+				if experiments.IsEnabled(`normalised-upload-paths`) {
+					// Convert any Windows paths to Unix/URI form
+					path = filepath.ToSlash(path)
+				}
+
+				// Build an artifact object using the paths we have.
+				artifact, err := a.build(path, p, literalPath)
+				if err != nil {
+					return err
+				}
+
+				artifacts = append(artifacts, artifact)
+
+
+				return nil
+			})
+			if err != nil {
+				return nil, err
+			}
+		}
+	}
+
+	for _, globPath := range prospects {
 		globPath = strings.TrimSpace(globPath)
 		if globPath == "" {
 			continue
--- a/agent/artifact_uploader_test.go
+++ b/agent/artifact_uploader_test.go
@@ -80,12 +80,29 @@ func TestCollect(t *testing.T) {
 			Sha1Sum:      "bd4caf2e01e59777744ac1d52deafa01c2cb9bfd",
 			Sha256Sum:    "fc5e8608c7772e4ae834fbc47eec3d902099eb3599f5191e40d9e3d9b3764b0e",
 		},
+		{
+			Name:         "output.log",
+			Path:         []string{"test", "fixtures", "artifacts", "nested", "output.log"},
+			AbsolutePath: filepath.Join(root, "test", "fixtures", "artifacts", "nested", "output.log"),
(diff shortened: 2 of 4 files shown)


jsoo1:
After some more looking at this, I think it is slightly more complicated than this (but not much).

jsoo1:
Ok I think this is a start now. I see that maybe in the past `filepath.Walk` might have been restrictively slow but now `filepath.WalkDir` exists 😄 .

jsoo1:
A brief test on our CI confirms this does what I want. Though the symlinks need some work and there could be cases of which I am unaware.

jsoo1:
Ok tests pass and I think are actually correct.

jsoo1:
Hi @sj26 thank you for your feedback!

> `BUILDKITE_ARTIFACT_PATHS` is intended to be a list of globs for artifact upload, not used as an input to shell scripts for commands like `mkdir`.

Hm. Why is that? I would assume that almost any env var in a shell would be subject to some use by a command.

> If you want a shared value for commands and for artifacts then perhaps you want to abstract that. I would suggest doing something like this instead, for example:
> 
> 
> 
> ```yaml
> 
> steps:
> 
> - env:
> 
>     - PREFIX=the/place/to/build
> 
>   artifact_paths:
> 
>     - ${PREFIX}/bin/*
> 
>   command: make PREFIX="${PREFIX}"
> 
> ```

I can see this being useful, thank you!

I would like to push back on this decision a bit, though.

While I could probably define my own env var for the purpose, my bigger papercuts are listed in #1570.  To recap:

My problems are with the  correctness and violation of the principle of least surprise.

1. A directory _is_ a valid glob pattern!
2. Because of that my expectations are violated when a directory I list in `$BUILDKITE_ARTIFACT_PATHS` is explicitly ignored for upload.
3. `$BUILDKITE_ARTIFACT_PATHS` is a misnomer. I get confused when I include a directory in the list (which is a path, as the name suggests) and it gets ignored.

I only list the use of `$prefix` as a potential upside of supporting directory listings.

In any case, we love Buildkite and use it heavily.  Thank you for your work!

sj26:
The agent puts all the parameters required to run the job into the environment and then invokes a bootstrap process which consumes a lot of those variables to run hooks etc to run your command and all the surrounding behaviour. `$BUILDKITE_ARTIFACT_PATHS` is one such env, and is consumed by the `buildkite-agent artifact upload` command during arifact upload after the command hook.

You can upload the binaries produced by make with something like this:

```yaml
steps:
- env:
    - PREFIX=the/place/to/build
  artifact_paths: ${PREFIX}/bin/*
  command: make PREFIX="${PREFIX}"
```

or you could be explicit and handle the upload yourself if you want more control:

```yaml
steps:
- env:
    - PREFIX=the/place/to/build
  command: |
    make PREFIX="${PREFIX}"
    buildkite-agent artifact upload "${PREFIX}/bin/*"
```

I reckon we'll add a recursive mode for artifact upload at some point. Those commands need a bit of a rethink for better ergonomics. Maybe something like:


```yaml
steps:
- env:
    - PREFIX=the/place/to/build
  command: |
    make PREFIX="${PREFIX}"
    # Doesn't work right now, but might at some point in the future
    buildkite-agent artifact upload -r "${PREFIX}/bin"
```

jsoo1:
@sj26

> The agent puts all the parameters required to run the job into the environment and then invokes a bootstrap process which consumes a lot of those variables to run hooks etc around your command to provide the surrounding behaviour. `$BUILDKITE_ARTIFACT_PATHS` is one such env, and is consumed by the `buildkite-agent artifact upload` command during arifact upload after the command hook.

Ok... can you expand a bit on how that relates to this conversation? Is that an additional maintenance burden? Sorry I'm having trouble seeing the relation.

> You can upload the binaries produced by make with something like this:
> 
> 
> 
> ```yaml
> 
> steps:
> 
> - env:
> 
>     - PREFIX=the/place/to/build
> 
>   artifact_paths: ${PREFIX}/bin/*
> 
>   command: make PREFIX="${PREFIX}"
> 
> ```

... but importantly not:

> ```yaml
>   artifact_paths: ${PREFIX}/bin
> ``` 


> or you could be explicit and handle the upload yourself if you want more control:
> 
> 
> 
> ```yaml
> 
> steps:
> 
> - env:
> 
>     - PREFIX=the/place/to/build
> 
>   command: |
> 
>     make PREFIX="${PREFIX}"
> 
>     buildkite-agent artifact upload "${PREFIX}/bin/*"
> 
> ```

But here the glob will expand  by the shell, first, no? You would need to use `buildkite-agent artifact upload "${PREFIX}/bin/"'*'` to give the agent sub command an in expanded glob.

> I reckon we'll add a recursive mode for artifact upload at some point. Those commands need a bit of a rethink for better ergonomics.

That's what this pr does, no?

jsoo1:
Perhaps I've been a little too polite here. Not allowing directories in the `artifact_paths` list is a major usability and correctness bug. Let's use your example to motivate why. Suppose our default `make` target begins with a single binary output. We start with your pipeline definition:

```yaml
steps:
- env:
    - PREFIX=the/place/to/build
  artifact_paths:
    - ${PREFIX}/bin/*
  command: make PREFIX="${PREFIX}"
```

Then one day we decide to add a `lib` output. Now we would have to go change the pipeline to make sure the `lib` directory is uploaded:


```diff
  steps:
  - env:
      - PREFIX=the/place/to/build
    artifact_paths:
      - ${PREFIX}/bin/*
+     - ${PREFIX}/lib/*
    command: make PREFIX="${PREFIX}"
```

DrJosh9000:
Closing due to staleness, but also noting that simply `${PREFIX}/**` should work for obtaining files and directories within a particular prefix.

PR #1573, opened 2022-03-15
update github.com/aws/aws-sdk-go to the latest v1 release

yob:
I'm not aware of any bugfixes or changes that we need urgently, but it's approaching 2 years since we last upgraded (PR #1234) and we don't want an important dep like this to trail too far behind. Mainly because the longer the lag, the harder it might be to get back to current.

Once this merges I might investigate whether it's worth enabling dependabot for go modules so the bumps can happen semi-regularly.

Note that I've confirmed the agent still builds, but I have not confirmed artifact uploading to AWS bits still work: artifact uploading to custom buckets, artifact downloading, and fetching ec2 tags.

/cc @pda @moskyb

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.14
 require (
 	cloud.google.com/go v0.0.0-20170217213217-65216237311a
 	github.com/DataDog/datadog-go v3.7.2+incompatible
-	github.com/aws/aws-sdk-go v1.32.10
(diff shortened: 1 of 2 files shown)


pda (approved):
Looks good.
I'm pretty confident a minor version bump will be fine.
I think we can merge it, and do more testing prior to it going into a tagged release.

PR #1579, opened 2022-03-17
Fix no-plugins handling

elruwen:
Before this change, the isSet method returns false even though the fields are set in the config file. Now the config file is honored.

Now I can disable repository hooks but enable plugins.


sj26:
Thanks for the contribution @elruwen, this looks great! I'll let one of our resident Golang experts take a look and hopefully get this merged for you.

moskyb (approved):
Great catch @elruwen! Thanks for submitting this, and sorry it took so long for us to get to this. I'll merge this once the build runs and roll it into the next version :)

moskyb:
Looks like past ben has me covered :) I'll merge this now.

Screen Shot 2022-05-16 at 9 52 23 AM

PR #1580, opened 2022-03-17
Create directories only accessible by current user and group

pda:
Buildkite Agent creates directories on the filesystem for a few reasons:

- if started with a `--build-path` that doesn't exist,
- when downloading artifacts,
- when checking out the codebase to be built,
- when checking out plugins,
- when checking out git mirrors,
- for various temporary files.

This pull request changes the mode passed to `os.MkdirAll` from `0777` to `0770`.

Note that the specified mode is masked by [umask](https://en.wikipedia.org/wiki/Umask) before the directory is created.
The default umask of most systems is `0022`, or at the very least `0002`.

    $ echo "$(uname): $(umask)"
    Darwin: 022

    $ for img in archlinux debian ubuntu centos alpine; do echo -n "$img: "; docker run --rm $img sh -c umask; done
    archlinux: 0022
    debian: 0022
    ubuntu: 0022
    centos: 0022
    alpine: 0022

Given a standard umask of `0022`:

- The previous mode of `0777` created directories as: `0755` (`drwxr-xr-x`)
- The new mode of `0770` creates directories as:      `0750` (`drwxr-x---`)

In neither case is the directory world-writable.

This patch drops the ability for _other users_ to read the directory regardless of the umask in effect during creation.

It continues to delegate _group_ access control to umask, so that system administrators can use that standard mechanism to control how different users in the same group can share resources. The standard `0022` umask provides a safe default of disallowing write access from other users in the same group.

This change is not relevant to Windows, where new directories inherits the ACL from their parent.

### DIFF
--- a/agent/download.go
+++ b/agent/download.go
@@ -121,7 +121,7 @@ func (d Download) try() error {
 			responseDump, err := httputil.DumpResponse(response, true)
 			if err != nil {
 				d.logger.Debug("\nERR: %s\n%s", err, string(responseDump))
-			}  else {
+			} else {
 				d.logger.Debug("\n%s", string(responseDump))
 			}
 		}
@@ -130,7 +130,7 @@ func (d Download) try() error {
 	}
 
 	// Now make the folder for our file
-	err = os.MkdirAll(targetDirectory, 0777)
+	err = os.MkdirAll(targetDirectory, 0770)
 	if err != nil {
 		return fmt.Errorf("Failed to create folder for %s (%T: %v)", targetFile, err, err)
 	}
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -133,7 +133,7 @@ func NewJobRunner(l logger.Logger, scope *metrics.Scope, ag *api.AgentRegisterRe
 	// TempDir is not guaranteed to exist
 	tempDir := os.TempDir()
 	if _, err := os.Stat(tempDir); os.IsNotExist(err) {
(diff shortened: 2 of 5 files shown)


yob (approved):
I wasn't aware of the umask behaviour you've described. Surprisingly really, given how long I've been using linux.

Creating the directions so they're not readable by other users is a nice incremental improvement I think.

Is there any potential for this to be a breaking change for a handful customers? Maybe if they have a job that does something unusual like sudo to a different user and assume it can read the files in a build dir?

pda:
Yeah I imagine it could break a setup like that. But by leaving the group bits as `07` (`rwx`) before umask, this PR allows the user/sysadmin to use Unix groups and umask to fix that, and they'll end up with a better and more secure system 😁

kymikoloco:
In migrating from Jenkins to BuildKite (agent `3.34.0`), Docker was rebuilding images even though the files in the context were equivalent. Turned out that Docker's engine takes into account the permissions of a file in the context when deciding when to invalidate the cache, and Jenkins was checking out files with `0660`, and BuildKite did `0640`.

We (previously) used a `post-checkout` hook to add the permissions during the migration (as I also didn't know that `umask` was how you decided these things)

moskyb:
as @kymikoloco (among others 🙃) have pointed out, this is breaking docker workflows in the agent a bit. we're gonna revert it and make a patch release, and give it another go at some point.

PR #1582, opened 2022-03-17
Test `buildkite artifact shasum`, refactor for testability.

pda:
Basic test coverage for the `buildkite artifact shasum` command.

We don't have an established way to test our CLI invocations, and I think the way we wire up `github.com/urfave/cli` would need substantial refactoring to make it possible.

So instead, this pull request extracts a ` func searchAndPrintSha1Sum(…)` from the CLI action and tests that. The main API interaction, output and logging is covered by the test.

I'm interested in shifting from SHA-1 to SHA-256, so some coverage would be valuable.

### DIFF
--- a/agent/artifact_uploader.go
+++ b/agent/artifact_uploader.go
@@ -2,6 +2,7 @@ package agent
 
 import (
 	"crypto/sha1"
+	"crypto/sha256"
 	"errors"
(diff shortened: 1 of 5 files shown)


moskyb (approved):
Love it 🎉

PR #1583, opened 2022-03-17
Artifact SHA-256 checksum alongside existing SHA-1

pda:
Begin to shift from SHA-1 to SHA-256 for artifact checksums.

- `buildkite-agent artifact upload` will calculate and submit both SHA-1 and SHA-256.
- `buildkite-agent artifact shasum --sha256` will print the SHA-256 hash (or error if there isn't one).

_Currently_ there's no server-side support for `sha256sum` in the API, so the uploaded SHA-256 will not be stored, and the `--sha256` option on `buildkite-agent artifact shasum` will always fail. But that support would be added if/when this PR merges.

- [ ] teach `--format` of `artifact search` to show SHA-256 (non-blocking for this PR)

### DIFF
--- a/clicommand/artifact_shasum.go
+++ b/clicommand/artifact_shasum.go
@@ -18,10 +18,11 @@ var ShasumHelpDescription = `Usage:
 
 Description:
 
-   Prints the SHA-1 hash for the single artifact specified by a search query.
+   Prints the SHA-1 or SHA-256 hash for the single artifact specified by a
+   search query.
 
-   The SHA-1 hash is fetched from Buildkite's API, having been generated
-   client-side by the agent during artifact upload.
+   The hash is fetched from Buildkite's API, having been generated client-side
+   by the agent during artifact upload.
 
    A search query that does not match exactly one artifact results in an error.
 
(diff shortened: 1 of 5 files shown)


pda:
Ideally we'd slip this into the imminent https://github.com/buildkite/agent/pull/1591 v3.35.0 release /cc @moskyb 
AFAIK we haven't tagged that yet, so it's marginally awkward but not too late.

moskyb:
#1591 isn't merged yet! feel free to slip this into v3.35, i'll review this now

clicommand/artifact_shasum.go
```diff
@@ -68,6 +74,10 @@ var ArtifactShasumCommand = cli.Command{
 	Usage:       "Prints the SHA-1 hash for a single artifact specified by a search query",
 	Description: ShasumHelpDescription,
 	Flags: []cli.Flag{
+		cli.BoolFlag{
+			Name:  "sha256",
+			Usage: "Request SHA-256 instead of SHA-1, errors if SHA-256 not available",
+		},
```
moskyb, line 80: is it worth preferentially fetching SHA256, and then returning SHA1 only if it isn't present, rather than making users decide? would that break any workflows?

moskyb (approved):
LGTM, one question about the user interface, but i'm happy to table that for a later date

clicommand/artifact_shasum.go
pda: It might break things, if e.g. people were comparing it against their own SHA-1.

PR #1588, opened 2022-03-21
Refactor retry machinery to allow the use of exponential backoff

moskyb:
**🤔 Problem:** The retry util we currently use has a couple of issues:
- There aren't any tests
- Retry configuration errors happen when you try to use the retrier, rather than when you configure it
- It only supports constant backoff - exponential backoff is really handy a lot of the time.

**✅ Solution:** This PR replaces the existing machinery (encapsulated in the `retry.Do` function and the `retry.Config` struct), with a struct called `retry.Retrier`. `Retrier` intentionally has a similar user interface to `retry.Do`, but it adds:
- Configurable backoff strategies - currently constant and exponential, but we could theoretically do some freaky stuff with it if we really wanted to
- Constructor panics when coding errors are made - eg, if a user puts a negative interval in to a constant strategy, that can never be valid and the program should complain loudly if it happens
- Tests! Hooray!

**💬 Things to discuss:** 
- the math i use in the `retry.Exponential` function means that i can't have a base lower than 1 second, as i convert the `time.Duration` into an integer number of seconds, but if i don't do that, then the math gets all wack because `time.Duration`s are integer numbers of nanoseconds - eg, `math.Pow(2*time.Second, 2) != math.Pow(2, 2) * time.Second`. is there a happy middle ground here? i'm embarrassed to say that my arithmetic skills are failing me here.
- How do we feel about the functional constructor? i feel like i was overdoing it by adding it, but... i kinda like it
- ~~I've changed the artifact batch creation to use exponential backoff, but should we change other operations to use exponential? if so, which ones?~~ This is no longer true, i'll change things to use expo backoff in another PR.

**This PR will be most easily reviewed commit-by-commit**

Closes PIP-234


pda:
Ooh nice, I like the look of it.
(I haven't fully reviewed it yet.)

retry/retrier.go
```diff
@@ -0,0 +1,209 @@
...
+			return err
+		}
+
+		// Perform the action the user has requested we retry
+		err = callback(r)
+		if err == nil {
+			return nil
+		}
+
+		time.Sleep(r.NextInterval())
```
tessereth, line 207: This loop is not in the order I expected it to be. I expected:

1. Call the callback
2. Return if success
4. MarkAttempt (because in my head it's a count of failures so it happens after the call, but it could be earlier)
3. Check ShouldGiveUp
5. Call NextInterval and sleep

I think the main differences are:

1. If we're going to give up, we should do so immediately, not after sleeping.
2. If I set MaxAttempts to 1, it should make one request but the current code won't make any requests.

retry/retrier.go
```diff
@@ -0,0 +1,209 @@
+package retry
+
+import (
+	"fmt"
+	"math"
+	"math/rand"
+	"time"
+)
+
+var random = rand.New(rand.NewSource(time.Now().UnixNano()))
```
tessereth, line 10: You may want to make it possible to pass in a custom random number generator for testing purposes.

retry/retrier.go
```diff
@@ -0,0 +1,209 @@
...
+		panic("retriers using the constant strategy that run forever must have an interval")
+	}
+	r.jitter = oldJitter // and now set it back to what it was previously
+
+	return r
+}
+
+func (r *Retrier) calculateJitter() time.Duration {
+	if r.jitter {
+		return time.Duration(random.Float32()) * jitterInterval
```
tessereth, line 130: This can go down as well as up. Does everything correctly handle `NextInterval` returning a negative number?

tessereth (commented):
First of all, YAY exponential backoff and jitter. I'm all for this.

I've only really checked out the `retrier.go` code in detail. I like it. But there's an annoying number of pointy bits in this kind of code. Have you explored pulling in a library for this rather than us rolling our own?

retry/retrier.go
moskyb: yep, these exact points came up in the mob review session we did the other day. the main reason i did it like this was to maintain compatibility with the old retrier, which behaved in the same way as laid out in the code. there's one test that fails if we shuffle around the order, but IMO i'd rather change some tests such that our code is semantically correct than make code that does weird stuff so that i don't have to change tests

retry/retrier.go
moskyb: good call, will add this!

moskyb:
@tessereth 
> Have you explored pulling in a library for this rather than us rolling our own?

yeah, i looked into this a bit - there weren't any libraries that i really loved, and none as nice as eg. python's [tenacity](https://github.com/jd/tenacity) or ruby's [retryable](https://github.com/nfedyashev/retryable). i suspect that it's the kind of thing that gets reimplemented privately a lot, because the libraries out there aren't awesome - QED, that's exactly what i did 😅

retry/retrier.go
moskyb: `NextInterval` is only ever used as an arg to time.Sleep, which just [returns immediately](https://github.com/golang/go/blob/master/src/runtime/time.go#L180=) if it's passed a negative interval, so if things go negative we should be all good.

can you explain what you mean by going down as well as up though? `random.Float64` [returns 0 or a positive float](https://pkg.go.dev/math/rand#Float64), though i actually do think that it might be good to have negative jitter be a thing

tessereth: Oh, I just assumed `random.Float64` would return any random float64. That's what you get for not reading the docs. Thanks for letting me know!

retry/retrier.go
moskyb: okay, so i finally got around to fixing this today, but in the latest push, we get the best of both worlds - clear logic in the retry loop, and we didn't need to change any of the existing tests :)

retry/retrier.go
```diff
@@ -0,0 +1,238 @@
...
+	return func(r *Retrier) {
+		r.strategyType = strategyType
+		r.intervalCalculator = strategy
+	}
+}
+
+// WithJitter enables jitter on the retrier, which will cause all of the retries to wait a random amount of time < 1 second
+// The idea here is to avoid thundering herds - retries that are in parallel will happen at slightly different times when
+// jitter is enabled, whereas if jitter is disabled, all the retries might happen at the same time, causing further load
+// on the system that we're tryung to do something with
```
waferbaby, line 92: Typo: 'trying'

waferbaby (approved):
Looks great to me!

sj26 (approved):
Love it

PR #1523, opened 2021-10-05
Reject pipeline uploads containing redacted vars

keithduncan:
Work in progress with @eleanorakh and @pzeballos to catch and prevent pipeline uploads from including interpolated redacted vars.

Fixes #1469

PR #1589, opened 2022-03-22
Make secret value rejection on pipeline upload optional

### LINKED ISSUE #1578: Feature flag to reject pipelines with redacted variables
buildkite-agent 3.34.0 introduced new behavior, rejecting pipeline uploads that contain redacted variables. As of now, it's impossible to continue to redact variables and allow a pipeline with secrets to be uploaded. To me, these are two distinct features that should be controlled separately.

I realize that having a secret in a pipeline in plain text is clearly not a best practice, but sometimes it is difficult to track down, especially if you're working with dynamic pipelines.

A new configuration flag like `reject_pipelines_with_secrets = false` that will allow pipelines to be uploaded even if they contain secrets. It would probably make sense for some strong warnings still. Maybe even say that "uploading pipelines with redacted vars is deprecated and won't be supported in the next major version" kind of thing. Something that gives a transition period to identify the issues and remediate before rejecting them outright.

If there are many pipelines that happen to upload with redacted vars, we need to update every single one ASAP (may not be possible to do in a timely manner)j, rollback the buildkite-agent version, or disable redacting vars, which still has a lot of value in hiding values in logs.

**Additional context**
I opened a ticket with support@buildkite.com and they sent me your way. Thanks in advance!

moskyb:
**🤔 Problem:** In https://github.com/buildkite/agent/pull/1523, we made it so that the `buildkite-agent pipeline upload` command would reject any pipeline yaml that had sensitive values in it. We shipped this as part of [v3.34](https://github.com/buildkite/agent/releases/tag/v3.34.0). This is an awesome security feature, and prevents a bunch of footguns, but unfortunately it's a breaking change, and we accidentally shipped it as part of a minor release.

**✅ Solution:** This PR reverts default redaction behaviour to how it was previously, with some slight changes:
- The newer behaviour (rejecting a pipeline upload if there's juicy secrets in it) is available with the `--reject-secrets` flag
- Even when the `--reject-secrets` flag isn't used, the command issues a warning that:
  - There are unredacted secrets in the pipeline
  - and that default behaviour in agent v4 will be to reject these pipeline uploads

Slightly related, it makes the `--dry-run` flag for the pipeline upload command work with redaction.

**Screenshots:**
Without `--reject-secrets`:

Screen Shot 2022-03-22 at 2 59 50 PM

With `--reject-secrets`:

Screen Shot 2022-03-22 at 3 00 10 PM

These are from my local machine using the `--dry-run` flag, using this pipeline:
```yaml
steps:
  - label: ":hammer: Example Script"
    command: "echo $COOL_PASSWORD"
    artifact_paths: "artifacts/*"
    agents:
      queue: "${BUILDKITE_AGENT_META_DATA_QUEUE:-default}"
```

Closes #1578 
Closes PIP-309

### DIFF
--- a/clicommand/pipeline_upload.go
+++ b/clicommand/pipeline_upload.go
@@ -57,6 +57,7 @@ type PipelineUploadConfig struct {
 	DryRun          bool     `cli:"dry-run"`
 	NoInterpolation bool     `cli:"no-interpolation"`
 	RedactedVars    []string `cli:"redacted-vars" normalize:"list"`
+	RejectSecrets   bool     `cli:"reject-secrets"`
 
 	// Global flags
 	Debug       bool     `cli:"debug"`
@@ -73,7 +74,7 @@ type PipelineUploadConfig struct {
 
 var PipelineUploadCommand = cli.Command{
 	Name:        "upload",
-	Usage:       "Uploads a description of a build pipeline adds it to the currently running build after the current job.",
+	Usage:       "Uploads a description of a build pipeline adds it to the currently running build after the current job",
 	Description: PipelineUploadHelpDescription,
 	Flags: []cli.Flag{
 		cli.BoolFlag{
@@ -96,6 +97,11 @@ var PipelineUploadCommand = cli.Command{
 			Name:   "no-interpolation",
 			Usage:  "Skip variable interpolation the pipeline when uploaded",
 			EnvVar: "BUILDKITE_PIPELINE_NO_INTERPOLATION",
+		},
+		cli.BoolFlag{
+			Name:   "reject-secrets",
+			Usage:  "When true, fail the pipeline upload early if the the pipeline contains secrets",
+			EnvVar: "BUILDKITE_AGENT_PIPELINE_UPLOAD_REJECT_SECRETS",
 		},
 
 		// API Flags
@@ -227,6 +233,34 @@ var PipelineUploadCommand = cli.Command{
 			l.Fatal("Pipeline parsing of \"%s\" failed (%s)", src, err)
 		}
 
+		if len(cfg.RedactedVars) > 0 {
+			needles := redaction.GetKeyValuesToRedact(shell.StderrLogger, cfg.RedactedVars, env.FromSlice(os.Environ()).ToMap())
+			serialisedPipeline, err := result.MarshalJSON()
+
+			if err != nil {
+				l.Fatal("Couldn’t scan the %q pipeline for redacted variables. This parsed pipeline could not be serialized, ensure the pipeline YAML is valid, or ignore interpolated secrets for this upload by passing --redacted-vars=''. (%s)", src, err)
+			}
+
+			stringifiedserialisedPipeline := string(serialisedPipeline)
+
+			secretsFound := make([]string, 0, len(needles))
+			for needleKey, needle := range needles {
+				if strings.Contains(stringifiedserialisedPipeline, needle) {
+					secretsFound = append(secretsFound, needleKey)
+				}
+			}
+
+			if len(secretsFound) > 0 {
+				if cfg.RejectSecrets {
+					l.Fatal("Pipeline %q contains values interpolated from the following secret environment variables: %v, and cannot be uploaded to Buildkite", src, secretsFound)
+				} else {
+					l.Warn("Pipeline %q contains values interpolated from the following secret environment variables: %v, which could leak sensitive information into the Buildkite UI.", src, secretsFound)
+					l.Warn("This pipeline will still be uploaded, but if you'd like to to prevent this from happening, you can use the `--reject-secrets` cli flag, or the `BUILDKITE_AGENT_PIPELINE_UPLOAD_REJECT_SECRETS` environment variable, which will make the `buildkite-agent pipeline upload` command fail if it finds secrets in the pipeline.")
+					l.Warn("The behaviour in the above flags will become default in Buildkite Agent v4")
+				}
+			}
+		}
+
 		// In dry-run mode we just output the generated pipeline to stdout
 		if cfg.DryRun {
 			enc := json.NewEncoder(os.Stdout)
@@ -239,23 +273,6 @@ var PipelineUploadCommand = cli.Command{
 			}
 
 			return
-		}
-
-		if len(cfg.RedactedVars) > 0 {
-			needles := redaction.GetKeyValuesToRedact(shell.StderrLogger, cfg.RedactedVars, env.FromSlice(os.Environ()).ToMap())
-			serialisedPipeline, err := result.MarshalJSON()
-
-			if err != nil {
-				l.Fatal("Couldn’t scan the %q pipeline for redacted variables. This parsed pipeline could not be serialized, ensure the pipeline YAML is valid, or ignore interpolated secrets for this upload by passing --redacted-vars=''. (%s)", src, err)
-			}
-
-			stringifiedserialisedPipeline := string(serialisedPipeline)
-
-			for needleKey, needle := range needles {
-				if strings.Contains(stringifiedserialisedPipeline, needle) {
-					l.Fatal("Couldn’t upload the %q pipeline. Refusing to upload pipeline containing the value interpolated from the %q environment variable. Ensure your pipeline does not include secret values or interpolated secret values.", src, needleKey)
-				}
-			}
 		}
 
 		// Check we have a job id set if not in dry run
--- a/redaction/redactor.go
+++ b/redaction/redactor.go
@@ -321,7 +321,7 @@ func GetKeyValuesToRedact(logger shell.Logger, patterns []string, environment ma
 
 			if matched {
 				if len(varValue) < RedactLengthMin {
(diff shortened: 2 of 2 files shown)


moskyb:
i've also tested that this works in a non-dry run environment, using an agent on my machine:

With `--reject-secrets`:

Screen Shot 2022-03-23 at 10 20 35 AM


Without `--reject-secrets`:

Screen Shot 2022-03-23 at 10 19 51 AM

clicommand/pipeline_upload.go
```diff
@@ -227,35 +233,46 @@ var PipelineUploadCommand = cli.Command{
...
+					secretsFound = append(secretsFound, needleKey)
 				}
 			}
+
+			if len(secretsFound) > 0 {
+				if cfg.RejectSecrets {
+					l.Fatal("Refusing to upload pipeline %q, as it contained values interpolated from the following secret environment variables: %v. Ensure your pipeline does not include secret values or interpolated secret values.", src, secretsFound)
+				} else {
+					l.Warn("The pipeline %q contains values interpolated from the following secret environment variables: %v.", src, secretsFound)
+					l.Warn("I'm still going to upload the pipeline, but if you'd like to to prevent this from happening, you can use the `--reject-secrets` cli flag, or the `BUILDKITE_PIPELINE_REJECT_SECRETS` environment variable, which will make the `buildkite-agent pipeline upload` command fail if it finds secrets in the pipeline.")
```
moskyb, line 258: how do we feel about first-person error messages? i personally like them (QED, i wrote this), but if that's not the ✨buildkite agent vibe✨ then i'm happy to edit this

clicommand/pipeline_upload.go
```diff
@@ -97,6 +98,11 @@ var PipelineUploadCommand = cli.Command{
 			Usage:  "Skip variable interpolation the pipeline when uploaded",
 			EnvVar: "BUILDKITE_PIPELINE_NO_INTERPOLATION",
 		},
+		cli.BoolFlag{
+			Name:   "reject-secrets",
+			Usage:  "When true, fail the pipeline upload early if the the pipeline contains secrets.",
```
ticky, line 103: Style nit: generally our multi-sentence usages have full stops, single sentences don't.

```suggestion
			Usage:  "When true, fail the pipeline upload early if the the pipeline contains secrets",
```

clicommand/pipeline_upload.go
```diff
@@ -227,35 +233,46 @@ var PipelineUploadCommand = cli.Command{
...
-			return
-		}
-
 		if len(cfg.RedactedVars) > 0 {
 			needles := redaction.GetKeyValuesToRedact(shell.StderrLogger, cfg.RedactedVars, env.FromSlice(os.Environ()).ToMap())
 			serialisedPipeline, err := result.MarshalJSON()
 
 			if err != nil {
-				l.Fatal("Couldn’t scan the %q pipeline for redacted variables. This parsed pipeline could not be serialized, ensure the pipeline YAML is valid, or ignore interpolated secrets for this upload by passing --redacted-vars=''. (%s)", src, err)
+				l.Fatal("Couldn't scan the %q pipeline for redacted variables. This parsed pipeline could not be serialized, ensure the pipeline YAML is valid, or ignore interpolated secrets for this upload by passing --redacted-vars=''. (%s)", src, err)
```
ticky, line 241: Not a fan of the pretty apostrophes?

clicommand/pipeline_upload.go
ticky: Hmm, doesn't seem quite right to my eye. Maybe more a "This pipeline will still be uploaded"

ticky (approved):

clicommand/pipeline_upload.go
eleanorakh: Yea, I'd prefer we didn't use first person here :)

clicommand/pipeline_upload.go
```diff
@@ -97,6 +98,11 @@ var PipelineUploadCommand = cli.Command{
 			Usage:  "Skip variable interpolation the pipeline when uploaded",
 			EnvVar: "BUILDKITE_PIPELINE_NO_INTERPOLATION",
 		},
+		cli.BoolFlag{
+			Name:   "reject-secrets",
+			Usage:  "When true, fail the pipeline upload early if the the pipeline contains secrets.",
+			EnvVar: "BUILDKITE_PIPELINE_REJECT_SECRETS",
```
pda, line 104: I wonder if this should be `BUILDKITE_AGENT_PIPELINE_REJECT_SECRETS` or `BUILDKITE_PIPELINE_UPLOAD_REJECT_SECRETS` 🤔

i.e. putting it in the `BUILDKITE_AGENT_` namespace, to separate it from things set by the Buildkite server, or making it clear that it only affects pipeline _uploads_.

I think there's a mixture of precedent. e.g:

- `BUILDKITE_PIPELINE_REPLACE` matches your new var name.
- `BUILDKITE_PIPELINE_NO_INTERPOLATION` also.
- `BUILDKITE_PIPELINE_UPLOAD_DRY_RUN` specifies `PIPELINE_UPLOAD` in the name.
- `BUILDKITE_AGENT_INCLUDE_RETRIED_JOBS` namespaces to `BUILDKITE_AGENT_`.

I think my preference would be `BUILDKITE_AGENT_PIPELINE_UPLOAD_REJECT_SECRETS` which is kinda too long but also very clear 🤷‍♂️

Here's some examples of server-specified build job environment vars with `BUILDKITE_PIPELINE_` namespace that I think we should avoid mixing these config flags up with:

```
BUILDKITE_PIPELINE_ID="350b5c5f-dafe-432b-a16a-4d730d3d81e1"
BUILDKITE_PIPELINE_NAME="Agent"
BUILDKITE_PIPELINE_PROVIDER="github"
BUILDKITE_PIPELINE_SLUG="agent"
```

redaction/redactor.go
```diff
@@ -321,7 +321,7 @@ func GetKeyValuesToRedact(logger shell.Logger, patterns []string, environment ma
 
 			if matched {
 				if len(varValue) < RedactLengthMin {
-					logger.Warningf("Value of %s below minimum length and will not be redacted", varName)
+					logger.Warningf("Value of %s below minimum length (%d bytes) and will not be redacted", varName, RedactLengthMin)
```
pda, line 324: Technically I _think_ the length limit is in Unicode [runes](https://go.dev/blog/strings), not bytes.
I think the 4-rune string “🔑🔑🔑🔑” would fall short of a `RedactLengthMin = 8`, even though it's 16 bytes.

pda (commented):
Good plan 👍🏼

redaction/redactor.go
pda: (but with “rune” being a Go-specific term, I guess we'd call them “characters”)

moskyb: `len("some string")` returns the number of bytes in a string, rather than the number of runes. if we wanted to count runes here (which wouldn't be unreasonable i think, it's unlikely but not out of the question that someone uses é in their password), we'd have use `utf8.RuneCountInString()` - see [this go playground](https://go.dev/play/p/dzWSRVQ4FPI)

clicommand/pipeline_upload.go
moskyb: yeah, agreed, `BUILDKITE_AGENT_PIPELINE_UPLOAD_REJECT_SECRETS` is probably the least namespace-conflicty, and also does exactly what it says on the tin.

might be worth adding "standardise envar namespaces" to our list of things to do for v4.

clicommand/pipeline_upload.go
moskyb: eh.... when i initially did this, i told myself it was for terminal compatibilty (and because my editor pointed it out to me), but realistically i think we can trust that basically everyone is using a terminal that is UTF-8 compatible. i also scatter emoji all throughout my codebases too, so i'm a bit of a hypocrite here :P

i'll revert it back to using the fancy apostrophe

clicommand/pipeline_upload.go
moskyb: cool! no worries, i'll use `this pipeline will still be uploaded`.

moskyb: 🤔 i might update the error message when we refuse to update as well, just so the wording is a little more consistent.

redaction/redactor.go
pda: Ah — thanks for fact checking my fake news :)

clicommand/pipeline_upload.go
pda: Looks like it's still `BUILDKITE_PIPELINE_UPLOAD_REJECT_SECRETS` in the code, which is fine if that's your decision, but it's different to your comment above 😉

pda (approved):
Looking good, up to you if you want to fix https://github.com/buildkite/agent/pull/1589#discussion_r832763127

clicommand/pipeline_upload.go
moskyb: oh whoops! you're totally right, i can't read 😶

moskyb: fixed :)

PR #1590, opened 2022-03-23
Support looking up meta-data by build id

lox:
Currently the agent supports looking up meta-data by job, but this can be awkward in triggered builds where you are trying to lookup meta-data on the original build.

buildkite-agent meta-data get --build 07cfd694-0c0f-4dd1-bc0a-bfbe0327222

This would require backend changes (implementation of `/build/{build-id}/data/get`), PR for discussion.

### DIFF
--- a/api/meta_data.go
+++ b/api/meta_data.go
@@ -46,6 +46,24 @@ func (c *Client) GetMetaData(jobId string, key string) (*MetaData, *Response, er
 	return m, resp, err
 }
 
+// Gets the meta data value by build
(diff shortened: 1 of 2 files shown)


ngan:
@buildkite-agent-releaser I would love to see this get added. We’re waiting on this exact future to unblock some quick and easy pipeline to pipeline communication. Thank you!

tessereth:
Looks like this has been implemented in https://github.com/buildkite/agent/pull/2025 and released in v3.46.0.

Thanks for the suggestion!

Issue #1391, opened 2021-03-05
A list of breaking changes we're holding back for 4.0


chloeruka:
This is an ongoing list of breaking changes being held back for the next major release of Buildkite agent.

- [ ] Change default shell signal from SIGTERM to SIGINT (so we can signal process groups safely)

See also: [Elastic Stack for AWS: A list of breaking changes we're holding back for 6.0](https://github.com/buildkite/elastic-ci-stack-for-aws/issues/792)

PR #1593, opened 2022-03-23
Disallow uploading pipelines containing interpolated secrets by default

moskyb:
Further to #1589, pipeline secret redaction should become default in Agent v4.

This PR makes it so that default behaviour is to disallow pipeline uploads containing interpolations of potentially secret environment variables. We add flag to `buildkite-agent pipeline upload` to allow uploading pipelines with these secrets, but note in the CLI help and the log output that this behaviour is insecure.

We won't merge this until we release Agent v4


yob:
> pipeline secret redaction should become default in Agent v4.

I don't know if this issue is complete, but there's an historical attempt to list the breaking  changes we're saving for v4 over in #1391

moskyb:
@yob already in there :)

DrJosh9000:
Closing in favour of #3897

PR #1597, opened 2022-03-24
Run docker builds on the elastic-builders buildkite queue

moskyb:
**🤔 Problem:** A recent change to buildkite's buildkite setup changed it so that only a subset of buildkite agent queues have access to ECR. this is a useful security change, but unfortunately, the docker builds for the agent run on the default queue

**✅ Solution:** Move our docker builds to the `elastic-builders` queue, which has access to ECR.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -117,6 +117,8 @@ steps:
     steps:
       - name: ":docker: {{matrix}} build"
         key: build-docker
+        agents:
+          queue: elastic-builders
         depends_on:
           - build-binary
           - set-metadata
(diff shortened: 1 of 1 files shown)


yob (approved):
Changing the elastic-builder seems reasonable to me ✅

Purely out of interest: why did you add the group as well?

moskyb:
ah, mostly as a "while i'm here" thing, which i'm happy to revert and make into its own PR, if you want

yob:
> i'm happy to revert this and make into its own PR, if you want

no no, I was just curious. I'm still learning about groups myself.

PR #1599, opened 2022-03-27
Clear up usage instructions around --disconnect-after-idle-timeout and --disconnect-after-job

### LINKED ISSUE #1598: --disconnect-after-idle-timeout=0 has unexpected behavior
When I set ScaleInIdlePeriod=0, I expect agents to be terminated immediately upon their job being completed. Instead, the behavior is observed, is that they are never terminated. This recently left over 60 instances running idle in my AWS account for a day.

I'd like for either the behavior to be updated, or the docs to reflect this behavior so others don't fall into this trap.

I've considered setting ScaleInIdlePeriod=1 for this use case. But the real problem was setting this to 0 and not knowing this meant my agents (and EC2 instances in this scenario), were never terminated.

**Additional context**

I am using the elastic-ci-stack, here is my terraform resource so you can see my cloudformation parameters:
```
resource "aws_cloudformation_stack" "data_flow_agent_stack" {
  name = "data-flows-curative-build-kite-agents"
  # https://github.com/buildkite/elastic-ci-stack-for-aws/blob/3fccc8458d3b91e05e599779195bcb70749b4eea/templates/aws-stack.yml#L93-L437
  parameters = {
    # 2 VCPU 4 GB Memory per agent
    AgentsPerInstance                     = "1"
    BootstrapScriptUrl                    = "s3://build-kite-secrets/bootstrap.sh"
    InstanceType                          = "c5.large"
    BuildkiteAgentTokenParameterStorePath = "/build_kite/agent_token"
    EnableDockerUserNamespaceRemap        = "true"
    BuildkiteQueue                        = "data_flows"
    VpcId                                 = data.aws_vpc.main.id
    ECRAccessPolicy                       = "full"
    Subnets                               = join(",", data.aws_subnet.private.*.id)
    MinSize                               = "0"
    MaxSize                               = "200"
    ScaleInIdlePeriod                     = 0
    SecretsBucket                         = aws_s3_bucket.build_kite_secrets.id
  }

  capabilities  = ["CAPABILITY_NAMED_IAM", "CAPABILITY_AUTO_EXPAND"]
  template_body = file("${path.module}/template.yaml")
  tags = {
    CostBucket = "ci-cd"
  }
}
```
My use case for this agent queue, is to execute a parallelized task that is seldom executed. I expect there to be 0 of these agents most of the time, then a big spike of around 25 in service, then back down to 0 after the tasks are complete.

I have forked the elastic-ci-stack template, but this particular queue is not using any of those changes.

moskyb:
**This PR:** Cleans up a bit of wording around the `--disconnect-after-idle-timeout` and `--disconnect-after-job` flags, noting that when set to 0, `--disconnect-after-idle-timeout` means "don't disconnect idle agents".

Closes #1598

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -215,13 +215,13 @@ var AgentStartCommand = cli.Command{
 		},
 		cli.BoolFlag{
 			Name:   "disconnect-after-job",
-			Usage:  "Disconnect the agent after running a job",
+			Usage:  "Disconnect the agent after running exactly one job. When used in conjunction with the --spawn flag, each worker booted will run exactly one job",
 			EnvVar: "BUILDKITE_AGENT_DISCONNECT_AFTER_JOB",
 		},
 		cli.IntFlag{
 			Name:   "disconnect-after-idle-timeout",
 			Value:  0,
-			Usage:  "If no jobs have come in for the specified number of seconds, disconnect the agent",
+			Usage:  "The maximum idle time in seconds to wait for a job before disconnecting. The default of 0 means no timeout",
 			EnvVar: "BUILDKITE_AGENT_DISCONNECT_AFTER_IDLE_TIMEOUT",
 		},
 		cli.IntFlag{


tessereth (approved):
LGTM

clicommand/agent_start.go
```diff
@@ -214,13 +214,13 @@ var AgentStartCommand = cli.Command{
 		},
 		cli.BoolFlag{
 			Name:   "disconnect-after-job",
-			Usage:  "Disconnect the agent after running a job",
+			Usage:  "Disconnect the agent as soon as the job they're running completes",
```
pda, line 217: What do you think about "Disconnect the agent after running exactly one job"?

(Aside: I think the agent has been “it” so far, probably never “they”. I guess that could be open to change?)

pda: The big question in my mind around these flags is: how does it interact with `--spawn=N` when the one agent process has multiple agent worker connections?

I'm not sure whether or not this is the right place to try and clarify that.

clicommand/agent_start.go
```diff
@@ -214,13 +214,13 @@ var AgentStartCommand = cli.Command{
...
 			EnvVar: "BUILDKITE_AGENT_DISCONNECT_AFTER_JOB",
 		},
 		cli.IntFlag{
 			Name:   "disconnect-after-idle-timeout",
 			Value:  0,
-			Usage:  "If no jobs have come in for the specified number of seconds, disconnect the agent",
+			Usage:  "If no jobs have come in for the specified number of seconds, disconnect the agent. Note that the default for this flag is 0, and means \"don't disconnect idle agents\". If you want to disconnect agents as soon as they become idle, use the --disconnect-after-job flag instead",
```
pda, line 223: I personally would lean towards brevity:

"The maximum idle time in seconds to wait for a job before disconnecting. The default of 0 means no timeout."

pda: This one carries the same extra questions about how it behaves in `--spawn` mode.

clicommand/agent_start.go
moskyb: so after digging into this a bit, the interaction is that these flags work on an agent-worker basis, rather than an agent basis - that is, this flag actually means "disconnect the agent after all of the workers it runs have completed exactly one job". 

this is kinda confusing, butttttttt i think it's okay to leave the interactions between `--spawn` and the flags whose docs i'm changing here, because it's unlikely(?) that they'll be used in concert. the upshot is that `--spawn 5 --disconnect-after-job` could be used as a "run exactly this many jobs and then exit, i suppose 🤷

moskyb: the other thing we could do is just encode this further context in the help text - 
```
when used in conjunction with the `--spawn` flag, each worker booted will run exactly one job.
```
... or something

clicommand/agent_start.go
moskyb: this one looks to be designed with `--spawn` in mind, where `--disconnect-after-job` isn't. when `--disconnect-after-idle-timeout` is specified, the agent will only disconnect if _all_ of the workers are idle - see [agent/agent_worker.go:234](https://github.com/buildkite/agent/blob/main/agent/agent_worker.go#L234=)

ticky (approved):

PR #1600, opened 2022-03-30
fix: don't warn about unredacted vars if var is empty

joscha:
This change removes the warning for an unredacted variable if said variable is empty (e.g. there is nothing to be redacted anyway).

We have a few variables that can be empty; each of these produces an insane amount of warnings in each lifecycle execution:

Infra__apply_terraform__48053

(^ this is only one variable and only a subset of lifecycles)


moskyb (approved):
love it 🎉

PR #1601, opened 2022-04-01
Revert "Create directories only accessible by current user and group"

moskyb:
Reverts buildkite/agent#1580

#1580 broke docker workflows for quite a few users, so we'll need to take another pass at folder permissions sometime in the future, in such a way that it doesn't break existing behaviour.


pda (approved):
Approved, but noting that this is a short-term workaround pending further investigation into controlling these directory permissions without breaking Docker workflows.

PR #1606, opened 2022-04-04
Fix race condition in bootstrap.go

moskyb:
**🤔 Problem:** In bootstrap.go, we were accessing a variable `ctx` in a goroutine, but further down in the function, the `ctx` variable was being wrapped with a different context and rebound to the wrapped context, so the goroutine was referring to a variable that didn't exist anymore.

It's worth noting that this race condition has been rearing its head more often recently, and i'm not entirely sure why - the code causing the race has been in place for almost 2 years, so i can only think that until recently we weren't often calling `ctx.Done()` on the bootstrap process? but that also doesn't make much sense, because that context should have `Done()` called on it every time the bootstrap exits successfully.

So, in short, i'm not 100% sure what's going on, but this change should hopefully fix the issue, or at the very least not make it any worse.

**✅ Solution:** Only start the goroutine listening on the context's done channel after the context has been wrapped.

Additionally, when running tests in CI, run them in race detector mode. This should catch most (but not necessarily all) data races within the agent's codebase, provided that they're covered by tests.

Note that we're adding the race detector tests as a soft-fail step in CI - there's a race condition in `bintest` that we're hitting in one of our tests that i don't want to fail the build, but we should also probably get around to fixing that at some point. 

Fixes PIP-381

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -19,6 +19,19 @@ steps:
         config: .buildkite/docker-compose.yml
         run: agent
 
+  - name: ":hammer: :oncoming_automobile: :linux: arm64"
+    key: test-race-linux-arm64
+    command: ".buildkite/steps/tests.sh -race"
+    artifact_paths: junit-*.xml
+    soft_fail:
+      - exit_status: "*"
+    agents:
+      queue: elastic-runners-arm64
+    plugins:
+      docker-compose#v3.0.0:
+        config: .buildkite/docker-compose.yml
+        run: agent
+
   - name: ":hammer: :linux: arm64"
     key: test-linux-arm64
     command: ".buildkite/steps/tests.sh"
--- a/.buildkite/steps/tests.sh
+++ b/.buildkite/steps/tests.sh
@@ -7,7 +7,7 @@ echo arch is "$(uname -m)"
 GO111MODULE=off go get gotest.tools/gotestsum
 
 echo '+++ Running tests'
-gotestsum --junitfile "junit-${OSTYPE}.xml" -- -count=1 -failfast ./...
+gotestsum --junitfile "junit-${OSTYPE}.xml" -- -count=1 -failfast "$@" ./...
 
 echo '+++ Running integration tests for git-mirrors experiment'
-TEST_EXPERIMENT=git-mirrors gotestsum --junitfile "junit-${OSTYPE}-git-mirrors.xml" -- -count=1 -failfast ./bootstrap/integration
+TEST_EXPERIMENT=git-mirrors gotestsum --junitfile "junit-${OSTYPE}-git-mirrors.xml" -- -count=1 -failfast "$@" ./bootstrap/integration
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -81,6 +81,11 @@ func (b *Bootstrap) Run(ctx context.Context) (exitCode int) {
 		b.shell.InterruptSignal = b.Config.CancelSignal
 	}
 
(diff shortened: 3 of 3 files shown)


tessereth (approved):
LGTM. I'd love to just replace the existing tests with the race-detector ones but I don't want to block running the race detector on fixing all the bugs. Perhaps you can add a comment to the pipeline.yml saying the `-race` tests should replace the non-race tests once they pass?

moskyb:
@tessereth yeah, agreed, i also want to always run the race detector when we run tests. i'll add that comment to the pipeline, but fixing the race condition in bintest probably means digging into bintest internals, which is a bit scary because bintest is essentially sorcery as far as i'm concerned.

regardless, i'll chuck it on the backlog and hopefully we'll find time and expertise to fix it sometime soon.

pda:
Do you have an example crash stack trace etc?
I'm curious what the problem looked like when it manifested.

> the ctx variable was being wrapped with a different context and rebound to the wrapped context, so the goroutine was referring to a variable that didn't exist anymore.

Is that exactly true? I'm a bit unclear on the semantics of reassigning a variable that another goroutine has in scope, but I would have thought it would be one of these outcomes:

- The goroutine keeps the original non-wrapped `ctx`, while `ctx` elsewhere becomes wrapped, or…
- The goroutine's `ctx` now points at the wrapped value.

I would guess it would be the first one? But I can't see a version of that where the `ctx` held in the goroutine doesn't exist anymore.

So… is the problem that the goroutine is working with the unwrapped context, but elsewhere the wrapped context is being used? Even then, I imagine that would be fine in many cases.

Is this something that only happens if OpenTracing is enabled?
I don't think that's a commonly used feature.

pda:
> Is this something that only happens if OpenTracing is enabled?
> I don't think that's a widely used feature.

Oh, I see we do `ctx = opentracing.ContextWithSpan(ctx, span)` regardless of whether tracing is enabled 🤔

https://github.com/buildkite/agent/blob/acedee3107d560ece2e0ac9bdc8a2ad0070408e5/bootstrap/bootstrap.go#L305-L306

pda:
> the code causing the race has been in place for almost 2 years

The code was written ~2 years ago (#1273) but only released in [v3.27.0](https://github.com/buildkite/agent/releases/tag/v3.27.0) in Feb 2021.
Still, quite a while ago.

yob:
We recently bumped the version of go used to compile official builds (in 3.34.0, #1549), could that have changed the timing and made a race more likely to go bad?

moskyb:
@pda here's a stack trace from one customer: 
```
panic: interface conversion: interface {} is , not chan struct {}
 
goroutine 18 [running]:
context.(*cancelCtx).Done(0x0)
	/usr/local/go/src/context/context.go:361 +0x1c6
github.com/buildkite/agent/v3/bootstrap.(*Bootstrap).Run.func1()
	/work/bootstrap/bootstrap.go:87 +0x32
created by github.com/buildkite/agent/v3/bootstrap.(*Bootstrap).Run
	/work/bootstrap/bootstrap.go:85 +0x252
```

there's definitely something funky going on with the context that has `Done` called on it on line 87, but to your point, i have no idea what it actually is. the thing about referencing a variable that's since been rebound is a hypothesis at best, albeit one that's supported by the fact that the race detector no longer complaining about that line when we switch the order around.

@yob that could be it. the one mention of this specific issue that i could find is [this issue on golang itself](https://github.com/golang/go/issues/51789), though one commenter says there that data races might be an issue, so i'm really unclear on what's up

pda (approved):
Seems like a good change, even if there's still some mystery around the specifics 👍🏼

PR #1612, opened 2022-04-13
Install gotestsum in the recommended way for Go 1.17+

moskyb:
It's been causing occasional issues when installing on Windows, so hopefully installing it in the [idiomatic way](https://github.com/gotestyourself/gotestsum#install=) will fix these


pda (approved):

moskyb:
this is blocked pending us upgrading the elastic stack we use to run the windows builds - they still use go 1.15 to run tests, so they don't have the `go install` command

.buildkite/steps/tests.sh
```diff
@@ -4,7 +4,7 @@ set -euo pipefail
 go version
 echo arch is "$(uname -m)"
 
-GO111MODULE=off go get gotest.tools/gotestsum
+go install gotest.tools/gotestsum@v1.8.0
```
pda, line 7: Am I correct in thinking image-builder already installs this, but installing it again is probably a reasonably fast no-op?

moskyb: Yep! pretty much. Also, for linux, we test in a go container, which doesn't have this installed. The windows images do; but we want to run the same script on each, ideally

pda: Ah yeah!

PR #1628, opened 2022-05-01
Use Alpine 3.15.4 in agent release images

moskyb:
We currently use Alpine 3.12 in our Alpine release images (which are the defaults, ie if you run `docker pull buildkite/agent`, it'll give you an image based on alpine). 3.12 is [EOL as of yesterday](https://alpinelinux.org/releases/), and is also different to the alpine that we use in the build process, so we should try to keep these in line.

This PR:
- Updates the version of Alpine we use from 3.12 to 3.15.4, the latest available
- Formats all of the dockerfiles in `packaging/docker` (my editor autoformatted the two i changed, so i figured it was best to do them all)
- Adds the `packaging/docker` dir to the list of things that dependabot should check for updates for - this means that hopefully this won't happen again :)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,6 +1,10 @@
 version: 2
 updates:
   - package-ecosystem: "docker"
+    directory: "/packaging/docker"
+    schedule:
+      interval: "monthly"
+  - package-ecosystem: "docker"
     directory: "/.buildkite"
     schedule:
       interval: "monthly"
--- a/packaging/docker/sidecar/Dockerfile
+++ b/packaging/docker/sidecar/Dockerfile
@@ -1,10 +1,10 @@
-FROM alpine:3.12
+FROM alpine:3.15.4
 
 RUN mkdir /buildkite \
-          /buildkite/builds \
-          /buildkite/hooks \
-          /buildkite/plugins \
-          /buildkite/bin
+  /buildkite/builds \
+  /buildkite/hooks \
+  /buildkite/plugins \
+  /buildkite/bin
 
 COPY buildkite-agent.cfg /buildkite/
 COPY buildkite-agent /buildkite/bin/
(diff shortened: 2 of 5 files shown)


sj26 (approved):
Seems like a good idea :+1:

packaging/docker/sidecar/Dockerfile
```diff
@@ -1,14 +1,14 @@
...
+  /buildkite/bin
 
 COPY buildkite-agent.cfg /buildkite/
 COPY buildkite-agent /buildkite/bin/
 
 FROM busybox:musl
 COPY --from=0 /buildkite /buildkite
-VOLUME /buildkite
\ No newline at end of file
+VOLUME /buildkite
```
pda: 💚

packaging/docker/ubuntu-18.04-linux/Dockerfile
```diff
@@ -5,20 +5,20 @@ ENV DOCKER_COMPOSE_VERSION=1.27.4
...
+    gnupg-agent \
+    jq \
+    openssh-client \
+    perl \
+    rsync \
+    software-properties-common \
     && curl -fsSL https://download.docker.com/linux/ubuntu/gpg | apt-key add - \
     && add-apt-repository \
-      "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable" \
+    "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable" \
```
pda, line 21: I find this change unfortunate, but not going to spend energy arguing with an autoformatter 😅

.github/dependabot.yml
```diff
@@ -1,5 +1,9 @@
 version: 2
 updates:
+  - package-ecosystem: "docker"
```
pda, line 3: 💚

PR #1630, opened 2022-05-05
SUPER DUPER WIP: Add opentelemetry support to the agent

moskyb:
/giphy it's happening


bootstrap/bootstrap.go
```diff
@@ -231,116 +238,54 @@ func (b *Bootstrap) extractTraceCtx() opentracing.SpanContext {
...
-		)
-		stopper = tracer.Stop
-	default:
-		b.shell.Commentf("An invalid tracing backend was given: %s. Tracing will not occur.", b.Config.TracingBackend)
-		fallthrough
-	case "":
-		t = opentracing.NoopTracer{}
-		stopper = func() {}
+	tags := Merge(GenericTracingExtras(b, *b.shell.Env), DDTracingExtras())
+	opts = slices.Grow(opts, len(tags))
```
moskyb, line 264: review note: i'm using the new in Go 1.18, experimental [`slices`](https://pkg.go.dev/golang.org/x/exp/slices) package to grow our slice to the correct size here.

it's experimental in that the API might change, rather than experimental in that it might not work (like this PR lol), but regardless, i'm happy to replace this with non-generic code if we're not comfortable using it.

moskyb: if you're wondering why i manually grow the slice rather than just appending willy nilly, when you call append on a slice in go, if there isn't enough capacity in the slice to add the element you want to add, it automatically doubles the capacity of the slice. This is fine in most cases (and honestly, probably fine here), but it can mean that we allocate more memory for a given slice than we strictly speaking need. 

Because we know how big the slice is going to get ahead of time (or at least, we can derive it from the sizes of other things) it pays to allocate only as much memory to the slice.

tracetools/span.go
```diff
@@ -2,28 +2,105 @@ package tracetools
...
+// operation name. It will also do some common/repeated setup on the span to keep code a little more DRY.
+// Will panic if tracerType is anything other than "datadog", "otel", or ""
+func StartSpanFromContext(ctx context.Context, operation string, tracerType string) (any, context.Context) {
+	switch tracerType {
+	case "datadog":
+		span, ctx := opentracing.StartSpanFromContext(ctx, operation)
+		span.SetTag(ddext.AnalyticsEvent, true) // Make the span available for analytics in Datadog
+		return span, ctx
+
+	case "otel":
```
moskyb, line 26: TODO: Constantise this and the valid backends in bootstrap.go so that we don't end up with two different sources of truth on what backends are valid

tracetools/span.go
```diff
@@ -2,28 +2,105 @@ package tracetools
...
-// code a little more DRY.
-func StartSpanFromContext(ctx context.Context, operation string) (opentracing.Span, context.Context) {
-	span, ctx := opentracing.StartSpanFromContext(ctx, operation)
-	// Make the span available for analytics in Datadog
-	span.SetTag(ddext.AnalyticsEvent, true)
-	return span, ctx
+// StartSpanFromContext will start a span appropriate to the given tracing backend from the given context with the given
+// operation name. It will also do some common/repeated setup on the span to keep code a little more DRY.
+// Will panic if tracerType is anything other than "datadog", "otel", or ""
+func StartSpanFromContext(ctx context.Context, operation string, tracerType string) (any, context.Context) {
```
moskyb, line 19:
```suggestion
func StartSpanFromContext(ctx context.Context, operation string, tracingBackend string) (any, context.Context) {
```

tracetools/span.go
```diff
@@ -2,28 +2,105 @@ package tracetools
...
+	case "otel":
+		ctx, span := otel.Tracer("buildkite_agent").Start(ctx, operation)
+		span.SetAttributes(attribute.String("analytics.event", "true"))
+		return span, ctx
+
+	case "":
+		return nil, ctx
+
+	default:
+		panic(fmt.Sprintf("Invalid tracing backend %q passed to tracetools.StartSpanFromContext", tracerType))
```
moskyb, line 35: i'm not sure that panicking is the right call here - it means that we crash if we get an invalid backend, rather than just not tracing. maybe silently dropping traces is bad too though?

tracetools/span.go
```diff
@@ -2,28 +2,105 @@ package tracetools
...
+	}
+}
+
+// Adds context to the span, either as opentracing tags, or as opentelemetry attributes
+// Will panic if span is anything other than an opentracing.Span, a trace.Span, or nil
+// Noop when span is nil
+func AddAttributesToSpan(span any, attributes map[string]string) {
+	switch span := span.(type) {
```
moskyb, line 43: most of these comments are to myself, but this one isn't: if anyone has any better ideas than "do a typeswitch every time you need to do something with a span" then i'm all ears, please come at me with better ideas than this

tracetools/span.go
```diff
@@ -2,28 +2,105 @@ package tracetools
...
+
+	case nil:
+		return
+
+	default:
+		panic(fmt.Sprintf("Invalid span type: %T passed to tracetools.AddAttributesToSpan", span))
+	}
 }
 
 // FinishWithError is syntactic sugar for opentracing APIs to add errors to a span
```
moskyb, line 62: TODO: Update this comment so that it isn't a lie

tracetools/span.go
cbandy: >  if anyone has any better ideas than "do a typeswitch every time you need to do something with a span" then i'm all ears

Have your own Span interface with implementations for OTel or OpenTracing.

```go
type Span interface {
  AddAttributes(map[string]string)
  FinishWithError(error)
  RecordError(error)
}

func StartSpanFromContext(…) (Span, context.Context) { … }
```

moskyb: yeah, that is the obvious solution in retrospect 😅 just a smidge embarassing that yesterday-ben didn't come up with that

moskyb: i'll adjust this to use interface funtimes. i have a kind of halway solution locallty at the moment, i don't think it'll be too hard to adapt

moskyb: alright, that's donezo, way nicer now. not even a single panic! thanks for the help @cbandy ^_^

moskyb:
I'm going to close this PR and re-open it with something based off of https://github.com/buildkite/agent/pull/1631, the rebase-hell is too deep for even a rebase enjoyer like myself

PR #1631, opened 2022-05-09
Modularise tracing backends

moskyb:
**🤔 Problem:** We're adding opentelemetry tracing support to the agent [soon](https://github.com/buildkite/agent/pull/1630), but the current state in the agent makes it difficult to add another tracing implementation, as the agent as it stands is very tied into opentracing and datadog.

**✅ Solution:** This PR makes the tracing initialisation process more modular, allowing us to more easily add other tracing backends in future. A quick rundown of the changes:
- Update the functions in tracetools/span.go to operate on an interface, able to be fulfilled by spans from multiple tracing implementations
- Refactor the tracer initialisation code in bootstrap/bootstrap.go:
  - Move to bootstrap/tracing.go
  - Instead of tracing with the OpenTracing `NoopTracer` when tracing is disabled, use a noop implementation of the interface from tracetools above. This allows us to not trace in a more vendor-agnostic way
  - Generally slice up the *Bootstrap.StartTracing() method into smaller, re-usable chunks where necessary


bootstrap/tracing.go
```diff
@@ -0,0 +1,155 @@
...
+		}
+
+		return b.startTracingDatadog(ctx)
+
+	case tracetools.BackendNone:
+		return &tracetools.NoopSpan{}, ctx, noopStopper
+
+	default:
+		b.shell.Commentf("An invalid tracing backend was provided: %q. Tracing will not occur.", b.Config.TracingBackend)
+		b.Config.TracingBackend = "" // Ensure that we don't do any tracing after this, some of the stuff in tracetools uses the bootstrap's tracking backend
```
moskyb, line 42: this step was necessary in previous implementations, because tracetools.StartSpanFromContext panicked when passed an invalid tracing backend, and this would stop that from happening. It's technically no longer necessary, but i still kinda like it from a defensive programming perspective. What do you think?

bootstrap/tracing.go
```diff
@@ -0,0 +1,155 @@
...
+	resourceName := b.OrganizationSlug + "/" + b.PipelineSlug + "/" + b.getLabel()
+	opts := []tracer.StartOption{
+		tracer.WithServiceName("buildkite_agent"),
+		tracer.WithSampler(tracer.NewAllSampler()),
+		tracer.WithAnalytics(true),
+		tracer.WithGlobalTag(ddext.ResourceName, resourceName),
+	}
+
+	tags := Merge(GenericTracingExtras(b, *b.shell.Env), DDTracingExtras())
+	opts = slices.Grow(opts, len(tags))
```
moskyb, line 68: > from [#1630](https://github.com/buildkite/agent/pull/1630#discussion_r865611722), which this PR was extracted from

review note: i'm using the new-in-Go-1.18, experimental [`slices`](https://pkg.go.dev/golang.org/x/exp/slices) package to grow our slice to the correct size here, because it's a lot clearer than using 
```Go
opts = append(opts, make([]tracer.StartOption, len(tags))...)[:len(opts)]
```

it's experimental in that the API might change, rather than experimental in that it might not work (like this PR lol), but regardless, i'm happy to replace this with non-generic code if we're not comfortable using it.

moskyb: if you're wondering why i manually grow the slice rather than just appending willy nilly (which is absolutely an option), when you call append on a slice in go, if there isn't enough capacity in the slice to add the element you want to add, it automatically doubles [0] the capacity of the slice. What this means in practice is allocating a whole other slice with the new capacity and copying everything over, which is less than ideal, but also not exactly bad.

This is fine in most cases (and honestly, probably fine here), but it can mean that we allocate more memory for a given slice than we strictly speaking need.

Because we know how big the slice is going to get ahead of time (or at least, we can derive it from the sizes of other things) it pays to allocate only as much memory to the slice.

[0] I think? I can't quite remember, but it grows it by some factor

tracetools/span.go
```diff
@@ -5,25 +5,78 @@ import (
...
+)
+
+// StartSpanFromContext will start a span appropriate to the given tracing backend from the given context with the given
+// operation name. It will also do some common/repeated setup on the span to keep code a little more DRY.
+// If an unknown tracing backend is specified, it will return a span that noops on every operation
+func StartSpanFromContext(ctx context.Context, operation string, tracingBackend string) (Span, context.Context) {
+	switch tracingBackend {
+	case BackendDatadog:
+		span, ctx := opentracing.StartSpanFromContext(ctx, operation)
+		span.SetTag(ddext.AnalyticsEvent, true) // Make the span available for analytics in Datadog
```
moskyb, line 23: I __think__ this line might not be necessary, as we add this tag during [tracer initialisation](https://github.com/buildkite/agent/pull/1631/files#diff-f80935c46a08d736109e6791c2b0c9a1b004a9c196a8a149b3a70d5c2d3d40c0R136), but it can't hurt to be double sure it's in there i guess? ¯\_(ツ)_/¯

sj26:
This is looking great, @moskyb!

moskyb:
Merging this PR is blocked by our windows build problem; the agents that run windows agent builds are running go 1.15, so they can't handle my incredibly powerful use of generics (the `maps` and `slices` libraries + the `any` keyword). I'm going to focus on updating our windows agents to get this build green.

sj26 (approved):
This looks great to me!

I'm not super across the golang specific details, and what to look for there. But given the test coverage and passing build it seems safe to merge and try. :+1:

bootstrap/tracing.go
cbandy: `slices.Grow` still uses the mechanics of `append`. It does not create a slice without more capacity than necessary: https://go.dev/play/p/cE5Fo_ydndQ

To control the capacity, one should use make and some [vanilla Go](https://go.dev/blog/slices):

```go
	tags := Merge(GenericTracingExtras(b, *b.shell.Env), DDTracingExtras())

	opts := make([]tracer.StartOption, 0, 4+len(tags))
	opts = append(opts,
		tracer.WithServiceName("buildkite_agent"),
		tracer.WithSampler(tracer.NewAllSampler()),
		tracer.WithAnalytics(true),
		tracer.WithGlobalTag(ddext.ResourceName, resourceName),
	)
	for k, v := range tags {
		opts = append(opts, tracer.WithGlobalTag(k, v))
	}
```

bootstrap/bootstrap.go
```diff
@@ -1532,9 +1423,9 @@ func (b *Bootstrap) runPreCommandHooks(ctx context.Context) error {
 
 // runCommand runs the command and adds tracing spans.
 func (b *Bootstrap) runCommand(ctx context.Context) error {
-	span, ctx := tracetools.StartSpanFromContext(ctx, "command")
+	span, ctx := tracetools.StartSpanFromContext(ctx, "command", b.Config.TracingBackend)
 	var err error
-	defer func() { tracetools.FinishWithError(span, err) }()
+	defer func() { span.FinishWithError(err) }()
```
moskyb, line 1428: All of these `defer func() { ... }` calls can be changed to 
```suggestion
	defer span.FinishWithError(err)

Comment (cbandy) [reply bootstrap/bootstrap.go:1428]: I'm not sure about that. The arguments to `defer` are evaluated at the `defer`. https://go.dev/blog/defer-panic-and-recover

The existing spelling is a closure that refers to the variable `err`.

File: bootstrap/tracing.go [hunk 5]
```diff
@@ -0,0 +1,155 @@
...
+
+// startTracingDatadog sets up tracing based on the config values. It uses opentracing as an
+// abstraction so the agent can support multiple libraries if needbe.
+func (b *Bootstrap) startTracingDatadog(ctx context.Context) (tracetools.Span, context.Context, stopper) {
+	resourceName := b.OrganizationSlug + "/" + b.PipelineSlug + "/" + b.getLabel()
+	opts := []tracer.StartOption{
+		tracer.WithServiceName("buildkite_agent"),
+		tracer.WithSampler(tracer.NewAllSampler()),
+		tracer.WithAnalytics(true),
+		tracer.WithGlobalTag(ddext.ResourceName, resourceName),
```
tessereth, line 64: This used to be set on the root span, not as a global tag. I think that means the resource tag will be added to all spans, and thus count every span as a separate job run event. That's probably bad.

bootstrap/tracing.go
```diff
@@ -0,0 +1,155 @@
...
+		"buildkite.retry":             retry,
+		"buildkite.parallel":          parallel,
+		"buildkite.rebuilt_from_id":   rebuiltFromID,
+		"buildkite.triggered_from_id": triggeredFromID,
+	}
+}
+
+func DDTracingExtras() map[string]any {
+	return map[string]any{
+		ddext.AnalyticsEvent:   true,
```
tessereth, line 136: This was previously set on the root span, not as a global tag. I imagine we need to keep that behaviour?

EDIT: Oh, I see we were adding this tag manually on every span in StartSpanFromContext. I guess it makes sense to add it globally.

bootstrap/bootstrap.go
moskyb: ah! you're right, good call

tracetools/span.go
```diff
@@ -5,25 +5,83 @@ import (
...
+		return &OpenTracingSpan{Span: span}, ctx
```
tessereth, line 29: Do you want to use NewOpenTracingSpan here?
```suggestion
		return NewOpenTracingSpan(span), ctx
```

bootstrap/bootstrap_test.go
```diff
@@ -78,35 +79,44 @@ func TestGetValuesToRedactEmpty(t *testing.T) {
...
-	}
-	span, ctx, stopper := b.startTracing(oriCtx)
+	assert.NoError(t, err)
+
+	span, _, stopper := b.startTracing(oriCtx)
+	assert.Equal(t, span, &tracetools.NoopSpan{})
+	span.FinishWithError(nil) // Finish the nil span, just for completeness' sake
+
+	// If you call opentracing.DefaultTracer() without having set it first, it returns a NoopTracer
+	// In this test case, we haven't touched opentracingt at all, so we get the NoopTracer
```
tessereth, line 97: The comment doesn't match the code here, should this be GlobalTracer?
```suggestion
	// If you call opentracing.GlobalTracer() without having set it first, it returns a NoopTracer
	// In this test case, we haven't touched opentracing at all, so we get the NoopTracer
```

tessereth (commented):
Other than concerns about globally adding the Resource tag, this looks great.

go.mod
```diff
@@ -3,8 +3,8 @@ module github.com/buildkite/agent/v3
 go 1.18
 
 require (
-	cloud.google.com/go v0.0.0-20170217213217-65216237311a
-	github.com/DataDog/datadog-go v3.7.2+incompatible
+	cloud.google.com/go v0.65.0
```
tessereth, line 6: I'm a little confused why all the dependencies changed in this PR. Could you add that to the PR description?

bootstrap/tracing.go
moskyb: ooh yes, good call!

bootstrap/tracing.go
moskyb: that's a good catch, i didn't think about that at all. will change.

bootstrap/bootstrap_test.go
moskyb: yep, you're right - this code was extracted from when there was also OpenTelemetry stuff in here. In Otel, it's called the DefaultTracer. Will shuffle this around to use the right nomenclature

go.mod
moskyb: this PR was extracted from the giga-PR that did all of this and also added opentelemetry, so I _think_ that this is a result of adding the opentel dependencies, and then they require updated other stuff? I could be wrong though. Regardless, i'll clean this up and make it so that we're only updating the Datadog stuff

moskyb: Cool, so I went back and made it so that A) We have minimal dependency changes, only the ones required by the dd-trace-go upgrade, and B) They all happen in the [`Update dd-trace-go`](https://github.com/buildkite/agent/pull/1631/commits/65f8d4fb1158a29dad3ce68387d43e021a88516e) commit.

I _think_ reason some of our direct dependencies (notably `cloud.google.com/go` and `github.com/DataDog/datadog-go`, but also some `golang.org/x/...` libraries) have changed is because both we and dd-trace-go depend on these libs, and dd-trace-go's dependent versions changed.

I've double checked all of the libs that have changed, and made sure that none of our use cases have been deprecated or removed.

bootstrap/tracing.go
```diff
@@ -0,0 +1,167 @@
...
+		b.shell.Commentf("An invalid tracing backend was provided: %q. Tracing will not occur.", b.Config.TracingBackend)
+		b.Config.TracingBackend = tracetools.BackendNone // Ensure that we don't do any tracing after this, some of the stuff in tracetools uses the bootstrap's tracking backend
+		return &tracetools.NoopSpan{}, ctx, noopStopper
+	}
+}
+
+func (b *Bootstrap) tracingResourceName() string {
+	label, ok := b.shell.Env.Get("BUILDKITE_LABEL")
+	if !ok {
+		label = "job"
```
tessereth, line 50: Should this branch also include the organization and pipeline slugs? I'm pretty sure it used to.

tessereth (approved):
Looks great except for one tiny comment. I'll pre-emptively approve so you don't have to come back to me.

bootstrap/tracing.go
moskyb: oh it did! lost in rebase hell, thanks for pointing this out

moskyb: oh, wait, these are just a couple of lines down now, [here](https://github.com/buildkite/agent/pull/1631/files/fec0ca97087532b9984909a969f50d7e48d9183c#diff-f80935c46a08d736109e6791c2b0c9a1b004a9c196a8a149b3a70d5c2d3d40c0R53)

tessereth: Oh, I just can't read. In my head this returned early. Never mind. 🙈

PR #1548, opened 2021-12-29
Migrate from OpenTracing to OpenTelemetry

rajatvig:
This moves the tracing setup from OpenTracing (now deprecated) to OpenTelemetry. The code is vendor agnostic and tracing data is sent over GRPC to the collector as configured in the environment.

Have removed the code in `tracetools` and inlined the span create and end in `bootstrap`.

There is a small demo setup of the OpenTelemetry Collector and Jaeger to see traces from the agents locally. 

This would very likely break any consumers of the current Datadog based tracing setup as it relies on a running Collector. However, as documented [here](https://docs.datadoghq.com/tracing/setup_overview/open_standards/), it is easy to get it stood up for Datadog.

PR #1632, opened 2022-05-10
Experimental OpenTelemetry Tracing Support

moskyb:
## 🤔 Problem: 

The times, they are a-changing, and so are distributed tracing standards. 

The CNCF, in their infinite wisdom, have [archived the OpenTracing project](https://www.cncf.io/blog/2022/01/31/cncf-archives-the-opentracing-project/), upon which our implementation of tracing in the agent was implemented. We'll continue to support OpenTracing tracing through datadog for the forseeable future, but people are rightly migrating to the newer standard of [OpenTelemetry](https://opentelemetry.io/). 

OpenTelemetry has a tracing provider that's well-defined, supports multiple backends (Datadog, Jaeger, etc), and is all-around a good and useful piece of software. We should provide functionality within the agent to trace CI runs using OpenTelemetry tracing.

There's a bit of an issue in that while OpenTracing and OpenTelemetry are similar, they're just different enough that it's a bit of a hassle cognitively to have them both in the same place. In this PR you'll see a bunch of stuff where we're doing things with OpenTelemetry spans that are _just slightly_ different to what we'd do with OpenTracing spans. It doesn't hurt (or more accurately, it does hurt) that they share an acronym, and have similar names. 

## 💬 Cool, so what does this PR do?

 Building on the work done in #1631, we add a new tracing backend, `opentelemetry`, and associated utilities in the `tracetools` and `bootstrap` packages in order to be able to use them. This new tracing backend is behind an experiment called `opentelemetry-tracing`, and **is not** considered stable - we may change the implementation without warning.

I've confirmed that this implementation can egest tracing data into a local Jaeger instance, and that those traces look relatively normal (to my admittedly untrained eye, anyway).

Screen Shot 2022-05-17 at 5 09 18 PM

> You can find a JSON dump of the pictured trace [here](https://gist.github.com/moskyb/c91988ca4f549e4cb89f89065bd288a4)

## 🙅‍♂️ What doesn't it do?

 Notably missing from the OpenTelemetry implementation are equivalents of the methods in [`tracetools/propagate.go`](https://github.com/buildkite/agent/blob/main/tracetools/propagate.go). A large part of this is that after looking into this for quite a while, i'm still not exactly sure of:
- What these functions actually do in the context of opentracing
- What the best way to implement them in an OpenTelemetry context is
- If this functionality gets used

With this in mind, I'm open to bundling this lack of functionality into "well, it's experimental" and only implementing it only if it gets asked for, but this is also kind of a cop out, so if you have more information on this please let me know!

## 🏴‍☠️ Acknowledgements: 

Huge thanks to @rajatvig and his awesome work on #1548 - this PR would be in much, much rougher shape without all their code that ~~i stole~~ inspired me


KevinGreen:
Hey Ben, this probably isn't the best place to make this comment but I've noticed that rebuilds of the same job will get grouped under the same root span in datadog. This ends up showing the one span as taking as long as the first build + the delay before I click the rebuild button + the second build. It would be nice if there was a way to have a root span per build instead of per pipeline+sha

moskyb:
@KevinGreen this is definitely an issue that i'm interested in fixing - could you open an issue on the repo and include a screenshot or two so that i know what i'm looking for?

bootstrap/tracing.go
```diff
@@ -165,3 +241,22 @@ func Merge(maps ...map[string]any) map[string]any {
 
 	return merged
 }
+
+func (b *Bootstrap) toAttributes(extras map[string]any) []attribute.KeyValue {
```
tessereth, line 245: This returns otel specific attributes right? Maybe this could be called `toOpenTelemetryAttributes` or something to be clear it's only for one provider?

bootstrap/tracing.go
```diff
@@ -165,3 +241,22 @@ func Merge(maps ...map[string]any) map[string]any {
...
+		switch v := v.(type) {
+		case string:
+			attrs = append(attrs, attribute.String(k, v))
+		case int:
+			attrs = append(attrs, attribute.Int(k, v))
+		case bool:
+			attrs = append(attrs, attribute.Bool(k, v))
+		default:
+			b.shell.Warningf("Unknown attribute type (key: %v, value: %v (%T)) passed when initialising OpenTelemetry. This is a bug, submit this error message at https://github.com/buildkite/agent/issues", k, v, v)
+			b.shell.Warningf("OpenTelemetry will still work as intended, but the attribute %v and its value above will not be included", v)
```
tessereth, line 257: Maybe not quite `as intended`?
```suggestion
			b.shell.Warningf("OpenTelemetry will still work, but the attribute %v and its value above will not be included", v)
```

tracetools/span.go
```diff
@@ -72,6 +83,39 @@ func (s *OpenTracingSpan) RecordError(err error) {
...
+
+// FinishWithError adds error information to the OpentelemetrySpan if error isn't nil, and records the span as having finished
+func (s *OpenTelemetrySpan) FinishWithError(err error) {
+	s.RecordError(err)
+	s.Span.End()
+}
+
+// RecordError records an error on the given span.
+// Will panic if span is anything other than an opentracing.Span, a trace.Span, or nil
+// noop when span or error are nil
```
tessereth, line 109: Is this comment accurate? Doesn't `s` have to be an OpenTelemetrySpan?

tessereth (approved):
Looks great! I'm so excited this is happening.

I don't suppose there's anything in `tracing.go` that's worth unit testing? I guess there aren't that many branches so if it works at all it's probably all working. /shrug

bootstrap/tracing.go
moskyb: heh, yes, good point :P

tracetools/span.go
moskyb: yeah, good point. this one is a leftover from renaming things, and they have such similar names that it's really easy to miss.

PR #1635, opened 2022-05-12
Log levels

moskyb:
**🤔 Problem:** Some of the time, users of the buildkite-agent might not necessarily want or need all of the logs we're spitting out, either for cost reasons (storing logs can get pricy) or for readability reasons (i don't necessarily care about everything the agent is doing all the time.

**✅ Solution:** Add a cli flag, `--log-level` and an accompanying envar `BUILDKITE_AGENT_LOG_LEVEL`, to set the level of the logger on the agent. Now, if I set `--log-level error`, the buildkite agent logger will only output logs of level error or higher. This flag is enabled for all CLI commands, except `bootstrap`.

This PR is best read commit-by-commit.

### DIFF
--- a/clicommand/global.go
+++ b/clicommand/global.go
@@ -48,8 +48,15 @@ var NoHTTP2Flag = cli.BoolFlag{
 
 var DebugFlag = cli.BoolFlag{
 	Name:   "debug",
-	Usage:  "Enable debug mode",
+	Usage:  "Enable debug mode. Synonym for `--log-level debug`. Takes precedence over `--log-level`",
 	EnvVar: "BUILDKITE_AGENT_DEBUG",
+}
+
+var LogLevelFlag = cli.StringFlag{
+	Name:   "log-level",
+	Value:  "notice",
+	Usage:  "Set the log level for the agent, making logging more or less verbose. Defaults to notice",
+	EnvVar: "BUILDKITE_LOG_LEVEL",
 }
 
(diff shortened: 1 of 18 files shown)


clicommand/global.go
```diff
@@ -48,10 +48,17 @@ var NoHTTP2Flag = cli.BoolFlag{
...
-	Usage:  "Enable debug mode",
+	Usage:  "Enable debug mode. Synonym for `--log-level debug`. Takes precedence over `--log-level`",
 	EnvVar: "BUILDKITE_AGENT_DEBUG",
 }
 
+var LogLevelFlag = cli.StringFlag{
+	Name:   "log-level",
+	Value:  "notice",
+	Usage:  "Set the log level for the agent, making logging more or less verbose. Defaults to notice",
+	EnvVar: "BUILDKITE_LOG_LEVEL",
```
sj26, line 59: The rest of these have `BUILDKITE_AGENT_` prefix — was that intentional? Do you see this as broader than just the agent?

sj26 (commented):
Looks pretty good to me! The env name doesn't fit the pattern of the rest, is that intentional?

I'll let one of our more dedicated gophers do a pass for approval.

clicommand/global.go
moskyb: 😶 you're right, i'll fix this up

tessereth (approved):
LGTM!

PR #1639, opened 2022-05-16
Fix go fmt

moskyb:
We recently [merged a PR](https://github.com/buildkite/agent/pull/1579) that had been opened before we had CI checking that gofmt was being fulfilled, and so wasn't rejected on that basis. This PR fixes that formatting error.


jordandcarter (approved):
Epic 😉

PR #1640, opened 2022-05-16
Make unreleased changelog entry for Agent 3.36

moskyb:
We're not releasing 3.36 yet, but I think it's a good idea to have an unreleased changelog, so that:
- It's easier to add to over time
- Our users know what's on the menu for the next version


moskyb:
Superseded by #1644

PR #1643, opened 2022-05-19
Skip pre-exit on cancel test on darwin

moskyb:
It doesn't work, and hasn't for ages :/


pda (approved):
😢

but 😁 because yeah it never works on my Mac

PR #1645, opened 2022-05-19
Add valid tracing backends to --tracing-backend usage text

moskyb:


clicommand/agent_start.go
```diff
@@ -450,7 +450,7 @@ var AgentStartCommand = cli.Command{
 		},
 		cli.StringFlag{
 			Name:   "tracing-backend",
-			Usage:  "The name of the tracing backend to use.",
+			Usage:  fmt.Sprintf("The name of the tracing backend to use. Valid values are: %q", maps.Keys(tracetools.ValidTracingBackends)),
```
pda, line 453: I was curious how this renders, and in checking discovered a bit of nondeterminism coming from Go's unordered maps:

```
$ for i in $(seq 100); do ./buildkite-agent start --help | grep tracing-backend | cut -c56- ; done | sort | uniq -c
  12 The name of the tracing backend to use. Valid values are: ["" "datadog" "opentelemetry"] [$BUILDKITE_TRACING_BACKEND]
  73 The name of the tracing backend to use. Valid values are: ["datadog" "opentelemetry" ""] [$BUILDKITE_TRACING_BACKEND]
  15 The name of the tracing backend to use. Valid values are: ["opentelemetry" "" "datadog"] [$BUILDKITE_TRACING_BACKEND]
```

The one with the empty string in the middle is particularly unusual :)

I think my preferred format would be:

> Enable tracing for build jobs by specifying a backend (opentelemetry, datadog)

pda: Also, with the experiment flag in mind:

https://github.com/buildkite/agent/blob/46279b29e992197246713ff7f170eaf5d27d7089/bootstrap/tracing.go#L52-L56

Maybe that needs reference in the help message…

> Enable tracing for build jobs by specifying a backend; "datadog" (deprecated, OpenTracing) or "opentelemetry" (requires --experiment opentelemetry-tracing)

moskyb: Heh, i hadn't even thought about unordered maps! Will hardcode this.

Re: the experiment flag, i'm not so keen to mark the datadog tracing as deprecated just yet; it still works, and i'm not keen to have both of our tracing backends be in limbo states (one experimental, the other deprecated). At the very least i'd be keen to get Otel stable before we deprecate datadog tracing.

pda: Sounds good 👍🏼

pda (approved):

moskyb:
Superseded by #1647

PR #1651, opened 2022-05-26
bootstrap/shell: round off command durations

kevinburkesegment:
It is unlikely that anyone will care about how much time a command
took more than about 5 significant digits:

- If you are expecting a command to take a few milliseconds, maybe you
care down to the microsecond.

- If you are expecting a command to take a few hours, maybe you care
how many minutes or seconds, but I very much doubt you care about how
many microseconds.

By default, Buildkite reports durations with extreme precision, which
can be distracting. Instead, round off durations so that at most 5
digits are present in the displayed duration.

Here's what the displayed output looks like, both before and after, for builds of different durations:

```
before: 1µs
after:  1µs

before: 10.07µs
after:  10.07µs

before: 100.47µs
after:  100.47µs

before: 1.00059ms
after:  1.0006ms

before: 10.02081ms
after:  10.021ms

before: 100.41318ms
after:  100.41ms

before: 1.00954425s
after:  1.0095s

before: 10.0612254s
after:  10.061s

before: 1m40.08240456s
after:  1m40.08s

before: 16m46.462033s
after:  16m46.5s

before: 2h46m47.60398084s
after:  2h46m50s

before: 27h57m22.63669287s
after:  27h57m20s
```

### DIFF
--- a/bootstrap/shell/shell.go
+++ b/bootstrap/shell/shell.go
@@ -461,6 +461,34 @@ type executeFlags struct {
 	PTY bool
 }
 
+func round(d time.Duration) time.Duration {
+	// The idea here is to show 5 significant digits worth of time.
+	// If your build takes 2 hours, you probably don't care about the timing
+	// being reported down to the microsecond.
+	switch {
+	case d < 100*time.Microsecond:
+		return d
+	case d < time.Millisecond:
+		return d.Round(10 * time.Nanosecond)
+	case d < 10*time.Millisecond:
+		return d.Round(100 * time.Nanosecond)
+	case d < 100*time.Millisecond:
+		return d.Round(time.Microsecond)
+	case d < time.Second:
+		return d.Round(10 * time.Microsecond)
+	case d < 10*time.Second:
+		return d.Round(100 * time.Microsecond)
+	case d < time.Minute:
+		return d.Round(time.Millisecond)
+	case d < 10*time.Minute:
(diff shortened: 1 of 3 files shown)


bootstrap/shell/shell.go
```diff
@@ -461,6 +461,34 @@ type executeFlags struct {
...
 
+func round(d time.Duration) time.Duration {
+	// The idea here is to show 5 significant digits worth of time.
+	// If your build takes 2 hours, you probably don't care about the timing
+	// being reported down to the microsecond.
+	switch {
+	case d < 100*time.Microsecond:
+		return d
+	case d < time.Millisecond:
+		return d.Round(time.Microsecond / 100)
```
moskyb, line 472: [nit]
could we change these round calls in general to:
```suggestion
		return d.Round(time.Nanosecond * 10)
```

and its equivalents for the other switch branches -  IMO it's a lot easier to reason about "round to the nearest 10 nanoseconds" rather than "round to the nearest microsecond / 100

moskyb (commented):
LGTM, more or less! would you mind adding a test for the round function? it'd make for a better spec for how it works than a comment, i think

bootstrap/shell/shell.go
kevinburkesegment: Done

kevinburkesegment:
> would you mind adding a test for the round function?

Done

moskyb (approved):

PR #1653, opened 2022-05-26
Fix nil pointer deref when using --log-format json

### LINKED ISSUE #1652: Boot time segfault on v3.36.0
We just updated our agent to v3.36.0 from v3.35.0 and we're now seeing a segfault when our agent Docker container starts. We're using `buildkite/agent:3.36.0-ubuntu-20.04` as the base for our docker image.

The stack trace suggests it's while creating a logger but I don't know for sure. Our docker containers are logging to journald but from the application's point of view it should still be writing to stdout as usual.

Shouldn't crash

We've downgraded back to v3.35.0 for now.

**Additional context**
Logs: image

moskyb:
A bit of code that was accidentally included in #1635 causes the agent to panic with a nil pointer dereference, because we're calling a method on a logger that doesn't exist.

This PR removes that log call. Oy vey.

Fixes #1652

### DIFF
--- a/clicommand/global.go
+++ b/clicommand/global.go
@@ -127,7 +127,6 @@ func CreateLogger(cfg interface{}) logger.Logger {
 
 		l = logger.NewConsoleLogger(printer, os.Exit)
 	case `json`:
-		l.Warn("FORMAT IS JSON")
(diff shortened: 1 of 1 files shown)


KevinGreen (approved):
Thanks for the quick fix!

yob (approved):

ticky (approved):

PR #1655, opened 2022-05-29
Refactor env.Environment from wrapping a map to being a map

moskyb:
The main reason behind this refactor is the `*Environment.ToMap()` method. The implication behind the naming of this method is that it returns a copy of the environment in a map representation, but what it actually does is just return the underlying map that Environment is implemented on top of. This is used in a variety of ways across the codebase, and we actually reach into Environment's pockets andmodify this underlying map a fair bit. This in mind, so it makes sense (to me, anyway) to remove the confusingly-named middleman.

One possible drawback of this approach is that it makes it easier to accidentally "manually" set a k/v pair in th env, ie by calling `env["foo"] = "bar"`, buuuuuuuuuuuut it was already pretty easy by calling ToMap(), so i'm happy with this tradeoff, but I'd be very interested in review comments about this.

The impetus for this change is that I'm looking into tracing environment propagation at the moment and got confused/frustrated/sidetracked by the nature of the `*Environment.ToMap()` method. It's not strictly necessary in order to do propagation stuff - we're more than capable of implementing that without this PR - but it is a nice quality of life change regardless.

### DIFF
--- a/env/environment.go
+++ b/env/environment.go
@@ -8,22 +8,20 @@ import (
 
 // Environment is a map of environment variables, with the keys normalized
 // for case-insensitive operating systems
-type Environment struct {
-	env map[string]string
-}
+type Environment map[string]string
 
-func New() *Environment {
-	return &Environment{env: map[string]string{}}
+func New() Environment {
+	return Environment{}
 }
 
 // FromSlice creates a new environment from a string slice of KEY=VALUE
-func FromSlice(s []string) *Environment {
-	env := &Environment{env: make(map[string]string, len(s))}
+func FromSlice(s []string) Environment {
+	env := make(Environment, len(s))
 
 	for _, l := range s {
-		parts := strings.SplitN(l, "=", 2)
-		if len(parts) == 2 {
-			env.Set(parts[0], parts[1])
+		key, val, ok := strings.Cut(l, "=")
+		if ok {
+			env.Set(key, val)
 		}
 	}
 
@@ -31,13 +29,13 @@ func FromSlice(s []string) *Environment {
 }
 
 // Get returns a key from the environment
-func (e *Environment) Get(key string) (string, bool) {
-	v, ok := e.env[normalizeKeyName(key)]
+func (e Environment) Get(key string) (string, bool) {
+	v, ok := e[normalizeKeyName(key)]
 	return v, ok
 }
 
 // Get a boolean value from environment, with a default for empty. Supports true|false, on|off, 1|0
-func (e *Environment) GetBool(key string, defaultValue bool) bool {
+func (e Environment) GetBool(key string, defaultValue bool) bool {
 	v, _ := e.Get(key)
 
 	switch strings.ToLower(v) {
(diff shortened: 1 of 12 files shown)


env/environment.go
```diff
@@ -8,36 +8,34 @@ import (
...
 // Environment is a map of environment variables, with the keys normalized
 // for case-insensitive operating systems
-type Environment struct {
-	env map[string]string
-}
+type Environment map[string]string
 
-func New() *Environment {
-	return &Environment{env: map[string]string{}}
+func New() Environment {
```
moskyb, line 13: If you're wondering why we're changing all of these pointer receivers (`*Environment`) to value receivers (`Environment`), it's because [all maps are kinda pointers to maps](https://dave.cheney.net/2017/04/30/if-a-map-isnt-a-reference-variable-what-is-it) in go, and having pointers-to-pointers-to-maps would add another layer of indirection and make everything a bit annoying.

env/environment.go
```diff
@@ -8,36 +8,34 @@ import (
...
-func FromSlice(s []string) *Environment {
-	env := &Environment{env: make(map[string]string, len(s))}
+func FromSlice(s []string) Environment {
+	env := make(Environment, len(s))
 
 	for _, l := range s {
-		parts := strings.SplitN(l, "=", 2)
-		if len(parts) == 2 {
-			env.Set(parts[0], parts[1])
+		key, val, ok := strings.Cut(l, "=")
```
pda, line 22: Ooh fancy Go 1.18 [`strings.Cut`](https://pkg.go.dev/strings#Cut) 👍🏼

env/environment.go
```diff
@@ -8,36 +8,34 @@ import (
 
 // Environment is a map of environment variables, with the keys normalized
 // for case-insensitive operating systems
```
pda, line 10: I imagine the “with the keys normalized” part might be the original reason for wrapping the map?
To guard against non-normalized reads/writes.
But as you say, the liberal passing of `.ToMap()`'s mutable result to other scopes renders that less useful.

env/export.go
```diff
@@ -35,9 +35,9 @@ var endsWithUnescapedQuoteRegex = regexp.MustCompile("([^\\\\]\"\\z|\\A\"\\z)")
 //     TMP=C:\Users\IEUser\AppData\Local\Temp
 //     USERDOMAIN=IE11WIN10
 //
-func FromExport(body string) *Environment {
+func FromExport(body string) Environment {
 	// Create the environment that we'll load values into
-	env := &Environment{env: make(map[string]string)}
+	env := Environment{}
```
pda, line 40: Is it a possible problem that we're not `make()`ing the `map` anymore?
I tend to forget the rules around `make()` and map/slice literals.
But testing it quickly… map literal like `Environment{}` or `map[string]string{}` doesn't seem to need `make()` 👌🏼

pda (approved):
Seems like a good simplification 👍🏼

env/export.go
moskyb: i went down a bit of a rabbithole investigating whether `m := make(map[K]V)` and `m := map[K]V{}` are exactly equivalent and the only answer i can come up with is "probably?". The spec mentions them being "functionally identical", which doesn't mean that they're actually identical ahahha. Regardless, functionally identical is identical enough for me.

env/environment.go
moskyb: my kingdom for mutabilty modifiers... time to rewrite in rust?

PR #1656, opened 2022-05-30
Release Agent Edge any time a main build completes

moskyb:
Currently, we don't do any real pre-releasing for the agent. We have three release "streams", `edge`, `beta`, and `stable`, but we only really ever use them when making a stable release, which is underutilising them, IMO.

This PR repurposes the `edge` release stream, and makes it so that whenever we successfully merge to main, we'll publish the `edge` release stream with whatever's in main at that time. The main benefit of this is that it makes it easier for customers to get at new features prior to their full release, whether that's because they like to live dangerously, or if they're acceptance testing something for us.

To use this release stream, customers can use the plug `edge` into the `BuildkiteAgentRelease` parameter in the elastic stack. I would argue that all of our agent builds should run on the latest version of the agent, but actually making that happen is out of scope for this PR.

Many thanks to @toothbrush for their inspiration on this!

### DIFF
--- a/.buildkite/steps/upload-release-steps.sh
+++ b/.buildkite/steps/upload-release-steps.sh
@@ -45,13 +45,15 @@ block_step() {
 YAML
 }
 
-output_steps_yaml() {
+edge_steps_yaml() {
   echo "steps:"
 
   trigger_step \
     "edge ${agent_version}.${build_version}" \
     "agent-release-edge"
+}
 
+output_steps_yaml() {
   block_step "beta?"
 
   trigger_step \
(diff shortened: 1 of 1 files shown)


pda:
> The main benefit of this

😁

pda:
This might flush out some setups that have been using edge but are actually depending on the stability that accidentally came from us not using the edge channel properly. That said, most of the time the `main` branch is in a good state.

pda (approved):

moskyb:
alright, let's merge it and see if it actually works :P

PR #1658, opened 2022-05-30
Move from cloud.google.com/go to cloud.google.com/go/compute

moskyb:
It's the only module we actually use in cloud.google.com/go, and it looks like it's published as its own module now.

This PR supersedes https://github.com/buildkite/agent/pull/1657, which has something strange going on.


moskyb:
wait, i'm silly, #1657 is all good, it's doing the exact same thing as this pr lol

PR #1667, opened 2022-06-06
Okay, I give up, everything has 777 permissions again.

moskyb:
We've tried [a bunch](https://github.com/buildkite/agent/pull/1580) of [solutions](https://github.com/buildkite/agent/pull/1616) to make file permissions for directories created by the agent a bit more restrictive, so that other processes can't subtly change what the agent is doing, but every time we've tried, it's broken docker workflows for our customers because of docker namespace remapping, and the fact that docker (rightly!) recommends that users don't use the root user inside containers.

This PR reverts all changes to directory permissions, and goes back to letting the umask dictate what permissions files get. This means that in most cases, things will get assigned 0755 perms, but this can be overridden by the user.

### DIFF
--- a/agent/download.go
+++ b/agent/download.go
@@ -133,6 +133,7 @@ func (d Download) try() error {
 	}
 
 	// Now make the folder for our file
+	// Actual file permissions will be reduced by umask, and won't be 0777 unless the user has manually changed the umask to 000
 	err = os.MkdirAll(targetDirectory, 0777)
 	if err != nil {
 		return fmt.Errorf("Failed to create folder for %s (%T: %v)", targetFile, err, err)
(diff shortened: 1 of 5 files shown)


tessereth (approved):
I am sad that we couldn't ship this but I get it. Eventually we'll figure out how to restrict the file permissions without breaking docker setups.

agent/download.go
```diff
@@ -133,7 +133,7 @@ func (d Download) try() error {
 	}
 
 	// Now make the folder for our file
-	err = os.MkdirAll(targetDirectory, 0775)
+	err = os.MkdirAll(targetDirectory, 0777)
```
pda, line 136: As discussed, maybe we should put a brief comment on each of these, something like:

```suggestion
	err = os.MkdirAll(targetDirectory, 0777) // actual file mode reduced by system umask
```

pda (approved):

PR #1671, opened 2022-06-09
Remove mirror dir if clone fails

### LINKED ISSUE #1670: Core dump when cloning large mirrors leaves corrupt mirrors directory
We have a very large git repo that occasionally segfaults when cloning the git mirror:

```
# Creating "/builds/default-i-02009a8845326e246-9/cash/redacted"
--
  | $ cd /builds/default-i-02009a8845326e246-9/cash/redacted
  | # Using git-mirrors experiment 🧪
  | $ cd /mirrors
  | # Cloning a mirror of the repository to "/mirrors/org-49461806-github-com-squareup-redacted-git"
  | $ git clone --mirror -v -- org-49461806@github.com:squareup/redacted.git /mirrors/org-49461806-github-com-squareup-redacted-git
  | Cloning into bare repository '/mirrors/org-49461806-github-com-squareup-redacted-git'...
  | POST git-upload-pack (gzip 168115 to 82512 bytes)
  | remote: Enumerating objects: 308734, done.
  | remote: Counting objects: 100% (663/663), done.
  | remote: Compressing objects: 100% (364/364), done.
  | remote: Total 308734 (delta 355), reused 527 (delta 277), pack-reused 308071
  | Receiving objects: 100% (308734/308734), 424.15 MiB \| 30.80 MiB/s, done.
  | Resolving deltas: 100% (196120/196120), done.
  | BUG: refs/files-backend.c:2956: initial ref transaction called with existing refs
⚠️ Warning: Checkout failed! signal: aborted (core dumped) (Attempt 1/3 Retrying in 2s)
```

This leaves a corrupted directory in mirrors that causes problems subsequently:

```
$ cd /mirrors
--
  | # Updating existing repository mirror to find commit 41ad94161f0aa65ed06c25e2323670129490c268
  | $ git --git-dir /mirrors/org-49461806-github-com-squareup-redacted-remote set-url origin org-49461806@github.com:squareup/redacted.git

proceeds to check out the repo again, but doesn't update the mirror
```

Seems like we should be deleting the mirror dir if the clone fails?

lox:
Fixes #1670

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -1216,6 +1216,10 @@ func (b *Bootstrap) updateGitMirror() (string, error) {
 		b.shell.Commentf("Cloning a mirror of the repository to %q", mirrorDir)
 		flags := "--mirror " + b.GitCloneMirrorFlags
 		if err := gitClone(b.shell, flags, b.Repository, mirrorDir); err != nil {
+			b.shell.Commentf("Removing mirror dir %q due to failed clone", mirrorDir)
+			if err := os.RemoveAll(mirrorDir); err != nil {
+				b.shell.Errorf("Failed to remove \"%s\" (%s)", mirrorDir, err)
+			}
 			return "", err
 		}
 


bootstrap/bootstrap.go
```diff
@@ -1216,6 +1216,10 @@ func (b *Bootstrap) updateGitMirror() (string, error) {
 		b.shell.Commentf("Cloning a mirror of the repository to %q", mirrorDir)
 		flags := "--mirror " + b.GitCloneMirrorFlags
 		if err := gitClone(b.shell, flags, b.Repository, mirrorDir); err != nil {
+			b.shell.Commentf("Removing mirror dir %q due to failed clone", mirrorDir)
+			if err := os.RemoveAll(mirrorDir); err != nil {
```
moskyb, line 1220: i suspect this will cause issues with the lockfile we're creating above on [line 1208](https://github.com/buildkite/agent/pull/1671/files#diff-70e82e7432e3056b7466bb3e4fe16874c12cfbe7bbb71a15ad9d1997217e94c9R1208) - if we delete the lockfile before we're finished deleting everything else, something sharing the mirror dir might run into a very weird situation

bootstrap/bootstrap.go
```diff
@@ -1216,6 +1216,10 @@ func (b *Bootstrap) updateGitMirror() (string, error) {
...
+				b.shell.Errorf("Failed to remove \"%s\" (%s)", mirrorDir, err)
+			}
 			return "", err
```
moskyb, line 1223: Given that we're deleting the mirror dir, should we clone again by default? or is that already handled for us?

moskyb (changes requested):

bootstrap/bootstrap.go
lox: The lockfile is a peer of the mirror dir I think, not a child:

https://github.com/buildkite/agent/blob/1376fb737e8e33d83bd47d6404d8cc2cca458c45/bootstrap/bootstrap.go#L1208

So should be fine to return an error and run the deferred unlock, right?

bootstrap/bootstrap.go
lox: The checkout hook will automatically retry if it doesn't have a fatal error, I figure we can rely on this behaviour?

moskyb: Sounds good to me

bootstrap/bootstrap.go
moskyb: oh you're completely right, i hallucinated a `/` in there. as you were!

moskyb (approved):
Looks great! sorry about the wait on approving

PR #1675, opened 2022-06-14
Replace retry with roko

moskyb:
We recently extracted the retry module from the agent into its own go module, [roko](https://github.com/buildkite/roko).

This PR removes the existing retry library, and replaces it with roko. There's no functionality change between the two modules other than the module name.


clicommand/pipeline_upload.go
```diff
@@ -320,7 +320,7 @@ var PipelineUploadCommand = cli.Command{
 
 			return err
 			// On a server error, it means there is downtime or other problems, we
-			// need to retry. Let's retry every 5 seconds, for a total of 5 minutes.
+			// need to roko. Let's retry every 5 seconds, for a total of 5 minutes.
```
tessereth, line 323: Ah.... Maybe not this one, as good as it sounds 😅

tessereth (approved):

clicommand/pipeline_upload.go
moskyb: 🙈 omg

pda:
image

🙌🏼

sj26:
Nice 🙌

PR #1681, opened 2022-06-26
Add note about using `--step` with parallelized steps

nickofthyme:
Adds note about about using the `--step` option with parallelized steps to be added in https://github.com/buildkite/docs/pull/1607.

### DIFF
--- a/clicommand/artifact_download.go
+++ b/clicommand/artifact_download.go
@@ -76,7 +76,7 @@ var ArtifactDownloadCommand = cli.Command{
 		cli.StringFlag{
 			Name:  "step",
 			Value: "",
-			Usage: "Scope the search to a particular step by using either its name or job ID",
+			Usage: "Scope the search to a particular step by using either its name or job ID. See [section](/docs/agent/v3/cli-artifact#parallelized-steps) for parralelized steps.",
 		},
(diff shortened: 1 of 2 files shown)


moskyb (commented):
I'm really torn about this one. One one hand, i think it's a really good change for docs readability, and having docs links on hand in the CLI docs on buildkite.com/docs would be really handy (say that ten times fast!)

At the same time though, these are CLI docs for a CLI tool, and it's weird to have a linky-looking thing that doesn't go anywhere if you're wondering how artifact search works and you run `buildkite-agent artifact search --help` in your terminal.

I think on balance, i'm probably going to have to reject this PR, but with the caveat that i think that we (buildkite) need to have a bit of a think about how we generate the CLI docs on buildkite.com/docs - the two-step process is a bit clunky, and it would be great to be able to link out to other articles like you have here.

this is a great contribution, and i'm really sorry that we can't merge it :(

nickofthyme:
No that's fine. I think https://github.com/buildkite/docs/pull/1607 covers the important parts in the same place, I just thought it'd be good to point this nuance out in the `--step` option directly which led me here. But I'm fine with just merging the `/docs` PR.

PR #1686, opened 2022-06-28
Update proposed comment.

shepting:
Very minor nit.

I think it might make a little more sense for the key and value numbers to match.


moskyb (approved):
LGTM!

PR #1688, opened 2022-06-29
Add a shell message for flock-file-locks experiment

lox:
This adds a comment similar to the mirrors experiment when `flock-file-locks` is used. We're trying to troubleshoot lock failures, so this helps us be certain the experiment is enabled!


KevinGreen:
Hi @lox, thanks for adding this. Are you seeing the locking issue with this experiment turned on? I just opened https://github.com/buildkite/agent/pull/1685 which might be explained by the locking not working as expected in updateGitMirror but it's only happened twice since turning on the experiment in April and its typically worked well.

moskyb (approved):

PR #1692, opened 2022-07-01
Don't blow up when a hook exports an array

moskyb:
Previously, if a line in the output of `export -p` contained multiple options on the `declare` call (eg, there was an array in there - `declare -ax ARRAY=("red", "green", "blue")`), env.FromExport() would have some very weird behaviour - it could eat variables entirely, leading to them being unset in the build that the agent ran - more context for buildkite staff [here](https://3.basecamp.com/3453178/buckets/23112918/messages/5089354108#__recording_5089608441), but the TL;DR is that a hook was exporting an array variable, and it was causing other variables - important ones like `PATH` and `PWD` - to go missing.

**This PR:** Teaches env.FromExport to be able to introspect the args passed to each `declare` call, and allows it to ignore anything that it won't be able to parse. At the moment, it ignores indexed arrays (`declare -a`), associative arrays (`declare -A`), integers (`declare -i`) and links (`declare -n`)

### DIFF
--- a/env/export.go
+++ b/env/export.go
@@ -96,31 +96,31 @@ func FromExport(body string) Environment {
 			//     2. `FOO="open quote for new lines`
 			//     3. `FOO`
 			//
-			parts := strings.SplitN(line, "=\"", 2)
-			if len(parts) == 2 {
+			key, val, ok := strings.Cut(line, `="`)
+			if ok {
 				// If the value ends with an unescaped quote,
 				// then we know it's a single line environment
 				// variable (see example 1)
-				if endsWithUnescapedQuoteRegex.MatchString(parts[1]) {
+				if endsWithUnescapedQuoteRegex.MatchString(val) {
 					// Remove the `"` at the end
-					singleLineValueWithQuoteRemoved := strings.TrimSuffix(parts[1], `"`)
+					singleLineValueWithQuoteRemoved := strings.TrimSuffix(val, `"`)
 
 					// Set the single line env var
-					env.Set(parts[0], unquoteShell(singleLineValueWithQuoteRemoved))
+					env.Set(key, unquoteShell(singleLineValueWithQuoteRemoved))
 				} else {
 					// We're in an example 2 situation,
 					// where we need to keep keep the
 					// environment variable open until we
 					// encounter a line that ends with an
 					// unescaped quote
-					openKeyName = parts[0]
-					openKeyValue = []string{parts[1]}
+					openKeyName = key
+					openKeyValue = []string{val}
 				}
 			} else {
 				// Since there wasn't an `=` sign, we assume
 				// it's just an environment variable without a
 				// value (see example 3)
-				env.Set(parts[0], "")
+				env.Set(key, "")
 			}
 		}
 
--- a/env/export.go
+++ b/env/export.go
@@ -5,8 +5,22 @@ import (
 	"strings"
 )
 
-var posixExportLineRegex = regexp.MustCompile("\\Adeclare \\-x ([a-zA-Z_]+[a-zA-Z0-9_]*)(=\")?(.+)?\\z")
-var endsWithUnescapedQuoteRegex = regexp.MustCompile("([^\\\\]\"\\z|\\A\"\\z)")
+var (
+	posixExportLineRegex = regexp.MustCompile("\\Adeclare \\-[aAfFgiIlnrtux]+ ([a-zA-Z_]+[a-zA-Z0-9_]*)(=\")?(.+)?\\z")
+	//                                                             ^
+	//                                                             |
+	//                                           all of the available options for declare
+	endsWithUnescapedQuoteRegex = regexp.MustCompile("([^\\\\]\"\\z|\\A\"\\z)")
+
+	// There are a bunch of types of bash variable that we want to ignore, becuase supporting them
+	// is either a pain, or useless, or both.
+	disallowedDeclareOpts = map[rune]struct{}{
+		'a': {}, // declare -a - it's an array, and can't be used in the environment anyway
+		'A': {}, // declare -A - it's an associative array, and also can't be used in the environment
+		'n': {}, // declare -n - it's a reference to another variable, and will be a pain to sort out
+		'i': {}, // declare -i - it's an integer, and will break our parser. We theoretically could support this, but it's a pain
+	}
+)
 
 // FromExport parses environment variables from a shell export of environment variables. On
 // *nix it looks like this:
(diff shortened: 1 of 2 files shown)


env/export.go
```diff
@@ -87,40 +101,61 @@ func FromExport(body string) Environment {
...
+					// There should be at least three elements to the command; the format is:
+					// declare -x VARNAME="VALUE"
+					// If there's anything less than that, something is seriously wonky with this line and we won't be able to parse it anyway, so just skip it
+					continue
+				}
+
+				// Massive assumption here! We're assuming that that the options passed to declare have been coalesced -
+				// ie, that the command looks like `declare -axtn VARNAME="BLARGH"` rather than `declare -x -a -t -n VARNAME="BLARGH"`
+				// However, the output of `export -p` (which is what we're parsing here) is pretty consistent in coalescing the args
+				// God willing, this shouldn't change.
```
moskyb, line 116: he says, tempting fate

env/export.go
```diff
@@ -5,8 +5,22 @@ import (
...
+	//                                           all of the available options for declare
+	endsWithUnescapedQuoteRegex = regexp.MustCompile("([^\\\\]\"\\z|\\A\"\\z)")
+
+	// There are a bunch of types of bash variable that we want to ignore, becuase supporting them
+	// is either a pain, or useless, or both.
+	disallowedDeclareOpts = []string{
+		"a", // declare -a - it's an array, and can't be used in the environment anyway
+		"A", // declare -A - it's an associative array, and also can't be used in the environment
+		"n", // declare -n - it's a reference to another variable, and will be a pain to sort out
+		"i", // declare -i - it's an integer, and will break our parser. We theoretically could support this, but it's a pain
```
moskyb, line 21: The full set of options accepted by bash `declare` is:
```
      -a	to make NAMEs indexed arrays (if supported)
      -A	to make NAMEs associative arrays (if supported)
      -i	to make NAMEs have the `integer' attribute
      -l	to convert the value of each NAME to lower case on assignment
      -n	make NAME a reference to the variable named by its value
      -r	to make NAMEs readonly
      -t	to make NAMEs have the `trace' attribute
      -u	to convert the value of each NAME to upper case on assignment
      -x	to make NAMEs export
```
(from `help declare`)

So we're not disallowing everything other than `-x`, but everything that haven't disallowed shouldn't have too much impact on our parser, i think

env/export.go
```diff
@@ -5,8 +5,22 @@ import (
 	"strings"
 )
 
-var posixExportLineRegex = regexp.MustCompile("\\Adeclare \\-x ([a-zA-Z_]+[a-zA-Z0-9_]*)(=\")?(.+)?\\z")
-var endsWithUnescapedQuoteRegex = regexp.MustCompile("([^\\\\]\"\\z|\\A\"\\z)")
+var (
+	posixExportLineRegex = regexp.MustCompile("\\Adeclare \\-[aAfFgiIlnrtux]+ ([a-zA-Z_]+[a-zA-Z0-9_]*)(=\")?(.+)?\\z")
+	//                                                             ^
+	//                                                             |
+	//                                           all of the available options for declare
```
pda, line 12: Is it possible that being strict about matching a known set of options could break things in future if some version of bash or bash-emulating shell provides other (harmless) options?

env/export.go
pda: 👍🏼

env/export.go
```diff
@@ -87,40 +101,61 @@ func FromExport(body string) Environment {
 				continue
 			}
 
-			// Trim the `declare -x ` off the start of the line
-			line = strings.TrimPrefix(line, "declare -x ")
+			if strings.HasPrefix(line, "declare") {
+				command := strings.Split(line, " ")
```
pda, line 105: Could splitting on `" "` here and then doing `line = strings.Join(command[2:], " ")` later mangle env values with multiple contiguous spaces? Might we want to `strings.SplitN(line, " ", 3)` or similar so that we're only splitting the preamble/name, not the value?

env/export.go
```diff
@@ -87,40 +101,61 @@ func FromExport(body string) Environment {
...
 
 			// We now have a line that can either be one of these:
 			//
 			//     1. `FOO="bar"`
 			//     2. `FOO="open quote for new lines`
 			//     3. `FOO`
 			//
-			parts := strings.SplitN(line, "=\"", 2)
-			if len(parts) == 2 {
+			key, val, ok := strings.Cut(line, "=\"")
```
pda, line 134: Backtick string to avoid escaping? 🙃

pda (commented):
Looking good, a few early comments but I haven't finished reviewing yet.

env/export.go
pda: Heh, this crossed my mind too.
I think if we ran into that, it would be time to explore a non-`export -p`-based solution 😅

env/export.go
```diff
@@ -5,8 +5,22 @@ import (
...
+var (
+	posixExportLineRegex = regexp.MustCompile("\\Adeclare \\-[aAfFgiIlnrtux]+ ([a-zA-Z_]+[a-zA-Z0-9_]*)(=\")?(.+)?\\z")
+	//                                                             ^
+	//                                                             |
+	//                                           all of the available options for declare
+	endsWithUnescapedQuoteRegex = regexp.MustCompile("([^\\\\]\"\\z|\\A\"\\z)")
+
+	// There are a bunch of types of bash variable that we want to ignore, becuase supporting them
+	// is either a pain, or useless, or both.
+	disallowedDeclareOpts = []string{
```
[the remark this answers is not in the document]
moskyb: my general policy is to avoid using regex wherever possible - they're really tricky to debug, and require a different skillset than ordinary programming to parse, IMO.

what i might do though is turn this into a map to make the process of searching for disallowed opts slightly faster and easier to read

env/export.go
moskyb: ooooooooh that's a really good callout, i never would've thought of that! will change it up.

pda (approved):
Excellent!

sj26:
So good!

env/export.go
moskyb: just coming back to this, because i'm now realising that i forgot to, post-merge: i think it's _probably_ fine - it's significantly more permissive now than it used to be - but IMO in the medium-term i think we need to move away from parsing the output of `export -p`, and move to parsing the output of `printenv` or something a bit less complex. I think if something comes along to further break this workflow, that should be the kick in the pants we need to properly refactor this.

A low-hanging fruit version of this might just be refactoring the logic of this parser from what we currently have:
```Go
firstLine := lines[0]
if looksLikeExportDashPOutput(firstLine) {
	for line := range lines {
		// Parse the entire file like it's export -p
	} 
} else {
  // It's _definitely_ "THIS=that" style output, parse it like that
}
```
to something like 
```Go
for line := range lines {
	k, v := ParseExportLine(line)
	env.Set(k, v)
}

func ParseExportLine(line string) (string, string, error) {
	switch {
		case LooksLikeExportDashP(line):
			return ParseExportDashP(line)
		case LooksLikeEnv(line):
			return ParseEnv(line)
		default:
			fmt.Println("Failed to parse line:", line)
			return "", "", fmt.Errorf("Failed to parse line: %v", line)
	}
	return parts[0], parts[1]
}
```

that is, currently, anything that fails to get parsed using our magic export -p regex is assumed to be env-style (eg `THIS=that`), so any failure cases fall through the cracks. We should probably update this parser to be aware of error cases in general, which it's currently not.

this is all a problem for future benno though - at the moment, i think what we have is acceptable, but i agree that it could be better

PR #1694, opened 2022-07-03
Use non-conflicting docker networks

moskyb:
When running multiple jobs on the same machine, they use the same docker network by default, which means that the two running jobs can run into a race condition where they both create a docker network with the same name, leading to a build failure (see https://buildkite.com/buildkite/agent/builds/4420#0181ba48-c82c-4b92-bac9-5bff1743a373).

~~This PR adjusts the build script to add a random number to the end of the docker network name, meaning that we have a 32767x lower chance of a network name collision~~

This PR uses the COMPOSE_PROJECT_NAME environment variable set to the job ID of the build we're running to avoid network name collisions

### DIFF
--- a/.buildkite/docker-compose.yml
+++ b/.buildkite/docker-compose.yml
@@ -28,3 +28,7 @@ services:
 
 volumes:
   ruby-2-6-agent-gem-cache: ~
+
+networks:
+  default:
+    name: buildkite-agent-${RAND}
--- a/.buildkite/docker-compose.yml
+++ b/.buildkite/docker-compose.yml
@@ -28,7 +28,3 @@ services:
 
 volumes:
   ruby-2-6-agent-gem-cache: ~
-
-networks:
-  default:
-    name: buildkite-agent-${RAND}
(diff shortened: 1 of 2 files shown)


tessereth (approved):

tessereth (approved):
I approve this idea in principle, once you fix whatever bug is breaking the build. Feel free to merge if there's no major changes or bump it back to me if you have major changes and want a second pair of eyes.

sj26:
Is this something we should be addressing in the docker-compose plugin, too?

sj26:
I was thinking about this a little more. I'm pretty sure docker-compose will use the project name as the name of the default network, and we go to lengths in the docker-compose plugin to use a distinct project name for each docker compose stack:

https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/blob/970b15d9036d5a14199c9cce69dd0c4c8da4c69e/lib/shared.bash#L81-L85

https://docs.docker.com/compose/networking/

image

moskyb:
oooh this is a good callout, thanks @sj26!

moskyb:
i might just steal what we do in the docker-compose plugin and do that in our `ruby-env` script

sj26 (approved):
Nice.

PR #1698, opened 2022-07-07
Use more words, less emoji in pipeline step names

moskyb:
Pop quiz! What do you think this CI step does, based on its name?
image
If you answered "Run the tests in a linux ARM64 environment with the race detector enabled", you're correct! You win the prize of knowing deep inside that you're a very intelligent person.

In all seriousness, the names of our CI steps are a bit crap at the moment, the high amount of emoji and the low amount of text means that it all turns into heiroglyphics, a little bit.

This PR removes some of the emoji (leaving a reasonable number in, for flavour), and gives each step a relatively descriptive name in english, as well.

Suggestions more than welcome on better step names!

### DIFF
--- a/.buildkite/pipeline.release-experimental.yml
+++ b/.buildkite/pipeline.release-experimental.yml
@@ -1,7 +1,7 @@
 steps:
   - wait
 
-  - name: ":s3: publish binaries"
+  - name: ":s3: Publish Edge Binaries"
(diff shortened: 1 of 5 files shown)


tessereth (approved):
Looks nice ✨

pda (approved):
Thanks!

PR #1699, opened 2022-07-07
OpenTelemetry Polishing

moskyb:
**This PR:** Takes a couple of the things we've learned from tracing our own agent builds and implements them in the opentelemetry-tracing experiment. We've:
- Added a Deployment Environment attribute, `ci`.
- Renamed the root span for each build from `job.run` (which doesn't say a whole lot about what's happening!) to something dynamic per job, in the format `<buildkite org>/<pipeline slug>/<job key or label>`. For example, this might end up as `buildkite/agent/go-fmt` or `rails/rails/rspec`.
- Added job label and key attributes to all spans generated by the agent, making them easier to group

### DIFF
--- a/bootstrap/tracing.go
+++ b/bootstrap/tracing.go
@@ -220,6 +220,16 @@ func GenericTracingExtras(b *Bootstrap, env env.Environment) map[string]any {
 		triggeredFromID = "n/a"
 	}
 
+	jobLabel, has := env.Get("BUILDKITE_LABEL")
+	if !has || jobLabel == "" {
+		jobLabel = "n/a"
+	}
+
+	jobKey, has := env.Get("BUILDKITE_STEP_KEY")
+	if !has || jobKey == "" {
+		jobKey = "n/a"
+	}
+
 	return map[string]any{
 		"buildkite.agent":             b.AgentName,
 		"buildkite.version":           agent.Version(),
@@ -227,6 +237,8 @@ func GenericTracingExtras(b *Bootstrap, env env.Environment) map[string]any {
 		"buildkite.org":               b.OrganizationSlug,
 		"buildkite.pipeline":          b.PipelineSlug,
 		"buildkite.branch":            b.Branch,
+		"buildkite.job_label":         jobLabel,
+		"buildkite.job_key":           jobKey,
 		"buildkite.job_id":            b.JobID,
 		"buildkite.job_url":           jobURL,
 		"buildkite.build_id":          buildID,


sj26:
Nice!

> Added a Deployment Environment attribute, ci.

Tell me more about this — what does it mean? Would anybody want it to be _not_ `ci`?

> Renamed the root span

Tell me more about this — is the name meant to be a descriptive label? Or a "type" of span? I thought those things would be (and seem to be) attributes of a span, to be faceted on?

> Added job label and key attributes to all spans generated by the agent, making them easier to group

I notice we're backfilling "n/a" on some of these. Would it be better to omit them when not present?

pda:
> Renamed the root span for each build from `job.run` … to … `<buildkite org>/<pipeline slug>/<job key or label>`. For example, this might end up as `buildkite/agent/go-fmt` or `rails/rails/rspec`.

Does it matter if job label has spaces, emoji, non-uniqueness etc?

moskyb:
> Tell me more about [adding a deployment environment key] — what does it mean? Would anybody want it to be not ci?

i don't think so? the `deployment.environment` key that we add is a predefined key in Otel that's meant to tell the trace consumer (Datadog in our case) what the execution environment is for the traces being received. Usually it'd be `production` or `staging` or whatever, but neither of those really make sense for the buildkite agent. In this case i'm happy to make some assumptions about what our users might want or need, and if we turn out to be wrong, we can adjust this later.

The reasoning for adding this is that in Datadog, the traces show up scoped by environment:
image
and having them in `env:none` feels pretty clunky. Datadog should naturally convert `deployment.environment` into the datadog-native `env` tags, as implied (but not outright stated 🙃) [here](https://docs.datadoghq.com/tracing/setup_overview/open_standards/otel_collector_datadog_exporter/).

> Tell me more about [renaming the root span] — is the name meant to be a descriptive label?

Yes. Again in datadog, this is what most of our traces look like:
CleanShot 2022-07-08 at 13 48 16@2x
where the column with all the `SELECT features . value FROM ...` is the span name. AFAICT, the name of the span should tell you what a span is doing, and while `job.run` doesn't _not_ do that - it is, in fact, a job being run - it's not very useful at a glance to see:
image
These are very clearly different jobs - they have very different run times, and when we dig into them, it's difficult to tell which is which, or how/that they are different "resources".

Additionally, from a datadog-specific perspective, the resource name (which comes from the span name) is the only thing that gets indexed past the 15 minute live-query interval by default, so if i wanted to track the total runtime of a specific step over time, this is the easiest way to do it (we can also add query facets for span attributes, but this is extra config effort.

IMO changing up the root span name brings us in line with other tracing integrations.

> I notice we're backfilling "n/a" on some of these. Would it be better to omit them when not present?

Probably? But there was already an established standard for `n/a` so i just followed that. Happy to go with omission if you're keen.

moskyb:
@pda

> Does it matter if job label has spaces, emoji, non-uniqueness etc?

not AFAICT. as shown above, SQL spans have the query as their span name, which both have spaces and are pretty non-unique. Re: emoji, they seem to be fine:
CleanShot 2022-07-08 at 14 33 56@2x

sj26:
I can see somebody wanting to split deployment environment if they were using agents for production deploys or something. But let's add config if and when that happens.

sj26 (approved):
Lovely ✨

PR #1702, opened 2022-07-11
Promote opentelemetry tracing to mainline feature status

moskyb:
🎉

**This PR:** Promotes opentelemetry tracing from being an experimental feature to being... a feature. It's here! We'll continue to iterate on this over time, but we're happy enough with the implementation as it stands to make it a mainline feature.


tessereth (approved):
🥳

PR #1703, opened 2022-07-12
Better race testing

moskyb:
**This PR:** 
- Makes the test step with the race detector enabled hard-fail
- Skips the current failing race test - the logic here is that we don't want to add any further data races to the code. The only one that currently gets caught by the race detector is a bug in a library we use for testing, and we don't want new data races to get missed because this step is currently soft-failing

### DIFF
--- a/race/disabled.go
+++ b/race/disabled.go
@@ -1,0 +1,9 @@
+//go:build !race
+
+package race
+
+// We can use this constant to determine whether or not the Go Race Detector is enabled. If the race detector is enabled,
+// the `race` build tag is passed, so this file won't be compiled, where the other one (./enabled.go) will be.
+// That is, if the race detector is enabled, the constant below will be false. Otherwise, it'll be true.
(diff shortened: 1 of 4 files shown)


race/disabled.go
```diff
@@ -0,0 +1,9 @@
+//go:build !race
+
+package race
+
+// We can use this constant to determine whether or not the Go Race Detector is enabled. If the race detector is enabled,
+// the `race` build tag is passed, so this file won't be compiled, where the other one (./enabled.go) will be.
+// That is, if the race detector is enabled, the constant below will be false. Otherwise, it'll be true.
+// It's kinda weird, but it's also how go itself does it: https://github.com/golang/go/blob/master/src/internal/race/race.go#L15
```
tessereth, line 8: Huh, that's an interesting way to do it... But it's not dumb if it works.

tessereth (approved):

race/disabled.go
moskyb: i feel like "prior art from the creators of the language" is about as carte blanche as it gets to do janky shit, and yet... it still feels incredibly janky.

PR #1705, opened 2022-07-14
Otel polishing 2: Electric Boogaloo

moskyb:
**This PR:** Makes a bunch more changes to the OpenTelemetry implementation (a few of which get backported to the datadog tracing implementation too!). Technically, we [de-experimented opentelemetry](https://github.com/buildkite/agent/pull/1702) a couple of days ago, so _technically_ we shouldn't make any breaking changes, but the de-experimentation hasn't been shipped yet so 🤷

These changes mostly only affect the opentel integration; those that also affect the datadog implementation will be noted in the diff. There shouldn't be any backwards incompatible changes.

The changes we've made are:
- Rename the tracing service name from `buildkite_agent` to `buildkite-agent`
- Trace each part of the artifact upload phase individually - previously we had one chonky span for the whole step, now we have individual spans for pre-upload, upload, and post-upload
before:
CleanShot 2022-07-14 at 15 16 33@2x

after:
CleanShot 2022-07-14 at 15 15 50@2x

- Rename hook spans from `hook.execute` for every hook to `hook.$SCOPE.$NAME`
before: 
image

after: 
image

- For hooks where the scope is `plugin`, add plugin attributes to their hook span:
CleanShot 2022-07-14 at 15 27 59@2x

- For all "phases" of the bootstrap (environment, checkout, command, artifact), either create a phase-level span or rename the existing phase-level span to be called `phase/$phase_name`. This has the effect of grouping spans in the phase together - ie `pre-command`, `command` and `post-command` all get grouped under one parent span. This is what it looked like before:
CleanShot 2022-07-14 at 15 34 35@2x

and after: 
CleanShot 2022-07-14 at 15 37 49@2x

The phase thing is a little bit confusing, and i'm happy to revert that to the existing behaviour - the change makes trace easier to parse IMO, but the change itself is a bit confusing.

**This PR will be most easily reviewed commit-by-commit**

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -1010,7 +1010,7 @@ func (b *Bootstrap) CheckoutPhase(ctx context.Context) error {
 				roko.WithMaxAttempts(3),
 				roko.WithStrategy(roko.Constant(2*time.Second)),
 			).Do(func(r *roko.Retrier) error {
-				err := b.defaultCheckoutPhase()
+				err := b.defaultCheckoutPhase(ctx)
 				if err == nil {
 					return nil
 				}
@@ -1214,17 +1214,26 @@ func (b *Bootstrap) updateGitMirror() (string, error) {
 
 // defaultCheckoutPhase is called by the CheckoutPhase if no global or plugin checkout
 // hook exists. It performs the default checkout on the Repository provided in the config
-func (b *Bootstrap) defaultCheckoutPhase() error {
+func (b *Bootstrap) defaultCheckoutPhase(ctx context.Context) error {
+	span, _ := tracetools.StartSpanFromContext(ctx, "repo-checkout", b.Config.TracingBackend)
+	span.AddAttributes(map[string]string{
+		"checkout.repo_name": b.Repository,
+		"checkout.refspec":   b.RefSpec,
+		"checkout.commit":    b.Commit,
+	})
+	var err error
+	defer func() { span.FinishWithError(err) }()
+
 	if b.SSHKeyscan {
 		addRepositoryHostToSSHKnownHosts(b.shell, b.Repository)
 	}
 
-	var err error
 	var mirrorDir string
 
 	// If we can, get a mirror of the git repository to use for reference later
 	if experiments.IsEnabled(`git-mirrors`) && b.Config.GitMirrorsPath != "" && b.Config.Repository != "" {
 		b.shell.Commentf("Using git-mirrors experiment 🧪")
+		span.AddAttributes(map[string]string{"checkout.is_using_git_mirrors": "true"})
 
 		// Skip updating the Git mirror before using it?
 		if b.Config.GitMirrorsSkipUpdate {
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -1446,7 +1446,8 @@ func (b *Bootstrap) resolveCommit() {
 
 // runPreCommandHooks runs the pre-command hooks and adds tracing spans.
 func (b *Bootstrap) runPreCommandHooks(ctx context.Context) error {
-	span, ctx := tracetools.StartSpanFromContext(ctx, "pre-command hooks", b.Config.TracingBackend)
+	spanName := b.implementationSpecificSpanName("pre-command", "pre-command hooks")
+	span, ctx := tracetools.StartSpanFromContext(ctx, spanName, b.Config.TracingBackend)
 	var err error
 	defer func() { span.FinishWithError(err) }()
 
@@ -1484,7 +1485,8 @@ func (b *Bootstrap) runCommand(ctx context.Context) error {
 
 // runPostCommandHooks runs the post-command hooks and adds tracing spans.
 func (b *Bootstrap) runPostCommandHooks(ctx context.Context) error {
-	span, ctx := tracetools.StartSpanFromContext(ctx, "post-command hooks", b.Config.TracingBackend)
+	spanName := b.implementationSpecificSpanName("post-command", "post-command hooks")
+	span, ctx := tracetools.StartSpanFromContext(ctx, spanName, b.Config.TracingBackend)
 	var err error
 	defer func() { span.FinishWithError(err) }()
 
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -1805,7 +1805,7 @@ func (b *Bootstrap) artifactPhase(ctx context.Context) error {
 		return nil
 	}
 
-	spanName := b.implementationSpecificSpanName("artifact", "artifact upload")
+	spanName := b.implementationSpecificSpanName("artifacts", "artifact upload")
 	span, ctx := tracetools.StartSpanFromContext(ctx, spanName, b.Config.TracingBackend)
 	var err error
 	defer func() { span.FinishWithError(err) }()
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -472,7 +472,8 @@ func addRepositoryHostToSSHKnownHosts(sh *shell.Shell, repository string) {
 // setUp is run before all the phases run. It's responsible for initializing the
 // bootstrap environment
 func (b *Bootstrap) setUp(ctx context.Context) error {
-	span, ctx := tracetools.StartSpanFromContext(ctx, "environment", b.Config.TracingBackend)
+	spanName := b.implementationSpecificSpanName("phase/environment", "environment")
+	span, ctx := tracetools.StartSpanFromContext(ctx, spanName, b.Config.TracingBackend)
 	var err error
 	defer func() { span.FinishWithError(err) }()
 
@@ -952,7 +953,8 @@ func (b *Bootstrap) createCheckoutDir() error {
 // CheckoutPhase creates the build directory and makes sure we're running the
 // build at the right commit.
 func (b *Bootstrap) CheckoutPhase(ctx context.Context) error {
-	span, ctx := tracetools.StartSpanFromContext(ctx, "checkout", b.Config.TracingBackend)
+	spanName := b.implementationSpecificSpanName("phase/checkout", "checkout")
+	span, ctx := tracetools.StartSpanFromContext(ctx, spanName, b.Config.TracingBackend)
 	var err error
 	defer func() { span.FinishWithError(err) }()
 
@@ -1504,6 +1506,9 @@ func (b *Bootstrap) runPostCommandHooks(ctx context.Context) error {
 
 // CommandPhase determines how to run the build, and then runs it
 func (b *Bootstrap) CommandPhase(ctx context.Context) (error, error) {
+	span, ctx := tracetools.StartSpanFromContext(ctx, "phase/command", b.Config.TracingBackend)
+	var err error
+	defer func() { span.FinishWithError(err) }()
 	// Run pre-command hooks
 	if err := b.runPreCommandHooks(ctx); err != nil {
 		return err, nil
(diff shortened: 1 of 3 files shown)


bootstrap/bootstrap.go
```diff
@@ -1737,16 +1771,40 @@ func (b *Bootstrap) writeBatchScript(cmd string) (string, error) {
...
+	}
+
+	return nil
+}
+
+// Run the pre-artifact hooks
+func (b *Bootstrap) preArtifactHooks(ctx context.Context) error {
+	span, ctx := tracetools.StartSpanFromContext(ctx, "pre-artifact", b.Config.TracingBackend)
 	var err error
 	defer func() { span.FinishWithError(err) }()
```
moskyb, line 1806: The new spans for pre- and post-artifact hooks only add more information, and don't remove or change any existing information, so should be backwards compatible in datadog tracing.

sj26 (commented):
I might let other folks go deep on this one, but on the whole it looks good to me!

bootstrap/bootstrap.go
```diff
@@ -384,7 +390,11 @@ func (b *Bootstrap) executeGlobalHook(ctx context.Context, name string) error {
 	if err != nil {
 		return err
 	}
-	return b.executeHook(ctx, "global", name, p, nil)
+	return b.executeHook(ctx, HookConfig{
+		Scope: "global",
+		Name:  name,
+		Path:  p,
+	})
```
pda, line 399: This is nicer than trying to understand what `"global", name, p, nil` means :)

bootstrap/bootstrap.go
```diff
@@ -725,18 +739,20 @@ func (b *Bootstrap) VendoredPluginPhase(ctx context.Context) error {
 func (b *Bootstrap) executePluginHook(ctx context.Context, name string, checkouts []*pluginCheckout) error {
 	for _, p := range checkouts {
 		hookPath, err := hook.Find(p.HooksDir, name)
-		// os.IsNotExist() doesn't unwrap wrapped errors (as at Go 1.13).
-		// agent is still go pre-1.13 compatible (I think) so we're avoiding errors.Is().
-		// In future somebody should check if os.IsNotExist() has added support for
-		// error unwrapping, or change this code to errors.Is(err, os.ErrNotExist)
-		if os.IsNotExist(err) {
+		if errors.Is(err, os.ErrNotExist) {
```
pda, line 745: somebody should and somebody did, thanks!

bootstrap/bootstrap.go
```diff
@@ -227,6 +228,7 @@ func (b *Bootstrap) executeHook(ctx context.Context, hookCfg HookConfig) error {
 		"hook.name":    hookCfg.Name,
 		"hook.command": hookCfg.Path,
 	})
+	span.AddAttributes(hookCfg.SpanAttributes)
```
pda, line 231: Does this handle `hookCfg.SpanAttributes` being `nil`?

bootstrap/bootstrap.go
```diff
@@ -472,7 +472,8 @@ func addRepositoryHostToSSHKnownHosts(sh *shell.Shell, repository string) {
 // setUp is run before all the phases run. It's responsible for initializing the
 // bootstrap environment
 func (b *Bootstrap) setUp(ctx context.Context) error {
-	span, ctx := tracetools.StartSpanFromContext(ctx, "environment", b.Config.TracingBackend)
+	spanName := b.implementationSpecificSpanName("phase/environment", "environment")
```
pda, line 475: Would `.` rather than `/` match other names better?
e.g. `phase.environment` instead of `phase/environment`.

bootstrap/tracing.go
```diff
@@ -73,7 +73,7 @@ func (b *Bootstrap) ddResourceName() string {
 // abstraction so the agent can support multiple libraries if needbe.
 func (b *Bootstrap) startTracingDatadog(ctx context.Context) (tracetools.Span, context.Context, stopper) {
 	opts := []tracer.StartOption{
-		tracer.WithServiceName("buildkite_agent"),
+		tracer.WithServiceName("buildkite-agent"),
```
pda, line 76: 👍🏼

Will this affect existing OpenTracing/Datadog usage?

pda (approved):
Looking good!

bootstrap/bootstrap.go
moskyb: Yep - both implementations of this method iterate over the map, which just does nothing on a nil map. I had to double check this, i hate nil maps so much.

bootstrap/tracing.go
moskyb: nah, this only affects opentel. datadog initialision happens in the `startTracingDatadog` method above, which keeps the underscored name. the two methods are mutually exclusive, only one of them can (should?) be called per bootstrap run.

bootstrap/bootstrap.go
moskyb: i had this thought as well, but in the opposite direction - the root span is `$org/$pipeline/$step_key`, and hook span names are `hook.$scope.$name`, so maybe we should rename the hook spans to `hook/$scope/$name`, and have everything be slash-delimited. 

imo slashes (`phase/environment`) better convey "this is the phase and the phase is called environment", but that could be my rails brain coming through, and that logic breaks down in the hook span case - `hook/plugin/command` doesn't really convey "this is a hook and its scope is plugin and its scope is command", so it's all a bit fuzzy really.

i'm happy to defer to your judgement here - i don't really have a strong opinion either way.

moskyb: another concern i have in this vague realm is the usability/understandability of the word "phase" - it's used in the agent internals, but i'm not sure it's a piece of jargon that we've actually defined, and i don't think it's in the docs. optimistically, i think that it's relatively easy to grok - it's a chunk of the agent's execution cycle for a job - but my judgement is probably clouded here by my familiarity with the agent.

moskyb: i agree with all of these - i think sticking to a more semantic name is good for consistency, i'll update all of these.

the one that might be a little tricky is calling the phase span eg just `command` for the command phase, because then you run into the issue of what do you call the span covering actually running the command, which is a child of the `command` span and a sibling of `pre-command` and `post-command`. I think we probably have a bit of an out here in that we could probably just get a bit more specific with this span name - eg `command execution` for command `repo checkout` for checkout, `artifact upload` for artifact, so i think it should be fine.

I'll implement this tomorrow morning :)

pda: Ah I hadn't found that naming advice, it's good 👍🏼

bootstrap/bootstrap.go
```diff
@@ -1785,7 +1784,7 @@ func (b *Bootstrap) artifactPhase(ctx context.Context) error {
 		return nil
 	}
 
-	spanName := b.implementationSpecificSpanName("phase/artifact", "artifact upload")
+	spanName := b.implementationSpecificSpanName("artifact", "artifact upload")
```
pda, line 1787: maybe `artifacts` plural?

bootstrap/bootstrap.go
```diff
@@ -1476,7 +1474,8 @@ func (b *Bootstrap) runPreCommandHooks(ctx context.Context) error {
 
 // runCommand runs the command and adds tracing spans.
 func (b *Bootstrap) runCommand(ctx context.Context) error {
-	span, ctx := tracetools.StartSpanFromContext(ctx, "command", b.Config.TracingBackend)
+	spanName := b.implementationSpecificSpanName("command-exec", "command")
+	span, ctx := tracetools.StartSpanFromContext(ctx, spanName, b.Config.TracingBackend)
```
pda, line 1478: Hrm, is this span even necessary?

In the switch statement below, it'll either create a `plugin command hook` or a `repository command hook` or an `agent command hook`, _or_ it will run the `defaultCommandPhase()`. AFAIK only that final case (which is the 99% default) is lacking a span.

I wonder if we should replace this `command-exec` span with a `default command hook` span just for that final/default case?

bootstrap/bootstrap.go
```diff
@@ -209,47 +209,57 @@ func (b *Bootstrap) Cancel() error {
...
+	Path           string
+	Env            env.Environment
+	SpanAttributes map[string]string
+}
+
 // executeHook runs a hook script with the hookRunner
-func (b *Bootstrap) executeHook(ctx context.Context, scope string, name string, hookPath string, extraEnviron env.Environment) error {
-	span, ctx := tracetools.StartSpanFromContext(ctx, "hook.execute", b.Config.TracingBackend)
+func (b *Bootstrap) executeHook(ctx context.Context, hookCfg HookConfig) error {
+	spanName := b.implementationSpecificSpanName(fmt.Sprintf("hook.%s.%s", hookCfg.Scope, hookCfg.Name), "hook.execute")
```
pda, line 222: Oh, I think we might need to map legacy/internal `hookCfg.Scope` names to [publicly documented hook location names](https://buildkite.com/docs/agent/v3/hooks#hook-locations).

- `local` → `repository`
- `global` → `agent`

bootstrap/bootstrap.go
moskyb: yep, sounds good!

bootstrap/bootstrap.go
pda: (and we're going to change the format to e.g. `agent environment hook`, right?)

bootstrap/bootstrap.go
moskyb: oh wait, actually, we don't even need to do that, there's already a span for the default command hook. I can't remember why i added this, ahaha

bootstrap/bootstrap.go
moskyb: yep yep! all done now

bootstrap/bootstrap.go
```diff
@@ -209,47 +209,75 @@ func (b *Bootstrap) Cancel() error {
...
+	SpanAttributes map[string]string
+}
+
+func (b *Bootstrap) tracingImplementationSpecificHookScope(scope string) string {
+	if b.TracingBackend != tracetools.BackendOpenTelemetry {
+		return scope
+	}
+
+	// The scope names local and global are confusing, and different to what we document
+	// however, we nee to
```
pda, line 226: … interrupted train of thought? 😁

bootstrap/bootstrap.go
```diff
@@ -209,47 +209,75 @@ func (b *Bootstrap) Cancel() error {
...
 
+type HookConfig struct {
+	Name           string
+	Scope          string
+	Path           string
+	Env            env.Environment
+	SpanAttributes map[string]string
+}
+
+func (b *Bootstrap) tracingImplementationSpecificHookScope(scope string) string {
```
pda, line 220: There's probably a backlog TODO in here to rename the hook scopes (or “locations”?) throughout the codebase so that they match the documented names.

pda (approved):

bootstrap/bootstrap.go
moskyb: 🙈

bootstrap/bootstrap.go
moskyb: 🤔 i wonder if we should just do it now (or soon, in another, less-dense PR). I don't thiiiiink it would break anything?

bootstrap/bootstrap.go
moskyb: interrupted _and_ misspelled

bootstrap/bootstrap.go
pda: Yeah I think “now”, but probably the “sleevie” version of now 😉
Definitely separate pull request, and probably easier to get this one merged first.

pda (approved):
🚀

PR #1706, opened 2022-07-15
Include a list of enabled features in the register request

moskyb:
**This PR:** Adds a smallish list of feature enabled to the agent register request, allowing us at buildkite a bit more visibility into what features of the agent our users actually used. This telemetry is on by default, but doesn't collect any sensitive info, and can be disabled using the `--no-telemetry` flag.

### DIFF
--- a/api/agents.go
+++ b/api/agents.go
@@ -2,29 +2,19 @@ package api
 
 // AgentRegisterRequest is a call to register on the Buildkite Agent API
 type AgentRegisterRequest struct {
-	Name               string        `json:"name"`
-	Hostname           string        `json:"hostname"`
-	OS                 string        `json:"os"`
-	Arch               string        `json:"arch"`
-	ScriptEvalEnabled  bool          `json:"script_eval_enabled"`
-	IgnoreInDispatches bool          `json:"ignore_in_dispatches"`
-	Priority           string        `json:"priority,omitempty"`
-	Version            string        `json:"version"`
-	Build              string        `json:"build"`
-	Tags               []string      `json:"meta_data"`
-	PID                int           `json:"pid,omitempty"`
-	MachineID          string        `json:"machine_id,omitempty"`
-	FeatureUsage       *FeatureUsage `json:"feature_usage,omitempty"`
-}
-
-type FeatureUsage struct {
-	TracingBackend             string   `json:"tracing_backend"`
-	Experiments                []string `json:"experiments"`
-	DisconnectAfterJobEnabled  bool     `json:"disconnect_after_job_enabled"`
-	DisconnectAfterIdleEnabled bool     `json:"disconnect_after_idle_timeout_enabled"`
-	Shell                      string   `json:"shell"`
-	PluginsEnabled             bool     `json:"plugins_enabled"`
-	AcquireJobEnabled          bool     `json:"acquire_job_enabled"`
+	Name               string   `json:"name"`
+	Hostname           string   `json:"hostname"`
+	OS                 string   `json:"os"`
+	Arch               string   `json:"arch"`
+	ScriptEvalEnabled  bool     `json:"script_eval_enabled"`
+	IgnoreInDispatches bool     `json:"ignore_in_dispatches"`
+	Priority           string   `json:"priority,omitempty"`
+	Version            string   `json:"version"`
+	Build              string   `json:"build"`
+	Tags               []string `json:"meta_data"`
+	PID                int      `json:"pid,omitempty"`
+	MachineID          string   `json:"machine_id,omitempty"`
+	Features           []string `json:"features"`
 }
 
 // AgentRegisterResponse is the response from the Buildkite Agent API
--- a/api/agents.go
+++ b/api/agents.go
@@ -14,7 +14,7 @@ type AgentRegisterRequest struct {
 	Tags               []string      `json:"meta_data"`
 	PID                int           `json:"pid,omitempty"`
 	MachineID          string        `json:"machine_id,omitempty"`
-	FeatureUsage       *FeatureUsage `json:"feature_usage"`
+	FeatureUsage       *FeatureUsage `json:"feature_usage,omitempty"`
 }
 
 type FeatureUsage struct {
--- a/api/agents.go
+++ b/api/agents.go
@@ -2,18 +2,29 @@ package api
 
 // AgentRegisterRequest is a call to register on the Buildkite Agent API
 type AgentRegisterRequest struct {
-	Name               string   `json:"name"`
-	Hostname           string   `json:"hostname"`
-	OS                 string   `json:"os"`
-	Arch               string   `json:"arch"`
-	ScriptEvalEnabled  bool     `json:"script_eval_enabled"`
-	IgnoreInDispatches bool     `json:"ignore_in_dispatches"`
-	Priority           string   `json:"priority,omitempty"`
-	Version            string   `json:"version"`
-	Build              string   `json:"build"`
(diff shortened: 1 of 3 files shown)


api/agents.go
```diff
@@ -2,18 +2,29 @@ package api
...
+}
+
+type FeatureUsage struct {
+	TracingBackend             string   `json:"tracing_backend"`
+	Experiments                []string `json:"experiments"`
+	DisconnectAfterJobEnabled  bool     `json:"disconnect_after_job_enabled"`
+	DisconnectAfterIdleEnabled bool     `json:"disconnect_after_idle_timeout_enabled"`
+	Shell                      string   `json:"shell"`
+	PluginsEnabled             bool     `json:"plugins_enabled"`
+	AcquireModeEnabled         bool     `json:"acquire_mode_enabled"`
```
pda, line 27: Maybe this?

```suggestion
	AcquireJobEnabled         bool     `json:"acquire_job_enabled"`
```

pda (approved):
Excellent 👌🏼

I guess we'll want to add it to some docs / README / something too, before it lands in a release.

api/agents.go
```diff
@@ -2,18 +2,29 @@ package api
...
+	Arch               string        `json:"arch"`
+	ScriptEvalEnabled  bool          `json:"script_eval_enabled"`
+	IgnoreInDispatches bool          `json:"ignore_in_dispatches"`
+	Priority           string        `json:"priority,omitempty"`
+	Version            string        `json:"version"`
+	Build              string        `json:"build"`
+	Tags               []string      `json:"meta_data"`
+	PID                int           `json:"pid,omitempty"`
+	MachineID          string        `json:"machine_id,omitempty"`
+	FeatureUsage       *FeatureUsage `json:"feature_usage"`
```
sj26, line 17: Could we call this "telemetry" instead? Feels like a more general concept, and matches the arg name.

api/agents.go
```diff
@@ -2,18 +2,29 @@ package api
...
+}
+
+type FeatureUsage struct {
+	TracingBackend             string   `json:"tracing_backend"`
```
sj26, line 21: Should this be omitted when empty?

```suggestion
	TracingBackend             string   `json:"tracing_backend,omitempty"`
```

api/agents.go
```diff
@@ -2,18 +2,29 @@ package api
...
+	Experiments                []string `json:"experiments"`
```
sj26, line 22: Nice

api/agents.go
moskyb: ah yes! it should, it's been so long since i serialised a struct, i completely forgot.

api/agents.go
moskyb: eh... i think telemetry covers a larger concept that what we have here - ie, if we were sending back job timings to buildkite (for buildkite to see and not the client), that would also be another type of telemetry, that i think we would want to disable that using the same flag that disables this. Does that make sense? I feel like i haven't explained it that well ahah

sj26: Yeah, true. I wish we didn't already use the term "meta_data" 😬 

Maybe feature_usage works. It just feels odd in the backend.

moskyb: maybe `agent_feature_usage` is a little bit less odd?

api/agents.go
moskyb: actually, thinking on this, i think it makes sense to leave that as `""` if there's no tracing backend configured - i think it's marginally clearer in that case that tracing is disabled?

the other option would be to have two things in here, `tracing_enabled bool` and `tracing_backend bool omitempty`

pda: Or, `open_tracing_enabled` and `open_telemetry_enabled` bools.
I think that's my favourite.

api/agents.go
sj26: Maybe the other direction, what about just `features`?

sj26: I'm overthinking it.

I've adjusted the backend to accept `feature_usage` as a json blob.

sj26 (approved):
Nice, and thanks!

api/agents.go
```diff
@@ -2,18 +2,29 @@ package api
...
+	Arch               string        `json:"arch"`
+	ScriptEvalEnabled  bool          `json:"script_eval_enabled"`
+	IgnoreInDispatches bool          `json:"ignore_in_dispatches"`
+	Priority           string        `json:"priority,omitempty"`
+	Version            string        `json:"version"`
+	Build              string        `json:"build"`
+	Tags               []string      `json:"meta_data"`
+	PID                int           `json:"pid,omitempty"`
+	MachineID          string        `json:"machine_id,omitempty"`
+	FeatureUsage       *FeatureUsage `json:"feature_usage,omitempty"`
```
sj26, line 17: I'd like to call this parameter `features`, and make it an array of simple strings.

clicommand/agent_start.go
```diff
@@ -94,6 +94,7 @@ type AgentStartConfig struct {
 	NoPlugins                   bool     `cli:"no-plugins"`
 	NoPluginValidation          bool     `cli:"no-plugin-validation"`
 	NoPTY                       bool     `cli:"no-pty"`
+	NoTelemetry                 bool     `cli:"no-telemetry"`
```
sj26, line 97: I thin "no telemetry" is too broad. Folks might expect us not to send operating system or process information under this name.

I'd suggest instead we use `--no-features-reporting` to be more specific.

clicommand/agent_start.go
```diff
@@ -414,6 +457,11 @@ var AgentStartCommand = cli.Command{
 			Usage:  "Send metrics to DogStatsD for Datadog",
 			EnvVar: "BUILDKITE_METRICS_DATADOG",
 		},
+		cli.BoolFlag{
+			Name:   "no-telemetry",
+			Usage:  "Disables sending usage statistics back to the buildkite mothership. We use the information gathered from telemetry to figure out what features are getting used.",
+			EnvVar: "BUILDKITE_AGENT_NO_TELEMETRY",
+		},
```
sj26, line 464: This one needs to be updated I think?

sj26 (approved):
Looks good! I noticed the bool flag didn't catch up on the rename, but otherwise this looks great to go

clicommand/agent_start.go
```diff
@@ -130,6 +131,48 @@ type AgentStartConfig struct {
...
+	if asc.GitMirrorsPath != "" {
+		features = append(features, "git-mirrors")
+	}
+
+	if asc.AcquireJob != "" {
+		features = append(features, "acquire-job")
+	}
+
+	if asc.TracingBackend == tracetools.BackendDatadog {
+		features = append(features, "tracing-datadog")
```
sj26, line 146:
```suggestion
		features = append(features, "datadog-tracing")
```

clicommand/agent_start.go
```diff
@@ -130,6 +131,48 @@ type AgentStartConfig struct {
...
+	}
+
+	if asc.TracingBackend == tracetools.BackendOpenTelemetry {
+		features = append(features, "tracing-opentelemetry")
```
sj26, line 150:
```suggestion
		features = append(features, "opentelemetry-tracing")
```

sj26:
I tried this out end-to-end against a development version of Buildkite and observed it recording feature usage 🙌

pda (approved):
Looks good to me, I've tested it locally.
Ideally some tests would be nice. Not sure if we have an existing place to put them for this layer of the agent, though.

PR #1709, opened 2022-07-18
Don't propagate HookExitErrors to tracing

moskyb:
At current, when tracing a job using opentel, if that job uses the docker plugin (amongst some others), the trace always reports an error, even if the buildkite job passed:
CleanShot 2022-07-18 at 16 08 06@2x

When these errors happen as part of the bootstrap process, we ignore them, but because the `err` variable is getting set, it gets captured by tracing.

**This PR:** Updates the code to set `err = nil` if we run across a `HookExitError`, which stops these errors from being captured by tracing.

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -323,7 +323,7 @@ func (b *Bootstrap) executeHook(ctx context.Context, hookCfg HookConfig) error {
 		case *hook.HookExitError:
 			// ...because the hook called exit(), tsk we ignore any changes
 			// since we can't discern them but continue on with the job
-			break
+			err = nil // set err to nil, otherwise the HookExitError will be captured by tracing
 		default:
 			// ...because something else happened, report it and stop the job
 			return errors.Wrapf(err, "Failed to get environment")


bootstrap/bootstrap.go
```diff
@@ -323,7 +323,7 @@ func (b *Bootstrap) executeHook(ctx context.Context, hookCfg HookConfig) error {
 		case *hook.HookExitError:
 			// ...because the hook called exit(), tsk we ignore any changes
 			// since we can't discern them but continue on with the job
-			break
+			err = nil // set err to nil, otherwise the HookExitError will be captured by tracing
```
pda, line 326: Took me a moment to find this 82 lines up and understand how that `err` was propagating given it wasn't mentioned after the `switch`:

```go
	var err error
```

Do you think the `break` you removed was never needed? Seems like a C/etc habit of breaking each case to stop fall-through, but Go never does that anyway, right?

pda: Something still feels a bit off with `err = nil` but I think your patch is the right small change to make for now.

Hrmm. In the `default:` case of `return errors.Wrapf(err, "Failed to get environment")` does the `defer func() { span.FinishWithError(err) }()` get the wrapper error, or the unwrapped error? Does it matter? I guess it gets the unwrapped error, but the wrapping doesn't look terribly important for the trace.

pda (approved):
Very nice 👍🏼

TIL `HookExitError`

bootstrap/bootstrap.go
pda: ¹ The `defer` up the top propagating `err` which is subsequently set from multiple different sources in the following ~100 lines is what feels not-quite-right to me, but it's probably fine, and maybe even the best approach.

moskyb: > Do you think the break you removed was never needed?

i don't think so; go breaks by default, unless `passthrough` is the last statement in your case. what's interesting is that originally i just added `err = nil` to the existing case, leaving the break intact, but my editor gave me a warning:
CleanShot 2022-07-18 at 17 16 33@2x
(the tooltip is covering `err = nil`)

... but it didn't have the warning previously. odd.

re: the wrapped error, yeah, i think the trace will get the unwrapped error, but i also agree that that's probably fine.

moskyb: > The defer [...] is what feels not-quite-right to me, but it's probably fine

yeah; i agree. i think it's probably the best way to do this in this case, as the only other option would be to call `span.FinishWithError()` every time we return? i guess?

maybe there's some middle ground like 
```Go
func executeHook(...) error {
  defer span.Finish()
  // ...
  if err != nil {
    span.SetError(err)
    return err
  }
}
```

but that also still leaves the door open for someone to accidentally leave a `span.SetError` call out by accident... i'm honestly not sure that there's a 100% nice solution for this

pda: My brain kept wondering if there's a better way.

Might this be better?

```diff
diff --git a/bootstrap/bootstrap.go b/bootstrap/bootstrap.go
index c5d4df00..9eb768c1 100644
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -243,6 +243,7 @@ func (b *Bootstrap) executeHook(ctx context.Context, hookCfg HookConfig) error {
 	span, ctx := tracetools.StartSpanFromContext(ctx, spanName, b.Config.TracingBackend)
 	var err error
 	defer func() { span.FinishWithError(err) }()
+
 	span.AddAttributes(map[string]string{
 		"hook.type":    scopeName,
 		"hook.name":    hookCfg.Name,
@@ -314,26 +315,20 @@ func (b *Bootstrap) executeHook(ctx context.Context, hookCfg HookConfig) error {
 	b.shell.Env.Set("BUILDKITE_LAST_HOOK_EXIT_STATUS", "0")
 
 	// Get changed environment
-	changes, err := script.Changes()
-	if err != nil {
+	changes, changesErr := script.Changes()
+	// Hooks calling exit() prevent us from capturing environment changes.
+	// This manifests as HookExitError, but the job should continue.
+	var hookExitError *hook.HookExitError
+	if changesErr != nil && errors.As(changesErr, &hookExitError) {
 		// Could not compute the changes in environment or working directory
-		// for some reason...
-
-		switch err.(type) {
-		case *hook.HookExitError:
-			// ...because the hook called exit(), tsk we ignore any changes
-			// since we can't discern them but continue on with the job
-			err = nil // set err to nil, otherwise the HookExitError will be captured by tracing
-		default:
-			// ...because something else happened, report it and stop the job
-			return errors.Wrapf(err, "Failed to get environment")
-		}
-	} else {
-		// Hook exited successfully (and not early!) We have an environment and
-		// wd change we can apply to our subsequent phases
-		b.applyEnvironmentChanges(changes, redactors)
+		// for some reason... report it and stop the job.
+		return errors.Wrapf(changesErr, "Failed to get environment")
 	}
 
+	// Hook exited successfully (and not early!) We have an environment and
+	// wd change we can apply to our subsequent phases
+	b.applyEnvironmentChanges(changes, redactors)
+
 	return nil
 }
```

Resulting in this code:

```go
	// Get changed environment
	changes, changesErr := script.Changes()
	// Hooks calling exit() prevent us from capturing environment changes.
	// This manifests as HookExitError, but the job should continue.
	var hookExitError *hook.HookExitError
	if changesErr != nil && errors.As(changesErr, &hookExitError) {
		// Could not compute the changes in environment or working directory
		// for some reason... report it and stop the job.
		return errors.Wrapf(changesErr, "Failed to get environment")
	}

	// Hook exited successfully (and not early!) We have an environment and
	// wd change we can apply to our subsequent phases
	b.applyEnvironmentChanges(changes, redactors)
```

I think I like it better because it follow the “if error, early exit returning the error” pattern used elsewhere in this `Bootstrap.executeHook()` method, and simplifies the state machine that is `err` in this method (it can only transition from `nil`→something, not back again).

I _don't_ really like Go's [`errors.As()`](https://pkg.go.dev/errors#As) and the way it requires a temporary variable to compare against, but that does seem to be the idiomatic Go way to type-check the error in an `if` condition.

🤷‍♂️

pda: Actually, instead of `return errors.Wrapf(…)` it would need to be:

```go
err = errors.Wrapf(changesErr, "Failed to get environment")
return err
```

But that's still pretty good, and maintains the “err only transitions from nil to non-nil” simplicity.

pda:
@moskyb and I did some digging into where this `HookExitError` is coming from.

It actually seems like a bug, perhaps in the docker-compose plugin, or maybe in the ScriptWrapper, which has been hidden by the fact that we silently ignore that error.

It's worth noting that while `HookExitError` implies that the hook did an early exit, this is only detected indirectly by the absence of any environment variables [captured by the ScriptWrapper after the hook runs](https://github.com/buildkite/agent/blob/abe508d438e6028356f16a8102ad7783fc0d2a79/hook/scriptwrapper.go#L139).

This error condition prevents the environment from being updated, which (a) is never a problem for a docker-compose plugin command hook, and (b) means it would definitely have been noticed if it was happening in the kind of hooks that _do_ want to change environment.

I would be inclined to leave the error on the OpenTelemetry trace, because I think an error is happening, and hope that we can fix the underlying problem in due course.

moskyb:
further to paul's comment, i'm gonna close this PR.

PR #1712, opened 2022-07-18
Split bootstrap

moskyb:
the bootstrap.go file is currently about 4k lines long, and is basically impossible to navigate with anything other than ctrl-f and judicious use of go to definition.

This PR splits the bootstrap file up into 6(!) files - one for the main bootstrap process, still called bootstrap.go, and one each for:
- Checkout-related stuff (`checkout.go`)
- Artifacts (`artifacts.go`)
- Plugin handling (`plugin.go`)
- Command running (`command.go`)
- Hook handling (`hook.go`)

It also does a minor clean up in the `*Bootstrap.Run()` function


lox:
Another approach to this might be to actually split out the phases as different structs in different files. 🤔

moskyb:
@lox yeah, i agree, but that's a lot more difficult than just splitting up a file and calling it a day 😅

i'd love to have more functional separation between bootstrap phases as well as having them just live in different files, but i think it achieves a different thing that this PR does, which is just make the bootstrap process more navigable from a "having files open in your editor" standpoint

bootstrap/artifacts.go
```diff
@@ -0,0 +1,98 @@
+package bootstrap
+
+import (
+	"context"
+
+	"github.com/buildkite/agent/v3/tracetools"
+)
+
+func (b *Bootstrap) artifactPhase(ctx context.Context) error {
```
DrJosh9000, line 9: A named return arg can help simplify some of the code in this function:

`func (b *Bootstrap) artifactPhase(ctx context.Context) (err error) {
...snip...

    // note this next line sees the returned value, even though it is never assigned with =

    if err := b.preArtifactHooks(ctx); err != nil {
        return err
    }

 	if err := b.uploadArtifacts(ctx); err != nil {
 		return err
 	}

 	return b.postArtifactHooks(ctx)
}`

DrJosh9000: Oh, feel free to ignore me - splitting the existing code should come before cleanups.

PR #1713, opened 2022-07-19
Less hideous script wrapper

moskyb:
The ScriptWrapper part of the bootstrap is really hard to understand, but honestly, a large part of that is because the scripts themselves are really hard to read. At current, they use string concatenation nonsense that's just really difficult to decipher.

This PR takes the string concatenation nonsense and replaces it with Go Templates, as well as adding tests to catch regressions in script generation

### DIFF
--- a/hook/scriptwrapper.go
+++ b/hook/scriptwrapper.go
@@ -13,29 +13,10 @@ import (
 )
 
 const (
-	hookExitStatusEnv = `BUILDKITE_HOOK_EXIT_STATUS`
-	hookWorkingDirEnv = `BUILDKITE_HOOK_WORKING_DIR`
+	hookExitStatusEnv = "BUILDKITE_HOOK_EXIT_STATUS"
+	hookWorkingDirEnv = "BUILDKITE_HOOK_WORKING_DIR"
+
 )
-
-// Hooks get "sourced" into the bootstrap in the sense that they get the
-// environment set for them and then we capture any extra environment variables
-// that are exported in the script.
-
-// The tricky thing is that it's impossible to grab the ENV of a child process
-// before it finishes, so we've got an awesome (ugly) hack to get around this.
-// We write the ENV to file, run the hook and then write the ENV back to another file.
-// Then we can use the diff of the two to figure out what changes to make to the
-// bootstrap. Horrible, but effective.
-
-// ScriptWrapper wraps a hook script with env collection and then provides
-// a way to get the difference between the environment before the hook is run and
-// after it
-type ScriptWrapper struct {
-	hookPath      string
-	scriptFile    *os.File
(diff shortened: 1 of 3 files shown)


hook/scriptwrapper.go
```diff
@@ -180,6 +162,36 @@ func NewScriptWrapper(opts ...scriptWrapperOpt) (*ScriptWrapper, error) {
 	return wrap, nil
 }
 
+func (wrap *ScriptWrapper) GenerateCmdDotExeScript(pathToHook string) string {
```
pda: I believe we call these batch scripts elsewhere

pda (approved):
That is so much nicer 👌🏼

Here's a fake diff just comparing the powershell before/after:

```diff
-return `$ErrorActionPreference = "STOP"` + "\n" +
-	`Get-ChildItem Env: | Foreach-Object {"$($_.Name)=$($_.Value)"} | Set-Content "` + wrap.beforeEnvFile.Name() + `"` + "\n" +
-	pathToHook + "\n" +
-	`if ($LASTEXITCODE -eq $null) {$Env:` + hookExitStatusEnv + ` = 0} else {$Env:` + hookExitStatusEnv + ` = $LASTEXITCODE}` + "\n" +
-	`$Env:` + hookWorkingDirEnv + ` = $PWD | Select-Object -ExpandProperty Path` + "\n" +
-	`Get-ChildItem Env: | Foreach-Object {"$($_.Name)=$($_.Value)"} | Set-Content "` + wrap.afterEnvFile.Name() + `"` + "\n" +
-	`exit $Env:` + hookExitStatusEnv
+	powershellScript = `$ErrorActionPreference = "STOP"
+Get-ChildItem Env: | Foreach-Object {"$($_.Name)=$($_.Value)"} | Set-Content "{{.BeforeEnvFileName}}"
+{{.PathToHook}}
+if ($LASTEXITCODE -eq $null) {$Env:BUILDKITE_HOOK_EXIT_STATUS = 0} else {$Env:BUILDKITE_HOOK_EXIT_STATUS = $LASTEXITCODE}
+$Env:BUILDKITE_HOOK_WORKING_DIR = $PWD | Select-Object -ExpandProperty Path
+Get-ChildItem Env: | Foreach-Object {"$($_.Name)=$($_.Value)"} | Set-Content "{{.AfterEnvFileName}}"
+exit $Env:BUILDKITE_HOOK_EXIT_STATUS`
```

Also, this might encourage making the code more readable, adding comments etc if we decide that's a good idea, which would have been a pain and added noise in the old approach.

hook/scriptwrapper.go
moskyb: Ooh good call!

PR #1717, opened 2022-07-20
Add Code of Conduct

moskyb:
Our lovely agent is actually a pretty big opensource project! It's awesome to have so many lovely contributors, but it's still important to have a reasonable set of expectations of the community (buildkite employees included!), and have a framework upon which to enforce those expectations, if necessary.

To my knowledge, we've never been in a situation where we'd have needed to invoke the CoC, but it's much better to have these things and not need them than it is to need them and not have them.

This PR adds a code of conduct to this repo, adapted from the [Contributor Covenant v2.1](https://www.contributor-covenant.org/version/2/1/code_of_conduct/).

See also: https://github.com/buildkite/elastic-ci-stack-for-aws/pull/1038

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,132 @@
+
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in our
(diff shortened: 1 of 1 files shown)


pda:
> much better to have these things and not need them than it is to need them and not have them

👌🏼

pda (approved):

PR #1719, opened 2022-07-21
Update bundler version to be compatible with Ruby 2.7.6 alpine image

moskyb:
For some reason, after [updating our deployer image](https://github.com/buildkite/agent/pull/1716), debian releases are now failing, as it appears that bundler has gone away as part of that process.

This PR runs `gem install bundler` before running `bundle install` in the debian deploy process.

It'd be nice to get rid of the gemfile completely - this is a go project after all, but that's a problem for tomorrow.

### DIFF
--- a/.buildkite/docker-compose.yml
+++ b/.buildkite/docker-compose.yml
@@ -15,10 +15,10 @@ services:
       - "BUILDKITE_BUILD_PATH=/buildkite"
 
   ruby:
-    image: ruby:2.6
+    image: ruby:2.7.6
     volumes:
       - ..:/work
-      - ruby-2-6-agent-gem-cache:/usr/local/bundle
+      - ruby-2-7-6-agent-gem-cache:/usr/local/bundle
     working_dir: /work
     command: /bin/bash
     environment:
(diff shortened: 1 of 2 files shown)


yob (approved):
Is this running ruby in the deploytools image? I'm very surprised bundler is missing there, it ruby 2.7 it's shipped with the standard  ruby install and is usually Just There.

Maybe alpine has broken it out to a dedicated package or something? Anyway, I assume this should work around it.

moskyb:
@yob really good point - i looked into this and found that the issue isn't actually that we don't have bundler installed, we totally do. The problem is that bundler is locked to 1.16.3, where ruby 2.7, which the deploytools container is now shipped with 2.1.4. I'll update our gemfile.lock so that it'll accept a newer version of bundler.

Actual error message is [here](https://buildkite.com/buildkite/agent-release-edge/builds/433#01821e20-2c4a-4c5f-a2b8-3581b0f56533) it'd be good if i actually read the error message before submitting a fix lol

yob (approved):

Issue #1720, opened 2022-07-21
--log-level error includes warning


ChrisBr:
I'm calling `buildkite-agent pipeline upload --log-level error` and the output includes

```
2022-07-21 12:45:57 WARN   Error running git rev-parse "58a5bebc9abd8ce45e3d29e1b608c56cf8e3d88f": exit status 128
```

I would've expected this to be the log level order

https://github.com/buildkite/agent/blob/c355645bdd5c754de0a1c7238bbe27d95a9c7a8f/logger/log.go#L158-L171

So setting the log level to error should only include error and fatal logs, no?

ChrisBr:
Setting the log level to fatal will exclude the warning.

PR #1721, opened 2022-07-21
Fix warning log level

ChrisBr:
`ERROR` should have a higher log level than `WARNING`.

Fix https://github.com/buildkite/agent/issues/1720


moskyb (approved):

moskyb:
oooh thank you @ChrisBr! good catch!

PR #1722, opened 2022-07-21
Fix order of level names for logger.Level.String()

moskyb:
We changed the iota placement of the `logger.Level` const in #1720 to fix log levels, but in doing so forgot to chage the placement in the `levelNames` array. This means that if someone were to use `--log-level error` it would behave like error, but say it was using warning logging.

I really wish that go had real enums.


tessereth (approved):

pda (approved):
It's like a gun that can only work when pointed at your own feet

Issue #1723, opened 2022-07-24
Retry logic not working in 3.37.0


ChrisBr:
We just updated from 3.35.2 to 3.37.0.

We have a build which is getting cancelled but all jobs are stuck in an endless loop retrying.

```
kubectl logs $AGENT_IT builder | grep "Buildkite rejected the call to start the job" | wc -l
     758
```

So there are 758 retries on this agent!

We got hundreds of these messages and the job is stuck for >1h.
```
{"ts":"2022-07-24T07:02:38Z","level":"INFO","msg":"Assigned job 01822a27-1e37-4968-8126-79aaecef5a44. Accepting...","agent":"shopify-build-docker-6fb69d7c48-mphf9"}
{"ts":"2022-07-24T07:02:38Z","level":"INFO","msg":"Starting job 01822a27-1e37-4968-8126-79aaecef5a44","agent":"shopify-build-docker-6fb69d7c48-mphf9"}
{"ts":"2022-07-24T07:02:38Z","level":"WARN","msg":"Buildkite rejected the call to start the job (PUT https://agent-shopify.buildkite.com/v3/jobs/01822a27-1e37-4968-8126-79aaecef5a44/start: 422 This job can't be started while the build is in a `canceling` state.)","agent":"shopify-build-docker-6fb69d7c48-mphf9"}
{"ts":"2022-07-24T07:02:38Z","level":"ERROR","msg":"Failed to run job: PUT https://agent-shopify.buildkite.com/v3/jobs/01822a27-1e37-4968-8126-79aaecef5a44/start: 422 This job can't be started while the build is in a `canceling` state.","agent":"shopify-build-docker-6fb69d7c48-mphf9"}
{"ts":"2022-07-24T07:02:46Z","level":"INFO","msg":"Assigned job 01822a27-1e37-4968-8126-79aaecef5a44. Accepting...","agent":"shopify-build-docker-6fb69d7c48-mphf9"}
{"ts":"2022-07-24T07:02:46Z","level":"INFO","msg":"Starting job 01822a27-1e37-4968-8126-79aaecef5a44","agent":"shopify-build-docker-6fb69d7c48-mphf9"}
[...]

PR #1724, opened 2022-07-24
Halt retrying

ChrisBr:
We just updated from 3.35.2 to 3.37.0.

We have a build which is getting cancelled but all jobs are stuck in an endless loop retrying.

```
kubectl logs $AGENT_IT builder | grep "Buildkite rejected the call to start the job" | wc -l
     758
```

So there are 758 retries on this agent!

We got hundreds of these messages and the job is stuck for >1h.
```
{"ts":"2022-07-24T07:02:38Z","level":"INFO","msg":"Assigned job 01822a27-1e37-4968-8126-79aaecef5a44. Accepting...","agent":"shopify-build-docker-6fb69d7c48-mphf9"}
{"ts":"2022-07-24T07:02:38Z","level":"INFO","msg":"Starting job 01822a27-1e37-4968-8126-79aaecef5a44","agent":"shopify-build-docker-6fb69d7c48-mphf9"}
{"ts":"2022-07-24T07:02:38Z","level":"WARN","msg":"Buildkite rejected the call to start the job (PUT https://agent-shopify.buildkite.com/v3/jobs/01822a27-1e37-4968-8126-79aaecef5a44/start: 422 This job can't be started while the build is in a `canceling` state.)","agent":"shopify-build-docker-6fb69d7c48-mphf9"}
{"ts":"2022-07-24T07:02:38Z","level":"ERROR","msg":"Failed to run job: PUT https://agent-shopify.buildkite.com/v3/jobs/01822a27-1e37-4968-8126-79aaecef5a44/start: 422 This job can't be started while the build is in a `canceling` state.","agent":"shopify-build-docker-6fb69d7c48-mphf9"}
{"ts":"2022-07-24T07:02:46Z","level":"INFO","msg":"Assigned job 01822a27-1e37-4968-8126-79aaecef5a44. Accepting...","agent":"shopify-build-docker-6fb69d7c48-mphf9"}
{"ts":"2022-07-24T07:02:46Z","level":"INFO","msg":"Starting job 01822a27-1e37-4968-8126-79aaecef5a44","agent":"shopify-build-docker-6fb69d7c48-mphf9"}
{"ts":"2022-07-24T07:02:46Z","level":"WARN","msg":"Buildkite rejected the call to start the job (PUT https://agent-shopify.buildkite.com/v3/jobs/01822a27-1e37-4968-8126-79aaecef5a44/start: 422 This job can't be started while the build is in a `canceling` state.)","agent":"shopify-build-docker-6fb69d7c48-mphf9"}
{"ts":"2022-07-24T07:02:46Z","level":"ERROR","msg":"Failed to run job: PUT https://agent-shopify.buildkite.com/v3/jobs/01822a27-1e37-4968-8126-79aaecef5a44/start: 422 This job can't be started while the build is in a `canceling` state.","agent":"shopify-build-docker-6fb69d7c48-mphf9"}
```

I believe it might be related to https://github.com/buildkite/agent/pull/1675

I'm not entirely sure this will fix the error but the example in [roko says this](https://github.com/buildkite/roko)

```go
r := roko.NewRetrier(
  roko.WithMaxAttempts(3),                           // Only try 3 times, then give up
  roko.WithStrategy(roko.Constant(5 * time.Second)), // Wait 5 seconds between attempts
)

err := r.Do(func(r *roko.Retrier) error {
  err := canFail()
  if err.Is(errorUnrecoverable) {
    r.Break()  // Give up, we can't recover from this error
    return err // We still need to return from this function, Break() doesn't halt this callback
    // return nil would be appropriate too, if we don't want to handle this error further
  }
})
```

Close https://github.com/buildkite/agent/issues/1723

### DIFF
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -693,6 +693,7 @@ func (r *JobRunner) startJob(startedAt time.Time) error {
 			} else {
 				r.logger.Warn("Buildkite rejected the call to start the job (%s)", err)
 				rtr.Break()
+				return err
 			}
 		}
 


ChrisBr:
Hmm not too sure that will fix it 🤔 The retry should be halted anyway and the final return should return the error. Might be a complete red herring.

moskyb:
@ChrisBr yeah, i think the retrier thing might be a red herring - it'll already return the 422 error, without the early return you've added.

I'm pretty sure that this is an issue with the buildkite backend - i think what's happening here is:
- Our dispatcher doesn't know that it shouldn't dispatch jobs in the `cancelling` state, so it tries to dispatch this job to the agent our unhappy agent here
- The agent receives the cancelling job, and says "yep, sure thing!" and tries to accept the job with the buildkite backend
- Buildkite tells the agent that it's not allowed to accept a job in the `cancelling` state
- The agent says "oh bummer, okay", spits out the error message above, and goes back to polling for work
- Repeat ad nauseam

we have a couple of bits of work ongoing to fix this issues and issues like it - in the meantime, manually cancelling the job in the BK ui should resolve this issue, but i realise that it's not an ideal solution. i'll get back to you when we've made some progress on the underlying issue.

ChrisBr:
> I'm pretty sure that this is an issue with the buildkite backend - i think what's happening here is:

What I'm just surprised is that this agent gets assigned the same job again and again. We had a few hundred agents running, shouldn't there be a bit of a mix if it's coming from the backend?

> in the meantime, manually cancelling the job in the BK ui should resolve this issue, but i realise that it's not an ideal solution.

Yeah that's not ideal as we got builds with a few 100 steps in parallel.

I never saw this before so it was just a bit of an alarm bell as it haapened just after the agent upgrade.

moskyb:
> What I'm just surprised is that this agent gets assigned the same job again and again

i think this is happening because the dispatcher has already assigned this agent to this job, and the agent accepted it, so when the agent pings for work again, the dispatcher says "i've already assigned this one to you".

> Yeah that's not ideal as we got builds with a few 100 steps in parallel.

yep, i totally understand - we're working on it, and i hope to be able to update you soon

ChrisBr:
Closing this one then as it seems a backend issue.

PR #1727, opened 2022-07-25
Update OpenTelemetry libs to 1.8.0

moskyb:
Supersedes #1726 and #1725 

All of the opentel libraries have the same version numbers and get pushed at the same time, so we should probably update them all in one go.


tessereth (approved):

go.mod
```diff
@@ -23,7 +23,7 @@ require (
 	github.com/qri-io/jsonschema v0.0.0-20180607150648-d0d3b10ec792
 	github.com/rjeczalik/interfaces v0.1.1
 	github.com/sergi/go-diff v1.0.0 // indirect
-	github.com/stretchr/testify v1.7.3
+	github.com/stretchr/testify v1.8.0
```
pda, line 26: https://github.com/stretchr/testify/compare/v1.7.3...v1.8.0
Seems okay at a glance

But I wonder why testify is being updated in this when nothing else non-otel related is. All good, though.

PR #1730, opened 2022-07-28
Load the AWS Shared Credentials for s3 operations

### LINKED ISSUE #1729: Support AWS_PROFILE for artifact uploads
Currently the [buildkite-agent artifact upload](https://buildkite.com/docs/agent/v3/cli-artifact#using-your-private-aws-s3-bucket-credentials) does not support named AWS profiles

To utilise assumed roles in a Buildkite CI stack ; add supported flags such as: BUILDKITE_S3_PROFILE which would utilise a AWS profile to upload objects to a S3 bucket

Currently I have to extract the non-assumed role and pass it into the step; this breaks our model of jobs assuming roles to-do their work

**Additional context**
Was discussed on Cashapp/Block slack with @lox and Jarryd Tilbrook

lox:
Currently the agent doesn't respect `AWS_PROFILE` or the shared credentials files in `~/.aws`. This is because this wasn't initially supported in the AWS SDK. 

I think this would probably be a breaking change, although the impacted folks would likely be very few. 

Added behind an experiment, I think this would make a sensible future default. 

Closes #1729.


moskyb:
this lgtm, but i probably wouldn't even bother with putting it behind an experiment - i don't think adding support for profiles is a breaking change. If they already exist on the agent machine then that's probably what the user was expecting the agent to use anyway.

lox:
Updated to remove the experiment @moskyb

agent/s3.go
```diff
@@ -74,6 +75,14 @@ func awsS3Session(region string) (*session.Session, error) {
 	return sess, nil
 }
 
+func sharedCredentialsProvider() credentials.Provider {
+	// If empty SDK will default to environment variable "AWS_PROFILE"
+	// or "default" if environment variable is also not set.
+	awsProfile := os.Getenv("BUILDKITE_S3_PROFILE")
+
+	return &credentials.SharedCredentialsProvider{Filename: "", Profile: awsProfile}
```
moskyb, line 83: [super nit] I think having the Filename explicitly blank here is a little confusing, even though that's what it defaults(?) to

```suggestion
	return &credentials.SharedCredentialsProvider{Profile: awsProfile}
```

moskyb (approved):

agent/s3.go
lox: I tend to agree, I copied this code from the AWS SDK, but let's ditch it.

PR #1731, opened 2022-07-28
Do not retry on not found resource

leakingtapan:
We found Buildkite agent retry on 404 when downloading artifacts, but this is unnecessary. Since we have extended retry (up to 30 mins), the current behavior (retry on 404) largely slows down the progress to fail fast. So add the small patch to skip retry on  404 for artifact download.


leakingtapan:
@lox

lox:
I'm no longer Buildkite staff @leakingtapan! Just a contributor like yourself.

leakingtapan:
ah, got it. Whom should I seek approval for?

leakingtapan:
@moskyb could you take a look? thx

DrJosh9000:
Closing due to staleness. If this is still something that's worth considering, please holler.

PR #1732, opened 2022-07-31
Cache S3 clients between artifact downloads

moskyb:
**🤔 Problem:** Currently, when downloading multiple artifacts in the agent, we initialise an S3 client for each artifact that we're downloading. This is problematic for a couple of reasons, first of which is that the credential initialisation actually [involves a call to S3](https://github.com/buildkite/agent/blob/main/agent/s3.go#L132-L135), so if there's a lot of artifacts to download, the the `ListObjects` calls made by each credential initialisation will contribute to our rate limits.

The other issue is that we don't need to create an S3 client per call, and it actually hurts us in a couple of ways; first and foremost, as reported by a customer, each client initialisation first has to authenticate with AWS. We've seen evidence that when a lot of artifacts are being uploaded, this authentication process being performed more often than necessary leads to the AWS EC2 Instance Metadata Service rate-limiting the instance asking for creds, leading to some artifacts not being uploaded.

There are other less-important issues with using an S3 client per artifact upload, mostly that it's just kinda wasteful and takes up more memory than necessary.

**✅ Solution:** When uploading multiple artifacts to S3, share an instance of the S3 client between calls. 

There's also a "while i'm here" or two hanging around in this PR, namely:
- Increase the artifact up/download concurrency limit from the completely arbitrary "the number of CPUs on this machine" to the still arbitrary but much higher "CPUs*10". I'm not sure why the limit was set to the number of CPUs, it seems to ignore what goroutines are best at, but 🤷. We considered letting artifact up/downloads have unbounded concurrency, as this would make them waaaaaaaay faster, but rate limiting on S3 is a thing, and it's probably best to keep these things bounded for now
- Rename our Buildkite-specific AWS credentials provider from `credentialsProvider` to `buildkiteCredentialsProvider`. 

**This PR is best reviewed commit-by-commit, and "Hide Whitespace" on (i dedented a block and it makes the diff confusing)**

CleanShot 2022-08-01 at 10 28 07@2x

### DIFF
--- a/agent/artifact_downloader.go
+++ b/agent/artifact_downloader.go
@@ -123,7 +123,7 @@ func (a *ArtifactDownloader) Download() error {
 				}
 
 				err = NewS3Downloader(a.logger, S3DownloaderConfig{
-					S3Client:    s3Clients[artifact.UploadDestination],
+					S3Client:    s3Clients[bucketName],
 					Path:        path,
 					S3Path:      artifact.UploadDestination,
 					Destination: downloadDestination,
--- a/agent/artifact_downloader.go
+++ b/agent/artifact_downloader.go
@@ -9,6 +9,7 @@ import (
 	"runtime"
 	"strings"
 
+	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/buildkite/agent/v3/logger"
 	"github.com/buildkite/agent/v3/pool"
 )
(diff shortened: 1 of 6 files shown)


agent/s3.go
```diff
@@ -39,17 +41,18 @@ func (e *credentialsProvider) Retrieve() (creds credentials.Value, err error) {
...
+		return credentials.Value{}, errors.New("BUILDKITE_S3_ACCESS_KEY_ID or BUILDKITE_S3_ACCESS_KEY not found in environment")
 	}
+
 	if creds.SecretAccessKey == "" {
-		err = errors.New("BUILDKITE_S3_SECRET_ACCESS_KEY or BUILDKITE_S3_SECRET_KEY not found in environment")
+		return credentials.Value{}, errors.New("BUILDKITE_S3_SECRET_ACCESS_KEY or BUILDKITE_S3_SECRET_KEY not found in environment")
 	}
 
 	e.retrieved = true
-	return
```
moskyb, line 49: i think named returns are my least favourite feature in all of go. i have removed them from the files i've touched in this PR.

agent/s3_downloader.go
```diff
@@ -12,8 +12,11 @@ import (
 )
 
 type S3DownloaderConfig struct {
+	//
```
pda: // 👋🏼

agent/artifact_downloader.go
```diff
@@ -9,6 +9,7 @@ import (
 	"runtime"
 	"strings"
 
+	"github.com/aws/aws-sdk-go/service/s3"
```
pda, line 12: It's a slight shame the the S3-specific internals leak out to this layer.

I guess the alternative would be an s3_downloader context or wrapped client or something… but that would add complexity and it's not even in a separate package, so all good.

agent/artifact_downloader.go
```diff
@@ -76,74 +77,96 @@ func (a *ArtifactDownloader) Download() error {
...
+			}
+
+			// Handle downloading from S3, GS, or RT
+			if strings.HasPrefix(artifact.UploadDestination, "s3://") {
+				// We want to have as few S3 clients as possible, as creating them is kind of an expensive operation
+				// But it's also theoretically possible that we'll have multiple artifacts with different S3 buckets, and each
+				// S3Client only applies to one bucket, so we need to store the S3 clients in a map, one for each bucket
+				bucketName, _ := ParseS3Destination(artifact.UploadDestination)
+				if _, has := s3Clients[bucketName]; !has {
+					p.Lock()
```
pda, line 110: I think this is open to a race condition, although perhaps a harmless one…

When multiple goroutines spawn concurrently for the same bucket, they'll each find is missing from the hash, and then take it in turns to wait/acquire the lock, instantiate a client, and place it in the `s3Clients` map.

When they pass `S3Client:    s3Clients[bucketName]` later on, it'll be unpredictable which instance will win.

I think the lock/unlock needs to go outside/around the `if _, has := s3Clients[bucketName]` bit?

pda: I would be tempted to do an entirely separate `for _, artifact := range artifacts {` prior to this one, just to prepare a `map[bucketstring]client` that the concurrent workers can treat as entirely read-only.

agent/s3_downloader.go
moskyb: this is fixed up in a later commit 😶

agent/artifact_downloader.go
moskyb: I think you're right - part of me is like "oh no, it's O(2n) instead of O(n)!" but it pretty much doesn't matter at all compared to the amount of IO/etc we're doing. I'll do a cheeky lil pre-loop here.

agent/artifact_downloader.go
```diff
@@ -76,81 +78,112 @@ func (a *ArtifactDownloader) Download() error {
...
+	s3Clients := map[string]*s3.S3{}
+
+	for _, artifact := range artifacts {
+		if !strings.HasPrefix(artifact.UploadDestination, "s3://") {
+			continue
 		}
 
-		p.Wait()
+		bucketName, _ := ParseS3Destination(artifact.UploadDestination)
+		if _, has := s3Clients[bucketName]; !has {
```
tessereth, line 178: This go-ism took me way too long to understand. But I think I got there. This runs the block if `bucketName` isn't already in the map.

agent/s3_downloader_test.go
```diff
@@ -11,12 +11,12 @@ func TestS3DowloaderBucketPath(t *testing.T) {
 	t.Parallel()
 
 	s3Uploader := NewS3Downloader(logger.Discard, S3DownloaderConfig{
```
tessereth, line 13: Not relevant to this PR but I noticed this in passing through this whole file..
```suggestion
	s3Downloader := NewS3Downloader(logger.Discard, S3DownloaderConfig{
```

tessereth (approved):
This all seems legit to me.

agent/s3_downloader_test.go
moskyb: oh nooooo

moskyb: will fix ahahaha

agent/artifact_downloader.go
moskyb: yeah, something like `maps.HasKey(s3Clients, bucketName)` would be too simple and readable aye

(`maps.HasKey` does not exist)

pda: I tend to use `ok` e.g. `if _, ok := s3Clients[bucketName]; !ok {` which is no less mysterious but I thought it was more Go-idiomatic, but now I'm not sure.

pda: [This](https://go.dev/doc/effective_go#maps) implies `ok` is idiomatic:

image

----

But the next paragraph implies it can be any suitable word 🤷‍♂️
image

moskyb: yeah, i've mostly used `ok` in the past, but `has` is fairly common in the agent codebase:
```
cache-s3-creds (~/b/agent) rg "has :="
tracetools/propagate.go
37:	s, has := env[EnvVarTraceContextKey]

clicommand/agent_start.go
698:		if _, has := tracetools.ValidTracingBackends[cfg.TracingBackend]; !has {

agent/pipeline_parser.go
65:	if tracing, has := p.Env.Get(tracetools.EnvVarTraceContextKey); has {
67:		if envMap, has := mapSliceItem("env", pipeline); has {

agent/artifact_downloader.go
178:		if _, has := s3Clients[bucketName]; !has {

bootstrap/tracing.go
44:		if _, has := os.LookupEnv("DD_TRACE_STARTUP_LOGS"); !has {
193:	if attemptStr, has := env.Get("BUILDKITE_RETRY_COUNT"); has {
200:	if parallelStr, has := env.Get("BUILDKITE_PARALLEL_JOB"); has {
206:	rebuiltFromID, has := env.Get("BUILDKITE_REBUILT_FROM_BUILD_NUMBER")
211:	triggeredFromID, has := env.Get("BUILDKITE_TRIGGERED_FROM_BUILD_ID")
216:	jobLabel, has := env.Get("BUILDKITE_LABEL")
221:	jobKey, has := env.Get("BUILDKITE_STEP_KEY")
```

and i kinda like it from a readability standpoint - i think the `val, ok := someVar.(string)` convention makes a bunch of sense because the `ok` is "did the type cast work ok", but in the case of `val, ok := someMap[key]`, it's not really whether something is ok or not, it's whether it's present.

PR #1733, opened 2022-08-01
Fix unowned plugin reference

moskyb:
We had a report on [hackerone (only available to buildkite staff unfortunately!)](https://hackerone.com/reports/1653984#), reporting that one of our plugin definitions in the plugin tests refers to a org/repo combo that doesn't belong to buildkite. 

While this **currently poses no security risk** as we don't actually clone or run the plugins as part of the tests, it's not out of the question that we might run/clone plugins as part of tests in the future, so we should change this reference to something that we own, just in case.

### DIFF
--- a/agent/plugin/plugin_test.go
+++ b/agent/plugin/plugin_test.go
@@ -63,9 +63,9 @@ func TestCreateFromJSON(t *testing.T) {
 			}},
 		},
 		{
-			`["github.com/buildkite-unofficial/ping#master"]`,
 			[]*Plugin{&Plugin{
-				Location:      `github.com/buildkite-unofficial/ping`,
+			`["github.com/buildkite-plugins/fake-plugin#master"]`,
+				Location:      `github.com/buildkite-plugins/fake-plugin`,
 				Version:       `master`,
 				Scheme:        ``,
 				Configuration: map[string]interface{}{},
--- a/agent/plugin/plugin_test.go
+++ b/agent/plugin/plugin_test.go
@@ -18,7 +18,7 @@ func TestCreateFromJSON(t *testing.T) {
 	}{
 		{
 			`[{"https://github.com/buildkite-plugins/docker-compose#a34fa34":{"container":"app"}}]`,
-			[]*Plugin{&Plugin{
+			[]*Plugin{{
 				Location:      `github.com/buildkite-plugins/docker-compose`,
 				Version:       `a34fa34`,
 				Scheme:        `https`,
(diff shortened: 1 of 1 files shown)


agent/plugin/plugin_test.go
```diff
@@ -55,25 +55,25 @@ func TestCreateFromJSON(t *testing.T) {
...
 				Version:       `a34fa34`,
 				Scheme:        `file`,
 				Configuration: map[string]interface{}{},
 			}},
 		},
 		{
-			`["github.com/buildkite-unofficial/ping#master"]`,
-			[]*Plugin{&Plugin{
-				Location:      `github.com/buildkite-unofficial/ping`,
+			`["github.com/buildkite-plugins/fake-plugin#master"]`,
```
yob, line 66: Does the agent have special behaviour for handling plugins in the `github.com/buildkite-plugins` org vs. plugins in third party GH orgs?

I assume this is OK because none of the tests are checking the special buildkite-plugins form  (`docker-compose#v3.7.0`), but I thought it was worth checking.

sj26: Yeah I think that's the purpose of these tests — they should not be in `buildkite` or `buildkite-plugins`.

moskyb: ah good call! in that case, could someone with github admin superpowers create a buildkite org that's neither buildkite nor buildkite-plugins, and then we could use that?

pda: Do we need to register another GitHub org for testing this?
Or paramerise some test things so that we can use an existing one that we do already control?

pda: > Does the agent have special behaviour for handling plugins in the github.com/buildkite-plugins org vs. plugins in third party GH orgs?

TIL this happens server-side, not agent-side.

```ruby
    case paths.length
    when 1
      "github.com/buildkite-plugins/#{add_plugin_suffix(paths.first, uri.fragment)}"
    when 2
      "github.com/#{paths.first}/#{add_plugin_suffix(paths.last, uri.fragment)}"
    else
      source
    end
```

pda: (So, I think it's fine that we use `github.com/buildkite-plugins` for testing, unless we're really worried about future-proofing it against future agent-side handling of this)

sj26: Good idea! Maybe you could use this demo org for now?

https://github.com/hashikite

tessereth (approved):
I'm going to make the executive decision to approve this and leave it up to @moskyb whether you want to change the url based on the comments.

moskyb:
i might leave it as is for now. if we move the repo-handling agentside, we can update this test.

PR #1745, opened 2022-08-12
Consume per-job agent tokens

moskyb:
Ooooh, super spicy secret new feature!

We have a new feature in the works where agents will get assigned a buildkite agent API token for each job that they run, have that token only work on operations for that job, and make that token have a limited lifetime.

This means that if a bad actor exfiltrates a buildkite API token from a job running in your buildkite account, they won't be able to use that token to mess with stuff that's not on that job, and also the token they exfiltrate will expire.

This PR does the agent-side of this feature - consuming a token sent by the `jobs#accept` response, and using that token for operations within the job.

We're still working on the backend side of this work, and we won't merge this until that part of the work is done.

### DIFF
--- a/agent/agent_worker.go
+++ b/agent/agent_worker.go
@@ -493,12 +493,12 @@ func (a *AgentWorker) AcceptAndRunJob(job *api.Job) error {
 	return a.RunJob(accepted)
 }
 
-func (a *AgentWorker) RunJob(job *api.Job) error {
+func (a *AgentWorker) RunJob(acceptResponse *api.Job) error {
 	jobMetricsScope := a.metrics.With(metrics.Tags{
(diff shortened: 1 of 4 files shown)


agent/agent_worker.go
```diff
@@ -493,12 +493,12 @@ func (a *AgentWorker) AcceptAndRunJob(job *api.Job) error {
 	return a.RunJob(accepted)
 }
 
-func (a *AgentWorker) RunJob(job *api.Job) error {
+func (a *AgentWorker) RunJob(acceptResponse *api.Job) error {
```
tessereth, line 496: If one was being pedantic, this could also be an `acquireResponse`. But they have exactly the same format and I still agree this name is better than just `job`.

tessereth (approved):
Looks great

lox:
Ermagerd.

PR #1744, opened 2022-08-12
Update buildkite agent plugin repo cloning for gitlab

brianbentson:
This is a re-creation of https://github.com/buildkite/agent/pull/1641 because "re-opens" are disabled for normal users.

PR #1746, opened 2022-08-14
Fix multiple-nested plugin repos on gitlab

moskyb:
**Context:** https://github.com/buildkite/agent/pull/1641, https://github.com/buildkite/agent/pull/1744

**🤔 Problem:** Currently, when cloning plugins from Gitlab, we assume that the repo path structure is the same as github (that is, `org/repo`), and consider anything not matching that path to be an invalid plugin reference. The trouble is, in Gitlab, there can be multiple levels of nesting in a repo path, with the pattern `some-org/some/hierarchy/of/teams/repo` - for example `https://gitlab.com/example-corp/devops/aws/lambda/example-buildkite-plugin`. If one of these multi-nested repos is used as a BK plugin reference, we'll erroneously say that it's an invalid plugin path, and fail the pipeline.

**✅ Solution:** This PR adds an additional branch to our plugin reference checking gear, allowing any level of nesting if the plugin URL starts with `gitlab.com`.

Note that this issue doesn't affect privately-hosted gitlab instances, which weren't affected by the buggy behaviour. These already worked, and should still work.

Many thanks to @brianbentson for writing the actual code in this PR! There were some git shenanigans that required me to pull it into a non-forked branch, but this is entirely their work.

### DIFF
--- a/agent/plugin/plugin.go
+++ b/agent/plugin/plugin.go
@@ -279,7 +279,6 @@ func (p *Plugin) Label() string {
 		return p.Location
 	}
 }
-
 func (p *Plugin) constructRepositoryHost() (string, error) {
 	if p.Location == "" {
 		return "", fmt.Errorf("Missing plugin location")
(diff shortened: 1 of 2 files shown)


tessereth (approved):
This all looks great to me, thanks for the contribution @brianbentson!

brianbentson:
Many thanks! Out of curiosity how long until this ends up in a release?

moskyb:
without putting any guarantees on it, probably soonish? we tend to have a monthly-ish release cycle, with heavy emphasis on the 'ish' 😅

PR #1753, opened 2022-08-26
CI: bundle update for CI ruby dependencies

pda:
The CI/CD pipeline for this repository uses some Ruby components to manage `deb`/`rpm` packages.

This pull request updates the gems using `bundle update`. There's only two top-level dependencies (`fpm` and `deb-s3`) but this mostly updates transitive dependencies.

We'll probably only know _for sure_ about unexpected breaking changes after merging to `main` and running a release, but if it's broken we can revert easily enough.


gchan (approved):
> We'll probably only know for sure about unexpected breaking changes after merging to main and running a release, but if it's broken we can revert easily enough.

Sounds good that we can revert if there are any issues

Some change logs

git 
https://github.com/ruby-git/ruby-git/releases/tag/v1.12.0

jmespath
https://github.com/jmespath/jmespath.rb/blob/main/CHANGELOG.md#161-2022-03-07

dotenv
https://github.com/bkeepers/dotenv/blob/master/Changelog.md

pda:
Thanks @gchan!

PR #1761, opened 2022-09-02
Retry Disconnect API calls

pda:
We've had some situations where the Agent Disconnect API call fails (e.g. HTTP 500 or 502), which leaves the agent in a connected state on the server side; this is normally cleared within 5 minutes, but in some situations could take longer. During that time, jobs may be assigned to the “ghost” agent, preventing them actually running on another connected agent.

This pull request retries the Disconnect API call up to 4 times, with a constant 1 second delay between tries. This keeps the total duration under 4-5 seconds; disconnection should remain fast as part of agent shutdown, even if the API fails.

This PR also introduces test coverage for `AgentWorker.Disconnect`, both without and with failure retries.

Alongside this, I've opened an issue at https://github.com/buildkite/roko/issues/2 so that roko's sleep function can be set globally during tests, without threading it through the code under test.

### DIFF
--- a/agent/agent_worker.go
+++ b/agent/agent_worker.go
@@ -83,6 +83,11 @@ type AgentWorker struct {
 	// When this worker runs a job, we'll store an instance of the
 	// JobRunner here
 	jobRunner *JobRunner
+
+	// retrySleepFunc is useful for testing retry loops fast
+	// Hopefully this can be replaced with a global setting for tests in future:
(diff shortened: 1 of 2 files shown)


jordandcarter (approved):
Looks good, nice use of retrySleepFunc in the test 👌

tessereth (approved):

PR #1767, opened 2022-09-21
build agent binaries for windows/arm64

### LINKED ISSUE #1282: [Feature Request] buildkite agent for windows arm64 (Windows on ARM)
ie.  https://docs.microsoft.com/en-us/windows/arm/ <3

yob:
This adds arm64 windows binaries to our official build matrix. Originally requested in #1282, but at the time we were nervous about enabling because:

* none of us had access to windows/arm64 systems
* it required a golang 1.17+, and we were compiling the agent with 1.16

These days we're compiling the agent with golang 1.18 so windows/arm64 is a target that Just Works. An arm64 windows environment is also super easy to boot using parallels 18 on an Apple Silicon machine, and we have plenty of those.

I tested this by manually cross compiling compiling a windows/arm64 binary on my linux laptop and then using it inside an arm parallels VM.

Compilation on Linux:

    GOOS=windows GOARCH=arm64 go build -o buildkite-agent-win-arm64 .

On buildkite:

Fixes #1282

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -107,9 +107,6 @@ steps:
           - with: { os: openbsd, arch: arm64 }
             skip: "arm64 OpenBSD is not currently supported"
 
-          - with: { os: windows, arch: arm64 }
-            skip: "arm64 Windows is not currently supported"
-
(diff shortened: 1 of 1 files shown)


pda (approved):
Thanks!

PR #1771, opened 2022-09-27
Create alpine k8s image

dabarrell:
Creates a version of the alpine-linux docker image that includes kubectl and kustomize, for use in kubernetes agent stacks, or orchestrate pods.

One design choice that could go the other way is that it pulls the most recent stable version of both binaries. Pinning versions would make the dockerfile a little simpler, but would require more upkeep.

It uses a multi-stage build to separate the download of the binaries in order to minimise side-effects that come from curling + executing a bash install script from `kubernetes-sigs/kustomize`. This script is the simplest way to pull the most recent stable version (otherwise we have to parse the releases ourselves). Another option is to create a local version of this script (https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh) that we can strip back. Or pin the script using a commit, rather than `master`.


pda (approved):

sj26:
```
$ docker pull 445615400570.dkr.ecr.us-east-1.amazonaws.com/agent:alpine-k8s-build-4579
alpine-k8s-build-4579: Pulling from agent
530afca65e2e: Pull complete
a0cb4edaf23f: Pull complete
96e48acfbd1c: Pull complete
4b8a8ac45ae9: Pull complete
b69cc3264740: Pull complete
5a216eaf465d: Pull complete
97abcf8c5665: Pull complete
6375d0e02512: Pull complete
Digest: sha256:8a5ed7802bc8d555d1b39cff921a6fbcf530411f444f55475a5a78192cbac83a
Status: Downloaded newer image for 445615400570.dkr.ecr.us-east-1.amazonaws.com/agent:alpine-k8s-build-4579
445615400570.dkr.ecr.us-east-1.amazonaws.com/agent:alpine-k8s-build-4579

$ docker run --rm -it --entrypoint sh 445615400570.dkr.ecr.us-east-1.amazonaws.com/agent:alpine-k8s-build-4579

/ # kubectl version
WARNING: This version information is deprecated and will be replaced with the output from kubectl version --short.  Use --output=yaml|json to get the full version.
Client Version: version.Info{Major:"1", Minor:"25", GitVersion:"v1.25.2", GitCommit:"5835544ca568b757a8ecae5c153f317e5736700e", GitTreeState:"clean", BuildDate:"2022-09-21T14:33:49Z", GoVersion:"go1.19.1", Compiler:"gc", Platform:"linux/amd64"}
Kustomize Version: v4.5.7
The connection to the server localhost:8080 was refused - did you specify the right host or port?
```

Seems legit!

sj26 (approved):

packaging/docker/alpine-linux-k8s/Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+COPY --from=downloader /downloads/kubectl /usr/local/bin/kubectl
+COPY --from=downloader /downloads/kustomize /usr/local/bin/kustomize
+
+ENV BUILDKITE_AGENT_CONFIG=/buildkite/buildkite-agent.cfg
+
+RUN mkdir -p /buildkite/builds /buildkite/hooks /buildkite/plugins \
+    && curl -Lfs -o /usr/local/bin/ssh-env-config.sh https://raw.githubusercontent.com/buildkite/docker-ssh-env-config/master/ssh-env-config.sh \
+    && chmod +x /usr/local/bin/ssh-env-config.sh
+
+COPY ./buildkite-agent.cfg /buildkite/buildkite-agent.cfg
```
thephw, line 40: Because of how we mount the ConfigMap volumes in kubernetes it'd be useful if the config file and environment variable were at:
`/buildkite/config/buildkite-agent.cfg`

The volume mounts basically expect to own everything in the directory it mounts to and because `/buildkite` already has subdirectories for `/buildkite/builds`, `/buildkite/hooks`, and `/buildkite/plugins` we override the config file location in the agent deployment configuration. We could continue to do it in the deployment configuration -- but it might make more sense to do it here in the stack-specific agent configuration.

sj26: Good idea! But I reckon we can do that as a next step.

Issue #1663, opened 2022-06-06
Propegate TraceID and SpanID to steps


ojkelly:
**Is your feature request related to a problem? Please describe.**
The new OpenTelemetry (OTEL) experiment is helpful, but it does not make it possible to get the current traceId and spanId.

If we could get access to these, we can then link our OTEL instrumentation inside each step with the trace that exists for the whole job.

**Describe the solution you'd like**
Extend the existing `BUILDKITE_TRACE_CONTEXT` found in https://github.com/buildkite/agent/blob/848a053b3131ebced93d829c3215b31bddd19f72/tracetools/propagate.go to each step, when a tracing backend is used.

**Describe alternatives you've considered**
I've hacked around the codebase to enable access to TraceID and SpanID as individual variables, but the OTEL libraries are all focussed around propagating from a single trace context.

**Additional context**
This helps with getting a deeper understanding of what's happening over the course of a whole job. And helps with linking telemetry from each step together.

I'm willing to work on a PR up for this, if it's likely to be merged.

ojkelly:
I've found a very relevant issue from the Otel repo https://github.com/open-telemetry/opentelemetry-specification/issues/740.

It goes into a bit of depth about other CI tooling, and how they've mostly settled on using `TRACEPARENT` and `TRACECONTEXT` as the envars to share trace information.

moskyb:
hey @ojkelly! this is definitely something that we're looking to work on - it was intentionally left out of the initial release so that we could get something out, but we'll definitely be looking to include it in the near future. There's some internal work that i'm doing that's ahead of trace propagation in the queue, but it will be coming soon.

I've also been across the issue you linked - whatever we end up doing, it'll probably end up being complaint with the (non-) standard laid out there.

ozdenyilmaz:
Hi everyone,

I am from the product team at Buildkite. Unfortunately this is still in backlog for us, so it will be a while before we pick this piece of work. The team is working on security work like signed agents that will delay this being from picked up. Sorry about that.

jlisee:
I just saw this it's really cool!  This open up this question now :) to get all tracing and span work in one place:
- https://github.com/buildkite/agent/issues/4180

PR #1775, opened 2022-10-04
Use base64+json for trace context encoding

goodspark:
Back when I added trace propagation to the agent, I was short-sighted and used the golang-binary format. Since the trace context was meant to be propagated between builds of a pipeline, it was safe to assume the consumers would always be able to decode golang-binary formats.

But now I want to trace additional spans within non-Golang code and link it to the parent trace of the pipeline build. This will let me gather hollistic metrics about specific failure modes and performance of various commands/tooling running in my repos.

So I picked JSON, since it's a format all modern languages can easily parse without the need for a special library.

TODOs
- [x] Tests
- [x] Test a build and verify traces go through and trace context is properly passed to subprocesses

### DIFF
--- a/tracetools/propagate.go
+++ b/tracetools/propagate.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/base64"
 	"encoding/gob"
+	"encoding/json"
 
 	"github.com/opentracing/opentracing-go"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
@@ -15,15 +16,21 @@ const EnvVarTraceContextKey = "BUILDKITE_TRACE_CONTEXT"
 
 // EncodeTraceContext will serialize and encode tracing data into a string and place
 // it into the given env vars map.
-func EncodeTraceContext(span opentracing.Span, env map[string]string) error {
+func EncodeTraceContext(span opentracing.Span, env map[string]string, useJsonTraceContext bool) error {
 	textmap := tracer.TextMapCarrier{}
 	if err := span.Tracer().Inject(span.Context(), opentracing.TextMap, &textmap); err != nil {
 		return err
 	}
 
 	buf := bytes.NewBuffer([]byte{})
-	enc := gob.NewEncoder(buf)
-	if err := enc.Encode(textmap); err != nil {
+	// Unfortunately there's no general Encoder interface so we can't DRY the Encode call.
+	var err error
+	if useJsonTraceContext {
+		err = json.NewEncoder(buf).Encode(textmap)
+	} else {
+		err = gob.NewEncoder(buf).Encode(textmap)
+	}
+	if err != nil {
 		return err
 	}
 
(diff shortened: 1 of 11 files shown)


tracetools/propagate.go
```diff
@@ -22,8 +22,7 @@ func EncodeTraceContext(span opentracing.Span, env map[string]string) error {
 	}
 
 	buf := bytes.NewBuffer([]byte{})
-	enc := gob.NewEncoder(buf)
-	if err := enc.Encode(textmap); err != nil {
+	if err := json.NewEncoder(buf).Encode(textmap); err != nil {
```
goodspark, line 25: Switching immediately means there will be a period when people might lose trace linking (but not traces) while they roll out the new agent across their CI fleet. Seemed simpler than trying to plumb through all the env vars and configs, which was what I started with but was having trouble getting working.

tracetools/propagate.go
```diff
@@ -3,7 +3,7 @@ package tracetools
 import (
```
goodspark, line 3: It'll also be interesting to see if we can use the proposed env vars in https://github.com/open-telemetry/opentelemetry-specification/issues/740 (which is also mentioned in https://github.com/buildkite/agent/issues/1663) as a move towards more standardization. This will help others reusing libraries native to whatever language/tool they're using while still maintaining trace linking across process boundaries.

tracetools/propagate.go
moskyb: so i'm not an opentracing expert, but the value of the `BUILDKITE_TRACE_CONTEXT` env var is something that downstream (ie, non-buildkite agent) processes have to decode this to get a trace context, right? it's not just an opaque value?

if this is the case, would the fact that we're switching encodings here means that tracing setups that previously worked would have to change their decoding strategy? if so, this would be a breaking change, and i don't think we'd be able to ship it.

there are other options, i think - we could expose a second env var like `BUILDKITE_TRACE_CONTEXT_BASE64` or something. We could also do the other thing and wire up all the CLI flags and configs and such, though i agree with you that that's not a super great solution; not least of which because it's a pain to implement, but it's also good to keep the number of options on the agent down where we can.

moskyb (commented):
the code of this looks great, but i have some concerns about the backward-compatibility of it. Provided we can sort those out, i'd love to get this into the agent - more, more normal encoding methods is a good thing!

tracetools/propagate.go
goodspark: I wound up adding an arg and plumbing it through. Turns out the code I had before was mostly fine, I just had typos that was breaking some agent-to-bootstrap plumbing.

agates4:
I'm excited about this PR - it would help us send additional traces through to Datadog!

goodspark:
This has been sitting around for months. Any chance someone from Buildkite can review this?

smocherla-brex:
Hey folks, is there anything we could to merge this PR? This would be really useful to add custom trace spans in our commands running in CI.

PR #1777, opened 2022-10-06
retry 5xx responses when attempting to start a job

jonahbull:
We occasionally see 502 responses from the Agent API when starting a job:
```
Failed to run job: PUT https://agent.buildkite.com/v3/jobs/0183a95c-c17e-4b0f-8058-e405362c6dc7/start: 502
```
Since those responses likely indicate a transient issue, I think it makes sense to attempt to retry those and other 5xx responses for this call rather than failing and having the job be re-scheduled. Open to any suggestions or input y'all might have.

### DIFF
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -683,20 +683,22 @@ func (r *JobRunner) executePreBootstrapHook(hook string) (bool, error) {
 }
 
 // Starts the job in the Buildkite Agent API. We'll retry on connection-related
-// issues, but if a connection succeeds and we get an error response back from
+// issues, but if a connection succeeds and we get an client error response back from
 // Buildkite, we won't bother retrying. For example, a "no such host" will
-// retry, but a 422 from Buildkite won't.
+// retry, but an HTTP response from Buildkite that isn't retryable won't.
 func (r *JobRunner) startJob(startedAt time.Time) error {
 	r.job.StartedAt = startedAt.UTC().Format(time.RFC3339Nano)
 
 	return roko.NewRetrier(
-		roko.WithMaxAttempts(30),
-		roko.WithStrategy(roko.Constant(5*time.Second)),
+		roko.WithMaxAttempts(7),
+		roko.WithStrategy(roko.Exponential(2*time.Second, 0)),
 	).Do(func(rtr *roko.Retrier) error {
-		_, err := r.apiClient.StartJob(r.job)
+		response, err := r.apiClient.StartJob(r.job)
 
 		if err != nil {
-			if api.IsRetryableError(err) {
+			if response != nil && api.IsRetryableStatus(response) {
+				r.logger.Warn("%s (%s)", err, rtr)
+			} else if api.IsRetryableError(err) {
 				r.logger.Warn("%s (%s)", err, rtr)
 			} else {
 				r.logger.Warn("Buildkite rejected the call to start the job (%s)", err)
--- a/api/retryable.go
+++ b/api/retryable.go
@@ -3,9 +3,12 @@ package api
 import (
 	"io"
 	"net"
+	"net/http"
 	"net/url"
 	"strings"
 	"syscall"
+
+	"golang.org/x/exp/slices"
 )
 
 var retrableErrorSuffixes = []string{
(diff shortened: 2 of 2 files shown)


agent/job_runner.go
```diff
@@ -683,20 +683,22 @@ func (r *JobRunner) executePreBootstrapHook(hook string) (bool, error) {
...
+// issues, but if a connection succeeds and we get an client error response back from
 // Buildkite, we won't bother retrying. For example, a "no such host" will
-// retry, but a 422 from Buildkite won't.
+// retry, but an HTTP >= 400 <= 500 response from Buildkite won't.
 func (r *JobRunner) startJob(startedAt time.Time) error {
 	r.job.StartedAt = startedAt.UTC().Format(time.RFC3339Nano)
 
 	return roko.NewRetrier(
 		roko.WithMaxAttempts(30),
 		roko.WithStrategy(roko.Constant(5*time.Second)),
```
moskyb, line 694: If we're going to be retrying when we get 5xx errors, that might be when buildkite is experiencing an outage. we should probably change this to be a [`roko.Exponential`](https://pkg.go.dev/github.com/buildkite/roko#Exponential) strategy so that we don't accidentally nuke the buildkite backend when it's already having a bad day.

api/retryable.go
```diff
@@ -18,6 +19,23 @@ var retrableErrorSuffixes = []string{
...
+}
+
+// IsRetryableStatus returns true if the response's StatusCode is one that we should retry.
+func IsRetryableStatus(r *Response) bool {
+	for _, status := range retryableStatuses {
+		if r.StatusCode == status {
+			return true
+		}
+	}
+	return false
```
moskyb, line 36: [super nit] we have the magic of generics now, so we could replace this with
```suggestion
	slices.Contains(retryableStatuses, status)
```

if you really wanted to, but this is perfectly fine as is.

moskyb (commented):
i love this! i've bikeshedded the review a tad, which i'm really sorry about, but this would be an awesome change to bring into the agent.

agent/job_runner.go
jonahbull: Ah, good call. I've switched to the `roko.Exponential` strategy in the latest commit. Do you think 30 attempts with a 5 second exponential backoff is too much?

api/retryable.go
```diff
@@ -18,6 +19,23 @@ var retrableErrorSuffixes = []string{
 	io.EOF.Error(),
 }
 
+var retryableStatuses = []int{
+	http.StatusInternalServerError, // 500
+	http.StatusBadGateway,          // 502
+	http.StatusServiceUnavailable,  // 503
+	http.StatusGatewayTimeout,      // 504
+}
```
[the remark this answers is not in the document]
jonahbull: Good idea, added.

api/retryable.go
jonahbull: :+1: I've embraced the magic of generics in the latest commit.

agent/job_runner.go
```diff
@@ -683,20 +683,22 @@ func (r *JobRunner) executePreBootstrapHook(hook string) (bool, error) {
...
 	).Do(func(rtr *roko.Retrier) error {
-		_, err := r.apiClient.StartJob(r.job)
+		response, err := r.apiClient.StartJob(r.job)
 
 		if err != nil {
-			if api.IsRetryableError(err) {
+			if response != nil && api.IsRetryableStatus(response) {
+				r.logger.Warn("%s (%s)", err, rtr)
```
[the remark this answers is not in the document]
jonahbull: I could be wrong, but it looks like there's some response wrapping that happens in `client.doRequest`
```go
	err = checkResponse(resp)
	if err != nil {
		// even though there was an error, we still return the response
		// in case the caller wants to inspect it further
		return response, err
	}
```
so if it's a non-2xx response we can expect err to be non-nil.

api/retryable.go
moskyb: ✨

agent/job_runner.go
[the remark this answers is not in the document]
moskyb: aaah true! good call, i should've read the code 😅

agent/job_runner.go
moskyb: i've done a bit of math, and if we did 30 attempts with a 5 second backoff, the the delay between try 29 and try 30 would be 29 Trillion Years, which feels maybe a bit too long 😅

I think 
```Go
		roko.WithMaxAttempts(7),
		roko.WithStrategy(roko.Exponential(2*time.Second, 0)),
```

which will wait 255s (~= 4min) total is a good balance - we wait a bit longer than the current constant delay, and try fewer times.

jonahbull: 😅 yes that's a bit much, I adjusted it downward as you suggested

moskyb:
awesome work @jonahbull! thank so much ^_^

jonahbull:
Thanks @moskyb! Appreciate your quick and helpful feedback 🙏🏻

Issue #1770, opened 2022-09-22
Feature Request: Agent startup hook.


donallfg:
**Is your feature request related to a problem? Please describe.**
I want to run some code on agent startup, and on agent shutdown to acquire/release storage resources. There is already an agent-shutdown hook that is executed that we can use for releasing resources, but we would like a startup hook to acquire them.

**Describe the solution you'd like**
https://buildkite.com/docs/agent/v3/hooks#agent-lifecycle-hooks - An additional agent lifecycle hook, `agent-startup` that gets triggered during the agent startup process.

**Describe alternatives you've considered**
Job lifecycle hooks aren't suitable because they run on every job. 
We oculd wrap the buildkite-agent start in our custom code, and launch from there, but we need to deploy code to do that, _and_ code to handle the shutdown in that case too
We could use infrastructure lifecycle hooks, e.g. https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/autoscaling_lifecycle_hook which will likely be our fallback.

moskyb:
🤔 i think i quite like this idea. we currently have [something similar](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/master/packer/linux/conf/bin/bk-install-elastic-stack.sh#L209-L213) in the elastic-stack called (confusingly enough) the bootstrap script, which does something similar, but outside the bounds of the agent itself.

I think that given the existence of the `pre-exit` hook, it probably makes sense to have something symmetrical for agent startup. I'll raise this with the team internally :)

donalmacc:
Hey. Original requester here (although on my personal account rather than my work account.) I took a stab at implementing this last night, and sent a PR https://github.com/buildkite/agent/pull/1778

moskyb:
amazing, thanks so much @donalmacc! awesome work, i'll get reviewing :)

PR #1778, opened 2022-10-11
Add an agent-startup hook that is fired at the same time as the agent…

donalmacc:
…-shutdown hook is registered.

Utilise the agent-shutdown hook logic for both of the "agent lifecycle" hooks, except if the hook fails shutdown the agent .

Implement the same test suite for agent startup and agent shutdown, as the same rules apply.

Closes: https://github.com/buildkite/agent/issues/1770

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -894,6 +894,11 @@ var AgentStartCommand = cli.Command{
 		// Agent-wide shutdown hook. Once per agent, for all workers on the agent.
 		defer agentShutdownHook(l, cfg)
 
+		// Once the shutdown hook has been setup, trigger the startup hook.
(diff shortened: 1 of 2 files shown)


donalmacc:
Hi, wondering if there's been any update here?

moskyb:
argh sorry @donalmacc! i was reviewing this and then got distracted by something.... a month ago. this all looks really good, but lemme get my reviewer glasses back on and give this a proper look.

again, really sorry that this slipped through the cracks

moskyb (approved):

donallfg:
No need to apologise at all, I totally understand the practicalities. Thanks for taking the fix, and I'm looking forward to updating our plugin to use these hooks rather than our current shim :)

Issue #1780, opened 2022-10-12
Wrong environment variable parsing with ANSI-C Quoting


morxa:
We set secrets in an environment hook (following [this guide](https://buildkite.com/docs/pipelines/secrets)).

One of the secrets is an [ANSI-C quoted string](https://www.gnu.org/software/bash/manual/html_node/ANSI_002dC-Quoting.html), e.g., the following:

```
MYUSERNAME="user"
export MYUSERNAME
MYSECRETKEY=$'this\nis\na\nkey'
export MYSECRETKEY
```

The buildkite agent parses the environment variable `MYSECRETKEY` incorrectly. The output of the build (_Running global environment hook_ to be more precise) shows:
```
# MYSECRETKEY=$'this\nis\na\nkey' added
```
As you can see, the string is treated as part of the environment variable name, rather than as its value. **This causes secrets to be leaked to anyone who can access the build log.**

[Here](https://buildkite.com/fawkesrobotics/fawkes-build/builds/1973#0183cd72-f23e-40ac-aaa4-b38abe46b8fb) is a full example. The environment hook is the following script:
```
#!/usr/bin/env bash

set -eu

HOST=
ORG=
REPO=

MYUSERNAME="user"
export MYUSERNAME
MYSECRETKEY=$'this\nis\na\nkey'
export MYSECRETKEY

export
```

If you open _Running global environment hook_, you can see the output and which variables are exported. I added an `export` to the end of the script to debug the environment variables. I also added another `export` to the script that uses the environment variable, which was set by the buildkite agent (IIUC). Here, the secret is incorrectly escaped:
```
declare -x MYSECRETKEY="\$'this\\nis\\na\\nkey'="
```

This is actually how we found the issue, because the secret was no  longer working.

Looking at the agent code, the following assumption seems to be invalid:
https://github.com/buildkite/agent/blob/765795ecb61d68cd1b470f6593fc9f79c67027c9/env/export.go#L128-L132

The buildkite agent version is v3.39.0.

sj26:
Thanks for the details report, @morxa.

We had trouble replicating this until we upgraded to bash 5.2, which was [released about 16 days ago](https://lists.gnu.org/archive/html/info-gnu/2022-09/msg00012.html). Are you using bash 5.2?

moskyb:
hey @morxa! thanks for reporting this, this seems like a pretty ugly bug to come across, and this description of the error is really well written.

[...]

PR #1781, opened 2022-10-13
[WIP] Replace export -p with a dedicated agent command for printing env

moskyb:
Issue: #1780

### DIFF
--- a/hook/scriptwrapper.go
+++ b/hook/scriptwrapper.go
@@ -230,6 +230,13 @@ func (wrap *ScriptWrapper) Changes() (HookScriptChanges, error) {
 		return HookScriptChanges{}, fmt.Errorf("Failed to read \"%s\" (%s)", wrap.afterEnvFile.Name(), err)
 	}
 
+	// An empty afterEnvFile indicates that the hook early-exited from within the
+	// ScriptWrapper, so the working directory and environment changes weren't
+	// captured.
+	if len(afterEnvContents) == 0 {
+		return HookScriptChanges{}, &HookExitError{hookPath: wrap.hookPath}
+	}
+
 	var (
 		beforeEnv env.Environment
 		afterEnv  env.Environment
@@ -249,10 +256,6 @@ func (wrap *ScriptWrapper) Changes() (HookScriptChanges, error) {
 		if err != nil {
 			return HookScriptChanges{}, fmt.Errorf("failed to unmarshal after env file: %w, file contents: %q", err, string(afterEnvContents))
 		}
-	}
-
-	if afterEnv.Length() == 0 {
-		return HookScriptChanges{}, &HookExitError{hookPath: wrap.hookPath}
 	}
 
 	diff := afterEnv.Diff(beforeEnv)
--- a/hook/scriptwrapper.go
+++ b/hook/scriptwrapper.go
@@ -2,8 +2,8 @@ package hook
 
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -20,26 +20,26 @@ const (
 
 	batchScript = `@echo off
 SETLOCAL ENABLEDELAYEDEXPANSION
-SET > "{{.BeforeEnvFileName}}"
+buildkite-agent env > "{{.BeforeEnvFileName}}"
 CALL "{{.PathToHook}}"
 SET BUILDKITE_HOOK_EXIT_STATUS=!ERRORLEVEL!
 SET BUILDKITE_HOOK_WORKING_DIR=%CD%
-SET > "{{.AfterEnvFileName}}"
+buildkite-agent env > "{{.AfterEnvFileName}}"
 EXIT %BUILDKITE_HOOK_EXIT_STATUS%`
 
 	powershellScript = `$ErrorActionPreference = "STOP"
-Get-ChildItem Env: | Foreach-Object {"$($_.Name)=$($_.Value)"} | Set-Content "{{.BeforeEnvFileName}}"
+buildkite-agent env | Set-Content "{{.BeforeEnvFileName}}"
 {{.PathToHook}}
 if ($LASTEXITCODE -eq $null) {$Env:BUILDKITE_HOOK_EXIT_STATUS = 0} else {$Env:BUILDKITE_HOOK_EXIT_STATUS = $LASTEXITCODE}
 $Env:BUILDKITE_HOOK_WORKING_DIR = $PWD | Select-Object -ExpandProperty Path
-Get-ChildItem Env: | Foreach-Object {"$($_.Name)=$($_.Value)"} | Set-Content "{{.AfterEnvFileName}}"
+buildkite-agent env | Set-Content "{{.AfterEnvFileName}}"
 exit $Env:BUILDKITE_HOOK_EXIT_STATUS`
 
-	bashScript = `export -p > "{{.BeforeEnvFileName}}"
+	bashScript = `buildkite-agent env > "{{.BeforeEnvFileName}}"
 . "{{.PathToHook}}"
 export BUILDKITE_HOOK_EXIT_STATUS=$?
 export BUILDKITE_HOOK_WORKING_DIR=$PWD
-export -p > "{{.AfterEnvFileName}}"
+buildkite-agent env > "{{.AfterEnvFileName}}"
 exit $BUILDKITE_HOOK_EXIT_STATUS`
 )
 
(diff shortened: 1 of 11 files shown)


hook/scriptwrapper.go
```diff
@@ -20,26 +20,26 @@ const (
...
 EXIT %BUILDKITE_HOOK_EXIT_STATUS%`
 
 	powershellScript = `$ErrorActionPreference = "STOP"
-Get-ChildItem Env: | Foreach-Object {"$($_.Name)=$($_.Value)"} | Set-Content "{{.BeforeEnvFileName}}"
+buildkite-agent env | Set-Content "{{.BeforeEnvFileName}}"
 {{.PathToHook}}
 if ($LASTEXITCODE -eq $null) {$Env:BUILDKITE_HOOK_EXIT_STATUS = 0} else {$Env:BUILDKITE_HOOK_EXIT_STATUS = $LASTEXITCODE}
 $Env:BUILDKITE_HOOK_WORKING_DIR = $PWD | Select-Object -ExpandProperty Path
-Get-ChildItem Env: | Foreach-Object {"$($_.Name)=$($_.Value)"} | Set-Content "{{.AfterEnvFileName}}"
+buildkite-agent env | Set-Content "{{.AfterEnvFileName}}"
```
moskyb, line 35: The windows scripts are untested as yet; smoke testing the agent on windows is a bit of a pain when i only have a mac dev box, but we'll get this sorted prior to merge

pda:
I wonder about putting `map[string]string` between the `os.Environ()` output and the JSON representation 🤔

Go's non-deterministic ordering of `map` makes the output unstable; harder to test / diff / etc. I think at a minimum we should make the ordering stable, somehow. Unless I'm misremembering, and Go `map`s or JSON serialization thereof is already stable, but I don't think so.

More subtle: buildkite-agent is run in diverse environments (pun unintended) by people doing diverse things. AFAIK Unix/POSIX/computerey “environment” is loosely defined. At a code level, Go's `os.Environ()`, Linux's `environ()` call, and other implementations all agree that environment is a list of strings. But what can we assume about the strings?

[Linux says](https://man7.org/linux/man-pages/man7/environ.7.html):

> By convention, the strings in environ have the form "name=value".
> The name is case-sensitive and may not contain the character "=".
> The value can be anything that can be represented as a string.
> The name and the value may not contain an embedded null byte
> ('\0'), since this is assumed to terminate the string.

[POSIX.1-2017 says](https://pubs.opengroup.org/onlinepubs/9699919799/)…

> These strings have the form `name=value`; names shall not contain the character '='. For values to be portable across systems conforming to POSIX.1-2017, the value shall be composed of characters from the portable character set (except NUL and as indicated below). There is no meaning associated with the order of strings in the environment. If more than one string in an environment of a process has the same name, the consequences are undefined.

So Linux suggests that strings are only `name=value` by convention, implying they might be something different.
And neither Linux nor POSIX are specific about what happens if there's duplicate `name`s.

I guess I'm wondering if `buildkite-agent env` should serialize to:
- an object like `{"name": "value", …}` object, implicitly deduped from `os.Environ()`…
    - in random order via Go's `map[string]string`,
    - in defined order e.g. sorted by `name`,
    - in the order they came out of `os.Environ()`,
- an array of (name, value) tuples like `[[name, value], …]` after splitting each line by first `=`…
    - in defined order e.g. sorted by `name`,
    - in the order they came out of `os.Environ()`,
- an array like `["name=value", …]`
    - verbatim from `os.Environ()`

I would lean towards either:
- `[[name, value], …]` in the order that `os.Environ()` provided.
- `{"name": "value", …}` sorted by `name`.

The former gives fairly useful uncontroversial/unopinionated lossless representation of the environment as provided by the OS. The latter gives a slightly more useful shape but is slightly more opinionated re. deduping and ordering.

hook/scriptwrapper.go
```diff
@@ -20,26 +20,26 @@ const (
...
 exit $Env:BUILDKITE_HOOK_EXIT_STATUS`
 
-	bashScript = `export -p > "{{.BeforeEnvFileName}}"
+	bashScript = `buildkite-agent env > "{{.BeforeEnvFileName}}"
```
pda, line 38: Does this introduce a dependency on (the correct / currently-used version of) `buildkite-agent` being in `$PATH`? Or do we already do that elsewhere? Might need to provide an absolute path or something?

I'm wondering what happens if I `go build -o buildkite-agent && buildkite-agent start` on a system that has an older version installed in `$PATH`… I think it'll be surprising if the running version calls the older version, especially if the older version doesn't _have_ the `env` subcommand.

hook/scriptwrapper.go
```diff
@@ -220,18 +220,29 @@ func (wrap *ScriptWrapper) Close() {
 
 // Changes returns the changes in the environment and working dir after the hook script runs
 func (wrap *ScriptWrapper) Changes() (HookScriptChanges, error) {
-	beforeEnvContents, err := ioutil.ReadFile(wrap.beforeEnvFile.Name())
+	beforeEnvContents, err := os.ReadFile(wrap.beforeEnvFile.Name())
```
pda, line 223: Ah I see [`ioutil.ReadFile` is deprecated](https://pkg.go.dev/io/ioutil@go1.19.2#ReadFile) 👍🏼

moskyb: all of ioutil is apparently! i learned this yesterday. we should probably go through and replace things, but the compatibility guarantee means that we probably won't have to until \*checks watch\* ever

hook/scriptwrapper.go
moskyb: So looking at other parts of the bootstrap, we definitely do depend on the agent being on the PATH:
https://github.com/buildkite/agent/blob/main/bootstrap/bootstrap.go#L1866-L1868

but i think you're right to question this, and we should probably stay away from assuming the agent is on the path in the future, it'll only cause trouble when testing things locally.

We should be able to get away with using `os.Executable()` for this with no worries

moskyb:
re: ordering in the environment, when serializing maps, `json.Marshal()` sorts the maps by key alphabetically: https://pkg.go.dev/encoding/json#Marshal:
> Map values encode as JSON objects. The map's key type must either be a string, an integer type, or implement encoding.TextMarshaler. The map keys are sorted and used as JSON object keys

So we get ordering for free! that's nice. I verified this myself:
```
$ AARDVARK="yep!" ZEBRA="nope!" buildkite-agent env --pretty
{
  "AARDVARK": "yep!",
  "BUILDKITE_API_TOKEN": "nice try!",
  # ...
  "ZEBRA": "nope!",
  "__CFBundleIdentifier": "com.googlecode.iterm2",
  "__CF_USER_TEXT_ENCODING": "0x1F5:0:15"
}
```

moskyb:
re: ordering in the environment, when serializing maps, `json.Marshal()` sorts the maps by key alphabetically: https://pkg.go.dev/encoding/json#Marshal:
> Map values encode as JSON objects. The map's key type must either be a string, an integer type, or implement encoding.TextMarshaler. The map keys are sorted and used as JSON object keys

So we get ordering for free! that's nice. I verified this myself:
```
$ AARDVARK="yep!" ZEBRA="nope!" buildkite-agent env --pretty
{
  "AARDVARK": "yep!",
  "BUILDKITE_API_TOKEN": "nice try!",
  # ...
  "ZEBRA": "nope!",
  "__CFBundleIdentifier": "com.googlecode.iterm2",
  "__CF_USER_TEXT_ENCODING": "0x1F5:0:15"
}
# Bloody good of them to have animals at each end of the alphabet i reckon
```

part of the reason that i decided on "big ol json map" as the encoding mechanism here is because the `env.Environment` type is just a `map[string]string` with a bit of sauce on it, so it makes the encode/decode cycle of this nice and easy.

hook/scriptwrapper.go
triarius: Was it golangci-lint that let you know?

hook/scriptwrapper.go
triarius: Can we avoid making a change to windows if this is only known to be a bash issue so far?

hook/scriptwrapper.go
moskyb: oh weird! i _thought_ we could use `os.Executable()` here, but for some reason it doesn't appear to play nicely with [bintest](https://github.com/buildkite/bintest), and causes every(?) test with bintest to fail with errors like
```
hooks_integration_test.go:450: fork/exec /opt/homebrew/bin/git: resource temporarily unavailable
# or
ssh_test.go:88: Compile of _bintest_06ec6d25081df5c7f7c59c7397831c7eae5da454/main.go failed: go: error obtaining buildID for go tool compile: fork/exec /opt/homebrew/Cellar/go/1.19.2/libexec/pkg/tool/darwin_arm64/compile: resource temporarily unavailable
# or
shell_test.go:265: Error running `/bin/pwd`: fork/exec /bin/pwd: resource temporarily unavailable
```

... and then the test hangs for 10 minutes, then times out, and everything is terrible forever. This only happens if i add an `os.Executable()` call to the scriptwrapper.

I think i'm going to close this can of worms by just calling `buildkite-agent env` directly, as there's prior art there, but i might make a note of it in a comment somewhere?

hook/scriptwrapper.go
moskyb: it might have been 😅

hook/scriptwrapper.go
moskyb: i'm kinda torn on this. i agree that making a minimal change (ie not messing with the windows scripts) will make this change easier, but it would also be really nice to have one really consistent way of getting the environment - otherwise we're waiting for powershell to pull a fast one on us in the same way that bash has.

the other thing is that not changing the windows implementation to be the same makes deleting the existing code harder - it's all currently tied up in the [`env.FromExport`](https://github.com/buildkite/agent/blob/main/env/export.go#L52) method, which does a bunch of malarkey to figure out what kind of output it's getting, and then tries to parse it. if we could delete that whole method, it'll be a much simpler change to make.

i'm happy to smoke test this change on a windows box in AWS if that makes things easier

hook/scriptwrapper.go
```diff
@@ -220,29 +220,50 @@ func (wrap *ScriptWrapper) Close() {
...
+		if err != nil {
+			return HookScriptChanges{}, fmt.Errorf("failed to unmarshal after env file: %w, file contents: %q", err, string(afterEnvContents))
+		}
+	}
+
 	diff := afterEnv.Diff(beforeEnv)
 
 	// Pluck the after wd from the diff before removing the key from the diff
 	afterWd := ""
 	if afterWd == "" {
```
pda, line 265: lol

pda (approved):

triarius (approved):
LGTM

hook/scriptwrapper.go
moskyb: for future travellers: we decided to prioritise the fix for the security issue, as that's more pressing than having consistent code, and the issue shouldn't affect windows machines at all, so it's probably fine to leave them as is for now.

We'll backport the `buildkite-agent env` stuff into windows very soon though, as it's definitely nice to have consistent code, and it lets us delete a whole bunch of code. yay!

PR #1782, opened 2022-10-14
Bump version and CHANGELOG for v3.39.1

triarius:


triarius:
Doing this as a backport instead

PR #1786, opened 2022-10-14
Add queue name tag to job metrics

francoiscampbell:
Adds the name of the queue(s) that the agent is listening on as a tag on job-level metrics.

The goal is to be able to split the `buildkite.queue.duration` metric by queue, to make easy to determine if a particular queue's agent pool is overloaded or underscaled, but the tag may be useful for other metrics that fall under `jobMetricsScope` as well.

If an agent is listening on a single queue, the tag will be `queue: <queue>`, for multiple queue, the tag will be `queue: <queue1>,<queue2>`, etc.


francoiscampbell:
Atlantis plan

pda:
Thanks @francoiscampbell!

In theory this could increase custom metric costs for some people, especially if they have unbounded dynamic queue names. But I don't think that's common, and don't think it'll be a problem in reality.

Unfortunately this isn't a well-tested code path, but I've tested it locally, and can see the correct `queue:…` tag on the statsd UDP packets 👌🏼

For future reference, I used:
```sh
ruby -r socket -e 's = UDPSocket.new; s.bind("127.0.0.1", 8125); loop { d, _ = s.recvfrom(1024); puts d if d.start_with?("buildkite") }'
```

pda:
Merged in cc28e099

PR #1785, opened 2022-10-14
Bump version and CHANGELOG for 3.39.1

triarius:
## [v3.39.1](https://github.com/buildkite/agent/tree/v3.39.1) (2022-10-14)
[Full Changelog](https://github.com/buildkite/agent/compare/v3.39.0...v3.39.1)

### Security
- Fix security issue where in certain situations, agents using bash v5.2 could have some environment variables leak to job logs. See the [issue](https://github.com/buildkite/agent/issues/1780) and [our security advisory](https://buildkite.com/changelog/167-security-advisory-for-buildkite-agent-and-bash-5-dot-2) for more information. [#1781](https://github.com/buildkite/agent/pull/1781) (@moskyb + @triarius + @pda) (@moskyb)

PR #1787, opened 2022-10-17
Embed agent version in its own file

steveh:
This might not be desirable, but: In #1785 we had to add a comment purely to avoid Go linting combining two variable declarations. This is because a build step parses Go in a fragile way to extract the version.

By making the version plain text we can avoid this step.


.buildkite/steps/extract-agent-version-metadata.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -euo pipefail
 
-agent_version=$(awk -F\" '/var baseVersion string = "/ {print $2}' agent/version.go)
+agent_version=$(<"agent/VERSION")
```
moskyb, line 4: my bash-fu is leaving me: is this the same as having 
```
agent_version=$(cat agent/VERSION)
```

if so, can we use cat rather than the hard-to-google bash syntax?

agent/VERSION
```diff
@@ -0,0 +1 @@
+3.39.0
```
moskyb, line 1: i got a little bit confused by this initially, thinking "shouldn't this be 3.39.1"?, but then i realised that 3.39.1 is on another branch, that isn't up do date with master. non-linear development is confusing sometimes.

.buildkite/steps/extract-agent-version-metadata.sh
steveh: The term-of-art is [input redirection](https://www.gnu.org/software/bash/manual/html_node/Redirections.html#Redirecting-Input). Shell purists would [throw a fit](https://en.wikipedia.org/wiki/Cat_(Unix)#Useless_use_of_cat), but cat is more obvious?

moskyb: readability > purism, imo. shell purists can stay mad

steveh: Rebased, and added `xargs` to trim newlines

agent/VERSION
steveh: I initially did it as 3.39.1 before realising I was targeting `main` :D

.buildkite/steps/extract-agent-version-metadata.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -euo pipefail
 
-agent_version=$(awk -F\" '/var baseVersion string = "/ {print $2}' agent/version.go)
+agent_version=$(cat agent/VERSION | xargs) # xargs to trim trailing newlines
```
triarius, line 4: Nit: I don't think `xargs` is necessary:

> The shell shall expand the command substitution by executing command in a subshell environment (see [Shell Execution Environment](https://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html#tag_02_12)) and replacing the command substitution (the text of command plus the enclosing "$()" or backquotes) with the standard output of the command, removing sequences of one or more <newline>s at the end of the substitution. Embedded <newline>s before the end of the output shall not be removed; however, they may be treated as field delimiters and eliminated during field splitting, depending on the value of IFS and quoting that is in effect.

Source: https://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html#tag_02_06_03

Experimental evidence:
```
🕙 16:37:47 ❯ docker run --rm -it bash:latest
bash-5.2# printf $(printf "hello\n")
hellobash-5.2# printf $(printf "\nhello\n")
hellobash-5.2# printf "$(printf "\nhello\n")"

hellobash-5.2#
```

steveh: TIL! Rebased.

triarius (approved):
Thanks for bearing with us! I think this is a good improvement to how things are at the moment so should be good to merge.

PR #1790, opened 2022-10-17
When homebrew releases fail, print the body of the API call that failed

moskyb:
Our homebrew release step wasn't working for a little while, because we'd misconfigured the branch protection settings on the homebrew-buildkite repo. The failure only printed out the status code (409? why?), which made it harder to diagnose the issue.

This PR replaces the curl `--fail` flag with `--fail-with-body`, which does the same thing, but means that it'll print the response when the command fails, making issues with this process easier to diagnose in the future.


triarius (approved):
LGTM

PR #1791, opened 2022-10-18
Make windows hooks process env changes the same way as unix

moskyb:
This uses the new buildkite-agent env command introduced in [#1781](https://github.com/buildkite/agent/pull/1781), and uses it for hooks on windows machines as well. This change was scoped out of the initial fix, as windows machines weren't affected by the bug, and the effort of testing them would add time between the issue being reported and a fix being issued.

This PR ports the behaviour that we were previously using for hooks running on unix machines to windows machines as well, making the code a bunch nicer and allowing us to delete a whole bunch of useless env processing code.


pda (approved):
This looks good to me, assuming some automatic and perhaps manual testing passes.

Love the deletion of all that old env code 👌🏼

moskyb:
The tests here are failing because apparently some windows env vars start with an `=` 🙃

The fix here could either be to silently ignore them, or to special case them in some way. I would slightly prefer to special case them - i don't think our previous implementation properly handled these, but it would probably be good to have the option available.

hook/scriptwrapper_test.go
```diff
@@ -110,11 +104,11 @@ func TestHookScriptsAreGeneratedCorrectlyOnWindowsBatch(t *testing.T) {
 	// See: https://pkg.go.dev/fmt > ctrl-f for "%%"
 	scriptTemplate := `@echo off
 SETLOCAL ENABLEDELAYEDEXPANSION
-SET > "%s"
+buildkite-agent env > "%s"
```
DrJosh9000, line 107: Checking the exact contents of the wrapper script in a test is a change detector, and there's a test for the functionality of the wrapper above. Is there any value beyond testing the file was written and is nonempty?

moskyb: yeah, i get where you're coming from here. these tests were originally added as regression tests when we replaced [this nonsense](https://github.com/buildkite/agent/blob/345e259bc18bdd62cc86407acb75d9a17aa77415/hook/scriptwrapper.go#L117-L141) with the current template solution. i think there's a little bit of value in making double sure that we don't accidentally mess with the templates, but i agree that the value isn't huge.

PR #1802, opened 2022-10-25
Apply `gofmt -s` simplifications

DrJosh9000:
Beyond the usual formatting of `gofmt`, the `-s` flag applies some straightforward simplifications, such as removing redundant types. While useful from time to time, it is usually unnecessary to enforce it on every change, so I left the gofmt test alone.


moskyb (approved):
:chefkiss:

PR #1803, opened 2022-10-25
buildkite-agent env has --from-env-file and --print flags

pda:
Adds two flags to the `buildkite-agent env` command introduced in #1781.

By default, `buildkite-agent env` loads `os.Environ()` and prints it as a JSON object; it's mainly intended for internal use to capture a snapshot of environment before a hook runs, and detect changes afterwards.

The new `--from-env-file` boolean flag sources the environment from the file named by `$BUILDKITE_ENV_FILE` instead of from `os.Environ()`, using `bufio.Scanner` and `strconv.Unquote()` to safely and correctly parse the quoting/escaping done by `%q` when the file was written:
https://github.com/buildkite/agent/blob/2aaefb699828b512ad62cd7791104b47bad46b96/agent/job_runner.go#L490

The new `--print NAME` flag causes a single environment value to be printed in its raw unescaped form followed by a newline, instead of printing all vars as a JSON object.

Combined, this solves the problem of safely parsing `BUILDKITE_ENV_FILE` (which cannot be safely evaluated in a shell if it may contain untrusted user input) to access a single variable, and allows for usage like this in a pre-command hook or similar:

```bash
if [[ $(buildkite-agent env --from-env-file --print BUILDKITE_BRANCH) != "main" ]]; then
  echo "This agent only builds branch 'main'"
  exit 42
fi
```

### DIFF
--- a/clicommand/env.go
+++ b/clicommand/env.go
@@ -1,9 +1,11 @@
 package clicommand
 
 import (
+	"bufio"
 	"encoding/json"
 	"fmt"
 	"os"
+	"strconv"
 	"strings"
 
 	"github.com/urfave/cli"
@@ -32,14 +34,33 @@ var EnvCommand = cli.Command{
 			Usage:  "Pretty print the JSON output",
 			EnvVar: "BUILDKITE_AGENT_ENV_PRETTY",
 		},
+		cli.BoolFlag{
+			Name:  "from-env-file",
+			Usage: "Source environment from file described by $BUILDKITE_ENV_FILE",
+		},
+		cli.StringFlag{
+			Name:  "print",
+			Usage: "Print a single environment variable by `NAME` as raw text followed by a newline",
+		},
 	},
 	Action: func(c *cli.Context) error {
-		env := os.Environ()
-		envMap := make(map[string]string, len(env))
+		var envMap map[string]string
 
-		for _, e := range env {
-			k, v, _ := strings.Cut(e, "=")
-			envMap[k] = v
+		if c.Bool("from-env-file") {
+			envMap = mustLoadEnvFile(os.Getenv("BUILDKITE_ENV_FILE"))
+		} else {
+			env := os.Environ()
+			envMap = make(map[string]string, len(env))
+
+			for _, e := range env {
+				k, v, _ := strings.Cut(e, "=")
+				envMap[k] = v
+			}
+		}
+
+		if name := c.String("print"); name != "" {
+			fmt.Println(envMap[name])
+			return nil
 		}
 
 		var (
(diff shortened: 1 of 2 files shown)


clicommand/env.go
```diff
@@ -32,14 +34,33 @@ var EnvCommand = cli.Command{
...
+		if c.Bool("env-file") {
+			envMap = mustLoadEnvFile(os.Getenv("BUILDKITE_ENV_FILE"))
+		} else {
+			env := os.Environ()
+			envMap = make(map[string]string, len(env))
+
+			for _, e := range env {
+				k, v, _ := strings.Cut(e, "=")
+				envMap[k] = v
+			}
```
triarius, line 58: I suspect @moskyb's efforts to also use this command in windows will cause a merge conflict here.

triarius (approved):
LGTM, but someone will like have merge conficts.

clicommand/env.go
```diff
@@ -32,20 +35,41 @@ var EnvCommand = cli.Command{
...
-		env := os.Environ()
-		envMap := make(map[string]string, len(env))
+		var envMap map[string]string
+		var err error
 
-		for _, e := range env {
-			k, v, _ := strings.Cut(e, "=")
-			envMap[k] = v
+		if c.Bool("from-env-file") {
+			envMap, err = loadEnvFile(os.Getenv("BUILDKITE_ENV_FILE"))
```
sj26, line 52: This feels weird to me. I expect `--from-env-file` to take a value, but default to an environment variable, like our other options. But if we defaulted to the env file then it would break dumping of the current env as json.

sj26:
This feels like the same domain but a pretty orthogonal use case to the underlying command to dump the current env as json. Should it be a separate command?

```json
buildkite-agent env get NAME
```

pda:
> Maybe the recently introduced command should also be a subcommand?
> 
> ```
> buildkite-agent env dump --json
> ```

Yeah I like that. Since the recently-added `buildkite-agent env` is private/internal usage only (hopefully), we're free to evolve it to support other use-cases around it in a coherent way 👍🏼

clicommand/env.go
pda: Yep. I like your idea of moving the current “dump entire process env as JSON” use-cases into a subcommand/flag(s), which then frees up the `buildkite-agent env` namespace to support some other environment-related use-cases.

sj26:
The nice thing about subcommands is if we peer forwards to being able to set env from things which are not bash hooks, that could use `buildkite-agent env set NAME VALUE` to talk to the parent process and inject env without sourcing and env dumping shenanigans at all.

PR #1805, opened 2022-10-25
Enable overriding buildkite-agent url in `install.ps1`

staticfloat:
When using the very helpful `buildkite-agent` installation powershell script, it is occasionally necessary to install some other version, rather than the current latest release.  While it would be cool to support specifying an actual version, the simplest method of this is to simply provide the exact URL to be downloaded.  Example invocation:

```
$env:buildkiteAgentUrl = "https://github.com/buildkite/agent/releases/download/v3.39.0/buildkite-agent-windows-amd64-3.39.0.zip"

iex ((New-Object System.Net.WebClient).DownloadString('https://raw.githubusercontent.com/staticfloat/agent/patch-1/install.ps1'))
```

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -2,6 +2,7 @@ $installDir = "C:\buildkite-agent"
 $beta = $env:buildkiteAgentBeta
 $token = $env:buildkiteAgentToken
 $tags = $env:buildkiteAgentTags
+$url = $env:buildkiteAgentUrl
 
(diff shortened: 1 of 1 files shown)


DrJosh9000 (approved):
Thanks for this! Sorry it took a while.

mwrock:
FYI: This merge breaks current calls to https://raw.githubusercontent.com/buildkite/agent/master/install.ps1 when `$version` is not populated.

PR #1808, opened 2022-10-27
WIP: Allow the redactor to redact values with non-ascii characters

moskyb:


redaction/redactor.go
```diff
@@ -134,10 +167,13 @@ func (redactor *Redactor) Write(input []byte) (int, error) {
...
 	doneTo := 0
 
+	pretty.Println(redactor.table)
+
 	// For so long as the safe consumption index is inside the current input array
 	for cursor < len(input) {
 		ch := input[cursor]
-		skip := redactor.table[ch].skip
+		fmt.Printf("ch: %c\n", rune(ch))
+		skip := redactor.table.Skip(rune(ch))
```
moskyb, line 176: this solution doesn't work because of this line. the problem is that we store all of the skip values as a `map[rune]int`, but the redactor looks through its input byte-by-byte. the issue here is that when we call `ch := input[cursor]`, we're selecting the `cursor`th _byte_ from input, which might only represent some of the "character" that we're looking for.

moskyb:
superseded by the much more sensible #1809. Thanks @DrJosh9000!

PR #1809, opened 2022-10-27
redaction: treat needles as byte slices

DrJosh9000:
redactor.go:100 ranges over a string (needle), and in Go, ranging over a string iterates each _rune_ each of which can be a single or multibyte code point. But the rest of the redactor deals in bytes, particularly the table array.

This also updates the misleadingly-named test to (a) a better name and (b) actually look at the redacted output.


pda (approved):
👌🏼

PR #1810, opened 2022-10-27
Key JUnit files with the step ID, rather than the OS

moskyb:
This means that every JUnit file will have a unique file name, should stop the test annotate step from failing intermittently

### DIFF
--- a/.buildkite/steps/tests.sh
+++ b/.buildkite/steps/tests.sh
@@ -7,7 +7,7 @@ echo arch is "$(uname -m)"
 go install gotest.tools/gotestsum@v1.8.0
 
 echo '+++ Running tests'
-gotestsum --junitfile "junit-${OSTYPE}.xml" -- -count=1 -failfast "$@" ./...
+gotestsum --junitfile "junit-${BUILDKITE_JOB_ID}.xml" -- -count=1 -failfast "$@" ./...
 
 echo '+++ Running integration tests for git-mirrors experiment'
-TEST_EXPERIMENT=git-mirrors gotestsum --junitfile "junit-${OSTYPE}-git-mirrors.xml" -- -count=1 -failfast "$@" ./bootstrap/integration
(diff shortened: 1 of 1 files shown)


.buildkite/steps/tests.sh
```diff
@@ -1,13 +1,13 @@
...
 
 go install gotest.tools/gotestsum@v1.8.0
 
 echo '+++ Running tests'
-gotestsum --junitfile "junit-${OSTYPE}.xml" -- -count=1 -failfast "$@" ./...
+gotestsum --junitfile "junit-${BUILDKITE_STEP_KEY}.xml" -- -count=1 -failfast "$@" ./...
 
 echo '+++ Running integration tests for git-mirrors experiment'
-TEST_EXPERIMENT=git-mirrors gotestsum --junitfile "junit-${OSTYPE}-git-mirrors.xml" -- -count=1 -failfast "$@" ./bootstrap/integration
+TEST_EXPERIMENT=git-mirrors gotestsum --junitfile "junit-${BUILDKITE_STEP_KEY}-git-mirrors.xml" -- -count=1 -failfast "$@" ./bootstrap/integration
```
triarius, line 13: In general, if you expect one variable to not be set, I think `set +u` is too broad. you should use a pattern like
```bash
JUNIT_FILE_KEY=${BUILDKITE_STEP_KEY:-$OSTYPE}
```
instead

triarius (commented):
I don't think its best to ignore an unset `BUILDKITE_STEP_KEY`. But I really don;t know why it could be unset 🤷‍♂️

.buildkite/steps/tests.sh
tessereth: You could use BUILDKITE_JOB_ID. That's guaranteed to exist and be unique. I'm unsure why BUILDKITE_STEP_KEY wasn't defined in that example though...

moskyb: Yeah, i'm not sure why that happened either... regardless, i've updated it to keep `set -u` and use `BUILDKITE_JOB_ID` instead

triarius (approved):
🎉

PR #1811, opened 2022-10-31
env: handle '='-prefixed vars

### LINKED ISSUE #1804: agent v3.39.1 incorrectly parses special Windows environment variables when running bash hooks
When running bash plugin hooks on windows, the new `buildkite-agent env` tool incorrectly parses some special environment variable names such as `=C`, which results in JSON entries such as the following:

```
.\buildkite-agent.exe env --pretty
{                                                                                                                                            
  "": "C:=C:\\Users\\Administrator",
...
}
```

This later surfaces with very strange "environment variable changed" messages in our build logs such as the following:
image

Note that I am showing the `buildkite-agent env` invocation by setting `set -x` in our bash hook script, and to show that since we're using bash hooks, we don't use the typical powershell environment collector, which hasn't changed in v3.39.1.  To get these environment variables, you need to invoke `cmd.exe` at some point, [which will automatically create the variables](https://devblogs.microsoft.com/oldnewthing/20100506-00/?p=14133).  An easy reproducer is here, launched from within `bash` since it provides a convenient way of stripping out most variables:

```
$ env -i C:\\Windows\\System32\\cmd.exe /c ".\buildkite-agent.exe env --pretty"
{
  "": "C:=C:\\Users\\Administrator\\AppData\\Local\\Temp\\bk",
  "COMSPEC": "C:\\Windows\\System32\\cmd.exe",
  "MSYSTEM": "MINGW64",
  "PATH": "C:\\Program Files\\Git\\usr",
  "PATHEXT": ".COM;.EXE;.BAT;.CMD;.VBS;.JS;.WS;.MSC",
  "PROMPT": "$P$G",
  "SYSTEMROOT": "C:\\Windows",
  "WINDIR": "C:\\Windows"
}
```

I don't have much `go` experience, so I'm not certain what happens to the data after this point, but I can say that my updates to `PATH` do not get properly merged from this point on, despite the fact that there's no error in the build UI.

[I would like the special variables that start with `=`](https://ss64.com/nt/syntax-variables.html#:~:text=Undocumented%20Dynamic%20variables%20(read%20only)) to be parsed correctly, or dropped.  These variables are not necessary for my use cases, but since they are _technically_ legal (you can't set them via `set` in the cmd terminal, but you _can_ set them via `SetEnvironmentVariable`, so a parent process could set these and then launch `buildkite-agent start`).  I personally am fine with them being dropped.

It is somewhat difficult to ensure that these variables are not set when I invoke `buildkite-agent`, so at the moment I am locking our infrastructure to using `buildkite-agent v3.39.0`.

DrJosh9000:
Windows can have interestingly-named environment variables (#1804). This update changes both the env command and env subpackage to handle variables with names starting with '='.

It is debatable whether to filter out these variables or preserve them. Preserving them feels like the least surprising option to me. I did some brief experimenting on Windows and didn't uncover huge issues. But I'm willing to be convinced otherwise.

I've also taken the opportunity to slightly simplify the env command implementation. 

Fixes #1804

### DIFF
--- a/clicommand/env.go
+++ b/clicommand/env.go
@@ -39,12 +39,9 @@ var EnvCommand = cli.Command{
 		envMap := make(map[string]string, len(envn))
 
 		for _, e := range envn {
-			k, v, ok := env.Split(e)
-			if !ok {
-				fmt.Fprintf(c.App.ErrWriter, "Invalid environment variable from os.Environ: %q\n", e)
-				os.Exit(2)
+			if k, v, ok := env.Split(e); ok {
+				envMap[k] = v
 			}
-			envMap[k] = v
 		}
 
 		enc := json.NewEncoder(c.App.Writer)
--- a/clicommand/env.go
+++ b/clicommand/env.go
@@ -4,8 +4,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
-	"strings"
 
+	"github.com/buildkite/agent/v3/env"
 	"github.com/urfave/cli"
 )
 
@@ -13,8 +13,9 @@ const envDescription = `Usage:
   buildkite-agent env [options]
 
 Description:
-   Prints out the environment of the current process as a JSON object, easily parsable by other programs. Used when
-   executing hooks to discover changes that hooks make to the environment.
+   Prints out the environment of the current process as a JSON object, easily
+   parsable by other programs. Used when executing hooks to discover changes
+   that hooks make to the environment.
 
 Example:
    $ buildkite-agent env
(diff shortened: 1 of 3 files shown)


DrJosh9000:
tl;dr: I have further confidence.

I did some manual testing with this branch merged with #1791, and got funky with environment.bat in order to generate an =var that isn't overridden. e.g.

```
SUBST D: C:\Windows
D:
CD System32
```

They were passed through to the build step successfully.

clicommand/env.go
```diff
@@ -34,38 +35,27 @@ var EnvCommand = cli.Command{
...
-			k, v, _ := strings.Cut(e, "=")
+		envn := os.Environ()
+		envMap := make(map[string]string, len(envn))
+
+		for _, e := range envn {
+			k, v, ok := env.Split(e)
+			if !ok {
+				fmt.Fprintf(c.App.ErrWriter, "Invalid environment variable from os.Environ: %q\n", e)
+				os.Exit(2)
+			}
```
moskyb, line 46: we're changing the failure dynamics here - previously, we'd ignore any envar that we weren't able to parse, and now we'll fail the command.

this change is all good, i'm also keen to make sure that this it's here intentionally, and isn't just an artifact of debugging or something

DrJosh9000: I'm not sure why I was feeling paranoid about it here. os.Environ has to produce well-formed strings and the only way it wouldn't is if things are severely broken on the machine... but then all bets are off anyway.

env/environment_test.go
```diff
@@ -209,3 +209,26 @@ func TestEnvironmentApply(t *testing.T) {
...
+	t.Parallel()
+	tests := []struct {
+		in          string
+		name, value string
+		ok          bool
+	}{
+		{"key=value", "key", "value", true},
+		{"equalsign==", "equalsign", "=", true},
+		{"=Windows=Nonsense", "=Windows", "Nonsense", true},
+		{"=Bonus=Windows=Nonsense", "=Bonus", "Windows=Nonsense", true},
```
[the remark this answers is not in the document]
DrJosh9000: Done.

moskyb (approved):

PR #1814, opened 2022-10-31
bootstrap/shell: re-enable TestInterrupt under -race

DrJosh9000:
The race in bintest that causes TestInterrupt to fail should now be fixed.

While I'm here, replace uses of io/ioutil (deprecated).


moskyb (approved):
🎉

PR #1815, opened 2022-11-03
env: Disallow variables starting with =

DrJosh9000:
To maintain the same behaviour as before, we've instead decided to drop them instead of preserve them. 

This changes the outcome for #1804, but #1804 remains fixed.


moskyb (approved):
🎉

it's a bummer that we had to drop this for compatibility, but i'm glad we got here.

PR #1817, opened 2022-11-06
Replace deprecated ioutil calls with equivalents

DrJosh9000:
io/ioutil was deprecated as of Go 1.16. Equivalent functionality is now in the io or os packages. 

Do this cleanup in one go so I stop doing it piece-by-piece.


triarius (approved):
Thanks! 🙏

PR #1822, opened 2022-11-08
bootstrap: Tidy up t.Error/t.Fatal (part 2)

DrJosh9000:
This is the second chunk in a series of tidy-ups. This one focuses on tests in the bootstrap subpackage, including a few tests in bootstrap/integration (but not all), and applying comments from https://github.com/golang/go/wiki/TestComments (excluding the comment about assert libraries, which is a battle to fight later).

`TestCleaningAnExistingCheckout`(checkout_integration_test.go), appears to be buggy: test.txt is seemingly not removed, and existence was not properly checked, so it erroneously passed. It is skipped (for now).

### DIFF
--- a/api/client_test.go
+++ b/api/client_test.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/http/httptest"
+	"strings"
 	"testing"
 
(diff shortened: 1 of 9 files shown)


triarius (approved):
LGTM

PR #1824, opened 2022-11-09
bootstrap: Tidy up t.Error/t.Fatal (part 3)

DrJosh9000:
This is the third chunk in a series of tidy-ups. This one focuses on
tests in the bootstrap subpackage, and applies some comments from
https://github.com/golang/go/wiki/TestComments.

### DIFF
--- a/bootstrap/shell/test.go
+++ b/bootstrap/shell/test.go
@@ -13,7 +13,7 @@ import (
 func NewTestShell(t *testing.T) *Shell {
 	sh, err := New()
 	if err != nil {
-		t.Fatal(err)
+		t.Fatalf("shell.New() error = %v", err)
 	}
 
 	sh.Logger = DiscardLogger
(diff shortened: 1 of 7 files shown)


bootstrap/shell/test.go
```diff
@@ -13,7 +13,7 @@ import (
 func NewTestShell(t *testing.T) *Shell {
 	sh, err := New()
 	if err != nil {
-		t.Fatal(err)
+		t.Fatalf("shell.New() error = %v", err)
```
triarius, line 16: What's your perspective on converting these to be black box test? From a cursory glance, I think for this file at least, it should be straightforward.

triarius (approved):
LGTM 💖

My comment is more for later discussion

bootstrap/shell/test.go
DrJosh9000: I'm in favour of decreasing how aware the tests are of how the code operates (something like "test what it does, not how it does it" - there's a good quote somewhere I can't seem to find...). Probably a higher priority would be increasing coverage on other critical agent pieces.

PR #1826, opened 2022-11-10
Tidy up t.Error/t.Fatal (part 4)

DrJosh9000:
This should be the last tidy-up focused on t.Error and t.Fatal calls.


triarius (approved):
LGTM, thanks!

PR #1827, opened 2022-11-10
Add `buildkite-agent oidc request-token` command

triarius:
I decided to have just 1 method abstracting the API call for when there is and is not an audience. ~~The way I chose to implement the method gives room to extend it if we decided Buildkite's API should allow multiple audiences.~~

~~Per the spec, `aud` can be EITHER an array or a string: https://openid.net/specs/openid-connect-core-1_0.html#IDToken. So it is certainly possible for us to make that choice in the future, even if we do not want to do that now. Indeed, there may be services that would require multiple audience claims in the single JWT that we may want to allow integration with. Moreover, I have seen `aud` be an array in the wild: https://community.auth0.com/t/audience-value-in-returned-bearer-token-jwt-is-a-list/6418.~~

~~However, as a note for if we do enable such support, we should ensure that the API issues a JWT with a string valued `aud` if there is only a single audience as some JWT libraries don't easily support validating JWTs with an array valued `aud`: https://github.com/dgrijalva/jwt-go/issues/445 and so many implementations may break when using them.~~


api/client_test.go
```diff
@@ -1,4 +1,4 @@
-package api
+package api_test
```
triarius, line 1: I wanted to use a private method in this file in a black box test, so I made this file a black box test too.

api/oidc.go
```diff
@@ -0,0 +1,40 @@
...
+	"fmt"
+)
+
+var ErrAudienceTooLong = errors.New("the API only supports at most one element in the audience")
+
+type OidcToken struct {
+	Token string `json:"token"`
+}
+
+func (c *Client) OidcToken(jobId string, audience ...string) (*OidcToken, *Response, error) {
```
pda, line 14: Does this `audience ...string` aspect of the method signature scale to handling future user-specified args (e.g. expiry/lifetime/TTL)?

I wonder if it should receive these parameters as a struct?
I wonder if `type oidcTokenRequest struct` should be public and received as an arg?
Not sure 🤷‍♂️

api/oidc.go
```diff
@@ -0,0 +1,40 @@
...
+	type oidcTokenRequest struct {
+		Audience string `json:"audience,omitempty"`
+	}
+
+	m := &oidcTokenRequest{}
+	switch len(audience) {
+	case 0:
+	case 1:
+		m.Audience = audience[0]
```
pda, line 23: I initially read this as C-style switch fall-through, where cases `0` and `1` are handled the same, and that left me wondering why it didn't panic on `audience[0]` for an empty slice.

Maybe it's worth making it clearer that `case 0:` is a no-op?

(Alternatively, maybe the complexity of handling multiple audiences is YAGNI since we may never support that?)

clicommand/oidc_token.go
```diff
@@ -0,0 +1,147 @@
...
+   $ buildkite-agent oidc token --audience sts.amazonaws.com
+
+   Prints the environment passed into the process
+`
+	backoffSeconds = 2
+	maxAttempts    = 5
+)
+
+var OidcTokenCommand = cli.Command{
+	Name:        "token",
```
pda, line 53: I wonder if something verby like `request-token` or `get-token` would be clearer than the noun `token`?

The verb “request” would align with https://openid.net/specs/openid-connect-core-1_0.html#TokenRequest

> A Client makes a Token Request by presenting its Authorization Grant (in the form of an Authorization Code) to the Token Endpoint …

main.go
```diff
@@ -90,6 +90,13 @@ func main() {
 				clicommand.MetaDataKeysCommand,
 			},
 		},
+		{
+			Name:  "oidc",
+			Usage: "Interact with Buildkite OIDC",
```
pda, line 95: I'm not sure, but perhaps this will help some people:

```suggestion
			Usage: "Interact with Buildkite OpenID Connect (OIDC)",
```

api/oidc.go
triarius: Yeah, that's a good idea. I will do that.

api/oidc.go
triarius: I didn't consider it to be YAGNI as it's in the spec, and we don't control whether 3rd parties would read the spec and think it would be a good idea to require multiple audiences on a OIDC token they consume. Then our implementation would not work with such a service.

But this it not really the right place to have a debate about it, bk/bk needs to support it too. So I'll implement it with at most audience in the agent for now.

api/oidc.go
```diff
@@ -0,0 +1,32 @@
+package api
+
+import (
+	"fmt"
+)
+
+type OidcToken struct {
```
DrJosh9000, line 7: I believe initialisms, such as "OIDC", should be in their natural-looking case. So this would be "OIDCToken". Similarly with the other types and functions. https://github.com/golang/go/wiki/CodeReviewComments#initialisms

api/oidc.go
```diff
@@ -0,0 +1,32 @@
...
+	}
+
+	u := fmt.Sprintf("jobs/%s/oidc/tokens", methodReq.JobId)
+	httpReq, err := c.newRequest("POST", u, m)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	t := &OidcToken{}
+	resp, err := c.doRequest(httpReq, t)
```
DrJosh9000, line 30: Are the token and response ever useful if err != nil? If so that would make using this method more complicated.
Otherwise to avoid potential confusion ("I called OIDCToken and got both an error and some other stuff?") I would insert an `if err != nil { return nil, nil, err }` in here.

api/oidc_test.go
```diff
@@ -0,0 +1,143 @@
...
+	"github.com/buildkite/agent/v3/logger"
+	"github.com/google/go-cmp/cmp"
+)
+
+func newOidcTokenServer(
+	t *testing.T,
+	accessToken, oidcToken, path string,
+	expectedBody []byte,
+) *httptest.Server {
+	t.Helper()
```
DrJosh9000, line 22: Breaking function declarations across multiple lines feels weird to me. It's the sort of thing that leads to indentation-level confusion.
And many other Go folks think there's no problem with long lines anyway:
https://github.com/golang/go/wiki/CodeReviewComments#line-length

api/oidc_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+	return httptest.NewServer(http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+		switch req.URL.Path {
+		case path:
+			if got, want := authToken(req), accessToken; got != want {
+				http.Error(
```
DrJosh9000, line 28: Similarly here, and below - is it so uncomfortably long to put it on one line?

api/oidc_test.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+	}))
+}
+
+func TestOidcToken(t *testing.T) {
+	const jobId = "b078e2d2-86e9-4c12-bf3b-612a8058d0a4"
+	const oidcToken = "eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiYWRtaW4iOnRydWUsImlhdCI6MTUxNjIzOTAyMn0.NHVaYe26MbtOYhSKkoKYdFVomg4i8ZJd8_-RU8VNbftc4TSMb4bXP3l3YlNWACwyXPGffz5aXHc6lty1Y2t4SWRqGteragsVdZufDn5BlnJl9pdR_kdVFUsra2rWKEofkZeIC4yWytE58sMIihvo9H1ScmmVwBcQP6XETqYd0aSHp1gOa9RdUPDvoXQ5oqygTqVtxaDr6wUFKrKItgBMzWIdNZ6y7O9E0DhEPTbE9rfBo6KTFsHAZnMg4k68CDp2woYIaXbmYTWcvbzIuHO7_37GT79XdIwkm95QJ7hYC9RiwrV7mesbY4PAahERJawntho0my942XheVLmGwLMBkQ"
+	const accessToken = "llamas"
+
+	for _, testData := range []struct {
```
DrJosh9000, line 81: This one's a nit: The main downside to ranging over an inline-struct-slice like this is that the "for" is many lines above the loop body.  So by the time your eyes get to the loop body, you almost forget that it's a loop.
The alternative is `tests := []struct{...}{...}; for _, test := range tests {...}`

api/oidc_test.go
```diff
@@ -0,0 +1,143 @@
...
+			if token, resp, err := client.OidcToken(testData.OidcTokenRequest); err != nil {
+				if !errors.Is(err, testData.Error) {
+					t.Fatalf(
+						"OidcToken(%v) got error = %v, want error = %v",
+						testData.OidcTokenRequest,
+						err,
+						testData.Error,
+					)
+				}
+			} else if !cmp.Equal(token, testData.OidcToken) {
```
DrJosh9000, line 136: Testing both error and non-error cases in a table-driven test has the downside that often, you want to shortcut checks in one and not another. Which tends to end up with nested or chained conditionals like this 😛  It's an argument for splitting the tests up into two: one for non-errors and one for errors.

api/oidc_test.go
```diff
@@ -0,0 +1,143 @@
...
+	"net/http/httptest"
+	"testing"
+
+	"github.com/buildkite/agent/v3/api"
+	"github.com/buildkite/agent/v3/logger"
+	"github.com/google/go-cmp/cmp"
+)
+
+func newOidcTokenServer(
+	t *testing.T,
```
DrJosh9000, line 18: It looks like the `t` argument is only used for calling `t.Helper()`. If so there's likely no point in passing in the `t` at all (compare with https://github.com/golang/go/wiki/TestComments#mark-test-helpers - the `readFile` example calls `t.Fatal` but `t.Helper` ensures the line in the test itself is identified as the problem.)

api/oidc_test.go
```diff
@@ -0,0 +1,143 @@
...
+			client := api.NewClient(logger.Discard, api.Config{
+				UserAgent: "Test",
+				Endpoint:  server.URL,
+				Token:     accessToken,
+				DebugHTTP: true,
+			})
+
+			if token, resp, err := client.OidcToken(testData.OidcTokenRequest); err != nil {
+				if !errors.Is(err, testData.Error) {
+					t.Fatalf(
```
DrJosh9000, line 129: The loop body is not in a subtest (in the `t.Run` sense) so calling `t.Fatalf` will abort the whole table of tests. 
Perhaps wrap the closure up at the start of of the loop body in `t.Run` so that it "keeps going"? https://github.com/golang/go/wiki/TestComments#keep-going

api/oidc_test.go
```diff
@@ -0,0 +1,143 @@
...
+						"OidcToken(%v) got error = %v, want error = %v",
+						testData.OidcTokenRequest,
+						err,
+						testData.Error,
+					)
+				}
+			} else if !cmp.Equal(token, testData.OidcToken) {
+				t.Fatalf("OidcToken(%v) got token = %v, want %v", testData.OidcTokenRequest, token, testData.OidcToken)
+			} else if resp.StatusCode != http.StatusOK {
```
DrJosh9000, line 138: This `else` is redundant - the `t.Fatalf` would exit the test.

Consider checking both the response token and `resp.StatusCode`, with `t.Errorf`, but not in an `else` branch.

clicommand/odic_request_token.go
```diff
@@ -0,0 +1,148 @@
...
+		DebugHTTPFlag,
+
+		// Global flags
+		NoColorFlag,
+		DebugFlag,
+		LogLevelFlag,
+		ExperimentsFlag,
+		ProfileFlag,
+	},
+	Action: func(c *cli.Context) error {
```
DrJosh9000, line 83: Something that all the other commands do, that I would like to undo in most of them, is to change these Action closures into named top-level functions. A few of them "feel" too big and complicated to be an inline func, so it would pay to reduce the indentation level even by one tab.

clicommand/odic_request_token.go
```diff
@@ -0,0 +1,148 @@
...
+		// The configuration will be loaded into this struct
+		cfg := OidcTokenConfig{}
+
+		loader := cliconfig.Loader{CLI: c, Config: &cfg}
+		warnings, err := loader.Load()
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "%s\n", err)
```
DrJosh9000, line 90: `c.App.ErrWriter` instead of `os.Stderr`?  I think we should be leaning on `cli.App.Writer` and `cli.App.ErrWriter` more - it would provide a way to insert our own io.Writers in order to test the output of the subcommands.

clicommand/odic_request_token.go
```diff
@@ -0,0 +1,148 @@
...
+		}
+
+		// Setup any global configuration options
+		done := HandleGlobalFlags(l, cfg)
+		defer done()
+
+		// Create the API client
+		client := api.NewClient(l, loadAPIClientConfig(cfg, "AgentAccessToken"))
+
+		// Find the meta data value
```
DrJosh9000, line 108: It would be a bit shorter to print the token within the retry func as soon as the request succeeded. (You could declare `token, resp, err := client.OIDCToken(req)` on 121.)

DrJosh9000 (commented):
It's looking pretty good! Still, have some comments.

api/oidc_test.go
triarius: The `t` is used. `t.Error` is called when the body of the request does not match what is expected.

api/oidc_test.go
triarius: I'll change to using the `t.Error` followed by `continue` pattern.

api/oidc.go
triarius: `doRequest` returns the response in cases where `err != nil`. The rationale seems to be that it allows inspection of the status code and retrying if necessary. See https://github.com/buildkite/agent/blob/1a9d87e123f3b2c07c86d763a05e870a6395da0c/api/client.go#L260-L261

The practice in this codebase seems to be to do the retrying in the `clicommand` method and not the `api` one, and inspecting the status code from the bubbled up response. See https://github.com/buildkite/agent/blob/1a9d87e123f3b2c07c86d763a05e870a6395da0c/api/meta_data.go#L60-L65 and https://github.com/buildkite/agent/blob/1a9d87e123f3b2c07c86d763a05e870a6395da0c/clicommand/meta_data_get.go#L107-L119

So the respone SHOULD be passed up or the retry won't work as expected. The token is not needed not however, so I have changed it to pass a nil pointer on error now.

I don't really think this is a good practice though (see below). However, I do think it's better to consistently follow a bad practice than to mix different practices. So I think we should follow this for now, and have a piece of work to unravel `http.doRequest`. I haven't thought too deeply about what that should be, but here are some initial thoughts:

I think there is no good reason for `api.doRequest` to return an `http.Response`. Instead, I think the relavant parts of the response that are needed for retry handling or otherwise should be returned as part of a structured error. Indeed, there already is a error type that purports to do this: `api.ErrorReponse`, although it contains a pointer to `http.Response` 🤷‍♂️. Either way returning a pointer to `http.Response` is potentially misleading as the body of the `http.Response` is closed in `api.doRequest`, so it will look like there is no body if we try to use it outside the `api.doRequest` method.

api/oidc_test.go
triarius: I think this line would be “uncomfortably long” if the function declaration were not wrapped. It's a lot easier to read a function declaration that has been intentionally wrapped for each argument than one that has been arbitrarily wrapped by the code editor. So I prefer to do it this way.

api/oidc_test.go
triarius: Given that it is already quite indented, I do think that this line would be uncomfortably long if it were not wrapped.

api/oidc_test.go
triarius: Ok, I have changed it to declare the table explicitly.

api/oidc_test.go
triarius: My bad. I tried to do too much in one test and hurt myself in my confusion 😵. I'll make separate tables for errors and successes.

clicommand/odic_request_token.go
triarius: I agree. However, I prefer to consistently follow a bad practice until we decide to explicitly adopt a better one.

clicommand/odic_request_token.go
triarius: Done

api/oidc.go
DrJosh9000: That's fair. `net/http` itself does this sort of thing.

api/oidc_test.go
DrJosh9000: Ah, I missed that when I read too quickly. It does raise a question:

_Should it_ call `t.Error`? Without it, the test should still fail when the request mismatches and the test server returns a 400. With it, the test server is doing more than merely faking the real server?

No need to block on this one.

api/oidc_test.go
DrJosh9000: Okay - I won't hold you up with a debate around wrapping. It's not unreadable. I'll go away and think about it.

clicommand/odic_request_token.go
triarius: I think it's nicer to think of `token` as the value that the retrier produces. I removed `resp` as there is no reason for it to escape the retrier.

DrJosh9000 (approved):
LGTM 🥳

This is good. I think anything else I could comment would be nitpicking.

api/oidc_test.go
triarius: I think I have a solution for this. I'll make a new PR for it though.

PR #1828, opened 2022-11-14
Tidy up context.Context

DrJosh9000:
Some of the core code (shell.go, process.go) stored contexts and cancellation funcs in structs. The context package doc is clear about not doing this: https://pkg.go.dev/context. This PR removes context fields, and then makes all the consequent changes to pass context through function args instead.


triarius (approved):
Awesome work!

PR #1832, opened 2022-11-15
Add support for requesting custom lifetime to `buildkite-agent oidc request-token`

triarius:
This adds a `--lifetime` parameter to the `buidlkite-agent oidc request-token` command. This will be interpreted as an integer number of seconds. The API will add this to the current time (at the API server) and use this value for the `exp` claim in the OIDC token. Note that specifying `0` (or omitting) will result in the API's default lifetime being used instead.


DrJosh9000 (approved):
LGTM 🎉

PR #1834, opened 2022-11-16
Support contexts in API client

DrJosh9000:
`http.Request` can be interrupted using a context. This PR adds context args to newRequest and newFormRequest, and then makes all the consequent changes needed to pass a context into those.


triarius (approved):
Awesome 🪅

PR #1835, opened 2022-11-16
Set ErrWriter

triarius:
We have started to use the cli.Context's App.ErrWriter in some places
now, but it is created to be nil. This results in runtime panics if we
try to write to it.

### DIFF
--- a/main.go
+++ b/main.go
@@ -116,6 +116,8 @@ func main() {
 		clicommand.BootstrapCommand,
 	}
 
+	app.ErrWriter = os.Stderr
+
 	// When no sub command is used
 	app.Action = func(c *cli.Context) {
 		cli.ShowAppHelp(c)


main.go
```diff
@@ -116,6 +116,8 @@ func main() {
 		clicommand.BootstrapCommand,
 	}
 
+	app.ErrWriter = os.Stderr
```
DrJosh9000, line 119: app.Reader is also not set by NewApp, which we could potentially use. 
For completeness should we ensure all three are set appropriately?

DrJosh9000 (approved):
LGTM 😎

(Add this to the list of reasons we should upgrade urfave/cli: https://github.com/urfave/cli/blob/v1.22.10/app.go#L116-L129 vs https://github.com/urfave/cli/blob/v2.23.5/app.go#L141-L155)

main.go
triarius: I think v1.22.10 does not have a Reader, so just setting the ErrWriter should be enough for now.

PR #1836, opened 2022-11-16
Add restrictions on possible lifetime values at the clicommand level

triarius:
The API does not allow non-positive lifetimes. However, a combination of the cli framework we depend on, and the configuration parsing we have written means that there is no good way to distinguish between omitted configuration values and those that are set to the zero value for their type (e.g. 0 for integers). So we have to let the cli take a 0 valued lifetime, but that will not be passed on the API.

### DIFF
--- a/clicommand/oidc_request_token.go
+++ b/clicommand/oidc_request_token.go
@@ -58,14 +58,12 @@ var OIDCRequestTokenCommand = cli.Command{
 	Flags: []cli.Flag{
 		cli.StringFlag{
 			Name:  "audience",
-			Value: "",
-			Usage: "The audience that will consume the OIDC token",
+			Usage: "The audience that will consume the OIDC token. The API will supply a default audience if it is ommited.",
 		},
 		cli.IntFlag{
 			Name:  "lifetime",
 			Value: 0,
-			Usage: `The time (in seconds) the OIDC token will be valid for before expiry.
-Specfying 0 will result the in the API substituting its default value for the lifetime instead of 0.`,
+			Usage: "The time (in seconds) the OIDC token will be valid for before expiry. Must be a non-negative integer. Ommitting this flag or specifying 0 will result the in the API substituting its default value and NOT a non-expriing token.",
 		},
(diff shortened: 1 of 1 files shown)


clicommand/oidc_request_token.go
```diff
@@ -58,18 +58,14 @@ var OIDCRequestTokenCommand = cli.Command{
...
-			Value: "",
-			Usage: "The audience that will consume the OIDC token",
+			Usage: "The audience that will consume the OIDC token. The API will supply a default audience if it is ommited.",
 		},
 		cli.IntFlag{
 			Name:  "lifetime",
-			Value: 0,
-			Usage: `The time (in seconds) the OIDC token will be valid for before expiry.
-Specfying 0 will result the in the API substituting its default value for the lifetime instead of 0.`,
+			Usage: "The time (in seconds) the OIDC token will be valid for before expiry. Must be a non-negative integer. Ommitting this flag or specifying 0 will result the in the API substituting its default value and NOT a non-expriing token.",
```
DrJosh9000, line 65: Spelling: Omitting, non-expiring. There's an extra "the" ("the in the API").
I'm wondering if it could be phrased better... "If the flag is omitted or set to 0, the API will choose a default finite lifetime."?

clicommand/oidc_request_token.go
```diff
@@ -58,18 +58,14 @@ var OIDCRequestTokenCommand = cli.Command{
 	Flags: []cli.Flag{
 		cli.StringFlag{
 			Name:  "audience",
-			Value: "",
-			Usage: "The audience that will consume the OIDC token",
+			Usage: "The audience that will consume the OIDC token. The API will supply a default audience if it is ommited.",
```
DrJosh9000, line 61: Spelling: omitted

clicommand/oidc_request_token.go
triarius: That's more clear, thanks!

DrJosh9000 (approved):
LGTM 🥳

PR #1841, opened 2022-11-20
Install docker compose plugin in agent alpine docker images

ajoneil:
The `docker-compose` alpine package contains docker compose v1 as a standalone tool.
The `docker-cli-compose` alpine package contains docker compose v2 as a docker plugin.

Not having `docker-cli-compose` in the agent docker image causes it to fail with the docker compose buildkite plugin when using `cli-version: 2`.


triarius (approved):
Hi @ajoneil, thanks for your contribution!

While this does not provide docker compose v2 for the ubuntu based images, we think this is a good step in the right direction. I have edited the title of your PR to indicated that it is just for the alpine images, though.

PR #1847, opened 2022-11-23
agent/plugin: Doc comments and cleanups

DrJosh9000:
This PR adds a package comment and doc comments to all exported top-level identifiers to the plugin package. It also makes some opportunistic cleanups:

- All the tests are more consistent, using table-driven tests throughout
- ValidateResult stores errors directly rather than error strings, so as to be more useful under the forthcoming Go 1.20 multi-errors Unwrap convention
- New sentinel error values enable more semantic error checking
- Consolidated regexps: declaration and compilation are within single var block, and previously duplicated regexps are un-duplicated.

### DIFF
--- a/agent/plugin/definition.go
+++ b/agent/plugin/definition.go
@@ -14,8 +14,15 @@ import (
 	"github.com/qri-io/jsonschema"
 )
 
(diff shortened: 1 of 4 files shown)


triarius (approved):
🪅 yay for table tests!

PR #1848, opened 2022-11-24
🤫 WIP: Secret experiment

moskyb:
This is something that's been on my mind for a little while: how do we make secrets awesome in the agent?

At the moment, we rely on build scripts to fetch secrets at run time, but what if we could make it part of the agent lifecycle, and handle everything all nice?

This PR is an experiment to that end. It pretends that there are elements in the pipeline.yml that look like this:
```yaml
secret_providers:
  - id: ssm
    type: aws-ssm
    config:
      iam_role: arn:aws:iam::123456789012:role/buildkite-role
      assume_via_oidc: true

  - id: vault
    type: vault_kvv2
    config:
      address: https://vault.example.com
      mount_path: secret
      auth_type: kubernetes
      auth_config:
        role: buildkite
        service_account_token_path: /var/run/secrets/kubernetes.io/serviceaccount/token

secrets: # Global to build, every step has these secrets injected
  - env_var: WIDGET_CORP_API_KEY
    key: path/to/widget-corp-api-key
    provider_id: vault

  - file: ~/.ssh/id_rsa
    key: path/to/ssh-key
    provider_id: ssm

# and/or

steps:
  # ...

  - name: deploy
    key: deploy
    command: ".buildkite/steps/deploy.sh"
    secrets: # Local to step, merged with globals serverside
      - env_var: GITHUB_DEPLOY_TOKEN
        key: /path/to/github-deploy-token
        provider_id: vault
```

and provides the agent a way to fetch those secrets in such a way that the secret store itself is abstracted away from the buildkite backend - that is, the backend would have no access to customer secrets - and in such a way that multiple stores can be used.

This PR is a super duper POC: there's only one secret backend (SSM parameter store) implemented, there are no tests, and some things are completely broken - using OIDC to log into aws, and writing secrets to files.

Hopefully this PR shows more or less what i'd like to get at, but not necessarily soon.

### DIFF
--- a/secrets/provider_registry.go
+++ b/secrets/provider_registry.go
@@ -48,42 +48,32 @@ func NewProviderRegistryFromJSON(config ProviderRegistryConfig, jsonIn string) (
 
 func (pr *ProviderRegistry) FetchAll(configs []SecretConfig) ([]Secret, []error) {
 	secrets := make([]Secret, 0, len(configs))
-	secretsCh := make(chan Secret)
-
 	errors := make([]error, 0, len(configs))
-	errorsCh := make(chan error)
 
-	var wg sync.WaitGroup
+	var (
+		wg  sync.WaitGroup
+		mtx sync.Mutex
+	)
+
+	wg.Add(len(configs))
 	for _, c := range configs {
-		wg.Add(1)
 		go func(config SecretConfig) {
 			defer wg.Done()
 
 			secret, err := pr.Fetch(config)
-			if err != nil {
+			mtx.Lock()
+			defer mtx.Unlock()
 
-				errorsCh <- err
+			if err != nil {
+				errors = append(errors, err)
 				return
 			}
-			secretsCh <- secret
+
+			secrets = append(secrets, secret)
 		}(c)
 	}
 
-	go func() {
-		for err := range errorsCh {
-			errors = append(errors, err)
-		}
-	}()
-
-	go func() {
-		for secret := range secretsCh {
-			secrets = append(secrets, secret)
-		}
-	}()
-
 	wg.Wait()
-	close(secretsCh)
-	close(errorsCh)
 
 	return secrets, errors
 }
--- a/secrets/provider_registry.go
+++ b/secrets/provider_registry.go
@@ -1,0 +1,131 @@
+package secrets
+
+import (
+	"encoding/json"
+	"fmt"
+	"sync"
+
+	"github.com/buildkite/agent/v3/bootstrap/shell"
+	"github.com/puzpuzpuz/xsync"
+)
+
+type ProviderRegistry struct {
+	config ProviderRegistryConfig
+	// We have the two maps here because we only want to initialise a provider if a secret using that provider is used
+	// We shouldn't boot up any secrets providers that won't get used
+	candidates *xsync.MapOf[string, providerCandidate] // Candidates technically doesn't need to be a sync map as it's never altered after initialization, but i've made it one just for symmetry
+	providers  *xsync.MapOf[string, Provider]
+}
+
+type ProviderRegistryConfig struct {
+	Shell *shell.Shell
+}
+
+// NewProviderRegistryFromJSON takes a JSON string representing a slice of secrets.RawProvider, and returns a ProviderRegistry,
+// ready to be used to fetch secrets.
+func NewProviderRegistryFromJSON(config ProviderRegistryConfig, jsonIn string) (*ProviderRegistry, error) {
+	var rawProviders []providerCandidate
+	err := json.Unmarshal([]byte(jsonIn), &rawProviders)
+	if err != nil {
+		return nil, fmt.Errorf("unmarshalling secret providers: %w", err)
+	}
+
+	candidates := xsync.NewMapOf[providerCandidate]()
+	for _, provider := range rawProviders {
+		if _, ok := candidates.Load(provider.ID); ok {
+			return nil, fmt.Errorf("duplicate provider ID: %s. Provider IDs must be unique", provider.ID)
+		}
+
+		candidates.Store(provider.ID, provider)
+	}
+
+	return &ProviderRegistry{
+		config:     config,
+		candidates: candidates,
+		providers:  xsync.NewMapOf[Provider](),
+	}, nil
+}
+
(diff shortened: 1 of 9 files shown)


secrets/provider_registry.go
```diff
@@ -0,0 +1,131 @@
...
+
+	"github.com/buildkite/agent/v3/bootstrap/shell"
+	"github.com/puzpuzpuz/xsync"
+)
+
+type ProviderRegistry struct {
+	config ProviderRegistryConfig
+	// We have the two maps here because we only want to initialise a provider if a secret using that provider is used
+	// We shouldn't boot up any secrets providers that won't get used
+	candidates *xsync.MapOf[string, providerCandidate] // Candidates technically doesn't need to be a sync map as it's never altered after initialization, but i've made it one just for symmetry
```
moskyb, line 16: generics! 🎉

secrets/provider_registry.go
```diff
@@ -0,0 +1,131 @@
...
+	providers  *xsync.MapOf[string, Provider]
+}
+
+type ProviderRegistryConfig struct {
+	Shell *shell.Shell
```
moskyb, line 21: i don't love passing the bootstrap's shell into the provider registry, but it would be nice for it to be able to log things, and when it creates [`secrets.EnvSecret`s](https://github.com/buildkite/agent/pull/1848/files#diff-f427d9a2346cf79f74bece1e3483ad9d6c0cc7dad4cf7818de4db9f9658855f9R26), it needs to be able to pass the shell's env into them so that they can store themselves. This seems a bit nasty and i'd appreciate some C&C on how to make it a bit better

secrets/provider_registry.go
```diff
@@ -0,0 +1,131 @@
...
+		for err := range errorsCh {
+			errors = append(errors, err)
+		}
+	}()
+
+	go func() {
+		for secret := range secretsCh {
+			secrets = append(secrets, secret)
+		}
+	}()
```
moskyb, line 82: maybe this is getting too goroutiney. do we really need to fetch things in parallel? we can, so we might as well, was my justification, but i think the reality is that the number of secrets in each pipeline is gonna be pretty small, and they could probably be serialised.

secrets/secret_config.go
```diff
@@ -0,0 +1,38 @@
...
+import "fmt"
+
+type SecretConfig struct {
+	Key        string `json:"key"`
+	ProviderID string `json:"provider_id"`
+	EnvVar     string `json:"env_var"`
+	File       string `json:"file"`
+}
+
+func (s SecretConfig) Validate() error {
```
moskyb, line 12: whoops, i haven't used this function. nice to have around, though

secrets/provider_registry.go
moskyb: more external libraries 👎

secrets/secret.go
```diff
@@ -0,0 +1,69 @@
...
+func newFileSecret(config SecretConfig, value string) Secret {
+	return FileSecret{
+		FilePath: config.File,
+		Value:    value,
+	}
+}
+
+func (s FileSecret) Store() error {
+	return os.WriteFile(s.FilePath, []byte(s.Value), 0777)
+}
```
moskyb, line 63: this is broken at the moment, and i can't quite figure out why. i'll think about it a bit

secrets/provider_registry.go
DrJosh9000: The question will be: is there a significant benefit to using the library over either the inbuilt map with a sync.Mutex/sync.RWMutex, or using sync.Map (are there going to be that many concurrent map reads/writes that the performance is noticeably better?)

secrets/provider_registry.go
DrJosh9000: If you keep the parallelisation, might want to at least cap the number of worker goroutines at runtime.NumCPU or similar.

Also is it worth returning each error, or should it treat the whole operation as failed if any of them fail?

Also while you've been out, I've been busy sprinkling context.Context args through the codebase - can requests be cancelled early through use of a context somehow?

secrets/provider_registry.go
moskyb: yeah, this was originally implemented with a normal map and a `sync.RWMutex`, but there's a lot of reads and writes that require opening and closing the mutex a whole bunch, and the code got really ugly.

then i tried a `sync.Map`, but it's not typesafe, and casting everything gets really ugly...

my (completely unresearched) vibe on the number of secrets per pipeline is P95<10, P99<100, so i feel like we're unlikely to see a level of concurrency where we'd see a lot of lock contention, regardless of implementation.

the other thing is that i don't think my locking dynamics here are necessarily correct; i think that two secrets could initialise the same provider at the same time, but then only one of them would get stored, which isn't fab. i think this code needs more work regardless of what locking system we use.

TL;DR: i was experimenting a bit. i'm happy to make these stdlib `sync.Map`s, or a regular map behind a mutex, but i do kinda like having the fancy typesafe concurrent map.

secrets/provider_registry.go
moskyb: > If you keep the parallelisation, might want to at least cap the number of worker goroutines at runtime.NumCPU or similar

what would the benefit of limiting concurrency be? given that calls to secret managers are going to be IO bound, i'm happy to let greenthreading/goroutine magic handle throttling our concurrency.

> Is it worth returning each error, or should it treat the whole operation as failed if any of them fail?

I think it's probably most useful to do both - ie, any `len(errors) > 0` should be treated as a failure, but it's also nice to be able to tell the user exactly which secret failed to be fetched, and for what reason.

> can requests be cancelled early through use of a context somehow?

yep, i think it's totally reasonable to change the `Provider.Fetch` interface to accept a context. i'll make that change now :)

DrJosh9000: > what would the benefit of limiting concurrency be? given that calls to secret managers are going to be IO bound, i'm happy to let greenthreading/goroutine magic handle throttling our concurrency.

That's the happy path - what if they each open a connection simultaneously, exhausting TCP source ports? Or there's a config error?

secrets/secret.go
moskyb: turns out it was a race condition caused by me trying to get too fancy with channels 🙃 mutexes 4 lyf

DrJosh9000:
Closing due to staleness 😔

PR #1852, opened 2022-11-29
process: Fix flaky "tester-signal" tests

DrJosh9000:
These tests are occasional flakes because they rely on a sleep to let the subprocess set up the signal-catching. The fix instead waits for the subprocess to print "Ready" to stdout before interrupting/cancelling.


triarius (approved):
This is amazing!

PR #1854, opened 2022-11-30
agent: Timeout on finishJob and onUploadChunk

DrJosh9000:
This uses `context.WithTimeout` to enforce a timeout on `finishJob` and `onUploadChunk`. 

This is not very different from swapping `roko.TryForever` for `roko.WithMaxAttempts` and a suitable limit, but has two upsides: 1) there is no maths needed to figure out how many attempts to limit to in order to reach a particular duration, and 2) the context will abort both the roko sleep and any (probably doomed) HTTP request in progress.

Why 48 hours? It seems like a long enough time.

### DIFF
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -705,8 +705,8 @@ func (r *JobRunner) startJob(ctx context.Context, startedAt time.Time) error {
 	})
 }
 
-// Finishes the job in the Buildkite Agent API. This call will keep on retrying
-// forever until it finally gets a successfull response from the API.
+// finishJob finishes the job in the Buildkite Agent API. If the FinishJob call
+// cannot return successfully, this will retry for a long time.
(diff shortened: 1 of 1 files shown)


triarius (approved):
Very glad we got the contexts in so we could do this!

pda:
> there is no maths needed to figure out how many attempts to limit to in order to reach a particular duration


Ideally I think that should be solved in roko, by being able to specify a total retry duration and an interval function/strategy.

> the context will abort both the roko sleep and any (probably doomed) HTTP request in progress


That seems good and important: we're trying to prevent “agent got stuck forever” scenarios, we should be as interrupty as possible rather than waiting for a try to (hopefully) finish.

DrJosh9000:
> Ideally I think that should be solved in roko, by being able to specify a total retry duration and an interval function/strategy.

Yes - as discussed, sounds like potential for some new options in roko.

PR #1855, opened 2022-11-30
Ensure that unrecoverable errors for Hearbeat and Ping stop the agent

moskyb:
**Context:** A [recent incident](https://www.buildkitestatus.com/incidents/m8f0dnxtmjbz) at Buildkite caused an issue to occur with some agents where their connection state was invalidated,  and were marked as ‘lost’ in the buildkite backend. 

this caused them to get into a halfway-between state where the agent is still alive, but still unsuccessfully trying to talk to the buildkite agent API. because the agent is already past the point in its code where it would usually try to authenticate with the API, it just loops trying to connect to an API that will never successfully authenticate its requests.

the buildkite API was returning a 401 to all requests from the agent, and the agent would say "okay :)" and then retry a bit later, blissfully unaware that no requests would ever succeed.

when this  happens, it leaves the agent as a bit of a zombie; it’s still running, but it can’t pick up work and is basically useless to everyone.

**This PR:** Makes the agent die whenever it receives a non-retryable status from either the `Ping()` and `Heartbeat()` calls. This means that if agent connection states ever get invalidated in the future, the agents will fully die instead of becoming zombies.

It looks like this when it happens:
image

### DIFF
--- a/agent/agent_worker.go
+++ b/agent/agent_worker.go
@@ -2,6 +2,7 @@ package agent
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"strconv"
@@ -90,6 +91,30 @@ type AgentWorker struct {
 	retrySleepFunc func(time.Duration)
 }
 
+type errUnrecoverable struct {
+	action   string
+	response *api.Response
+	err      error
+}
+
+func (e *errUnrecoverable) Error() string {
+	status := "unknown"
+	if e.response != nil {
+		status = e.response.Status
+	}
+
+	return fmt.Sprintf("%s failed with unrecoverable status: %s, mesage: %q", e.action, status, e.err)
+}
+
+func (e *errUnrecoverable) Is(other error) bool {
+	_, ok := other.(*errUnrecoverable)
+	return ok
+}
+
+func (e *errUnrecoverable) Unwrap() error {
+	return e.err
+}
+
 // Creates the agent worker and initializes its API Client
 func NewAgentWorker(l logger.Logger, a *api.AgentRegisterResponse, m *metrics.Collector, apiClient APIClient, c AgentWorkerConfig) *AgentWorker {
 	return &AgentWorker{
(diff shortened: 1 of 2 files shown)


agent/agent_worker.go
```diff
@@ -362,18 +402,44 @@ func (a *AgentWorker) Heartbeat(ctx context.Context) error {
 // Performs a ping that checks Buildkite for a job or action to take
 // Returns a job, or nil if none is found
 func (a *AgentWorker) Ping(ctx context.Context) (*api.Job, error) {
-	ping, _, err := a.apiClient.Ping(ctx)
-	if err != nil {
+	ping, resp, pingErr := a.apiClient.Ping(ctx)
+	// wait a minute, where's my if err != nil block? TL;DR look for pingErr ~20 lines down
```
DrJosh9000, line 406: Good explanation 👍

Is err renamed to pingErr to make finding it across 20 lines easier? (That's reasonable, mostly curious)

The other thing I was wondering, and this is not necessarily a suggestion here, but could the ergonomics of apiClient.Ping be changed to make the gap unnecessary? I can see lifting errUnrecoverable into the API client, but handling ping.Message and ping.Action is definitely too business-logicy for that layer, which is the issue.

api/retryable.go
```diff
@@ -29,6 +29,10 @@ var retryableStatuses = []int{
 	http.StatusGatewayTimeout,      // 504
 }
 
+func IsFailure(r *Response) bool {
```
DrJosh9000, line 32: I would like more doc comments, please and thankyou 🙏

```suggestion
// IsFailure reports if the response represents a failure.
func IsFailure(r *Response) bool {
```

There's probably a better description than that, though?

DrJosh9000 (commented):
Looks pretty good!

agent/agent_worker.go
moskyb: > Is err renamed to pingErr to make finding it across 20 lines easier?

got it in one 👍

> but could the ergonomics of apiClient.Ping be changed to make the gap unnecessary?

yeah, i investigated this a little bit, and couldn't find a way to make it nice. i think part of the issue is that the apiclient exposes non-2xx status codes as go errors, which is something that i think is a little weird, and makes the ergonomics of failure cases a little strange. changing that now would be a major PITA though, so i think it's probably best to leave it for now

DrJosh9000 (approved):
😎🙌

PR #1860, opened 2022-12-01
Kubernetes job runner

benmoss:
Opening this up to get more 👀, still a work in progress, but building this out as part of our work in https://github.com/buildkite/agent-stack-k8s/tree/v2


benmoss:
@moskyb wdyt about reviewing this? Would you rather it be broken up into smaller PRs? I'm sure there will be more changes coming, but we shouldn't let this branch get too long in the tooth

PR #1861, opened 2022-12-05
Update dependabot.yml to include packaging Dockerfiles

triarius:
It was not updating the packaging Dockerfiles which are in subdirectories

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,7 +1,23 @@
 version: 2
 updates:
   - package-ecosystem: "docker"
-    directory: "/packaging/docker"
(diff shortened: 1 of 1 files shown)


DrJosh9000:
My only quibble is that "Update dependabot.yml" won't be very useful in `git log` output.

triarius:
Hmm, it looks like this is not possible as dependabot does not support recur sing into subdirectories: https://github.com/dependabot/dependabot-core/issues/2178

I'm going to hardcode the paths for all current docker images for now. But it looks like generating a dependabot.yaml is the only solution at this stage.

DrJosh9000 (approved):
LGTM 🥳

PR #1869, opened 2022-12-06
bootstrap: Replace mock ssh with container preparation

DrJosh9000:
This mainly changes a test from invoking a bintest mock, to invoking the real `ssh` and testing the string parsing logic separately. 

`resolveGitHost` is refactored into two parts: "invoking `ssh -G`" and "parsing the output" (`hostFromSSHG`).

The new tests will detect if SSH ever formats its output differently in a breaking way, and removes two instances of bintest, which may be the source of test flakes.

### DIFF
--- a/bootstrap/shell/test.go
+++ b/bootstrap/shell/test.go
@@ -26,20 +26,20 @@ func NewTestShell(t *testing.T) *Shell {
 
 	// Windows requires certain env variables to be present
 	if runtime.GOOS == "windows" {
-		sh.Env = env.FromSlice([]string{
-			//			"PATH=" + os.Getenv("PATH"),
-			"SystemRoot=" + os.Getenv("SystemRoot"),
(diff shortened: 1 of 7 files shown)


bootstrap/shell/test.go
```diff
@@ -26,20 +26,20 @@ func NewTestShell(t *testing.T) *Shell {
...
+			"ProgramData": os.Getenv("ProgramData"),
+		}
 	} else {
-		sh.Env = env.FromSlice([]string{
-			"PATH=" + os.Getenv("PATH"),
-		})
+		sh.Env = env.Environment{
+			"PATH": os.Getenv("PATH"),
+		}
 	}
```
moskyb, line 43: 👍 love it

moskyb (approved):
CleanShot 2022-12-07 at 12 43 30@2x

:chefskiss:

PR #1870, opened 2022-12-06
Add agent tags from ECS container metadata

francoiscampbell:
Adds selected ECS metadata as agent tags. The metadata that I felt was useful to add was:
- Container name
- Docker image
- ECS task ARN

Tested a custom build on our infrastructure:
image

### DIFF
--- a/agent/ecs_meta_data.go
+++ b/agent/ecs_meta_data.go
@@ -1,0 +1,32 @@
+package agent
+
+import (
+	"context"
+	metadata "github.com/brunoscheufler/aws-ecs-metadata-go"
+	"net/http"
+)
+
+type ECSMetadata struct {
+}
+
+func (e ECSMetadata) Get() (map[string]string, error) {
+	metaData := make(map[string]string)
+
+	ecsMeta, err := metadata.GetContainer(context.Background(), &http.Client{})
+	if err != nil {
+		return metaData, err
+	}
+
+	switch m := ecsMeta.(type) {
+	case *metadata.ContainerMetadataV3:
+		metaData["ecs:container-name"] = m.DockerName
+		metaData["ecs:image"] = m.Image
+		metaData["ecs:task-arn"] = m.Labels.EcsTaskArn
+	case *metadata.ContainerMetadataV4:
+		metaData["ecs:container-name"] = m.DockerName
+		metaData["ecs:image"] = m.Image
+		metaData["ecs:task-arn"] = m.Labels.EcsTaskArn
+	}
+
+	return metaData, nil
+}
--- a/agent/ecs_meta_data.go
+++ b/agent/ecs_meta_data.go
@@ -2,6 +2,7 @@ package agent
 
 import (
 	"context"
+	"fmt"
 	metadata "github.com/brunoscheufler/aws-ecs-metadata-go"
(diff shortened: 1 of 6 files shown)


agent/tags_test.go
```diff
@@ -82,6 +82,32 @@ func TestFetchingTagsFromEC2Tags(t *testing.T) {
 		[]string{"custom_tag=true"})
 }
 
+func TestFetchingTagsFromECS(t *testing.T) {
+	fetcher := &tagFetcher{
+		ecsMetaDataDefault: func() (map[string]string, error) {
+			return map[string]string{
+				`ecs:container-name`: "ecs-buildkite-agent-blahblah",
+				`ecs:image`:          "buildkite/agent",
```
moskyb, line 90: super nit here:
```suggestion
				"ecs:container-name": "ecs-buildkite-agent-blahblah",
				"ecs:image":          "buildkite/agent",
```

i realise that there's a bunch of prior art in the repo of using backtick strings for hash keys(? i guess? seems like a weird choice from my forebears), but it's best to avoid them unless we actually need them in future, i think.

agent/ecs_meta_data.go
```diff
@@ -0,0 +1,32 @@
...
+	switch m := ecsMeta.(type) {
+	case *metadata.ContainerMetadataV3:
+		metaData["ecs:container-name"] = m.DockerName
+		metaData["ecs:image"] = m.Image
+		metaData["ecs:task-arn"] = m.Labels.EcsTaskArn
+	case *metadata.ContainerMetadataV4:
+		metaData["ecs:container-name"] = m.DockerName
+		metaData["ecs:image"] = m.Image
+		metaData["ecs:task-arn"] = m.Labels.EcsTaskArn
+	}
```
moskyb, line 29: looking at the code in [aws-ecs-metadata-go](https://github.com/brunoscheufler/aws-ecs-metadata-go"), it shouldn't be possible for it to return anything other than `ContainerMetadataV3` or `ContainerMetadataV4`, but it might still be worth adding 

```suggestion
	default:
		return fmt.Errorf("ecs metadata returned unknown type %T", m)
	}
```

just in case something changes

moskyb (commented):
thanks for your contribution @francoiscampbell! i have two minor nits to fix, but after that i reckon we're all good to go - this seems like it'd be super useful for people running the agent on ECS!

reviewing this PR makes me think that we should rejig the way the agent gets tags at some point - atm users have to tell the agent `--tags-from-here --tags-from-there --tags-from-another-place` to get those tags. maybe we could make it autodetect what kind of environment it's running in in future? i dunno.

☝️ none of this is actually relevant to this review - in no way am i expecting you to implement this! i'm just thinking about how we could do things better in future.

francoiscampbell:
Thanks for the comments, updated

moskyb:
i've rebased this from main and am running CI on it now. thanks so much for your contribution @francoiscampbell!

francoiscampbell:
Hey @moskyb, I've rebased this, are we still interested in merging?

moskyb:
hey, sorry! this got caught up in the end-of-year rush. CI is passing, and i'll merge it now :)

moskyb (approved):

PR #1871, opened 2022-12-06
Remove unnecessary uses of backtick strings

moskyb:
There's a very weird style within the agent codebase of using backtick strings for a bunch of weird stuff that i don't understand the motivation for. In general, in go, backtick strings should be used only when the user wants unescaped output, or for when adding escaping to a string would make it harder to read.

Within the agent codebase, backtick strings seem to get used for:
- map keys
- command strings
- case block conditions

... and much more. it's kinda ugly, and it causes contributors to follow the established precedent, which makes the codebase uglier as time goes on.

**This PR:** Takes many of the useless uses of backtick strings and replaces them with nice, normal double quotes.

This PR was implemented by running 
```Regex
`([^"\\{}]*?)` # Look for all backtick strings that don't contain ", \, { or } (valid uses of a backtick string)
(?![^"]*") # That aren't surrounded by a normal double quote string - ie, a string like "hello `there`" should remain unchanged
```
across the codebase, and replacing with `"$1"`.

I then did a manual check of all the changes - there were a bunch of false positives - backtick strings in comments should stick around, and i couldn't be bothered making the regex even more horrible.

**This PR is going to be super boring to review, but it's important to do it carefully.** There's every chance that the regex is wrong in some subtle way, and we've screwed everything up forever.


DrJosh9000 (approved):
Thank you 🙌

PR #1872, opened 2022-12-07
Use stdlib errors instead of github.com/pkg/errors

DrJosh9000:
github.com/pkg/errors is archived, and the README says it is in maintenance mode. The standard library `errors` package is built-in.

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -7,6 +7,7 @@ package bootstrap
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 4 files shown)


moskyb (changes requested):
i'm 100% on board with removing pkg/errors, but i'm a little bit hesitant about changing the error messages here.

while the standing wisdom is to leave errors unacapitalised, not to use `Error: ...` language etc, the errors in `bootstrap.go` tend to be printed out verbatim, and are clearer as they are than if they were corrected to be more "go standard"

the difference between
```
ERROR 🚨: Could not open /path/to/some/file for appending: ...
# and
ERROR 🚨: open /path/to/some/file for appending: ...
```
is slight, but i think it's important to present errors to humans (in this case people reading the build logs to see why their build broke) in grammared english, rather than (in my opinion) oversimplifying error messages so that the compiler is happier with us.

i'm absolutely not going to die on this hill - this is mostly bikeshedding, and as mentioned above, the difference is pretty slight. i do think it's something that's worth thinking and chatting about, though.

DrJosh9000:
Thanks. I've changed the error strings back. I was changing them a bit mindlessly - but I'd love to change them all, and in doing so make sure they are printed with some adorning context.

PR #1873, opened 2022-12-08
An in-built hierarchical status page

DrJosh9000:
I wanted to "see" what the agent itself thinks it is doing, without watching the log output, independently of what the service thinks the agent is doing. So I started implementing a /status handler inside the health check server.

The new endpoint is behind both the `--health-check-addr` flag and the `inbuilt-status-page` experiment.

The structure of the page is hierarchical, using `<details> / <summary>` tags - kind of a fusion of cohost.org with a typical Google /statusz page. To aid in not cluttering the main code that needs to report statuses, the hierarchy is based on contexts, which are--helpfully--already threaded through most of the agent. Adding an item creates a new context to which sub-items can be added. There are two kinds of items. Simple items report only a string, but benefit from a setStatus function that takes care of all locking. The more serious kind of item is based on a custom HTML template fragment, and assumes the caller is in control of locking the data needed to fill the template. If you make a mistake in writing the template, or the callback returns an error, the error is reported through the status page itself, rather than logging or crashing the program.

Screenshot 2022-12-14 at 5 48 59 pm


triarius (commented):
Looks amazing! Just one minor thing about the test, happy to be convinced it's not necessary.

PR #1649, opened 2022-05-24
Integrate Test Analytics

keithpitt:
As per the title

PR #1881, opened 2022-12-15
Use test-collector plugin

DrJosh9000:
A new attempt at using Test Analytics via the collector plugin (see #1649).
For now the collectors are using a single TA token. Perhaps we should do it with multiple tokens like #1649 has?


triarius (approved):
🎉

Issue #1886, opened 2022-12-20
Add ability to suppress fatal if `buildkite-agent artifact search ...` doesn't find any results


MatthewDolan:
**Is your feature request related to a problem? Please describe.**
When running `buildkite-agent artifact search ...`, if no results are found, the process fatals and an error message is printed. Example:

```
2022-12-20 18:39:39 INFO   Searching for artifacts: "${search_query}"
2022-12-20 18:39:40 FATAL  No matches found for "${search_query}"
🚨 Error: The command exited with status 1
```

The behavior was surprising. The code for this behavior can be found at https://github.com/buildkite/agent/blob/d392bca3c57702f211ea0a9cfd94caaefcd5536c/clicommand/artifact_search.go#L167-L169

**Describe the solution you'd like**
It would be helpful to add an option (e.g. `--allow-empty-results`) which suppresses the fatal exit. Instead, it would return am empty result (i.e. no lines of output).

**Describe alternatives you've considered**
- Considered having the calling script catch / trap the exit code, but that felt less intuitive than returning a successful empty result set.
- Considered changing the behavior `buildkite-agent artifact search ...` (without adding an option) but that feels too disruptive to existing scripts which might be relying on this behavior.

**Additional context**
This would be useful for builds which might alter behavior based on whether an artifact is present. Example:

```
artifact_path=$(buildkite-agent artifact search "${search_query}" --format "%p")
if [[ -n "${artifact_path}" ]]; then
  echo "Searched for plan artifact and found: ${artifact_path}"
  buildkite-agent artifact download "${artifact_path}" .
else
  echo "No artifact to download."
fi
```

A concrete example might be a `terraform apply` step which conditionally downloads and applies a prior plan if one is present, otherwise it applies `--auto-approve` if no plan is present.

PR #1887, opened 2022-12-20
feat: add flag to allow empty results with doing an artifact search

MatthewDolan:
See issue https://github.com/buildkite/agent/issues/1886 for motivation for the change.

This felt like the easiest way to address it, although it was tricky to find a name for the option that matched the problem well. Alternative name suggestions would be appreciated and I would be happy to update the PR.

### DIFF
--- a/clicommand/artifact_search.go
+++ b/clicommand/artifact_search.go
@@ -51,6 +51,7 @@ type ArtifactSearchConfig struct {
 	Step               string `cli:"step"`
 	Build              string `cli:"build" validate:"required"`
 	IncludeRetriedJobs bool   `cli:"include-retried-jobs"`
+	AllowEmptyResults  bool   `cli:"allow-empty-results"`
 	PrintFormat        string `cli:"format"`
 
 	// Global flags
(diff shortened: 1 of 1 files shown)


clicommand/artifact_search.go
```diff
@@ -165,7 +170,11 @@ Format specifiers:
 		}
 
 		if len(artifacts) == 0 {
-			l.Fatal(fmt.Sprintf("No matches found for %q", cfg.Query))
+			if cfg.AllowEmptyResults {
+				l.Info(fmt.Sprintf("No matches found for %q", cfg.Query))
+			} else {
+				l.Fatal(fmt.Sprintf("No matches found for %q", cfg.Query))
```
moskyb, line 176: this isn't your fault as it's inspired by pre-existing code, but our logger functions automatically Sprintf-ify things:
```suggestion
			if cfg.AllowEmptyResults {
				l.Info("No matches found for %q", cfg.Query)
			} else {
				l.Fatal("No matches found for %q", cfg.Query)
```

moskyb (approved):
LGTM! there's a minor nit here that you're welcome to fix or not :)

PR #1891, opened 2022-12-24
add git-checkout-flags

### LINKED ISSUE #1888: git-checkout-flags
Our host machines don't have git-lfs installed but we invoke git-lfs from within a container. This becomes problematic because git-lfs installs global hooks which break the `git checkout` on the host.

 I would prefer to checkout with ` --no-verify` by default here without maintaining a special preparing checkout. Should be similar to `git-fetch-flags`.

 In our pre-checkout hook, we already attempt to remove this file to some success, but sometimes it fails. I don't recall approaching this from the git-lfs side i.e. preventing it from installing hooks in the first place etc., but I think `--no-verify` is the least potentially problematic solution.

jmelahman:
Unfortunately, turns out `git checkout` doesn't have a `--no-verify` option, so not immediately useful to me. Opening in case it's useful for others.

Closes #1888

### DIFF
--- a/agent/agent_configuration.go
+++ b/agent/agent_configuration.go
@@ -11,6 +11,7 @@ type AgentConfiguration struct {
 	GitMirrorsLockTimeout      int
 	GitMirrorsSkipUpdate       bool
 	PluginsPath                string
+	GitCheckoutFlags           string
 	GitCloneFlags              string
 	GitCloneMirrorFlags        string
 	GitCleanFlags              string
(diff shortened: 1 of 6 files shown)


triarius:
Thanks for the contribution @jmelahman. If you can rebase onto the main branch and rebase away the `typo` and `fmt` commits, we can get this merged.

jmelahman:
> Thanks for the contribution @jmelahman. If you can rebase onto the main branch and rebase away the `typo` and `fmt` commits, we can get this merged.

Done. Thanks

triarius (approved):

Issue #1880, opened 2022-12-14
BUILDKITE_AGENT_ACQUIRE_JOB with job in wait status exits


ChrisBr:
If you start an agent with `BUILDKITE_AGENT_ACQUIRE_JOB` and the job is still waiting for other jobs to finish, the agent will stop immediately with `422 Job $ID is no longer eligible to be assigned an agent`. 

We would like to provision agents when they're still waiting for performance reason so I would expect the agent to just be idle until the job becomes scheduled or die if it doesn't after `BUILDKITE_AGENT_DISCONNECT_AFTER_IDLE_TIMEOUT`.

Also the error message is in the past (`no longer eligible`) but the job was never eligible.

PR #1894, opened 2022-12-30
Retry AcquireJob when the job is locked

triarius:
An upcoming change to the API will return 423 Locked when the job is in the waiting state.
This PR will allow the agent to retry according to `Retry-After` headers sent by the API in this situation. This will time out after 270s (4 min 30 s).

If there were no delays or jitter, the attempts would happen at t = 0, 1, 2, 4, ..., 128s
after the initial one. Therefore, there are 9 attempts taking at least 255s. If the jitter
always hit the max of 1s, then another 8s is added to that. This is still comfortably within
the timeout of 270s, and the bound seems tight enough so that the agent is not wasting time
waiting for a retry that will never happen.

In order to activate this behaviour on the API, the header `X-Buildkite-Lock-Acquire-Job` has to be set to 1, which this PR does. However, a feature flag also needs to be activated, so this ability is by no means widely available.

This is what it looks like when run:
```
2023-01-04 17:08:43 INFO   quirinius Connecting to Buildkite...
2023-01-04 17:08:43 INFO   quirinius Attempting to acquire job 01857b65-437d-4701-955a-44dc38de8bb6...
2023-01-04 17:08:43 WARN   quirinius The job is waiting for a dependency (PUT http://agent.buildkite.localhost/v3/jobs/01857b65-437d-4701-955a-44dc38de8bb6/acquire: 423 Job 01857b65-437d-4701-955a-44dc38de8bb6 is not yet eligible to be assigned an agent)
2023-01-04 17:08:45 WARN   quirinius The job is waiting for a dependency (PUT http://agent.buildkite.localhost/v3/jobs/01857b65-437d-4701-955a-44dc38de8bb6/acquire: 423 Job 01857b65-437d-4701-955a-44dc38de8bb6 is not yet eligible to be assigned an agent)
2023-01-04 17:08:47 WARN   quirinius The job is waiting for a dependency (PUT http://agent.buildkite.localhost/v3/jobs/01857b65-437d-4701-955a-44dc38de8bb6/acquire: 423 Job 01857b65-437d-4701-955a-44dc38de8bb6 is not yet eligible to be assigned an agent)
2023-01-04 17:08:52 WARN   quirinius The job is waiting for a dependency (PUT http://agent.buildkite.localhost/v3/jobs/01857b65-437d-4701-955a-44dc38de8bb6/acquire: 423 Job 01857b65-437d-4701-955a-44dc38de8bb6 is not yet eligible to be assigned an agent)
2023-01-04 17:09:01 WARN   quirinius The job is waiting for a dependency (PUT http://agent.buildkite.localhost/v3/jobs/01857b65-437d-4701-955a-44dc38de8bb6/acquire: 423 Job 01857b65-437d-4701-955a-44dc38de8bb6 is not yet eligible to be assigned an agent)
2023-01-04 17:09:18 WARN   quirinius The job is waiting for a dependency (PUT http://agent.buildkite.localhost/v3/jobs/01857b65-437d-4701-955a-44dc38de8bb6/acquire: 423 Job 01857b65-437d-4701-955a-44dc38de8bb6 is not yet eligible to be assigned an agent)
2023-01-04 17:09:50 WARN   quirinius The job is waiting for a dependency (PUT http://agent.buildkite.localhost/v3/jobs/01857b65-437d-4701-955a-44dc38de8bb6/acquire: 423 Job 01857b65-437d-4701-955a-44dc38de8bb6 is not yet eligible to be assigned an agent)
2023-01-04 17:10:54 WARN   quirinius The job is waiting for a dependency (PUT http://agent.buildkite.localhost/v3/jobs/01857b65-437d-4701-955a-44dc38de8bb6/acquire: 423 Job 01857b65-437d-4701-955a-44dc38de8bb6 is not yet eligible to be assigned an agent)
2023-01-04 17:13:02 WARN   quirinius The job is waiting for a dependency (PUT http://agent.buildkite.localhost/v3/jobs/01857b65-437d-4701-955a-44dc38de8bb6/acquire: 423 Job 01857b65-437d-4701-955a-44dc38de8bb6 is not yet eligible to be assigned an agent)
2023-01-04 17:13:43 INFO   quirinius Disconnecting...
2023-01-04 17:13:43 INFO   quirinius Disconnected
2023-01-04 17:13:43 FATAL  Failed to acquire job: context deadline exceeded
```

Fixes: #1880


agent/agent_worker.go
```diff
@@ -537,20 +547,36 @@ func (a *AgentWorker) AcquireAndRunJob(ctx context.Context, jobId string) error
...
+					// Buildkite rejected the finish for some reason.
+					a.logger.Warn("Buildkite rejected the call to acquire the job (%s)", err)
+					r.Break()
+				case http.StatusLocked:
+					// If the API returns with a 423, the job is in the waiting state
+					a.logger.Warn("The job is waiting for a dependency (%s)", err)
+					duration, errParseDuration := time.ParseDuration(response.Header.Get("Retry-After") + "s")
+					if errParseDuration != nil {
+						duration = time.Second + time.Duration(rand.Int63n(int64(time.Second)))
+					}
```
moskyb, line 570: i was going to write a comment about using `strconv.Atoi` to avoid using the weird `+ "s"` thingo, but as it turns out, everything that did that was longer and/or jankier than what you've written here. good code! 👍

agent/agent_worker_test.go
```diff
@@ -106,3 +108,67 @@ func TestDisconnectRetry(t *testing.T) {
...
+
+			if backoff_seq > 4 {
+				cancel()
+			}
+
+			exptectedSleeps = append(exptectedSleeps, sleep)
+
+			rw.Header().Set("Retry-After", fmt.Sprintf("%f", delay))
+			rw.WriteHeader(http.StatusLocked)
+			fmt.Fprintf(rw, `{"message": "Job waitinguuid is not yet eligible to be assinged" }`)
```
moskyb, line 140: super nit:
```suggestion
			fmt.Fprintf(rw, `{"message": "Job waitinguuid is not yet eligible to be assigned"}`)
```

agent/agent_worker_test.go
```diff
@@ -106,3 +108,67 @@ func TestDisconnectRetry(t *testing.T) {
...
+	defer server.Close()
+
+	client := api.NewClient(logger.Discard, api.Config{
+		Endpoint: server.URL,
+		Token:    "llamas",
+	})
+
+	l := logger.NewBuffer()
+
+	retrySleeps := make([]time.Duration, 0)
```
moskyb, line 154: this is equivalent to 
```Go
retrySleeps := []time.Duration{}
```
and looks a bit odd, but isn't actually a problem, making it a [nit] i guess

api/client.go
```diff
@@ -139,12 +139,22 @@ func (c *Client) FromPing(resp *Ping) *Client {
...
 // Relative URLs should always be specified without a preceding slash. If
 // specified, the value pointed to by body is JSON encoded and included as the
 // request body.
-func (c *Client) newRequest(ctx context.Context, method, urlStr string, body interface{}) (*http.Request, error) {
+func (c *Client) newRequest(
+	ctx context.Context,
+	method, urlStr string,
+	body interface{},
+	headers ...Header,
+) (*http.Request, error) {
```
moskyb, line 157: this method signature is probably getting to the point where it's worth factoring it into a request struct or something:
```Go
type Request struct {
	Method  string
	URL     string
	Body    any
	Headers []Header
}

// ...

func (c *Client) newRequest(ctx context.Context, req Request) (*http.Request, error) { 
	// ...
}
```
... but for now i don't think we need to change it, and it would mean a bunch refactoring in everything that uses this method (which is basically everything in this package)

triarius: Yeah, we specifically chose this array of headers signature so that we didn't have to change every method call. I'll make a ticket to do this refactor.

agent/agent_worker.go
triarius: It would be nice if `time.ParseDuration` had a cleaner way to specify a unit.

agent/agent_worker_test.go
triarius: Done

agent/agent_worker_test.go
```diff
@@ -106,3 +108,61 @@ func TestDisconnectRetry(t *testing.T) {
...
+		agent:              nil,
+		apiClient:          client,
+		agentConfiguration: AgentConfiguration{},
+		retrySleepFunc:     retrySleepFunc,
+	}
+
+	err := worker.AcquireAndRunJob(ctx, "waitinguuid")
+	assert.ErrorContains(t, err, "423")
+
+	// the last Retry-After is not recorded as the retries exists before using it
```
triarius, line 162:
```suggestion
	// the last Retry-After is not recorded as the retries loop exits before using it
```

moskyb (approved):
looks great! awesome work 🎉

PR #1895, opened 2022-12-31
Add git config protocol.file.allow=always to tests

triarius:
This was disabled upstream due to a [security vulnerability in git](https://github.blog/2022-10-18-git-security-vulnerabilities-announced/#cve-2022-39253). That caused the test TestCheckingOutLocalGitProjectWithSubmodules to fail. It has only been passing in CI because we run a version of git in CI that did not receive the patch.

One solution would have been to set `protocol.file.allow=always` globally. However, that would mean that anyone who ran the tests would expose their environment to that vulnerability, and permanently if the global config were not unset afterwards.

Thus, in this PR, the config is only applied for the commands that need it. As one such command was in the default checkout hook, a configuration option was added so that the config would only be enabled in the test that needed it.

There is also some very minor clone cleanup in a separate commit.

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -1416,7 +1416,7 @@ func (b *Bootstrap) defaultCheckoutPhase(ctx context.Context) error {
 		for _, config := range b.GitSubmoduleCloneConfig {
 			args = append(args, "-c", config)
 		}
-		args = append(args, []string{"submodule", "update", "--init", "--recursive", "--force"}...)
+		args = append(args, "submodule", "update", "--init", "--recursive", "--force")
 		if err := b.shell.Run(ctx, "git", args...); err != nil {
 			return err
 		}
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -1412,7 +1412,12 @@ func (b *Bootstrap) defaultCheckoutPhase(ctx context.Context) error {
 			}
 		}
 
-		if err := b.shell.Run(ctx, "git", "submodule", "update", "--init", "--recursive", "--force"); err != nil {
+		args := []string{}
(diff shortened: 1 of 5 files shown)


bootstrap/bootstrap.go
```diff
@@ -1412,7 +1412,12 @@ func (b *Bootstrap) defaultCheckoutPhase(ctx context.Context) error {
 			}
 		}
 
-		if err := b.shell.Run(ctx, "git", "submodule", "update", "--init", "--recursive", "--force"); err != nil {
+		args := []string{}
+		for _, config := range b.GitSubmoduleCloneConfig {
+			args = append(args, "-c", config)
+		}
+		args = append(args, []string{"submodule", "update", "--init", "--recursive", "--force"}...)
```
DrJosh9000, line 1419:
```suggestion
		args = append(args, "submodule", "update", "--init", "--recursive", "--force")
```

DrJosh9000 (commented):
Looking good!

bootstrap/bootstrap.go
triarius: I should have known there would be a simpler way to do that in go 🤦‍♂️

bootstrap/config.go
```diff
@@ -83,6 +83,9 @@ type Config struct {
 	// Flags to pass to "git clean" command
 	GitCleanFlags string `env:"BUILDKITE_GIT_CLEAN_FLAGS"`
 
+	// Flags to pass to "git" when submodule init commands are invoked
```
triarius, line 86:
```suggestion
	// Config key-value pairs to pass to "git" when submodule init commands are invoked
```

DrJosh9000 (approved):
LGTM 👍

PR #1901, opened 2023-01-09
Build and Publish Docker Images for linux/arm64

triarius:
Currently, we only publish Docker images for amd64 on DockerHub. This introduces multi architecture images and publishes an additional arm64 image. This will allow customers to run the agent in Docker on arm64 based machines.

In this PR, we made the decision to build both linux/amd64 and linux/arm64 images on an intel EC2 instance, and push both images to the same tag in the registry as a [manifest list](https://docs.docker.com/registry/spec/manifest-v2-2/#manifest-list). `buildx` provides a way to do this using `qemu`.

We then introduce a new step in CI where each docker image is tested on EC2 instances capable of running the images natively (i.e. graviton instances are used to test the linux/arm64 images).

Finally, in packaging, both images are copied from ECR to dockerhub using the [skopeo](https://github.com/containers/skopeo) tool from the podman developers, that is capable of coping the images between repos without having to hand roll a manifest list.


.buildkite/steps/build-docker-image.sh
```diff
@@ -1,10 +1,13 @@
-#!/bin/bash
-set -euo pipefail
+#!/usr/bin/env bash
+
+set -Eeufo pipefail
```
moskyb, line 3: TIL `set -E`
>`-E`
>If set, any trap on ERR is inherited by shell functions, command substitutions, and commands executed in a subshell environment. The ERR trap is normally not inherited in such cases.

moskyb (approved):
LGTM! love your work 🎉

matthewborden:
Images are being published for both architectures here: https://hub.docker.com/r/buildkite/agent/tags?page=1&name=edge-build-5071.

I was able to successfully run the arm64 image on my M1 Macbook Pro:

```
docker run buildkite/agent:edge-build-5071-ubuntu-18.04 --help
Usage:

  buildkite-agent <command> [options...]

Available commands are:

  start       Starts a Buildkite agent
  annotate    Annotate the build page within the Buildkite UI with text from within a Buildkite job
  annotation  Make changes an annotation on the currently running build
  artifact    Upload/download artifacts from Buildkite jobs
  meta-data   Get/set data from Buildkite jobs
  oidc        Interact with Buildkite OpenID Connect (OIDC)
  pipeline    Make changes to the pipeline of the currently running build
  step        Get or update an attribute of a build step
  env         Prints out the environment of the current process as a JSON object
  bootstrap   Run a Buildkite job locally
  help        Shows a list of commands or help for one command

Use "buildkite-agent <command> --help" for more information about a command.
```

.buildkite/steps/build-docker-image.sh
```diff
@@ -91,14 +91,7 @@ docker buildx build --progress plain --builder "$builder_name" --platform linux/
...
 # Sanity check test before pushing. Only works for current arch. CI will test all arches as well.
-case $variant in
-sidecar)
-  echo "Skipping tests for sidecar variant"
-  ;;
-*)
-  .buildkite/steps/test-docker-image.sh "$variant" "$image_tag" "$(uname -m)"
-  ;;
-esac
+.buildkite/steps/test-docker-image.sh "$variant" "$image_tag" "$(uname -m)"
```
moskyb, line 74: 🎉 love it

moskyb (approved):
Awesome work. Before merging, could you rebase this down into a couple of commits?

.buildkite/steps/build-docker-image.sh
triarius: No probs @moskyb, @matthewborden I'll do the rebasing

triarius:
Rebased to reduce commits. The diff between the old and new is something that was applied to main in between:
```
🕙 16:10:14 ❯ git diff 0e0250..346fc9d
status/status.go --- Go
34         // Errors ignored below, as the status page is "best effort".                                                                                        34         // Errors ignored below, as the status page is "best effort".
35         hostname, _ = os.Hostname()                                                                                                                          35         hostname, _ = os.Hostname()
36         username    = func() string {                                                                                                                        36         username    = func() string {
37                 user, _ := user.Current()                                                                                                                    37                 user, err := user.Current()
..                                                                                                                                                              38                 if err != nil {
..                                                                                                                                                              39                         return fmt.Sprintf("unknown (uid=unknown; error=%v)", err)
..                                                                                                                                                              40                 }
38                 return fmt.Sprintf("%s (%s)", user.Username, user.Uid)                                                                                       41                 return fmt.Sprintf("%s (uid=%s)", user.Username, user.Uid)
39         }()                                                                                                                                                  42         }()
40         exepath, _ = os.Executable()                                                                                                                         43         exepath, _ = os.Executable()
41         startTime  = time.Now()                                                                                                                              44         startTime  = time.Now()
```

PR #1904, opened 2023-01-10
[WIP] agent config generics refactor thing

moskyb:


moskyb:
we're probably going to do something slightly similar to this as part of our ongoing project to make testing the CLI layer better cc/ @triarius

triarius:
> we're probably going to do something slightly similar to this as part of our ongoing project to make testing the CLI layer better cc/ @triarius 

Yeah after we refactor the Configs to be created in a unified manner we can abstract out common fields.

PR #1905, opened 2023-01-10
Replace all usages of interface{} with any

moskyb:
We currently use a mix of `interface{}` and `any` to refer to values of any type in our codebase. For the sake of consistency, i've run an `s/interface{}/any/g` over the codebase.

This is the world's most useless change, i'm not really fussed if it doesn't make it in, but i do much prefer `any` over `interface{}`. Knowing that the empty interface means any type (and, indeed that the type `interface{}` refers to an empty interface and not something else, and that an empty interface matches any type) requires a bit more go knowledge than seeing the word `any` and thinking "oh that can be any type"


DrJosh9000 (approved):
Yes...ha ha, ha... yes!!

PR #1910, opened 2023-01-16
status: Handle user.Current error

### LINKED ISSUE #1906: Agent panics when user not found
I'm running the agent in a container where the user is not in `/etc/passwd`, and so `user.Current()` returns an error, but since that is [ignored and then Sprintf-ed](https://github.com/buildkite/agent/blob/c8419aec2fc867284be8d56c63ff0825e521daa0/status/status.go#L36-L39), we get this panic:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x30cbd8]

goroutine 1 [running]:
github.com/buildkite/agent/v3/status.glob..func1()
        /Users/mossity/workspace/superorbital/agent-stack-k8s/agent/status/status.go:38 +0x28
github.com/buildkite/agent/v3/status.init()
        /Users/mossity/workspace/superorbital/agent-stack-k8s/agent/status/status.go:39 +0x5c
```

cc @DrJosh9000 

Repro:
```bash
GOOS=linux go build -o bk
docker run -v $PWD:/workspace --rm --entrypoint /workspace/bk bitnami/kubectl:latest
```

DrJosh9000:
Fixes #1906.

### DIFF
--- a/status/status.go
+++ b/status/status.go
@@ -34,8 +34,11 @@ var (
 	// Errors ignored below, as the status page is "best effort".
 	hostname, _ = os.Hostname()
 	username    = func() string {
-		user, _ := user.Current()
-		return fmt.Sprintf("%s (%s)", user.Username, user.Uid)
+		user, err := user.Current()
+		if err != nil {
+			return fmt.Sprintf("unknown (uid=unknown; error=%v)", err)
+		}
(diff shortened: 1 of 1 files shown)


moskyb (approved):
🎉 send it

PR #1918, opened 2023-01-17
Make dependabot ignore ubuntu docker image versions 

triarius:
We could pin the major and minor versions, but that may not work https://github.com/dependabot/dependabot-core/pull/6115.
Instead, because there is no patch version, dependabot should actually
never make a PR for these Dockerfiles. Thus, we can ignore them entirely


triarius:
Hmm, this might not work. But I have a plan.

triarius:
I guess it cannot hurt?

DrJosh9000 (approved):
Sounds good to me

PR #1920, opened 2023-01-18
CLI: 'env' command moved to 'env dump' subcommand

pda:
Move the `buildkite-agent env` command (which dumps environment to stdout) to `buildkite-agent env dump`, making `buildkite-agent env` just a container for this and any future subcommands relating to environment.

As suggested by @sj26 in October: https://github.com/buildkite/agent/pull/1803#issuecomment-1296327269

More recently the idea of a `buildkite-agent env set <key> <value>` command came up in [Re: 💡 Agent sockets - Pipelines Dispatch ⚙️](https://3.basecamp.com/3453178/buckets/27608512/messages/5721116595#__recording_5725223825)

It's relatively safe to assume that the only usage of the existing `buildkite-agent env` command is by the agent itself, making it essentially a private/internal API which we can safely change.

### DIFF
--- a/clicommand/env_dump.go
+++ b/clicommand/env_dump.go
@@ -9,8 +9,8 @@ import (
 	"github.com/urfave/cli"
 )
 
-const envDescription = `Usage:
-  buildkite-agent env [options]
+var EnvDumpHelpDescription = `Usage:
+  buildkite-agent env dump [options]
 
 Description:
    Prints out the environment of the current process as a JSON object, easily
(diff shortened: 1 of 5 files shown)


clicommand/env_dump.go
```diff
@@ -9,29 +9,31 @@ import (
 	"github.com/urfave/cli"
 )
 
-const envDescription = `Usage:
-  buildkite-agent env [options]
+var EnvDumpHelpDescription = `Usage:
```
DrJosh9000, line 12: Why change const to var, and why change the capitalisation (making it exported)?

main.go
```diff
@@ -82,6 +82,13 @@ func main() {
 				clicommand.ArtifactShasumCommand,
 			},
 		},
+		{
+			Name:  "env",
+			Usage: "Something something environment variables",
+			Subcommands: []cli.Command{
+				clicommand.EnvDumpCommand,
+			},
+		},
```
triarius, line 91:
```suggestion
		{
			Name:  "env",
			Usage: "Something something environment variables",
			Subcommands: []cli.Command{
				clicommand.EnvDumpCommand,
			},
			Action: clicommand.EnvDumpAction,
		},
```

I think something like this (with appropriate changes in `clicommand/env_dump.go`) should work to allow `buildkite-agent env` and `buildkite-agent env dump` to be synonymous.

Not only does it preserve backward compatibility (up to the flag changes you made), but I actually think this is desirable behaviour in the long term.

clicommand/env_dump.go
pda: Ah — I used the `clicommand/meta_data_*.go` as an example of a collection of subcommands, so followed the patterns therein rather than editing the old `env.go`. I guess the result was close enough that git decided it was an edit-diff rather than a delete-file / new-file.

main.go
pda: > but I actually think this is desirable behaviour in the long term

I wonder. I don't think dumping the process' current env is very useful outside the one weird use-case we have internally, which we hope to eliminate via control sockets. I can't imagine a user actually wanting to use `buildkite-agent env` to see their environment as JSON. I'd rather it show the usage and list of subcommands.

clicommand/env_dump.go
pda: All the other collections-of-subcommands follow this pattern to, so I think I'll stick with it, and we can change them all at once if we prefer a different style.

DrJosh9000: SGTM

DrJosh9000 (approved):
If we're making what is _technically_ a breaking change, you'd appreciate if it went out as soon as possible (in the next release)?

pda:
Yeah I guess the sooner the better, but not urgent if we feel it needs more discussion.

pda:
> `agent.Expect("env") -> agent.Expect("env", "dump")`

Thanks — I was just looking into that, and now I don't have to finish doing that :)

DrJosh9000:
It feels strange to be able to do that 😅

PR #1923, opened 2023-01-19
only use the ctx w/ timeout where necessary

leathekd:
We noticed today that many of our builds were failing with an `agent lost` comment.  

Screen Shot 2023-01-19 at 10 31 48 AM

On inspection we found that they were failing after 4 minutes and 30 seconds... which is suspiciously similar to the timeout introduced in https://github.com/buildkite/agent/pull/1894 .  Sure enough, our failing builds were all running on the v3.43.0 build.

I think the issue is that the `ctx` with the timeout that is created for the `acquireJob` loop is ultimately passed to the call to `RunJob`.  

This PR renames the context with the timeout and uses it where it seems appropriate while passing the original `ctx` on to `RunJob`.

Please let me know if I've got it wrong.  Happy to iterate on this if I've misread!  Thanks.

### DIFF
--- a/agent/agent_worker.go
+++ b/agent/agent_worker.go
@@ -532,7 +532,7 @@ func (a *AgentWorker) AcquireAndRunJob(ctx context.Context, jobId string) error 
 	// always hit the max of 1s, then another 8s is added to that. This is still comfortably within
 	// the timeout of 270s, and the bound seems tight enough so that the agent is not wasting time
 	// waiting for a retry that will never happen.
-	ctx, cancel := context.WithTimeout(ctx, 270*time.Second)
+	timeoutCtx, cancel := context.WithTimeout(ctx, 270*time.Second)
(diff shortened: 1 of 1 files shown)


moskyb:
hey @leathekd! this is a great spot. lemme run CI and do some local tests, but once it's verified i'll merge this and make a point release. thanks so much for your contribution!

moskyb (approved):
awesome work @leathekd! thanks so much ^_^

leathekd:
Thanks for the quick turnaround @moskyb!

moskyb:
approximately as quick as your turnaround fixing our bugs 😅

i'll release v3.43.1 in a hot minute, i'm just waiting for the aussies to wake up to review my PR.

triarius:
:facepalm: thanks for jump on this so quickly everyone!

Issue #1928, opened 2023-01-24
Buildkite Agent 3.43.x changed tini version & path


willhughes-au:
Buildkite Agent v3.43.x has changed the path and version of tini. 

Buildkite Agent 3.42.0 has tini 0.19.0 available at /sbin/tini:

```
$ docker run -it --entrypoint=/bin/sh buildkite/agent:3.42.0-ubuntu
# /sbin/tini
tini (tini version 0.19.0 - git.de40ad0)
Usage: tini [OPTIONS] PROGRAM -- [ARGS] | --version
```

In 3.43.1 that path is no longer there, and an older version of tini is at another path.

```
$ docker run -it --entrypoint=/bin/sh buildkite/agent:3.43.1-ubuntu
# /sbin/tini
/bin/sh: 1: /sbin/tini: not found
# which tini
/usr/bin/tini
# tini 
tini (tini version 0.18.0)
Usage: tini [OPTIONS] PROGRAM -- [ARGS] | --version
```

triarius:
Hi @willhughes-au. Thanks for reporting this to us. It does look like this was caused by a recent changed that we made to the agent's ubuntu-20.04 docker image to install `tini` from the ubuntu repos rather than downloading the binary from github. We consider installing packages from a distribution's repositories to be the better practice, so we are not keen to revert this change.

[We can create a symlink for you to restore the old path](https://github.com/buildkite/agent/pull/1934) and release a new version. In the meantime, it should be quite straightforward to work around this by using the new path, or even just `tini` to execute it instead. For example:
```
docker run -it --entrypoint tini buildkite/agent:3.43.1-ubuntu -- buildkite-agent --version
```
should give you the output:
```
buildkite-agent version 3.43.1, build x
```

Having said that, if you use the default entrypoint - located at /usr/local/bin/buildkite-agent-entrypoint`, it should already be invoking `tini` at the correct path. So I am a little curious if:
1. There is a particular reason why you need the path to be `/sbin/tini` and not `/usr/bin/tini`?
2. There is a particular reason why the default entrypoint, is not working for you?


[...]

PR #1934, opened 2023-01-31
Restore old tini path on ubuntu 20.04

triarius:
Tini used to be downloaded from github to /usr/sbin/tini. Now we install it
from the ubuntu repositories which installs to /usr/bin/tini.

Both are in the $PATH, so there should be no issue if a user invokes the
command as `tini`. Also, the default entrypoint is using the new path to
tini, so users starting the docker container without overriding the default
entrypoint should not have experienced any issues.

Fixes: #1928


DrJosh9000 (approved):
LGTM! 🤘

PR #1937, opened 2023-02-01
Give better errors when config loading fails

### LINKED ISSUE #1936: Improve error log when configuration file is poorly formatted
In `/etc/buildkite-agent/buildkite-agent.cfg` I accidentally set the token to:
```
token="{
    "ARN": "<arn>",
    "Name": "<name>",
    "VersionId": "<version>",
    "SecretString": "<secret>",
    "VersionStages": [
        "AWSCURRENT"
    ],
    "CreatedDate": 1234567890.123
}"
```

That's an obvious error on my end for not parsing the secret out and setting it.

The report is that the buildkite-agent failed with this log:
```
Jan 31 05:04:28 systemd[1]: Started Buildkite Agent.
Jan 31 05:04:28 buildkite-agent[7046]: Can't separate key from value
Jan 31 05:04:28 systemd[1]: buildkite-agent.service: Main process exited, code=exited, status=1/FAILURE
```

I believe it's trying to [load](https://github.com/buildkite/agent/blob/2dda4e362f4fabd986e21f54f9dba91306d94e13/cliconfig/file.go#L20-L59) the `buildkite-agent.cfg` and failing to parseLine: [code](https://github.com/buildkite/agent/blob/2dda4e362f4fabd986e21f54f9dba91306d94e13/cliconfig/file.go#L123).

Unfortunately the [Can't separate key from value](https://www.google.com/search?q=can%27t+separate+key+from+value&oq=Can%27t+separate+key+from+value) error message is a common docker failure which is misleading.

I'd appreciate that the failure included 'Can't separate key from value while parsing buildkite-agent.cfg' or so.

N/A

moskyb:
Closes #1936 

**This PR:** Generally gives the error handling for config loading a lick of paint, using error wrapping appropriately, and giving much more context when config loading fails.

As reported in #1936, previously, a malformed config file would cause the agent to simply print "Can't separate key from value" (to stdout).

Now, the same malformed config file prints out
```
Error loading config: loading config file: parsing config line: can't separate key from value in string "        \"AWSCURRENT\"", no valid separators (= or :) found
```
to stderr.

This should give users more info to be able to fix their configuration issues.

Additionally in this PR, i've modified a couple of the existing error messages to be a bit clearer, and made them all lowercase, so that go vet stops giving me lil yellow squiggly lines:
CleanShot 2023-02-02 at 11 02 51@2x

### DIFF
--- a/cliconfig/file.go
+++ b/cliconfig/file.go
@@ -3,6 +3,7 @@ package cliconfig
 import (
 	"bufio"
 	"errors"
+	"fmt"
 	"os"
 	"strings"
 
@@ -24,13 +25,13 @@ func (f *File) Load() error {
 	// Figure out the absolute path
 	absolutePath, err := f.AbsolutePath()
 	if err != nil {
-		return err
+		return fmt.Errorf("getting absolute path for %s: %w", f.Path, err)
 	}
 
 	// Open the file
 	file, err := os.Open(absolutePath)
 	if err != nil {
-		return err
+		return fmt.Errorf("opening file %s: %w", f.Path, err)
 	}
 
 	// Make sure the config file is closed when this function finishes
@@ -44,11 +45,11 @@ func (f *File) Load() error {
 	}
 
 	// Parse each line
-	for _, fullLine := range lines {
+	for lineNum, fullLine := range lines {
 		if !isIgnoredLine(fullLine) {
 			key, value, err := parseLine(fullLine)
 			if err != nil {
-				return err
+				return fmt.Errorf("parsing config line %d: %w", lineNum+1, err)
 			}
 
 			f.Config[key] = value
@@ -82,10 +83,9 @@ func (f File) Exists() bool {
 //
 // The project is released under an MIT License, which can be seen here:
 // https://github.com/joho/godotenv/blob/master/LICENCE
-func parseLine(line string) (key string, value string, err error) {
+func parseLine(line string) (key, value string, err error) {
 	if len(line) == 0 {
-		err = errors.New("zero length string")
-		return
+		return "", "", errors.New("zero length string")
 	}
 
 	// ditch the comments (but keep quoted hashes)
@@ -120,21 +120,18 @@ func parseLine(line string) (key string, value string, err error) {
 	}
 
 	if len(splitString) != 2 {
-		err = errors.New("Can't separate key from value")
-		return
+		return "", "", fmt.Errorf("can't separate key from value in string %q, no valid separators (= or :) found", line)
 	}
 
 	// Parse the key
 	key = splitString[0]
-	if strings.HasPrefix(key, "export") {
-		key = strings.TrimPrefix(key, "export")
-	}
-	key = strings.Trim(key, " ")
+	key = strings.TrimPrefix(key, "export")
+	key = strings.TrimSpace(key)
 
 	// Parse the value
 	value = splitString[1]
 	// trim
-	value = strings.Trim(value, " ")
+	value = strings.TrimSpace(value)
 
 	// check if we've got quoted values
 	if strings.Count(value, "\"") == 2 || strings.Count(value, "'") == 2 {
@@ -147,7 +144,7 @@ func parseLine(line string) (key string, value string, err error) {
 		value = strings.Replace(value, "\\n", "\n", -1)
 	}
 
-	return
+	return key, value, nil
 }
 
 func isIgnoredLine(line string) bool {
--- a/cliconfig/loader.go
+++ b/cliconfig/loader.go
@@ -50,7 +50,7 @@ func (l *Loader) Load() (warnings []string, err error) {
 			l.File = &file
 		} else {
 			absolutePath, _ := file.AbsolutePath()
-			return warnings, fmt.Errorf("A configuration file could not be found at: %q", absolutePath)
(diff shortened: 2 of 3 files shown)


cliconfig/file.go
```diff
@@ -82,10 +83,9 @@ func (f File) Exists() bool {
 //
 // The project is released under an MIT License, which can be seen here:
 // https://github.com/joho/godotenv/blob/master/LICENCE
-func parseLine(line string) (key string, value string, err error) {
```
moskyb, line 85: i've mostly changed this because i loathe named returns in go, but i'm happy to change things back if we want

cliconfig/file.go
```diff
@@ -120,19 +120,16 @@ func parseLine(line string) (key string, value string, err error) {
 	}
 
 	if len(splitString) != 2 {
-		err = errors.New("Can't separate key from value")
-		return
+		return "", "", fmt.Errorf("can't separate key from value in string %q, no valid separators (= or :) found", line)
```
moskyb, line 123: i love my two lil guys 
CleanShot 2023-02-02 at 11 04 22@2x

just vibing. hanging out

cliconfig/file.go
DrJosh9000: In my experience the main problem is bare returns. All your changes to the return statements are good!

Giving the return args names to aid documentation is usually a plus when there are multiple of the same type. 

But also `(key, value string, err error)` would have been slightly neater to begin with.

cliconfig/file.go
DrJosh9000: Love it!

cliconfig/file.go
```diff
@@ -48,7 +49,7 @@ func (f *File) Load() error {
 		if !isIgnoredLine(fullLine) {
 			key, value, err := parseLine(fullLine)
 			if err != nil {
-				return err
+				return fmt.Errorf("parsing config line: %w", err)
```
DrJosh9000, line 52: Should it include the line number?

cliconfig/file.go
```diff
@@ -120,19 +120,16 @@ func parseLine(line string) (key string, value string, err error) {
...
 	}
 
 	// Parse the key
-	key = splitString[0]
-	if strings.HasPrefix(key, "export") {
-		key = strings.TrimPrefix(key, "export")
-	}
+	key := splitString[0]
+	key = strings.TrimPrefix(key, "export")
 	key = strings.Trim(key, " ")
```
DrJosh9000, line 129: [Optional] Should we trim other kinds of spaces? `strings.TrimSpace`?

cliconfig/loader.go
```diff
@@ -225,7 +225,7 @@ func (l Loader) setFieldValueFromCLI(fieldName string, cliName string) error {
 				} else if fieldKind == reflect.Int {
```
DrJosh9000, line 225: [Optional] Time to switch these if..else if..else if.. chains out for a neater looking switch?

DrJosh9000 (commented):
This is excellent! A couple of comments for you, some are just complaints about the existing code though

cliconfig/file.go
moskyb: ooh that's a good idea! yes, will implement it

cliconfig/loader.go
moskyb: yep, good call. i had the same thought going through the code initially

cliconfig/file.go
moskyb: oh yeah, bare returns are 90% of the problem. I might modify to get the best of both worlds

cliconfig/file.go
moskyb: yeah, i think that's probably a good call. some people might be using (shudder) tabs for indentation.

cliconfig/file.go
```diff
@@ -120,16 +120,13 @@ func parseLine(line string) (key string, value string, err error) {
...
 	}
 
 	// Parse the key
 	key = splitString[0]
-	if strings.HasPrefix(key, "export") {
-		key = strings.TrimPrefix(key, "export")
-	}
-	key = strings.Trim(key, " ")
+	key = strings.TrimPrefix(key, "export")
+	key = strings.TrimSpace(key)
```
DrJosh9000, line 129: Maybe we should do the same for `value` below

cliconfig/loader.go
```diff
@@ -119,14 +119,14 @@ func (l *Loader) Load() (warnings []string, err error) {
 				// Error if they specify the deprecated version and the new version
 				if !l.fieldValueIsEmpty(renamedToFieldName) {
 					renamedFieldValue, _ := reflections.GetField(l.Config, renamedToFieldName)
-					return warnings, fmt.Errorf("Can't set config option `%s=%v` because `%s=%v` has already been set", cliName, value, renamedFieldCliName, renamedFieldValue)
+					return warnings, fmt.Errorf("couldn't set config option `%s=%v, `%s=%v` has already been set", cliName, value, renamedFieldCliName, renamedFieldValue)
```
DrJosh9000, line 122: The closing ` went missing?

DrJosh9000 (approved):
LGTM! 🥳

With two minor nits

cliconfig/loader.go
moskyb: oop, good spot 🦅👁️

PR #1939, opened 2023-02-06
clicommand: Descriptions unexported and const

DrJosh9000:
The CLI command descriptions aren't ever changed, nor are they used outside the clicommand package.

For consistency, the templates in main.go are updated the same way.


triarius (approved):
LGTM

PR #1940, opened 2023-02-06
ansi-timestamps: Compute prefixes at end-of-line

DrJosh9000:
This PR changes how the `Prefixer` works, so that the callback to produce a prefix is called for _the line that just finished_, rather than for _the next line_. Since we don't know when the current line is finished, input is buffered by the Prefixer, which must also be flushed after use (in case there is no terminating newline).

This makes the timestamps produced with the `ansi-timestamps` experiment consistent with the other options.


process/prefixer.go
```diff
@@ -1,83 +1,73 @@
...
-// This regex matches "Erase in Line" escape characters:
+// This regex matches either newline, or one of the "Erase in Line" escape
+// sequences:
 // [K Clear from cursor to the end of the line
 // [1K Clear from cursor to beginning of the line
 // [2K Clear entire line
-var lineBreakEscapeCharRegex = regexp.MustCompile(`^\[[012]?K`)
+var lineBreakEscapeRE = regexp.MustCompile(`\n|\x1b\[[012]?K`)
 
+// Prefixer inserts prefixes generated by a callback before each line.
```
triarius, line 15: Maybe make a mention here that you should always Flush()

triarius (approved):
Amazing work! LGTM

process/prefixer.go
DrJosh9000: Done

PR #1943, opened 2023-02-07
Add current-job api

moskyb:
**This PR:** Exposes an API through the agent's job executor (aka "the bootstrap") that allows running jobs to mutate or introspect their own state. 

This API currently exposes three endpoints:
- `GET /api/current-job/v0/env` to get the environment of the currently running job
- `PATCH /api/current-job/v0/env` to update the job's environment
- `DELETE /api/current-job/v0/env` to remove certain keys from the job's environment

On unix machines, this API is exposed via a unix domain socket, and on windows, TBD.

**Still to do:**
- [ ] ~~Figure out how to serve the `jobapi.Server` on windows (probably a normal TCP port or something? UDSes are available on recent builds of windows, but we don't want to have to replay on that)~~ we're going to do this in a separate PR
- [x] Write tests for `jobapi.Server`
- [x] Actually run the `jobapi.Server` from the job executor + do some manual testing that it looks all good
- [ ] ~~Create a `jobapi.Client` thingy to allow other bits of the agent (and potentially other golang projects 👀) to use this API~~ We're also going to do this in a separate PR (and potentially later on)


jobapi/middleware.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"encoding/json"
+	"net/http"
+	"strings"
+)
+
+func (s *Server) authMdlw(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		defer next.ServeHTTP(w, r)
```
moskyb, line 11: 🤔 should we not call the next middleware in the stack if we're sending an unauthorised response?

jobapi/doc.go
```diff
@@ -0,0 +1,3 @@
+// jobapi provides an API with which to interact with and mutate the currently executing job
+// Pronunciation: /ˈdʒɑbe̞pi/ /joh-bah-pee/
```
moskyb, line 2: [nit]

the IPA here is wrong: 
```suggestion
// Pronunciation: /ˈdʒɑbapi/ /joh-bah-pee/
```

jobapi/middleware_test.go
```diff
@@ -0,0 +1,100 @@
...
+			auth:     "",
+			wantCode: http.StatusUnauthorized,
+			wantBody: map[string]string{"error": "authorization header is required"},
+		},
+	}
+
+	for _, c := range cases {
+		c := c
+		t.Run(c.title, func(t *testing.T) {
+			t.Parallel()
```
moskyb, line 58: do the child test cases need to be `t.Parallel`ed if the [parent test is already parallel?](https://github.com/buildkite/agent/pull/1943/files#diff-6d895c7a60f16288c48e247e795f18cb4634b9b1def94d2bec65fc8c67885e1bR20)

jobapi/tokens.go
```diff
@@ -0,0 +1,23 @@
...
+	b := make([]byte, length)
+	for i := range b {
+		bigRand, err := rand.Int(rand.Reader, bigLettersLen)
+		if err != nil {
+			return "", fmt.Errorf("generating crypto-random big.Int: %w", err)
+		}
+		b[i] = letters[bigRand.Int64()%int64(lettersLen)] // mod letterslen just in case rand.Int returns a number larger than lettersLen, don't want to have a panic
+	}
+	return string(b), nil
+}
```
moskyb, line 23: This potentially feels like i'm getting a little bit too fancy with all the `big.Int`s and the cryptorandoms and such. happy to make this a slightly more normal random function if it makes sense.

jobapi/server.go
```diff
@@ -0,0 +1,108 @@
...
+	}()
+
+	// Trigger graceful shutdown
+	err := s.httpSvr.Shutdown(shutdownCtx)
+	if err != nil {
+		return fmt.Errorf("shutting down Job API server: %w", err)
+	}
+
+	return nil
+}
```
moskyb, line 95: I need to go back and double check this code - it's ~stolen from~ inspired by the Chi docs, but i haven't actually checked that it works the way i expect it to yet.

jobapi/server.go
```diff
@@ -0,0 +1,108 @@
...
+	"net/http"
+	"os"
+	"time"
+
+	"github.com/buildkite/agent/v3/env"
+)
+
+type Server struct {
+	SocketPath string
+	environ    *env.Environment
```
moskyb, line 17: [All maps in go are reference types](https://go.dev/blog/maps), and are implicitly pointers anyway, butttttt i do kinda like the `*` in the struct (and the constructor below) to call out that "hello, this is a mutable thing and we're going to be mutating it", which might be slightly less clear without it?

on the other hand, it might just be confusing and say "man, why did that benno person do that" to future travellers.

moskyb: Suggestion: rewrite the whole thing in rust for those sweet sweet mutability markers
```Rust
struct Server {
  socket_path: String,
  environ: &mut HashMap<String, String>,
  token: String,
  started: bool,
}
```

jobapi/routes.go
```diff
@@ -0,0 +1,134 @@
...
+	protected := make([]string, 0, len(candidates))
+	for _, c := range candidates {
+		if _, ok := protectedEnv[c]; ok {
+			protected = append(protected, c)
+		}
+	}
+	return protected
+}
+
+func returnError(w http.ResponseWriter, err error, code int) {
```
moskyb, line 131: this should potentially be renamed to `writeError` or something, it doesn't return anything and having 
```Go
returnError(w, errors.New("blah"), 500)
return
```
repeats itself and looks a little funny

jobapi/routes.go
```diff
@@ -0,0 +1,134 @@
...
+		updated := make([]string, len(req.Env))
+		protected := checkProtected(maps.Keys(req.Env))
+
+		if len(protected) > 0 {
+			err = fmt.Errorf("the following environment variables are protected, and cannot be modified: % v", protected)
+			returnError(w, err, http.StatusUnprocessableEntity)
+			return
+		}
+
+		for k, v := range req.Env {
```
moskyb, line 71: At current, when the json `{"foo": null}` is unmarshalled into a `map[string]string`, it comes out as `map[string]string{"foo": ""}`, which is a different (and valid) thing to do with the patch API, whereas we don't want to allow people to set env to nil using the patch endpoint - they should be doing that via the delete endpoint.

instead, we should update the Env field of the `EnvUpdateRequest` struct to be a `map[string]*string`, and in this function, check that none of the values are nil. if any are, we should fail the entire operation prior to updating any env.

jobapi/middleware.go
```diff
@@ -0,0 +1,49 @@
...
+
+		enc := json.NewEncoder(w)
+
+		auth := r.Header.Get("Authorization")
+		if auth == "" {
+			w.WriteHeader(http.StatusUnauthorized)
+			enc.Encode(ErrorResponse{Error: "authorization header is required"})
```
moskyb: should probably update these to use `returnError`

jobapi/tokens.go
DrJosh9000: What about something like

```go
func generateToken(minBits int) (string, error) {
    b := make([]byte, (minBits+7)/8)
    rand.Read(b)
    return base64.URLEncoding.EncodeToString(b), nil
}
```

jobapi/server.go
DrJosh9000: Yeah, IIRC there was something of a discussion pre-Go 1.0 about whether maps ought to carry the * to signify they're all references. Since that discussion is ancient history at this point, and it's a similar situation if it were an interface type, and we're not rewriting in Rust (yet), I recommend avoiding the *.

jobapi/middleware.go
```diff
@@ -0,0 +1,47 @@
+package jobapi
+
+import (
+	"errors"
+	"net/http"
+	"strings"
+)
+
+func AuthMdlw(token string) func(http.Handler) http.Handler {
```
DrJosh9000, line 9: I don't want to nitpick about naming and doc comments, but also I very strongly *want* to nitpick about naming and doc comments...

jobapi/routes.go
```diff
@@ -0,0 +1,148 @@
...
+	"github.com/go-chi/chi/v5"
+	"github.com/go-chi/chi/v5/middleware"
+	"golang.org/x/exp/maps"
+)
+
+var protectedEnv map[string]struct{}
+
+func init() {
+	protectedEnv = make(map[string]struct{}, len(agent.ProtectedEnv))
+	for _, p := range agent.ProtectedEnv {
```
DrJosh9000, line 18: Why isn't ProtectedEnv a `map[string]struct{}` already?

jobapi/routes.go
```diff
@@ -0,0 +1,148 @@
...
+	r.Route("/api/current-job/v0", func(r chi.Router) {
+		r.Get("/env", s.getEnv())
+		r.Patch("/env", s.patchEnv())
+		r.Delete("/env", s.deleteEnv())
+	})
+
+	return r
+}
+
+func (s *Server) getEnv() http.HandlerFunc {
```
DrJosh9000, line 41: What about:

```go
r.Route(..., func(r chi.Router) {
    r.Get("/env", s.getEnv)   // no parentheses on s.getEnv
    ...
})

func (s *Server) getEnv(w http.ResponseWriter, _ *http.Request) {
     // no func generator needed
}

**Review by DrJosh9000 — COMMENTED** [review]
Some comments for you!

File: jobapi/middleware.go (hunk 10 above)
Comment (moskyb) [reply jobapi/middleware.go:9]: doit

File: jobapi/routes.go (hunk 12 above)
Comment (moskyb) [reply jobapi/routes.go:41]: oh yeah, good call. i think i used the func generator pattern for the middlewares because some of them actually need stuff passed into them and then blindly used them for the route functions too. 

will update

File: jobapi/routes.go (hunk 11 above)
Comment (moskyb) [reply jobapi/routes.go:18]: 🤔 excellent question. i hate using `init()` so i might just change the `agent.ProtectedEnv`

File: jobapi/middleware_test.go (hunk 3 above)
Comment (DrJosh9000) [reply jobapi/middleware_test.go:58]: If you want them to be parallel with each other, then I believe so: https://pkg.go.dev/testing#T.Run

File: jobapi/routes.go [hunk 13]
```diff
@@ -0,0 +1,133 @@
...
+	protected := make([]string, 0, len(candidates))
+	for _, c := range candidates {
+		if _, ok := agent.ProtectedEnv[c]; ok {
+			protected = append(protected, c)
+		}
+	}
+	return protected
+}
+
+func writeError(w http.ResponseWriter, err error, code int) {
```
DrJosh9000, line 130: I noticed that 4 out of 9 times, `writeError` is called with a string wrapped in an error - I'm guessing solely to satisfy the type of `err` here? Since this is more or less a convenience function, I reckon it makes sense to be a bit looser with the arg type. Use `any` and stringify it with `fmt.Sprint`?

jobapi/middleware.go
DrJosh9000: Alright. 😈

Could you 💐please💐 add (throughout) doc comments to the top-level exported identifiers. This doesn't need to happen right away, but ideally before the PR is ready to go.

```go
// AuthMdlw returns a middleware that checks for the given bearer token
// before invoking the next handler.
```

I'm also wondering why `Mdlw` is abbreviated... I suppose I can understand if you say vowels are expensive?

moskyb: > I'm also wondering why Mdlw is abbreviated... I suppose I can understand if you say vowels are expensive?
you know, i don't know. i thiiiiiiiiiink that it comes from some http library that i used yonks ago that called its middleware package `mdlw`, but it's a bad reason to abbreviate 😅

bootstrap/bootstrap.go
```diff
@@ -85,6 +88,42 @@ func (b *Bootstrap) Run(ctx context.Context) (exitCode int) {
...
+		srv, token, err := jobapi.NewServer(jobAPISocketPath, b.shell.Env)
+		if err != nil {
+			b.shell.Errorf("Error creating job API server: %v", err)
+			return 1
+		}
+
+		b.shell.Env.Set("BUILDKITE_AGENT_JOB_API_SOCKET", jobAPISocketPath)
+		b.shell.Env.Set("BUILDKITE_AGENT_JOB_API_TOKEN", token)
+
+		err = srv.Start()
```
DrJosh9000, line 107:
```suggestion
		if err := srv.Start(); err != nil {
```

I prefer this kind of thing; for one reason, the scope of `err` is limited to the `if`, which it makes it clear there's no further processing of the value for `err`.

jobapi/server.go
```diff
@@ -0,0 +1,133 @@
...
+		}
+	}()
+
+	// Trigger graceful shutdown
+	err := s.httpSvr.Shutdown(shutdownCtx)
+	if err != nil {
+		return fmt.Errorf("shutting down Job API server: %w", err)
+	}
+
+	err = os.Remove(s.SocketPath)
```
DrJosh9000, line 104: I think the more correct cleanup action is to call `Close` on the listener (created above in `Start`). The default behaviour of `net.UnixListener`'s `Close` is to unlink the file if it was created by the `net` package (and can be changed with [SetUnlinkOnClose](https://pkg.go.dev/net#UnixListener.SetUnlinkOnClose)).

bootstrap/bootstrap.go
```diff
@@ -85,6 +88,42 @@ func (b *Bootstrap) Run(ctx context.Context) (exitCode int) {
 	defer stopper()
 	defer func() { span.FinishWithError(err) }()
 
+	if experiments.IsEnabled("job-api") {
+		if runtime.GOOS == "windows" {
+			b.shell.Errorf("The job API is not currently supported on Windows. It's on its way though, promise!")
+			return 1
+		}
+
+		jobAPISocketPath := path.Join(os.TempDir(), fmt.Sprintf("buildkite-agent-%d.sock", os.Getpid()))
```
DrJosh9000, line 97: I'm a bit worried by `os.TempDir` here: It defaults to `/tmp`, which is usually world-writable, so something else could (usually accidentally) start messing with our socket.

It turns out there's [a standardised place](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s15.html) for Unix sockets: `/run` (or a subdirectory). Using this creates a new problem for us, which is, there might not be an existing subdirectory writable by the agent.

jobapi/server_test.go
```diff
@@ -0,0 +1,359 @@
...
+	e.Set("CAPITAL", "quito")
+
+	return e
+}
+
+func testServer(e env.Environment) (*jobapi.Server, string, error) {
+	sockNum := random.Int() % 100_000
+	sockName := path.Join(os.TempDir(), fmt.Sprintf("jobapi-test-%d.sock", sockNum))
+	return jobapi.NewServer(sockName, e)
+
```
DrJosh9000, line 40: (nit) Bonus newline?

DrJosh9000 (commented):
Hey! I wrote some more comments for you...

bootstrap/bootstrap.go
```diff
@@ -85,6 +88,42 @@ func (b *Bootstrap) Run(ctx context.Context) (exitCode int) {
...
+		srv, token, err := jobapi.NewServer(jobAPISocketPath, b.shell.Env)
+		if err != nil {
+			b.shell.Errorf("Error creating job API server: %v", err)
+			return 1
+		}
+
+		b.shell.Env.Set("BUILDKITE_AGENT_JOB_API_SOCKET", jobAPISocketPath)
```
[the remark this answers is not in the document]
moskyb: good catch, it does not. switching this to `os.Setenv` makes it work though

jobapi/server.go
moskyb: you are correct! i also found that calling `http.Server.Shutdown()` automatically calls `Close` on all of the listeners for that server, so there's not even anything we need to do.

jobapi/middleware.go
DrJosh9000: Yeah, probably not

jobapi/routes.go
```diff
@@ -0,0 +1,147 @@
...
+
+	resp := EnvUpdateResponse{
+		Added:   added,
+		Updated: updated,
+	}
+
+	resp.Normalize()
+
+	json.NewEncoder(w).Encode(resp)
+	w.WriteHeader(http.StatusOK)
```
DrJosh9000, line 96: w.WriteHeader needs to come before writing to w ... if for no other reason than it logs a warning about a superflouous WriteHeader call.

bootstrap/api.go
```diff
@@ -0,0 +1,58 @@
...
+		b.shell.Warningf("We'll continue to run your job, but you won't be able to use the Job API")
+		return cleanup, nil
+	}
+
+	socketPath, err := jobapi.NewSocketPath()
+	if err != nil {
+		return cleanup, fmt.Errorf("creating job API socket path: %v", err)
+	}
+
+	srv, token, err := jobapi.NewServer(socketPath, b.shell.Env)
```
DrJosh9000, line 33: I think there's a problem here (similar to the missing `SOCKET` and`TOKEN` vars): `setUp` overwrites `b.shell.Env` with a new map, but because that's after the call to `startJobAPI`,`NewServer` gets (a reference to) the old map. Changes from `PATCH` and `DELETE` can be seen with `GET`, but won't be present as 100% Genuine environment variables in later hooks/plugins.

EXPERIMENTS.md
```diff
@@ -58,3 +58,20 @@ Changes the file lock implementation from github.com/nightlyone/lockfile to gith
 When the experiment is enabled the agent will use different lock files from agents where the experiment is disabled, so agents with this experiment enabled should not be run on the same host as agents where the experiment is disabled.
 
 **Status**: Being tested, but it's looking good. We plan to switch lock implementations over the course of a couple of releases, switching over in such a way that nothing gets broken.
+
+### `job-api`
+
+Exposes a local API for the currently running job to introspect and mutate its state in the form of environment variables. This allows you to write scripts, hooks and plugins in languages other than bash, using them to interact with the agent.
```
DrJosh9000, line 64: > This allows you to write scripts, hooks and plugins in languages other than bash ...

[🚨 Pedantry mode activated 🚨]

I don't think this is strictly 100% true. The API definitely makes it possible for almost any program running under a hook or plugin script to alter environment variables, but the job executor still assumes that hooks and plugins are _scripts_.

I don't know if it's worth changing the text here, though.

Maybe we should start thinking about the counterpart change (another experiment?) to make the job executor run hooks and plugins directly (without sourcing through the scriptwrapper)? (Or does this exist already?)

bootstrap/integration/bootstrap_tester.go
```diff
@@ -44,7 +44,7 @@ type BootstrapTester struct {
 }
 
 func NewBootstrapTester() (*BootstrapTester, error) {
-	homeDir, err := os.MkdirTemp("", "home")
+	homeDir, err := os.MkdirTemp("/tmp", "home")
```
DrJosh9000, line 47: Maybe worth a comment mentioning the Linux-path-length-for-Unix-sockets problem?

EXPERIMENTS.md
moskyb: > Maybe we should start thinking about the [...] change to make the job executor run hooks and plugins directly

I definitely agree that we should start thinking about it but.... i suspect we might not actually be able to do it. some commands are already commands that we wrap in a script (eg, `echo "hello"`), but i suspect we'll probably have to keep the scriptwrapper around for the forseeable future as a backwards-compatibility thing... we have no way of knowing whether something is a script that will use our API, or a script that's expecting to have its env captured, so i suspect we'll need to keep wrapping everything in the scriptwrapper regardless

DrJosh9000 (approved):
🎉 🎉 🎉 🥳 🎊 🪩 🍕

A bunch of remaining comments are minor nits, I leave it up to you whether to apply, raise a follow-up, or ignore 🥺

jobapi/middleware_test.go
```diff
@@ -0,0 +1,100 @@
...
+func TestAuthMdlw(t *testing.T) {
+	t.Parallel()
+
+	token := "llamas"
+	cases := []struct {
+		title    string
+		auth     string
+		wantCode int
+		wantBody map[string]string
+	}{
```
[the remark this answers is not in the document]
moskyb: ooh this is a very good call, will add!

jobapi/server.go
```diff
@@ -0,0 +1,152 @@
...
+	r := s.router()
+	l, err := net.Listen("unix", s.SocketPath)
+	if err != nil {
+		return fmt.Errorf("listening on socket: %w", err)
+	}
+
+	s.httpSvr = &http.Server{Handler: r}
+	go func() {
+		_ = s.httpSvr.Serve(l)
+	}()
```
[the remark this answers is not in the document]
moskyb: if a func returns an err that i'm not tracking, i much prefer to have `_ = mightFail()`, as it makes my intentions clear - it shows that i know that this function might return an error, but that i explicitly don't care. happy to change either way, but it's a readability thing that i like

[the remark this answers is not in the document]
DrJosh9000: That's fair, and tend to agree that `_ =` to denote deliberately ignored errors is nice. I'm less used to it. 

Though I think `http.Server.Serve` is a bit unusual in that it _always_ returns a non-nil error.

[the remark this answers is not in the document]
moskyb: oh that's true, i hadn't realised that... i think we're probably still fine to ignore it though, unless you disagree?

jobapi/socket.go
```diff
@@ -0,0 +1,40 @@
...
+
+	"github.com/mitchellh/go-homedir"
+)
+
+var random = rand.New(rand.NewSource(time.Now().UnixNano()))
+
+// NewSocketPath generates a path to a socket file (without actually creating the file itself) that can be used with the
+// job api.
+// These files are located in a subdirectory of the home dir (`~/.buildkite-agent/sockets/job-api` on unix machines),
+// and are named after the PID of the process that runs this function, with a random number appended to the end.
```
triarius, line 18: Given then issues we've had with the length of this path in testing, does it make sense to make the PATH to the socket completely configurable? A user with a very long username could find themselves having such issues and may need to find somewhere else to put it. We should give them a heads-up about the 108 character thing too.

jobapi/socket.go
```diff
@@ -0,0 +1,40 @@
...
+		return "", fmt.Errorf("finding home directory: %w", err)
+	}
+
+	path := filepath.Join(home, ".buildkite-agent", "sockets", "job-api")
+	err = os.MkdirAll(path, 0700)
+	if err != nil {
+		return "", fmt.Errorf("creating socket directory: %w", err)
+	}
+
+	sockNum := random.Int63() % 100_000
```
triarius, line 38: Should this be cryptographic random? Probably not, but might be good to write down why.

triarius (commented):
Nice work! I have some early comments I want to get in.

bootstrap/bootstrap.go
```diff
@@ -85,6 +88,42 @@ func (b *Bootstrap) Run(ctx context.Context) (exitCode int) {
...
+		b.shell.Env.Set("BUILDKITE_AGENT_JOB_API_TOKEN", token)
+
+		err = srv.Start()
+		if err != nil {
+			b.shell.Errorf("Error starting Job API server: %v", err)
+			return 1
+		}
+
+		b.shell.Commentf("Job API server started at %s", jobAPISocketPath)
+		b.shell.Commentf("Token: %s", token)
```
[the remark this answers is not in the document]
moskyb: It looks like between the time that you reviewed this and hit the send button, i got rid of these - they were mostly for debugging while i was working on it :)

jobapi/socket.go
moskyb: nah, this doesn't need to be crypto random - it's just to avoid collisions if one process starts more than one socket (which should never happen in theory, but it can be useful in tests)

jobapi/server_test.go
```diff
@@ -0,0 +1,380 @@
...
+	if err != nil {
+		t.Fatalf("srv.Start() = %v", err)
+	}
+
+	info, err := os.Stat(srv.SocketPath)
+	if err != nil {
+		t.Fatalf("os.Stat(%q) error = %v", srv.SocketPath, err)
+	}
+
+	isSocket := info.Mode()&os.ModeSocket != 0
```
DrJosh9000, line 72: I just discovered https://github.com/golang/go/issues/33357, so this check might need skipping on Windows

moskyb: 😭

DrJosh9000: I'm having trouble even calling `os.Stat` on the socket on Windows. But connecting to it is fine.

jobapi/server.go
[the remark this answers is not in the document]
DrJosh9000: Ignoring it is fine, with either syntax.

jobapi/server_test.go
DrJosh9000: Here's what I ended up with in the client: https://github.com/buildkite/agent/blob/3f9c89d4f519dc89e8ad0b98ee40d1d4bbcd1861/jobapi/client.go#L51-L69

PR #1945, opened 2023-02-08
OSS acknowledgements

DrJosh9000:
This adds a compilation of all the licenses and notices of dependencies used to build the agent (ACKNOWLEDGEMENTS.md), and a script to produce the compilation.

I've also made a compressed version, embedded it into clicommand, and added a new subcommand for printing it. This will add around 20KB to the binary.


scripts/generate-acknowledgements.sh
```diff
@@ -0,0 +1,57 @@
...
+fi
+
+OUTPUT_FILE="ACKNOWLEDGEMENTS.md"
+
+# Ensure modules are downloaded
+go mod download
+
+# Get go-licenses tool
+go install github.com/google/go-licenses@latest
+GO_LICENSES="${HOME}/go/bin/go-licenses"
```
triarius, line 16: Some people might have installed this by some other means. Some, not necessarily the same, people might also have a different GOPATH.

```suggestion
if ! command -v go-licenses >/dev/null; then
	go install github.com/google/go-licenses@latest
	GO_LICENSES="$(go env GOPATH)/bin/go-licenses"
else
	GO_LICENSES="$(command -v go-licences)"
fi
```

triarius (commented):
This is great! I think it is a good idea to have a CI step that generates this and checks there's no diff from what's output by the built binary.

scripts/generate-acknowledgements.sh
triarius: This is all the testing I've done:
```
if ! command -v go-licenses >/dev/null; then    go install github.com/google/go-licenses@latest;   GO_LICENSES="$(go env GOPATH)/bin/go-licenses"; else    GO_LICENSES="$(command -v go-licenses)"; fi
[narthana@quirinius ~]$ echo $GO_LICENSES
```

triarius (approved):
Thanks!

DrJosh9000:
> This is great! I think it is a good idea to have a CI step that generates this and checks there's no diff from what's output by the built binary.

I wonder how often Dependabot will be blocked by it. The probability is probably pretty low that a dependency will change licenses, but it could happen. Dependencies could also add or change notice files.

DrJosh9000:
Actually, that's probably a good thing. If a dependency changes licenses, right now we probably wouldn't notice!

I'll follow up with another PR to add the CI step, in order for it to have a file to diff against.

PR #1946, opened 2023-02-08
Remove upstart support

sj26:
The last version of Ubuntu to use Upstart was 14.10. New installations should be using systemd. Old installations will retain a copied-in upstart unit. The logic in here to juggle the init systems is no longer required.


DrJosh9000 (approved):
LGTM! 🎉

PR #1947, opened 2023-02-08
Restart with systemd

sj26:
Currently an upgrade via the debian package will send a TERM to all buildkite-agent processes and hope they restart. If they're run via systemd, a TERM will cause the agent to stop gracefully and exit with a 0 status. Our systemd unit is configured to [only restart on failures](https://github.com/buildkite/agent/blob/ef099a377634744decd66967fd0da4d9c1d2aa08/packaging/linux/root/usr/share/buildkite-agent/systemd/buildkite-agent.service#L13), so it won't restart automatically. (Otherwise the agent won't stop when asked to from the dashboard.)

Try to be smarter when restarting after an upgrade on debian-based systems with systemd. Most folks will be running the agent using systemd, so try to restart it nicely using systemctl. Fall back to the old behaviour.

Another option could be to introduce a dedicated "restart" signal, but that feels too heavyweight here.

I need to test this, but I need a debian package built to do so. Using CI to build a package felt easiest.

### DIFF
--- a/packaging/linux/scripts/after-install-and-upgrade.sh
+++ b/packaging/linux/scripts/after-install-and-upgrade.sh
@@ -95,11 +95,19 @@ TXT
   echo "and then you can start your agent by running \"$START_COMMAND\""
 fi
 
-# Crude method of causing all the agents to restart
+# Restart on upgrade, if we can
 if [ "$OPERATION" = "upgrade" ] ; then
-  # Restart agents that have a command line containing
-  # "buildkite-agent v1.2.3.4" or "buildkite-agent start"
-  pkill -f 'buildkite-agent (v|start)' > /dev/null 2>&1 || true
+  if [ $BK_SYSTEMD_EXISTS -eq 0 ] && systemctl is-active buildkite-agent --quiet; then
+    # Try using systemd, if the unit is active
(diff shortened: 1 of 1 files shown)


packaging/linux/scripts/after-install-and-upgrade.sh
```diff
@@ -95,24 +95,18 @@ TXT
...
 
-# Crude method of causing all the agents to restart
+# Restart on upgrade, if we can
 if [ "$OPERATION" = "upgrade" ] ; then
-  # Restart agents that have a command line containing
-  # "buildkite-agent v1.2.3.4" or "buildkite-agent start"
-  pkill -f 'buildkite-agent (v|start)' > /dev/null 2>&1 || true
-fi
-
-# Make sure all the folders created are owned by the buildkite-agent user #
```
DrJosh9000, line 105: Is this part still relevant?

sj26: Yeah, bum merge on my part 🙈

DrJosh9000 (approved):
LGTM 🥳

sj26:
I upgraded using the debian package from the build and it restarted the process using systemd nicely.

PR #1948, opened 2023-02-12
fix: dedup arguments to git when fetching submodule

yesudeep:
For a git repository that uses submodules, the agent currently appends the same arguments per submodule to the git command line, which is less than ideal and results in the following error (for 2 submodules):

```
$ git submodule update --init --recursive --force submodule update --init --recursive --force
error: pathspec 'submodule' did not match any file(s) known to git
error: pathspec 'update' did not match any file(s) known to git
error: pathspec '--init' did not match any file(s) known to git
error: pathspec '--recursive' did not match any file(s) known to git
error: pathspec '--force' did not match any file(s) known to git
```

The goal of this commit is to ensure the arguments are reinitialized per submodule in order to prevent this situation.

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -1410,6 +1410,8 @@ func (b *Bootstrap) defaultCheckoutPhase(ctx context.Context) error {
 			b.shell.Warningf("Failed to enumerate git submodules: %v", err)
 		} else {
 			for idx, repository := range submoduleRepos {
+				// Ensure we don't append duplicate arguments.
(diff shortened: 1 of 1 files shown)


triarius:
Thanks for picking up on and fixing this @yesudeep. We reverted the PR that introduced this: https://github.com/buildkite/agent/pull/1949.

We're going to let the author of that PR resubmit it with this error in mind. Unfortunatly, that means this PR is superseeded. Please let us know if you're having further issues though.

PR #1959, opened 2023-02-17
use reference clones for submodules redux

jonahbull:
Add support for using reference repositories for submodules in addition to regular repositories.

This is an improved version of #1935 which was reverted in #1949.

It fixes the following issues discovered after #1935 was merged:

* Each submodule repo would append the arguments `submodule update --init --recursive --force` to the git command.

* The reference repos for submodules didn't seem to be be created on the first run. This is in contrast to what happens when using a reference repo for the top level.

* Submodule checkout behavior was affected even when the `git-mirrors` experiment wasn't enabled

A brief overview of other changes:

* `updateGitMirror` now takes the repository to update as a parameter.

* Since we need it in two places now, I pulled the code for getting and/or updating the `mirrorDir` into its own method, `getOrUpdateMirrorDir`.

The tests pass, and I did some testing on my own for various scenarios outside of our specific use case this time.

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -1149,9 +1149,9 @@ func hasGitCommit(ctx context.Context, sh *shell.Shell, gitDir string, commit st
 	return true
 }
 
-func (b *Bootstrap) updateGitMirror(ctx context.Context) (string, error) {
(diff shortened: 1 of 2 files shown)


DrJosh9000 (approved):
LGTM ✅

Thanks for persisting!

PR #1960, opened 2023-02-19
fix 353 - fix this Error when using ssh : 'launchctl on macos fails with "Could not find domain for"'

tcptps:
fix 353 see https://github.com/buildkite/docs/issues/353#issuecomment-1435758739
edit: 
The Error occurs because by default (without LimitLoadToSessionType) launchd tries to load the agent into the Aqua context but this context doesnt exist when you are only logged in using ssh. 
The -S Background option tells launchd to load the agent into the Background Context. See changes in this PR for matching changes to the docs: https://github.com/buildkite/docs/pull/1875#issue-1590513215

### DIFF
--- a/templates/launchd.plist
+++ b/templates/launchd.plist
@@ -54,5 +54,12 @@
       <key>BUILDKITE_AGENT_CONFIG</key>
       <string>/Users/your-build-user/.buildkite-agent/buildkite-agent.cfg</string>
     </dict>
+    <key>LimitLoadToSessionType</key>
+    <array>
+      <string>Aqua</string>
+      <string>LoginWindow</string>
+      <string>Background</string>
+      <string>StandardIO</string>
+    </array>
   </dict>
 </plist>


templates/launchd.plist
```diff
@@ -54,5 +54,12 @@
       <key>BUILDKITE_AGENT_CONFIG</key>
       <string>/Users/your-build-user/.buildkite-agent/buildkite-agent.cfg</string>
     </dict>
+    <key>LimitLoadToSessionType</key>
+    <array>
+      <string>Aqua</string>
+      <string>LoginWindow</string>
+      <string>Background</string>
+      <string>StandardIO</string>
+    </array>
```
DrJosh9000, line 63: I often see `<string>System</string>` as well (for instance Homebrew creates it). I'm not sure it's necessary though.

templates/launchd.plist
```diff
@@ -54,5 +54,12 @@
       <key>BUILDKITE_AGENT_CONFIG</key>
       <string>/Users/your-build-user/.buildkite-agent/buildkite-agent.cfg</string>
     </dict>
+    <key>LimitLoadToSessionType</key>
```
DrJosh9000, line 57: For consistency can there be a blank line between the previous item and this new one?

DrJosh9000 (commented):
Thanks for the PR, this looks pretty good! 

One general comment: can the PR title and description be more descriptive? (353 is a bug in another repo, and it's helpful when browsing old PRs to get a sense of "what" and "why" without having to dig into the changes.)

DrJosh9000 (approved):
Thanks again for the PR! I'm happy for this to merge as-is, and I'll follow up with any tweaks.

PR #1963, opened 2023-02-21
Check that a pipeline upload error actually comes from the JSON marshaller and exit early

triarius:
In all cases where the pipeline cannot be marshaled to JSON, there is no point to retry the upload.
Retries will keep the agent busy for about 5 min, which is a waste of customers' compute time.

### DIFF
--- a/clicommand/pipeline_upload.go
+++ b/clicommand/pipeline_upload.go
@@ -3,6 +3,7 @@ package clicommand
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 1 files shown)


clicommand/pipeline_upload.go
```diff
@@ -315,6 +316,13 @@ var PipelineUploadCommand = cli.Command{
...
 
+				if jsonerr := new(json.MarshalerError); errors.As(err, &jsonerr) {
+					l.Error("Unrecoverable error, skipping retries")
+					r.Break()
+
+					return jsonerr
+				}
+
 				// 422 responses will always fail no need to retry
 				if apierr, ok := err.(*api.ErrorResponse); ok && apierr.Response.StatusCode == 422 {
```
triarius, line 327: While this part should also use the `errors.As` pattern as above, there is another PR out there where I have made this change: https://github.com/buildkite/agent/pull/1927

When this is merged, that should have merge conflicts. But I will take care of it.

DrJosh9000 (approved):
LGTM ✅

PR #1964, opened 2023-02-21
agent/s3.go: Fix nil pointer dereference

DrJosh9000:
At this point in `NewS3Client`, `sess` is unset, leading to a fun and useful panic if no region is discoverable (e.g. you aren't using AWS.)


triarius (approved):
Nice work!

PR #1973, opened 2023-02-24
Fix extra pipeline uploaded in fallback upload flow

triarius:
I made a snafu in https://github.com/buildkite/agent/pull/1927 that meant that if the API only supported the old upload flow, the pipeline would be uploaded once more than necessary.

This is evidenced by the failure of the test in this build: https://buildkite.com/buildkite/agent/builds/5316.

This is not impacting the correctness of uploads, as they are deduped by a UUID would have been the same between all the uploads. However, it is not necessary to perform the extra upload, so let's remove it.

### DIFF
--- a/agent/pipeline_uploader.go
+++ b/agent/pipeline_uploader.go
@@ -30,14 +30,31 @@ type PipelineUploader struct {
 	RetrySleepFunc func(time.Duration)
 }
 
+// AsyncUploadFlow will first attempt to perform an async pipeline upload and depending
+// on the API's response it will poll for the upload's status.
+//
+// There are 3 "routes" that are relevant
+// 1. Async Route:  /jobs/:job_uuid/pipelines?async=true
+// 2. Upload Route: /jobs/:job_uuid/pipelines
+// 3. Status Route: /jobs/:job_uuid/pipelines/:upload_uuid
+//
+// In this method, the agent will first upload the pipeline to the Async Route.
+// Then, depending on the response it will behave differetly
+//
+// 1. The Async Route responds 202: poll the Status Route until the upload has beed "applied"
+// 2. The Async Route responds with other 2xx: exit, the upload succeeded synchronously
+// 3. The Async Route responds with other xxx: retry uploading the pipeine to the Async Route
 func (u *PipelineUploader) AsyncUploadFlow(ctx context.Context, l logger.Logger) error {
 	result, err := u.pipelineUploadAsyncWithRetry(ctx, l)
 	if err != nil {
 		return fmt.Errorf("Failed to upload and accept pipeline: %w", err)
 	}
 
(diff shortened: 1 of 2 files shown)


agent/pipeline_uploader_test.go
```diff
@@ -140,9 +140,54 @@ func TestFallbackPipelineUpload(t *testing.T) {
...
+			num529s:         60,
+			expectedSleeps:  genSleeps(59, 5*time.Second),
+			expectedUploads: 60,
+		},
+		{
+			name:            "61_529s",
+			num529s:         61,
+			expectedSleeps:  genSleeps(59, 5*time.Second),
+			expectedUploads: 60,
+			errStatus:       func(i int) *int { return &i }(529),
```
DrJosh9000, line 181: ahaha. 😬

What if `errStatus` was typed `int`, and only compared if it is != 0 ?

agent/pipeline_uploader.go
```diff
@@ -30,14 +30,31 @@ type PipelineUploader struct {
 	RetrySleepFunc func(time.Duration)
 }
 
+// AsyncUploadFlow will first attempt to perform an async pipeline upload and depending
+// on the API's response it will poll for the upload's status.
+//
+// There are 3 "routes" that are relevant
+// 1. Async Route:  /jobs/:job_uuid/pipelines?async=true
+// 2. Upload Route: /jobs/:job_uuid/pipelines
```
DrJosh9000, line 38: "Sync" route? (I'm thinking that "Upload" is also what the Async Route does)

agent/pipeline_uploader.go
```diff
@@ -30,14 +30,31 @@ type PipelineUploader struct {
...
+// 3. Status Route: /jobs/:job_uuid/pipelines/:upload_uuid
+//
+// In this method, the agent will first upload the pipeline to the Async Route.
+// Then, depending on the response it will behave differetly
+//
+// 1. The Async Route responds 202: poll the Status Route until the upload has beed "applied"
+// 2. The Async Route responds with other 2xx: exit, the upload succeeded synchronously
+// 3. The Async Route responds with other xxx: retry uploading the pipeine to the Async Route
```
DrJosh9000, line 46: typo

```suggestion
// 3. The Async Route responds with other xxx: retry uploading the pipeline to the Async Route
```

DrJosh9000 (approved):
LGTM 🌱

agent/pipeline_uploader.go
triarius: It's hard to name things.

agent/pipeline_uploader_test.go
triarius: I was debating whether to do this instead, but then I did not :)

PR #1981, opened 2023-02-27
templates: Tidyups from #1960

DrJosh9000:
Add a blank line, add System as a session type, and copy the change into the other template.


triarius (approved):
LGTM

PR #1982, opened 2023-02-27
Revert "Use sh instead of bash in RunScript"

DrJosh9000:
Reverts buildkite/agent#1974

This might need further consideration.

When testing on a (what I consider fairly stock-standard) Debian machine I have handy, my (what I consider perfectly reasonable) command hook:

```shell
#!/bin/bash
set -euo pipefail
...snip...
```

fails with the error:

```
/tmp/buildkite-agent-bootstrap-hook-runner-3400400749: 2: set: Illegal option -o pipefail
```

because it turns out that `/bin/sh` is symlinked to `dash`, not `bash`, and I think `RunScript` passes scripts directly to whatever shell binary it picks.


triarius (approved):
Thanks for picking this up! I'll take it back to the drawing board.

PR #1990, opened 2023-03-03
Use flock to lock qemu-binfmt installation

triarius:
There was a race contition in the previous code that caused test flake. More details are in the comment.

### DIFF
--- a/.buildkite/steps/build-docker-image.sh
+++ b/.buildkite/steps/build-docker-image.sh
@@ -60,12 +60,36 @@ builder_name=$(docker buildx create --use)
 # shellcheck disable=SC2064 # we want the current $builder_name to be trapped, not the runtime one
 trap "docker buildx rm $builder_name || true" EXIT
 
-# Needed for buildking multiarch
+# Needed a more recent version of QEMU than what's on the host for building multiarch images.
 # See https://github.com/docker/buildx/issues/314
 QEMU_VERSION=6.2.0
 echo "--- Installing QEMU $QEMU_VERSION"
-docker run --privileged --userns=host --rm "tonistiigi/binfmt:qemu-v$QEMU_VERSION" --uninstall qemu-*
-docker run --privileged --userns=host --rm "tonistiigi/binfmt:qemu-v$QEMU_VERSION" --install all
+# There isn't a good way to see if this script has already installed qemu-binfmt.
+# So we unintstall and install it each time. This would create a race condition between the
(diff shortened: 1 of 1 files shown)


.buildkite/steps/build-docker-image.sh
```diff
@@ -60,12 +60,36 @@ builder_name=$(docker buildx create --use)
...
 
-# Needed for buildking multiarch
+# Needed a more recent version of QEMU than what's on the host for building multiarch images.
 # See https://github.com/docker/buildx/issues/314
 QEMU_VERSION=6.2.0
 echo "--- Installing QEMU $QEMU_VERSION"
-docker run --privileged --userns=host --rm "tonistiigi/binfmt:qemu-v$QEMU_VERSION" --uninstall qemu-*
-docker run --privileged --userns=host --rm "tonistiigi/binfmt:qemu-v$QEMU_VERSION" --install all
+# There isn't a good way to see if this script has already installed qemu-binfmt.
+# So we unintstall and install it each time. This would create a race condition between the
```
DrJosh9000, line 68:
```suggestion
# So we uninstall and install it each time. This would create a race condition between the
```

.buildkite/steps/build-docker-image.sh
```diff
@@ -60,12 +60,36 @@ builder_name=$(docker buildx create --use)
...
+# uninstall and install if multiple agents are running on an agent's host.
+#
+# Thus, we use flock(1) to hold an exclusive lock while this is hapening.
```
DrJosh9000, line 71:
```suggestion
# Thus, we use flock(1) to hold an exclusive lock while this is happening.
```

.buildkite/steps/build-docker-image.sh
```diff
@@ -60,12 +60,36 @@ builder_name=$(docker buildx create --use)
...
+# ```bash
+# (
+#   flock -x "$variable"
+#   commands to run under lock
+# ) {variable}>filename
+# ```
+# This first set of round braces run the commands in a subshell. In the subshell, flock(1) is
+# executed with a file descriptor number. This will wait for the lock. The lock will be released
+# by the kernel when the subshell exists (and the file descriptor is closed).
+# See https://linux.die.net/man/1/flock, in particualr the "third form".
```
DrJosh9000, line 82:
```suggestion
# See https://linux.die.net/man/1/flock, in particular the "third form".
```

DrJosh9000 (approved):
✅ 

...With some typo fixes

PR #1993, opened 2023-03-06
Also restart instantiated systemd services

triarius:
We recently modified the after-upgrade scripts of the Linux packages to restart the systemd service if it is running. However, this was not effective in restarting systemd services instantiated from the template we provide.

I also did some clean up:
- Some exported bash variables that did not need to be exported.
- The presence of `DefaultInstance` in the non-templated systemd unit. This created a warning in the journal: 
```
Mar 06 16:09:06 quirinius systemd[1]: /usr/lib/systemd/system/buildkite-agent.service:21: DefaultInstance=>
```

### Edit
Now that the logic has been altered, here is a table showing the possible outputs of `systemctl is-active buildkite-agent 'buildkite-agent@*'`

| `buildkite-agent` | `buildkite-agent@1` | `is-active`<br />exit code |
|---|---|---|
| stopped | stopped | 1 |
| started | stopped | 0 |
| stopped | started | 0 |
| started | started | 0 |

From `man systemctl`, it seems `try-restart` does not fail or attempt to restart stopped services.

### DIFF
--- a/packaging/linux/scripts/after-install-and-upgrade.sh
+++ b/packaging/linux/scripts/after-install-and-upgrade.sh
@@ -97,15 +97,26 @@ fi
 
 # Restart on upgrade, if we can
 if [ "$OPERATION" = "upgrade" ] ; then
-  if [ $BK_SYSTEMD_EXISTS -eq 0 ] && systemctl is-active buildkite-agent --quiet; then
-    # Try restarting with systemd, if the unit is active
-    systemctl try-restart buildkite-agent
-  elif [ $BK_SYSTEMD_EXISTS -eq 0 ] && systemctl is-active 'buildkite-agent@*' --quiet; then
-    # For instantiations of the template. Trying to find one pattern that covers
-    # all instantiations and the non-templated version is possible, it is
-    # `buildkite-agent*`, however that would also cover other service names such
-    # as `buildkite-agentless`. It's safer to do two, more restrictive restarts.
-    systemctl try-restart 'buildkite-agent@*'
+  # Try restarting with systemd, if the unit is active.
+  #
+  # Why two patterns?
+  #
+  # The second pattern is for instantiations of the template.
+  # Trying to find one pattern that covers such instantiations and the base
+  # unit is possible, it is `buildkite-agent*`. However that would also cover
+  # other service names such as `buildkite-agentless`.
+  # It's safer to use two, more restrictive patterns.
+  #
+  # Does using two patterns in the same invocation work?
+  #
+  # `is-active` is used to check if any of the base or template instance
+  # services are running.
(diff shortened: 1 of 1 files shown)


DrJosh9000 (approved):
🤓

packaging/linux/scripts/after-install-and-upgrade.sh
```diff
@@ -97,9 +97,19 @@ fi
...
+      systemctl try-restart buildkite-agent
+    fi
+
+    # For instantiations of the template. Trying to find one pattern that covers
+    # all instantiations and the non-templated version is possible, it is
+    # `buildkite-agent*`, however that would also cover other service names such
+    # as `buildkite-agentless`. It's safer to do two, more restrictive restarts.
+    if systemctl is-active 'buildkite-agent@*' --quiet; then
+      systemctl try-restart 'buildkite-agent@*'
+    fi
```
sj26, line 112: I think this has a secondary change that means folks with systemd installed but not using it to manage buildkite-agent might not see a restart after this change.

There's also no need to check is-active when using try-restart, so if you want to double down on this conditional you can just do a try-restart command.

But I'd suggest:

```suggestion
  # Try restarting with systemd, if the unit is active, either as a base unit or as an instance of the template.
  if [ $BK_SYSTEMD_EXISTS -eq 0 ] && ( systemctl is-active buildkite-agent 'buildkite-agent@*' --quiet ); then
    systemctl try-restart buildkite-agent 'buildkite-agent@*'
```

triarius: > I think this has a secondary change that means folks with systemd installed but not using it to manage buildkite-agent might not see a restart after this change where they would have via the pkill previously.

Good catch. I'll fix.

For the suggestion, on my machine, the pattern that matches the instances of the template unit does not match the base unit and vice versa. So I think 2 of everything is needed.

We could have an `||` with `is-active`.

sj26: Yeah, neat. Using some `||` magic for the detection would be good. I don't think there's any harm to running two `try-restart` commands without switching between, they're meant to fail silently if the unit isn't running.

triarius: I avoided the `||` because I don't think I know how to make parens work in bash, and it would be too long without factoring out the systemd exsistence check.

triarius: Hmm, some people might have both the base unit and the template instances started, and this won't catch the latter in that case. I'll think of something else then.

triarius: Luckily, I think putting the patterns in the same `systemctl` invocation will work.

PR #1995, opened 2023-03-06
WIP: Shell flexibility

DrJosh9000:
There's a long history of attempting to generalise from Bash to sh. Bash has been assumed (in a way that ignores whatever the script author might have wanted) since [2015](https://github.com/buildkite/agent/commit/918d5a0d0d73085943771915e6cdf387fbd5b507). 
Tim tried to switch it to sh in #448, but it was reverted #452. More recently @triarius tried it in #1974, but I reverted it (#1982) because one of my agents had some hooks that assumed Bash (`set -o pipefail` is not plain sh) on a host where `sh` was not a symlink to Bash (a reasonable choice). 

But the more container images do not include bash, the more pressing the need to generalise to `sh`.

So the key change here is to try finding Bash, and fall back to sh with a warning if Bash can't be found.

I considered adding a new flag or experiment, but it all felt wrong. There's an old and widely-accepted way to set the interpreter for a script: the shebang line. The point of assuming Bash to begin with was to enable shebang-less scripts - if someone was to purposely put a shebang line into a script, perhaps we should try to respect it.


utils/file.go
```diff
@@ -29,3 +32,40 @@ func FileExists(filename string) bool {
...
+	return line, nil
+}
+
+// IsPOSIXShell attempts to detect POSIX-compliant shells (e.g bash, sh, zsh)
+func IsPOSIXShell(shell []string) bool {
+	bin := filepath.Base(shell[0])
+
+	if filepath.Base(shell[0]) == "env" {
+		bin = filepath.Base(shell[1])
+	}
```
moskyb, line 63: are there other shebang-isms that we should be checking for here? `#!/path/to/shell` and `#!/usr/bin/env shell` are the only two major ones i can think of, and are both handled here, but  are there others?

hook/scriptwrapper.go
```diff
@@ -126,21 +129,45 @@ func NewScriptWrapper(opts ...scriptWrapperOpt) (*ScriptWrapper, error) {
...
-	var isBashHook bool
-	var isPwshHook bool
+	// Extract any shebang line from the hook to copy into the wrapper.
+	shebang, err := utils.ShebangLine(wrap.hookPath)
+	if err != nil {
+		return nil, fmt.Errorf("Hook path could not be read: %w", err)
+	}
+
+	// Previously we assumed Bash, because the wrapper relied on a Bash-ism.
+	// By using `bk-agent env dump`, the wrapper is compatible with /bin/sh.
```
sj26, line 139: 💚

utils/file.go
DrJosh9000: It's possible on Linux to chain interpreters by making the shebang line refer to another script (with its own shebang line, which could be another script, which ...). I don't know who does this, and I'm inclined to avoid parsing arbitrary chains of scripts unless we come across a use-case.

It's possible for binaries to be in odd places (e.g. I think maybe Solaris puts bash somewhere unusual?) but `filepath.Base` handles that.

It's possible for the interpreter path to be a symlink to one of the supported shell binaries, but with a different name (e.g. `/usr/local/bin/frobulator -> /bin/bash`). The real-world use-cases I know of are symlinking `/bin/sh` to Bash or something else, but that would be handled already. I think following symlinks to resolve the name would be a reasonable addition to this function, but also would be as far as we should go.

hook/scriptwrapper.go
```diff
@@ -126,21 +129,45 @@ func NewScriptWrapper(opts ...scriptWrapperOpt) (*ScriptWrapper, error) {
...
+	//
+	// But if the shebang specifies something weird like Ruby 🤪
+	// the wrapper won't work. Stick to POSIX shells for now.
+	if shebang != "" {
+		// Splitting the shebang by spaces (ignoring quotes, etc) should be
+		// good enough to determine the POSIX-ness of the interpreter.
+		shebs := strings.Fields(strings.TrimPrefix(shebang, "#!"))
+		if !utils.IsPOSIXShell(shebs) {
+			return nil, fmt.Errorf("Hook starts with an unsupported shebang line %q", shebang)
+		}
```
triarius, line 152: I think it's better to split the shebang line in the `utils.IsPOSIXShell` method. It seems like it should know more about what to do with the shebang line than this package.

Also, the two methods in `utils` seem quite related. It might be a good idea to make the package name more specific. A `shebang` package perhaps?

utils/file.go
```diff
@@ -29,3 +32,39 @@ func FileExists(filename string) bool {
...
+		return "", nil
+	}
+	return line, nil
+}
+
+// IsPOSIXShell attempts to detect POSIX-compliant shells (e.g bash, sh, zsh)
+func IsPOSIXShell(shell []string) bool {
+	bin := filepath.Base(shell[0])
+	if bin == "env" {
+		bin = filepath.Base(shell[1])
```
triarius, line 61: If the shebang were just `#!/usr/bin/env` (which is an error that would have been silently ignored before), I think we will panic here. So explicit bounds checking might be necessary.

triarius (commented):
Completely agree with the approach 🎉, I think it's a great change to make.

Have some thoughts on the implementation, however.

hook/scriptwrapper.go
DrJosh9000: Sounds good to me! Done.

utils/file.go
DrJosh9000: Done.

triarius (approved):
Amazing!

shellscript/shellscript.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+	return line, nil
+}
+
+// IsPOSIXShell attempts to detect POSIX-compliant shells (e.g bash, sh, zsh)
+// from either a plain command line, or a shebang line.
+//
+// Examples:
+//   - IsPOSIXShell("/bin/sh") == true
+//   - IsPOSIXShell("/bin/fish") == false
```
moskyb, line 41:
```suggestion
//   - IsPOSIXShell("#!/bin/sh") == true
//   - IsPOSIXShell("#!/bin/fish") == false
```

moskyb (approved):
looking great! awesome work @DrJosh9000

PR #2001, opened 2023-03-08
Update zglob to v0.0.4

moskyb:
The version of [zglob](https://github.com/mattn/go-zglob) we're using is positively ancient, and doesn't support useful globbing features like negation or bracketing.

This PR updates our version of go-zglob to the latest, v0.0.4. This update includes both negation and bracketing. Yay!

PR #2002, opened 2023-03-08
Prefer `errors.Is(err, <whatever>)` over `err == <whatever>`

moskyb:
While investigating #2001, we ran into an issue where a direct error comparison `err == os.ErrNotExist` was causing test failures, where `errors.Is(err, os.ErrNotExist)` was okay. This is because `errors.Is` will unwrap errors, and return true if any of the unwrapped errors match the second arg.

**This PR:** Replaces most of the usages of `err == <whatever>` with `errors.Is(err, <whatever>)`. This makes these comparisons more robust, and establishes a better SOP within the agent codebase.

It also adds error wrapping to the artifact uploader, which also didn't quite fit in #2001, and is kinda-sorta more relevant to this PR.

### DIFF
--- a/agent/artifact_uploader.go
+++ b/agent/artifact_uploader.go
@@ -71,17 +71,17 @@ func (a *ArtifactUploader) Upload(ctx context.Context) error {
 	// Create artifact structs for all the files we need to upload
 	artifacts, err := a.Collect()
 	if err != nil {
-		return err
+		return fmt.Errorf("collecting artifacts: %w", err)
 	}
 
 	if len(artifacts) == 0 {
 		a.logger.Info("No files matched paths: %s", a.conf.Paths)
-	} else {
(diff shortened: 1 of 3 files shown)


agent/artifact_uploader.go
```diff
@@ -71,17 +70,17 @@ func (a *ArtifactUploader) Upload(ctx context.Context) error {
...
-		return err
+		return fmt.Errorf("collecting artifacts: %w", err)
 	}
 
 	if len(artifacts) == 0 {
 		a.logger.Info("No files matched paths: %s", a.conf.Paths)
-	} else {
-		a.logger.Info("Found %d files that match \"%s\"", len(artifacts), a.conf.Paths)
+		return nil
+	}
```
DrJosh9000, line 79: 👍👍👍👍👍👍👍👍👍

DrJosh9000 (approved):
🥹

triarius (approved):
Yay!

PR #2003, opened 2023-03-08
Add tags from env variables provided by the controller in agent-stack-k8s if kuberenetes-exec experiment is enabled

triarius:
These are set by the controller in the companion PR: https://github.com/buildkite/agent-stack-k8s/pull/137


sj26 (approved):
Nice, and nicely tested :+1:

PR #1967, opened 2023-02-23
Add new priority modes to --spawn-with-priority

DrJosh9000:
In #1929 it was pointed out that, when using `--spawn-with-priority`, the workload tends to be unevenly spread. Agents with fewer spawn will not have any spawn with as high a priority as agents with more spawn. Because work is assigned to available agents with the highest priority first, work ends up assigned only to agents with more spawn, and no work is assigned to the agents with fewer spawn (until the agents with more spawn become busy). This PR changes `--spawn-with-priority` to be a string flag, letting the user choose the priority assignment.

Example commands:

```shell
# Existing behaviour: Start 5 agents with default priority.
buildkite-agent start --build-path=~/tmp --spawn=5

# No longer valid (no value provided for --spawn-with-priority)
buildkite-agent start --build-path=~/tmp --spawn=5 --spawn-with-priority

# New behaviour: Start 5 agents with priorities [0, 1, 2, 3, 4].
buildkite-agent start --build-path=~/tmp --spawn=5 --spawn-with-priority=ascending

# New behaviour: Start 5 agents with priorities [0, -1, -2, -3, -4].
buildkite-agent start --build-path=~/tmp --spawn=5 --spawn-with-priority=descending

# New behaviour: Start 5 agents with priorities [42, 43, 44, 45, 46].
buildkite-agent start --build-path=~/tmp --spawn=5 --spawn-with-priority=ascending --priority=42

# New behaviour: Start 5 agents with priorities [42, 41, 40, 39, 38].
buildkite-agent start --build-path=~/tmp --spawn=5 --spawn-with-priority=descending --priority=42
```

Fixes #1929

Edit: changed from using an experiment to a single flag
Edit 2: changed from `inverse` to a variety of options.
Edit 3: fixed the example commands
Edit 4: changed away from the bool-like flag (fully breaking change)

Issue #1929, opened 2023-01-24
Distribute jobs more evenly across hosts


nick-f:
**Is your feature request related to a problem? Please describe.**

We have hosts that can have different numbers of spawned agents. The priority for these is set by the spawn ID with the [spawn-with-priority](https://buildkite.com/docs/agent/v3/cli-start#spawn-with-priority) option.

The way that priorities work now are that higher numbered priorities are used first.

If `hostA` has 1 spawned agent running and `hostB` has 3 spawned agents running, `hostB` is going to be running at least 2 or maybe 3 tests while `hostA` is sitting idle waiting for jobs to be assigned to it.

Assuming all agents are idle, the order that jobs are assigned is:

1. `hostB agent3`
2. `hostB agent2`
3. `hostA agent1` or `hostB agent1`
4. `hostA agent1` or `hostB agent1` (whichever was not given a job before)

**Describe the solution you'd like**

The next agent would be chosen based on the spawned agent utilisation of each host.

`hostA` has 1 spawned agent with 1 job running (100% utilisation)
`hostB` has 3 spawned agents with 1 job running (33% utilisation)
`hostC` has 5 spawned agents with 1 job running (20% utilisation)

The next host to be assigned work would be `hostC` because the current utilisation is the lowest. The agent on `hostC` that is given the work is determined based on the priority.

(Ideally that spawned agent prioritisation could also be flipped so `hostC agent1` would be the first to be used instead of `hostC agent5`. Having that as a configuration option would be ace! I can split that out into a separate feature request if needed.)

`hostA` has 1 spawned agent with 1 job running (100% utilisation)
`hostB` has 3 spawned agents with 1 job running (33% utilisation)
`hostC` has 5 spawned agents with 2 jobs running (40% utilisation)

Now, with `hostC` utilisation at 40%, the next host to be assigned a job would be `hostB`.

**Describe alternatives you've considered**

I've spoken with Jarryd from Buildkite about this issue, but there doesn't appear to be any existing solutions for this use case.

Setting host priority doesn't work for situations where there are, say, two agents on a host. If that host is meant to be used first due to host priority, then the same situation would occur as the original problem, where one host is doing all the work while the other is sitting idle.

**Additional context**

We set the number of spawn agents in each host's config.

There are a variety of hardware profiles for our hosts, so some can only run one agent at a time, some run 3, and we're about to start trialling hosts that should be able to run 6 or more agents 🤞

triarius:
[...]

PR #2004, opened 2023-03-09
Add `descending-spawn-priority` experiment.

DrJosh9000:
This is the MVP of #1967, so that we have a workaround landed for #1929.

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -900,8 +900,14 @@ var AgentStartCommand = cli.Command{
 			registerReq.Name = strings.ReplaceAll(cfg.Name, "%spawn", strconv.Itoa(i))
 
 			if cfg.SpawnWithPriority {
-				l.Info("Assigning priority %s for agent %d", strconv.Itoa(i), i)
-				registerReq.Priority = strconv.Itoa(i)
+				p := i
+				if experiments.IsEnabled("descending-spawn-priority") {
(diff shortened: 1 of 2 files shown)


triarius (approved):
LGTM

clicommand/agent_start.go
```diff
@@ -900,8 +900,14 @@ var AgentStartCommand = cli.Command{
 			registerReq.Name = strings.ReplaceAll(cfg.Name, "%spawn", strconv.Itoa(i))
 
 			if cfg.SpawnWithPriority {
-				l.Info("Assigning priority %s for agent %d", strconv.Itoa(i), i)
-				registerReq.Priority = strconv.Itoa(i)
+				p := i
+				if experiments.IsEnabled("descending-spawn-priority") {
+					// This experiment helps jobs be assigned across all hosts
+					// in cases where the value of --spawn varies between hosts.
+					p = -i
```
gw0, line 907: Why just use negative values starting with `-1` instead of subtracting from what `priority` is set in the configuration? With something in the lines of:
```
p = cfg.Priority - i + 1
```

I would expect it to behave like mentioned in the PR https://github.com/buildkite/agent/pull/1967:
```
# New behaviour: Start 5 agents with priorities [42, 41, 40, 39, 38].
buildkite-agent start --build-path=~/tmp --spawn=5 --spawn-with-priority=descending --priority=42
```

PR #2005, opened 2023-03-09
Add buildx to docker images

triarius:
Some of the docker image building solutions on Kubernetes involve a [buildkit](https://docs.docker.com/build/buildkit/) service. The plugin `buildx` is used to give the docker CLI access to these, but it is not installed on our default agent image, even though the docker CLI is. If it were, we could instruct customers to use the agent image as the image in their image building step.


DrJosh9000 (approved):
🌴😎🍹

PR #2006, opened 2023-03-10
Add an option to tee job logs to a non-temp file

goodspark:
We want to collect job logs and import them. I think this will let us collect the contents (provided the job logs stay around long enough to be collected by our daemons).

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -67,6 +67,7 @@ type AgentStartConfig struct {
 	BootstrapScript             string   `cli:"bootstrap-script" normalize:"commandpath"`
 	CancelGracePeriod           int      `cli:"cancel-grace-period"`
 	EnableJobLogTmpfile         bool     `cli:"enable-job-log-tmpfile"`
+	JobLogFilepath              string   `cli:"job-log-filepath"`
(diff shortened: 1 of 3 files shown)


clicommand/agent_start.go
```diff
@@ -769,6 +776,7 @@ var AgentStartCommand = cli.Command{
 			DisconnectAfterIdleTimeout: cfg.DisconnectAfterIdleTimeout,
 			CancelGracePeriod:          cfg.CancelGracePeriod,
 			EnableJobLogTmpfile:        cfg.EnableJobLogTmpfile,
+			JobLogTmpDirPath:           cfg.JobLogTmpDirPath,
 			Shell:                      cfg.Shell,
 			RedactedVars:               cfg.RedactedVars,
 			AcquireJob:                 cfg.AcquireJob,
```
goodspark, line 782: After testing this a bit more, it might only work for long running jobs with our log collector. I wonder how yall would feel if we just exposed a different flag that had the agents write all job logs to a persistent file (which could be rotated by the system)?

goodspark:
I think I'll close this. My primary use case is either running the agent in a container or I have a VM that's already setup to pipe the agents' stdout to a log. In either case, stdout is the common denominator. I can prep another PR to route job logs to the parent stdout.

PR #2007, opened 2023-03-10
Fix DD trace setup warning

goodspark:
We're seeing this in our job logs:
```
2023/03/10 00:53:18 Datadog Tracer v1.46.0 WARN: ddtrace/tracer: deprecated config WithServiceName should not be used with `WithService` or `DD_SERVICE`; integration service name will not be set.
```


moskyb (approved):
TY @goodspark!

PR #2009, opened 2023-03-10
Add option for outputting structured logs for collection and searching

goodspark:
We want to collect our logs into a separate system that makes it easier to search and run queries.
We also want to ability to filter logs based on metadata like the pipeline it came from.

To facilitate simpler log collection, this adds a field that allows job logs to be written out to the stdout of the agent's process. This is more idiomatic for log collection in containerized systems as well as systems that have process managers like systemd or launchd (which have fields for routing stdout to different log files).


process/structure.go
```diff
@@ -0,0 +1,37 @@
...
+	w        io.Writer
+	metadata map[string]string
+	encoder  *json.Encoder
+}
+
+func NewStructure(w io.Writer, metadata map[string]string) *Structure {
+	var metadata_copy = map[string]string{}
+	for k, v := range metadata {
+		metadata_copy[k] = v
+	}
```
moskyb, line 21: why copy the map here? if it's necessary for some reason, this can be replaced with:
```suggestion
	metadataCopy := maps.Clone(metadata)
```
but i suspect it can be left out

process/structure.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"encoding/json"
+	"io"
+	"strings"
+	"time"
+)
+
+// Structure outputs messages written to it in a structured format, including metadata.
+type Structure struct {
```
moskyb, line 11: maybe this could be
```suggestion
type StructuredWriter struct {
```

moskyb (changes requested):
hey, this is a really really cool feature to add, and i'd love to include it in the agent. it makes a lot of use cases for "i want to analyse my job logs" a whole bunch easier, and frankly, i'm surprised we don't have it already. awesome work all around 🎉

i've made a couple of suggestions, mostly around keeping consistency with the agent's existing json logger. there's a couple of formatting nits and such in there too, but this is awesome work, and i'd love to get it shipped 🙌

process/structure.go
```diff
@@ -0,0 +1,37 @@
...
+func (s *Structure) Write(data []byte) (n int, err error) {
+	// When writing as a structured log, trailing newlines and carriage returns generally don't make sense.
+	var msg = strings.TrimRight(string(data), "\r\n")
+	var log = map[string]string{"message": msg, "timestamp": time.Now().UTC().Format(time.RFC3339)}
+	for k, v := range s.metadata {
+		log[k] = v
+	}
+	if err = s.encoder.Encode(log); err != nil {
+		return 0, err
+	}
```
moskyb, line 35: i was thinking of making a comment to the tune of "should we be calling `json.Marshal` (or equivalent) for every line? is the method used in the agent json logger more efficient?", but as it turns out, using the json library is notably more efficient:

```Go
package main_test

import (
	"encoding/json"
	"fmt"
	"strings"
	"testing"
)

func makeTestMap() map[string]string {
  // ...
}

func BenchmarkUnmarshall(b *testing.B) {
	for i := 0; i < b.N; i++ {
		mm := makeTestMap()
		marsh, _ = json.Marshal(mm)
	}
}

func BenchmarkStringBuilder(b *testing.B) {
	for i := 0; i < b.N; i++ {
		mm := makeTestMap()
		b := strings.Builder{}
		for k, v := range mm {
			b.WriteString(fmt.Sprintf("%q:%q,", k, v))
		}

		s = fmt.Sprintf("{%s}", strings.TrimSuffix(b.String(), ","))
	}
}
```

```
^_^ (~/b/scratch-go) go test -bench=.
goos: darwin
goarch: arm64
pkg: github.com/moskyb/scratch-go
BenchmarkUnmarshall-10       	  491529	      2445 ns/op
BenchmarkStringBuilder-10    	  365670	      3185 ns/op
PASS
ok  	github.com/moskyb/scratch-go	3.413s
```

process/structure.go
goodspark: You know, I might just try to adapt or wrap the existing Logger + JSONWriter code into the job logger stuff. It's basically the achieving the same end result.

moskyb: yeah, i deleted the comment because i realised that i'd misread the benchmark and that the thing that i had said was quicker was actually slower, and i'd just misread the benchmark ahaha.

regardless, i think either approach is is totally fine, we can probably spare 1000 nanoseconds here or there

clicommand/agent_start.go
```diff
@@ -743,6 +753,10 @@ var AgentStartCommand = cli.Command{
 			l.Fatal("The given tracing backend %q is not supported. Valid backends are: %q", cfg.TracingBackend, maps.Keys(tracetools.ValidTracingBackends))
 		}
 
+		if cfg.LogFormat == "json" && !cfg.WriteJobLogsToStdout {
+			l.Fatal("JSON log output doesn't make sense without writing job logs to the agent's stdout.")
+		}
+
```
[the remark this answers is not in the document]
goodspark:
```suggestion
```

PR #2010, opened 2023-03-13
Set a non-zero exit status when a job is cancelled in Kubernetes

triarius:
If a job is cancelled via the Buildkite API, for example, by clicking the cancel button on the website and the bootstrap container in the Kubernetes pod has not started, the exit status will be 0. This causes the build to pass in situations where it should not.

In this PR, we detect if child containers of the agent start container have not run and provide a non-zero exit status. In addition, the agent start container will emit a faked job log that informs the user that a container may not have run. Otherwise, if the Buildkite job was automatically cancelled, as will be the case after https://github.com/buildkite/agent-stack-k8s/pull/143 is merged, the user would not have a good idea what the cause of the cancellation was.

## Screenshot
2023-03-15T21:57:25,215979237+11:00

### DIFF
--- a/kubernetes/kubernetes.go
+++ b/kubernetes/kubernetes.go
@@ -165,12 +165,21 @@ func (r *Runner) WaitStatus() process.WaitStatus {
 			return waitStatus{Code: client.ExitStatus}
 		}
 
+		// use an unusual status code to distinguish this unusual state
 		if client.State == stateUnknown {
-			// use an unusual status code to distinguish this unusual state
 			ws.Code += -10
 		}
 	}
 	return ws
+}
+
+func (r *Runner) ClientStateUnknown() bool {
+	for _, client := range r.clients {
+		if client.State == stateUnknown {
+			return true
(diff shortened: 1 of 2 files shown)


kubernetes/kubernetes.go
```diff
@@ -159,17 +159,29 @@ func (w waitStatus) Signaled() bool {
...
 	for _, client := range r.clients {
 		if client.ExitStatus != 0 {
 			return waitStatus{Code: client.ExitStatus}
 		}
-		// just return any ExitStatus if we don't find any "interesting" ones
-		ws = waitStatus{Code: client.ExitStatus}
+
+		// use an unusual status code to distinguish this unusual state
+		if client.State == stateUnknown {
+			ws.Code += -10
```
moskyb, line 170: [super nit]

```suggestion
			ws.Code -= 10
```

moskyb (approved):
LGTM! one nit, but feel free to merge as it

PR #2014, opened 2023-03-14
Avoid holding full job logs

DrJosh9000:
The `String` method on `bytes.Buffer` does not consume the buffer in the same way that `Read` does - it leaves the contents in place. Because `output` is never `Read` or `Truncate`-d, it never shrinks. Also Go almost certainly copies the internal `[]byte` into a new `string` (or vice-versa).

It appears the agent never needs the whole copy of the logs, though.

This replaces the buffer implementation with a new one (in a new file, with a test) that _only_ has `Write` and `ReadAndTruncate` methods. This implementation, directly working on a byte slice, is (imo) at least as simple as a `bytes.Buffer` (but then you have to know how much you want to read, which, yes, you can get with `Len`, or you can use `io.ReadAll`, but it's about the same).

The output chunking logic is reworked, since it no longer has to track where in the (full) output buffer the next chunk will be.

Finally, it now uses `[]byte` throughout the data path, avoiding any potential copying to and from `string`.

### DIFF
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -352,7 +352,7 @@ func (r *JobRunner) Run(ctx context.Context) error {
 			environmentCommandOkay = false
 
 			// Ensure the Job UI knows why this job resulted in failure
-			r.logStreamer.Process("pre-bootstrap hook rejected this job, see the buildkite-agent logs for more details")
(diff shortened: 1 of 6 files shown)


moskyb (approved):
🎉 awesome work

pda (commented):
Very nice 👌🏼

PR #2015, opened 2023-03-15
Why did we have three groups of go modules?

moskyb:
we now have two groups of go modules, one for direct dependencies and one for indirect, as ~~god~~ rsc intended


triarius (approved):
🧹

DrJosh9000 (approved):

PR #2016, opened 2023-03-15
ansi-timestamps: Compute prefixes at start of line

DrJosh9000:
Looking at some of the log display code, we may need ANSI timestamps to reflect the start of each line, rather than the end. (sigh)

Fortunately, this admits a buffer-free implementation!

This PR should avoid the over-eager computation of the _next_ line's prefix, which was the main bug fixed with #1940.


triarius (approved):
Nice

PR #2017, opened 2023-03-16
Make env.Environment.Merge and .Apply operate in-place

moskyb:
This prevents the bootstrap from accidentally copying the environment when processing the output of a scriptwrapper, which was preventing environment updates from the job api from showing up


triarius (approved):
I think this needs a followup with a test to demonstrate how the job-api broke.

moskyb:
@triarius sounds good 👍

PR #2023, opened 2023-03-20
Re-generate code

triarius:
I noticed some code was supposed to be generated, but I had been manually adding to it. So I overhauled some of the code generation and created a CI job to ensure code generation had been done.

~~One notable addition I made was to use code generation for the `VERSION` file that was previously handcrafted. While I have not automated this yet, this should bring us a step closer to automating the tag generation process.~~

Read the individual commits for more details.


version/gen_version.sh
moskyb, line 1: does this generator change anything about the release process? i can imagine that it means that instead of manually editing the version file, CI could update itself on new tag creation, is that right?

moskyb (approved):
LGTM! love clearing out some cruft

version/gen_version.sh
moskyb: oosh, the UX for github's "comment on file" feature leaves a little to be desired...

triarius: Having had a look at it, I think this will break the release process because there will be at least one commit where the tag does not match what's in the version file. I think we should eliminate the version file all together. But that's for another day.

I'm going to revert the generation of the version file and merge the rest.

PR #2028, opened 2023-03-22
ArtifactUploader API calls: faster timeout & retry

pda:
When ArtifactUploader calls `UpdateArtifacts` (`PUT /v3/jobs/{uuid}/artifacts`), in the uncommon event of something stalling server-side or network-wise, the server and the client each take 60 seconds to time out, and the retry would wait a further 5 seconds. `UpdateArtifacts` is idempotent and called inside a retry loop, so we're better off timing out sooner and retrying sooner.

The p99 response time for this endpoint is ~50ms, so timing out after 5 seconds (100x the p99) seems more than reasonable. And reducing the initial retry delay (but backing off exponentially) also keeps things moving faster.

This means if `buildkite-agent artifact upload …` is invoked with a reasonable _external_ timeout of e.g. 30 or 60 seconds, a single server/network stall won't use up all that time, and a repeated server/network stall will be more visible due to the timeouts being logged.

Only the verified-idempotent `UpdateArtifacts` API call is updated in this pull request, although the same treatment could be applied to other endpoints, e.g. the API call that creates the artifact batch is also idempotent, ~but out of scope for this PR~.

Update: I've also applied a timeout (10 seconds) to `ArtifactBatchCreator` making the `CreateArtifacts` call; that one has a longer p99/max duration server side, hence the longer context timeout for this one. I've verified it is indeed idempotent. Also, the agent splits large numbers of artifacts into batches of up to 30, and this timeout is per-batch, so it won't scale beyond that with the total number of artifacts being uploaded.

### DIFF
--- a/agent/artifact_uploader.go
+++ b/agent/artifact_uploader.go
@@ -327,10 +327,14 @@ func (a *ArtifactUploader) upload(ctx context.Context, artifacts []*api.Artifact
 
 				// Update the states of the artifacts in bulk.
 				err := roko.NewRetrier(
-					roko.WithMaxAttempts(10),
-					roko.WithStrategy(roko.Constant(5*time.Second)),
+					// TODO: roko.ExponentialSubsecond(500*time.Millisecond) WithMaxAttempts(10)
+					// Meanwhile, 8 roko.Exponential(2sec) attempts is 1,2,4,8,16,32,64 seconds delay (~2 mins)
+					roko.WithMaxAttempts(8),
+					roko.WithStrategy(roko.Exponential(2*time.Second, 0)),
 				).DoWithContext(ctx, func(r *roko.Retrier) error {
-					if _, err := a.apiClient.UpdateArtifacts(ctx, a.conf.JobID, statesToUpload); err != nil {
+					ctxShort, cancel := context.WithTimeout(ctx, 5*time.Second)
+					defer cancel()
+					if _, err := a.apiClient.UpdateArtifacts(ctxShort, a.conf.JobID, statesToUpload); err != nil {
 						a.logger.Warn("%s (%s)", err, r)
 						return err
 					}
--- a/agent/artifact_uploader.go
+++ b/agent/artifact_uploader.go
@@ -275,9 +275,10 @@ func (a *ArtifactUploader) upload(ctx context.Context, artifacts []*api.Artifact
 
 	// Create the artifacts on Buildkite
 	batchCreator := NewArtifactBatchCreator(a.logger, a.apiClient, ArtifactBatchCreatorConfig{
-		JobID:             a.conf.JobID,
-		Artifacts:         artifacts,
-		UploadDestination: a.conf.Destination,
+		JobID:                  a.conf.JobID,
+		Artifacts:              artifacts,
+		UploadDestination:      a.conf.Destination,
+		CreateArtifactsTimeout: 10 * time.Second,
 	})
 
 	artifacts, err = batchCreator.Create(ctx)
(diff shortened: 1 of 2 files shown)


agent/artifact_uploader.go
```diff
@@ -327,8 +327,10 @@ func (a *ArtifactUploader) upload(ctx context.Context, artifacts []*api.Artifact
 
 				// Update the states of the artifacts in bulk.
 				err := roko.NewRetrier(
-					roko.WithMaxAttempts(10),
-					roko.WithStrategy(roko.Constant(5*time.Second)),
+					// TODO: roko.ExponentialSubsecond(1*time.Second) WithMaxAttempts(10)
+					// Meanwhile, 8 roko.Exponential(2sec) attempts is 1,2,4,8,16,32,64 seconds delay (~2 mins)
+					roko.WithMaxAttempts(8),
+					roko.WithStrategy(roko.Exponential(2*time.Second, 0)),
 				).DoWithContext(ctx, func(r *roko.Retrier) error {
```
triarius, line 334: You should be able to use `context.WithTimeout` to avoid having to set the timeout on the http client. Just make sure you make a new context and only use it in this request!

pda: Oh nice, thanks 👍🏼

agent/artifact_uploader.go
```diff
@@ -327,8 +327,10 @@ func (a *ArtifactUploader) upload(ctx context.Context, artifacts []*api.Artifact
 
 				// Update the states of the artifacts in bulk.
 				err := roko.NewRetrier(
-					roko.WithMaxAttempts(10),
-					roko.WithStrategy(roko.Constant(5*time.Second)),
+					// TODO: roko.ExponentialSubsecond(1*time.Second) WithMaxAttempts(10)
+					// Meanwhile, 8 roko.Exponential(2sec) attempts is 1,2,4,8,16,32,64 seconds delay (~2 mins)
+					roko.WithMaxAttempts(8),
+					roko.WithStrategy(roko.Exponential(2*time.Second, 0)),
```
triarius, line 333:
```suggestion
				).DoWithContext(ctx, func(r *roko.Retrier) error {
					timeOutCtx := context.WithTimeout(ctx, 500 * time.Millisecond)
					
					if _, err := a.apiClient.UpdateArtifacts(timeOutCtx, a.conf.JobID, statesToUpload); err != nil {
```

You should be able to use `context.WithTimeout` to avoid having to set the timeout on the http client. Just make sure you make a new context and only use it in this request!

PS don't accept the suggestion, it's not going to work because the diffs can't extend to unedited portions of the file.

agent/artifact_uploader.go
triarius, line 334:
```
				).DoWithContext(ctx, func(r *roko.Retrier) error {
					timeoutCtx, cancel := context.WithTimeout(ctx, 500 * time.Millisecond)
					defer cancel()

					if _, err := a.apiClient.UpdateArtifacts(timeoutCtx, a.conf.JobID, statesToUpload); err != nil {
```

You should be able to use `context.WithTimeout` to avoid having to set the timeout on the http client. Just make sure you make a new context and only use it in this request attempt!

pda:
Added to PR description:

> Update: I've also applied a timeout (10 seconds) to `ArtifactBatchCreator` making the `CreateArtifacts` call; that one has a longer p99/max duration server side, hence the longer context timeout for this one. I've verified it is indeed idempotent. Also, the agent splits large numbers of artifacts into batches of up to 30, and this timeout is per-batch, so it won't scale beyond that with the total number of artifacts being uploaded.

pda:
> I've also applied a timeout (10 seconds) to ArtifactBatchCreator making the CreateArtifacts call

Here's what that looks like when running in a build job, with a simulated slow response from CreateArtifacts:

image

agent/artifact_uploader.go
triarius: We may want to look into something like this: https://go.dev/blog/http-tracing. Instrumenting the client with [http tracing](https://pkg.go.dev/net/http/httptrace) when in debug mode should give an even better view of where this gets stuck.

Looks like you can also localise it using contexts, so we may want to make it super specific to start with.

triarius (approved):
LGTM. Maybe make a follow-ups for the subsecond backoff and httptrace if you think you need them.

pda:
These artifact retry delay have been further hastened by #2069

PR #2030, opened 2023-03-27
Warn on unknown experiments

moskyb:
**This PR:** Makes it so that if a user enables an unknown experiment (ie, one that's not in the codebase), we'll emit a warning on the logger.

Additionally, for us, it creates a canonical list of available experiments in code, meaning that to find which experiments are in progress, we no longer have to do a global search across the repo for `experiments.IsEnbabled(` to find all of them.

### DIFF
--- a/experiments/experiments.go
+++ b/experiments/experiments.go
@@ -4,11 +4,26 @@
 // It is intended for internal use by buildkite-agent only.
 package experiments
 
-var experiments = make(map[string]bool)
+var (
+	Available = map[string]struct{}{
+		"job-api":                       {},
+		"kubernetes-exec":               {},
+		"ansi-timestamps":               {},
+		"git-mirrors":                   {},
+		"flock-file-locks":              {},
+		"resolve-commit-after-checkout": {},
+		"descending-spawn-priority":     {},
+		"inbuilt-status-page":           {},
+	}
+
+	experiments = make(map[string]bool, len(Available))
+)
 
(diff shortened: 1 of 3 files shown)


experiments/experiments.go
```diff
@@ -4,11 +4,26 @@
...
+	experiments = make(map[string]bool, len(Available))
+)
 
 // Enable a particular experiment in the agent.
-func Enable(key string) {
+func Enable(key string) (known bool) {
 	experiments[key] = true
+	_, known = Available[key] // is the experiment they've enabled one that we know of?
+	return known
 }
```
triarius, line 27: I know a lot of people are vehemently against named return values. Is it necessary to trigger them in this case 😛

triarius (approved):
This will be great to have!

experiments/experiments.go
moskyb: dent-you-either-die-a-hero-or-you-live-long-enough-to-see-yourself-become-the-villain

i'm a known named-return hater, buttttttttttt..... i think in this case the actual meaning of the return value might not be as clear as if it weren't named - if i were to see a function `func Enable(key string) bool`, i think my first guess of what the return value means would be "was it actually enabled or not". just this once, i'll ~enter the named return radicalisation pipeline~ allow it

triarius: I'm on team "change the method name if the return value is ambiguous"

PR #2032, opened 2023-03-29
Promote git-mirrors experiment to full functionality

moskyb:
Use of the git mirrors experiment will now be solely controlled by the --git-mirrors-dir flag

### DIFF
--- a/experiments/experiments.go
+++ b/experiments/experiments.go
@@ -4,6 +4,14 @@
 // It is intended for internal use by buildkite-agent only.
 package experiments
 
+import (
+	"fmt"
+
+	"github.com/buildkite/agent/v3/logger"
+)
+
+type State string
+
 const (
 	PolyglotHooks              = "polyglot-hooks"
 	JobAPI                     = "job-api"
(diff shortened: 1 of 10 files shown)


triarius (approved):
LGTM, thanks for cleaning this up. I assume we need to remove it from the docs website too.

experiments/experiments.go
```diff
@@ -30,14 +42,39 @@ var (
 		NormalisedUploadPaths:      {},
 	}
 
+	Promoted = map[string]string{
+		GitMirrors: fmt.Sprintf("The %s experiment has been promoted to a stable feature. You can safely remove the `--experiment %s` flag without affecting anything", GitMirrors, GitMirrors),
```
moskyb, line 46: the `without affecting anything` wording is a bit clunky here, would appreciate pointers on how to england good

triarius (approved):
🎉

experiments/experiments.go
moskyb: nah, i like it 💪

triarius (approved):

PR #2034, opened 2023-03-31
humanize bytes in the artifact upload/download log

xxchan:


moskyb (approved):
this is a lovely usability change @xxchan! thanks for your contribution 💚

PR #2040, opened 2023-04-03
Experiment: Polyglot hooks

moskyb:
**This PR:** Removes the requirement that hooks for the buildkite-agent be written in a shell scripting language a la bash. With the use of this branch, hooks can also either be compiled executable binaries (ie those produced by languages like Golang or Rust), or any interpreted scripting language (ie python, ruby, lua), so long as an appropriate shebang is specified at the top of the file.

note that neither binary hooks nor non-shell script hooks have environment variable changes captured as bash script hooks do; they _must_ use the Job API to modify the environment of the job.

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -297,39 +297,20 @@ func (b *Bootstrap) executeHook(ctx context.Context, hookCfg HookConfig) error {
 
 	b.shell.Headerf("Running %s hook", hookName)
 
-	isBinary, err := shellscript.IsBinaryExecutable(hookCfg.Path)
+	hookType, err := hook.Type(hookCfg.Path)
 	if err != nil {
-		return fmt.Errorf("checking if %s hook is a binary: %w", hookName, err)
+		return fmt.Errorf("determining hook type for %q hook: %w", hookName, err)
 	}
 
-	shebangLine, err := shellscript.ShebangLine(hookCfg.Path)
-	if err != nil {
-		return fmt.Errorf("reading shebang line for %s hook: %w", hookName, err)
-	}
-
-	switch {
-	case isBinary:
+	switch hookType {
+	case hook.TypeBinary, hook.TypeScript:
 		// It's a binary, so we'll just run it directly, no wrapping needed or possible
 		return b.runUnwrappedHook(ctx, hookName, hookCfg)
-	case shellscript.IsPOSIXShell(shebangLine):
+	case hook.TypeShell:
 		// It's definitely a shell script, wrap it so that we can snaffle the changed environment variables
 		return b.runWrappedShellScriptHook(ctx, hookName, hookCfg)
-	case shebangLine != "" && !shellscript.IsPOSIXShell(shebangLine):
-		// It's an interpretable-something with a shebang line, and it almost certainly isn't a shell script,
-		// so it won't be capable of setting environment variables (outside of the job api), so just run it directly
-		return b.runUnwrappedHook(ctx, hookName, hookCfg)
-	case shebangLine == "":
-		// If there's no shebang line in the hook, and it's not a binary, so let's just assume that it's a shell script
-		// The OS will complain if we do something stupid, so we'll just give it a shot
-		return b.runWrappedShellScriptHook(ctx, hookName, hookCfg)
 	default:
-		return fmt.Errorf(`the buildkite agent wasn't able to determine the type of hook it was trying to run.
-This is a bug, please contact support@buildkite.com and/or submit an issue on the github.com/buildkite/agent repo with the following information:
-Hook scope: %q
-Hook path: %q
-Hook type: %q
-Shebang line: %q
-Is binary?: %t`, hookCfg.Scope, hookCfg.Path, hookName, shebangLine, isBinary)
+		return fmt.Errorf("unknown hook type %q for %q hook", hookType, hookName)
 	}
 }
 
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -297,6 +297,10 @@ func (b *Bootstrap) executeHook(ctx context.Context, hookCfg HookConfig) error {
 
 	b.shell.Headerf("Running %s hook", hookName)
 
+	if !experiments.IsEnabled(experiments.PolyglotHooks) {
+		return b.runWrappedShellScriptHook(ctx, hookName, hookCfg)
+	}
+
 	hookType, err := hook.Type(hookCfg.Path)
 	if err != nil {
 		return fmt.Errorf("determining hook type for %q hook: %w", hookName, err)
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -297,11 +297,56 @@ func (b *Bootstrap) executeHook(ctx context.Context, hookCfg HookConfig) error {
 
 	b.shell.Headerf("Running %s hook", hookName)
 
+	isBinary, err := shellscript.IsBinaryExecutable(hookCfg.Path)
+	if err != nil {
+		return fmt.Errorf("checking if %s hook is a binary: %w", hookName, err)
+	}
+
+	shebangLine, err := shellscript.ShebangLine(hookCfg.Path)
+	if err != nil {
+		return fmt.Errorf("reading shebang line for %s hook: %w", hookName, err)
+	}
+
+	switch {
+	case isBinary:
+		// It's a binary, so we'll just run it directly, no wrapping needed or possible
+		return b.runUnwrappedHook(ctx, hookName, hookCfg)
+	case shellscript.IsPOSIXShell(shebangLine):
+		// It's definitely a shell script, wrap it so that we can snaffle the changed environment variables
+		return b.runWrappedShellScriptHook(ctx, hookName, hookCfg)
+	case shebangLine != "" && !shellscript.IsPOSIXShell(shebangLine):
+		// It's an interpretable-something with a shebang line, and it almost certainly isn't a shell script,
+		// so it won't be capable of setting environment variables (outside of the job api), so just run it directly
+		return b.runUnwrappedHook(ctx, hookName, hookCfg)
+	case shebangLine == "":
+		// If there's no shebang line in the hook, and it's not a binary, so let's just assume that it's a shell script
+		// The OS will complain if we do something stupid, so we'll just give it a shot
+		return b.runWrappedShellScriptHook(ctx, hookName, hookCfg)
+	default:
+		return fmt.Errorf(`the buildkite agent wasn't able to determine the type of hook it was trying to run.
+This is a bug, please contact support@buildkite.com and/or submit an issue on the github.com/buildkite/agent repo with the following information:
+Hook scope: %q
+Hook path: %q
+Hook type: %q
+Shebang line: %q
+Is binary?: %t`, hookCfg.Scope, hookCfg.Path, hookName, shebangLine, isBinary)
+	}
+}
+
+func (b *Bootstrap) runUnwrappedHook(ctx context.Context, hookName string, hookCfg HookConfig) error {
+	environ := hookCfg.Env.Copy()
+
+	environ.Set("BUILDKITE_HOOK_PHASE", hookCfg.Name)
+	environ.Set("BUILDKITE_HOOK_PATH", hookCfg.Path)
+	environ.Set("BUILDKITE_HOOK_TRIGGERED_FROM", hookCfg.Scope)
+
+	return b.shell.RunWithEnv(ctx, environ, hookCfg.Path)
+}
+
+func (b *Bootstrap) runWrappedShellScriptHook(ctx context.Context, hookName string, hookCfg HookConfig) error {
 	redactors := b.setupRedactors()
 	defer redactors.Flush()
 
-	// We need a script to wrap the hook script so that we can snaffle the changed
-	// environment variables
 	script, err := hook.NewScriptWrapper(hook.WithHookPath(hookCfg.Path))
 	if err != nil {
 		b.shell.Errorf("Error creating hook script: %v", err)
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -321,9 +321,9 @@ func (b *Bootstrap) executeHook(ctx context.Context, hookCfg HookConfig) error {
 			// Regardless: not supported right now, or potentially ever.
 			sheb, _ := shellscript.ShebangLine(hookCfg.Path) // we know this won't error because it must have a shebang to be a script
 
-			err := fmt.Sprintf(`when trying to run the hook at %q, the agent found that it was a script with a shebang that isn't for a shellscripting language - in this case, %q.
+			err := fmt.Errorf(`when trying to run the hook at %q, the agent found that it was a script with a shebang that isn't for a shellscripting language - in this case, %q.
 Hooks of this kind are unfortunately not supported on Windows, as we have no way of interpreting a shebang on Windows`, hookCfg.Path, sheb)
-			return fmt.Errorf(err)
+			return err
 		}
 
 		// It's a script, and we can rely on the OS to figure out how to run it (because we're not on windows), so run it
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -13,6 +13,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"runtime"
 	"strconv"
 	"strings"
 	"time"
@@ -307,12 +308,45 @@ func (b *Bootstrap) executeHook(ctx context.Context, hookCfg HookConfig) error {
 	}
 
 	switch hookType {
-	case hook.TypeBinary, hook.TypeScript:
+	case hook.TypeScript:
+		if runtime.GOOS == "windows" {
+			// We use shebangs to figure out how to run scripts, and Windows has no way to interpret a shebang
+			// ie, on linux, we can just point the OS to a file of some sort and say "run that", and as part of that it will try to
+			// read a shebang, and run the script using the interpreter specified. Windows can't do this, so we can't run scripts
+			// directly on Windows
+
+			// Potentially there's something that we could do with file extensions, but that's a bit of a minefield, and would
+			// probably mean that we have to have a list of blessed hook runtimes on windows only... or something.
+
+			// Regardless: not supported right now, or potentially ever.
+			sheb, _ := shellscript.ShebangLine(hookCfg.Path) // we know this won't error because it must have a shebang to be a script
+
+			err := fmt.Sprintf(`when trying to run the hook at %q, the agent found that it was a script with a shebang that isn't for a shellscripting language - in this case, %q.
+Hooks of this kind are unfortunately not supported on Windows, as we have no way of interpreting a shebang on Windows`, hookCfg.Path, sheb)
+			return fmt.Errorf(err)
+		}
+
+		// It's a script, and we can rely on the OS to figure out how to run it (because we're not on windows), so run it
+		// directly without wrapping
+		if err := b.runUnwrappedHook(ctx, hookName, hookCfg); err != nil {
+			return fmt.Errorf("running %q script hook: %w", hookName, err)
+		}
+
+		return nil
+	case hook.TypeBinary:
 		// It's a binary, so we'll just run it directly, no wrapping needed or possible
-		return b.runUnwrappedHook(ctx, hookName, hookCfg)
+		if err := b.runUnwrappedHook(ctx, hookName, hookCfg); err != nil {
+			return fmt.Errorf("running %q binary hook: %w", hookName, err)
+		}
+
+		return nil
 	case hook.TypeShell:
 		// It's definitely a shell script, wrap it so that we can snaffle the changed environment variables
-		return b.runWrappedShellScriptHook(ctx, hookName, hookCfg)
+		if err := b.runWrappedShellScriptHook(ctx, hookName, hookCfg); err != nil {
+			return fmt.Errorf("running %q shell hook: %w", hookName, err)
+		}
+
+		return nil
 	default:
 		return fmt.Errorf("unknown hook type %q for %q hook", hookType, hookName)
 	}
--- a/bootstrap/integration/hooks_integration_test.go
+++ b/bootstrap/integration/hooks_integration_test.go
@@ -492,6 +492,10 @@ func TestPreExitHooksFireAfterCancel(t *testing.T) {
 }
 
 func TestPolyglotScriptHooksCanBeRun(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("script hooks aren't supported on windows")
+	}
+
 	path, err := exec.LookPath("ruby")
 	if err != nil {
 		t.Fatal("error finding path to ruby executable: %w", err)
--- a/bootstrap/integration/hooks_integration_test.go
+++ b/bootstrap/integration/hooks_integration_test.go
@@ -12,6 +12,7 @@ import (
 	"time"
 
 	"github.com/buildkite/agent/v3/bootstrap/shell"
+	"github.com/buildkite/agent/v3/experiments"
 	"github.com/buildkite/bintest/v3"
 )
 
@@ -500,7 +501,7 @@ func TestPolyglotScriptHooksCanBeRun(t *testing.T) {
 		t.Fatal("ruby not found in $PATH. This test requires ruby to be installed on the host")
 	}
 
-	defer experiments.WithUndo(experiments.PolyglotHooks)()
+	defer experiments.EnableWithUndo(experiments.PolyglotHooks)()
 
 	tester, err := NewBootstrapTester()
 	if err != nil {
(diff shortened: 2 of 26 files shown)


bootstrap/shell/shell.go
```diff
@@ -282,6 +282,30 @@ func (s *Shell) Run(ctx context.Context, command string, arg ...string) error {
...
+
+	cmd.Env = append(cmd.Env, environ.ToSlice()...)
+
+	return s.executeCommand(ctx, cmd, s.Writer, executeFlags{
+		Stdout: true,
+		Stderr: true,
+		PTY:    s.PTY,
+	})
+}
+
```
moskyb, line 308: this function is just the `Run` method and the `RunWithPrompt` methods garglemeshed together with an env-y thing going on in the middle. earlier on, i tried to make a nice refactor so that you could compose bits and bobs of the different `Run*` methods together, but it wasn't super important to make this change. i'd like to clean it up at some point though

env/environment.go
```diff
@@ -197,6 +197,10 @@ func (e *Environment) Apply(diff Diff) {
 
 // Copy returns a copy of the env
 func (e *Environment) Copy() *Environment {
+	if e == nil {
+		return New()
+	}
+
```
moskyb, line 203: gotta catch 'em all ('em being nil pointer dereference panics)

hook/scriptwrapper.go
```diff
@@ -145,9 +145,9 @@ func NewScriptWrapper(opts ...scriptWrapperOpt) (*ScriptWrapper, error) {
 	//
 	// But if the shebang specifies something weird like Ruby 🤪
 	// the wrapper won't work. Stick to POSIX shells for now.
-	if shebang != "" && !shellscript.IsPOSIXShell(shebang) {
-		return nil, fmt.Errorf("hook starts with an unsupported shebang line %q", shebang)
-	}
+	// if shebang != "" && !shellscript.IsPOSIXShell(shebang) {
+	// 	return nil, fmt.Errorf("hook starts with an unsupported shebang line %q", shebang)
+	// }
```
moskyb, line 150: this doesn't _technically_ have to be commented out anymore - it's still an error if we try to scriptwrap a non-shell hook, but we've kind of lifted this logic out to [here](https://github.com/buildkite/agent/pull/2040/files?diff=unified&w=0#diff-70e82e7432e3056b7466bb3e4fe16874c12cfbe7bbb71a15ad9d1997217e94c9R313). Maybe it makes sense to keep this in as a defensive programming thing?

shellscript/shellscript.go
```diff
@@ -3,13 +3,85 @@ package shellscript
...
 	"os"
 	"path/filepath"
 	"strings"
 
 	"github.com/buildkite/shellwords"
+	"golang.org/x/exp/slices"
 )
 
+// Magic numbers for various types of executable files
+var (
```
moskyb, line 16: i'd like to thank the peanut gallery in the #serious-engineering channel for helping me with all of this binary nonsense

shellscript/shellscript.go
```diff
@@ -3,13 +3,85 @@ package shellscript
 
 import (
 	"bufio"
+	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
 
 	"github.com/buildkite/shellwords"
+	"golang.org/x/exp/slices"
```
DrJosh9000, line 12: Until `slices` is not in `exp`, use `bytes.Equal` instead?

bootstrap/bootstrap.go
```diff
@@ -293,11 +293,56 @@ func (b *Bootstrap) executeHook(ctx context.Context, hookCfg HookConfig) error {
...
+	if err != nil {
+		return fmt.Errorf("checking if %s hook is a binary: %w", hookName, err)
+	}
+
+	sb, err := shellscript.ShebangLine(hookCfg.Path)
+	if err != nil {
+		return fmt.Errorf("reading shebang line for %s hook: %w", hookName, err)
+	}
+
+	switch {
```
DrJosh9000, line 306: This looks fairly robust to me, but I wonder: do we want to give the user the ability (say, with a flag) to always run unwrapped?

hook/scriptwrapper.go
triarius: We can also defend ourselves with a test. I think that's more robust, so would prefer to delete the comments.

moskyb: oh totally, if this gets out of WIP i'd either de-comment it or remove these lines entirely.

when you say we can defend ourselves with a test, do you mean adding a test that asserts that if we try to wrap a non-shell-script, it complains?

triarius: Yeah, exactly. I's image, we would want to test a whole table of binaries and scripts and verify that they run. We should be able to tease out a test that fails when the scriptwrapper is run on the wrong type of binary from this.

You might have to think a bit about making an interface. I think that's a good thing to do, though.

bootstrap/bootstrap.go
```diff
@@ -293,11 +293,56 @@ func (b *Bootstrap) executeHook(ctx context.Context, hookCfg HookConfig) error {
 
 	b.shell.Headerf("Running %s hook", hookName)
 
+	isBinary, err := shellscript.IsBinaryExecutable(hookCfg.Path)
+	if err != nil {
+		return fmt.Errorf("checking if %s hook is a binary: %w", hookName, err)
+	}
+
+	sb, err := shellscript.ShebangLine(hookCfg.Path)
```
pda, line 301: I think `sb` is a bit short/cryptic for the size of the scope that we're in here (I love short vars for short scopes), and alongside the longer-and-more-descriptive `isBinary`. Perhaps it could be called `shebang` instead, which is still short but with descriptiveness similar to `isBinary`, and also helps clarify that while `isBinary` is a boolean, this one is a string?

shellscript/shellscript.go
```diff
@@ -3,13 +3,85 @@ package shellscript
...
 )
 
+// Magic numbers for various types of executable files
+var (
+	// Linux and many other Unix-like systems use ELF binaries
+	ELFMagic = []byte{0x7F, 0x45, 0x4C, 0x46}
+
+	// Windows uses MS-DOS MZ-style executables. This includes PE32 and PE32+ (which is 64-bit 🙃) binaries,
+	// which a are what modern windowses tend to use.
```
pda, line 21: It hates the windowses, precious.

shellscript/shellscript.go
```diff
@@ -3,13 +3,85 @@ package shellscript
...
+// The file at the given path is assumed to be an executable, and executability is not checked.
+func IsBinaryExecutable(path string) (bool, error) {
+	f, err := os.Open(path)
+	if err != nil {
+		return false, fmt.Errorf("open file %q: %w", path, err)
+	}
+
+	defer f.Close()
+	r := bufio.NewReader(f)
+	magic, err := r.Peek(4)
```
pda, line 63: I was surprised that `Peek` was the only method that you could ask for a specific number of bytes and get them as a slice… I assumed there's be a `Read…` method for that, which would have made slightly more sense in this context given we don't really care about not advancing the reader, but… `Peek` it is!

bootstrap/bootstrap.go
```diff
@@ -293,11 +293,56 @@ func (b *Bootstrap) executeHook(ctx context.Context, hookCfg HookConfig) error {
...
+	case sb != "" && !shellscript.IsPOSIXShell(sb):
+		// It's an interpretable-something with a shebang line, and it almost certainly isn't a shell script,
+		// so it won't be capable of setting environment variables (outside of the job api), so just run it directly
+		return b.runUnwrappedHook(ctx, hookName, hookCfg)
+	case sb == "":
+		// If there's no shebang line in the hook, and it's not a binary, so let's just assume that it's a shell script
+		// The OS will complain if we do something stupid, so we'll just give it a shot
+		return b.runWrappedShellScriptHook(ctx, hookName, hookCfg)
+	default:
+		return fmt.Errorf(`the buildkite agent wasn't able to determine the type of hook it was trying to run.
```
pda, line 322: I'm interested in (but not opposed to) the reasoning behind magic-number-detecting known executable binary formats, compared to just trusting the `+x` file mode to decide if it's executable, and checking for known-shell shebang to do the env wrapping.

Is it because we support wrapped shell hooks without a shebang?

But in this implementation, it'll even try to run binary-but-not-known-executable files as wrapped shell scripts. I wonder if that's a problem.

I wonder if this could work:

- if `isFileModeExecutable(path) == false` then `return`
- if `IsPOSIXShell(shebang)` then `runWrappedShellScriptHook`
- else `runUnwrappedHook`

That would delegate more trust/control to the `+x` file mode, and mean we don't need to maintain a (beautiful 💚) definition of known executable magic numbers.

There's quite likely some factors I'm not considering.

pda: Hmm… I guess a bit difference there is that we'd no longer try running non-shell-shebanged executables as _wrapped_ hooks, which would perhaps break lots of things, if there's non-shebanged shell hooks out in the wild. And I'm sure there it.

So perhaps what I'm suggesting is

- case `IsPOSIXShell(shebang)`: `runWrappedShellScriptHook`
- case `hasShebang || isBinary`: `runUnwrappedHook`
- default (case `isValidUTF8`): `runWrappedShellScriptHook`

Note that in this version, `isBinary` means “isn't valid UTF-8 (and thus also isn't ASCII)” which implies “more likely a compiled program than a shell script”, and there's no magic number detection.

Just a thought 🤷‍♂️

triarius: I suspect checking for the `x` bit won't work on Windows. But that does not mean we should not respect it on unicies.

hook/scriptwrapper.go
moskyb: yep, that sounds like a good idea. when we get to properly testing this, i'll do that.

shellscript/shellscript.go
moskyb: you can also do:
```Go
b := make([]byte, 4)
err := f.Read(b)
```
which will read at most 4 bytes into b, as will
```Go
bytes, err := ioutil.ReadAll(io.LimitReader(f, 4))
```

perhaps either of those is clearer? i'm no 100% sure

shellscript/shellscript.go
```diff
@@ -3,13 +3,85 @@ package shellscript
...
+
+	BinaryMagicks = [][]byte{
+		ELFMagic,
+		MZMagic,
+		MachOFatMagic,
+		MachOThin32Magic,
+		MachOThin64Magic,
+		MachOThin32ReverseMagic,
+		MachOThin64ReverseMagic,
+	}
```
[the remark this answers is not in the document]
moskyb: yep, good call

bootstrap/bootstrap.go
moskyb: > ... we'd no longer try running non-shell-shebanged executables as wrapped hooks, which would perhaps break lots of things, if there's non-shebanged shell hooks out in the wild. And I'm sure there it.

yep, this was a big part of the motivation here. in fact, on windows, the only way to identify a batch/powershell script is the file extension - they don't use shebangs over in windowsland, so maintaining the default of "it's not shebanged, i don't know what it is" being "assume it's a shellscript" is sensible in this case IMO.

---

re: alternate ways of detecting executable binaries:

i agree that taking care of a big list of executable magic numbers is a pain in the butt, but i can't think of a better cross-platform way to check if something is an executable - checking if the file encoding is funky and there's an `x` permission bit set is fine on unices, but as @triarius mentions, it probably won't work on windows.

additionally, checking that a file is invalid UTF-8 is necessary, but not sufficient in proving that it's a binary. to bring up windows again, the default encoding for Powershell scripts is Windows-1252, which would fail this check and cause the agent to try to run the powershell scripts unwrapped, which would be a problem.

i guess my point here is that imo, having a very narrow definition of what a binary hook looks like (it must have the right magic number) is a benefit to us; it allows us to define a strict API on what will be executed as a binary hook. if it isn't a valid executable for your platform, we'll try to run it in a different way and it'll break. 

if we've missed binary types in our big list, then we can add their signatures, but were we to use a less exact way of identifying binaries, it would open us up to all sorts of horrible bugs, encoding and otherwise

shellscript/shellscript.go
pda: Ooh fancy LimitReader composition 👌🏼

I like `r.Peek(4)` better though, because it's less words and still quite clear 👍🏼

bootstrap/bootstrap.go
pda: Makes sense! 👍🏼

shellscript/shellscript.go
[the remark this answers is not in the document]
pda: Does that make it more complex / fragile / harder to understand?
I don't think there'd be a big benefit to skipping those checks?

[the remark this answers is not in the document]
moskyb: yeah, i was thinking this after i commented tbh. it's a lot more code/test complexity to save ~5 loop iterations, i might leave it as is tbh.

bootstrap/bootstrap.go
DrJosh9000: On reflection, it's probably likely to just break people for minimal gain.

lox:
Love this!

> note that neither binary hooks nor non-shell script hooks have environment variable changes captured as bash script hooks do; they must use the Job API to modify the environment of the job.

Yeah, this is my only concern. I wonder if this will be an unexpected sharp edge. I guess that folks assume that bash scripts are simply sourced (vs the magic of wrapped hooks) and they won't expect that behaviour for non-shell scripts. 

I assume the future is all Job API and no more magical environment capture?

moskyb:
@lox 

> I assume the future is all Job API and no more magical environment capture?

in an "as the limit tends to infinity" sense, i think so. i doubt that we'd ever fully remove the environment functionality anytime in the forseeable future; it's just too useful and easy to set up.

i suspect that if we merged this, we'd try to make non-bash hooks the golden path, and try to build some nicer tooling around them, and encourage it that way, but keep env capture around as a simple mode? i'm not 100% sure, i may or may or may not have thought that far ahead 😅

bootstrap/integration/hooks_integration_test.go
```diff
@@ -488,3 +490,85 @@ func TestPreExitHooksFireAfterCancel(t *testing.T) {
...
+	})
+
+	tester.RunAndCheck(t)
+
+	if !strings.Contains(tester.Output, "hi there from golang 🌊") {
+		t.Fatalf("tester.Output %s does not contain expected output: %q", tester.Output, "hi there from golang 🌊")
+	}
+}
+
+func which(prog string) bool {
```
DrJosh9000, line 562: Could this be replaced with `shell.LookPath` (does the ~same things)?

hook/type_test.go
```diff
@@ -0,0 +1,99 @@
...
+			}
+
+			if hookType != c.expectedHookType {
+				t.Fatalf("Expected hook type %q, got %q", c.expectedHookType, hookType)
+			}
+		})
+	}
+}
+
+func hookFixture(wd string, fixturePath []string) []string {
```
DrJosh9000, line 97: If this was variadic...

```suggestion
func hookFixture(wd string, fixturePath ...string) []string {
```

...then the implementation would be the same, but the call sites could avoid the `[]string{ ... }` wrapper...

```
hookFixture(root, "some", "path", "that", "doesnt", "exist")
```

hook/type_test.go
```diff
@@ -0,0 +1,99 @@
...
+	if !errors.Is(err, os.ErrNotExist) {
+		t.Errorf("expected os.ErrNotExist, got %v", err)
+		return true
+	}
+
+	return false
+}
+
+func TestHookType(t *testing.T) {
+	wd, _ := os.Getwd()
```
DrJosh9000, line 38: Add a comment that this is because test/fixtures/hook is at the root of the repo, but the test working directory is not?

DrJosh9000 (commented):
Looking pretty good 🥞

hook/type_test.go
moskyb: oh man that's way nicer, i'll do that

experiments/experiments.go
```diff
@@ -34,6 +36,11 @@ var (
 	experiments = make(map[string]bool, len(Available))
 )
 
+func WithUndo(key string) func() {
+	Enable(key)
+	return func() { Disable(key) }
+}
+
```
triarius, line 43: I also think it's a bit evil to enable the experiment as a side effect and not mention it in the method name. Then when you're reading the code 
```go
defer experiments.WithUndo(experiments.GitMirrors)()
```
It reads like you're deferring the experiment ("with undo", whatever that is).

Maybe this is more clear?
```suggestion
func EnableReturningUndo(key string) func() {
	Enable(k)
	return func() { Disable(key) }
}
```

Then I would use it like this:
```go
undo := experiments.EnableReturningUndo(experiments.GitMirrors)
defer undo()
```

env/environment.go
triarius: I sort of expect a copy of `nil` to be `nil`, but it's probably more pragmatic for it to be `New()`.

bootstrap/integration/test-binary-hook/main.go
```diff
@@ -0,0 +1,31 @@
...
+	c, err := jobapi.NewDefaultClient(context.TODO())
+	if err != nil {
+		log.Fatalf("error: %v", fmt.Errorf("creating job api client: %w", err))
+	}
+
+	_, err = c.EnvUpdate(context.TODO(), &jobapi.EnvUpdateRequest{
+		Env: map[string]string{
+			"OCEAN":    "Pacífico",
+			"MOUNTAIN": "chimborazo",
+		},
```
triarius, line 23: Good thinking to test the jobapi as well 🧠

bootstrap/shell/shell.go
triarius: Yes please

scripts/generate-test-binaries.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+indir=$1
+outdir=$2
+binary_name=$3
+
+for os in "darwin" "linux" "windows"; do
+  for arch in "amd64" "arm64"; do
+    echo "Building test binary for $os/$arch"
+    GOOS=$os GOARCH=$arch go build -o "$outdir/$binary_name-$os-$arch" "$indir"
```
triarius, line 10: I think you should try to make these binaries reproducible so that if I run this script on my machine, it will generate the exact same files. You will at least `-trimpath` and `CGO_ENABLED=0`.

Alternatively, don't check in the binaries at all? Maybe we can make a `testing.Run` to generate them for the tests that need them. 

Anyway, this is all non-blocking, but I might try to fix it myself if I encounter problems.

hook/type.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+const (
+	TypeShell   = "shell"
+	TypeBinary  = "binary"
+	TypeScript  = "script" // Something with a shebang that isn't a shell script, but is an interpretable something-or-other
+	TypeUnknown = "unknown"
+)
+
+func Type(path string) (string, error) {
```
triarius, line 17: It might be better to call this `FromPath`. Writing code like `type = type.Type` triggers me back to when I was forced to learn Java in Uni.

hook/scriptwrapper.go
triarius: Have we gotten to properly testing this?

triarius (commented):
Thank you for your service @moskyb 🫡

I think my only blocking comment is for the scriptwrapper tests. It may no longer be necessary to write them, though because of some combination of other tests. If so, I think you should delete the commented code. If not, you should still delete it. We are all pretty good at git archaeology.

hook/scriptwrapper.go
moskyb: very good call! when i said:

> oh totally, if this gets out of WIP i'd either de-comment it or remove these lines entirely.

apparently i was lying :P

i'll bang out some tests for this now

scripts/generate-test-binaries.sh
moskyb: yeah, i think you're right actually, we should probably generate these on the fly like we do with the integration test. it might be a little slow, but tests are pretty fast as is.

hook/type.go
moskyb: i mean in this case it would be `type := hook.Type`, but i do see where you're coming from

triarius: Whoops, yeah, `hook.Type` is fine.

scripts/generate-test-binaries.sh
moskyb: this is done :)

triarius: Sorry I bikeshedded you into this,

```


C:\buildkite-agent\builds\agent-runners-windows-amd64-i-050772c7c334529b8-1\buildkite\agent\test\fixtures\hook\test-binary-windows-arm64
--



    type_test.go:98: Failed to remove test-binary-hook: remove 
C:\buildkite-agent\builds\agent-runners-windows-amd64-i-050772c7c334529b8-1\buildkite\agent\test\fixtures\hook\test-binary-linux-amd64:
 The process cannot access the file because it is being used by another 
process.
```

Maybe this just skip the removal on Windows, it should be cleaned out at the next checkout anyway?

It's a bit concerning that it says "cannot access". If that prevents executing the hook concurrently, we may have a problem. But I suspect it's just because we are trying to delete it and there a process hanging onto the binary.

moskyb: alternatively, i think i can just yeet them into the tempdir and let the os sort it out

triarius (approved):

bootstrap/bootstrap.go
```diff
@@ -297,11 +298,74 @@ func (b *Bootstrap) executeHook(ctx context.Context, hookCfg HookConfig) error {
...
+
+			// Potentially there's something that we could do with file extensions, but that's a bit of a minefield, and would
+			// probably mean that we have to have a list of blessed hook runtimes on windows only... or something.
+
+			// Regardless: not supported right now, or potentially ever.
+			sheb, _ := shellscript.ShebangLine(hookCfg.Path) // we know this won't error because it must have a shebang to be a script
+
+			err := fmt.Sprintf(`when trying to run the hook at %q, the agent found that it was a script with a shebang that isn't for a shellscripting language - in this case, %q.
+Hooks of this kind are unfortunately not supported on Windows, as we have no way of interpreting a shebang on Windows`, hookCfg.Path, sheb)
+			return fmt.Errorf(err)
```
DrJosh9000, line 326: Suppose a % sneaks into the string stored in `err`...

```suggestion
			err := fmt.Errorf(`when trying to run the hook at %q, the agent found that it was a script with a shebang that isn't for a shellscripting language - in this case, %q.
Hooks of this kind are unfortunately not supported on Windows, as we have no way of interpreting a shebang on Windows`, hookCfg.Path, sheb)
			return err
```

bootstrap/integration/hooks_integration_test.go
```diff
@@ -488,3 +490,85 @@ func TestPreExitHooksFireAfterCancel(t *testing.T) {
...
 }
+
+func TestPolyglotScriptHooksCanBeRun(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("script hooks aren't supported on windows")
+	}
+
+	path, err := exec.LookPath("ruby")
+	if err != nil {
+		t.Fatal("error finding path to ruby executable: %w", err)
```
DrJosh9000, line 501: IIRC `%w` is only for `fmt.Errorf`, and also `t.Fatal` doesn't do formatting verbs at all.

```suggestion
		t.Fatalf("error finding path to ruby executable: %v", err)
```

This might be caught with `go vet`?

bootstrap/integration/hooks_integration_test.go
```diff
@@ -488,3 +490,85 @@ func TestPreExitHooksFireAfterCancel(t *testing.T) {
...
+	defer tester.Close()
+
+	filename := "environment"
+	script := []string{
+		"#!/usr/bin/env ruby",
+		`puts "ohai, it's ruby!"`,
+	}
+
+	if err := os.WriteFile(filepath.Join(tester.HooksDir, filename), []byte(strings.Join(script, "\n")), 0755); err != nil {
+		t.Fatalf("os.WriteFile(%q, script, 0755 = %v", filename, err)
```
DrJosh9000, line 523:
```suggestion
		t.Fatalf("os.WriteFile(%q, script, 0755) = %v", filename, err)
```

bootstrap/integration/hooks_integration_test.go
```diff
@@ -488,3 +490,85 @@ func TestPreExitHooksFireAfterCancel(t *testing.T) {
...
+	}
+
+	tester.RunAndCheck(t)
+
+	if !strings.Contains(tester.Output, "ohai, it's ruby!") {
+		t.Fatalf("tester.Output %s does not contain expected output: %q", tester.Output, "ohai, it's ruby!")
```
DrJosh9000, line 529: Format this with %q as well? (tester.Output being an empty string seems like a likely test failure mode)

bootstrap/integration/hooks_integration_test.go
```diff
@@ -488,3 +490,85 @@ func TestPreExitHooksFireAfterCancel(t *testing.T) {
...
+			c.Exit(1)
+		} else {
+			c.Exit(0)
+		}
+	})
+
+	tester.RunAndCheck(t)
+
+	if !strings.Contains(tester.Output, "hi there from golang 🌊") {
+		t.Fatalf("tester.Output %s does not contain expected output: %q", tester.Output, "hi there from golang 🌊")
```
DrJosh9000, line 572: As above

DrJosh9000 (changes requested):

bootstrap/integration/hooks_integration_test.go
moskyb: when i changed `%w` for `%v` it came up with the lil squiggly yellow line, i'm guessing that it doesn't see the %w verb as relevant to calls to `t.Fatal`? fair enough i suppose.

fixed now, btw

DrJosh9000 (approved):
Awesome! Ship it 🚢

PR #2042, opened 2023-04-04
Local Agent API, leadership, and locking

DrJosh9000:
This PR adds an experiment that adds _another_ local socket! The new API exposes primitives that can be used to implement mutexes for concurrent processes, in a cross-platform way. This can be used to solve problems to do with multiple agents trying to run jobs concurrently, that need to handle shared resources, such as preventing concurrent `.ssh/known-hosts` edits, or multiple simultaneous `docker prune` calls. (Though the latter is probably better done as a cron job.)

Unlike the Job API, there is intended to be only one leader (and therefore only one leader _socket_) per host. This handles multiple agents, whether they are parallelised with `--spawn` or though some other mechanism such as systemd unit templates. Each agent runs a socket, and the leader is the one with the symlink pointing at its socket.

Unlike consensus-protocol-based systems (see Paxos, Raft), no attempt is made to synchronise state between multiple agents. If the leader dies, it takes the lock state with it. If multiple agents were running, another agent will realise the leader is dead and assume leadership, but will start from a blank state. This is hardly ideal, but all of this is local to a single host. We can probably incorporate Raft via a library in a future PR.

Finally, some new CLI subcommands exercise the new API to implement mutex (`lock acquire`, `lock release`) and [do-once](https://pkg.go.dev/sync#Once) (`lock do`, `lock done`).


clicommand/agent_start.go
```diff
@@ -761,6 +762,17 @@ var AgentStartCommand = cli.Command{
...
+			// Try to be the leader - no worries if not.
+			leaderSvr, err := leaderapi.NewServer(leaderapi.LeaderSocketPath)
+			if err != nil {
+				l.Warn("Couldn't become leader: %v", err)
+			}
+			if leaderSvr != nil {
+				defer leaderSvr.Shutdown(ctx)
+			}
+		}
+
```
triarius, line 775: What happens when a leader dies? It looks like nothing will be listening on the socket and the clients will all fail. That's not too bad, but I think some kind of promotion procedure would be useful.

triarius (commented):
Just some general comments for now. Will save my nitpicking till it's no longer draft.

clicommand/agent_start.go
triarius: Also a good idea to print an info log when this does become the leader.

DrJosh9000: Done

clicommand/lock_acquire.go
```diff
@@ -0,0 +1,73 @@
...
+
+`
+
+type LockAcquireConfig struct{}
+
+var LockAcquireCommand = cli.Command{
+	Name:        "acquire",
+	Usage:       "Acquires a lock from the agent leader",
+	Description: lockAcquireHelpDescription,
+	Action:      lockAcquireAction,
```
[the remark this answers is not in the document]
DrJosh9000: Done

clicommand/agent_start.go
```diff
@@ -22,6 +22,7 @@ import (
 	"github.com/buildkite/agent/v3/cliconfig"
 	"github.com/buildkite/agent/v3/experiments"
 	"github.com/buildkite/agent/v3/hook"
+	"github.com/buildkite/agent/v3/leaderapi"
```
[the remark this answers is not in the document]
DrJosh9000: Yes!

clicommand/agent_start.go
```diff
@@ -1110,3 +1116,67 @@ func defaultSocketsPath() string {
...
+		defer canc()
+
+		cli, err := agentapi.NewClient(ctx, leaderPath)
+		if err != nil {
+			return err
+		}
+		return cli.Ping(ctx)
+	}
+
+	for range time.Tick(100 * time.Millisecond) {
```
moskyb, line 1174: TIL the assign-less `for range` expression! totally makes sense that you can do it though

internal/agentapi/client.go
```diff
@@ -0,0 +1,69 @@
...
+// Ping pings the server. It returns a non-nil error if the ping fails, or the
+// response timestamp is more than 100 milliseconds different to time.Now.
+func (c *Client) Ping(ctx context.Context) error {
+	var resp PingResponse
+	if err := c.sc.Do(ctx, "GET", "http://agent/api/leader/v0/ping", nil, &resp); err != nil {
+		return err
+	}
+	if time.Since(resp.Now) > 100*time.Millisecond {
+		return fmt.Errorf("ping timestamp %v too old", resp.Now)
+	}
```
moskyb, line 39: i think this timestamp checking is a good idea in general, but is there a specific failure case that it'll catch? my understanding is that if the ping takes >100ms, it'll cause a leadership coup, is that right?

moskyb (approved):
for a PR with a lot of Ls in the title, i call this one a big W

internal/socket/client.go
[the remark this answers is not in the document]
DrJosh9000: That's what I'm thinking. Put common socket stuff in internal/socket, make the new stuff use it in this PR, then change Job API in a follow-up.

internal/agentapi/client.go
DrJosh9000: Not really, no. It feels like if the agent is up and running, it ought to be able to respond to a local ping that quickly. Possibly it could be slow if the system as a whole is so overloaded, but then churning the leader won't help.

[Side note: I like 100ms because it's simultaneously "blink of an eye" for a human and "a long time" for a computer.]

internal/socket/client.go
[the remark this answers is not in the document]
moskyb: sounds fab 👍

internal/agentapi/client.go
moskyb: sweet 👍 i think it's still worthwhile to keep this around, i was just interested to know if there was something i'm missing

DrJosh9000:
@triarius I clicked the "Re-request review" button in case you wanted to pick some nits, as you mentioned 😀

triarius (commented):
This is great! I haven't done a very thorough review, but I don't think there are any blockers.

clicommand/agent_start.go
```diff
@@ -1110,3 +1116,71 @@ func defaultSocketsPath() string {
...
+		}
+		if d == path {
+			// It's me! Don't bother pinging.
+			return nil
+		}
+
+		ctx, canc := context.WithTimeout(ctx, 100*time.Millisecond)
+		defer canc()
+
+		cli, err := agentapi.NewClient(ctx, leaderPath)
```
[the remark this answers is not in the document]
DrJosh9000: Un-shadowed

internal/socket/server_test.go
```diff
@@ -0,0 +1,159 @@
...
+		{
+			method:     http.MethodPatch,
+			url:        "http://yn/yes",
+			wantStatus: http.StatusMethodNotAllowed,
+			want:       "method not allowed\n",
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.method+" "+test.url, func(t *testing.T) {
```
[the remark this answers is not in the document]
DrJosh9000: Done - conveniently this reminded me of `t.Cleanup`.

PR #2043, opened 2023-04-06
Add *_PRIVATE_KEY to auto-redacted vars

moskyb:
Private keys are private! We should avoid leaking them into the logs if possible. Potentially we should just redact anything that ends with `*_KEY`? We can get to that later


pda (approved):
Seems very reasonable (and I'm surprised it wasn't there previously).

matthewborden (approved):

PR #2050, opened 2023-04-17
Update roko v1.0.3 -> v1.1.0

moskyb:
This gives us access to the new `ExponentialSubsecond` strategy, as well as the ability to manually set the next interval.


triarius (approved):
Yay!

PR #2055, opened 2023-04-19
Run go-licenses for multiple GOOS values.

DrJosh9000:
#2051 demonstrates that go mod updates can cause license changes. No biggie, run the generate-acknowledgements scripts and push. 

Except that's not enough - the output depends on GOOS. (And probably GOARCH too.) So when the acknowledgements are updated on a Mac, any Linux or Windows specific packages licenses could be lost. 

There's also an issue where it thinks the license for `golang.org/x/sys` is at `golang.org/x/sys/unix/LICENSE` when `golang.org/x/sys/LICENSE` is right there. Sometimes. (Particularly, it prefers one on our dev machines and the other in the build container.)

Running go-licenses for all the agent GOOSes (geese?), and merging the results, picks up both paths, as well as the missing license for go-winio, and should be more robust.


moskyb:
DenshaDeD_ch01p16-17

scripts/generate-acknowledgements.sh
```diff
@@ -21,12 +21,22 @@ fi
...
 export TEMPDIR="$(mktemp -d /tmp/generate-acknowledgements.XXXXXX)" || exit 1
-rmdir "${TEMPDIR}"
-"${GO_LICENSES}" save . --save_path="${TEMPDIR}"
+export COMBINEDIR="${TEMPDIR}/combined"
+mkdir -p "${COMBINEDIR}"
+
+# go-licenses output can vary by GOOS.
+# https://github.com/google/go-licenses/issues/187
+# Run it three times for the main OSes, and combine the results.
+for goos in darwin linux windows ; do
```
triarius, line 30: I think we have to put all the GOOSes we publish binaries for in the [release](https://github.com/buildkite/agent/releases) (at least).

By my estimation, that's:
```suggestion
for goos in darwin linux windows freebsd dragonfly netbsd openbsd ; do
```

Keeping this in sync with the buildkite yaml for the release pipeline is also possible with some `yq` like tool.

triarius (commented):
Nice finds!

scripts/generate-acknowledgements.sh
DrJosh9000: Done (though without any automated syncing).

triarius (approved):
LGTM 💖 The lack of syncing may bite us one day. Maybe https://github.com/google/go-licenses/issues/187 will be resolved with a PR by then.

PR #2056, opened 2023-04-20
BUILDKITE_GIT_CLONE_MIRROR_FLAGS not working

ppatwf:
Ran into this when trying to get clone flags working in mirror mode. I tried using the environment variable listed in the docs (https://buildkite.com/docs/agent/v3/configuration), but it wasn't working.

I got the `BUILDKITE_GIT_CLONE_FLAGS` working properly in the same method and when I ran a regular clean it worked just fine, but the `BUILDKITE_GIT_CLONE_MIRROR_FLAGS` never worked. Dug through the agent code and saw this `env` was missed during a refactor so I figured that was the issue.

### DIFF
--- a/bootstrap/config.go
+++ b/bootstrap/config.go
@@ -81,7 +81,7 @@ type Config struct {
 	GitFetchFlags string `env:"BUILDKITE_GIT_FETCH_FLAGS"`
 
 	// Flags to pass to "git clone" command for mirroring
-	GitCloneMirrorFlags string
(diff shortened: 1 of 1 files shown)


triarius (changes requested):
Thanks for your contribution @ppatwf. I think we will accept this change, but we need a bit more context in the PR description before we can proceed.

ppatwf:
@triarius sorry! first time contributing back to OSS in a while and was heads down so I didn't pay enough attention to the PR. There's an updated description. Let me know if I can add any more context

triarius (approved):
Awesome, thanks @ppatwf

PR #2065, opened 2023-05-03
Pack the build version into the right package

moskyb:
The version.go file was moved in a previous PR, but the linker flag packing the build version in at build-time was still pointing at the old path


DrJosh9000 (approved):
🙏

PR #2067, opened 2023-05-03
Give a nicer error when empty strings are used as metadata values

### LINKED ISSUE #2066: Unable to set empty or blank values in metadata.
The `meta-data set` command is documented to

> Set arbitrary data on a build using a basic key/value store.

A limit of 100 kilobytes and a recommendation of less than 1kb for value size are documented elsewhere.

However, I have been unable to set values of `""`, `" "`, `"\t"`, or `"\n\n\n"` using this command. Attempts to run commands like

```
buildkite-agent meta-data set cluster-config-names ' '
```

result in errors like

```
2023-05-03 08:52:19 WARN   POST https://agent.buildkite.com/v3/jobs/0187e0ce-81e6-4c53-af13-2b5eecff6ed2/data/set: 400 Value is missing (Attempt 1/10 Retrying in 5s)
```

no errors are produced when using a similar command to set a non-blank value.

Ideally, I'd like to be able to set arbitrary (but small) values in metadata. Failing that, I'd like to know what the limits are on the contents of metadata values.

moskyb:
closes #2066 

we already disallow blank metadata keys on the serverside, this will just short-circuit that error and give a nice error.

it also adds a note to the docs saying that a value for the metadata must be set.

### DIFF
--- a/clicommand/meta_data_set.go
+++ b/clicommand/meta_data_set.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/buildkite/agent/v3/api"
@@ -23,6 +24,9 @@ Description:
 
    You can supply the value as an argument to the command, or pipe in a file or
    script output.
+
+   The value must be a non-empty string, and strings containing only whitespace
+   characters are not allowed.
 
 Example:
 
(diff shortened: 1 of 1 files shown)


triarius (approved):
🎉

DrJosh9000 (approved):
Nice!

PR #2068, opened 2023-05-04
Gracefully cancel checkout by default

moskyb:
This feature/fix was originally committed as an experiment, but we want to turn it on by default - it seems fairly safe


triarius (approved):
LGTM

PR #2069, opened 2023-05-04
ArtifactBatchCreator & ArtifactUploader: use roko.ExponentialSubsecond

pda:
For `ArtifactBatchCreator` & `ArtifactUploader` retries, use the new `roko.ExponentialSubsecond` from https://github.com/buildkite/roko/pull/8 as intended in https://github.com/buildkite/agent/pull/2028 (at that point `ExponentialSubsecond` wasn't merged into roko yet).

This means more retries happen sooner, with a gentler growth curve, and the retry limit is reached sooner; the total retry wait across 10 attempts is ~33 seconds. Assuming the requests are timing out after 10 seconds, that's 100+33 = 133 seconds (just over 2 minutes). I think that's a good amount of time to try for.

[Here's](https://www.wolframalpha.com/input?i=y%3D500%5E%28x%2F16+%2B+1%29+for+x+from+0+to+8) the **new curve** of `ExponentialSubsecond` for 10 attempts (9 “between” delays) with a 500ms start, growing to ~10 seconds:
(Y axis is **milliseconds**)

image

[Here's](https://www.wolframalpha.com/input?i=y%3D2%5Ex+for+x+from+0+to+6) the **previous curve** of `roko.Exponential(2*time.Second, 0)` for 8 attempts, growing to ~60 seconds:
(Y axis is in **seconds**)

image

Here's a failure retrying to exhaustion with the new strategy:

image

And here's one succeeding after a few timeouts:

image

Related:
- https://github.com/buildkite/roko/pull/8
- https://github.com/buildkite/agent/pull/2028

### DIFF
--- a/agent/artifact_batch_creator.go
+++ b/agent/artifact_batch_creator.go
@@ -76,11 +76,8 @@ func (a *ArtifactBatchCreator) Create(ctx context.Context) ([]*api.Artifact, err
 
 		// Retry the batch upload a couple of times
 		err = roko.NewRetrier(
-			// TODO: e.g. roko.ExponentialSubsecond(500*time.Millisecond) WithMaxAttempts(10)
-			// see: https://github.com/buildkite/roko/pull/8
-			// Meanwhile, 8 roko.Exponential(2sec) attempts is 1,2,4,8,16,32,64 seconds delay (~2 mins)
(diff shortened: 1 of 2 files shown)


DrJosh9000 (approved):
🪴

moskyb (approved):
📈

PR #2072, opened 2023-05-05
Add a flag to `artifact upload` to allow skipping symlinks when uploading files

triarius:
TL;DR: There is a new flag `--upload-skip-symlinks` that prevents files that are symlinks from being uploaded. The flag `--follow-symlinks` has been deprecated and renamed to `--glob-resolve-follow-symlinks`.

# Description

Some users may misinterpret specifying `false` to the `--follow-symlinks` flag to mean that the agent will not upload symlinks.

However, this is not the case. Until this PR, the agent will always upload all files that match the glob provided to the `artifact upload` subcommand. The `--follow-symlinks` flag only affected whether symlinked directories would be followed when resolving the glob.

However, I think it is a reasonable expectation that `--follow-symlinks=false` WOULD prevent symlinked files from being uploaded. Indeed, on the `aws s3 sync` command, [that is the case](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/s3/sync.html#options). 

However, I think we are constrained from changing the behaviour of the existing flag. Some customers may have been uploading artifacts unaware that they are actually symlinks. As the flag is confusing, we should deprecate it. However, it is not straightforward if we should replace it with one flag that both does not follow symlinks when resolving the glob, and skips uploading files that are symlinks, or create a flag for each of these.

I decided on the former so that there will be an exact replacement for the deprecated `--follow-symlinks` flag, `--glob-resolve-follow-symlinks`. The new feature of not uploading symlinks is enabled by the `--upload-skip-symlink` flag.

This gives the follow table
glob-resolve-follow-symlinks|upload-skip-symlinks| outcome
------|-----|--------
 F | F | glob does not enter symlinked dirs and symlinked files are not uploaded
 F | T | glob does not enter symlinked dirs and symlinked files are uploaded
 T | F | glob does enter symlinked dirs and symlinked files are not uploaded
 T | T | glob does enter symlinked dirs and symlinked files are uploaded

As all four possibilities are meaningfully different, I think it's justified to have 2 flags even if the backward compatibility need not be maintained.

# Uploading just the symlink
The binary nature of the `--upload-skip-symlinks` flag precludes a third behaviour where instead of uploading the file behind the symlink, the agent could endeavour to preserve the symlink.
 
While this is an interesting use case, S3 does not support symlinks. An implementation of the feature would have to choose a representation of the symlink in the object storage, and on downloading such a representation, reconstruct the symlink on the target file system. Also, the upload and download could occur on different file systems and by agents on different operating systems, each with differing levels of implementations of symlinks, exposing us to many edge cases.  I think that if users want to preserve what's on the file system to that extent, they are better served by creating a tarball.

## Implementation Note:
The implementation uses a call to `lstat` to determine if the file is a link. There is a previous call to `stat` to determine if the file is a directory. I thought about ways to combine these, but it seems both calls need to be made to determine if a path is a symlink to a directory.

### DIFF
--- a/agent/artifact_uploader.go
+++ b/agent/artifact_uploader.go
@@ -46,10 +46,10 @@ type ArtifactUploaderConfig struct {
 	DebugHTTP bool
 
 	// Whether to follow symbolic links when resolving globs
-	FollowSymlinks bool
+	GlobResolveFollowSymlinks bool
 
 	// Whether to not upload symlinks
-	NoUploadSymlinks bool
+	UploadSkipSymlinks bool
 }
 
 type ArtifactUploader struct {
@@ -127,7 +127,7 @@ func (a *ArtifactUploader) Collect() (artifacts []*api.Artifact, err error) {
 		// Resolve the globs (with * and ** in them), if it's a non-globbed path and doesn't exists
 		// then we will get the ErrNotExist that is handled below
 		globfunc := zglob.Glob
-		if a.conf.FollowSymlinks {
+		if a.conf.GlobResolveFollowSymlinks {
 			// Follow symbolic links for files & directories while expanding globs
 			globfunc = zglob.GlobFollowSymlinks
 		}
@@ -159,7 +159,7 @@ func (a *ArtifactUploader) Collect() (artifacts []*api.Artifact, err error) {
 				continue
 			}
 
-			if a.conf.NoUploadSymlinks && isSymlink(absolutePath) {
+			if a.conf.UploadSkipSymlinks && isSymlink(absolutePath) {
 				a.logger.Debug("Skipping symlink %s", file)
 				continue
 			}
--- a/agent/artifact_uploader.go
+++ b/agent/artifact_uploader.go
@@ -47,6 +47,9 @@ type ArtifactUploaderConfig struct {
 
 	// Whether to follow symbolic links when resolving globs
 	FollowSymlinks bool
+
+	// Whether to not upload symlinks
+	NoUploadSymlinks bool
 }
 
 type ArtifactUploader struct {
@@ -86,6 +89,14 @@ func (a *ArtifactUploader) Upload(ctx context.Context) error {
 	}
 
 	return nil
+}
+
+func isSymlink(path string) bool {
+	fi, err := os.Lstat(path)
+	if err != nil {
+		return false
+	}
+	return fi.Mode()&os.ModeSymlink != 0
 }
 
 func isDir(path string) bool {
@@ -145,6 +156,11 @@ func (a *ArtifactUploader) Collect() (artifacts []*api.Artifact, err error) {
 			// Ignore directories, we only want files
 			if isDir(absolutePath) {
 				a.logger.Debug("Skipping directory %s", file)
+				continue
+			}
+
+			if a.conf.NoUploadSymlinks && isSymlink(absolutePath) {
+				a.logger.Debug("Skipping symlink %s", file)
 				continue
 			}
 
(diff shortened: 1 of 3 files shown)


clicommand/artifact_upload.go
```diff
@@ -147,12 +153,13 @@ var ArtifactUploadCommand = cli.Command{
...
-			ContentType:    cfg.ContentType,
-			DebugHTTP:      cfg.DebugHTTP,
-			FollowSymlinks: cfg.FollowSymlinks,
+			JobID:            cfg.Job,
+			Paths:            cfg.UploadPaths,
+			Destination:      cfg.Destination,
+			ContentType:      cfg.ContentType,
+			DebugHTTP:        cfg.DebugHTTP,
+			FollowSymlinks:   cfg.FollowSymlinks,
+			NoUploadSymlinks: !cfg.UploadSymlinks,
```
triarius, line 162: We need to invert the bool as the existing behaviour is equivalent to setting `NoUploadSymlinks` to false, and when creating a new `agent.ArtifactUploaderConfig` struct, the field will be `false` by default.

agent/artifact_uploader.go
```diff
@@ -47,6 +47,9 @@ type ArtifactUploaderConfig struct {
 
 	// Whether to follow symbolic links when resolving globs
 	FollowSymlinks bool
+
+	// Whether to not upload symlinks
+	NoUploadSymlinks bool
```
triarius, line 52: We need to invert the field as existing behaviour is equivalent to setting `NoUploadSymlinks` to false, and when creating a new `ArtifactUploaderConfig` struct, the field will be `false` by default.

moskyb (changes requested):
everything here LGTM, but maybe while we're here it'd be worth adding a flag that does the same thing as what `--follow-symlinks` does, but with a better name, and using the existing `deprecated-and-renamed-to` gear in our CLI lib. maybe calling it `--glob-resolve-symlinks` or similar?

ldionne:
Thanks for fixing this! IMO the documentation could be a bit more clear on the fact that symlinks are skipped entirely, i.e. it’s not just that the symlinked-to file won’t be uploaded, but the symlink itself won’t be uploaded. Maybe something like: « Note that symlinks themselves are never uploaded because some object storages can’t represent them. —upload-skip-symlinks only controls whether the symlink targets are uploaded under the symlink’s name. If you want to upload a full directory hierarchy and preserve its structure, a tarball can be used instead. »

This might be a bit long winded, but you get the idea.

clicommand/artifact_upload.go
```diff
@@ -103,6 +112,21 @@ var ArtifactUploadCommand = cli.Command{
...
 		},
+		cli.BoolFlag{
+			Name:   "glob-resolve-follow-symlinks",
+			Usage:  "Follow symbolic links to directories while resolving globs. Note: this will not prevent symlinks to files from being uploaded. Use --upload-skip-symlinks to do that",
+			EnvVar: "BUILDKITE_AGENT_ARTIFACT_GLOB_RESOLVE_FOLLOW_SYMLINKS",
+		},
+		cli.BoolFlag{
+			Name:   "upload-skip-symlinks",
+			Usage:  "After the glob has been resolved to a list of files to upload, skip uploading those that are symlinks to files",
+			EnvVar: "BUILDKITE_ARTIFACT_UPLOAD_FOLLOW_SYMLINKS",
```
moskyb, line 123: the name and the envar are disagreeing here; one of them is skip, the other is follow. i'd default to them both being follow, given the prior art with glob-resolve, but i'm not super fussed

moskyb (changes requested):

clicommand/artifact_upload.go
triarius: upload-follow-symlinks was an early draft. I've replaced (and negated) it to upload-skip-symlinks

moskyb (approved):
love your work 💚

PR #2074, opened 2023-05-08
Wrap errors in git checkout process

moskyb:
When the checkout process errors out, it's usually to a command that we run failing. However, when these commands fail, it's usually with an unhelpful error like `exit status 128`, which isn't really helpful for diagnosing the actual error that occurred.

This PR updates the checkout process to wrap all of the errors coming out, making it easier to diagnose exactly where the problem actually is.


triarius (approved):
Looks great!

DrJosh9000 (approved):
Excellent ✅✅✅

PR #2076, opened 2023-05-09
fix: 🪶 experiment `normalised-upload-paths` was left out of the list of available experiments

MatthewDolan:
We are seeing this warning in our Buildkite jobs:

```
WARN   Unknown experiment enabled: "normalised-upload-paths"
```

It looks like the experiment is present and does do something.

I would hypothesize that this is just an accidental oversight to leave this out of the list.

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -1308,7 +1308,7 @@ func (b *Bootstrap) defaultCheckoutPhase(ctx context.Context) error {
 	var mirrorDir string
 
 	// If we can, get a mirror of the git repository to use for reference later
-	if experiments.IsEnabled(`git-mirrors`) && b.Config.GitMirrorsPath != "" && b.Config.Repository != "" {
+	if experiments.IsEnabled(experiments.GitMirrors) && b.Config.GitMirrorsPath != "" && b.Config.Repository != "" {
 		b.shell.Commentf("Using git-mirrors experiment 🧪")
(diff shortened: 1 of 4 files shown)


bootstrap/bootstrap.go
```diff
@@ -1308,7 +1308,7 @@ func (b *Bootstrap) defaultCheckoutPhase(ctx context.Context) error {
 	var mirrorDir string
 
 	// If we can, get a mirror of the git repository to use for reference later
-	if experiments.IsEnabled(`git-mirrors`) && b.Config.GitMirrorsPath != "" && b.Config.Repository != "" {
```
moskyb, line 1311: aha, i was foolish in searching only for `"git-mirrors"`. nice catch!

moskyb (approved):
ah, i can't believe i missed this! thanks so much @MatthewDolan, nice catch

moskyb:
> I would hypothesize that this is just an accidental oversight to leave this out of the list.

you would hypothesize correctly 😅

MatthewDolan:
Sorry. I fixed on last typo I noticed in the variable name (I had used the American spelling) so it disabled auto-merge. Thanks for the review!

moskyb:
ah thank you!

bootstrap/bootstrap.go
lox: Sorry, bad habit of mine is using `` for const-ish strings.

lox:
Any chance of getting a release of this out ASAP @moskyb?

PR #2077, opened 2023-05-15
Socket refactor

DrJosh9000:
`internal/socket` was written in a way to enable an easy refactor of `jobapi` to use it. Here's the refactor!


internal/socket/utils.go
```diff
@@ -54,3 +56,10 @@ func GenerateToken(len int) (string, error) {
 	withEqualses := base64.URLEncoding.EncodeToString(b)
 	return strings.TrimRight(withEqualses, "="), nil // Trim the equals signs because they're not valid in env vars
 }
+
+// WriteError writes an error as an ErrorResponse (JSON-encoded). The err value
+// is converted to a string with fmt.Sprint.
+func WriteError(w http.ResponseWriter, err any, code int) {
+	w.WriteHeader(code)
+	json.NewEncoder(w).Encode(ErrorResponse{Error: fmt.Sprint(err)})
```
moskyb, line 64: should we check the error returned by encode? i think it's probably not worth bothering, but perhaps worth thinking about?

if not, i'm a proponent of
```suggestion
	_ = json.NewEncoder(w).Encode(ErrorResponse{Error: fmt.Sprint(err)})
```

jobapi/payloads.go
```diff
@@ -1,11 +1,13 @@
...
 
-// Error response is the response body for any errors that occur
-type ErrorResponse struct {
-	Error string `json:"error"`
-}
+	"github.com/buildkite/agent/v3/internal/socket"
+)
+
+// ErrorResponse is the response body for any errors that occur.
+type ErrorResponse = socket.ErrorResponse
```
moskyb, line 10: would this cause problems for anyone importing the client library if it relies on an internal module?

moskyb (approved):

jobapi/payloads.go
DrJosh9000: It shouldn't cause problems. I tested it to be sure 🤓

internal/socket/utils.go
DrJosh9000: So, I did it (and added logging for a couple of other errors). It's... fine.

PR #2078, opened 2023-05-16
Persist Go module cache in a bind mount

DrJosh9000:
To speed up tests, store the Go module cache in a common bind mount. Given the design of Go modules, I expect this to be safe and robust.

While it would work in a named volume with a fixed name, the Docker compose plugin helpfully cleans up named volumes in its exit hook.


triarius (approved):
🚀

PR #2096, opened 2023-05-18
Always test in race mode

moskyb:
We currently have one step in CI for running the tests in race mode, but it only runs those race tests against the linux-arm64 target.

We originally added the one race test because there were certain tests that we knew contained data races, and wanted to be able to skip them only in race mode. Now though, i think we've resolved all of those data races (largely due to @DrJosh9000's amazing work 🎉), so we should just always test in race mode.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -46,20 +46,6 @@ steps:
         files: "junit-*.xml"
         format: "junit"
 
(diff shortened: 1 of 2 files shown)


moskyb:
turns out `-race` requires CGO, which requires gcc to be installed on the machine, which won't work for windows. sigh.

DrJosh9000:
What a beautiful world it would have been.

PR #2098, opened 2023-05-18
Update all alpine images to 3.18

moskyb:
Also, remove docker-compose v1 and replace it with a symlink to docker compose v2. docker-compose v1 is no longer offered on the apkn registry on alpine versions >3.18, so we'll have to make do.


triarius (approved):
Amazing 🐳

PR #2110, opened 2023-05-23
Rename polyglot hook env BUILDKITE_HOOK_TRIGGERED_FROM -> BUILDKITE_HOOK_SCOPE

moskyb:
It was a very weird decision for me to call it `TRIGGERED_FROM`, idk why i did that ¯\\\_(ツ)_/¯


DrJosh9000 (approved):
Yeah, sounds better to me!

moskyb:
i wonder if we should add those envars to all hook runs? a decision for later

PR #2112, opened 2023-05-23
Update github.com/qri-io/jsonschema to 0.2.1

DrJosh9000:
This is #2027 redone with contexts passed in via arg, and including the config via `% #v` in the test error output.

Supersedes #2027

### DIFF
--- a/agent/plugin/definition_test.go
+++ b/agent/plugin/definition_test.go
@@ -98,11 +98,11 @@ func TestDefinitionValidatesConfiguration(t *testing.T) {
 	res := validator.Validate(context.Background(), def, cfg)
 
 	if res.Valid() {
-		t.Errorf("validator.Validate(def, %v).Valid() = true, want false", cfg)
+		t.Errorf("validator.Validate(def, % #v).Valid() = true, want false", cfg)
 	}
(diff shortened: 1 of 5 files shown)


agent/plugin/definition_test.go
```diff
@@ -123,17 +126,18 @@ func TestDefinitionWithoutAdditionalProperties(t *testing.T) {
...
+	cfg := map[string]any{
 		"alpacas": "definitely",
 		"camels":  "never",
-	})
+	}
+	res := validator.Validate(context.Background(), def, cfg)
 
 	if res.Valid() {
-		t.Errorf("validator.Validate(def, {llamas:always,camels:never}).Valid() = true, want false")
+		t.Errorf("validator.Validate(def, %v).Valid() = true, want false", cfg)
```
moskyb, line 136:
```suggestion
		t.Errorf("validator.Validate(def, % #v).Valid() = true, want false", cfg)
```

makes the difference between:
```
map[alpacas:definitely camels:never]
```
and
```
map[string]string{"alpacas":"definitely", "camels":"never"}
```

the latter of which is much easier to copypaste if necessary, and also much easier to read, IMO

moskyb (approved):

PR #2114, opened 2023-05-25
lock acquire/do: fix timeout logic

DrJosh9000:
I got my logic backwards in an earlier PR, then never tested it.


moskyb (approved):
send it

PR #2115, opened 2023-05-25
Integration tests: Use filepath.Clean in the subDirMatcher

DrJosh9000:
This should fix the test failures on Windows.


moskyb (approved):
🤙

Issue #2100, opened 2023-05-22
environment variable name transformation question


jkburges:
We are using the zacharymctague/aws-ssm plugin to populate env vars from AWS SSM, pipeline.yml snippet:

```
    plugins:
      - zacharymctague/aws-ssm#v1.0.0:
          parameters:
            BUNDLE_GEMS__CONTRIBSYS__COM: /ci/sidekiq_pro_api_key
```

Now the sidekiq gem, for which we are configuring the env var, expects it to be as written - i.e. with the double underscores - but I think the Buildkite agent code is transforming it to in env var BUNDLE_GEMS_CONTRIBSYS_COM (single underscores), which is not what we want.

We can probably work around this but I was wondering why this manipulation (replacing multiple underscores with single) is happening?

Code which currently does it I think: https://github.com/buildkite/agent/blob/main/agent/plugin/plugin.go#L211, but the logic seems to have been introduced originally at https://github.com/buildkite/agent/commit/acf4f7d16ef6d5e0afec952b098790d396156b6d#diff-ed8ae83218edc2884a09b5742c7c139b1a1340b38ddb31045bc42507e70c9fb7R177.

triarius:
Hi @jkburges, thanks for reporting this!

You can probably tell from the code that there is a series of transformation made to turn yaml keys into environment variables. Some of these are to replace characters that can't appear in environment variable names, though it is by no means comprehensive. The reasoning for replacing consecutive underscores with a single underscore is not as clear, but I don't think it is out of any necessity, so not applying this transformation is an option.

I've discussed this with the team, and we've decided that to maintain compatibility, we think the best way to this is to (at least for a period) export both versions of the variable name - one allowing consecutive underscores, and only not. Then, the SSM plugin should be able to fill in both with the same value from the SSM parameter store.

In the meantime, I've [forked](https://github.com/triarius/aws-ssm-buildkite-plugin) our version of the plugin, https://github.com/buildkite-plugins/aws-ssm-buildkite-plugin, to use the map of environment variable names to parameter names from the JSON exported as `BUILDKITE_PLUGIN_CONFIGURATION`. This means the `jq` is a dependency of the plugin, so it may not be something we can upstream easily. However, it is available by default on the Elastic CI Stack.

You should be able to use it with something like:
```yaml
    plugins:
      - triarius/aws-ssm#050f04af5f932c5beb4e5b68b214b894fce0ba64:
          parameters:
[...]

PR #2116, opened 2023-05-25
Preserve plugin config env var names with consecutive underscores

triarius:
When an env var name, that was derived from a plugin config, had consecutive underscores, the consecutive underscores were replaced by a single one. (For later reference, this will be called the underscore collapsing transformation). Now, both versions of the variable before and after this transformation are exported.

We also log a warning about this and log both the deprecated variable name and its replacement.
We need to keep the deprecated version around, as these environment variables are how plugins read their config without depending on a tool like `jq` to parse the config.
So removing them would mean that plugins could be inadvertently misconfigured, and they would have to be rewritten to start using the replacement environment variables instead.

Implementing this was not that straightforward, as the function that creates the environment slice from the plugin config is recursive and did not take a logger or return errors. It is recursive as that is the most straightforward way to turn a tree like object such as 
```yaml
parameter:
  ci:
    _api_key: /path
    db:
      host: localhost
      port: 5432
```
Into the list of environment variables, similar to:
```bash
PARAMETER_CI_API_KEY=/path
PARAMETER_CI_DB_HOST=localhost
PARAMETER_CI_DB_PORT=5432
```
I could have adapted it to do something when it hits a leaf node and the path to root had consecutive underscores, but I think that would complicate the recursion too much.

Instead, I just removed the underscore collapsing transformation and added a pass over the result that "fans-out" environment variable names with consecutive underscores into two environment variables with and without the underscore collapsing transformation applied.

So in the example above, there will now be an additional environment variable:
```bash
PARAMETER_CI__API_KEY=/path
PARAMETER_CI_API_KEY=/path
PARAMETER_CI_DB_HOST=localhost
PARAMETER_CI_DB_PORT=5432
```

Also, for passing the errors back to bootstrap, I made two custom error types. It's arguably a little overengineered, but I tried to find a way to use error wrapping and/or `errors.Join` to do the job, but they weren't powerful enough, or I was holding them wrong. There could be some generics libraries we could import to eliminate a lot of the code written for the error types. I would be glad to do this.

Fixes: #2100

#### Before
2023-05-26T15:14:46,282867146+10:00

#### After
2023-05-26T15:14:26,392941516+10:00

### DIFF
--- a/agent/plugin/error.go
+++ b/agent/plugin/error.go
@@ -22,6 +22,10 @@ func (e *DeprecatedNameErrors) IsEmpty() bool {
 // Errors returns the contained set of errors in sorted order
 func (e *DeprecatedNameErrors) Errors() []DeprecatedNameError {
 	if e == nil {
+		return nil
+	}
+
+	if e.errs == nil {
 		return []DeprecatedNameError{}
 	}
 
--- a/agent/plugin/error.go
+++ b/agent/plugin/error.go
@@ -1,0 +1,129 @@
+package plugin
+
+import (
+	"errors"
+	"fmt"
+	"sort"
+	"strings"
+)
+
+// DeprecatedNameErrors contains an aggregation of DeprecatedNameError
+type DeprecatedNameErrors struct {
+	errs []DeprecatedNameError
+}
+
+// Errors returns the contained set of errors in sorted order
+func (e *DeprecatedNameErrors) Errors() []DeprecatedNameError {
+	if e == nil {
+		return []DeprecatedNameError{}
+	}
+
+	errs := e.errs
+	sort.Slice(errs, func(i, j int) bool {
+		if errs[i].old == errs[j].old {
+			return errs[i].new < errs[j].new
+		}
+		return errs[i].old < errs[j].old
+	})
+
+	return errs
+}
+
+// Len is the length of the underlying slice or 0 if nil
+func (e *DeprecatedNameErrors) Len() int {
+	if e == nil {
+		return 0
+	}
+	return len(e.errs)
+}
+
+// Error returns each contained error on a new line
+func (e *DeprecatedNameErrors) Error() string {
+	builder := strings.Builder{}
+	for i, err := range e.Errors() {
+		_, _ = builder.WriteString(err.Error())
+		if i < len(e.errs)-1 {
+			_, _ = builder.WriteRune('\n')
+		}
+	}
+	return builder.String()
+}
+
+// Append adds DeprecatedNameError contained set and returns the reciver.
+// Returning the reveiver is necessary to support appending to nil. So this
+// should be used just like the builtin `append` function.
+func (e *DeprecatedNameErrors) Append(errs ...DeprecatedNameError) *DeprecatedNameErrors {
+	if e == nil {
+		return &DeprecatedNameErrors{errs: errs}
+	}
+
+	e.errs = append(e.errs, errs...)
+
+	return e
+}
+
+// Is returns true if and only if a error that is wrapped in target
+// contains the same set of DeprecatedNameError as the receiver.
+func (e *DeprecatedNameErrors) Is(target error) bool {
+	if e == nil {
+		return target == nil
+	}
+
+	var targetErr *DeprecatedNameErrors
+	if !errors.As(target, &targetErr) {
+		return false
+	}
+
+	dict := make(map[DeprecatedNameError]int, len(e.errs))
+	for _, err := range e.errs {
+		if c, exists := dict[err]; !exists {
+			dict[err] = 1
+		} else {
+			dict[err] = c + 1
+		}
+	}
+
+	for _, err := range targetErr.errs {
+		c, exists := dict[err]
+		if !exists {
+			return false
+		}
+		dict[err] = c - 1
+	}
+
+	for _, v := range dict {
+		if v != 0 {
+			return false
+		}
+	}
+
+	return true
+}
+
+// DeprecatedNameError contains information about environment variable names that
+// are deprecated. Both the deprecated name and its replacement are held
+type DeprecatedNameError struct {
+	old string
+	new string
+}
+
+func NewDeprecatedNameError(oldName, newName string) DeprecatedNameError {
+	return DeprecatedNameError{old: oldName, new: newName}
+}
+
+func (e *DeprecatedNameError) Error() string {
+	return fmt.Sprintf(" deprecated: %q\nreplacement: %q\n", e.old, e.new)
+}
+
+func (e *DeprecatedNameError) Is(target error) bool {
+	if e == nil {
+		return target == nil
+	}
+
+	var targetErr *DeprecatedNameError
+	if !errors.As(target, &targetErr) {
+		return false
+	}
+
+	return e.old == targetErr.old && e.new == targetErr.new
+}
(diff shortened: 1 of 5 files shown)


bootstrap/bootstrap.go
```diff
@@ -855,21 +855,30 @@ func (b *Bootstrap) executePluginHook(ctx context.Context, name string, checkout
...
 		}
 
-		env, _ := p.ConfigurationToEnvironment()
-		err = b.executeHook(ctx, HookConfig{
+		envMap, err := p.ConfigurationToEnvironment()
+		if dnerr := (&plugin.DeprecatedNameErrors{}); err != nil && errors.As(err, &dnerr) {
+			b.shell.Logger.Headerf("Deprecated environment variables for plugin %s", p.Plugin.Name())
+			b.shell.Logger.Warningf(
+				"The configuration would export environment variables with the following names in a deprecated manner",
+			)
```
moskyb, line 863: it might be worth adding a "but don't worry, everything is fine for now" thing, like:
```suggestion
			b.shell.Logger.Warningf(
				"The configuration would export environment variables with the following names in a deprecated manner. We'll export both the deprecated and the replacement names, but this behaviour may change in future",
			)
```

moskyb: also: super nit:
```suggestion
			b.shell.Logger.Warningf("The configuration would export environment variables with the following names in a deprecated manner")
```

moskyb (approved):

agent/plugin/plugin.go
```diff
@@ -257,6 +271,26 @@ func (p *Plugin) ConfigurationToEnvironment() (*env.Environment, error) {
...
+		k, v, ok := strings.Cut(kv, "=")
+
+		if !ok { // this is impossible as the array contains strings with an "=" by construction
+			continue
+		}
+
+		noConsecutiveUnderScoreKey := consecutiveUnderscoreRE.ReplaceAllString(k, "_")
+		if k != noConsecutiveUnderScoreKey {
+			envSlice = append(envSlice, fmt.Sprintf("%s=%s", noConsecutiveUnderScoreKey, v))
+			err = errors.Join(err, &DeprecatedNameError{old: noConsecutiveUnderScoreKey, new: k})
```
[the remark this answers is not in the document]
triarius: In 1.20. But it wasn't powerful enough for me here :(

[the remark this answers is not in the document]
triarius: For testing, I needed a way to make sure that the expected set of errors is exactly the set of the actual errors. But I could only see how to `errors.Is` on an error created with `errors.Join` in a way that allows you to test that the expected set of errors is a subset of the actual errors.

agent/plugin/error.go
```diff
@@ -0,0 +1,123 @@
...
+	})
+
+	return errs
+}
+
+// Error returns each contained error on a new line
+func (e *DeprecatedNameErrors) Error() string {
+	builder := strings.Builder{}
+	for i, err := range e.Errors() {
+		_, _ = builder.WriteString(err.Error())
```
DrJosh9000, line 47: Builder.WriteString (and WriteRune) are [documented](https://pkg.go.dev/strings#Builder.WriteString) as returning nil error always, so the underscores seem extra unnecessary here.

agent/plugin/plugin.go
```diff
@@ -242,33 +237,62 @@ func walkConfigValues(prefix string, v any, into *[]string) error {
 	return fmt.Errorf("Unknown type %T %v", v, v)
 }
 
+// The input should be a slice of Env Variables of the form `k=v` where `k` is the variable name
+// and `v` is its value. If it can be determined that any of the env variables names is part of a
+// deprecation, either as the deprecated variable name or its replacement, append both to the
+// returned slice, and also append a deprecation error to the error value. If there are no
+// deprecations, the error value is guaranteed to be nil.
+func fanOutDeprectatedEnvVarNames(envSlice []string) ([]string, error) {
```
DrJosh9000, line 245:
```suggestion
func fanOutDeprecatedEnvVarNames(envSlice []string) ([]string, error) {
```

agent/plugin/error.go
```diff
@@ -0,0 +1,123 @@
...
+
+// IsEmpty return true if and only if `e` contains no errors
+func (e *DeprecatedNameErrors) IsEmpty() bool {
+	return e == nil || len(e.errs) == 0
+}
+
+// Errors returns the contained set of errors in sorted order
+func (e *DeprecatedNameErrors) Errors() []DeprecatedNameError {
+	if e == nil {
+		return []DeprecatedNameError{}
```
DrJosh9000, line 25: Why not `return nil`?

bootstrap/bootstrap.go
```diff
@@ -859,8 +859,10 @@ func (b *Bootstrap) executePluginHook(ctx context.Context, name string, checkout
 		if dnerr := (&plugin.DeprecatedNameErrors{}); err != nil && errors.As(err, &dnerr) {
 			b.shell.Logger.Headerf("Deprecated environment variables for plugin %s", p.Plugin.Name())
 			b.shell.Logger.Printf(
-				"The configuration would export environment variables with the following names in a deprecated manner." +
-					"We'll export both the deprecated and the replacement names, but this behaviour may change in future",
+				"The way that environment variables are derived from the plugin configuration is changing." +
+					"We'll export both the deprecated and the replacement names for now." +
+					"You may be able avoid this by removing consecutive underscore, hypen, or whitespace" +
+					"characters in your plugin configuration.",
```
moskyb, line 865:
```suggestion
				"The way that environment variables are derived from the plugin configuration is changing. " +
					"We'll export both the deprecated and the replacement names for now. " +
					"You may be able avoid this by removing consecutive underscore, hypen, or whitespace " +
					"characters in your plugin configuration.",
```

moskyb (approved):
one tiny whitespace thing, but other than that this LGTM 🎉

bootstrap/bootstrap.go
triarius: Good catch!

agent/plugin/error.go
triarius: Yep, and I made it handle when `e.errs` is `nil` too.

agent/plugin/error.go
triarius: I don't think it's too distracting to leave it this way. It makes the life of static analysis tools easier.

pda (commented):
I haven't looked closely at the implementation, but I like what it does 👍🏼
And it looks like there's good test coverage.

bootstrap/bootstrap.go
```diff
@@ -858,12 +858,12 @@ func (b *Bootstrap) executePluginHook(ctx context.Context, name string, checkout
 		envMap, err := p.ConfigurationToEnvironment()
 		if dnerr := (&plugin.DeprecatedNameErrors{}); err != nil && errors.As(err, &dnerr) {
 			b.shell.Logger.Headerf("Deprecated environment variables for plugin %s", p.Plugin.Name())
-			b.shell.Logger.Printf(
-				"The way that environment variables are derived from the plugin configuration is changing. " +
-					"We'll export both the deprecated and the replacement names for now. " +
-					"You may be able avoid this by removing consecutive underscore, hypen, or whitespace " +
-					"characters in your plugin configuration.",
-			)
+			b.shell.Logger.Printf("%s", strings.Join([]string{
```
DrJosh9000, line 861: Just to understand, I'm guessing the problem with using a single string literal is how long it would be?

triarius: Yep. Wrapped lines are irritating for many reasons

triarius:
The log message has changed somewhat after all the review. Here's what it looks like now in a very similar context to what was in the original issue. I've expanded the relevant sections, they are not expanded by default.

2023-05-31T23:43:44,606079698+10:00

PR #2118, opened 2023-05-27
Report more AWS metadata

david-poirier:
I would find it useful to have additional AWS metadata for attribution and reporting purposes.

This PR adds:

- Account ID
- Architecture


agent/ec2_meta_data.go
```diff
@@ -41,10 +41,13 @@ func (e EC2MetaData) Get() (map[string]string, error) {
...
 
-	metaData["aws:instance-id"] = document.InstanceID
-	metaData["aws:instance-type"] = document.InstanceType
+	metaData["aws:account-id"] = document.AccountID
 	metaData["aws:ami-id"] = document.ImageID
+	metaData["aws:architecture"] = document.Architecture
 	metaData["aws:availability-zone"] = document.AvailabilityZone
+	metaData["aws:instance-id"] = document.InstanceID
+	metaData["aws:instance-type"] = document.InstanceType
+	metaData["aws:private-ip"] = document.PrivateIP
```
moskyb, line 50: hey! these are all really good changes, but i wonder a little bit about including the private IP in the default set of tags for AWS instances...

while in terms of private info, the private IP isn't _that_ sensitive, it is a little bit sensitive. a lot of our customers operate the agent in highly regulated and/or sensitive environments, and might see this as a leak of sensitive information to a third party (the Buildkite backend).

i think for now we should leave the private IP out of the default tags, but as a workaround, you could add this to your agent startup script to populate it at when the agent launches:

```bash
IMDSv2_TOKEN=$(curl -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 21600")
PRIVATE_IP=$(curl -H "X-aws-ec2-metadata-token: $IMDSv2_TOKEN" http://169.254.169.254/latest/dynamic/instance-identity/document | jq -r .privateIp)
buildkite-agent start <your other args here> --tags "aws:private-ip=$PRIVATE_IP"
```

moskyb (changes requested):
hey @david-poirier, thanks for your contribution! this looks great; i have one note about including the private IP in the default tags, but once that's resolved, this LGTM and we can get it shipped 🚀

moskyb (changes requested):
hey @david-poirier, thanks for your contribution! this looks great; i have one note about including the private IP in the default tags, but once that's resolved, this LGTM and we can get it shipped 🚀

agent/ec2_meta_data.go
david-poirier, line 50:
```suggestion
```

david-poirier: nw, I dropped it ;)
thanks for the suggested workaround 👍🏼

moskyb (approved):
love it 🎉

PR #2124, opened 2023-05-30
Internalise utils

DrJosh9000:
One day, utils will be cleaned up. Until then, we should reduce its exposure to the world.


triarius (approved):
Yas!

PR #2129, opened 2023-05-31
Delete race

DrJosh9000:
Nothing uses it?


moskyb (approved):
we used to, when there were tests that we needed to skip in race mode, but someone (👀) squashed all of those

PR #2131, opened 2023-05-31
De-experimentify flock-file-locks

DrJosh9000:
As @moskyb wrote elsewhere:

> when we release this, we should include a note that says that agents on this version will be incompatible with agents of other versions on the same instance sharing the same git repo - all agents will have to be upgraded at the same time, otherwise they won't see each other's locks


moskyb (approved):
🐦‍⬛🐦🕊️📁🔒

moskyb (approved):
🐦‍⬛🐦🕊️📁🔒

lox:
For what it's worth, flock's did not do what they said on the tin for us with workloads running inside docker containers (agent on host, bootstrap in containers). File locks did not survive outside of the containers.

PR #2132, opened 2023-05-31
De-experimentify ansi-timestamps

DrJosh9000:
The future is now!


moskyb (changes requested):
hey, this looks awesome and i would looooooooooooooove to get rid of the header times streamer, but...

i think for compatibility we need to 
- add an option to disable ansi timestamps (`--no-ansi-timestamps`) or similar
- retain the ability to add timestamp lines (and allow them to be used with ANSI timestamps, something that was previously not allowed)

i'd like to commit to getting to the point where this PR currently is when we do another major version/versioning switchup (which is feeling more and more like a real thing as time goes on), but for now i think we probably have to retain the old behaviour as an opt-in :/

lox:
I am SO happy to see this.

agent/job_runner.go
```diff
@@ -209,58 +203,13 @@ func NewJobRunner(l logger.Logger, scope *metrics.Scope, ag *api.AgentRegisterRe
...
-				runner.headerTimesStreamer.Scan(line)
-			})
-			if err != nil {
-				l.Error("[JobRunner] Encountered error %v", err)
-			}
-		}()
-	}
+	// Prefix each line with an ANSI escape containing the millisecond timestamp
+	// of the line (the artist formerly known as "ansi-timestamps").
+	prefixer := process.NewPrefixer(runner.output, func() string {
```
lox, line 208: Assume this new technology handles long lines?

DrJosh9000: It should - it is implemented in a buffer-free way.

DrJosh9000:
ansi-timestamps is officially de-experimented in #2133, but deleting all this code (if only iin theory) was fun 😈

PR #2133, opened 2023-06-01
De-experimentify ansi-timestamps (take 2)

DrJosh9000:
Instead of enabling the current ANSI timestamp behaviour with no opt-out, this one provide an escape hatch to the old defaults with --no-ansi-timestamps.

I thought for a bit about how to make ANSI timestamps and TimestampLines work together. It could avoid having a header times streamer for the purpose of streaming header times, but would still need the logic in the header times streamer to distinguish headers from non-headers, as well as something else to buffer the first few bytes of each line (at least). That bit needs more thought, so for now, continue to not mix ANSI timestamps and TimestampLines.

### DIFF
--- a/agent/agent_configuration.go
+++ b/agent/agent_configuration.go
@@ -24,6 +24,7 @@ type AgentConfiguration struct {
 	PluginValidation           bool
 	LocalHooksEnabled          bool
 	RunInPty                   bool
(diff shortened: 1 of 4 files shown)


triarius (approved):
Nice 🎉

PR #2135, opened 2023-06-01
Reduce header regexps

DrJosh9000:
headerRegex (now headerRE) was used to find headers. Because it was anchored on both ends (^...$), it (morally) had to run through the whole string in order to match, though I hope that the RE2-like regexp engine optimises (.+)?$ to nothing.

We can say the same for headerExpansionRE. Matching `\s+` in between ^^^ and +++ just seems unnecessary - we document it as "^^^ +++". So I changed it to `\s`.

In both cases the new regexps only need to process either 4 or 7 bytes in order to match. 

The argument for removing matches of ansiColourRE applies to header expansions as well, so the fact that isHeaderExpansion didn't do the same thing is arguably a bug. Instead, I rearranged things so headerExpansionRE can test the same decolourised line (inside Scan), removing both isHeader and isHeaderExpansion.

### DIFF
--- a/agent/header_times_streamer.go
+++ b/agent/header_times_streamer.go
@@ -150,15 +150,17 @@ func (h *headerTimesStreamer) Upload(ctx context.Context) {
 	timesToUpload := len(times)
 
 	// Do we even have some times to upload
-	if timesToUpload > 0 {
-		// Call our callback with the times for upload
-		h.logger.Debug("[HeaderTimesStreamer] Uploading header times %d..%d", c, length-1)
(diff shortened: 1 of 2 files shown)


triarius (commented):
LGTM, I approve 🎉 Going to just leave a comment to check something. I'll formerly approve it later. It still needs a rebase before merging, though.

triarius (approved):
LGTM 🎉

triarius:
The thing I was checking is whether this PR would appear here https://github.com/pulls/review-requested

PR #2137, opened 2023-06-02
Wipe checkout directory on `git checkout` and `git fetch` failure and retry

triarius:
There are a certain class of errors from git in the `defaultCheckoutPhase` that would cause a retry, but before the retry, the checkout directory would be removed. This allows cleaning up in case an unclean exit from git had left the checkout directory in an unrecoverable state.

This PR adds failures in `git fetch` and `git checkout` when the exit code is 128 to that class of error. It also fixes the detection of errors for which the check directory needed to be clean, as previously, the detection would fail if the errors were wrapped, and they are wrapped in the `defaultCheckoutPhase` method.

I've done some refactoring to reduce the indentation level of the code, so I recommend reviewing with whitespace changes hidden.

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -1171,6 +1171,7 @@ func (b *Bootstrap) CheckoutPhase(ctx context.Context) error {
 					case gitErrorClean:
 					case gitErrorCleanSubmodules:
 					case gitErrorClone:
+					case gitErrorFetch:
 						// do nothing, this will fall through to destroy the checkout
 
 					default:
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -1160,7 +1160,7 @@ func (b *Bootstrap) CheckoutPhase(ctx context.Context) error {
 					b.shell.Warningf("Checkout failed! %s (%s)", err, r)
 
 					// Specifically handle git errors
-					if ge, ok := err.(*gitError); ok {
+					if ge := new(gitError); errors.As(err, &ge) {
 						switch ge.Type {
 						// These types can fail because of corrupted checkouts
 						case gitErrorClone:
@@ -1186,7 +1186,6 @@ func (b *Bootstrap) CheckoutPhase(ctx context.Context) error {
 					if err := b.createCheckoutDir(); err != nil {
 						return err
 					}
-
 				}
 
 				return err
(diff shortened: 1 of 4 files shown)


bootstrap/bootstrap.go
```diff
@@ -1133,75 +1133,81 @@ func (b *Bootstrap) CheckoutPhase(ctx context.Context) error {
...
-							// do nothing, this will fall through to destroy the checkout
-
-						default:
-							return err
-						}
-					}
+		if b.Config.Repository == "" {
+			b.shell.Commentf("Skipping checkout, BUILDKITE_REPO is empty")
+			break
+		}
```
triarius, line 1139: This used to be on 1197. I've inverted the condition to decrease the indentation.

bootstrap/git.go
```diff
@@ -52,6 +59,11 @@ func gitCheckout(ctx context.Context, sh shellRunner, gitCheckoutFlags, referenc
 		if strings.HasPrefix(err.Error(), "fatal: reference is not a tree: ") {
 			return &gitError{error: err, Type: gitErrorCheckoutReferenceIsNotATree}
 		}
+
+		// 128 is extremely broad, but it seems permissions errors, network unreaable errors etc, don't result in it
```
DrJosh9000, line 63:
```suggestion
		// 128 is extremely broad, but it seems permissions errors, network unreachable errors etc, don't result in it
```

bootstrap/git.go
```diff
@@ -127,6 +139,10 @@ func gitFetch(ctx context.Context, sh shellRunner, gitFetchFlags, repository str
 	}
 
 	if err := sh.Run(ctx, "git", commandArgs...); err != nil {
+		// 128 is extremely broad, but it seems permissions errors, network uncreachable errors etc, don't result in it
```
DrJosh9000, line 142:
```suggestion
		// 128 is extremely broad, but it seems permissions errors, network unreachable errors etc, don't result in it
```

bootstrap/integration/bootstrap_tester.go
```diff
@@ -211,19 +213,19 @@ func (b *BootstrapTester) writeHookScript(m *bintest.Mock, name string, dir stri
 		body = "#!/bin/sh\n" + strings.Join(append([]string{m.Path}, args...), " ")
 	}
 
-	if err := os.MkdirAll(dir, 0700); err != nil {
+	if err := os.MkdirAll(dir, 0o700); err != nil {
```
DrJosh9000, line 216: 👍👍👍

DrJosh9000 (approved):
I like this!

PR #2146, opened 2023-06-06
Use Agent API for locking mirrors and known_hosts

DrJosh9000:
If the Agent API experiment is enabled and working, then shell will now use it for locking instead of `flock`. 

At least two reasons for this:

1. The API socket can be bind-mounted into a container and "just work", whereas flock is harder to make work correctly across a container boundary.
2. Locking and unlocking in the Agent API has no disk cost, so can be polled more frequently, so should be slightly faster.

### DIFF
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -732,7 +732,7 @@ func (w LogWriter) Write(bytes []byte) (int, error) {
 func (r *JobRunner) executePreBootstrapHook(ctx context.Context, hook string) (bool, error) {
 	r.logger.Info("Running pre-bootstrap hook %q", hook)
 
-	sh, err := shell.New()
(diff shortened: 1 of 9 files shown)


moskyb (commented):
i like this but.... a [post on the community slack](https://buildkite-community.slack.com/archives/C02T53V9H/p1688981483343859) reminded me that we probably can't use the agent locking service for locking mirrors, as mirrors are sharable through [E|Z]FS, and mirror locks through the agent API wouldn't be propagated through to the other users of the shared git mirror :/

DrJosh9000:
Perhaps the choice of "lock providers" should be configurable? 

Oooh. What about making it a _hook_?

DrJosh9000:
Closing due to staleness... and it probably didn't have legs, per @moskyb 's comment.

PR #2154, opened 2023-06-13
Rewrite the redactor

### LINKED ISSUE #2062: Support multi-line secrets redaction
The redactor does not support redacting multiline secrets. We heavily use multiline secrets for various agent set up scripts, like distributing SSH keys for git access.

https://sourcegraph.com/github.com/buildkite/agent/-/blob/redaction/redactor.go?L240
Update the redactor to support redacting multiline secrets. 

We could clean up the env vars immediately after using them (eg SSH_KEY >> ssh_file; unset SSH_KEY). 
However, our customers are distributed and empowered to freely create new CI workflows to suit their needs. It would be impossible to enforce this clean up in a maintainable way.

DrJosh9000:
The old redactor is fast, and pretty good - why a whole rewrite? How is this any better?

1. Using a Algorithm With A Name (in the existing case, Boyer-Moore) is a double-edged sword. On one hand, it's a known quantity and we can compare implementations. But paradoxically, because it's got Names associated with it, maintainers have a _lack_ of confidence in their ability to do any kind of maintenance here. It Looks Complicated.
2. Requiring a manual `Flush` before a `Reset` is a possibly dangerous API.
3. Lack of multi-line secret redaction is Really Not Good. Lots of secrets stretch over multiple lines.
4. Not handling nested and overlapping secrets is Surprising, which is even worse.

So, here is a new implementation. Finding the Name of this new algorithm is left as an exercise for the reader. It handles multiple lines and nested and overlapping secrets, and should be more accessible to future maintainers. You can call `Reset` to change the secrets it is redacting mid-stream without need to call `Flush` first. It has many more tests, and also has a fuzz test (try `go test ./internal/redactor -fuzz=FuzzRedactor`).

The Boyer-Moore based redactor still exists in a test file, in order to produce benchmarks:

```
% go test  ./internal/redactor -bench='Benchmark.*Redactor'                                         
goos: darwin
goarch: arm64
pkg: github.com/buildkite/agent/v3/internal/redactor
BenchmarkBMRedactor-10             94035             10878 ns/op
BenchmarkRedactor-10              105124             10529 ns/op
PASS
ok      github.com/buildkite/agent/v3/internal/redactor 2.789s
```

EDIT: After stealing an optimisation from the old redactor, the new redactor is about as fast as the old one!

Fixes #2062


bootstrap/bootstrap.go
```diff
@@ -2088,40 +2088,36 @@ func (b *Bootstrap) setupRedactors() redaction.RedactorMux {
 		b.shell.Commentf("Enabling output redaction for values from environment variables matching: %v", b.Config.RedactedVars)
 	}
 
-	var mux redaction.RedactorMux
+	var mux redactor.Mux
```
triarius, line 2091: 💜

internal/redactor/redactor.go
```diff
@@ -0,0 +1,425 @@
...
+	// Intermediate buffer to account for partially-written non-secrets.
+	// (i.e. we began redacting in case we're in the middle of a secret, but
+	// we might not be).
+	buf []byte
+
+	// Current redaction states - if we have begun redacting a potential secret
+	// there will be at least one of these.
+	// nextStates is the next set of states. To avoid creating millions of new
+	// slices, Write alternates between these two.
+	states, nextStates []state
```
moskyb, line 51: can you go into a bit more detail (here in the PR, rather than in code) as to why we have the two state arrays and we swap through them? i'm not 100% sure i understand how/why this helps us avoid creating millions of new slices

internal/redactor/redactor.go
```diff
@@ -0,0 +1,425 @@
...
+
+			// It matched!
+			s.matched++
+
+			// Have we fully matched this needle?
+			if s.matched < len(s.needle) {
+				// This state survives for another byte.
+				r.nextStates = append(r.nextStates, s)
+				continue
+			}
```
moskyb, line 125: if we have a partial match on a write - ie we have a needle `foobar`, and we're writing `fomenting` to the redactor - does this mean that the `state` for the partial match on `fo` will stick around until we've processed the entire string that we're writing? i don't think this is a flaw, just confirming if i'm reading the code right - it's implied(?) by the code, but not explicit.

moskyb (approved):
i've had a fairly deep read of this PR and it looks good to me, as far as i can tell. i think i understand 95% of what's going on, and i have a couple of informational questions about the stuff i'm not clear on.

awesome work! this is way more understandable than the previous implementation. i propose we call it the `Deprez Redactor`.

internal/redactor/redactor.go
moskyb: 🤔 given that `states` is a pretty vague name, (and if i'm reading what these slices actually do correctly), would it be slightly clearer to name these `partialMatches` (and `nextPartialMatches` i guess? `partialMatches1` and partialmatches2`?) and `redact` something like `completeMatches`?

internal/redactor/redactor.go
```diff
@@ -0,0 +1,425 @@
...
+	r.buf = r.buf[bufidx:]
+
+	// Because redactions refer to buffer positions, and the buffer shrank,
+	// update the redaction ranges to point at the correct locations in the
+	// buffer. We also move them to the start of the slice to avoid allocation.
+	rem := len(r.redact[done+1:]) // remaining ranges
+	for i, rg := range r.redact[done+1:] {
+		r.redact[i] = rg.sub(bufidx)
+	}
+	r.redact = r.redact[:rem]
```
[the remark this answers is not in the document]
DrJosh9000: Actually, no... in the loop `i` takes the values 0 through `done` (not `done+1` and up), so it moves the items down at the same time as updating them. If we reduce the slice length before then, we lose access to the items.

internal/redactor/redactor.go
```diff
@@ -0,0 +1,425 @@
...
+	if flushTo == 0 || len(r.buf) == 0 {
+		return nil
+	}
+
+	bufidx := 0 // where we are up to in the buffer
+
+	// Stop when we're out of redactions, or the next one is after flushTo.
+	// Track the last range processed.
+	done := -1
+	for ri, rg := range r.redact {
```
moskyb, line 201: [whatever the question equivalent of a nit is]
`ri` is `redact index`, but what does `rg` stand for? is it a contraction of `range`?

internal/redactor/redactor.go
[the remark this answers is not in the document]
moskyb: ahh, right!! of course, that makes total sense.

internal/redactor/redactor.go
DrJosh9000: It should start matching on `f`, continue matching on `o`, but then on `m`, the first `if c != s.needle[s.matched] { continue }` should drop it.

internal/redactor/redactor.go
```diff
@@ -0,0 +1,425 @@
...
+		r.states, r.nextStates = r.nextStates, r.states[:0]
+	}
+
+	// 3. Merge overlapping redaction ranges.
+	// Because they were added from start to end, they are in order.
+	r.redact = mergeOverlaps(r.redact)
+
+	// 4. Write as much of the buffer as we can without spilling incomplete
+	//    matches.
+	flushTo := len(r.buf)
```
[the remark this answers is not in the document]
DrJosh9000: I like this suggestion. `limit` for the variable, and `flushUpTo` for the method?

DrJosh9000:
I've updated with some renaming and additional comments.

> i've had a fairly deep read of this PR and it looks good to me, as far as i can tell. 

Thanks! For reviewing, and for taking the time to form a deeper understanding of it.

> awesome work! this is way more understandable than the previous implementation. 

😁

> i propose we call it the `Deprez Redactor`.

😳

internal/redactor/redactor.go
DrJosh9000: I added more comment. I also went with `partialMatches`, `nextMatches`, and `completedMatches`, thoughts?

internal/redactor/redactor.go
DrJosh9000: Given renaming `r.redact` to `r.completedMatches`, then `rg` seemed better as `match`.

internal/redactor/redactor.go
moskyb: fantastic - love it. much better than `partialMatches1` and `partialMatches2` 😅

internal/redactor/redactor.go
moskyb: right, gotcha - the bumped partial match is explicitly added to `nextPartialMatches` if the next byte matches a couple lines down from that, and if it isn't explicitly added to `nextPartialMatches`, it'll get dropped.

bootstrap/bootstrap.go
```diff
@@ -448,8 +448,8 @@ func (b *Bootstrap) applyEnvironmentChanges(changes hook.HookScriptChanges, reda
 	b.shell.Env.Apply(changes.Diff)
 
 	// reset output redactors based on new environment variable values
-	redactors.Flush()
-	redactors.Reset(redaction.GetValuesToRedact(b.shell, b.Config.RedactedVars, b.shell.Env.Dump()))
+	//redactors.Flush()
```
moskyb, line 451:
```suggestion
```

internal/redactor/redactor.go
```diff
@@ -0,0 +1,432 @@
...
+	// Keep the remainder of the buffer where it is. A future append might
+	// create a new buffer, letting the old one be GC-ed.
+	r.buf = r.buf[bufidx:]
+
+	// Because redactions refer to buffer positions, and the buffer shrank,
+	// update the redaction ranges to point at the correct locations in the
+	// buffer. We also move them to the start of the slice to avoid allocation.
+	rem := len(r.completedMatches[done+1:]) // number of remaining matches
+	for i, match := range r.completedMatches[done+1:] {
+		// Note that i ranges from 0 to done, but rg ranges the values:
```
moskyb, line 264:
```suggestion
		// Note that i ranges from 0 to done, but match ranges the values:
```

moskyb (approved):
two more extremely tiny nits, but this is good to merge IMO. awesome work 🎉

PR #2156, opened 2023-06-15
Adjust HTTP response errors to separate the response code and the error

moskyb:
A customer was running into an error:
```
POST https://agent.buildkite.com/v3/register: 400 Agents can only be part of a single queue when registering to a cluster (Attempt 14/30 Retrying in 10s)
```
which due to a set of quite funny coincidences, makes it seem like there can only be 400 agents per cluster queue, when what's actually happening is that the log is telling us that that POST got a 400 Bad Request response code, and was telling the user that each agent is only allowed to be part of a single cluster queue.

This commit adjusts the error so that it will now be:
```
POST https://agent.buildkite.com/v3/register: 400 Bad Request: Agents can only be part of a single queue when registering to a cluster (Attempt 14/30 Retrying in 10s)
```
hopefully making the various clauses in that sentence a little easier to parse out

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -299,12 +299,12 @@ type ErrorResponse struct {
 }
 
 func (r *ErrorResponse) Error() string {
-	s := fmt.Sprintf("%v %v: %d",
+	s := fmt.Sprintf("%v %v: %s",
 		r.Response.Request.Method, r.Response.Request.URL,
-		r.Response.StatusCode)
+		r.Response.Status)
 
 	if r.Message != "" {
-		s = fmt.Sprintf("%s %v", s, r.Message)
+		s = fmt.Sprintf("%s: %v", s, r.Message)
 	}
 
 	return s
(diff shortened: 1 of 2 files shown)


api/client.go
```diff
@@ -299,12 +299,12 @@ type ErrorResponse struct {
...
 func (r *ErrorResponse) Error() string {
-	s := fmt.Sprintf("%v %v: %d",
+	s := fmt.Sprintf("%v %v: %s",
 		r.Response.Request.Method, r.Response.Request.URL,
-		r.Response.StatusCode)
+		r.Response.Status)
 
 	if r.Message != "" {
-		s = fmt.Sprintf("%s %v", s, r.Message)
+		s = fmt.Sprintf(": %s %v", s, r.Message)
```
DrJosh9000, line 307: Is that right? if `r.Message != ""` then `s` looks like `: <method> <url>: Bad Request <message>`

moskyb: oh shoot, no, you're right, it should be 
```suggestion
		s = fmt.Sprintf("%s: %v", s, r.Message)
```

moskyb: i had read it as concatenating, it's actually sprintfing, shows me for not actually reading the code i'm changing

DrJosh9000 (approved):
🌈❇️

jmandel1027:
Thanks guys!

Issue #1799, opened 2022-10-24
Collect timestamp lines alongside log temp file


agates4:
**Is your feature request related to a problem? Please describe.**
When the `BUILDKITE_ENABLE_JOB_LOG_TMPFILE` environment variable is set to true, 
and `BUILDKITE_TIMESTAMP_LINES` is also set to true, 
then the generated tmp file **does not** include the timestamp lines that we'd expect it to have.

**Describe the solution you'd like**
I would like the log tmpfile to have the timestamp lines that we'd also see normally in the UI output.

**Describe alternatives you've considered**
Actually, using the `curl` api DOES include the timestamp lines. 

**Additional context**
This would be great so that we can upload and associate logs with our Datadog CI/CD pipelines!

agates4:
BTW I think #758 is complete

triarius:
Hey we agree this should be fixed. We'll attempt to get to it but can't guarantee a time-frame. In the mean time, any pull request would be welcome.

triarius:
We are now tracking this internally in our backlog. As always, PRs are welcome.

PR #2157, opened 2023-06-15
Update job log file to include line transforms

chasestarr:
Fixes https://github.com/buildkite/agent/issues/1799

Combines `runner.output` writes with `tmpFile` instead of `allWriters` so that the temp file can get any line transforms like ansi-timestmap prefixes.

I'm not quite sure how to manually test or write an automated test for this change, currently going on intuition. I would appreciate any instructions for doing so.


DrJosh9000 (approved):
Looks great @chasestarr ! Thanks so much.

Ideally we would have better testing for changes like this. We'll think about how to make this better.

PR #2158, opened 2023-06-15
Internalise pipeline parser

DrJosh9000:
`agent` is overloaded with stuff. Making a separate `pipeline` package improves the spelling IMO (`pipeline.Parser` vs `agent.PipelineParser`). 

Since it's not intended for external consumption (in current form, anyway), hide it in `internal`.


moskyb (approved):
🙏 doing the lord's work

PR #2160, opened 2023-06-19
`Buildkite-Timeout-Milliseconds` request header

pda:
Introduce a new `Buildkite-Timeout-Milliseconds` request header, set as a positive integer by `api.Client.newRequest()` when its context has a deadline in the future.

This allows the server to be aware of the client-side timeout/deadline (e.g. in #2028), and e.g. configure its own downstream timeouts accordingly.

Thanks @mhaylock for the idea behind this.

Related:
- #2028

### DIFF
--- a/api/api_internal_test.go
+++ b/api/api_internal_test.go
@@ -1,0 +1,43 @@
+package api
+
+import (
+	"context"
+	"strconv"
+	"testing"
+	"time"
+
+	"github.com/buildkite/agent/v3/logger"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNewRequestBuildkiteTimeoutMilliseconds(t *testing.T) {
+	c := NewClient(logger.NewBuffer(), Config{})
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	r, err := c.newRequest(ctx, "GET", "/foo", nil)
+	assert.NoError(t, err)
+
+	value := r.Header.Get("Buildkite-Timeout-Milliseconds")
+	ms, err := strconv.ParseInt(value, 10, 64)
+	assert.NoError(t, err)
+
+	// Allow a generous 1000ms between setting the timeout and the header being set.
+	if ms <= 9_000 || ms > 10_000 {
+		t.Errorf("Expected Buildkite-Timeout-Milliseconds to reflect 10 second timeout, got %q (%d ms)", value, ms)
+	}
+}
+
+func TestNewRequestWithoutBuildkiteTimeoutMilliseconds(t *testing.T) {
+	c := NewClient(logger.NewBuffer(), Config{})
+
+	ctx := context.Background() // no timeout/deadline
+
+	r, err := c.newRequest(ctx, "GET", "/foo", nil)
+	assert.NoError(t, err)
+
+	if value, ok := r.Header["Buildkite-Timeout-Milliseconds"]; ok {
+		t.Errorf("Expected no Buildkite-Timeout-Milliseconds header, got %q", value)
+	}
+}
--- a/api/client.go
+++ b/api/client.go
@@ -15,6 +15,7 @@ import (
 	"net/http/httputil"
 	"net/url"
 	"reflect"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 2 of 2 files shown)


api/api_internal_test.go
pda, line 1: So I explicitly _do_ want to test the non-exported `api.Client.newRequest` method, and the existing `api/*_test.go` files are `package api_test` (which is good!)… so perhaps creating an `api_internal_test.go` is sane for this use-case? I'm not sure. /cc @DrJosh9000

api/client.go
```diff
@@ -172,6 +173,15 @@ func (c *Client) newRequest(
 
 	req.Header.Add("User-Agent", c.conf.UserAgent)
 
+	// If our context has a timeout/deadline, tell the server how long is remaining.
+	// This may allow the server to configure its own timeouts accordingly.
+	if deadline, ok := ctx.Deadline(); ok {
+		ms := deadline.Sub(time.Now()).Milliseconds()
+		if ms > 0 {
+			req.Header.Add("Buildkite-Timeout-Milliseconds", strconv.FormatInt(ms, 10))
```
pda, line 181: Is `Buildkite-Timeout-Milliseconds` a good header name?

Is putting the units in the header name good? It makes parsing the value much simpler.

Is milliseconds a good time granularity? It was either that or seconds, which would probably be adequate but maybe not? And opened more (minor) questions about integer-rounding etc.

I also considered side-stepping the precision question and making the name shorted with `Buildkite-Timeout` and ISO 8601 duration value e.g. `PT10S` (10 seconds) but I couldn't see a good stdlib formatter for those, and it makes parsing it less simple, and I don't think they're _very_ widely used nor very pretty/readable 😬

I've specifically avoided the [deprecated](https://datatracker.ietf.org/doc/html/rfc6648) `X-` prefix, which we use in one or two other places. I'd rather (gradually, one day) drop the `X-` from those vs adding it here for consistency. (/cc @triarius 😁)

yob:
fancy!

api/api_internal_test.go
pda: One thing I like about black-box testing in Go is the way it illustrates the external (i.e. actual) usage of the API with the package name as part of the call (`api.NewClient()` vs `NewClient()`). And mixed with TDD it can lead to nicer naming etc, avoiding stuttering pitfalls etc.

api/client.go
```diff
@@ -172,6 +173,15 @@ func (c *Client) newRequest(
 
 	req.Header.Add("User-Agent", c.conf.UserAgent)
 
+	// If our context has a timeout/deadline, tell the server how long is remaining.
+	// This may allow the server to configure its own timeouts accordingly.
+	if deadline, ok := ctx.Deadline(); ok {
+		ms := deadline.Sub(time.Now()).Milliseconds()
```
DrJosh9000, line 179:
```suggestion
		ms := time.Until(deadline).Milliseconds()
```

api/client.go
DrJosh9000: If I were in control of the other end of the request, I'd be chucking the value through something like `time.ParseDuration`, which can interpret suffixes like `ms`. So my vote would be for `Buildkite-Timeout: 123ms`. `Buildkite-Timeout-Milliseconds` is otherwise fine with me.

triarius: Makes sense to not have `X-`, I very much regret introducing it 😅

There seems to be an assumption of seconds for various headers:

`Retry-After` is either a [date or in seconds](https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.3). `Cache-Control` seems to assume seconds for its sub-values as well. `Age` [assumes seconds](https://datatracker.ietf.org/doc/html/rfc9111#field.age).

That's very confusing, though. I prefer the units in the header name approach to that. It is possible to put the units in the header value, i.e.
```
Buildkite-Timeout: 100ms
```
And we could default to seconds if there is no unit.

But my feelings are not that strong on this matter.

triarius: ^ I guess @DrJosh9000 and I have the same position

pda: Yeah… I'd like to use a common encoding of the duration/units, but…

```ruby
ActiveSupport::Duration.parse("100ms")
# ActiveSupport::Duration::ISO8601Parser::ParsingError: Invalid ISO 8601 duration: "100ms"
```

Ruby / Rails / ActiveSupport likes ISO 8610, Go stdlib likes a nicer but non-standard form, so I reckon `Buildkite-Timeout-Milliseconds` with an integer value ends up simplest 😢

DrJosh9000 (approved):

api/client.go
pda: Hah:
image

— https://stackoverflow.com/questions/657309/how-to-parse-days-hours-minutes-seconds-in-ruby

Hello twelve-years-ago-@sj26 👋🏼

pda: (chronic_duration hasn't been updated for about a decade, GitLab have a fork https://rubygems.org/gems/gitlab_chronic_duration which hasn't been updated for 4 years, and I'd be inclined to not add libraries just for this anyway)

pda:
Rebased onto https://github.com/buildkite/agent/pull/2166 to get a green build

PR #2168, opened 2023-06-22
WIP: Parser update & command signing

DrJosh9000:
There are like 5 different ideas here; I intend to break this into multiple smaller PRs.

### DIFF
--- a/internal/pipeline/step.go
+++ b/internal/pipeline/step.go
@@ -113,20 +113,30 @@ func (c *CommandStep) interpolate(pr *Parser) error {
 
 func (CommandStep) stepTag() {}
 
-// signedFields returns the contents of fields to sign.
-func (c *CommandStep) signedFields(version string) (map[string]string, error) {
-	switch version {
-	case "v1":
-		// These fields, all these fields, and only these fields.
-		// The map will be sorted by Sign.
-		// Do not change the contents of old versions.
-		return map[string]string{
-			"command": c.Command,
-		}, nil
-
-	default:
-		return nil, fmt.Errorf("unsupported version %q", version)
+// signedFields returns the default fields for signing.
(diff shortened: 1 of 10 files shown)


internal/pipeline/step.go
```diff
@@ -0,0 +1,294 @@
...
+	}
+	if err := n.Decode(&full); err != nil {
+		return err
+	}
+
+	// Normalise into Command.
+	c.Command = full.Command
+	if c.Command == "" {
+		c.Command = strings.Join(full.Commands, "\n")
+	}
```
moskyb, line 83: i wonder if we should think about doing this the other way around - ie have `commands` be the "canonical" representation of commands. 

my reasoning here is that the `command` -> `commands` normalisation eg:
```yaml
command: >
  echo hello
  echo there
```
to 
```yaml
commands: ["echo hello", "echo there"]
```
feels like it retains more specificity than the other way around.

DrJosh9000: Thinking through this...

We upload `command` or `commands` but get back `BUILDKITE_COMMAND`. 

It should be easy to ensure consistency when computing signatures for `command` (1 string) and `BUILDKITE_COMMAND` (1 string). 

If we split `command` into `commands` (N strings), then we need to split `BUILDKITE_COMMAND` into the same N strings for the signatures to be consistent. If the user supplies `commands` and an element of `commands` contains multiple lines then this won't be consistent - we'd need to normalise by splitting each element of `commands`. Or we'd need to `strings.Join(commands, "\n")` solely for the benefit of computing signatures.

moskyb: yep, that makes total sense to me 👍 very happy to leave as is

DrJosh9000:
Superseded by #2200 (and the work that led up to it).

PR #2172, opened 2023-06-27
De-experimentify wording on the status page

DrJosh9000:
It's not an experiment anymore, but the rest of the yellow div is correct.


dabarrell (approved):
Noticed this in agent sync too! LGTM

PR #2173, opened 2023-06-28
Add ordered.Map

DrJosh9000:
In the process of trying to parse pipelines with a greater semantic understanding (in order to do things with them, like sign command steps), but retain compatibility, I kept hitting the need for YAML unmarshaling into an *ordered* map. For example, even though the "env" block is essentially unordered, we step through it *in order* so that variables can interpolate into one another *in order*. Another example is plugin config: we reflect it to plugins both via env vars, but also as a JSON blob, and there's a fair chance some plugin is going to be (maybe even deliberately) relying on ordering within its config.

So I implemented an ordered map. It's the next evolution of yamltojson, in that it provides the same sort of order-preserving translation when you unmarshal YAML and marshal JSON, but you can also work with the contents of the map in a more map-like way. It's almost like the good old days of yaml.v2, but better!

Under the hood, ordered.Map is based on a slice of tuples, and an index (map from keys to slice indexes).

To be extra trendy, ordered.Map is generic. However, for now, unmarshaling requires string keys, and the only two uses are Map[string, string] and Map[string, any]. 

If you tell yaml.v3 to unmarshal into any, it will give you a map[string]any for mapping nodes. So any trickery is actually around co-opting UnmarshalYAML to prefer unmarshaling as a *Map[string,any] instead.

The unmarshalling tests are copied fairly closely from yamltojson, so I'm reasonably confident it's about as good as yamltojson is.

### DIFF
--- a/internal/ordered/map.go
+++ b/internal/ordered/map.go
@@ -1,0 +1,434 @@
+// Package ordered implements an ordered map type.
+package ordered
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+
+	"github.com/buildkite/agent/v3/internal/yamltojson"
+	"github.com/google/go-cmp/cmp"
+	"gopkg.in/yaml.v3"
+)
+
+var _ interface {
+	json.Marshaler
+	json.Unmarshaler
+	yaml.IsZeroer
+	yaml.Marshaler
+	yaml.Unmarshaler
+} = (*Map[string, any])(nil)
+
+// Map is an order-preserving map with string keys. It is intended for working
+// with YAML in an order-preserving way (off-spec, strictly speaking) and JSON
+// (more of the same).
+type Map[K comparable, V any] struct {
+	items []Tuple[K, V]
+	index map[K]int
+}
+
+// MapSS is a convenience alias to reduce keyboard wear.
+type MapSS = Map[string, string]
+
+// MapSA is a convenience alias to reduce keyboard wear.
+type MapSA = Map[string, any]
+
+// NewMap returns a new empty map with a given initial capacity.
+func NewMap[K comparable, V any](cap int) *Map[K, V] {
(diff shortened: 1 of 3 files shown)


internal/ordered/map.go
```diff
@@ -0,0 +1,446 @@
...
+		Key:   k,
+		Value: v,
+	})
+}
+
+// Replace replaces an old key in the same spot with a new key and value.
+// If the old key doesn't exist in the map, the item is inserted at the end.
+// If the new key already exists in the map (and isn't equal to the old key),
+// then it is deleted.
+func (m *Map[K, V]) Replace(old, new K, v V) {
```
triarius, line 107: Is the main reason this exists to avoid an unnecessary compact and allocation, compared to deleting `old` first and setting new later? I'd advertise that you can do that in the doc comments.

triarius (approved):
Brilliant as usual @DrJosh9000. I have some small suggestions, but otherwise I'm happy to merge.

internal/ordered/map.go
```diff
@@ -0,0 +1,446 @@
...
+				return err
+			}
+		}
+		om.Set(key, v)
+		return nil
+	})
+}
+
+// Tuple is used for storing values in Map.
+type Tuple[K comparable, V any] struct {
```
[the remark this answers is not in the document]
DrJosh9000: Easy!

internal/ordered/map.go
DrJosh9000: The other problem is that if you delete `old` and then set `new`, then `new` won't be in the place of `old` if it isn't already there. So it's an easy way to change a key in-place. I'll add a comment.

moskyb (approved):
@triarius said it - awesome work as usual @DrJosh9000 ^_^ no notes

PR #2174, opened 2023-06-30
Add option to configure location of Job Log tmp file

yhartanto:
Extending from this job log tmp feature https://github.com/buildkite/agent/pull/1564

We need the ability to move job log files into a different folder than tmp to share them. This will allow us to use different containers between job and runner and pre-exit hooks for example.


moskyb (changes requested):
hey, the code here LGTM, but it's failing go fmt checks -- could you run `go fmt ./...` on your machine and commit the results?

yhartanto:
@moskyb thank you sir, I updated now.

moskyb (approved):

PR #2180, opened 2023-07-03
Replace ParserResult with Pipeline

DrJosh9000:
`Pipeline` uses an `ordered.Map[string, string]` for `Env`, and a new
`ordered.Slice` for `Steps`. Unmarshaling into `ordered.Slice` prefers
`ordered.Map[string, any]` over `map[string]any`.

Why not `[]*ordered.Map[string, any]`? Because of steps like `"wait"`, which are
scalars.

`Pipeline` stores anything else in a `RemainingFields` which is an ordinary map.
I can't think of a reason to preserve ordering in fields that are not `Env` or
`Steps`, but if there is something, I would like to hear about it!

Plus a couple of related things:

* Interpolation is updated to work with the new types.
* Interpolating keys in the `Env` map is moved to `interpolateEnvBlock` and
  interpolated keys replace the old keys, rather than preserving uninterpolated keys into the output.
* Changed the parser API from a `Parser` struct taking fully-read input, to having `Parse` be a function consuming an`io.Reader` and an env map.
* Tweaked the tests to use the updated ordering for the top-level pipeline, and
  format the remarshaled JSON to make diffs easier to comb through.

### DIFF
--- a/internal/ordered/slice.go
+++ b/internal/ordered/slice.go
@@ -1,0 +1,30 @@
+package ordered
+
+import (
+	"fmt"
+
+	"gopkg.in/yaml.v3"
+)
+
+var _ yaml.Unmarshaler = (*Slice)(nil)
+
+// Slice is []any, but unmarshaling into it prefers *Map[string,any] over
+// map[string]any.
+type Slice []any
+
+// UnmarshalYAML unmarshals sequence nodes. Any mapping nodes are unmarshaled
+// as *Map[string,any].
+func (s *Slice) UnmarshalYAML(n *yaml.Node) error {
+	if n.Kind != yaml.SequenceNode {
+		return fmt.Errorf("line %d, col %d: unsupported kind %x for unmarshaling Slice (want %x)", n.Line, n.Column, n.Kind, yaml.SequenceNode)
+	}
+	seen := make(map[*yaml.Node]bool)
+	for _, c := range n.Content {
+		cv, err := decode(seen, c)
+		if err != nil {
+			return err
+		}
+		*s = append(*s, cv)
+	}
+	return nil
+}
--- a/internal/pipeline/interpolate.go
+++ b/internal/pipeline/interpolate.go
@@ -1,0 +1,134 @@
+package pipeline
+
+import (
+	"github.com/buildkite/agent/v3/internal/ordered"
+	"github.com/buildkite/interpolate"
+)
+
+// This file contains helpers for recursively interpolating all the strings in
+// pipeline objects.
+
+// selfInterpolater describes types that can interpolate themselves in-place.
+// They can call interpolate.Interpolate on strings, or
+// interpolate{Slice,Map,OrderedMap,Any} on their other contents, to do this.
+type selfInterpolater interface {
+	interpolate(interpolate.Env) error
+}
+
+// interpolateAny interpolates (almost) anything in-place. It returns the same
+// type it is passed. When passed a string, it returns a new string. Anything
+// it doesn't know how to interpolate is returned unaltered.
+func interpolateAny(env interpolate.Env, o any) (any, error) {
+	switch o := any(o).(type) {
+	case selfInterpolater:
+		return o, o.interpolate(env)
+
+	case string:
+		return interpolate.Interpolate(env, o)
+
+	case []any:
+		return o, interpolateSlice(env, o)
+
+	case []string:
+		return o, interpolateSlice(env, o)
+
+	case ordered.Slice:
+		return o, interpolateSlice(env, o)
+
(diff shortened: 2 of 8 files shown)


internal/pipeline/interpolate.go
```diff
@@ -0,0 +1,134 @@
...
+// They can call interpolate.Interpolate on strings, or
+// interpolate{Slice,Map,OrderedMap,Any} on their other contents, to do this.
+type selfInterpolater interface {
+	interpolate(interpolate.Env) error
+}
+
+// interpolateAny interpolates (almost) anything in-place. It returns the same
+// type it is passed. When passed a string, it returns a new string. Anything
+// it doesn't know how to interpolate is returned unaltered.
+func interpolateAny(env interpolate.Env, o any) (any, error) {
```
moskyb, line 21: could we make this generic? ie 
```suggestion
func interpolateAny[O any](env interpolate.Env, o O) (O, error) {
```
it would mean that we wouldn't have to do the casting further down this file - ie [here](https://github.com/buildkite/agent/pull/2180/files#diff-fc2d5994b8045c71c63083e091e94bc3edad2627c9072efd61283637b57973c5R68) we could just have `s[i] = inte`, but i wonder if Go's type system is smart enough to be disentangling all of the type casts/switches etc...

internal/pipeline/parser.go
```diff
@@ -1,184 +1,86 @@
...
-		if err != nil {
-			return nil, fmt.Errorf("%s: %v", errPrefix, err)
-		}
-		pipeline = pp
+// Parse parses a pipeline and applies interpolation. If nil envMap is provided,
+// interpolation is skipped. Otherwise, Parse mutates envMap to contain
+// variables from the pipeline's env block.
+// To interpolate only using variables defined the pipeline, pass an empty but
+// non-nil envMap.
+func Parse(src io.Reader, envMap *env.Environment) (*Pipeline, error) {
```
moskyb, line 21: i wonder if from a defensive programming perspective, the difference between an empty env and a nil env is a bit slim... perhaps we could split this into two methods:
```Go
func Parse(src io.Reader) (*Pipeline, error)
func (p *Pipeline) Interpolate(envMap *env.Environment) error
```

and then let the caller handle the decision of whether or not to interpolate.

internal/ordered/slice.go
```diff
@@ -0,0 +1,30 @@
...
+	"fmt"
+
+	"gopkg.in/yaml.v3"
+)
+
+var _ yaml.Unmarshaler = (*Slice)(nil)
+
+// Slice is []any, but unmarshaling into it prefers *Map[string,any] over
+// map[string]any.
+type Slice []any
```
moskyb, line 13: given that `ordered.Map` is generic, does it make sense to make `ordered.Slice` generic too? that way we should make the `Steps` field on the `Pipeline` struct a `ordered.Slice[Step]`, which should make using the struct a fair bit easier.

in terms of code rhyming, it makes a bit of sense as well, but it might very well be technically infeasible for whatever reason

internal/pipeline/interpolate.go
DrJosh9000: > i wonder if Go's type system is smart enough to be disentangling all of the type casts/switches etc...

It is not.

AFAIK there is presently no way to switch on `O` or `o.(type)`. You can do an ordinary type switch on an interface type, like `switch o := any(o).(type)`, and whilst `o` in each case will have the stronger type, the compiler won't see what you're doing when it gets to `return`. If you're lucky and the case is an interface, you can type assert `o.(O)`, but it rejects conversions like `O(o)`.

internal/ordered/slice.go
DrJosh9000: What I have in the other PR was `Steps`, which fingerprints each one and unmarshals the right kind of `Step`. I didn't want to get into the weeds of defining each kind of `Step` just yet, so I went with `[]any`. It felt a bit off putting something as general as an `ordered.Map`-biased `[]any` parser in the pipeline parser, but since it will be upgraded with specific step types later, that's probably where it belongs.

internal/pipeline/interpolate.go
moskyb: ah, curses. that's a drag.

internal/ordered/slice.go
moskyb: sweet, that makes total sense - i think i might not have realised that `Step` wasn't in this PR yet 😅

internal/pipeline/parser.go
DrJosh9000: Done!

moskyb (approved):
🎉 send it

PR #2181, opened 2023-07-03
Model steps and plugins

DrJosh9000:
Adds:

- a `Step` interface,
- five implementing step types,
- `Steps`, responsible for fingerprinting each step to pick the right type for each,
- `Plugin`, which represents individual plugin configs, and
- `Plugins`, which handles unmarshaling either sequences or maps.

This PR throws out most ordering for our own types (though explicitly preserving plugin config ordering). This is most of the churn in the `wantJSON` test data. Normalising plugin maps into "sequence of one-item maps" form accounts for one test. 

Because `Steps` is responsible for fingerprinting step types, I've added diff checks against typed parser output.

This also adds a `Contains` method to `ordered.Map`, and a partial unmarshaling mode to handle `ordered.Map[string, *yaml.Node]`. (It gets given the raw value nodes. Previously, it would have asked `*yaml.Node` to decode itself into `*yaml.Node`, which doesn't work.)

### DIFF
--- a/internal/ordered/map.go
+++ b/internal/ordered/map.go
@@ -76,6 +76,15 @@ func (m *Map[K, V]) Get(k K) (V, bool) {
 	return m.items[idx].Value, true
 }
 
+// Contains reports if the map contains the key.
+func (m *Map[K, V]) Contains(k K) bool {
+	if m == nil {
+		return false
+	}
+	_, has := m.index[k]
+	return has
+}
+
 // Set sets the value for the given key. If the key exists, it remains in its
 // existing spot, otherwise it is added to the end of the map.
 func (m *Map[K, V]) Set(k K, v V) {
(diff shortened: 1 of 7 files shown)


internal/ordered/map.go
```diff
@@ -301,8 +310,10 @@ func (m *Map[K, V]) UnmarshalJSON(b []byte) error {
 // UnmarshalYAML unmarshals a YAML mapping node into this map. It only supports
 // K = string. Where yaml.v3 typically infers map[string]any for unmarshaling
 // mappings into any, this method chooses *Map[string, any] instead.
+// If V = *yaml.Node, then the value nodes are not decoded. This is useful for
+// a shallow unmarshaling step.
 func (m *Map[K, V]) UnmarshalYAML(n *yaml.Node) error {
-	om, ok := (any)(m).(*Map[string, V])
```
moskyb, line 305: i'm commenting on the old diff here, but is `(any)(m)` the same as `any(m)`? it's just a regular type conversion?

moskyb (approved):
looks bloody great

internal/ordered/map.go
DrJosh9000: Yeah, I can't remember exactly but I probably had `()` in there while I was putting an older revision together.

DrJosh9000:
😄😄😄

internal/ordered/map.go
moskyb: TIL that's even allowed!

PR #2184, opened 2023-07-03
Remove yamltojson

DrJosh9000:
The only part we care about anymore is `RangeMap`, which can hide inside `ordered`. 

It would be nice to have a better solution for agent/plugin/definition.go, but this is not hugely worse than it was.

(Kill your darlings.)


moskyb (approved):
:chefskiss:

PR #2185, opened 2023-07-04
Clarify code around creating a process group

triarius:
This PR should be a no-op.

When the agent creates processes (for bootstrap, hooks or commands)
it creates a process group for the process. This is done so that
signals can be sent to the process group without affecting the process
that created the process group.

However, when launching a process in a pseudo-terminal (a PTY), there
was an error caused by explicit telling the process to be created in a
group, because the library that creates the PTY also creates a process
group (by some other means, see [here](https://github.com/creack/pty/blob/0d412c9fbeb14954aa65830dcfdb84005cd0cd96/start.go#L22)).

So we created a check for if the process was in a PTY before
doing this. But we appear to be checking this twice on non-windows
systems - both inside and outside the `setupProcessGroup` method.

So in this commit, I have removed the check from the outside and applied
it inside the method both on windows and non-windows.

Also, the only documentation of why we needed to do this was a dead link,
which I have found the new location of, and I've updated the comments to
detail enough of the contents of the dead link so that we won't miss it
if the new link goes dead.

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -127,10 +127,7 @@ func (p *Process) Run(ctx context.Context) error {
 	p.command = exec.Command(p.conf.Path, p.conf.Args...)
 
 	// Setup the process to create a process group if supported
-	// See https://github.com/kr/pty/issues/35 for context
-	if !p.conf.PTY {
-		p.setupProcessGroup()
-	}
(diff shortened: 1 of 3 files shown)


DrJosh9000 (approved):
Nice work!

PR #2186, opened 2023-07-04
Add Signature, Sign, and Verify

DrJosh9000:
The main design consideration was how to get objects to describe their default set of fields to sign, and to fetch ~arbitrary values for verifying, because Signature includes the list of signed fields. The rest pretty much flows from that.

Signatures can be attached to command steps (to begin with) and cover just the command itself. The core piece (throwing fields and values at a hash) is the same between Sign and Verify, so I implemented both.

I made the algorithm interface big enough to handle both symmetric and asymmetric schemes, but have only implemented HMAC-SHA256 here.

### DIFF
--- a/internal/pipeline/sign.go
+++ b/internal/pipeline/sign.go
@@ -1,0 +1,218 @@
+package pipeline
+
+import (
+	"bytes"
+	"crypto/hmac"
+	"crypto/sha256"
+	"encoding/base64"
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"hash"
+	"io"
+	"sort"
+)
+
+// Signature models a signature (on a step, etc).
+type Signature struct {
+	Algorithm    string   `json:"algorithm" yaml:"algorithm"`
+	SignedFields []string `json:"signed_fields" yaml:"signed_fields"`
+	Value        string   `json:"value" yaml:"value"`
+}
+
+// Sign computes a new signature for an object containing values using a given
+// signer.
+func Sign(sf SignedFielder, signer Signer) (*Signature, error) {
+	values := sf.SignedFields()
+	if len(values) == 0 {
+		return nil, errors.New("sign: no fields to sign")
+	}
+
+	// Ensure this part is the same as in Verify...
+	writeLengthPrefixed(signer, signer.AlgorithmName())
+	fields, err := writeFields(signer, values)
+	if err != nil {
+		return nil, err
+	}
+	// ...end
+
+	sig, err := signer.Sign()
+	if err != nil {
+		return nil, err
+	}
+	return &Signature{
+		Algorithm:    signer.AlgorithmName(),
+		SignedFields: fields,
+		Value:        base64.StdEncoding.EncodeToString(sig),
+	}, nil
+}
+
+// Verify verifies an existing signature against an object containing values
+// using the verifier.
+// (Verify does not create a new verifier based on the Algorithm field, in case
+// you want to use a non-standard algorithm, but it must match the verifier's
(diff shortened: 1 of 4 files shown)


internal/pipeline/sign.go
```diff
@@ -0,0 +1,186 @@
...
+	// called by Verify. The set of fields might differ from the default, e.g.
+	// when verifying older signatures computed with fewer fields or deprecated
+	// field names. signedFielder implementations should reject requests for
+	// values if "mandatory" fields are missing (e.g. signing a command step
+	// should always sign the command).
+	valuesForFields([]string) (map[string]string, error)
+}
+
+// Sign computes a new signature for a signedFielder.
+func Sign(sf signedFielder, algName string, key []byte) (*Signature, error) {
```
moskyb, line 86: i wonder if it would make more sense to have something like
```suggestion
// Sign computes a new signature for a signedFielder.
func Sign(sf signedFielder, scheme SigningScheme, key []byte) (*Signature, error) {
```

and then make the method to create a new signing scheme from an algorithm string public:
```Go
func NewSignatureScheme(algorithm string) (SignatureScheme, error) {
	switch algorithm {
	case "hmac-sha256":
		// ...
	case "rsa-pss":
		// ...
	default:
		return nil, fmt.Errorf("unsupported signing algorithm: %s", algorithm)
	}
}
```

that way we don't have to replicate the behaviour between the `Sign` and `Verify` methods, and leave deciding on the signature scheme to the caller

moskyb (changes requested):
the code here all LGTM, but i think we should put some (maybe relatively light?) testing in to generating signatures and ensuring they can be verified - if only to show what the API for using this functionality is

DrJosh9000:
> testing

Ah, I knew I was forgetting something here.

internal/pipeline/sign.go
moskyb: that said though, this is entirely stylistic, and i don't know that it makes a huge difference

internal/pipeline/sign.go
```diff
@@ -0,0 +1,186 @@
...
+	for _, f := range fields {
+		writeLV(h, f)
+		writeLV(h, values[f])
+	}
+
+	return fields, nil
+}
+
+// writeLV writes a length-prefixed string to h.
+func writeLV(h io.Writer, s string) {
```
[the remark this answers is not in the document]
DrJosh9000: Value, as in "Length Value". I'll think of a better name.

internal/pipeline/sign.go
```diff
@@ -0,0 +1,186 @@
...
+	sort.Strings(fields)
+
+	// If we blast strings at hash.Write, then you could get the same hash for
+	// different fields that happen to have the same data when concatenated.
+	// So write length-prefixed fields, and length-prefix the whole map.
+	binary.Write(h, binary.LittleEndian, uint32(len(fields)))
+	for _, f := range fields {
+		writeLV(h, f)
+		writeLV(h, values[f])
+	}
```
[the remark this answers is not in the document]
DrJosh9000: That's the idea, yeah.

internal/pipeline/sign.go
DrJosh9000: This has prompted some new thoughts...

DrJosh9000: I've taken the idea a bit further and split SignatureScheme into Signer and Verifier. Even though these are the same thing under the hood, it makes it slightly harder to reuse a signer for verifying.

moskyb (approved):
LGTM! awesome work 🎉

PR #2187, opened 2023-07-05
Die, Bootstrap, die: Rename package bootstrap -> job

moskyb:
#1958, unfortunately, became ungovernable, and got too large to reasonably work on or review.

This is the start of a series of work that will do a similar thing, but break it up into smaller, easier-to-review bits of work,

This PR specifically simply renames the package `bootstrap` to being called `job`.

**Open question:** Should we take this chance to internalise it?


DrJosh9000:
I vote for internalising it. I haven't thought too deeply, but I figure there's almost no chance of someone wanting to import it (outside the agent).

DrJosh9000 (approved):
I'm stamping this now because
- I love it
- in case you'd like to ship it and do the `internal` move later

DrJosh9000 (approved):
✅

PR #2189, opened 2023-07-06
Upgrade docker compose plugin to v4.14, use docker compose v2

moskyb:
TL;DR i saw https://github.com/buildkite/buildkite/pull/13006 and figured we should do that here too

We're using a wild number of versions of the docker compose plugin, all the way back to v1. This PR upgrades them all to the latest v4.14, and opts us into using docker compose v2


triarius (approved):
Thanks 💜

PR #2190, opened 2023-07-07
Parse command and commands the same

DrJosh9000:
The docs say `commands` is an alias for `command`. So one could write `command`’s value as multiple strings in a sequence, or write `commands`’ value as a single string.

Something that parses either single strings or sequences of strings into `[]string` is too generic for `pipeline`, so I put it in `ordered`.

### DIFF
--- a/internal/ordered/strings.go
+++ b/internal/ordered/strings.go
@@ -1,0 +1,39 @@
+package ordered
+
+import (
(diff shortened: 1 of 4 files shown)


triarius (approved):
Yeah, this is a wild one. LGTM

PR #2191, opened 2023-07-10
Fix plugin interpolation

DrJosh9000:
The bug is: `Plugins` was `[]Plugin`, but `*Plugin` was the thing that was `selfInterpolator`, so `interpolateAny` never called `*Plugin.interpolate`. Changing `Plugins` to be `[]*Plugin` neatly resolves that.

Plus various tweaks and comments on interpolation generally.

@moskyb was right about making `interpolateAny` generic, it was a matter of moving the type assertion.

### DIFF
--- a/internal/pipeline/interpolate.go
+++ b/internal/pipeline/interpolate.go
@@ -15,44 +15,58 @@ type selfInterpolater interface {
 	interpolate(interpolate.Env) error
 }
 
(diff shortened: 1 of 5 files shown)


moskyb (approved):
> @moskyb was right

as usual 💅

(this looks great btw)

PR #2192, opened 2023-07-10
Add a job log length limit

DrJosh9000:
Since [removing the need to hold the whole log in memory](https://github.com/buildkite/agent/pull/2014), there have been a few cases of long-running jobs logging lots of logs. This adds a (server-specified) limit, with a default of 1GiB. 

Why not just limit it server side? Because the job itself should get some kind of pushback once it approaches or hits the limit. Returning errors when writing to the buffer feels like the right place.

### DIFF
--- a/agent/log_streamer.go
+++ b/agent/log_streamer.go
@@ -11,12 +11,17 @@ import (
 	"github.com/buildkite/agent/v3/status"
 )
 
+const defaultLogMaxSize = 1024 * 1024 * 1024 // 1 GiB
+
 type LogStreamerConfig struct {
 	// How many log streamer workers are running at any one time
 	Concurrency int
 
-	// The maximum size of chunks
-	MaxChunkSizeBytes int
+	// The maximum size of each chunk
+	MaxChunkSizeBytes uint64
+
+	// The maximum size of the log
+	MaxSizeBytes uint64
 }
 
 type LogStreamer struct {
@@ -36,10 +41,10 @@ type LogStreamer struct {
 	queue chan *LogStreamerChunk
 
 	// Total size in bytes of the log
-	bytes int
+	bytes uint64
 
 	// Each chunk is assigned an order
-	order int
+	order uint64
 
 	// Every time we add a job to the queue, we increase the wait group
 	// queue so when the streamer shuts down, we can block until all work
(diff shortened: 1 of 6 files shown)


agent/log_streamer.go
```diff
@@ -11,6 +11,8 @@ import (
 	"github.com/buildkite/agent/v3/status"
 )
 
+const jobLogMaxSize = 1 << 30 // 1 GiB; fits in int32
```
pda, line 14: Personal opinion: I find the `1024 * 1024 * 1024` form (pronounced in my head as “kilobyte → megabyte → gigabyte”) easier to read/understand than the 1<<30 (even after lots of bit-shift-heavy side projects 😁).

The comment kinda fixes that, but then sometimes comments lie so it's nice to easily verify them.

pda (commented):
✨ 😁 ✨

agent/log_streamer.go
DrJosh9000: Makes sense, added.

agent/log_streamer.go
```diff
@@ -11,6 +11,8 @@ import (
 	"github.com/buildkite/agent/v3/status"
 )
 
+const jobLogMaxSize = 1024 * 1024 * 1024 // 1 GiB; fits in int32
```
pda, line 14: I think the server will need to tell the agent this number per-job, probably in the response to AcceptJob.
The limit might change over time, or vary by plan, etc.

DrJosh9000: Makes sense. I've added a field to `api.Job` and turned this constant into a default.

agent/log_streamer.go
```diff
@@ -97,14 +107,29 @@ func (ls *LogStreamer) Process(output []byte) error {
...
+
 	for len(output) > 0 {
+		// Have we exceeded the max size?
+		// (This check is also performed on the server side.)
+		if ls.bytes > ls.conf.MaxSizeBytes && !warnedAboutSize {
+			ls.logger.Warn("The job log has reached %s in size, which has "+
+				"exceeded the maximum size (%s). Further logs may be dropped "+
+				"by the server, and a future version of the agent will stop "+
+				"sending logs at this point.",
+				humanize.Bytes(ls.bytes), humanize.Bytes(ls.conf.MaxSizeBytes))
```
DrJosh9000, line 120: @matthewborden suggested this should just be a warning for now. Here's a cut of that.

triarius (approved):
LGTM 🎉

PR #2200, opened 2023-07-12
Add signatures

DrJosh9000:
- The pipeline can sign itself, given a signer. 
- `Signer` and `Verifier` now exposes more of `hash.Hash`, particularly so it can be `Reset`.
- pipeline upload `--dry-run` no longer requires the agent access token (which wasn't used in that path)
- pipeline upload now has a `--format` flag

Question: Is it time for `pipeline sign` ?

### DIFF
--- a/clicommand/pipeline_upload.go
+++ b/clicommand/pipeline_upload.go
@@ -20,6 +20,7 @@ import (
 	"github.com/buildkite/agent/v3/internal/redactor"
 	"github.com/buildkite/agent/v3/internal/stdin"
 	"github.com/urfave/cli"
+	"gopkg.in/yaml.v3"
 )
 
 const pipelineUploadHelpDescription = `Usage:
@@ -54,13 +55,16 @@ Example:
    $ ./script/dynamic_step_generator | buildkite-agent pipeline upload`
 
 type PipelineUploadConfig struct {
-	FilePath        string   `cli:"arg:0" label:"upload paths"`
-	Replace         bool     `cli:"replace"`
-	Job             string   `cli:"job"`
-	DryRun          bool     `cli:"dry-run"`
-	NoInterpolation bool     `cli:"no-interpolation"`
-	RedactedVars    []string `cli:"redacted-vars" normalize:"list"`
-	RejectSecrets   bool     `cli:"reject-secrets"`
+	FilePath         string   `cli:"arg:0" label:"upload paths"`
+	Replace          bool     `cli:"replace"`
+	Job              string   `cli:"job"` // required, but not in dry-run mode
+	DryRun           bool     `cli:"dry-run"`
+	DryRunFormat     string   `cli:"format"`
+	NoInterpolation  bool     `cli:"no-interpolation"`
+	RedactedVars     []string `cli:"redacted-vars" normalize:"list"`
+	RejectSecrets    bool     `cli:"reject-secrets"`
+	SigningKeyFile   string   `cli:"signing-key"`
+	SigningAlgorithm string   `cli:"signing-algorithm"`
 
 	// Global flags
 	Debug       bool     `cli:"debug"`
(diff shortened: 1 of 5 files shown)


clicommand/pipeline_upload.go
```diff
@@ -107,6 +117,17 @@ var PipelineUploadCommand = cli.Command{
 			Usage:  "When true, fail the pipeline upload early if the pipeline contains secrets",
 			EnvVar: "BUILDKITE_AGENT_PIPELINE_UPLOAD_REJECT_SECRETS",
 		},
+		cli.StringFlag{
+			Name:   "signing-key",
+			Usage:  "Path to a file containing a signing key. Passing this flag enables pipeline signing. The format of the file depends on the chosen signing algorithm",
+			EnvVar: "BUILDKITE_PIPELINE_SIGNING_KEY_FILE",
+		},
```
moskyb, line 124: a couple questions on this:
- for usability, we might wanna call this `--signing-key-file` (or `--signing-key-path`, which seems to be more standard in the agent? feels clunky to me though)
- do we want to support passing a signing key into the cli directly? eg `--signing-key "123deadbeef321"`
- the envar should probably be `BUILDKITE_PIPELINE_UPLOAD_SIGNING_KEY_FILE` (added `UPLOAD`)

clicommand/pipeline_upload.go
```diff
@@ -107,6 +117,17 @@ var PipelineUploadCommand = cli.Command{
...
+		cli.StringFlag{
+			Name:   "signing-algorithm",
+			Usage:  "Signing algorithm to use when signing parts of the pipeline. Available algorithms are: hmac-sha256",
+			Value:  "hmac-sha256",
```
moskyb, line 128: given that we're planning on adding asymmetric keys to the signing feature, it might be good to not default to anything from the start?

clicommand/pipeline_upload.go
```diff
@@ -107,6 +117,17 @@ var PipelineUploadCommand = cli.Command{
...
+			EnvVar: "BUILDKITE_PIPELINE_SIGNING_ALGORITHM",
```
moskyb, line 129: same as above:
```suggestion
			EnvVar: "BUILDKITE_PIPELINE_UPLOAD_SIGNING_ALGORITHM",
```

clicommand/pipeline_upload.go
```diff
@@ -287,14 +308,47 @@ var PipelineUploadCommand = cli.Command{
 			}
 		}
 
-		// In dry-run mode we just output the generated pipeline to stdout
+		if cfg.SigningKeyFile != "" {
```
moskyb, line 311: it might be good to add a dragons warning here - "this is experimental and will change! it might not work!" etc

clicommand/pipeline_upload.go
```diff
@@ -287,14 +308,47 @@ var PipelineUploadCommand = cli.Command{
...
+				enc := json.NewEncoder(os.Stdout)
+				enc.SetIndent("", "  ")
+				encode = enc.Encode
+
+			case "yaml":
+				encode = yaml.NewEncoder(os.Stdout).Encode
+
+			default:
+				l.Fatal("Unknown output format %q", cfg.DryRunFormat)
+			}
```
moskyb, line 346: 🎉 love it

moskyb (changes requested):

clicommand/pipeline_upload.go
DrJosh9000: Hmm, yeah, good point

internal/pipeline/pipeline.go
```diff
@@ -148,3 +148,22 @@ func (p *Pipeline) interpolateEnvBlock(envMap *env.Environment) error {
...
+		}
+
+		sig, err := Sign(cmd, signer)
+		if err != nil {
+			return err
+		}
+		cmd.Signature = sig
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
DrJosh9000: Sounds good. For now, because an error means no upload/output, I've leaned towards modifying in-place instead of building up a `newSteps`, and documenting that.

clicommand/pipeline_upload.go
DrJosh9000: - Consistency sounds good
- One danger with passing keys via flag is that keys could end up in shell history or process info (`ps -x`), but maybe that's not a problem?
- Sounds good

moskyb (approved):

PR #2203, opened 2023-07-13
Update URL for remote: look before leap

DrJosh9000:
I'm not certain yet, but the submodule-mirror-wrong-remote bug in older agents (see #2144) _might_ be related to recent "bad object" / "did not send all necessary objects" issues.

For submodules, older agents set the wrong remote URL in the directory. v3.49 (#2144) fixed it to set the correct remote. However, once upgraded to v3.49, the mirror dir is probably broken: the remote URL is updated correctly, but the local repo would still contain objects for the wrong remote. Afterwards, the repo is only fetched if it is the main repo. Woe betide you if you have multiple agents of different versions fighting over the mirror!

So this PR:

- uses `get-url` before using `set-url`, so that we can log a message
- if the remote URL changed, do a `fetch` for non-main repos to try to get out of bad object situations

### DIFF
--- a/internal/job/executor.go
+++ b/internal/job/executor.go
@@ -1327,9 +1327,10 @@ func (e *Executor) updateGitMirror(ctx context.Context, repository string) (stri
 
 	e.shell.Commentf("Updating existing repository mirror to find commit %s", e.Commit)
 
-	// Update the origin of the repository so we can gracefully handle repository renames
-	if err := e.shell.Run(ctx, "git", "--git-dir", mirrorDir, "remote", "set-url", "origin", repository); err != nil {
-		return "", err
+	// Update the origin of the repository so we can gracefully handle
+	// repository renames.
+	if err := e.updateRemoteURL(ctx, mirrorDir, repository); err != nil {
+		return "", fmt.Errorf("setting remote URL: %w", err)
 	}
 
 	if isMainRepository {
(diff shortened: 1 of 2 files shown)


internal/job/executor.go
```diff
@@ -1346,14 +1348,62 @@ func (e *Executor) updateGitMirror(ctx context.Context, repository string) (stri
...
+	}
+
+	gd := gitDir
+	if gd == "" {
+		gd = e.shell.Getwd()
+	}
+
+	e.shell.Commentf("Remote URL for git directory %s has changed (%s -> %s)!", gd, gotURL, repository)
+	e.shell.Commentf("This is usually because the repository has been renamed.")
+	e.shell.Commentf("If this is unexpected, you may see failures (e.g. 'bad object', 'did not send all necessary objects').")
```
triarius, line 1396: Nice message, I hope we don't need it 🤞

internal/job/executor.go
```diff
@@ -1365,13 +1415,10 @@ func (e *Executor) getOrUpdateMirrorDir(ctx context.Context, repository string)
...
-	} else {
-		mirrorDir, err = e.updateGitMirror(ctx, repository)
-		if err != nil {
-			return "", err
-		}
+		return mirrorDir, nil
 	}
-	return mirrorDir, nil
+
+	return e.updateGitMirror(ctx, repository)
```
triarius, line 1421: 🙏

moskyb (approved):

DrJosh9000:
I had another read-through of what it's doing, and it appears that, for mirrored submodules, nothing actually updates them, either in `updateGitMirror` or in the mirrored submodule loop. 

So for submodule mirrors, I think we need to fetch whether or not there was a rename, and this is might be why there are missing objects even on agents at v3.49 for a while. I get the feeling `--reference` is some subtle magic.

If a fetch brings in a repo with no common ancestry to what was there (say, due to the bug) then its probably worth `git gc` after fetch. Maybe `git fsck`? Arguably that should happen for the main repo too.

PR #2204, opened 2023-07-13
Artifacts package

moskyb:
Artifacts are very much their own thing, and take up a lot of space in the `agent` module without actually being used in it anywhere at all.

This PR takes all of the artifact gear out of the `agent` module, and puts it in its own (internal) module called `artifact`. It also renames some of the types and funcs within to remove [stuttering](https://michaelwhatcott.com/go-code-that-stutters/).

### DIFF
--- a/agent/aws.go
+++ b/agent/aws.go
@@ -13,7 +13,7 @@ var awsSess *session.Session
 // The aws sdk relies on being given a region, which is a breaking change for us
 // This applies a heuristic that detects where the agent might be based on the env
 // but also the local isntance metadata if available
-func awsRegion() (string, error) {
+func AWSRegion() (string, error) {
 	if r := os.Getenv("AWS_REGION"); r != "" {
 		return r, nil
 	}
@@ -39,7 +39,7 @@ func awsRegion() (string, error) {
 }
 
 func awsSession() (*session.Session, error) {
-	region, err := awsRegion()
+	region, err := AWSRegion()
 	if err != nil {
 		return nil, err
 	}
--- a/internal/artifact/uploader.go
+++ b/internal/artifact/uploader.go
@@ -1,4 +1,4 @@
-package agent
+package artifact
 
 import (
 	"github.com/buildkite/agent/v3/api"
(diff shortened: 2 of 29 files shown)


internal/artifact/uploader.go
```diff
@@ -1,10 +1,10 @@
-package agent
+package artifact
 
 import (
 	"github.com/buildkite/agent/v3/api"
 )
 
-type Uploader interface {
+type UploadDoer interface {
```
moskyb, line 7: taking any and all suggestions on a better name for this

surely this can't be it

moskyb: (there's a namespace collision with `artifact.Uploader`, which is the struct equivalent to `artifact.Downloader`, `.Searcher` etc)

triarius: I think it's more conventional to reserve the agent noun (ie the ones that end in "er") for interfaces and rename the structs.

They could be something like `artifact.DownloadRequest` and `artifact.UploadRequest` as I think they correspond to a request to the API.

I guess that means you should rename about half the structs in this package, but that's what refactors like this are for!

agent/aws.go
```diff
@@ -39,7 +39,7 @@ func awsRegion() (string, error) {
 }
 
 func awsSession() (*session.Session, error) {
-	region, err := awsRegion()
+	region, err := AWSRegion()
```
triarius, line 42: This really belongs somewhere else, but that's not for this PR 🤷‍♂️

internal/artifact/uploader.go
triarius: Or you could just combine the method names and agent nounifiy it: `URLUploader`, or `UploadURLer`. That's the approach `io.ReadWriter` takes.

triarius (approved):
Generally LGTM. Personally, I think more interfaces are needed here, but I'm happy to make incremental changes.

agent/aws.go
moskyb: yeah, agreed - i took a look at it and thought "a problem for another day"

DrJosh9000 (approved):

moskyb:
i'm gonna close this - the conflicts have grown too powerful in the time since i opened it. i'd still like to get it done at some point tho

PR #2205, opened 2023-07-14
Use a unique name for each agent started using the systemd template unit file

DavidGregory084:
Relates to https://github.com/buildkite/docs/issues/2252.

The docs currently advise users that they can [run multiple agents using the same configuration](https://buildkite.com/docs/agent/v3/ubuntu#running-multiple-agents) by using the systemd template unit file.

However, this results in starting multiple agents with the same name, which the [configuration docs](https://buildkite.com/docs/agent/v3/configuration#name) advise against doing.

This PR uses [unit file specifiers](https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Specifiers) to give each agent a unique name based on the hostname and the systemd service instance name.


DrJosh9000:
Thanks @DavidGregory084! Looks good to me.

DrJosh9000 (approved):

PR #2206, opened 2023-07-16
Simplify command phase

triarius:
In preparation for some other work I intend to do in this area, simplify the CommandPhase method.

I aimed to make this PR a no-op, but technically, I've changed the order of some of the statements, which does change the internal state and the order of some log statements. But there should be minimal externally visible effects.

I recommend you review this PR with a side-by-side diff.

### DIFF
--- a/internal/job/executor.go
+++ b/internal/job/executor.go
@@ -1717,48 +1717,62 @@ func (e *Executor) runPostCommandHooks(ctx context.Context) error {
 }
 
 // CommandPhase determines how to run the build, and then runs it
-func (e *Executor) CommandPhase(ctx context.Context) (error, error) {
+func (e *Executor) CommandPhase(ctx context.Context) (hookErr error, commandErr error) {
+	var preCommandErr error
+
 	span, ctx := tracetools.StartSpanFromContext(ctx, "command", e.ExecutorConfig.TracingBackend)
-	var err error
-	defer func() { span.FinishWithError(err) }()
-	// Run pre-command hooks
-	if err := e.runPreCommandHooks(ctx); err != nil {
-		return err, nil
-	}
+	defer func() {
+		span.FinishWithError(hookErr)
+	}()
 
-	// Run the actual command
-	commandExitError := e.runCommand(ctx)
-	var realCommandError error
-
-	// If the command returned an exit that wasn't a `exec.ExitError`
-	// (which is returned when the command is actually run, but fails),
-	// then we'll show it in the log.
-	if shell.IsExitError(commandExitError) {
-		if shell.IsExitSignaled(commandExitError) {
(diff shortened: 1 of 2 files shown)


internal/job/executor.go
```diff
@@ -1717,48 +1718,59 @@ func (e *Executor) runPostCommandHooks(ctx context.Context) error {
...
 // CommandPhase determines how to run the build, and then runs it
-func (e *Executor) CommandPhase(ctx context.Context) (error, error) {
+func (e *Executor) CommandPhase(ctx context.Context) (hookErr error, commandError error) {
+	var preCommandErr error
+
 	span, ctx := tracetools.StartSpanFromContext(ctx, "command", e.ExecutorConfig.TracingBackend)
-	var err error
-	defer func() { span.FinishWithError(err) }()
+	defer func() {
+		span.FinishWithError(hookErr)
```
triarius, line 1726: The span includes the command error if it is not nil when this method is returned. But, if there is both a post-command hook error and a command error, the post-command hook error would have finished the span. This seems like a bug. But as I'm aiming for a no-op refactor. I don't want to address that here.

internal/job/executor.go
```diff
@@ -1717,48 +1718,64 @@ func (e *Executor) runPostCommandHooks(ctx context.Context) error {
...
+	// Exit early if there was no error
+	if commandErr == nil {
+		return nil, nil
 	}
 
-	// Save the command exit status to the env so hooks + plugins can access it. If there is no error
-	// this will be zero. It's used to set the exit code later, so it's important
-	e.shell.Env.Set("BUILDKITE_COMMAND_EXIT_STATUS", fmt.Sprintf("%d", shell.GetExitCode(commandExitError)))
+	// Expand the job log header from the command to surface the error
+	e.shell.Printf("^^^ +++")
```
triarius, line 1759: This used to happen after then error logs. I don't think it will make a difference to the logs after terminal-to-html is applied. In both cases, the log group of the command is expanded. But obviously, some lines in the raw log will be permuted.

moskyb: i initially thought that there was a change in behaviour here where if pre-command hooks failed, their log groups wouldn't expand, but it seems like that was the case before anyway... kinda odd, but agree that a noop refactor is a good idea.

moskyb (commented):
🎉 awesome work, this is way easier to read

moskyb (approved):
oh whoops, forgot to ✅

internal/job/executor.go
triarius: Yeah, there's a few quirks like this. Probably crept in because the old code was hard to reason with.

DrJosh9000 (commented):
Nice work!

PR #2207, opened 2023-07-17
JobRunner cleanup

moskyb:
in advance of implementing job signature verification, this PR does a little bit of no-op cleanup on the JobRunner

### DIFF
--- a/agent/agent_worker.go
+++ b/agent/agent_worker.go
@@ -647,10 +647,15 @@ func (a *AgentWorker) RunJob(ctx context.Context, acceptResponse *api.Job) error
 	})
 
 	// Now that we've got a job to do, we can start it.
-	jr, err := NewJobRunner(a.logger, jobMetricsScope, a.agent, acceptResponse, a.apiClient, JobRunnerConfig{
+	jr, err := NewJobRunner(JobRunnerConfig{
+		Job:                acceptResponse,
+		APIClient:          a.apiClient,
+		Logger:             a.logger,
 		Debug:              a.debug,
 		DebugHTTP:          a.debugHTTP,
 		CancelSignal:       a.cancelSig,
+		MetricsScope:       jobMetricsScope,
+		JobStatusInterval:  time.Duration(a.agent.JobStatusInterval) * time.Second,
 		AgentConfiguration: a.agentConfiguration,
 		AgentStdout:        a.agentStdout,
 	})
--- a/agent/agent_worker.go
+++ b/agent/agent_worker.go
@@ -647,10 +647,8 @@ func (a *AgentWorker) RunJob(ctx context.Context, acceptResponse *api.Job) error
 	})
 
 	// Now that we've got a job to do, we can start it.
-	jr, err := NewJobRunner(JobRunnerConfig{
+	jr, err := NewJobRunner(a.logger, a.apiClient, JobRunnerConfig{
 		Job:                acceptResponse,
-		APIClient:          a.apiClient,
-		Logger:             a.logger,
 		Debug:              a.debug,
 		DebugHTTP:          a.debugHTTP,
 		CancelSignal:       a.cancelSig,
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -75,6 +75,21 @@ type JobRunnerConfig struct {
 	// The configuration of the agent from the CLI
 	AgentConfiguration AgentConfiguration
 
+	// The logger to use
+	Logger logger.Logger
+
+	// How often to check if the job has been cancelled
+	JobStatusInterval time.Duration
+
+	// A scope for metrics within a job
+	MetricsScope *metrics.Scope
+
+	// The job to run
+	Job *api.Job
+
+	// The APIClient that will be used when updating the job
+	APIClient APIClient
+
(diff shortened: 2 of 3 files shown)


agent/job_runner.go
```diff
@@ -326,6 +319,12 @@ func NewJobRunner(l logger.Logger, scope *metrics.Scope, ag *api.AgentRegisterRe
 			ClientCount: containerCount,
```
triarius, line 334: Happy to get this k8s stuff outta here, but probably best to do that when me remove the "experiment".

agent/job_runner.go
```diff
@@ -390,29 +389,26 @@ func (r *JobRunner) Run(ctx context.Context) error {
...
-	exitStatus := ""
-	signal := ""
-	signalReason := ""
+	exit := processExit{}
 
 	// Before executing the bootstrap process with the received Job env,
 	// execute the pre-bootstrap hook (if present) for it to tell us
 	// whether it is happy to proceed.
-	environmentCommandOkay := true
+	shouldRunJob := true
```
triarius, line 412: Love this rename

agent/agent_worker.go
```diff
@@ -647,10 +647,15 @@ func (a *AgentWorker) RunJob(ctx context.Context, acceptResponse *api.Job) error
 	})
 
 	// Now that we've got a job to do, we can start it.
-	jr, err := NewJobRunner(a.logger, jobMetricsScope, a.agent, acceptResponse, a.apiClient, JobRunnerConfig{
+	jr, err := NewJobRunner(JobRunnerConfig{
+		Job:                acceptResponse,
+		APIClient:          a.apiClient,
+		Logger:             a.logger,
```
triarius, line 653: I'm not a fan of having these in the config. IMO, config should be serialisable. It is a lot easier to test functions with serialisable inputs and outputs.

I get that `a.agentStdout` isn't serialisable either, but we shouldn't double down. Rather, it would be great if you could separate the serialisable and not serialisable parts.

agent/job_runner.go
```diff
@@ -149,32 +161,35 @@ type jobAPI interface {
...
+func NewJobRunner(conf JobRunnerConfig) (jobRunner, error) {
+	r := &JobRunner{
+		job:       conf.Job,
+		logger:    conf.Logger,
 		conf:      conf,
-		metrics:   scope,
-		apiClient: apiClient,
+		metrics:   conf.MetricsScope,
+		apiClient: conf.APIClient,
+	}
```
triarius, line 171: I don't like the redundancy of having these available in two places, inside conf and outside conf.

agent/job_runner.go
```diff
@@ -100,9 +115,6 @@ type JobRunner struct {
 	// The logger to use
 	logger logger.Logger
 
-	// The registered agent API record running this job
-	agent *api.AgentRegisterResponse
-
```
triarius, line 105: I guess this was not needed?

agent/job_runner.go
```diff
@@ -75,6 +75,21 @@ type JobRunnerConfig struct {
 	// The configuration of the agent from the CLI
 	AgentConfiguration AgentConfiguration
 
+	// The logger to use
+	Logger logger.Logger
```
triarius, line 79: If there's already logger in `JobRunner`, which contains `JobRunnerConfig`, why is there also one in `JobRunnerConfig`. I think we need to chose definitions of the structs where this sort of thing does not happen.

agent/job_runner.go
```diff
@@ -514,16 +484,66 @@ func (r *JobRunner) Run(ctx context.Context) error {
...
 	return nil
 }
 
+type processExit struct {
+	Status       string
+	Signal       string
+	SignalReason string
+}
+
+func (r *JobRunner) runJob(ctx context.Context) (exit processExit) {
```
triarius, line 501: Not sure if it's a huge no-no to put methods in their own file, but otherwise, I would love for these to be in a `run_job.go` file. We need to cut down on the length of this one.

triarius (commented):
Looking good, I have some initial thoughts

agent/job_runner.go
DrJosh9000: IMO, in the year 2023 CE, methods can totally go in their own files. We have multi-file search and "Jump to Definition". (Interleaving methods from different types is still bad though.)

agent/job_runner.go
moskyb: we pulled in the entire register response just to get the interval for how often we should check to see that if the job's been cancelled; i figured we can just pull that in directly

agent/agent_worker.go
moskyb: yeah, i wavered a little mid implementation thinking the same thing, and tentatively came down on the other side, but i think you're right.

is your preferred signature
```Go
NewThingo(a NonSerializable, b NonSerializable, cfg ThingoConfig)
```

agent/job_runner.go
moskyb: i might handle this in the verify PR... merge conflicts are going to be annoying otherwise

triarius (approved):
LGTM 🎉

DrJosh9000 (approved):
✅

PR #2209, opened 2023-07-17
Add experiment to avoid a recursive trap

triarius:
Some jobs are run as a bash script of the form:
```bash
trap "kill -- $$" INT TERM QUIT; <command>
```
We now understand this causes a bug (see the code comments), and we want to avoid it.

This PR adds an experiment to avoid adding the trap to the script.

### DIFF
--- a/internal/job/executor.go
+++ b/internal/job/executor.go
@@ -1976,18 +1976,20 @@ func (e *Executor) defaultCommandPhase(ctx context.Context) error {
 		return err
 	}
 
-	// Update: The comment below is of dubious validity, because:
-	//         1. The trap is recursive and leads to an infinite loop of signals and a segfault.
-	//         2. It just signals the intermediate shell again. If the intermediate shell swallowed
-	//            signals in the first place then signaling it again won't change that.
-	//         3. Propogating signals to child processes is handled by signalling their process
-	//            group.
-	// Therefore, we are phasing it out under an experiment.
-	//
+	// We added the `trap` below because we used to think that:
 	//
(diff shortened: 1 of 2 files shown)


internal/job/executor.go
```diff
@@ -1966,10 +1976,19 @@ func (e *Executor) defaultCommandPhase(ctx context.Context) error {
 		return err
 	}
 
+	// Update: The comment below is of dubious validity, because:
```
DrJosh9000, line 1979: Might be better to incorporate or replace the dubious comment...

```go
// The below is the "recursive trap". 
// The original idea was to ...
// But ..., so we are phasing it out under an experiment.
```

internal/job/executor.go
```diff
@@ -1828,9 +1828,19 @@ func (e *Executor) CommandPhase(ctx context.Context) (hookErr error, commandErr
...
 
 	switch {
-	case isExitError && isExitSignaled:
+	case isExitError && isExitSignaled && avoidRecursiveTrap:
+		// The recursive trap created a segfault that we were previously inadvertently suppressing
+		// in the next branch. Once the experiment is promoted, we should keep this branch in case
+		// to show the error to users.
+		e.shell.Errorf("The command was interrupted by a signal: %v", commandErr)
+
+		// although error is an exit error, it's not returned. (seems like a bug)
```
DrJosh9000, line 1840: `(seems like a bug)` seems like a bug.

DrJosh9000 (approved):
Ship it ✅

PR #2210, opened 2023-07-17
Verify step signatures

moskyb:
**This PR:** Enables the buildkite-agent to verify incoming jobs against signatures uploaded in their pipeline uploads.

This is done by a process (vaguely):
- When accepting a job, the `Job` sent by the backend will (WIP) now include a `step` attribute, which contains a subset of information contained in the step upload - at current, it's the step's command and signature, though as we sign more elements of the signature, this will expand to include other elements
- After accepting the job, but before kicking off the bootstrap process, the agent computes a signature for the `step`, and compares it to the signature shipped in the accept payload
- With the step verified, we need to check that the `Job`'s elements match the `Step`'s
- If they match, run the job
- If they don't (or there's a signature mismatch, or any other error happens during this process), fail the job with exit status -1, and an informational signal reason - this is all handled very similarly to the existing `pre-bootstrap` hook

### DIFF
--- a/agent/integration/test_helpers.go
+++ b/agent/integration/test_helpers.go
@@ -39,6 +39,8 @@ func mockBootstrap(t *testing.T) *bintest.Mock {
 }
 
 func runJob(t *testing.T, j *api.Job, server *httptest.Server, cfg agent.AgentConfiguration, bs *bintest.Mock) {
+	t.Helper()
+
 	l := logger.Discard
 
 	// minimal metrics, this could be cleaner
--- a/agent/integration/test_helpers.go
+++ b/agent/integration/test_helpers.go
@@ -1,0 +1,170 @@
+package integration
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"sync"
+	"testing"
+
+	"github.com/buildkite/agent/v3/agent"
+	"github.com/buildkite/agent/v3/api"
+	"github.com/buildkite/agent/v3/logger"
+	"github.com/buildkite/agent/v3/metrics"
+	"github.com/buildkite/bintest/v3"
+)
+
+func mockBootstrap(t *testing.T) *bintest.Mock {
+	t.Helper()
+
+	// tests run using t.Run() will have a slash in their name, which will mess with paths to bintest binaries
+	name := strings.ReplaceAll(t.Name(), "/", "-")
+	bs, err := bintest.NewMock(fmt.Sprintf("buildkite-agent-bootstrap-%s", name))
+	if err != nil {
+		t.Fatalf("bintest.NewMock() error = %v", err)
+	}
+
+	return bs
+}
+
+func runJob(t *testing.T, j *api.Job, server *httptest.Server, cfg agent.AgentConfiguration, bs *bintest.Mock) {
+	l := logger.Discard
+
+	// minimal metrics, this could be cleaner
+	m := metrics.NewCollector(l, metrics.CollectorConfig{})
+	scope := m.Scope(metrics.Tags{})
+
+	// set the bootstrap into the config
+	cfg.BootstrapScript = bs.Path
+
+	client := api.NewClient(l, api.Config{
+		Endpoint: server.URL,
+		Token:    "llamasrock",
+	})
+
+	jr, err := agent.NewJobRunner(l, client, agent.JobRunnerConfig{
+		Job:                j,
+		AgentConfiguration: cfg,
+		MetricsScope:       scope,
+	})
+	if err != nil {
+		t.Fatalf("agent.NewJobRunner() error = %v", err)
+	}
+
+	if err := jr.Run(context.Background()); err != nil {
+		t.Errorf("jr.Run() = %v", err)
+	}
+}
+
+type testAgentEndpoint struct {
+	calls map[string][][]byte
+	mtx   sync.Mutex
+}
+
+func createTestAgentEndpoint() *testAgentEndpoint {
+	return &testAgentEndpoint{
+		calls: make(map[string][][]byte, 4),
+	}
+}
+
+func (tae *testAgentEndpoint) finishesFor(t *testing.T, jobID string) []api.Job {
+	t.Helper()
+	tae.mtx.Lock()
+	defer tae.mtx.Unlock()
+
+	endpoint := fmt.Sprintf("/jobs/%s/finish", jobID)
+	finishes := make([]api.Job, 0, len(tae.calls))
+
+	for _, b := range tae.calls[endpoint] {
+		var job api.Job
+		err := json.Unmarshal(b, &job)
+		if err != nil {
+			t.Fatalf("decoding accept request body: %v", err)
+		}
+		finishes = append(finishes, job)
+	}
+
+	return finishes
+}
+
+func (tae *testAgentEndpoint) chunksFor(t *testing.T, jobID string) []api.Chunk {
+	t.Helper()
+	tae.mtx.Lock()
+	defer tae.mtx.Unlock()
+
+	endpoint := fmt.Sprintf("/jobs/%s/chunks", jobID)
+	chunks := make([]api.Chunk, 0, len(tae.calls))
+
+	for _, b := range tae.calls[endpoint] {
+		var chunk api.Chunk
+		err := json.Unmarshal(b, &chunk)
+		if err != nil {
+			t.Fatalf("decoding accept request body: %v", err)
+		}
+		chunks = append(chunks, chunk)
+	}
+
+	return chunks
+}
+
+func (t *testAgentEndpoint) server(jobID string) *httptest.Server {
+	return httptest.NewServer(http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+		t.mtx.Lock()
+		defer t.mtx.Unlock()
+
+		b, _ := io.ReadAll(req.Body)
+		t.calls[req.URL.Path] = append(t.calls[req.URL.Path], b)
+
+		switch req.URL.Path {
+		case "/jobs/" + jobID:
+			rw.WriteHeader(http.StatusOK)
+			fmt.Fprintf(rw, `{"state":"running"}`)
+		case "/jobs/" + jobID + "/start":
+			rw.WriteHeader(http.StatusOK)
+		case "/jobs/" + jobID + "/chunks":
+			rw.WriteHeader(http.StatusCreated)
+		case "/jobs/" + jobID + "/finish":
+			rw.WriteHeader(http.StatusOK)
+		default:
+			http.Error(rw, fmt.Sprintf("not found; method = %q, path = %q", req.Method, req.URL.Path), http.StatusNotFound)
+		}
+	}))
+}
+
+func mockPreBootstrap(t *testing.T, hooksDir string) *bintest.Mock {
+	t.Helper()
+
+	mock, err := bintest.NewMock(fmt.Sprintf("buildkite-agent-pre-bootstrap-hook-%s", t.Name()))
+	if err != nil {
+		t.Fatalf("bintest.NewMock() error = %v", err)
+	}
+
+	hookScript := filepath.Join(hooksDir, "pre-bootstrap")
+	body := ""
+
+	if runtime.GOOS == "windows" {
+		// You may be tempted to change this to `@%q`, but please do not. bintest doesn't like it when things change.
+		// (%q escapes backslashes, which are windows path separators and leads to this test failing on windows)
+		body = fmt.Sprintf(`@"%s"`, mock.Path)
+		hookScript += ".bat"
+	} else {
+		body = "#!/bin/sh\n" + mock.Path
+	}
+
+	if err := os.MkdirAll(hooksDir, 0o700); err != nil {
+		t.Fatalf("creating pre-bootstrap hook mock: os.MkdirAll() error = %v", err)
+	}
+
+	if err := os.WriteFile(hookScript, []byte(body), 0o777); err != nil {
+		t.Fatalf("creating pre-bootstrap hook mock: s.WriteFile() error = %v", err)
+	}
+
+	return mock
+}
--- a/agent/integration/test_helpers.go
+++ b/agent/integration/test_helpers.go
@@ -1,6 +1,8 @@
 package integration
 
 import (
+	"bytes"
+	"compress/gzip"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -10,6 +12,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strconv"
 	"strings"
 	"sync"
 	"testing"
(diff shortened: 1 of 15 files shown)


agent/run_job.go
```diff
@@ -85,7 +175,52 @@ func (r *JobRunner) Run(ctx context.Context) error {
 	go r.jobLogStreamer(cctx, &wg)
 	go r.jobCancellationChecker(cctx, &wg)
 
-	exit = r.runJob(cctx) // Ignore gostaticcheck here, the value of exit is captured by the deferred cleanup function above
+	exit = r.runJob(cctx)
+
+	return nil
+}
+
+func (r *JobRunner) verifyJob(verifier pipeline.Verifier) error {
```
triarius, line 183: Time for another file?

agent/run_job.go
```diff
@@ -85,7 +175,52 @@ func (r *JobRunner) Run(ctx context.Context) error {
...
+	step := r.conf.Job.Step
+
+	if _, ok := step.RemainingFields["matrix"]; ok {
+		r.logger.Warn("Signing/Verification of matrix jobs is not currently supported")
+		r.logger.Warn("Watch this space 👀")
+
+		return nil
```
triarius, line 190: Nice

triarius (commented):
Looking good.

I think all my comments are just various degrees of nits.

But I would encourage exploring what putting this stuff in a different package would look like.

agent/integration/test_helpers.go
```diff
@@ -0,0 +1,167 @@
...
+
+		switch req.URL.Path {
+		case "/jobs/" + jobID:
+			rw.WriteHeader(http.StatusOK)
+			fmt.Fprintf(rw, `{"state":"running"}`)
+		case "/jobs/" + jobID + "/start":
+			rw.WriteHeader(http.StatusOK)
+		case "/jobs/" + jobID + "/chunks":
+			req.
+				rw.WriteHeader(http.StatusCreated)
```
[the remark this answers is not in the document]
moskyb: oh, fully just committed halfway through writing some tests, should be fixed now (and replaced with genuinely some of the ugliest code i've ever written)

agent/integration/test_helpers.go
```diff
@@ -0,0 +1,171 @@
...
+			fmt.Fprintf(rw, `{"state":"running"}`)
+		case "/jobs/" + jobID + "/start":
+			rw.WriteHeader(http.StatusOK)
+		case "/jobs/" + jobID + "/chunks":
+			sequence := req.URL.Query().Get("sequence")
+			seqNo, _ := strconv.Atoi(sequence)
+			r, _ := gzip.NewReader(bytes.NewBuffer(b))
+			uz, _ := io.ReadAll(r)
+			t.logChunks[seqNo] = string(uz)
+			rw.WriteHeader(http.StatusCreated)
```
moskyb, line 133: error handling? what's that?

agent/run_job.go
```diff
@@ -85,7 +175,52 @@ func (r *JobRunner) Run(ctx context.Context) error {
 	go r.jobLogStreamer(cctx, &wg)
 	go r.jobCancellationChecker(cctx, &wg)
 
-	exit = r.runJob(cctx) // Ignore gostaticcheck here, the value of exit is captured by the deferred cleanup function above
+	exit = r.runJob(cctx)
```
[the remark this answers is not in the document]
moskyb: hiding my shame

agent/run_job.go
```diff
@@ -2,18 +2,45 @@ package agent
...
+	return fmt.Sprintf("invalid signature: %v", e.underlying)
+}
+
+func (e *invalidSignatureError) Unwrap() error {
+	return e.underlying
+}
+
+func (e *invalidSignatureError) Is(target error) bool {
+	_, ok := target.(*invalidSignatureError)
+	return ok
```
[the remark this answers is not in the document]
moskyb: ooh, yes, good call - i always get confused between `Is` and `As`

agent/run_job.go
```diff
@@ -62,6 +102,56 @@ func (r *JobRunner) Run(ctx context.Context) error {
...
+
+	default: // some other error
+		r.logger.Warn("Job verification failed with error: %s", err.Error())
+		r.logStreamer.Process([]byte(fmt.Sprintf("⚠️ WARNING: Job verification failed with error: %s\n", err.Error())))
+		exit.Status = -1
+		exit.SignalReason = "job_verification_failed_with_error"
+
+		return nil
+
+	}
```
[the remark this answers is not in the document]
moskyb: i've cleaned this block up significantly, and made it way easier to read and less repetitive - lemme know what you think of the new implementation

agent/run_job.go
moskyb: good call, donezo

agent/run_job.go
```diff
@@ -23,6 +50,19 @@ func (r *JobRunner) Run(ctx context.Context) error {
...
+		if err != nil {
+			return fmt.Errorf("failed to read job verification key: %w", err)
+		}
+
+		verifier, err = pipeline.NewVerifier("hmac-sha256", verificationKey)
+		if err != nil {
+			return fmt.Errorf("failed to create job verifier: %w", err)
+		}
+	}
+
```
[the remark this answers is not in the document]
moskyb: this is fixed in the JWX implementation, i might leave it until then if that's all good?

DrJosh9000 (commented):
Good stuff ...but still reviewing...

agent/verify_job.go
```diff
@@ -0,0 +1,95 @@
...
+import (
+	"errors"
+	"fmt"
+
+	"github.com/buildkite/agent/v3/internal/pipeline"
+)
+
+var ErrNoSignature = errors.New("job had no signature to verify")
+
+type invalidSignatureError struct {
```
DrJosh9000, line 12: When we get to signature schemes that separate message digest and signature, it might be helpful (for the poor saps diagnosing signing failures) to separate "digest didn't match message" and "the digest matched the message but the signature isn't valid for the digest" - both cases seem like "invalidSignatureError" though so I don't know if I'm arguing for any change here.

agent/run_job.go
```diff
@@ -23,6 +27,19 @@ func (r *JobRunner) Run(ctx context.Context) error {
 
 	r.startedAt = time.Now()
```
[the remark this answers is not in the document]
moskyb: ah, this gets passed into the `startJob` call that we send to the agent API, so this has to be here i think

agent/verify_job.go
```diff
@@ -0,0 +1,95 @@
...
+
+	if verifier == nil {
+		if step.Signature != nil {
+			return fmt.Errorf("job %q was signed with signature %q, but no verification key was provided, so the job can't be verified", r.conf.Job.ID, step.Signature.Value)
+		}
+
+		return nil // no signature, no verifier, no problem
+	}
+
+	if _, ok := step.RemainingFields["matrix"]; ok {
```
DrJosh9000, line 59: Since we need a semantic understanding of matrix vs non-matrix, perhaps `CommandStep` should grow a `Matrix` field (of type `any` I guess)?

clicommand/agent_start.go
```diff
@@ -34,6 +34,7 @@ import (
 	"github.com/mitchellh/go-homedir"
 	"github.com/urfave/cli"
 	"golang.org/x/exp/maps"
+	"golang.org/x/exp/slices"
```
DrJosh9000, line 37: Oh goodness. When did we start depending on `exp` packages?

At least these will be making their way into Go 1.21 largely the same.

internal/pipeline/command_step.go
DrJosh9000, line 1: If it's alright with you, I was thinking of a quick cleanup PR to move the step types into their own files - might be better to do that first.

clicommand/agent_start.go
moskyb: um, it's, um, it's [hard to say, exactly](https://github.com/buildkite/agent/commit/851073dc33c89a89f623cd0a0ce1e7b8a6394b79) when (May 5th 2022) this started or who (Benjamin J. Moskovitz) started it

internal/pipeline/command_step.go
moskyb: sounds great!

agent/verify_job.go
moskyb: ooh, good call

agent/integration/test_helpers.go
```diff
@@ -0,0 +1,174 @@
...
+	bs, err := bintest.NewMock(fmt.Sprintf("buildkite-agent-bootstrap-%s", name))
+	if err != nil {
+		t.Fatalf("bintest.NewMock() error = %v", err)
+	}
+
+	return bs
+}
+
+func runJob(t *testing.T, j *api.Job, server *httptest.Server, cfg agent.AgentConfiguration, bs *bintest.Mock) {
+	l := logger.Discard
```
triarius, line 42:
```suggestion
func runJob(t *testing.T, j *api.Job, server *httptest.Server, cfg agent.AgentConfiguration, bs *bintest.Mock) {
	t.Helper()

	l := logger.Discard
```

agent/run_job.go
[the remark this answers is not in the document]
triarius: Yeah, technically, the job still runs. It just fails the verification and prints the verification error into the job logs. From the point of view of the API, the job executed, just not completely and definitely not successfully. So the row in the job table still needs startedAt and finishedAt timestamps.

agent/run_job.go
```diff
@@ -62,6 +79,37 @@ func (r *JobRunner) Run(ctx context.Context) error {
...
+			exit.SignalReason = "job_verification_failed_invalid_signature"
+			return nil
+		}
+
+	default: // some other error
+		r.verificationFailureLogs(err, VerificationBehaviourBlock) // errors in verification are always fatal
+		exit.Status = -1
+		exit.SignalReason = "job_verification_failed_with_error"
+		return nil
+	}
```
triarius, line 111: I do think the `if/else` chain I suggested last time looks better here.

I, too, generally prefer `switches` to `if/else` chains, but here:

1. There is no added efficiency in using a `switch` as all the case statements have to be evaluated in order in the worst case as there is no value to switch on.
2. This statement leaks the `ise` variable. The `if/else` chain would scope it exclusively to the appropriate branch


But I'm not going to die on this hill.

agent/verify_job.go
triarius: I guess the go way is to let that be handled by unwrapping the underlying error.

triarius (commented):
This is shaping up to be amazing!

agent/verify_job.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+func (e *invalidSignatureError) Unwrap() error {
+	return e.underlying
+}
+
+func (e *invalidSignatureError) As(target any) bool {
+	_, ok := target.(*invalidSignatureError)
+	return ok
+}
```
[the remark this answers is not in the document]
moskyb: oh, how good, easy as

triarius (approved):
🎉

triarius:
This is great!

agent/run_job.go
DrJosh9000: If I were to change anything, it would be to replace `default:` with `case err != nil:`, and not make `err == nil` a case in the switch.

As in, the other side of "indent error flow" is "don't indent non-error flow".

agent/verify_job.go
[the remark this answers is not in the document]
DrJosh9000: "easy **as**" lol

Delete it tho?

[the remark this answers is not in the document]
moskyb: oh i coulda been sure that i did..... but i didn't

api/jobs.go
```diff
@@ -9,20 +9,34 @@ import (
...
+	ExitStatus         string               `json:"exit_status,omitempty"`
+	Signal             string               `json:"signal,omitempty"`
+	SignalReason       string               `json:"signal_reason,omitempty"`
+	StartedAt          string               `json:"started_at,omitempty"`
+	FinishedAt         string               `json:"finished_at,omitempty"`
+	RunnableAt         string               `json:"runnable_at,omitempty"`
+	ChunksFailedCount  int                  `json:"chunks_failed_count,omitempty"`
+}
+
+func (j Job) ValuesForFields(fields []string) (map[string]string, error) {
```
DrJosh9000, line 28: `Job` looks big enough that I'd make this receiver `*Job` (for not-copying-it-over-stack reasons, not for modifying-the-fields reasons).

agent/verify_job.go
```diff
@@ -0,0 +1,82 @@
...
+		return ErrNoSignature
+	}
+
+	// Verify the signature
+	if err := step.Signature.Verify(&step, verifier); err != nil {
+		return newInvalidSignatureError(err)
+	}
+
+	// Now that the signature of the job's step is verified, we need to check if the fields on the job match those on the
+	// step. If they don't, we need to fail the job
```
DrJosh9000, line 63: I was going to suggest it would be simpler to override `BUILDKITE_COMMAND` with `step.Command`, than to carefully compare each field.

But thinking about it, they would only differ if the backend (ours or the hypothetical attacker's) is doing crimes, in which case we probably shouldn't run anything.

Might be worth documenting that?

DrJosh9000 (approved):
Huzzah! 🧃

agent/verify_job.go
moskyb: just to double check, are saying that it might be good to document that the only reason that step and job fields should differ is if an attacker is doing crimes on the backend?

internal/pipeline/step_command.go
```diff
@@ -18,6 +18,7 @@ type CommandStep struct {
 	Command   string     `yaml:"command"`
 	Plugins   Plugins    `yaml:"plugins,omitempty"`
 	Signature *Signature `yaml:"signature,omitempty"`
+	Matrix    any        `yaml:"matrix,omitempty"`
```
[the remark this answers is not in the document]
moskyb: oooh good point

agent/verify_job.go
DrJosh9000: It might not be the only reason, but it definitely seems like a reason

internal/pipeline/step_command.go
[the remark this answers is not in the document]
DrJosh9000: ... and `interpolate` ...

agent/run_job.go
```diff
@@ -62,6 +79,49 @@ func (r *JobRunner) Run(ctx context.Context) error {
...
+				exit.Status = -1
+				exit.SignalReason = "job_verification_failed_no_signature"
+				return nil
+			}
+
+		case errors.As(err, &ise):
+			r.verificationFailureLogs(err, r.InvalidSignatureBehavior)
+			if r.InvalidSignatureBehavior == VerificationBehaviourBlock {
+				exit.Status = -1
+				exit.SignalReason = "job_verification_failed_invalid_signature"
```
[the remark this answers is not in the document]
triarius: If signal reason was designed for users to catch and retry, then I don't think we should be overloading its use to also display information in the timeline. Is there another way of creating a job event for signing failures? If not, I think we should examine building one when the time to build the timeline feature arrives.

PR #2216, opened 2023-07-18
Add signatures when marshalling steps to json

moskyb:
The pipeline is actually marshalled to json by the agent, and then back to yaml again on the backend, as it turns out.

### DIFF
--- a/internal/pipeline/step.go
+++ b/internal/pipeline/step.go
@@ -27,7 +27,6 @@ type Step interface {
 
 // CommandStep models a command step.
 //
-// If you add a field to this struct, ensure you add a corresponding block to the `MarshallJSON` method! It won't happen automatically.
 // Standard caveats apply - see the package comment.
 type CommandStep struct {
 	Command   string     `yaml:"command"`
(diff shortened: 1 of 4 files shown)


internal/pipeline/step.go
```diff
@@ -47,13 +48,19 @@ func (c *CommandStep) MarshalJSON() ([]byte, error) {
...
+
 	if c.Command != "" {
 		out["command"] = c.Command
 	}
+
 	if len(c.Plugins) > 0 {
 		out["plugins"] = c.Plugins
 	}
 
+	if c.Signature != nil {
```
DrJosh9000, line 60: Haha, whoops. 

🤔 It might be worth a tiny bit of reflection in a helper func that could be used across different structs, to cover the gap between yaml's `inline` feature and json's lack of it...

DrJosh9000 (approved):
Good catch!

internal/pipeline/step.go
moskyb: yeah, i wondered about this as well... i might give it a crack just for fun

moskyb: okay, i've had my fun in 9962f55, if it seems like it's overcomplicating things (it might be?) i can just yeet that commit

internal/pipeline/step.go
```diff
@@ -27,6 +27,7 @@ type Step interface {
 
 // CommandStep models a command step.
 //
+// If you add a field to this struct, ensure you add a corresponding block to the `MarshallJSON` method! It won't happen automatically.
```
DrJosh9000, line 30: Well this is no longer true. 😄

internal/job/_bintest_06ec6d25081df5c7f7c59c7397831c7eae5da454/main.go
```diff
@@ -0,0 +1,21 @@
+package main
```
DrJosh9000, line 1: Inadvertent addition?

moskyb: 😬 whoops!

PR #2217, opened 2023-07-18
Fix docker builds for Ubuntu 22.04

moskyb:
Python dependency snafus mean we have to constrain pyyaml to an older version, see: https://github.com/yaml/pyyaml/issues/724


DrJosh9000 (approved):
yay 🤪

Issue #2208, opened 2023-07-17
Known issues with git mirrors


DrJosh9000:
This issue exists to billboard known problems with git mirrors across different agent versions.

## Checkouts sometimes fail with 'bad object', 'did not send all necessary objects' errors

**Issue:** Checkouts sometimes fail because the mirror appears to contain the commit needed for a job, but the mirror is corrupt or missing necessary objects. The agent may try to repair the mirror, but checkout still fails on a retry.

**Affected versions:** Unknown.

**Status:** Investigating.

**Workarounds:** 

- Manually run `git fsck --full` followed by `git fetch origin` in the affected mirror directory.
- Delete the affected mirror directory, allowing the agent to make a fresh clone.
- Disable git mirrors.

## Mirrors of submodules not fetched

**Issue:** When the agent runs jobs that check out a repository that uses git submodules, the submodules should also be mirrored. The main repository mirror is updated (fetched) to ensure the commit or branch being built was present, but a bug meant that submodule mirrors was not fetched. As a result, submodule mirrors might become out -of-date, missing commits or objects needed by the main repo.

**Affected versions:** Between v3.44.0 and v3.49.0, inclusive.

**Status:** Fixed with #2203, which should be released as part of v3.50.

**Workarounds:** 

- Use an experimental agent build, containing a fix. 
- If each host only runs one agent, include a manual `git fetch origin` for all submodule mirror directories in your pipeline.
- Use a custom checkout hook.
- Disable git mirrors.

## Mirrors of submodules pull from wrong remote URL

**Issue:** When the agent runs jobs that check out a repository that uses git submodules, the submodules should also be mirrored. A bug caused the remote URL in the mirror directory for each submodule to be the wrong URL.

**Affected versions:** Between v3.44.0 and v3.48.0, inclusive.

**Status:** Fixed with #2144, but because the submodule mirror might not be re-fetched, it can still contain the wrong git history and objects. The missing fetch is fixed by #2203, which should be released as part of v3.50.

**Workarounds:** 

- Use an experimental agent build, containing a fix.
- If each host only runs one agent, update to at least agent v3.49 and include a manual `git fetch origin` for all submodule mirror directories in your pipeline.
- Use a custom checkout hook.
- Disable git mirrors.

PR #2218, opened 2023-07-18
Clean-retry on "fatal: bad object"

DrJosh9000:
On the theory that mirrors sometimes have objects gc-ed that are still referred to by existing checkout directories, make `fatal: bad object` an error that we can try to recover from by deleting the checkout directory and re-cloning.

I made `gitErrorFetchBadObject` its own enum value, in case we want to apply the recovery only when git mirrors are enabled.

cc #2208


triarius (approved):
Nice solution 🕵️🔍🧙

PR #2223, opened 2023-07-19
Add some notes about git mirrors

DrJosh9000:
This felt broadly useful, but I'm not 100% certain it belong here.

### DIFF
--- a/docs/git-mirror.md
+++ b/docs/git-mirror.md
@@ -1,0 +1,224 @@
+# How Git Mirrors Works
+
+## The goal
+
+The goal of Git mirrors is to:
+
+- reduce network bandwidth, and
+- reduce disk usage
+
+for machines running multiple agents side-by-side.[^1]
+
+[^1]: We also support agents across multiple machines sharing a Git mirror via a
+network share.
+
+How do we do this? By:
+
+- Mirroring the repository in a central location (the git mirror directory), and
+- Making each checkout use a `--reference` clone of the mirror.
+
+## Mirrors? `--mirror`? 🪞
+
+You're probably familiar with `git clone`: this gives you a copy of a repo you
+can use to make changes, with all the git magic hidden in a secret `.git`
+directory.
+
+You might also be familiar with `git clone --bare`: this exposes all the
+normally-hidden Git magic, at the expense of not checking out the main branch
+for you in the root of the directory. 
+
+So instead of:
+
+```shell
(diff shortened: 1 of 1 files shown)


pda:
Nice!

> You might guess that `git fetch` only ever adds new objects, but unfortunately,
> `git fetch` by default runs `git maintentance run --auto`, which sometimes
> cleans up old unreferenced objects.

Could we `git fetch --no-auto-maintenance`? Or is the corrupting fetch outside our control? Or is that maintenance there for a good reason? 😅

DrJosh9000:
> Could we git fetch --no-auto-maintenance? Or is the corrupting fetch outside our control? Or is that maintenance there for a good reason? 😅

Supposing we run --no-auto-maintenance, then I imagine (assuming commits and branches continue churning objects remotely) the mirror would continue to grow until either Git becomes too slow to use, or disk fills up.

pda:
> Supposing we run --no-auto-maintenance, then I imagine (assuming commits and branches continue churning objects remotely) the mirror would continue to grow until either Git becomes too slow to use, or disk fills up.

I can imagine use-cases where that's preferable to `git fetch` corrupting it, especially on CI agent machines which _probably_ have a finite lifetime before the mirror is thrown away. I imagine the growth of a typical non-garbage-collected git mirror isn't that much larger than a garbage-collected one, assuming most branches are merged and so most commits are kept forever. But there's probably details I'm not considering!

docs/git-mirror.md
```diff
@@ -0,0 +1,173 @@
+# How Git Mirrors Works
+
+## The goal
+
+The goal of Git mirrors is to:
+
+- reduce network bandwidth, and
+- reduce disk usage
+
+for machines running multiple agents side by side.
```
moskyb, line 10: maaaaaaaaybe it's worth talking about how "git mirror on a ZFS share" is also a supported workflow?

moskyb (approved):
this is awesome awesome work - the inner workings of git mirrors have always been a bit of a mystery to me, but this cleared a whole bunch of stuff up

docs/git-mirror.md
DrJosh9000: Congratulations! You have won one (1) footnote!

sj26:
Nice!

I have some really old hooks where I tried to use git worktrees for builds, if it's at all useful:

https://github.com/sj26/git-worktree-buildkite-hooks

PR #2225, opened 2023-07-19
Allow empty steps

DrJosh9000:
Some (usually automatically-produced) pipelines have no steps. Previously we didn't error on this case, but while changing the pipeline parser I added these checks. There's no reason for the agent to care about it, so remove the check.


DrJosh9000:
Tests are passing, so I probably didn't add one.

PR #2227, opened 2023-07-20
Increase test coverage for new parser

DrJosh9000:
- Add interpolation to group step parser test
- Add a block step

### DIFF
--- a/internal/pipeline/parser_test.go
+++ b/internal/pipeline/parser_test.go
@@ -185,24 +185,31 @@ func TestParserParsesNoSteps(t *testing.T) {
 }
 
 func TestParserParsesGroups(t *testing.T) {
+	envMap := env.FromSlice([]string{`ENV_VAR_FRIEND="friend"`})
+
 	input := strings.NewReader(`---
 steps:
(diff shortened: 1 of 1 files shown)


triarius (approved):
LGTM 🎉

moskyb (approved):

internal/pipeline/parser_test.go
```diff
@@ -185,24 +185,31 @@ func TestParserParsesNoSteps(t *testing.T) {
 }
 
 func TestParserParsesGroups(t *testing.T) {
+	envMap := env.FromSlice([]string{`ENV_VAR_FRIEND="friend"`})
+
 	input := strings.NewReader(`---
 steps:
   - group:
     steps:
```
dabarrell, line 193: I think this would fail without a label - needs to either be in the group key, or separate with group: null
```suggestion
  - group:
    label: "Group label"
    steps:
```

dabarrell: oh nevermind, this doesn't fail the upload. just verified locally

PR #2230, opened 2023-07-20
Fix an error where exit status wasn't getting captured by the cleanup process

moskyb:
tl;dr before:
```Go
defer cancel()
defer r.cleanup(cctx, &wg, exit)
```
after:
```Go
defer func(ctx context.Context, wg *sync.WaitGroup) {
	cancel()
	r.cleanup(ctx, wg, exit)
}(ctx, &wg)
```

previously we were capturing the value of `exit` at the time of the call to `defer`, now we're capturing the value at the time the defer is actually executed, which is what we wanted in the first place.

this PR also adds a regression test to ensure that this doesn't happen again

### DIFF
--- a/agent/integration/job_runner_integration_test.go
+++ b/agent/integration/job_runner_integration_test.go
@@ -1,7 +1,6 @@
 package integration
 
 import (
-	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -11,6 +10,8 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strconv"
+	"strings"
 	"testing"
 
 	"github.com/buildkite/agent/v3/agent"
@@ -56,13 +57,7 @@ func TestPreBootstrapHookRefusesJob(t *testing.T) {
 
 	runJob(t, j, server, agent.AgentConfiguration{HooksPath: hooksDir}, mb)
 
-	finishRequestBody := bytes.NewBuffer(e.calls["/jobs/my-job-id/finish"][0])
-
-	var job api.Job
-	err = json.NewDecoder(finishRequestBody).Decode(&job)
-	if err != nil {
-		t.Fatalf("decoding accept request body: %v", err)
-	}
+	job := e.finishesFor(t, jobID)[0]
 
 	if got, want := job.ExitStatus, "-1"; got != want {
 		t.Errorf("job.ExitStatus = %q, want %q", got, want)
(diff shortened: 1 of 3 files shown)


agent/integration/job_runner_integration_test.go
```diff
@@ -214,6 +248,22 @@ func createTestAgentEndpoint() *testAgentEndpoint {
 	}
 }
 
+func (tae *testAgentEndpoint) finishesFor(t *testing.T, jobID string) []api.Job {
+	endpoint := fmt.Sprintf("/jobs/%s/finish", jobID)
+	finishes := make([]api.Job, 0, len(tae.calls))
+
+	for _, b := range tae.calls[endpoint] {
+		var job api.Job
+		err := json.NewDecoder(bytes.NewBuffer(b)).Decode(&job)
```
DrJosh9000, line 257:
```suggestion
		err := json.Unmarshal(b, &job)
```

agent/integration/job_runner_integration_test.go
```diff
@@ -230,7 +280,7 @@ func (t *testAgentEndpoint) server(jobID string) *httptest.Server {
 		case "/jobs/" + jobID + "/finish":
 			rw.WriteHeader(http.StatusOK)
 		default:
-			http.Error(rw, fmt.Sprintf("not found; method = %q, path = %q", req.Method, req.URL.Path), http.StatusNotFound)
+			panic(fmt.Sprintf("not found; method = %q, path = %q", req.Method, req.URL.Path))
```
DrJosh9000, line 283: What's this change for?

DrJosh9000 (approved):
Looking good - I have one or two tweaks

agent/integration/job_runner_integration_test.go
triarius: I'm curious about this too. Is it because you need a json error message? I think you should craft one if so.

moskyb: this is a relic from when i was bughunting and flailing a bit; i think my logic was that if an endpoint that we haven't accounted for is hit it's a Very Bad Thing, and i don't think these errors are super well surfaced in the tests???? maybe? but also this doesn't really need to be here, i might just yeet it

agent/integration/job_runner_integration_test.go
moskyb: oh, duh

agent/integration/job_runner_integration_test.go
triarius: I think it's better to return a 4xx error in such circumstances and raise an error from the Test method. It will be easier to debug how the route was called then.

triarius: I guess the error handling could happen entirely in the non-test code. Hopefully, writing code that ignores errors is something we have gotten out of the habit of doing.

PR #2238, opened 2023-07-24
Two-phase pipeline parsing

DrJosh9000:
There's a bug where multiple [map merges](https://yaml.org/type/merge.html) aren't unmarshaled into structs in the way we've supported them.

To address this we need to go back to having greater control over the YAML unmarshaling process. It will also help if we, say, want to switch YAML library later on - if a new library supports the things we need to do, then we can simplify, and if it doesn't (or has subtle differences) then the intermediate processing can be the place where we make up the gap.

The key change here is from allowing `yaml.v3` to unmarshal into structs using its reflective logic to:

- Use yaml.v3 to parse into `*yaml.Node`, which provides all the relevant information in the document without further processing.
- Unmarshal that node using `ordered`, which handles merges and aliases (including multiple merges).
- Unmarshal the dynamic, weakly-typed structure made up of `*ordered.MapSA`, `[]any`, and basic types into each struct manually.

The last manual part could be converted into a `yaml.v3`-like or `encoding/json`-like reflection-based approach at some point.

### DIFF
--- a/internal/pipeline/parser_test.go
+++ b/internal/pipeline/parser_test.go
@@ -106,9 +106,9 @@ steps:
 			&CommandStep{
 				Command: "docker build .",
 				RemainingFields: map[string]any{
-					"agents": map[string]any{
-						"queue": "default",
-					},
+					"agents": ordered.MapFromItems(
+						ordered.TupleSA{Key: "queue", Value: "default"},
+					),
 					"name":              ":docker: building image",
 					"type":              "script",
 					"agent_query_rules": []any{"queue=default"},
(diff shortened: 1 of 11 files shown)


internal/pipeline/parser_test.go
```diff
@@ -106,23 +106,110 @@ steps:
...
+remainder: &remainder
+  name: ':docker: building image'
+  command: docker build .
+  agents:
+    queue: default
+
+steps:
+  - <<: *base_step
+    <<: *remainder
+`
```
triarius, line 176: Can we get a test for the official way to do it as well?
```yaml
steps:
- <<: [*base_step, *remainder]
```

internal/pipeline/parser_test.go
```diff
@@ -106,23 +106,110 @@ steps:
...
+    }
+  ]
+}`
+	if diff := cmp.Diff(string(gotJSON), wantJSON); diff != "" {
+		t.Errorf("marshalled JSON diff (-got +want):\n%s", diff)
+	}
+}
+
+func TestParserSupportsDoubleMerge(t *testing.T) {
+	const complexYAML = `---
```
triarius, line 161:
```suggestion
  t.Parallel()
	const complexYAML = `---
```

triarius (approved):
Amazing stuff!

internal/pipeline/parser_test.go
DrJosh9000: Done!

triarius (approved):

moskyb (approved):
fantastic work 🎉

PR #2240, opened 2023-07-24
Actually bump VERSION

DrJosh9000:
🤦


triarius (approved):
I would like to get rid of this need. I think the model in agent-stack-k8s is the future.

PR #2246, opened 2023-07-26
Check that go.mod is tidy in CI

moskyb:
i accidentally committed a go.mod with [`kr/pretty`](https://github.com/kr/pretty) (which i often use for debugging) the other day. this PR makes it so that committing a non-tidy go.mod file will fail CI

### DIFF
--- a/.buildkite/steps/test-go-fmt.sh
+++ b/.buildkite/steps/test-go-fmt.sh
@@ -8,4 +8,10 @@ if [[ $(gofmt -l ./ | head -c 1 | wc -c) != 0 ]]; then
   exit 1
 fi
 
+if [[ $(go mod tidy -v 2> >(wc -c)) != 0 ]]; then # go mod tidy -v outputs to stderr for some reason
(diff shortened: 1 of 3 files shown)


triarius (approved):
Thank you! I do remember seeing it, but I forgot to check where it was used 🤦‍♂️

.buildkite/steps/test-go-fmt.sh
```diff
@@ -8,4 +8,10 @@ if [[ $(gofmt -l ./ | head -c 1 | wc -c) != 0 ]]; then
   exit 1
 fi
 
+if [[ $(go mod tidy -v 2> >(wc -c)) != 0 ]]; then # go mod tidy -v outputs to stderr for some reason
```
DrJosh9000, line 11: `2> >(...)` is fun!

DrJosh9000 (approved):

.buildkite/steps/test-go-fmt.sh
moskyb: sigh.

PR #2241, opened 2023-07-24
Revert "Fix docker builds for Ubuntu 22.04"

DrJosh9000:
Reverts buildkite/agent#2217

Looks like https://github.com/yaml/pyyaml/issues/724 is fixed?

Fixes #2232

Issue #2232, opened 2023-07-20
Critical PyYAML vulnerability in latest buildkite/agent Ubuntu image


tenyo:
Our image scanner is catching a critical vulnerability in the `buildkite/agent:3-ubuntu` image since 2 days ago.
Likely related to https://github.com/buildkite/agent/pull/2217.

Scan results from latest image:
```
% trivy image buildkite/agent:3-ubuntu --severity "HIGH,CRITICAL"
2023-07-20T11:12:50.228-0400	INFO	Vulnerability scanning is enabled
2023-07-20T11:12:50.229-0400	INFO	Secret scanning is enabled
2023-07-20T11:12:50.229-0400	INFO	If your scanning is slow, please try '--scanners vuln' to disable secret scanning
2023-07-20T11:12:50.229-0400	INFO	Please see also https://aquasecurity.github.io/trivy/v0.43/docs/scanner/secret/#recommendation for faster secret detection
2023-07-20T11:13:09.257-0400	INFO	Detected OS: ubuntu
2023-07-20T11:13:09.257-0400	INFO	Detecting Ubuntu vulnerabilities...
2023-07-20T11:13:09.268-0400	INFO	Number of language-specific files: 2
2023-07-20T11:13:09.268-0400	INFO	Detecting gobinary vulnerabilities...
2023-07-20T11:13:09.272-0400	INFO	Detecting python-pkg vulnerabilities...

buildkite/agent:3-ubuntu (ubuntu 22.04)

Total: 0 (HIGH: 0, CRITICAL: 0)

2023-07-20T11:13:09.290-0400	INFO	Table result includes only package filenames. Use '--format json' option to get the full path to the package file.

Python (python-pkg)

Total: 1 (HIGH: 0, CRITICAL: 1)

┌───────────────────┬────────────────┬──────────┬───────────────────┬───────────────┬────────────────────────────────────────────┐
│      Library      │ Vulnerability  │ Severity │ Installed Version │ Fixed Version │                   Title                    │
├───────────────────┼────────────────┼──────────┼───────────────────┼───────────────┼────────────────────────────────────────────┤
│ PyYAML (METADATA) │ CVE-2020-14343 │ CRITICAL │ 5.3.1             │ 5.4           │ PyYAML: incomplete fix for CVE-2020-1747   │
│                   │                │          │                   │               │ https://avd.aquasec.com/nvd/cve-2020-14343 │
[...]

PR #2248, opened 2023-07-27
Remove docker-compose v1 from ubuntu 22.04 and replace with compatibility script

triarius:
Docker Compose V1 is very much deprecated. We don't offer it on the latest major version of the Elastic CI Stack (v6) and on the alpine docker images, we have been symlinking it to docker-compose v2.

This PR removes docker-compose v1 from the ubuntu 22.04 docker image and replaces it with the compatibility script that calls docker compose v2 with the `--compatbility` flag prepended to the arguments.
Also, it replaces the symlink in the alpine docker images with that script.

Fixes: \#2232
Closes: \#2241


DrJosh9000 (approved):
🚢

PR #2250, opened 2023-07-27
Make the the agent send a SIGTERM (configurable) before a SIGKILL to subprocesses

triarius:
The agent start process already does something similar to the job executor process when it detects that the buildkite job had been cancelled. However, every other process launched by the agent is unceremoniously killed with a SIGKILL when the job is cancelled. This has been causing issues with the checkout directory becoming corrupted if the job is cancelled during checkout.

This PR makes the agent first send a SIGTERM (which is configurable by the `cancel-signal` config option) and give the process a grace period to clean up before sending a SIGKILL. The grace period is also configurable, but it needs to be smaller than the grace period allowed to the job executor by the agent start process. I've tried to avoid confusion between the two - the grace period for the job executor is contained in the setting `cancel-grace-period`, while the grace periods for the sub processes of the job executor is called contained in the setting `signal-grace-period-seconds`. After it is parsed, it is converted to a `time.Duration` and passes around as `SignalGracePeriod`. But it seems inevitable that there will be confusion to the uninitiated. In a better world, `cancel-grace-period` would be `cancel-grace-period-seconds`. In an even better world, we would drop the units and parse them as durations from the config. 

Finally, I have set the default signal grace period to 0s. This effectively makes this PR a no-op. The plan is to dog-food it at a value of 9s on our internal pipelines before making that the default in a future version of the agent.

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -953,8 +953,8 @@ var AgentStartCommand = cli.Command{
 			}
 
 			// Create an agent worker to run the agent
-			workers = append(workers, agent.NewAgentWorker(l.WithFields(
-				logger.StringField("agent", ag.Name)),
+			workers = append(workers, agent.NewAgentWorker(
+				l.WithFields(logger.StringField("agent", ag.Name)),
 				ag,
 				mc,
 				client,
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -117,6 +117,7 @@ type AgentStartConfig struct {
 	SpawnWithPriority           bool     `cli:"spawn-with-priority"`
 	LogFormat                   string   `cli:"log-format"`
 	CancelSignal                string   `cli:"cancel-signal"`
+	SignalGracePeriodSeconds    int      `cli:"signal-grace-period-seconds"`
 	RedactedVars                []string `cli:"redacted-vars" normalize:"list"`
 
 	// Global flags
@@ -278,12 +279,7 @@ var AgentStartCommand = cli.Command{
 			Usage:  "The maximum idle time in seconds to wait for a job before disconnecting. The default of 0 means no timeout",
 			EnvVar: "BUILDKITE_AGENT_DISCONNECT_AFTER_IDLE_TIMEOUT",
 		},
-		cli.IntFlag{
-			Name:   "cancel-grace-period",
-			Value:  10,
-			Usage:  "The number of seconds a canceled or timed out job is given to gracefully terminate and upload its artifacts",
-			EnvVar: "BUILDKITE_CANCEL_GRACE_PERIOD",
-		},
+		cancelGracePeriodFlag,
 		cli.BoolFlag{
 			Name:   "enable-job-log-tmpfile",
 			Usage:  "Store the job logs in a temporary file ′BUILDKITE_JOB_LOG_TMPFILE′ that is accessible during the job and removed at the end of the job",
@@ -548,12 +544,8 @@ var AgentStartCommand = cli.Command{
 			Usage:  "Assign priorities to every spawned agent (when using --spawn) equal to the agent's index",
 			EnvVar: "BUILDKITE_AGENT_SPAWN_WITH_PRIORITY",
 		},
-		cli.StringFlag{
-			Name:   "cancel-signal",
-			Usage:  "The signal to use for cancellation",
-			EnvVar: "BUILDKITE_CANCEL_SIGNAL",
-			Value:  "SIGTERM",
-		},
+		cancelSignalFlag,
+		signalGracePeriodSecondsFlag,
 		cli.StringFlag{
 			Name:   "tracing-backend",
 			Usage:  `Enable tracing for build jobs by specifying a backend, "datadog" or "opentelemetry"`,
@@ -756,6 +748,12 @@ var AgentStartCommand = cli.Command{
 			}
 		}
 
+		if cfg.CancelGracePeriod <= cfg.SignalGracePeriodSeconds {
+			l.Fatal("cancel-grace-period must be greater than signal-grace-period-seconds")
+		}
+
+		signalGracePeriod := time.Duration(cfg.SignalGracePeriodSeconds) * time.Second
+
 		mc := metrics.NewCollector(l, metrics.CollectorConfig{
 			Datadog:              cfg.MetricsDatadog,
 			DatadogHost:          cfg.MetricsDatadogHost,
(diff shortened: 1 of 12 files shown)


process/main_test.go
triarius, line 1: The `TestMain` method has been extract into this file. When called with an appropriate environment variable, it acts as a subprocess calledn from the tests.

process/process.go
```diff
@@ -43,6 +43,10 @@ var signalMap = map[string]Signal{
 	"SIGTERM": SIGTERM,
 }
 
+var ErrNotWaitStatus = errors.New(
+	"unimplemented for systems where exec.ExitError.Sys() is not syscall.WaitStatus",
```
triarius, line 47: This error message isn't very useful for a user. But it was what was there before.

internal/job/executor.go
```diff
@@ -80,6 +80,7 @@ func (e *Executor) Run(ctx context.Context) (exitCode int) {
 		e.shell.PTY = e.ExecutorConfig.RunInPty
 		e.shell.Debug = e.ExecutorConfig.Debug
 		e.shell.InterruptSignal = e.ExecutorConfig.CancelSignal
```
triarius, line 82: Very annoying that this rename happened.

process/process.go
```diff
@@ -247,14 +269,16 @@ func (p *Process) Run(ctx context.Context) error {
 
 	// Convert the wait result into a native WaitStatus
 	if p.waitResult != nil {
```
triarius, line 271: The following is intended to be a no-op refactor

process/process.go
```diff
@@ -218,22 +223,39 @@ func (p *Process) Run(ctx context.Context) error {
...
 			case <-p.Done():
 				return
 			}
-		}()
-	}
+		// context was not canceled before the process terminated
+		case <-p.Done():
+			return
+		}
+	}()
```
triarius, line 258: This is the crux of the PR.

process/process.go
```diff
@@ -228,15 +230,24 @@ func (p *Process) Run(ctx context.Context) error {
...
+			select {
+			case <-p.Done(): // process exited after being interrupted
+				return
+			case <-time.After(p.conf.SignalGracePeriod):
+				p.logger.Warn("[Process] Has not terminated in time, killing. pid=%d", p.pid)
+				if err := p.Terminate(); err != nil {
+					p.logger.Error("[Process] error sending SIGKILL: %s", err)
+					return
+				}
+			}
```
triarius, line 250: This diff is the crux of the PR.

pda:
+1 we should absolutely never SIGKILL anything without SIGTERM and a grace period first 👍🏼

process/process.go
triarius: For historical reasons, `p.Terminate` is SIGKILL, `p.Interrupt` is (by default) SIGTERM.
https://github.com/buildkite/agent/blob/5c089a8242ca77f32075229eda64396b097f7be8/process/signal.go#L35-L50

clicommand/agent_start.go
```diff
@@ -756,6 +748,12 @@ var AgentStartCommand = cli.Command{
 			}
 		}
 
+		if cfg.CancelGracePeriod <= cfg.SignalGracePeriodSeconds {
```
DrJosh9000, line 751: Would be nice to make the names consistent (and even nicer if they were just `time.Duration` instead of `int`), but I can see how we got here. 🤷

clicommand/agent_start.go
```diff
@@ -954,16 +953,21 @@ var AgentStartCommand = cli.Command{
...
-					l.WithFields(logger.StringField("agent", ag.Name)), ag, mc, client, agent.AgentWorkerConfig{
-						AgentConfiguration: agentConf,
-						CancelSignal:       cancelSig,
-						Debug:              cfg.Debug,
-						DebugHTTP:          cfg.DebugHTTP,
-						SpawnIndex:         i,
-						AgentStdout:        os.Stdout,
-					}))
+			workers = append(workers, agent.NewAgentWorker(l.WithFields(
+				logger.StringField("agent", ag.Name)),
```
DrJosh9000, line 957: The parentheses seem wrongly-aligned here. It looks a lot like most of this is contained within `l.WithFields(...)` but actually it's only `logger.StringField(...)` and the rest are args to `agent.NewAgentWorker`.

```suggestion
			workers = append(workers, agent.NewAgentWorker(
				l.WithFields(logger.StringField("agent", ag.Name)),
```

clicommand/cancel_signal.go
```diff
@@ -0,0 +1,36 @@
...
+const (
+	defaultCancelGracePeriod = 10
+
+	// This will be increased to 9 in a future release of the agent.
+	defaultSignalGracePeriod = 0
+)
+
+var (
+	// cancel grace period must be stricly longer than signal grace period
+	_ uint = defaultCancelGracePeriod - defaultSignalGracePeriod - 1
```
DrJosh9000, line 14: Heh, clever use of untyped consts.

clicommand/cancel_signal.go
```diff
@@ -0,0 +1,36 @@
...
+
+const (
+	defaultCancelGracePeriod = 10
+
+	// This will be increased to 9 in a future release of the agent.
+	defaultSignalGracePeriod = 0
+)
+
+var (
+	// cancel grace period must be stricly longer than signal grace period
```
DrJosh9000, line 13:
```suggestion
	// cancel grace period must be strictly longer than signal grace period
```

clicommand/agent_start.go
```diff
@@ -1034,7 +1038,7 @@ func handlePoolSignals(ctx context.Context, l logger.Logger, pool *agent.AgentPo
 
 		for sig := range signals {
 			l.Debug("Received signal `%v`", sig)
-			setStatus(fmt.Sprintf("Recieved signal `%v`", sig))
+			setStatus(fmt.Sprintf("Received signal `%v`", sig))
```
DrJosh9000, line 1041: 😅

DrJosh9000 (approved):

clicommand/agent_start.go
triarius: Yeah, in next major version, I would want to change both to parse durations. There are already some args that are parsed as durations, but as these are related, and I was constrained by `cancel-grace-period`, I ended up with this.

clicommand/cancel_signal.go
triarius: I was tossing up between this and making it a test. Because both sources of truth are in the same file, I think it's more convenient to get the feedback from the compiler. And if you have a language server, you get it directly in the editor.

clicommand/agent_start.go
triarius: So rare for me to be on the other side of this 😂

PR #2251, opened 2023-07-27
Refactor some code in process.go

triarius:
More details available commit by commit

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -222,21 +222,23 @@ func (p *Process) Run(ctx context.Context) error {
 	}
 
 	// When the context finishes, terminate the process
-	if ctx != nil {
-		go func() {
-			select {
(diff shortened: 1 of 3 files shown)


process/process.go
```diff
@@ -43,6 +43,10 @@ var signalMap = map[string]Signal{
 	"SIGTERM": SIGTERM,
 }
 
+var ErrNotWaitStatus = errors.New(
+	"unimplemented for systems where exec.ExitError.Sys() is not syscall.WaitStatus",
```
triarius, line 47: This error message is not helpful to users, so should not really be exposed except under debug. However, it is more or less what was there before.

moskyb (approved):
🎉 love your work

DrJosh9000 (approved):
👍

PR #2252, opened 2023-07-28
Pipeline Signing/Verification with JWS

moskyb:
**This PR:** Updates the pipeline signing and verification process to use [JWS](https://en.wikipedia.org/wiki/JSON_Web_Signature) to sign and verify jobs being uploaded/run.

This gives us several benefits:
- We get the use of [a bunch](https://datatracker.ietf.org/doc/html/rfc7518#section-3) of new signing algorithms, including asymmetric ones
- It makes key rotations for agents a lot easier - users can load multiple verifiation keys onto an agent and it will attempt to verify incoming jobs using all of them. This means that users can switch out which key they use to sign jobs much more easily

This PR is dependant on #2210, and won't be merged until it is at the earliest


triarius (commented):
I think you should keep the Verifier and Signer interfaces, but define them in the agent package (where they were used). The Signer should have a method to sign, and a verifier should have a method to verify.

Then you can have a JWKS implementation for the signer and the verifier in the pipelines package that constructed from the information in the pipeline yaml and/or other configuration.

agent/job_runner.go
```diff
@@ -377,10 +381,10 @@ func NewJobRunner(l logger.Logger, apiClient APIClient, conf JobRunnerConfig) (j
...
 
 func (r *JobRunner) normalizeVerificationBehavior(behavior string) (string, error) {
-	if r.conf.AgentConfiguration.JobVerificationKeyPath == "" {
-		// We won't be verifying jobs, so it doesn't matter
-		return "if you're seeing this string, there's a problem with the job verification code in the agent. contact support@buildkite.com", nil
-	}
+	// if r.conf.AgentConfiguration.JobVerificationKeyPath == "" {
+	// 	// We won't be verifying jobs, so it doesn't matter
+	// 	return "if you're seeing this string, there's a problem with the job verification code in the agent. contact support@buildkite.com", nil
+	// }
```
moskyb, line 387: remove this

clicommand/pipeline_upload.go
```diff
@@ -283,25 +304,68 @@ var PipelineUploadCommand = cli.Command{
...
+				}
+
+				l.Fatal("Algorithm %s (%s) cannot be used for signing/verification", signingAlg, from)
+			}
+
+			if err := validateJWK(key, signingAlg); err != nil {
+				l.Fatal("Signing key ID %s is invalid: %v", cfg.SigningKeyID, err)
+			}
+
+			key.Set(jwk.AlgorithmKey, signingAlg)
```
moskyb, line 366: this algorithm wrangling might be the tricksiest part of the PR - currently, we allow users to either specify an algorithm in their JWK (under the `alg` key, supported by JWK), or allow them to specify an algorithm by passing the `--signing-algorithm` cli arg.

whatever happens, we end up validating that the key they've given us is capable of using that algorithm, then, if it wasn't already set, we set the key's `alg` to what they've passed us. it's a little complicated, but when we sign the pipeline later on, it allows us to just call `jws.Sign([]byte{"blah"}, jws.WithKey(key.Algorithm(), key)` rather than sidechanneling it all the way through. happy to update this logic, as it's a little bit complex

internal/pipeline/sign.go
```diff
@@ -56,34 +59,32 @@ func Sign(sf SignedFielder, signer Signer) (*Signature, error) {
...
 	values, err := sf.ValuesForFields(s.SignedFields)
 	if err != nil {
 		return fmt.Errorf("obtaining values for fields: %w", err)
 	}
 
-	// Ensure this part is the same as in Sign...
-	defer verifier.Reset()
-	writeLengthPrefixed(verifier, verifier.AlgorithmName())
-	if _, err := writeFields(verifier, values); err != nil {
+	signatureMaterial := &bytes.Buffer{}
```
DrJosh9000, line 72: If you repaint the front half of the bikeshed, repaint the back half too.

clicommand/pipeline_upload.go
```diff
@@ -398,3 +462,22 @@ func searchForSecrets(l logger.Logger, cfg *PipelineUploadConfig, environ map[st
 		}
 	}
 }
+
+func validateJWK(key jwk.Key, alg jwa.SignatureAlgorithm) error {
+	validKeyTypes := []jwa.KeyType{jwa.RSA, jwa.EC, jwa.OctetSeq}
```
DrJosh9000, line 467: Add a TODO to investigate OKP?

clicommand/pipeline_upload.go
```diff
@@ -398,3 +462,22 @@ func searchForSecrets(l logger.Logger, cfg *PipelineUploadConfig, environ map[st
...
+	if !slices.Contains(validKeyTypes, key.KeyType()) {
+		return fmt.Errorf("unsupported key type %s. Key type must be one of %v", key.KeyType(), validKeyTypes)
+	}
+
+	validAlgsForType := map[jwa.KeyType][]jwa.SignatureAlgorithm{
```
DrJosh9000, line 472: Add a note in a comment that this list deliberately excludes PKCS 1.5 (because it is discouraged, etc)

And probably also worth a note that we're waiting on the library to support Ed25519 (or whatever the JOSE folks are calling it)

agent/integration/job_verification_integration_test.go
```diff
@@ -203,42 +175,51 @@ func TestJobVerification(t *testing.T) {
...
+		t.Fatalf("setting kid: %v", err)
+	}
 
-	finish := e.finishesFor(t, job.ID)[0]
+	return key
+}
 
-	if got, want := finish.ExitStatus, "-1"; got != want {
-		t.Errorf("job.ExitStatus = %q, want %q", got, want)
+func setOf(t *testing.T, jwkes ...jwk.Key) jwk.Set {
```
DrJosh9000, line 199: [more idle bike-shedding]

set of what exactly?

Hungarian notation maybe - `jwksFromKeys(...)`?

DrJosh9000 (approved):
This is pretty great! 🎆🎇

I have mostly cosmetic comments for you.

clicommand/pipeline_upload.go
moskyb: so it turns out that OKP is Ed25519! easy peasy, i'll add that in today.

clicommand/pipeline_upload.go
moskyb: so it turns out that i had intended (at one point?) to not include RSA-PKCSv1.5 in this list, but it's totally in there at the moment - i'll clean it out. see my comment above about Ed25519 - it's very doable!

internal/pipeline/sign.go
```diff
@@ -24,30 +22,35 @@ type Signature struct {
 
 // Sign computes a new signature for an object containing values using a given
 // signer. Sign resets the signer after use.
-func Sign(sf SignedFielder, signer Signer) (*Signature, error) {
+func Sign(sf SignedFielder, key jwk.Key) (*Signature, error) {
+	signatureMaterial := &bytes.Buffer{}
```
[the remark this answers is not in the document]
moskyb: payloadddddd that's so much better how did i miss that

clicommand/pipeline_upload.go
```diff
@@ -283,25 +304,68 @@ var PipelineUploadCommand = cli.Command{
 			searchForSecrets(l, &cfg, envMap, result, src)
 		}
 
-		if cfg.SigningKeyPath != "" {
+		if cfg.JWKSFilePath != "" {
```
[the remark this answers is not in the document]
moskyb: donezo - extracted and tested :)

DrJosh9000 (approved):
🌟🌟🌟🚢

PR #2253, opened 2023-07-28
Even More Pipeline Parsing Fixes

moskyb:
**This PR:** Fixes two issues in the pipeline parser:
- [x] Block steps represented by a bare string were causing pipeline parsing to fail. Now, they'll be accepted and transformed into the equivalent `{"block": ""}`
- [x] We were only ever inferring step type from attributes (eg. command steps have a command attribute. Turns out, step types can be explicit too (eg. `type: "block"`). We should also support these steps.

### DIFF
--- a/.buildkite/steps/test-go-fmt.sh
+++ b/.buildkite/steps/test-go-fmt.sh
@@ -8,8 +8,12 @@ if [[ $(gofmt -l ./ | head -c 1 | wc -c) != 0 ]]; then
   exit 1
 fi
 
-if [[ $(go mod tidy -v 2> >(wc -c)) != 0 ]]; then # go mod tidy -v outputs to stderr for some reason
+tidy_output=$(go mod tidy -v 2>&1)
+
+if [[ "${#tidy_output}" -gt 0 ]]; then # go mod tidy -v outputs to stderr for some reason
   echo "The go.mod file is out of sync with the source code"
+  echo "Output of \`go mod tidy -v\`:"
+  echo "$tidy_output"
   echo "Please run \`go mod tidy\` locally, and commit the result."
   exit 1
 fi
--- a/internal/pipeline/maybe_scalar_steps.go
+++ b/internal/pipeline/maybe_scalar_steps.go
@@ -1,0 +1,112 @@
+package pipeline
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+
+	"github.com/buildkite/agent/v3/internal/ordered"
+	"github.com/buildkite/interpolate"
+)
+
+// In the buildkite pipeline yaml, some step types (broadly, wait steps, input steps and block steps) can be represented
+// either by a scalar string (ie "wait") or by a mapping with keys and values and such
+//
+// This behaviour is difficult to cleanly model in go, which leads to the somewhat odd structure of the structs defined
+// in this file - each type (WaitStep, InputStep) has a Scalar field which is set to the scalar value if the step was
+// if, during pipeline parsing, the step was represented as a scalar, and is left empty if the step was represented as
+// a mapping. In essence, if one of the fields of these structs is filled, the other should be empty.
+//
+// On the unmarshalling side, the differing types is handled by the steps parser - see the unmarshalStep() function in
+// ./steps.go - it infers the underlying type of the thing it's unmarshalling, and if it's a string, calls NewScalarStep()
+// (below) to create the appropriate struct. If it's a mapping, it creates the appropriate struct directly.
+//
+// On the marshalling side, the MarshalJSON() function on each struct handles the different cases. In general, if the
+// Scalar field is set, it marshals that, otherwise it marshals the other fields.
+//
+// In reading this file, you may have noticed that I mentioned that there are three types of step that can be represented
+// as a scalar, but there are only two structs defined here. This is because the third type, block steps, are represented
+// in exactly the same way as input steps, so they can share the same struct. This is liable to change in the future,
+// as conceptually they're different types, and it makes sense to have them as different types in go as well.
+//
+// Also also! The implementations for WaitStep and InputStep **almost**, but not quite identical. This is due to the behaviour
+// of marshalling an empty struct into into JSON. For WaitStep, it makes sense that the empty &WaitStep{} struct marshals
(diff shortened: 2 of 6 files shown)


.buildkite/steps/test-go-fmt.sh
```diff
@@ -10,6 +10,8 @@ fi
 
 if [[ $(go mod tidy -v 2> >(wc -c)) != 0 ]]; then # go mod tidy -v outputs to stderr for some reason
   echo "The go.mod file is out of sync with the source code"
+  echo "go mod tidy had output:"
+  go mod tidy -v
```
triarius, line 14: 💜

test.yml
triarius, line 1: stray file?

internal/pipeline/steps.go
```diff
@@ -43,45 +43,84 @@ func (s Steps) interpolate(env interpolate.Env) error {
...
-		case o.Contains("block") || o.Contains("input") || o.Contains("manual"):
-			step = make(InputStep)
+	default:
+		return nil, fmt.Errorf("unmarshaling step: unsupported type %T", o)
+	}
+}
 
-		case o.Contains("trigger"):
-			step = make(TriggerStep)
+func stepFromMap(o *ordered.MapSA) (Step, error) {
```
triarius, line 60: Nice!

triarius (commented):
LGTM, but I think there's a stray file you didn't intend to commit?

test.yml
moskyb: oop 😶

internal/pipeline/parser_test.go
```diff
@@ -392,6 +427,57 @@ steps:
...
+    "input"
+  ]
+}`
+
+	if diff := cmp.Diff(string(gotJSON), wantJSON); diff != "" {
+		t.Fatalf("marshalled JSON diff (-got +want):\n%s", diff)
+	}
+}
+
+// func
```
DrJosh9000, line 479: funcy

DrJosh9000 (approved):
Thanks for tackling this!

internal/pipeline/parser_test.go
moskyb: double oop 😶

internal/pipeline/step.go
```diff
@@ -29,6 +30,29 @@ type Step interface {
 	selfInterpolater
 }
 
+type ScalarStep string
```
DrJosh9000, line 33: My feelings about the types here: I kinda want for it so that the types are easy to assemble in order to produce pipelines with Go. "ScalarStep" describes "a shape of data" more than "a type of step", and also overlaps semantic meaning with WaitStep and InputStep. But it's not too hard to use.

moskyb: yeah, i had this thought as well... the other idea i had was something like 
```Go
type WaitStep struct {
	Scalar string `yaml:"-"`
	RemainingFields map[string]any `yaml:",inline"`
}
```
and then have the marshal return the value of `scalar` if it's defined, and a regular marshal if not. is that something that feels better to you?

DrJosh9000: Yeah sort of. I feel like if someone wants to write `WaitStep{}` it should marshal as `"wait"`, because that's the kind of step it is.

moskyb: this is updated in https://github.com/buildkite/agent/pull/2253/commits/a9621be2e078f366c3acd5dd1a266277115dcbf1 - would love your thoughts!

internal/pipeline/maybe_scalar_steps.go
```diff
@@ -0,0 +1,115 @@
...
+//
+// In reading this file, you may have noticed that I mentioned that there are three types of step that can be represented
+// as a scalar, but there are only two structs defined here. This is because the third type, block steps, are represented
+// in exactly the same way as input steps, so they can share the same struct. This is liable to change in the future,
+// as conceptually they're different types, and it makes sense to have them as different types in go as well.
+//
+// Also also! The implementations for WaitStep and InputStep **almost**, but not quite identical. This is due to the behaviour
+// of marshalling an empty struct into into JSON. For WaitStep, it makes sense that the empty &WaitStep{} struct marshals
+// to "wait", but with InputStep, there's no way to tell whether it should be marshalled to "input" or "block", which
+// have very different behaviour on the backend.
```
DrJosh9000, line 28: Excellent commentary 👏👏

internal/pipeline/maybe_scalar_steps.go
```diff
@@ -0,0 +1,115 @@
...
+
+func (s *WaitStep) unmarshalMap(m *ordered.MapSA) error {
+	if s.Contents == nil {
+		s.Contents = make(map[string]any, m.Len())
+	}
+
+	return m.Range(func(k string, v any) error {
+		s.Contents[k] = v
+		return nil
+	})
```
DrJosh9000, line 76: I just remembered that I added a `ToMap` method to `ordered.Map` -

so this _could_ be:

```suggestion
	s.Contents = m.ToMap()
```

internal/pipeline/maybe_scalar_steps.go
```diff
@@ -0,0 +1,115 @@
...
+// If scalar is empty, it marshals as the remaining fields
+func (s *WaitStep) MarshalJSON() ([]byte, error) {
+	if s.Scalar == "" {
+		if len(s.Contents) == 0 {
+			return json.Marshal("wait")
+		}
+		return json.Marshal(s.Contents)
+	}
+
+	return json.Marshal(s.Scalar)
```
DrJosh9000, line 61: Very low-priority hobby-horse of mine: 

```suggestion
	if s.Scalar != "" {
	  return json.Marshal(s.Scalar)
	}
	if len(s.Contents) == 0 {
		return json.Marshal("wait")
	}
	return json.Marshal(s.Contents)
```

DrJosh9000 (approved):
Good stuff!

internal/pipeline/maybe_scalar_steps.go
moskyb: 🙈 i can't believe i missed this

internal/pipeline/maybe_scalar_steps.go
moskyb: oh that's way easier

triarius (approved):

PR #2254, opened 2023-07-31
Add future-proofing UnknownStep type

DrJosh9000:
It occurred to me that the new parser has no way of representing steps that we haven't invented yet.

### DIFF
--- a/internal/pipeline/parser_test.go
+++ b/internal/pipeline/parser_test.go
@@ -615,6 +615,46 @@ func TestParserParsesTopLevelSteps(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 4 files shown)


triarius (commented):
LGTM, but I also think we should backtrack errors from unmarshalling known step types and fall back to presenting the step as an `UnknownStep` then as well. It will mean we can't sign the step, though, so maybe we should refuse to sign pipelines with unknown steps and display an error message to contact support.

moskyb:
hm, great minds: https://github.com/buildkite/agent/pull/2253/commits/fef91e32054d62caed2601ab7453fc2e09cbce37

(they're similar, if not actually the same)

triarius (approved):

PR #2255, opened 2023-07-31
Get mime types from github, rather than Apache's SVN Server

moskyb:
As of time of writing (2023-07-31), the Apache Foundation has an expired SSL cert on svn.apache.org, which is where we source our list of mime types as defined by httpd.

As a workaround, fetch them from the github mirror of the httpd source code instead


DrJosh9000 (approved):

triarius (approved):
💜

PR #2256, opened 2023-07-31
Store GOCACHE outside container

DrJosh9000:
This should enable test caching between runs on the same machine, which should slightly reduce time to run the pipeline.


moskyb (approved):

triarius (approved):
LGTM

PR #2260, opened 2023-07-31
Nil handling tweaks

DrJosh9000:
- Someone could write `env: null`, and this should be handled appropriately.
- Most of `Steps.unmarshalAny` should be skipped if `o == nil`, not just if `*s == nil` also.
- `ToMap` should handle nil receiver (like all the other methods).

### DIFF
--- a/internal/ordered/map.go
+++ b/internal/ordered/map.go
@@ -170,8 +170,12 @@ func (m *Map[K, V]) Delete(k K) {
 	}
 }
 
-// ToMap creates a regular (un-ordered) map containing the same data.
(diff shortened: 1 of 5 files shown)


moskyb (approved):

triarius (commented):
LGTM, but you should write a test case for this

DrJosh9000:
> LGTM, but you should write a test case for this

Added a parser test for `env: null` and a `ToMap` test. 

The other thing I found while doing that is defensive/an unlikely edge case (`Pipeline.Steps` would have to be non-nil before parsing, which we don't do).

moskyb (approved):

PR #2265, opened 2023-08-02
Fix data race in testAgentEndpoint

DrJosh9000:
I think racy access to `tae.calls` is being detected by the race detector.


moskyb (approved):
whoops - i think i actually have this change on #2210, but it hasn't made it into main yet

triarius (approved):

PR #2267, opened 2023-08-02
Reorganise step types among files

DrJosh9000:
I thought about continuing "(type)_step.go", but I think that grouping them under a common prefix makes them a little easier to see in file listings.

I also separated input, wait, and scalar (which includes both input and wait), but I'm OK to be pushed back on that.


moskyb (approved):
looks great 🎉

i especially love the `x_step.go` -> `step_x.go` change - way easier to find in file lists 🥳

PR #2268, opened 2023-08-03
Add --strict-single-hooks

DrJosh9000:
This is a partial revival of #1135 (successor to #1055), rolled back in #1161. It attempted to fix the discrepancy between only running one of plugin/local/global command/checkout hook, but allowing multiple _plugins_ to define a command/checkout hook (and running all of them).

Sadly both #1055 and #1135 broke enough pipelines that it had to be rolled back, but the dangers associated with letting multiple different things override checkouts and commands remain. ("Rock and a hard place")

This PR does two things differently to #1135:

* Wrap the enforcement of single hook types under a flag, and just log a warning and carry on otherwise if the flag is not true.
* Do the check after testing the plugin for containing the hook - not doing this was a bug in the previous attempt (fixed in #1156).

### DIFF
--- a/EXPERIMENTS.md
+++ b/EXPERIMENTS.md
@@ -97,3 +97,11 @@ We now understand this causes a bug, and we want to avoid it. Enabling this expe
 https://github.com/buildkite/agent/blob/40b8a5f3794b04bd64da6e2527857add849a35bd/internal/job/executor.go#L1980-L1993
 
 **Status:** Since the default behaviour is buggy, we will be promoting this to non-experiment soon™️.
+
+### `strict-single-hooks`
+
+We document that multiple checkout and command hooks will not run, only the first encountered will. A long-standing bug means this is not true for checkout and command hooks provided by plugins.
+
+This experiment enforces only running one checkout or command hook, even if multiple plugins define them. 
+
+**Status:** This experiment fixes a long-standing bug, so we intend to promote this to non-experiment when most pipelines and plugins have been updated.
--- a/EXPERIMENTS.md
+++ b/EXPERIMENTS.md
@@ -97,11 +97,3 @@ We now understand this causes a bug, and we want to avoid it. Enabling this expe
 https://github.com/buildkite/agent/blob/40b8a5f3794b04bd64da6e2527857add849a35bd/internal/job/executor.go#L1980-L1993
 
 **Status:** Since the default behaviour is buggy, we will be promoting this to non-experiment soon™️.
-
-### `strict-single-hooks`
-
-We document that multiple checkout and command hooks will not run, only the first encountered will. A long-standing bug means this is not true for checkout and command hooks provided by plugins.
-
-This experiment enforces only running one checkout or command hook, even if multiple plugins define them. 
-
-**Status:** This experiment fixes a long-standing bug, so we intend to promote this to non-experiment when most pipelines and plugins have been updated.
--- a/internal/job/executor.go
+++ b/internal/job/executor.go
@@ -850,18 +850,40 @@ func (e *Executor) VendoredPluginPhase(ctx context.Context) error {
 	return e.executePluginHook(ctx, "environment", vendoredCheckouts)
 }
 
+var singleHooks = map[string]bool{
+	"command":  true,
+	"checkout": true,
+}
+
 // Executes a named hook on plugins that have it
 func (e *Executor) executePluginHook(ctx context.Context, name string, checkouts []*pluginCheckout) error {
-	for _, p := range checkouts {
+	// Command and checkout hooks are a little different, in that we only execute
+	// the first one we see. We run the first one, and output a warning for all
+	// the subsequent ones.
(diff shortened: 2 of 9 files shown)


DrJosh9000:
Adding @pda specifically because of #1161.

internal/job/executor.go
```diff
@@ -850,18 +850,40 @@ func (e *Executor) VendoredPluginPhase(ctx context.Context) error {
...
+			} else {
+				e.shell.Warningf("The additional %s hook (%s plugin, position %d) will be ignored in a future version of the agent. "+
+					"To enforce single %s hooks now, enable the %q experiment", name, p.Plugin.Name(), i+1, name, experiments.StrictSingleHooks)
+			}
+		}
+		hookTypeSeen[name] = true
+
 		envMap, err := p.ConfigurationToEnvironment()
-		if dnerr := (&plugin.DeprecatedNameErrors{}); err != nil && errors.As(err, &dnerr) {
+		if dnerr := (&plugin.DeprecatedNameErrors{}); errors.As(err, &dnerr) {
```
triarius, line 886: 😅

internal/job/executor.go
```diff
@@ -850,18 +850,40 @@ func (e *Executor) VendoredPluginPhase(ctx context.Context) error {
...
+	// the subsequent ones.
+	hookTypeSeen := make(map[string]bool)
+
+	for i, p := range checkouts {
 		hookPath, err := hook.Find(p.HooksDir, name)
 		if errors.Is(err, os.ErrNotExist) {
 			continue // this plugin does not implement this hook
-		} else if err != nil {
+		}
+		if err != nil {
```
triarius, line 870: 😅

internal/job/executor.go
```diff
@@ -850,18 +850,40 @@ func (e *Executor) VendoredPluginPhase(ctx context.Context) error {
 	return e.executePluginHook(ctx, "environment", vendoredCheckouts)
 }
 
+var singleHooks = map[string]bool{
```
triarius, line 853: I think you need to give this a better name, or explain how it is used with a comment. Using the name of the experiment, `strict-single-hooks`, in some way would make this a bit more clear.

EXPERIMENTS.md
```diff
@@ -97,3 +97,11 @@ We now understand this causes a bug, and we want to avoid it. Enabling this expe
...
 
 **Status:** Since the default behaviour is buggy, we will be promoting this to non-experiment soon™️.
+
+### `strict-single-hooks`
+
+We document that multiple checkout and command hooks will not run, only the first encountered will. A long-standing bug means this is not true for checkout and command hooks provided by plugins.
+
+This experiment enforces only running one checkout or command hook, even if multiple plugins define them. 
+
+**Status:** This experiment fixes a long-standing bug, so we intend to promote this to non-experiment when most pipelines and plugins have been updated.
```
triarius, line 107: I don't think this condition will ever be met, and if it is, do we have any way of knowing?

It's also possible that for a large class of users, plugins running multiple checkout hooks is a feature they use to check out multiple repos with internal plugins, and/or multiple command hooks on plugins are a way to ensure multiple commands are run on the same agent in the same working directory without cleaning up in between.

For these reasons, I suggest ripping the band-aid off only with a major version bump.

EXPERIMENTS.md
```diff
@@ -97,3 +97,11 @@ We now understand this causes a bug, and we want to avoid it. Enabling this expe
 https://github.com/buildkite/agent/blob/40b8a5f3794b04bd64da6e2527857add849a35bd/internal/job/executor.go#L1980-L1993
 
 **Status:** Since the default behaviour is buggy, we will be promoting this to non-experiment soon™️.
+
+### `strict-single-hooks`
```
triarius, line 101: Can this be a flag instead of an experiment?

EXPERIMENTS.md
triarius: My envisioned lifecycle for this is:

v3:
`--strict-single-hook=false` is the default

v4: ???
`--strict-single-hook=true` is the default

v5:
remove the flag

DrJosh9000: I gave it some thought, and I think I agree with you.

triarius (approved):
LGTM, the flag being `false` should mean this is opt in

pda (approved):
Seems good 👍🏼

I can't remember exactly why we decided multiple hooks was Bad, but I can imagine some reasons. E.g. perhaps two plugins each wrap the underlying step's command, so it ends up being run twice.

I like the approach of warning of deprecation with opt-in to the upcoming enforced behaviour 👍🏼

triarius:
> I can't remember exactly why we decided multiple hooks was Bad, but I can imagine some reasons. E.g. perhaps two plugins each wrap the underlying step's command, so it ends up being run twice.

I think for command it's not so bad, and useful for some people even. But for checkout, it sure seems unnecessary to check out in two different ways twice.

PR #2269, opened 2023-08-03
Remove unwrapping in error `Is` methods

triarius:
I was doing some "light reading" of the go standard library, and noticed [this](https://cs.opensource.google/go/go/+/refs/tags/go1.20.7:src/errors/wrap.go;drc=4a0a2b33dfa3c99250efa222439f2c27d6780e4a;l=41):
```
// ... An Is method should only shallowly
// compare err and the target and not call Unwrap on either.
```

In the former implementations, calling `errors.As` could unwrap the errors, so this PR fixes this.

### DIFF
--- a/agent/plugin/error.go
+++ b/agent/plugin/error.go
@@ -18,14 +18,14 @@ func (e *DeprecatedNameErrors) IsEmpty() bool {
 	return e == nil || len(e.errs) == 0
 }
 
-// Errors returns the contained set of errors in sorted order
-func (e *DeprecatedNameErrors) Errors() []DeprecatedNameError {
+// Unwrap returns the a slice of errors in a stable order
+func (e *DeprecatedNameErrors) Unwrap() []error {
 	if e == nil {
 		return nil
 	}
 
 	if e.errs == nil {
-		return []DeprecatedNameError{}
+		return []error{}
 	}
 
 	errs := make([]DeprecatedNameError, 0, len(e.errs))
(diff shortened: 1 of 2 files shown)


agent/plugin/error.go
DrJosh9000, line 1: While we're here, you could update `Errors` to be `Unwrap` that returns `[]error`. I don't think any call sites of `Errors` use anything other than the `Error` method.

DrJosh9000 (approved):
Otherwise LGTM!

agent/plugin/error.go
```diff
@@ -41,13 +40,18 @@ func (e *DeprecatedNameErrors) Errors() []DeprecatedNameError {
 		return errs[i].old < errs[j].old
 	})
 
-	return errs
+	out := make([]error, 0, len(errs))
+	for i := range errs {
+		out = append(out, &errs[i])
+	}
```
DrJosh9000, line 46: I guess the alternative is to make `errs` have type `[]error`, then type-assert everything in the `sort.Slice` closure...

Hang on, does anything use the order of `e.errs`?

agent/plugin/error.go
triarius: Done

agent/plugin/error.go
triarius: I think it's nice for the user. I wrote tests for it, so they will break

triarius: Hmm. Maybe I didn't 🤷‍♂️. Still think it's nice for the user.

PR #2277, opened 2023-08-08
Refactor redactor into replacer

DrJosh9000:
This changes `redactor` into `replacer`. A summary:

1. The minimum redaction length, `ValuesToRedact`, and `VarsToRedact` are moved into a new package, `redact`
2. The `Redactor` and necessary friends are renamed and moved into `replacer`.
4. `pipeline_upload.go` is changed to use a `Replacer` to search through the JSON in a streaming manner.
5. The implementation of `flushUpTo` is changed to call a callback when processing matches instead of using a fixed `[]byte`.
6. `flushUpTo` also no longer eagerly flushes into the middle of matches.
7. Basically everything else is mechanical.


internal/replacer/replacer.go
```diff
@@ -190,51 +200,56 @@ func (r *Redactor) Flush() error {
...
-			fallthrough
+		}
 
-		case bufidx == match.from:
-			// A redacted range.
-			// Write a r.subst instead of the redacted range.
-			if _, err := r.dst.Write(r.subst); err != nil {
+		// Now handle the match itself.
+		// Call the replacement callback to get a replacement.
+		if repl := r.replacement(r.buf[match.from:match.to]); len(repl) > 0 {
```
moskyb, line 243: is "replace the thing matched with an empty string" (ie delete the matched string) a valid use case? i can see the semantics there being a little confusing if the user tried to do that

DrJosh9000: That should be supported - `replacement` should return `nil`. The `if` on this line is just to skip calling `r.dst.Write` with nil or empty.

DrJosh9000:
I'm going to add some more tests, since there is 0 coverage in most of `clicommand`.

internal/replacer/replacer.go
moskyb: sweet as

triarius (approved):
Nice. I think I know how to use this now.

PR #2279, opened 2023-08-09
Add ordered.Unmarshal, and use it

DrJosh9000:
The main affordance of `encoding/json`, `gopkg.in/yaml.v3`, etc is the ability to unmarshal data into a struct. This PR adds similar functionality to the `ordered` package.
    
Because `DecodeYAML` returns a variety of types, `Unmarshal` handles `*MapSA`, `[]any`, and a variety of scalars as sources.

The second commit applies the new `ordered.Unmarshal` to pipeline parsing.

### DIFF
--- a/internal/ordered/unmarshal.go
+++ b/internal/ordered/unmarshal.go
@@ -1,0 +1,375 @@
+package ordered
+
+import (
+	"errors"
+	"fmt"
+	"reflect"
+	"strings"
+
+	"gopkg.in/yaml.v3"
+)
+
+// Errors that can be returned by Unmarshal
+// (typically wrapped - use errors.Is).
+var (
+	ErrIntoNonPointer       = errors.New("cannot unmarshal into non-pointer")
+	ErrIntoNil              = errors.New("cannot unmarshal into nil")
+	ErrIncompatibleTypes    = errors.New("incompatible types")
+	ErrUnsupportedSrc       = errors.New("cannot unmarshal from src")
+	ErrMultipleInlineFields = errors.New(`multiple fields tagged with yaml:",inline"`)
+)
+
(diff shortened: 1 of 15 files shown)


triarius (approved):
Amazing!

internal/ordered/unmarshal.go
```diff
@@ -0,0 +1,432 @@
...
+
+	"gopkg.in/yaml.v3"
+)
+
+var (
+	errNotAPointer          = errors.New("cannot unmarshal into non-pointer")
+	errIntoNil              = errors.New("cannot unmarshal into nil")
+	errIncompatibleTypes    = errors.New("incompatible types")
+	errUnsupportedSrc       = errors.New("cannot unmarshal from src")
+	errMultipleInlineFields = errors.New(`multiple fields tagged with yaml:",inline"`)
```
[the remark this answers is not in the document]
DrJosh9000: This package is getting closer and closer to being useful enough to break out into its own repo, so I think I'll export these errors now.

internal/ordered/unmarshal.go
```diff
@@ -0,0 +1,375 @@
...
+	outlineKeys := make(map[string]struct{})
+
+	for _, field := range fields {
+		// Skip non-exported fields. This is conventional *and* correct.
+		if !field.IsExported() {
+			continue
+		}
+
+		// No worries if the tag is not there - apply defaults.
+		tag, _ := field.Tag.Lookup("yaml")
```
moskyb, line 281: were we to export this into its own library, would we want to give this its own struct tag? or is it too tied into the yaml.v3 ecosystem?

DrJosh9000: Yes, I think having its own tag would be an appropriate next step.

PR #2282, opened 2023-08-09
Create a panic log level

triarius:
Unlike the fatal log level, this will allow defers to run.

I'm mainly putting this up for discussion, I don't intend to merge as is.

We have a few options here, but the outcome that MUST happen is that `defer`s from cli command actions must be run when an error is encountered. Some options are:
1. This PR, always use `l.Panic`
2. Modify this PR to modify the `FATAL` log level to behave like `PANIC` does and remove panic
3. Always return errors from methods called by cli command actions, return those errors up to the main function and determine how they should be handled there.

3 is the most idiomatic solution, but it requires diving deep into the code and eliminating all `os.Exit` and `l.Fatal` calls. And also, it is likely that not all methods that call `os.Exit` or `l.Fatal` return errors that make it all the way to main.go. We would have to change the method signature and how their return values are handled if they call `l.Fatal`.

2 seems like the most pragmatic choice to me.

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -667,7 +667,7 @@ var AgentStartCommand = cli.Command{
 
 		// Remove any config env from the environment to prevent them propagating to bootstrap
 		if err := UnsetConfigFromEnvironment(c); err != nil {
-			l.Fatal("Failed to unset config from environment: %v", err)
+			l.Panic("Failed to unset config from environment: %v", err)
 		}
 
 		if cfg.JobVerificationKeyPath != "" {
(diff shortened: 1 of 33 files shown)


DrJosh9000:
Two big reasons to avoid `panic` in production code are the user-unfriendliness of its output if it reaches the runtime (and then stderr), and the existence of latent `recover` calls (that we don't write ourselves) - it's hard to guarantee that we will be the ones recovering any particular panic. I fondly remember a bug where a former colleague managed to convert any and all panics in a large section of a server into eventual deadlocks, and the crux was that [`fmt` recovers panics from within `Error` and `String` methods in order to format them](https://pkg.go.dev/fmt#hdr-Format_errors).

Exiting immediately has a purpose: prevent corrupted state from ruining user data by letting the OS clean up. If it's a service, then `systemd` or `launchd` or whatever can restart the process.

If we need a scheme where some cleanup code should run, like a defer, but don't want to thoroughly pass through errors up to main, then I would rather create a custom way to register cleanup code to run on exit (sort of like [`testing.T.Cleanup`](https://pkg.go.dev/testing#T.Cleanup)), which is used for normal-case `log.Exit` and ignored for `log.Fatal`.

triarius:
Fair enough. I took some steps to ensure that we would only recover from panics that we created, but I didn't anticipate that code in other packages might not respect that and recover panics we created.

I did a search, and it looks like we very rarely use `log.Fatal` outside the clicommand package. So I think we can pass up errors and deal with them in `main.go`.

PR #2283, opened 2023-08-09
Add a full agent version which includes the commit

triarius:
It looks like this when you build it locally:
```shell
❯ buildkite-agent --version
buildkite-agent version 3.54.0+x.87434487dac153b40549188b5885e6f95d080707
```
when built on CI, the `x` will be replaced with the Buildkite build number.

Previously, it would look like this:
```shell
❯ buildkite-agent --version
buildkite-agent version 3.54.0, build x
```

You can verify this by downloading the binary built in the build.

When built without all changes committed, it will look like this:
```shell
❯ buildkite-agent --version
buildkite-agent version 3.54.0+x.87434487dac153b40549188b5885e6f95d080707.dirty
```
and I've checked that the binary built in the build does not have this suffix.

While this is verbose, as it is a "full" version, I have not abbreviated it.

I've used the full version in the debug logs and the `--version` flag on the CLI. There are other places that a version is reported that I have decided not to change the version string for. Most notably: the User Agent and the version reported when registering the agent with the Buildkite API.

I think when we increase a major version, or otherwise change the versioning scheme, we should start to use the full version, or at least a version string that displays an abbreviation of the commit hash, in more places.

I've added a check to CI that the version is not for the version that is compiled in the build step. Only one build OS/Arch combination is tested. At the moment it is Linux amd64.

While I was in the area, I changed the CI step DAG to make the release depend on checks like this and go fmt, etc passing. I also saw that the test failure annotation step seems to only run when the tests pass, which defeats its purpose.


.buildkite/pipeline.yml
```diff
@@ -90,6 +90,7 @@ steps:
       - test-race-linux-arm64
       - test-linux-arm64
       - test-windows
+    allow_dependency_failure: true
```
triarius, line 93: Without this, the test failure annotation would not run if any test failed, which seems to be the opposite of its purpose.

.buildkite/pipeline.yml
```diff
@@ -248,11 +254,14 @@ steps:
   - name: ":pipeline: Upload Release Pipeline"
     key: upload-release-steps
     depends_on:
+      - check-go-fmt-mod-tidy
+      - check-code-generation
+      - check-version-string
```
triarius, line 259: We should run these before release.

version/version.go
```diff
@@ -17,19 +19,45 @@ import (
...
-var buildVersion string
+var buildNumber string
 
 func Version() string {
 	return strings.TrimSpace(baseVersion)
 }
 
-func BuildVersion() string {
-	if buildVersion == "" {
+func BuildNumber() string {
```
triarius, line 28: The old name is a bit ambiguous. `BuildVersion` could be `v3.53.6992`, but it was just `6992`. `BuildNumber` does not have this ambiguity.

.buildkite/steps/check-version-string.sh
DrJosh9000, line 1: This is a note of something I was wondering, then figured out.

I wondered why this script didn't always fail, because at least part of each binary build is to run `./scripts/generate-acknowledgements.sh`, which creates a new file that is then embedded into the binary, so surely every build is `dirty`. The answer:

`runtime/debug.ReadBuildInfo`'s `vcs.modified` info is based on `go build` running `git status --porcelain`, and its output is empty is because ACKNOWLEDGEMENTS.md.gz is conveniently in the .gitignore.

.buildkite/pipeline.yml
```diff
@@ -4,8 +4,8 @@ agents:
   queue: agent-runners-linux-amd64
 
 steps:
-  - name: ":go: go fmt"
-    key: test-go-fmt
+  - name: ":go: Check go-fmt and go-mod-tidy"
```
DrJosh9000, line 7: What do you think about combining this check with the generated files check?

DrJosh9000 (approved):
This is great!

.buildkite/pipeline.yml
triarius: It's a bit risky b/c the check for `gofmt` does not use git to check if the working tree is clean, but the check for `go generate` does. So the earlier test could pass but leave the working tree dirty, which will be a false positive for the latter test. I don't think that condition is activated, but I think it's safer to make all 3 checks assert that they are no-ops on the working directory (for committed files).

version/version.go
```diff
@@ -17,19 +19,48 @@ import (
...
+	var commit string
+	for _, setting := range info.Settings {
+		switch setting.Key {
+		case "vcs.revision":
+			commit = setting.Value
+		case "vcs.modified":
+			if setting.Value == "false" {
+				dirty = ""
+			}
+		}
```
triarius, line 51: I tend to prefer whatever has fewer indents. Here, that was the if/else if chain.

But it's a not opinion I hold very strongly.

version/version.go
```diff
@@ -3,33 +3,79 @@ package version
...
+// TODO-vNext: Include some prefix of the commit hash in the User-Agent header.
 func UserAgent() string {
-	return "buildkite-agent/" + Version() + "." + BuildVersion() + " (" + runtime.GOOS + "; " + runtime.GOARCH + ")"
+	return fmt.Sprintf(
+		"buildkite-agent/%s.%s (%s; %s)",
+		Version(),
+		BuildNumber(),
+		runtime.GOOS,
+		runtime.GOARCH,
+	)
```
triarius, line 80: Yeah, this looks much better!

DrJosh9000 (approved):
🚀

PR #2284, opened 2023-08-10
Bump go to 1.21

triarius:
I also increased the frequency of dependabot updates because dependabot should have done this, not me.

But since I was doing it, I removed the use of golang.org/x/exp/slices as it is in the stdlib now.


DrJosh9000:
I believe Go 1.21 needs Windows >= 10 and macOS >= 10.15. 

Probably time to clean up the README...

triarius:
Taking this back to the drawing board for now.

moskyb:
> I believe Go 1.21 needs Windows >= 10 and macOS >= 10.15.

i think we should investigate whether this means that the go toolchain requires these OSes, or if it's that go binaries require these OSes - i've always thought that it was the former, and if so, that's a development concern and totally fine, IMO.

if on the other hand binaries compiled by Go 1.21 require Windows >= 10 and macOS >=10.15, then we probably have a problem

DrJosh9000:
I have changed the policy that is written in the README. But. We need to keep the pin in this until

- we have let the change soak a bit, in case we need to change it again
- we have a better understanding of who still needs Windows Server 2012 R2, macOS 10.13 and macOS 10.14
- we've reached out to potentially impacted customers to see what their needs are

triarius:
@moskyb should we update the gomod to say `go 1.21` too?

moskyb:
oooh yes, good call

triarius (commented):
LGTM @moskyb 🎉

I can't approve it on GitHub as I created the PR, maybe if you approve it you can unblock the merge.

moskyb (approved):

PR #2286, opened 2023-08-13
Add more t.Parallel to shell tests

triarius:
Our previous attempt to do so in [#2218](https://github.com/buildkite/agent/pull/2218) did not succeed because the output from the command that did not make it into the error.

So, using the streaming "replacer" introduced by @DrJosh9000 in https://github.com/buildkite/agent/pull/2277, I've created a package to "sniff" the stdout/stderr stream that a command write to the job logs for a certain string. The package is called `olefator`.

I also created a method to run a command in a shell, `RunWithOlfactor` that uses the olfactor to sniff out a string in the output. It will return a `OlfactorError` if the command exited with a non-zero exit status and the output (the combined stdout/stderr stream) contained the "smell".


go.mod
```diff
@@ -44,6 +44,7 @@ require (
 	google.golang.org/api v0.136.0
 	gopkg.in/DataDog/dd-trace-go.v1 v1.53.0
 	gopkg.in/yaml.v3 v3.0.1
+	gotest.tools/v3 v3.5.0
```
triarius, line 47: This simplified writing tests a lot. It is like testify, but uses go-cmp under the hood. It's also made by the same [org](https://github.com/gotestyourself/) that make `gotestsum`, the tool we use to run go tests in CI.

docs/git-mirror.md
triarius, line 1: Nothing significant changed in this file.

internal/job/shell/shell.go
```diff
@@ -269,6 +274,48 @@ func (s *Shell) RunWithEnv(ctx context.Context, environ *env.Environment, comman
 	})
 }
 
+// RunWithOlfactor runs a command, writes stdout and stderr to the shell's writer,
```
triarius, line 277: This based off `RunWithoutPrompt`, so look at that for the correctness of this.

internal/replacer/replacer_test.go
```diff
@@ -1,4 +1,4 @@
-package replacer
+package replacer_test
```
triarius, line 1: Without this, there are cyclic dependencies in the shell package.
`shell` -> `olfactor` -> `replacer` -> `redact` -> `shell`

But the `replacer` only needs the redactor for tests, so making a separate `_test` package fixed this.

internal/job/shell/shell_test.go
triarius, line 1: I added more Parallel and did a bit of clean up of these tests.

internal/olfactor/olfactor_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+import (
+	"io"
+	"testing"
+
+	"github.com/buildkite/agent/v3/internal/olfactor"
+	"gotest.tools/v3/assert"
+)
+
+func TestOlfactor(t *testing.T) {
```
triarius, line 11: Most of `olfactor`s string decoding should be tested in the `replacer` package, so these tests are fairly simple.

internal/job/shell/olfactory_error.go
```diff
@@ -0,0 +1,42 @@
...
+
+// Unwrap returns the wrapped error
+func (e *OlfactoryError) Unwrap() error {
+	return e.inner
+}
+
+// Is returns true if the target is an OlfactoryError and the inner errors are equal
+func (e *OlfactoryError) Is(target error) bool {
+	terr, ok := target.(*OlfactoryError)
+	// the detected slices were sorted on the way in, so we can compare them directly
```
DrJosh9000, line 40: Slices? (strings?) Sorted?

(I haven't yet gone looking through the commit history to understand)

internal/job/shell/shell.go
```diff
@@ -269,6 +274,48 @@ func (s *Shell) RunWithEnv(ctx context.Context, environ *env.Environment, comman
...
+// and returns an error if it fails. If the process exits with a non-zero exit code,
+// and `smell` was written to the logger (i.e. the combined stream of stdout and stderr),
+// the error will be of type `olfactor.OlfactoryError`. If the process exits 0, the error
+// will be nil.
```
DrJosh9000, line 281:
```suggestion
// will be nil whether or not the output contained `smell`.
```

internal/job/shell/shell_test.go
```diff
@@ -297,17 +316,12 @@ func TestLockFileRetriesAndTimesOut(t *testing.T) {
...
-	defer cmd.Process.Kill()
+	assert.NilError(t, err)
+	defer func() { assert.NilError(t, cmd.Process.Kill()) }()
 
-	timeout := time.Second * 2
-	if _, err := sh.LockFile(context.Background(), lockPath, timeout); err != context.DeadlineExceeded {
-		t.Errorf("sh.LockFile(%q, %v) error = %v, want context.DeadlineExceeded", lockPath, timeout, err)
-	}
+	_, err = sh.LockFile(context.Background(), lockPath, 2*time.Second)
+	assert.ErrorIs(t, err, context.DeadlineExceeded)
```
DrJosh9000, line 324: I'm not really convinced this is any clearer or cleaner than it was before. Checking the output of `cmd.Process.Kill` is good though.

internal/job/shell/shell_test.go
```diff
@@ -335,24 +349,27 @@ func acquireLockInOtherProcess(lockfile string) (*exec.Cmd, error) {
 
 // TestAcquiringLockHelperProcess isn't a real test. It's used as a helper process
```
DrJosh9000, line 350: Is this comment still correct?

internal/olfactor/olfactor.go
```diff
@@ -0,0 +1,31 @@
+package olfactor
+
+import (
+	"io"
+
+	"github.com/buildkite/agent/v3/internal/replacer"
+)
+
+// Olfactor may be used for 'sniffing' the output of a command.
```
DrJosh9000, line 9: Maybe explain further ('sniffing' = searching for a string)

internal/job/shell/shell.go
```diff
@@ -269,6 +274,48 @@ func (s *Shell) RunWithEnv(ctx context.Context, environ *env.Environment, comman
...
+	}
+
+	cmd, err := s.buildCommand(command, arg...)
+	if err != nil {
+		s.Errorf("Error building command: %v", err)
+		return err
+	}
+
+	w, o := olfactor.New(io.Discard, smell)
+	mw := io.MultiWriter(s.Writer, w)
```
DrJosh9000, line 303: [starting debates where none are needed] Although there is no need for a dependency between `s.Writer` and `w`, could the `MultiWriter` be eliminated by passing `s.Writer` into `olfactor.New`?

DrJosh9000 (commented):
Pretty good stuff, I have a few minor quibbles

internal/job/shell/olfactory_error.go
triarius: Oops, my first draft took a slice of smells, but I realised it was YAGNI

internal/job/executor.go
```diff
@@ -1575,19 +1575,36 @@ func (e *Executor) defaultCheckoutPhase(ctx context.Context) error {
...
-			// By default `git fetch origin` will only fetch tags which are
-			// reachable from a fetches branch. git 1.9.0+ changed `--tags` to
-			// fetch all tags in addition to the default refspec, but pre 1.9.0 it
-			// excludes the default refspec.
-			gitFetchRefspec, _ := e.shell.RunAndCapture(ctx, "git", "config", "remote.origin.fetch")
-			if err := gitFetch(ctx, e.shell, gitFetchFlags, "origin", gitFetchRefspec, "+refs/tags/*:refs/tags/*"); err != nil {
+		// Otherwise fetch and checkout the commit directly.
+		if err := gitFetch(ctx, e.shell, gitFetchFlags, "origin", e.Commit); err == nil {
+			break // it worked, break out of the switch statement
+		} else if gerr := new(gitError); errors.As(err, &gerr) {
```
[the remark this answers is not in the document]
triarius: But the `err` is defined in the `if` statement's scope, so we need to `As` it in the same scope.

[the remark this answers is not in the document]
DrJosh9000: Ah yeah. I feel that's an argument for moving the `err := gitFetch` out of the `if`.

internal/job/shell/shell.go
triarius: Whoops, this is much clearer. The dependency is a good thing, I think. It means they are part of a path in a pipeline.

internal/job/shell/shell_test.go
```diff
@@ -361,7 +357,107 @@ func newShellForTest(t *testing.T) *shell.Shell {
...
+			); err == nil {
+				// if there is no error, we expect an 0 exit code
+				assert.Check(t, test.exitCode == 0)
+			} else if derr := new(shell.OlfactoryError); !errors.As(err, &derr) {
+				assert.Check(t, !test.expectSmell, "no smell detected, but %s was expected", test.smell)
+				assert.ErrorContains(t, err, "exit status 1")
+			} else {
+				// if there was a OlfactoryError, we expect the smell to match the test.smell
+				// and the exit code to match the test.exitCode
+				assert.ErrorContains(t, err, "error running command: exit status 1, detected: hi")
```
DrJosh9000, line 451: I think the main source of complication here is the need to test the combination of success and error against expected success/failure and expected error. `assert` makes it shorter but it still needs three branches.

Would it simplify if the test cases were split into two or more tables?

DrJosh9000 (approved):
I think this is pretty great. It fixes the bug, and improves a bunch of existing code.

I think it's more important to ship it than quibble further, so take this: 🥒

internal/job/shell/shell_test.go
triarius: I think technically it was never true. When the tests are run normally, it's now run in parallel and passes. Previously it was run in serial and the test passes.

I moved the functionality in to a branch from a TestMain function, which seems like the more idiomatic way to do this.

internal/job/shell/shell_test.go
triarius: Yeah, I might do that in a follow up

PR #2288, opened 2023-08-14
Use capacity hint in `concat`

DrJosh9000:
I assume this is what @moskyb intended to do here... but didn't.

([I, too, like to make values and never use them.](https://github.com/buildkite/agent/pull/2257))

### DIFF
--- a/clicommand/pipeline_upload.go
+++ b/clicommand/pipeline_upload.go
@@ -3,6 +3,7 @@ package clicommand
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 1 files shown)


clicommand/pipeline_upload.go
```diff
@@ -511,7 +512,7 @@ func concat[T any](a ...[]T) []T {
 		cap += len(s)
 	}
 
-	var result []T
+	result := make([]T, 0, cap)
```
triarius, line 515: So the `+=` prevented the compiler from deteecting this. Good to know!

triarius (approved):

PR #2292, opened 2023-08-15
Include plugins in command step signatures

DrJosh9000:
Plugins alter the execution of a step completely, so the list of plugins and their configurations must be included in the signature for it to be secure.

This PR currently encodes plugin configs into JSON. This is potentially not a great encoding (what stability guarantees do we really have?), but because this is how the whole pipeline is uploaded, it ought to round-trip well enough.

### DIFF
--- a/internal/pipeline/step_command.go
+++ b/internal/pipeline/step_command.go
@@ -1,7 +1,7 @@
 package pipeline
 
 import (
-	"errors"
+	"encoding/json"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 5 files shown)


internal/pipeline/step_command.go
```diff
@@ -89,25 +89,62 @@ func (c *CommandStep) unmarshalMap(m *ordered.MapSA) error {
 }
 
 // SignedFields returns the default fields for signing.
-func (c *CommandStep) SignedFields() map[string]string {
+func (c *CommandStep) SignedFields() (map[string]string, error) {
+	plugins := ""
+	if len(c.Plugins) > 0 {
+		// TODO: Reconsider using JSON here - is it stable enough?
```
triarius, line 95: Ordered Maps are an issue, yeah? I think we can overcome it if we marshal all the unordered stuff into ordered equivalents.

Another one could be floating point representation, and precision. If the backend or database changes the precision of a number, we could have an issue.

triarius (approved):
LGTM. Let's see what if anything breaks with the JSON representation of plugins.

internal/pipeline/step_command.go
DrJosh9000: > Ordered Maps are an issue, yeah? I think we can overcome it if we marshal all the unordered stuff into ordered equivalents.

My thinking on this is: the pipeline parser ultimately shoves whatever `ordered.DecodeYAML` produces for the plugin config into `Plugin.Config`, which will tend to be either `*ordered.Map[string, any]` or `nil`, preserving input order. As long as the backend does it the same, it should be equal.

> Another one could be floating point representation, and precision. If the backend or database changes the precision of a number, we could have an issue.

This kind of thing _does_ concern me.

PR #2294, opened 2023-08-15
Pin bk cli used in agent pipeline to a commit

triarius:
The main branch is going to undergo some changes. We should pin to a commit while this happens so that the agent CI pipeline does not break.


DrJosh9000 (approved):
🧩

PR #2295, opened 2023-08-15
Include pipeline env in step signatures

DrJosh9000:
Environment variables can completely disrupt how a step runs. So they need to be signed too.

The env is external to a step in the pipeline, but is highly relevant once a job is about to run. Therefore, all pipeline env vars are included as a set of fields in the signature, and are treated as required during verification.

Other env vars are not included, because many of them vary according to all sorts of factors. If you need an environment variable to be a particular value, put it in the pipeline env. If it needs to vary at pipeline upload time, use interpolation.


DrJosh9000:
This considers pipeline env but not step env. Please hold.

triarius (approved):
LGTM 🎉

PR #2298, opened 2023-08-17
Enable mount-buildkite-agent in release containers

DrJosh9000:
The default changed in docker#v5.0.0 from true (on Linux) to false.
Enabling this propagates the agent access token as well as bind-mounts
the host agent into the container.

(So the agent container version didn't really matter.)


triarius (approved):
Nice find 🎉

PR #2300, opened 2023-08-17
Update VERSION and CHANGELOG for v3.51.1

triarius:


DrJosh9000 (approved):
I'm ok with this being a point release, but I think env signing also landed after v3.51.0

PR #2301, opened 2023-08-20
Add experimental fix for agent workers can read plugin directories while they are being written to by other agent workers

triarius:
Previously, we took out an exclusive, filesystem lock while the plugin directory was being checked out or deleted. However, this is not sufficient, as one agent worker could be reading from the plugin directory while another is checking it out. This is likely to happen when jobs that use the same plugin at different versions are running on the same agent process, in separate workers. Or, when the `BUILDKITE_PLUGINS_ALWAYS_CLONE_FRESH` variable is set. This race condition can appear as plugins running hooks that are from different versions of the plugin, or not at all, and has been reported by customers.

To fix it, I have namespaced the plugin directory with the agent name. Then, the agent workers will have their own copy of the plugin. At the moment, most plugins are a collection of bash scripts, so checking them out is relatively lightweight, so it should not matter that each agent worker receives a separate copy of the plugin.

This new behaviour is behind an experiment `isolated-plugin-checkout`.

While I was in the area, I cleaned up some of the code in the `checkoutPlugin` method too.

### DIFF
--- a/internal/job/executor.go
+++ b/internal/job/executor.go
@@ -768,13 +768,12 @@ func (e *Executor) PluginPhase(ctx context.Context) error {
 		return nil
 	}
 
-	// NOTE: these are method expressions: https://go.dev/ref/spec#MethodExpr
-	checkoutPluginFunc := (*Executor).checkoutPlugin
+	checkoutPluginMethod := e.checkoutPlugin
 	if experiments.IsEnabled(experiments.IsolatedPluginCheckout) {
 		if e.Debug {
 			e.shell.Commentf("Using isolated plugin checkout")
 		}
-		checkoutPluginFunc = (*Executor).checkoutPluginIsolated
+		checkoutPluginMethod = e.checkoutPluginIsolated
 	}
 
 	checkouts := []*pluginCheckout{}
@@ -788,7 +787,7 @@ func (e *Executor) PluginPhase(ctx context.Context) error {
 			continue
 		}
 
-		checkout, err := checkoutPluginFunc(e, ctx, p)
+		checkout, err := checkoutPluginMethod(ctx, p)
 		if err != nil {
 			return fmt.Errorf("Failed to checkout plugin %s: %w", p.Name(), err)
 		}
--- a/internal/job/executor.go
+++ b/internal/job/executor.go
@@ -947,6 +947,8 @@ func (e *Executor) hasPluginHook(name string) bool {
 
 // Checkout a given plugin to the plugins directory and return that directory. Each agent worker
 // will checkout the plugin to a different directory, so that they don't conflict with each other.
+// Because the plugin directory is unique to the agent worker, we don't lock it. However, if
+// multiple agent workers have access to the plugin directory, they need to have different names.
 func (e *Executor) checkoutPluginIsolated(ctx context.Context, p *plugin.Plugin) (*pluginCheckout, error) {
 	// Make sure we have a plugin path before trying to do anything
 	if e.PluginsPath == "" {
--- a/internal/job/executor.go
+++ b/internal/job/executor.go
@@ -768,6 +768,15 @@ func (e *Executor) PluginPhase(ctx context.Context) error {
 		return nil
 	}
 
+	// NOTE: these are method expressions: https://go.dev/ref/spec#MethodExpr
+	checkoutPluginFunc := (*Executor).checkoutPlugin
+	if experiments.IsEnabled(experiments.IsolatedPluginCheckout) {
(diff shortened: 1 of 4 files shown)


internal/job/executor.go
```diff
@@ -952,28 +952,19 @@ func (e *Executor) checkoutPlugin(ctx context.Context, p *plugin.Plugin) (*plugi
 
 	// Ensure the plugin directory exists, otherwise we can't create the lock
 	// Actual file permissions will be reduced by umask, and won't be 0777 unless the user has manually changed the umask to 000
-	if err := os.MkdirAll(e.PluginsPath, 0777); err != nil {
+	if err := os.MkdirAll(e.PluginsPath, 0o777); err != nil {
```
DrJosh9000, line 955: Thank you 🙌

internal/job/executor.go
```diff
@@ -944,36 +944,28 @@ func (e *Executor) checkoutPlugin(ctx context.Context, p *plugin.Plugin) (*plugi
 		return nil, fmt.Errorf("Can't checkout plugin without a `plugins-path`")
 	}
 
-	// Get the identifer for the plugin
```
triarius, line 947: 2023-08-21T22:29:01,070784658+10:00

internal/job/executor.go
```diff
@@ -1033,7 +1029,11 @@ func (e *Executor) checkoutPlugin(ctx context.Context, p *plugin.Plugin) (*plugi
 		return nil, err
 	}
 	// Switch back to the previous working directory
-	defer e.shell.Chdir(previousWd)
+	defer func() {
+		if err == nil {
```
triarius, line 1033: Don't clobber an existing error.

internal/job/executor.go
```diff
@@ -1033,7 +1026,11 @@ func (e *Executor) checkoutPlugin(ctx context.Context, p *plugin.Plugin) (*plugi
 		return nil, err
 	}
 	// Switch back to the previous working directory
-	defer e.shell.Chdir(previousWd)
+	defer func() {
+		if err := e.shell.Chdir(previousWd); err != nil && e.Debug {
+			e.shell.Errorf("failed to switch back to previous working directory: %v", err)
+		}
```
triarius, line 1032: We probably should return the error here. If the working directory is wrong, later plugin checkouts might work. But if this is the last plugin, then we could end up in a difficult to debug state when the job runs.

But, since we have not returned the error for so long, I'm just going to log it.

triarius:
I'm going to make this new behaviour part of an experiment

internal/job/executor.go
```diff
@@ -768,6 +768,15 @@ func (e *Executor) PluginPhase(ctx context.Context) error {
...
 	}
 
+	// NOTE: these are method expressions: https://go.dev/ref/spec#MethodExpr
+	checkoutPluginFunc := (*Executor).checkoutPlugin
+	if experiments.IsEnabled(experiments.IsolatedPluginCheckout) {
+		if e.Debug {
+			e.shell.Commentf("Using isolated plugin checkout")
+		}
+		checkoutPluginFunc = (*Executor).checkoutPluginIsolated
+	}
```
DrJosh9000, line 778: Definitely my favourite secret feature of Go. 

Since `checkoutPluginFunc` is always called with `e` as the first arg, maybe it would make more sense to capture `e.checkoutPlugin` instead?

```suggestion
	// NOTE: these are method expressions: https://go.dev/ref/spec#MethodExpr
	checkoutPluginFunc := e.checkoutPlugin
	if experiments.IsEnabled(experiments.IsolatedPluginCheckout) {
		if e.Debug {
			e.shell.Commentf("Using isolated plugin checkout")
		}
		checkoutPluginFunc = e.checkoutPluginIsolated
	}
```

and then:

```go
checkout, err := checkoutPluginFunc(ctx, p)
```

triarius: Aww, then I guess they're no longer method expressions 😥

I've done this.

internal/job/executor.go
```diff
@@ -937,22 +945,153 @@ func (e *Executor) hasPluginHook(name string) bool {
 	return false
 }
 
+// Checkout a given plugin to the plugins directory and return that directory. Each agent worker
+// will checkout the plugin to a different directory, so that they don't conflict with each other.
```
DrJosh9000, line 949: "without needing to use locks"?

DrJosh9000 (approved):
Good job!

I think I'm fine with it being a new method that is basically a copy of the old one, because the other ways to do it add complexity.

PR #2302, opened 2023-08-20
Allow interpolation of env into group step name

triarius:
Before this commit, group steps like this:
```yaml
env:
  FOO: bar
steps:
- name: "group ${FOO}"
  steps:
  - commands: echo hello 
```
Would not interpolate the env variable into the group name. This was
impossible to detect server side as the interpolation happens in the
agent.

I've chosen to make a `Group` interface with only one implementation. It might not be strictly necessary - as an alternative, I could have just refined `Group` from `any` to `string` in the `(*GroupStep).interpolate` method. However, I think it is cleaner to not have to do type assertion in the `interpolate` methods. Now, the type of the `Group` field in a `GroupStep` is determined when the YAML node is unmarshalled. If it is not a string (or null), there will be an error. This will cause the step to be unmarshalled as an `UnknownStep` instead.

### DIFF
--- a/internal/ordered/map.go
+++ b/internal/ordered/map.go
@@ -220,6 +220,15 @@ var EqualSS = Equal[string, string]
 // EqualSA is a convenience alias to reduce keyboard wear.
 var EqualSA = Equal[string, any]
 
+func (m *Map[K, V]) Equal(target any) bool {
+	switch targetT := (target).(type) {
+	case *Map[K, V]:
+		return Equal(m, targetT)
+	default:
+		return false
+	}
+}
+
 // compact re-organises the internal storage of the Map.
 func (m *Map[K, V]) compact() {
 	pairs := make([]Tuple[K, V], 0, len(m.index))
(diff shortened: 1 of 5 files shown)


DrJosh9000 (commented):
If this is just to fix the missing interpolation of the group key, this seems quite complicated.

`interpolateAny` should be able to handle nil or string, e.g in `GroupStep.interpolate`:

```go
	grp, err := interpolateAny(env, g.Group)
	if err != nil {
		return err
	}
	g.Group = grp
```

DrJosh9000:
For making `GroupStep.Group` a stronger type, I wonder if we should have a "nullable / optional string" type in the `ordered` package that could be shared?

triarius:
> For making `GroupStep.Group` a stronger type, I wonder if we should have a "nullable / optional string" type in the `ordered` package that could be shared?

I think that would be useful. But in practice, I think we might then have to handle 2 kinds of `nil`.

If there were a type that could be, for example, a `nil`, a string or a map, it might be useful to avoid deciding which of the types a `null` would be marshalled into. But I think we could just handle that with a `*struct{}`.

triarius:
Closed in favour of https://github.com/buildkite/agent/pull/2303

PR #2303, opened 2023-08-21
Fix missing group interpolation

DrJosh9000:
This fixes the missing interpolation, ~~but does not make `GroupStep.Group` more type-safe~~ changes the type to `*string`, and adds support for interpolating `*string`s.

### DIFF
--- a/internal/pipeline/parser_test.go
+++ b/internal/pipeline/parser_test.go
@@ -14,7 +14,7 @@ import (
 func ptr[T any](x T) *T { return &x }
 
 func TestParserParsesYAML(t *testing.T) {
-	envMap := env.FromSlice([]string{`ENV_VAR_FRIEND=friend`})
+	envMap := env.FromSlice([]string{"ENV_VAR_FRIEND=friend"})
 	input := strings.NewReader("steps:\n  - command: \"hello ${ENV_VAR_FRIEND}\"")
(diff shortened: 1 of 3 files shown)


internal/pipeline/parser_test.go
```diff
@@ -11,6 +11,8 @@ import (
 	"github.com/google/go-cmp/cmp"
 )
 
+func ptr[T any](x T) *T { return &x }
+
 func TestParserParsesYAML(t *testing.T) {
 	envMap := env.FromSlice([]string{`ENV_VAR_FRIEND="friend"`})
```
triarius, line 17:
```suggestion
	envMap := env.FromSlice([]string{`ENV_VAR_FRIEND=friend`})
```
I think you should change the test data to be like this. `os.Environ()` won't put quotes if there was an environment variable like this, [even if the value contains spaces](https://go.dev/play/p/p8seo0PUTk6), and this map is meant to mimic its output. You'll have to change the expected results too.

triarius (approved):
Nice! Closes #2302

internal/pipeline/parser_test.go
DrJosh9000: Done - and where that removes `"`, I've updated <code>\`</code> strings to `"`.

PR #2309, opened 2023-08-21
Auto-create tracing spans for log groups

goodspark:
This will make it easy for CI users to trace and analyze their CI jobs without requiring additional work or scripting on their part. They can just look at tracing information on their tracing provider.

Example with Datadog tracing:
image

Associated CI job:
Screenshot 2023-08-21 173042

### DIFF
--- a/internal/job/shell/shell.go
+++ b/internal/job/shell/shell.go
@@ -72,6 +72,9 @@ type Shell struct {
 
 	// Amount of time to wait between sending the InterruptSignal and SIGKILL
 	SignalGracePeriod time.Duration
+
+	// Whether to auto create tracing spans for log groups
+	TraceLogGroups bool
 }
 
 // New returns a new Shell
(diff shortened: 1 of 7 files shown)


internal/job/shell/shell.go
```diff
@@ -532,6 +535,29 @@ func round(d time.Duration) time.Duration {
...
+	ctx  context.Context
+	span opentracing.Span
+}
+
+func (s *spanMakerWriter) Write(p []byte) (n int, err error) {
+	if bytes.HasPrefix(p, []byte("~~~ ")) || bytes.HasPrefix(p, []byte("--- ")) || bytes.HasPrefix(p, []byte("+++ ")) {
+		s.Close()
+		operationName := bytes.SplitN(p[4:], []byte("\n"), 1)[0]
+		s.span, s.ctx = opentracing.StartSpanFromContext(s.ctx, string(operationName))
+	}
```
goodspark, line 550: Something I'm not clear about - can we assume Write calls are for each line? Or will we get subsets of a line and/or multiple lines inside the given byte array?

goodspark: In testing, it seems at least there will be multiple lines. I'm not sure if it will be called mid-line. It'll be simpler to assume not, since we won't have to maintain a buffer of writes.

internal/job/shell/shell.go
```diff
@@ -532,6 +535,32 @@ func round(d time.Duration) time.Duration {
...
+	if bytes.HasPrefix(p, []byte("~~~ ")) || bytes.HasPrefix(p, []byte("--- ")) || bytes.HasPrefix(p, []byte("+++ ")) {
+		s.FinishIfActive()
+		operationName, _, _ := strings.Cut(string(p[4:]), "\r\n")
+		// We don't store the context bc we don't need to create child spans (yet). If we stored it, every log group would
+		// look like a child of the previous log group, where they're all more like siblings under the same parent span,
+		// since Buildkite itself has no concept of log group hierarchy.
+		s.span, _ = opentracing.StartSpanFromContext(s.ctx, operationName)
+	}
+	return s.w.Write(p)
+}
```
triarius, line 555: So the complete line is not guaranteed to be contained in a single `Write`. The `+++`, etc could occur in the middle of `p`. Even worse, it could be split across consecutive calls to `Write`.

I think what we need here is a `Writer` that has a state machine that will capture what's input between (for example) `+++` and then next `\n` and creates a new span as appropriate. It will also have to have an upper bound on how much it captures to prevent an adversary from crafting an input that will cause the agent to run out of memory. Of course, a limit is also useful because the tracing backend will have a limit on how long a span's name can be.

triarius (commented):
This is a great idea @goodspark, be we foresee some issues with the implementation.

We also would really appreciate in implementation for opentelemetry. It should not be that much more difficult.

DrJosh9000:
I feel bad about closing this, but it's likely going to need a full revisit in v4. We also have some plans for semantic sectioning of logs and OTel logging.

PR #2310, opened 2023-08-22
Publish RPMs: sync less data

DrJosh9000:
The core idea is to use `--recycle-pkglist` to skip recalculating metadata for all the old packages, which avoids the need to have the old packages there to begin with.


triarius (approved):
Nice, I think this should be pretty safe to merge.

You can merge this and create a branch to test it or keep adding commits to this branch and ask me to review later.

DrJosh9000:
I'm going to merge it and see what happens

PR #2311, opened 2023-08-22
Fix a flaky shell test

triarius:
The shell test ran the test binary in a separate process. The test binary is supposed to sleep forever, but the go runtime terminates it because all its goroutines are idle. This is because the test binary is not doing anything in the main goroutine. This PR fixes the issue by making the test binary do something in the main goroutine.

I've also kept the changes I made to make the test more debuggable. Here is the output from when I debugged. Note the output `fatal error: all goroutines are asleep - deadlock!`.
```shell
DONE 1 tests in 3.267s
gotestsum -- -count=1 -v ./internal/job/shell -run ^TestLockFileRetriesAndTimesOut$
✖  internal/job/shell (14ms)

=== Failed
=== FAIL: internal/job/shell TestLockFileRetriesAndTimesOut (0.01s)
    shell_test.go:319: acquiring lock in other process: /tmp/shelltest3604204178/my.lock
2023/08/22 13:20:35 Locking /tmp/shelltest3604204178/my.lock
2023/08/22 13:20:35 Acquired lock /tmp/shelltest3604204178/my.lock
fatal error: all goroutines are asleep - deadlock!

goroutine 1 [select (no cases)]:
github.com/buildkite/agent/v3/internal/job/shell_test.acquiringLockHelperProcess()
	/home/narthana/devel/buildkite/agent/internal/job/shell/main_test.go:49 +0x2d1
github.com/buildkite/agent/v3/internal/job/shell_test.TestMain(0xc0000666f0?)
	/home/narthana/devel/buildkite/agent/internal/job/shell/main_test.go:23 +0x4f
main.main()
	_testmain.go:77 +0x1c6
    shell_test.go:324: assertion failed: error is nil, not "context deadline exceeded" (context.DeadlineExceeded context.deadlineExceededError)

DONE 1 tests, 1 failure in 0.229s
```

### DIFF
--- a/internal/job/shell/main_test.go
+++ b/internal/job/shell/main_test.go
@@ -45,6 +45,12 @@ func acquiringLockHelperProcess() error {
 	}
 
 	log.Printf("Acquired lock %s", fileName)
-	// sleep forever
-	select {}
+	// sleep forever, but keep the main goroutine busy
+	c := make(chan struct{}, 1)
+	for {
+		select {
+		case <-c:
+		case c <- struct{}{}:
+		}
+	}
 }
--- a/internal/job/shell/main_test.go
+++ b/internal/job/shell/main_test.go
@@ -46,11 +46,7 @@ func acquiringLockHelperProcess() error {
 
 	log.Printf("Acquired lock %s", fileName)
 	// sleep forever, but keep the main goroutine busy
-	c := make(chan struct{}, 1)
 	for {
-		select {
-		case <-c:
-		case c <- struct{}{}:
-		}
+		time.Sleep(1 * time.Second)
 	}
 }
--- a/internal/job/shell/shell_test.go
+++ b/internal/job/shell/shell_test.go
@@ -316,15 +316,14 @@ func TestLockFileRetriesAndTimesOut(t *testing.T) {
 
 	lockPath := filepath.Join(dir, "my.lock")
 
-	cmd, err := acquireLockInOtherProcess(t, lockPath)
-	assert.NilError(t, err)
+	cmd := acquireLockInOtherProcess(t, lockPath)
 	defer func() { assert.NilError(t, cmd.Process.Kill()) }()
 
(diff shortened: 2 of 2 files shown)


internal/job/shell/main_test.go
```diff
@@ -45,6 +45,12 @@ func acquiringLockHelperProcess() error {
...
-	// sleep forever
-	select {}
+	// sleep forever, but keep the main goroutine busy
+	c := make(chan struct{}, 1)
+	for {
+		select {
+		case <-c:
+		case c <- struct{}{}:
+		}
+	}
```
DrJosh9000, line 55: Does the main goroutine have to be _that_ busy?

```suggestion
	for {
		time.Sleep(1 * time.Second)
	}
```

triarius: 😅

internal/job/shell/shell_test.go
```diff
@@ -316,24 +316,28 @@ func TestLockFileRetriesAndTimesOut(t *testing.T) {
...
 	cmd := exec.Command(os.Args[0], "--", lockfile)
 	cmd.Env = []string{"TEST_MAIN_WANT_HELPER_PROCESS=1"}
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
 
 	err := cmd.Start()
-	if err != nil {
-		return cmd, err
-	}
+	assert.NilError(t, err)
```
DrJosh9000, line 340: So this fails the test immediately if err != nil? (Looks like it... had to look it up since I'm not yet familiar with this particular assert lib). Then there would be no need to have an error return on the function.

DrJosh9000 (approved):

internal/job/shell/shell_test.go
triarius: Yeah, good point. I've made it not return the error. If the next assertion passes, then it's a false positive. So I think it's fine to exit early.

PR #2313, opened 2023-08-22
Update help text with suggestions from docs code review

triarius:
You may be wondering about the removal of line breaks from the code examples. This was done to reflected changes in the docs. It seems to still look good for the cli:
```shell
Usage:

   buildkite-agent lock release [key] [token]

Description:
   Releases the lock for the given key. This should only be called by the
   process that acquired the lock. To help prevent different processes unlocking
   each other unintentionally, the output from ′lock acquire′ is required as the
   second argument.

   Note that this subcommand is only available when an agent has been started
   with the ′agent-api′ experiment enabled.

Examples:

   #!/bin/bash
   token=$(buildkite-agent lock acquire llama)
   # your critical section here...
   buildkite-agent lock release llama "${token}"

Options:

   --no-color            Don't show colors in logging [$BUILDKITE_AGENT_NO_COLOR]
   --debug               Enable debug mode. Synonym for ′--log-level debug′. Takes precedence over ′--log-level′ [$BUILDKITE_AGENT_DEBUG]
   --log-level value     Set the log level for the agent, making logging more or less verbose. Defaults to notice. Allowed values are: debug, info, error, warn, fatal (default: "notice") [$BUILDKITE_AGENT_LOG_LEVEL]
   --experiment value    Enable experimental features within the buildkite-agent [$BUILDKITE_AGENT_EXPERIMENT]
   --profile value       Enable a profiling mode, either cpu, memory, mutex or block [$BUILDKITE_AGENT_PROFILE]
   --config value        Path to a configuration file [$BUILDKITE_AGENT_CONFIG]
   --lock-scope value    The scope for locks used in this operation. Currently only 'machine' scope is supported (default: "machine") [$BUILDKITE_LOCK_SCOPE]
   --sockets-path value  Directory where the agent will place sockets (default: "/home/narthana/.buildkite-agent/sockets") [$BUILDKITE_SOCKETS_PATH]
```

### DIFF
--- a/clicommand/env_unset.go
+++ b/clicommand/env_unset.go
@@ -26,7 +26,7 @@ Description:
 
 Examples:
 
-   Unsetting the variables ′LLAMA′ and ′ALPACA′)
+   Unsetting the variables ′LLAMA′ and ′ALPACA′:
 
    $ buildkite-agent env unset LLAMA ALPACA
    Unset:
--- a/clicommand/lock_done.go
+++ b/clicommand/lock_done.go
@@ -11,7 +11,7 @@ import (
 
 const lockDoneHelpDescription = `Usage:
 
-   buildkite-agent lock release [key]
+   buildkite-agent lock done [key]
 
 Description:
    Completes a do-once lock. This should only be used by the process performing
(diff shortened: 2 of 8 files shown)


clicommand/lock_done.go
```diff
@@ -11,7 +11,7 @@ import (
 
 const lockDoneHelpDescription = `Usage:
 
-   buildkite-agent lock release [key]
+   buildkite-agent lock done [key]
```
DrJosh9000, line 14: 😅

clicommand/env_unset.go
```diff
@@ -26,7 +26,7 @@ Description:
 
 Examples:
 
-   Unsetting the variables ′LLAMA′ and ′ALPACA′)
+   Unsetting the variables ′LLAMA′ and ′ALPACA′:
```
DrJosh9000, line 29: 😅

DrJosh9000 (approved):
😅😅😅

PR #2315, opened 2023-08-22
Add --skip-stat

DrJosh9000:
This should cut down on the thousands of (hopefully meaningless?) attempts to stat each and every previous RPM?


triarius (approved):
LGTM

PR #2317, opened 2023-08-23
Adjust cli help output to work better with documentation generation

triarius:
I've adjusted the documentation generation with some new features. We need to meet it half way in the agent code base though.

The biggest is that text that is indented by 4 spaces is treated as
a code block. So we need to unindent non-code and indent code.

See https://github.com/buildkite/docs/pull/2406 for more details.

If you check out the code in that PR, compile this version of the agent and run the script `scripts/update-agent-help.sh`, it should be a no-op.


DrJosh9000 (approved):
Prime

PR #2318, opened 2023-08-23
Artifact up/download to/from Azure Blob Storage

DrJosh9000:
This adds Azure Blob Storage as an accepted destination for artifacts. 

To start with, it uses [DefaultAzureCredential](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#readme-defaultazurecredential). This is fairly flexible, but unless the container is public, uploaded URLs cannot be accessed by default.

In the third commit I add support for account keys (shared key credentials), which can also be configured using a connection string. The shared key is mandatory for generating SAS (shared access signatures) URLs, which include a token allowing temporary access to a blob.


internal/artifact/azure_blob.go
```diff
@@ -0,0 +1,80 @@
...
+)
+
+// The domain suffix for Azure Blob storage.
+const azureBlobHostSuffix = ".blob.core.windows.net"
+
+// NewAzureBlobClient creates a new Azure Blob client.
+func NewAzureBlobClient(l logger.Logger, storageAccountName string) (*service.Client, error) {
+	url := fmt.Sprintf("https://%s%s/", storageAccountName, azureBlobHostSuffix)
+
+	// TODO: other credentials?
```
triarius, line 21: Did you do these? If not, maybe provide some links to docs on what the remaining credentials are.

triarius (commented):
Looks good in general, I have a suggestion which I think you should implement. I guess not all the proposed `AzureBlobLocation` struct is needed in all the places it would have to be plumbed to, but I think it's still useful to create this structure.

internal/artifact/azure_blob_downloader.go
```diff
@@ -0,0 +1,69 @@
...
+
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob"
+	"github.com/buildkite/agent/v3/logger"
+)
+
+// AzureBlobUploaderConfig configures AzureBlobDownloader.
+type AzureBlobDownloaderConfig struct {
+	Path        string
+	Repository  string
+	Destination string
```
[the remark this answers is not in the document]
DrJosh9000: The way the artifact downloader is currently structured is slightly annoying.

clicommand/global.go
```diff
@@ -103,6 +103,7 @@ var RedactedVars = cli.StringSliceFlag{
 		"*_PRIVATE_KEY",
 		"*_ACCESS_KEY",
 		"*_SECRET_KEY",
+		"*_CONNECTION_STRING",
```
triarius, line 106: Makes sense. It might be worth mentioning that connection string are frequently of the form `https://username:password@host`.

triarius (approved):
Thanks, this looks great!

PR #2319, opened 2023-08-23
Fix signature mismatches when steps have plugins

DrJosh9000:
The first part of this PR enables `CommandStep` to be unmarshaled directly from JSON consistently. This is mainly needed because `Plugin` wasn't set up to be unmarshaled from JSON, but reusing the YAML -> ordered path keeps it consistent.

The second part of this PR changes how `Plugin` is marshaled back to JSON. `FullSource` applies similar normalisations that the backend should be applying to plugin sources, and comes with the ~same suite of tests. 

Source normalisation of plugins is better suited to the marshaling phase than the parsing/unmarshaling phase, because:

* interpolation applies to plugin sources and configs, so putting it in the parsing phase would require reworking interpolation
* if/when `pipeline` is spun out into its own library, users would be writing `Plugin` literals with all the accepted kinds of `Source`, so normalisation would generally happen later.

There's also a couple of log output fumbles (`%w` is only supported in `fmt.Errorf`, not logger) and a fix to make it easier to hold `ordered.Unmarshal` when you have `yaml.Node`.

### DIFF
--- a/internal/pipeline/plugin.go
+++ b/internal/pipeline/plugin.go
@@ -2,6 +2,9 @@ package pipeline
 
 import (
 	"encoding/json"
+	"net/url"
+	"path"
+	"strings"
 
 	"github.com/buildkite/interpolate"
 	"gopkg.in/yaml.v3"
@@ -19,12 +22,13 @@ var (
 //
 // Standard caveats apply - see the package comment.
 type Plugin struct {
-	Name   string
+	Source string
 	Config any
 }
 
 // MarshalJSON returns the plugin in "one-key object" form, or "single string"
-// form (no config, only plugin name).
+// form (no config, only plugin name). Plugin sources are marshalled into "full"
+// form.
 func (p *Plugin) MarshalJSON() ([]byte, error) {
 	// NB: MarshalYAML (as seen below) never returns an error.
 	o, _ := p.MarshalYAML()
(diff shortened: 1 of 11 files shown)


internal/pipeline/plugin.go
```diff
@@ -19,40 +22,92 @@ var (
...
+	}
+
+	// Looks like an absolute or relative file path.
+	if strings.HasPrefix(p.Source, "/") || strings.HasPrefix(p.Source, ".") || strings.HasPrefix(p.Source, `\`) {
+		return p.Source
+	}
+
+	u, err := url.Parse(p.Source)
+	if err != nil {
+		return p.Source
```
triarius, line 66: I wonder if there's a way that we can carry across that `FullSource` had to fall back to being identical on a particular input. Then, when it comes time to sign the plugin, we can feed a warning to the job logs. It's perhaps not necessary right now, but I would want such a thing before we announce this is released.

triarius (approved):
Nice, falling back to not altering the source is a good strategy.

internal/pipeline/plugin.go
DrJosh9000: I think it's a bit tricky. Some inputs have to fall back because neither the pipeline-uploading agent nor the backend know what local or vendored plugins will be present. And some inputs must fall back because there needs to be a way to override the assumption that the plugin is in GitHub (and usually in buildkite-plugins). Failing to parse as a URL / URI could be a mistake, but I have the suspicion that some edge case will fall into it.

PR #2321, opened 2023-08-25
Handle the case when unmarshalling a step where there aren't any plugins

moskyb:
Previously, when unmarshalling a step that had no plugins, it would fail to accept the job:
```
2023-08-25 10:27:47 WARN   Bens-Macbook-Pro-1 Buildkite rejected the call to accept the job (unmarshalling CommandStep: unmarshaling plugins: got <nil>, want []any or *ordered.Map[string, any])
```

### DIFF
--- a/internal/pipeline/plugins.go
+++ b/internal/pipeline/plugins.go
@@ -31,6 +31,10 @@ func (p *Plugins) UnmarshalOrdered(o any) error {
 	}
 
 	switch o := o.(type) {
+	case nil:
+		*p = nil
+		return nil
+
 	case []any:
 		for _, c := range o {
 			switch ct := c.(type) {


triarius (commented):
LGTM, but needs a test

DrJosh9000 (approved):

internal/pipeline/plugins.go
```diff
@@ -31,6 +31,10 @@ func (p *Plugins) UnmarshalOrdered(o any) error {
 	}
 
 	switch o := o.(type) {
+	case nil:
+		*p = Plugins{}
```
moskyb, line 35: i'm rethinking this a little - this makes nil unmarshal to an empty slices, but it should probably unmarshal to nil, right? all slices being reference types and all that? does it matter that much? they both work the same.

DrJosh9000: I would write `nil` for these reasons:

- laziness
- when I want to remind the reader that `nil` is a valid slice value
- in order to distinguish "missing" instead of merely "empty"

I would write `[]something{}` for these reasons:

- when I want to remind the reader that checking for empty slices by comparing to `nil` instead of `len(slice) == 0` is not good enough
- in order to distinguish "not missing but has no items"
- maybe the slice has to pass through an interface, in which case untyped `nil` has a different meaning

triarius: I think nil -> nil is consistent with the old behaviour before any pipeline parsing was done by the agent, yeah? So let's do that.

moskyb: > in order to distinguish "missing" instead of merely "empty"

i think this one is the key reason to use nil - in this case, plugins are missing and not empty, and the empty case should be handled by the `*ordered.MapSA` branch in the typeswitch.

triarius (approved):
LGTM

PR #2324, opened 2023-08-27
Split checkout, artifact, and plugin phases out of executor.go

triarius:
The file is too big for me to navigate easily. Splitting it into multiple files will make it easier to work with.

We should split it into multiple packages, but that will require more work.

I recommend reviewing commit-by-commit.


moskyb:
having not read it yet: shit yes

triarius:
@benno I was reading the diff and it seems a bit strange. I'll try to make it easier to review.

moskyb (approved):
send it. so happy about this

DrJosh9000 (commented):
🤩

PR #2325, opened 2023-08-28
Retry fork/exec errors when running hook

triarius:
A customer reported that their jobs were unable to run a hook because the hook wrapper script was locked.

The hooks should have collision avoidant names, and they should have closed by the agent worker goroutine that created them prior to execution, which is in the same goroutine. So it's a bit of a mystery how the script wrapper could remain open.

We could just add more retry to work around this, but it would like to find the cause. So, I've done two things in this PR:

- Retry executing the hook if we get a fork/exec error
- If the retry is exhausted, log the error and try to determine what process has the file open

That latter will only work if:
- the process that has the file open is long-lived enough for the error handling to catch it with the file open, and
- the agent has permission to read the open file descriptors of the other processes

So we can't be certain what the other process is, but at least we should be able to prove or disprove that it is the agent itself.

Detecting other processes only works on Linux because it uses the [procfs](https://docs.kernel.org/filesystems/proc.html). For other Unixes, we can use `/dev/fd` to check that the agent process does not have the file open. I'm not sure if this will work if the two goroutines are scheduled on different OS threads, however.

### DIFF
--- a/internal/job/executor.go
+++ b/internal/job/executor.go
@@ -364,17 +364,17 @@ func (e *Executor) runUnwrappedHook(ctx context.Context, hookName string, hookCf
 	return e.shell.RunWithEnv(ctx, environ, hookCfg.Path)
 }
 
-func logOpenedHookInfo(l shell.Logger, hookName, path string) {
+func logOpenedHookInfo(l shell.Logger, debug bool, hookName, path string) {
 	switch {
 	case runtime.GOOS == "linux":
-		procPath, err := file.OpenedBy(l, path)
+		procPath, err := file.OpenedBy(l, debug, path)
 		if err != nil {
 			l.Errorf("The %s hook failed to run because and we could not find the process that had it opened", hookName)
 		} else {
 			l.Errorf("The %s hook failed to run because it was open by %s", hookName, procPath)
 		}
 	case utils.FileExists("/dev/fd"):
-		isOpened, err := file.IsOpened(l, path)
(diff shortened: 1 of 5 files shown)


internal/job/executor.go
```diff
@@ -363,6 +364,31 @@ func (e *Executor) runUnwrappedHook(ctx context.Context, hookName string, hookCf
...
 }
 
+func logOpenedHookInfo(l shell.Logger, debug bool, hookName, path string) {
+	switch {
+	case runtime.GOOS == "linux":
+		procPath, err := file.OpenedBy(l, debug, path)
+		if err != nil {
+			l.Errorf("The %s hook failed to run because and we could not find the process that had it opened", hookName)
+		} else {
+			l.Errorf("The %s hook failed to run because it was open by %s", hookName, procPath)
```
moskyb, line 374:
```suggestion
			l.Errorf("The %s hook failed to run the %s process has the hook file open", hookName, procPath)
```

moskyb: maybe something like `the %s process (PID: %d)` might be good? but would probably require a fair bit of plumbing so it's def not a blocker

triarius:
@moskyb after chatting with @pda, we decided to feed back errors from failing to close files that are opened in `NewScriptWrapper`. If the files are kept open by the agent, that's the likely cause.

moskyb (approved):

PR #2332, opened 2023-08-29
Adapt Olfactor to allow sniffing for multiple smells

triarius:
The olfactor used to take a single smell and sniff for it in an io stream. Now, a single olfactor can sniff for multiple smells.
The necessitated a method, `(*Olfactor.Smelt(smell string))`, which allows the user to query the olfactor for the smells it smelt.

The first implementation of this, using a hash map to store the detected smells, had a complication where if one of the smells was the prefix of another, only the longer smell would be reported being "smelt" by the olfactor.
To remedy this, I used a trie to store the smelt smells, but this also means that `(*Olafactor.Smelt(smell string))` will return `true` for any prefix of a smell that the `Olfactor` was created with. I think that keeps with the purpose of the olfactor: if you're searching an io stream for a string, you would want the search to be positive if any super string were written to it.


internal/job/shell/olfactory_error.go
```diff
@@ -34,8 +31,10 @@ func (e *OlfactoryError) Unwrap() error {
 	return e.inner
 }
 
-// Is returns true if the target is an OlfactoryError and the inner errors are equal
+// Is returns true if the target is an OlfactoryError and the smells match. If
+// you want to check the inner errors match, you should unwrap them and run
+// errors.Is on the unwrapped errors.
 func (e *OlfactoryError) Is(target error) bool {
 	terr, ok := target.(*OlfactoryError)
-	return ok && e.Smell == terr.Smell && errors.Is(e.inner, terr.inner)
```
triarius, line 40: I was a bit overzealous with error unwrapping here. I think in most cases, we don't care what the inner errors are. If we do, there's a way to figure it out.

internal/job/shell/shell_test.go
```diff
@@ -437,25 +429,33 @@ func TestRunWithoutPromptWithOlfactor(t *testing.T) {
...
+			assert.Assert(t, err != nil)
 			assert.Equal(t, test.output, out.String())
+
+			for _, smell := range test.smellsInOutput {
+				assert.Check(t, acmp.ErrorIs(err, shell.NewOlfactoryError(smell, nil)))
+			}
 		})
 	}
+
+	t.Run("smells_nothing_when_exit_0", func(t *testing.T) {
```
triarius, line 447: I could not fit this with the type of the other errors

internal/job/shell/shell_test.go
```diff
@@ -360,72 +360,64 @@ func newShellForTest(t *testing.T) *shell.Shell {
...
-			smell:       "hi",
-			expectSmell: false,
-			command:     []string{"bash", "-ec", "echo bye, how were you; false"},
-			output:      "bye, how were you\n",
-			exitCode:    1,
+			name:           "multiple_smells",
+			smellsToSniff:  []string{"bye", "how"},
+			command:        []string{"bash", "-ec", "echo bye, how were you; false"},
+			output:         "bye, how were you\n",
+			smellsInOutput: []string{"bye", "how"},
```
triarius, line 420: This is the only new test case.

internal/job/git.go
```diff
@@ -50,15 +53,15 @@ func gitCheckout(ctx context.Context, sh shellRunner, gitCheckoutFlags, referenc
 		return err
 	}
 	if !gitCheckRefFormat(reference) {
-		return fmt.Errorf("%q is not a valid git ref format", reference)
+		return fmt.Errorf("%q %w", reference, errInvalidRef)
```
triarius, line 56: I smuggled this in, forgive me.

internal/job/shell/shell.go
```diff
@@ -299,14 +299,17 @@ func (s *Shell) RunWithOlfactor(
...
 		Stdout: true,
 		Stderr: true,
 		PTY:    s.PTY,
 	}); err != nil {
-		if o.Smelt() {
-			return NewOlfactoryError(smell, err)
+		// wrap err with the smells that were smelt
+		for _, smell := range smells {
+			if o.Smelt(smell) {
+				err = NewOlfactoryError(smell, err)
```
DrJosh9000, line 311: I think the only issue here is that a linked list of OlfactoryErrors feels like the wrong data structure.

internal/trie/trie.go
DrJosh9000, line 1: I love a trie!

But we could consider changing the underlying replacer to report groups of strings it finds in each overlap range, which would give you a clean set of smells.

internal/job/shell/shell.go
triarius: Good point. Shall I put the trie in the error?

internal/trie/trie.go
triarius: For what it was designed for, the replacer only calling the closure after a complete match seems correct. I don't really want to mess with that.

Either way, I think the matching of any prefix behaviour is still desirable for the olfactor, so some ideas from a trie could still be incorporated into the replacement for the replacer.

internal/job/shell/shell.go
DrJosh9000: Maybe? So it would be up to the caller to check for an `OlfactoryError`, then query the trie in it to find out what it smelled?

But if you do that, then I wonder... what if we used `RunWithOlfactor` to check a particular smell is _present_, rather than absent? Then it would be better if `RunWithOlfactor` returned `(*Olfactor, error)`, so the caller could query the `Olfactor` for smells directly, and the error would only be non-nil when it failed to run.

internal/job/git_test.go
```diff
@@ -256,12 +259,12 @@ func (r *mockShellRunner) Run(_ context.Context, cmd string, args ...string) err
...
 	_ context.Context,
-	_ string,
+	_ []string,
 	cmd string,
 	args ...string,
-) error {
+) (*olfactor.Olfactor, error) {
 	r.got = append(r.got, append([]string{cmd}, args...))
-	return nil
+	return nil, nil
```
triarius, line 267: Because this is just a mock, I think it's okay to return a `nil` Olfactor.

internal/job/shell/shell.go
triarius: >  Then it would be better if RunWithOlfactor returned `(*Olfactor, error)`

Yeah, after implementing this, I agree. I've done this and eliminated `OlfactoryErrors` entirely. It's more ergonomic for the caller to have more control without layers of `error` wrapping in the way.

internal/job/git_test.go
DrJosh9000: Feels like, if this became a problem, it would be quick to find.

DrJosh9000 (approved):
🚢❇️

PR #2334, opened 2023-08-29
Update skopeo to v1.13.2

DrJosh9000:
https://github.com/containers/podman/discussions/19796

Perhaps we should just use `stable:v1`?


triarius (approved):
LGTM
> Perhaps we should just use stable:v1?

Sound good too

PR #2337, opened 2023-08-30
Fix ECR authentication failure

DrJosh9000:
https://github.com/containers/skopeo/issues/1909 summary:

Skopeo 1.11 changed behavior - when an `--authfile` flag or `REGISTRY_AUTH_FILE` env var is present, it read only that file for creds. We weren't setting that flag/var ourselves but the skopeo container does: 

https://github.com/containers/skopeo/blob/7f2f46e1b97df36b2b82d1b1d87c81b8b3d21601/contrib/skopeoimage/stable/Containerfile#L44

### DIFF
--- a/.buildkite/steps/publish-docker-image.sh
+++ b/.buildkite/steps/publish-docker-image.sh
@@ -15,7 +15,11 @@ dry_run() {
 }
 
 skopeo() {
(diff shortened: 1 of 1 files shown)


triarius (approved):
💖 for finding this!

PR #2339, opened 2023-08-30
Fix verification for steps with plugins, part 2

DrJosh9000:
Dogfooding showed that signed steps with plugins still tended to fail verification. After digging around with `--debug-http`, it turns out that plugin config maps/hashes tend to be reordered by the backend.

On the plus side: we're free to reorder keys within plugin configs! 🎉

Changes here:

- `pipeline.Plugins` can now be unmarshaled directly from JSON. For consistency it reuses the underlying `UnmarshalOrdered` using the same technique as for `CommandStep`.
- `ordered` now has `ToMapRecursive`, which replaces `*Map[string, any]` with `map[string]any` recursively.
- Each `Plugin` is unmarshaled using `ToMapRecursive`.
- `api.Job.ValuesForFields` now passes `BUILDKITE_PLUGINS` through a JSON unmarshal-marshal round-trip. The JSON marshaler writes (plain) map keys in sorted order. 

Net effect - signatures should verify.

### DIFF
--- a/internal/pipeline/sign_test.go
+++ b/internal/pipeline/sign_test.go
@@ -7,7 +7,6 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/buildkite/agent/v3/internal/ordered"
 	"github.com/lestrrat-go/jwx/v2/jwa"
 	"github.com/lestrrat-go/jwx/v2/jwk"
 )
@@ -22,12 +21,7 @@ func TestSignVerify(t *testing.T) {
 			},
 			{
 				Source: "another-plugin#v3.4.5",
-				Config: ordered.MapFromItems(
-					ordered.TupleSA{
-						Key:   "llama",
-						Value: "Kuzco",
-					},
-				),
+				Config: map[string]any{"llama": "Kuzco"},
 			},
 		},
 		Env: map[string]string{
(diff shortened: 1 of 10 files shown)


internal/pipeline/sign_test.go
```diff
@@ -22,12 +21,7 @@ func TestSignVerify(t *testing.T) {
 			},
 			{
 				Source: "another-plugin#v3.4.5",
-				Config: ordered.MapFromItems(
-					ordered.TupleSA{
-						Key:   "llama",
-						Value: "Kuzco",
-					},
-				),
+				Config: map[string]any{"llama": "Kuzco"},
```
triarius, line 24: Given that this PR was made to address the order of keys in plugin config, we should have a test that would have failed if the order weren't stable.

It might be impossible to write such a test using the `json` marshaller in go, but it's still worth having. The fact that `json.Marshal` sorts the keys before writing them is pretty crucial, so we should assert in our tests that it does to prevent regression.

triarius (approved):
SO the problem was that `ordered.Map` *was* used all the way down.

Nice fix!

internal/pipeline/sign_test.go
DrJosh9000: I've added a test that I think should capture encoding stability problems. It failed, but the reason was how I was holding Sign and Verify, so I made their argument processing more consistent.

PR #2340, opened 2023-08-31
Catch step env/job env edge case

DrJosh9000:
#2339 opened a possible edge case where if a job includes a step with an env var, but that env wasn't copied to the job, then the job could verify successfully with a different env than was intended during signing.

Even without #2339, this seems like a good defensive check to add anyway.


triarius (approved):
NIce catch!

Issue #2305, opened 2023-08-21
Glob pattern dist/**/* doesn't discover top level (within dist/) files


VasiliiSikora:
**Is your feature request related to a problem? Please describe.**
When using `buildkite-agent artifact download "dist/**/*" .` to download all artifacts in a compiled `dist` folder that has successfully been uploaded using `artifact_paths: "dist/**/*"` in the `pipeline.yml` file, the files downloaded do not contain files such as `dist/express.js` or `dist/worker.js`. 

This seems [unexpected for the given glob pattern](https://www.digitalocean.com/community/tools/glob?comments=true&glob=dist%2F%2A%2A%2F%2A&matches=false&tests=%2F%2F%20This%20will%20match%20as%20it%20ends%20with%20%27.js%27&tests=dist%2Ffolder%2Ffile.js&tests=%2F%2F%20This%20won%27t%20match%21&tests=dist%2Ffile.js).

**NOTE**: The above pattern does download all files that are contained in subdirectories, for example, dist/apps/express.js and dist/apps/worker.js are successfully downloaded with the above command.

**Describe the solution you'd like**
A solution that allows me to download files described above with the glob pattern `dist/**/*`. Otherwise, documentation to explain why this pattern does not work for `buildkite-agent`

**Describe alternatives you've considered**
I have found a workaround by using `buildkite-agent artifact download "dist/**" .`, however, unsure if this was intended behaviour or not.

**Additional context**
N/A

moskyb:
hey there! this is somewhat of a running issue - there's a bunch of edge cases and gotchas with artifact uploading and downloading. we have some work planned to work on making this better - we'll keep this issue open until we do, but in the meantime, it looks like your workaround is sufficient, if not quite desirable.

PR #2341, opened 2023-08-31
Provide new glob library with experiment

DrJosh9000:
Why a new glob library? 

- We've had issues reported where `artifact upload` tries to walk into subdirectories it has no permissions to enter, despite the pattern not matching those subdirectories at all.
- It's also inconvenient to have the glob bomb out half way with ambiguous-sounding errors.
- Syntax problems, e.g. #2305

Why _write_ a new glob library?

- I have doubts about the use of `fastwalk` (which is vendored in go-zglob).
- I have doubts that go-zglob could be adapted to fix some of the problems above (it translates glob patterns into regexes under the hood, which makes partial matching for candidate directories difficult, and the code harder to read).
- There are probably other glob libraries that would work, but wheee I'd already started

I realise I'm sneaking in a significant amount of unreviewed code through a new dependency. Please forgive me 🙏 I wanted to iterate quickly on design for the new library. At this point I believe it's at syntax and feature parity with go-zglob, has fewer bugs, and will provide a better experience. 

But because of the risks, it comes under an experiment.

### DIFF
--- a/agent/artifact_uploader.go
+++ b/agent/artifact_uploader.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -14,6 +15,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/DrJosh9000/zzglob"
 	"github.com/buildkite/agent/v3/api"
 	"github.com/buildkite/agent/v3/internal/artifact"
 	"github.com/buildkite/agent/v3/internal/experiments"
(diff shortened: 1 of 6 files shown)


agent/artifact_uploader.go
```diff
@@ -125,19 +125,26 @@ func (a *ArtifactUploader) Collect(ctx context.Context) (artifacts []*api.Artifa
...
+			}
+			if d != nil && d.IsDir() {
+				a.logger.Warn("Glob pattern %s matched a directory %s", globPath, path)
+				return nil
+			}
+			files = append(files, path)
+			return nil
+		}, a.conf.GlobResolveFollowSymlinks)
+		if err != nil {
+			return nil, fmt.Errorf("globbing pattern: %w", err)
```
triarius, line 147: I'm quite confident that this glob will be better, but I think the safe approach is to put it behind an experiment.

triarius (approved):

PR #2347, opened 2023-09-03
Pass errors back up to main.go from clicommand

triarius:
The methods in the clicommand package used to call `logger.Fatal` on some error conditions. However, as there is a `defer` in all of these methods, we should not call `logger.Fatal` unless in extreme circumstances, where it is not possible to continue the cleanup that occurs in the `defer`.

This PR replaces all Fatal logs in the cli Action methods with returns of errors. This will allow the `defer`s to be called. In the main package, these errors are caught, will cause the application to terminate with the appropriate exit code.


internal/job/integration/git.go
triarius, line 1: I've wrapped all the errors and enabled logging to make this easier to debug

internal/job/integration/main_test.go
```diff
@@ -29,11 +29,10 @@ func TestMain(m *testing.M) {
 		}
 
 		if err := app.Run(os.Args); err != nil {
-			fmt.Printf("%v\n", err)
-			os.Exit(1)
+			os.Exit(clicommand.ErrToExitCode(err))
```
triarius, line 32: This is needed for integration tests that check the exit code.

clicommand/errors.go
```diff
@@ -0,0 +1,73 @@
...
+func (e *SilentExitError) Code() int {
+	return e.code
+}
+
+func (e *SilentExitError) Is(target error) bool {
+	terr, ok := target.(*SilentExitError)
+	return ok && e.code == terr.code
+}
+
+func ErrToExitCode(err error) int {
```
DrJosh9000, line 61: I'd defensively catch `err == nil` here (and return 0).

Also, if nothing else is named `ExitCode`, that might be a nicer name for this function?

DrJosh9000 (commented):
Love it! Such an improvement

clicommand/artifact_shasum.go
```diff
@@ -109,18 +109,25 @@ var ArtifactShasumCommand = cli.Command{
...
 		ctx := context.Background()
 		ctx, cfg, l, _, done := setupLoggerAndConfig[ArtifactShasumConfig](ctx, c)
 		defer done()
 
 		if err := searchAndPrintShaSum(ctx, cfg, l, os.Stdout); err != nil {
-			l.Fatal(err.Error())
+			return err
 		}
+
+		return nil
```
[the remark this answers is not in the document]
triarius: Done!

clicommand/errors.go
triarius: So I had a think, and just calling it `ExitCode` is a bit deceptive as it's not a pure function. I've called it `PrintMessageAndReturnExitCode` now. This goes in the wrong direction in terms of method name length, but the right direction for clarity in my opinion.

DrJosh9000: That's fair.

internal/job/integration/executor_tester.go
```diff
@@ -334,11 +334,8 @@ func (e *ExecutorTester) ReadEnvFromOutput(key string) (string, bool) {
 func (e *ExecutorTester) RunAndCheck(t *testing.T, env ...string) {
 	t.Helper()
 
-	err := e.Run(t, env...)
-	t.Logf("Bootstrap output:\n%s", e.Output)
-
-	if err != nil {
-		t.Fatalf("BootstrapTester.Run(%q) = %v", env, err)
+	if err := e.Run(t, env...); shell.GetExitCode(err) != 0 {
+		assert.NilError(t, err, "bootstrap output:\n%s", e.Output)
```
DrJosh9000, line 338: Can you give me a bit more colour on this change?

If the exit code of err is nonzero, assert that... err == nil?

DrJosh9000 (approved):
🕵️🐿️

PR #2361, opened 2023-09-07
Add `allow-repositories` param to enable repository allow-listing

david-poirier:
The `allowed-repositories` command line param allows users to specify an allow-list of regular expression strings. E.g.:

```
buildkite-agent start \
    --build-path=/tmp/buildkite-builds \
    --token 123 \
    --allowed-repositories "^git@github.com:buildkite/.*"
```

If the `allowed-repositories` param is used and a job arrives with a `BUILDKITE_REPO` value that doesn't match an entry in the list then the job is rejected. 

If the `allowed-repositories` param is unused then the behaviour is unchanged from current.

### DIFF
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync"
@@ -567,6 +568,19 @@ type LogWriter struct {
 func (w LogWriter) Write(bytes []byte) (int, error) {
 	w.l.Info("%s", bytes)
 	return len(bytes), nil
+}
+
+func validateRepository(allowedRepos []string, pipelineRepo string) error {
+	if len(allowedRepos) == 0 {
+		return nil
+	}
+
+	for _, allowedRepo := range allowedRepos {
+		if match, _ := regexp.MatchString(allowedRepo, pipelineRepo); match {
+			return nil
+		}
+	}
+	return fmt.Errorf("repository %s isn't allowed", pipelineRepo)
 }
 
 func (r *JobRunner) executePreBootstrapHook(ctx context.Context, hook string) (bool, error) {
--- a/agent/agent_configuration.go
+++ b/agent/agent_configuration.go
@@ -24,6 +24,7 @@ type AgentConfiguration struct {
 	GitCleanFlags         string
 	GitFetchFlags         string
 	GitSubmodules         bool
+	AllowedRepositories   []string
 	SSHKeyscan            bool
 	CommandEval           bool
 	PluginsEnabled        bool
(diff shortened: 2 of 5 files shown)


agent/job_runner.go
```diff
@@ -569,6 +570,19 @@ func (w LogWriter) Write(bytes []byte) (int, error) {
 	return len(bytes), nil
 }
 
+func validateGitRepository(allowedRepos []string, pipelineRepo string) (bool, error) {
```
triarius, line 573: Given, you only ever return `true, nil` or `false, err` where `err != nil`. I think it makes more sense to remove the boolean return value altogether and just return an `error`.

triarius (changes requested):
Thanks for contributing this @david-poirier. We've discussed your rationale for why you don't want to use a pre-bootstrap hook for this, and have been persuaded.

I think we can get this if you change the name of the config to `allowed-repostories` or `allowed-repos` as you alluded to https://github.com/buildkite/agent/pull/2361#issuecomment-1709636409 and refactor `validateGitRepository` to have a signature like
```
func validateGitRepository(allowedRepos []string, pipelineRepo string) error
```

As the allowlist is for `BUILDKITE_REPO`, and the repos won't necessarily be git repositories, I think it makes sense to not mention git in the non-test code. In particular, the name `validateGitRepository` should drop the `Git` as well.

moskyb:
@david-poirier that string on the builds page (`Job never started...`) is a backend thing that's computed based on the `jobExit.SignalReason` (as well as the exit code) that you've set [here](https://github.com/buildkite/agent/pull/2361/files#diff-abe82fbce8aeb792bb04ab30004a198aab38006296721de63f02e0df6a1ca1a9R127). 

i'm happy to do any of the following:
- add a new signal reason (`invalid_repo`? or something?). this would require a backend change  on our end, but it'd be a pretty straightforward thing to do
- keep using the `agent_refused` signal reason - it makes sense, the agent has refused in a very similar way to how a pre-bootstrap hook might - and simply make the message under the logs more vague (eg `Job never started (because the agent refused the job` or something similar), and let the logs do the talking. this would also require an (even simpler) change on the backend
- keep things as-is and have the under-log message lie just a little bit, and hope that the logs make things clear that the repo isn't allowed

also!!! i want to mention that i love this PR (and the [issue attached](https://github.com/buildkite/agent/issues/2368)) - i agree that the `pre-bootstrap` hook can be pretty clunky, and i'm really psyched to see some proposed changes making good security practice easy.

triarius (approved):
Thanks! I'll leave it to @moskyb to decide what to do with the error string.

moskyb:
@david-poirier it's shipped now 🎉
CleanShot 2023-09-15 at 14 21 04@2x

moskyb:
@david-poirier i've just rebased this PR a wee bit, and will merge it when the build finishes. Thanks so much for your contribution!

PR #2365, opened 2023-09-13
Escape path segments in URLs made with fmt.Sprintf

DrJosh9000:
Making up URLs with `fmt.Sprintf` can be dicey. 

Most of these are no-ops in the normal case because job IDs and build IDs have nice formats. However, annotation contexts are arbitrary strings. 

Currently, an annotation context containing `.` can be created (passed through a JSON encoded request body) but not deleted (passed through the URL path). The reason was a bit surprising: Rails handles `.` in path segments specially. So `.` also needs to be escaped. Fortunately, this seems to be sufficient to fix the bug without any backend changes.

### DIFF
--- a/api/steps.go
+++ b/api/steps.go
@@ -18,7 +18,7 @@ type StepExportResponse struct {
 
 // StepExport gets an attribute from step
 func (c *Client) StepExport(ctx context.Context, stepIdOrKey string, stepGetRequest *StepExportRequest) (*StepExportResponse, *Response, error) {
-	u := fmt.Sprintf("steps/%s/export", stepIdOrKey)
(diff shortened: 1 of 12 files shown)


api/steps.go
```diff
@@ -45,7 +45,7 @@ type StepUpdate struct {
 
 // StepUpdate updates a step
 func (c *Client) StepUpdate(ctx context.Context, stepIdOrKey string, stepUpdate *StepUpdate) (*Response, error) {
-	u := fmt.Sprintf("steps/%s", stepIdOrKey)
+	u := fmt.Sprintf("steps/%s", railsPathEscape(stepIdOrKey))
```
triarius, line 48: This one seems a bit risky, but if it works for annotations, it should work here too.

triarius (approved):
🎉

PR #2382, opened 2023-09-21
Add Matrix parsing

DrJosh9000:
Here's my take on it. It's maybe a bit heavy in named types ("`MatrixAdjustmentWith`", very Enterprise Software™), but this accounts for all the ways to write a matrix that I could find through the docs and playing around.

`MatrixScalars` basically exists to validate input. I had previously included the capability to sort `MatrixScalars`, but it isn't needed yet. (Or probably at all. Ordering ought to be stable throughout. I thought a lot about ~~The Matrix Canon~~ canonicalising matrices today.) `MatrixScalars` and `MatrixAdjustmentWith` limit values to string, bool, and int because those are the types that the backend accepts (which, coincidentally, will make it easier to produce a canonical encoding of `Matrix` for signing.)


moskyb:
oh man, this is miles better than mine. i'll close it.

internal/pipeline/parser_test.go
```diff
@@ -1488,3 +1488,438 @@ steps:
...
+    }
+  ]
+}`,
+		},
+	}
+
+	for _, test := range tests {
+		test := test
+		t.Run(test.desc, func(t *testing.T) {
+			//t.Parallel()
```
moskyb, line 1905: should we parallelise these tests?

internal/pipeline/parser_test.go
```diff
@@ -1488,3 +1488,438 @@ steps:
...
+			gotJSON, err := json.MarshalIndent(got, "", "  ")
+			if err != nil {
+				t.Fatalf(`json.MarshalIndent(got, "", "  ") error = %v`, err)
+			}
+			if diff := cmp.Diff(string(gotJSON), test.wantJSON); diff != "" {
+				t.Errorf("marshalled JSON diff (-got +want):\n%s", diff)
+			}
+		})
+	}
+}
```
moskyb, line 1925: this file is getting really long - would it make sense to split it up a bit for easier navigation? happy for this to be done as part of a separate PR

internal/pipeline/step_command.go
```diff
@@ -25,7 +25,7 @@ type CommandStep struct {
 	Plugins   Plugins           `yaml:"plugins,omitempty"`
 	Env       map[string]string `yaml:"env,omitempty"`
 	Signature *Signature        `yaml:"signature,omitempty"`
-	Matrix    any               `yaml:"matrix,omitempty"`
+	Matrix    *Matrix           `yaml:"matrix,omitempty"`
```
moskyb, line 28: 🥳

moskyb (commented):
this is looking really good! i have a couple of comments and niggles, but otherwise it's looking 🔥

internal/pipeline/step_command_matrix.go
```diff
@@ -0,0 +1,225 @@
...
+
+	_ interface {
+		json.Marshaler
+		selfInterpolater
+	} = (*MatrixAdjustment)(nil)
+
+	_ = []interface {
+		json.Marshaler
+		ordered.Unmarshaler
+		yaml.Marshaler
```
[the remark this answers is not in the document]
DrJosh9000: An adjustment always has `with` and sometimes `skip` or `soft_fail`. That's a struct that `ordered.Unmarshal` knows how to fill. `MatrixAdjustmentWith` can either be a map or a single value, and presently `ordered.Unmarshal` can't create "singleton maps". That wouldn't be hard to add though...

internal/pipeline/parser_test.go
```diff
@@ -1488,3 +1488,438 @@ steps:
...
+			},
+			wantJSON: `{
+  "steps": [
+    {
+      "command": "echo {{matrix}}",
+      "matrix": {
+        "setup": [
+          "apple",
+          47,
+          true
```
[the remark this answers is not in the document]
DrJosh9000: That's possible, if there are no adjustments (which we also don't document in that case). Thinking...

internal/pipeline/parser_test.go
DrJosh9000: Done.

internal/pipeline/parser_test.go
DrJosh9000: I've moved the new tests, but the rest can remain for now.

internal/pipeline/parser_test.go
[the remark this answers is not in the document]
DrJosh9000: Done (if the matrix is simple, then it becomes a list again).

moskyb (approved):

PR #2390, opened 2023-09-25
Fix apt install awscli

moskyb:
It installs tzdata, which by default prompts for input when it can't discover what timezone you're in. We fix this by passing the magic variable DEBIAN_FRONTEND=noninteractive


DrJosh9000 (approved):
🙃

PR #2391, opened 2023-09-26
Matrix signing

DrJosh9000:
This includes matrix in the signature. However, any matrix that is not `nil` will (still) fail to verify.

Matrix is required (sneaking in a matrix could alter how a step should be interpreted), which will invalidate any existing signatures. ~~so I'm also taking this opportunity to centralise the canonicalisation of data that gets signed. `writeCanonical` is a bit ASN.1-inspired, but I hate ASN.1, so you get this instead.~~ We went with [JCS](https://rfc-editor.org/rfc/rfc8785.html) instead.

Like `Plugins`, `Matrix` is included in the signed data wholesale. Reasons this is up for debate:

- Data from a future form of matrix (new fields that would unmarshal into `RemainingFields` today) might or might not affect valid matrix combinations in the future, so to be safe we should include `RemainingFields` in the signature...
- `soft_fail` doesn't affect whether a matrix combination is valid, so that could be excluded...
- We only use truthiness of `skip`, so we could canonicalise it to `bool` for signing...

### DIFF
--- a/agent/verify_job.go
+++ b/agent/verify_job.go
@@ -136,6 +136,9 @@ func (r *JobRunner) verifyJob(keySet jwk.Set) error {
 				return newInvalidSignatureError(fmt.Errorf("job %q was signed with signature %q, but the value of BUILDKITE_PLUGINS (%q) does not match the value of step.plugins (%q)", r.conf.Job.ID, step.Signature.Value, jobPluginsNorm, stepPluginsNorm))
 			}
 
+		case "matrix": // compared indirectly through other fields
+			continue
+
 		default:
 			// env:: - skip any that were verified with Verify.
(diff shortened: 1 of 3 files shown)


sj26:
13591418-32fa-4406-89d1-ed6e34e805e2_text

triarius (commented):
I'm not sure that we should be maintaining a bespoke serialisation format. I'm going to need some strong arguments for why this is a good idea.

DrJosh9000:
> I'm not sure that we should be maintaining a bespoke serialisation format. I'm going to need some strong arguments for why this is a good idea.

As discussed, it's all JCS now.

moskyb (approved):
image

triarius (approved):
💚 LGTM

Issue #2368, opened 2023-09-14
Simple options to protect agents in case of Buildkite being compromised


david-poirier:
Given this scenario - buildkite gets compromised and malicious jobs are sent to some or all agents connected to the buildkite agent controller - I feel that the current (and proposed) mitigations are either insufficient or too complex.

Current advice (at https://buildkite.com/docs/agent/v3/securing#restricting-access-by-the-buildkite-agent-controller) is:

- use hooks to allow-list repos, plugins, and commands
- disable local hooks
- disable plugins
- disable command eval (which also disables plugins)

Using hooks does work but is difficult and error-prone at scale - ensuring they work as intended across different operating systems, and aren't removed, overwritten or broken by other engineers with limited security background is... challenging.

Disabling local hooks and plugins solves parts of the scenario but removes critical and commonly used agent functionality.

Disabling command eval solves a big part of the scenario, but by disabling plugins also removes critical functionality.

Pipeline signing (https://github.com/buildkite/buildkite-signed-pipeline) does work as well, but requires significant engineering effort to implement in its current form (and less but still significant effort once it's integrated into the agent).

---------------------------

I'd like to propose a simple set of agent configuration options that I believe can protect against the `buildkite gets compromised` scenario, without the complexity of hooks or pipeline signing.

- Protection from malicious repositories - An `allowed-repositories` command line param accepting an allow-list of regular expression strings. e.g.: `^git@github.com/buildkite/.*$`. If the param is used and a job arrives with a `BUILDKITE_REPO` value that doesn't match an entry in the list then the job is rejected - [merged PR here](https://github.com/buildkite/agent/pull/2361).
- Protection from malicious plugins - An `allowed-plugins` command line param accepting an allow-list of regular expression strings. e.g.: `^github.com/buildkite/.*$`. If the param is used and a job arrives with `BUILDKITE_PLUGINS` containing a plugin that doesn't match an entry in the list then the job is rejected. - [merged PR here](https://github.com/buildkite/agent/pull/2471)
- Protection from malicious environment variables - An `allowed-environment-variables` command line param accepting an allow-list of regular expression strings, e.g.: `^BUILD_.*$`. If the param is used then any environment variables that don't match an entry in the list are ignored. - [merged PR here](https://github.com/buildkite/agent/pull/2539)
[...]

PR #2406, opened 2023-10-03
Adding allowed environment variables option to only allow listed

CheeseStick:
## Context

Protection from malicious environment variables - An allowed-environment-variables command line param accepting an allow-list of regular expression strings, e.g., `^BUILDKITE_.*$` or `BUILDKITE_.*`. If the param is used, any environment variables that don't match an entry in the list, the agent will refuse to run a job.

This work is part of the [issue](https://github.com/buildkite/agent/issues/2368) about `buildkite-agent` security.


agent/job_runner_test.go
```diff
@@ -37,3 +37,35 @@ func TestValidateRepository(t *testing.T) {
 	err = validateRepository(allowedRepos, pipelineRepo)
 	require.NoError(t, err)
 }
+
+func TestAllowedEnvironmentVariables(t *testing.T) {
+	// Allow list without regex
+	var envVarsWithoutRegex = map[string]string{
```
triarius, line 43:
```suggestion
	envVarsWithoutRegex := map[string]string{
```

This is more idomatic go.

clicommand/agent_start.go
```diff
@@ -213,6 +214,10 @@ func (asc AgentStartConfig) Features(ctx context.Context) []string {
 		features = append(features, "allowed-repositories")
 	}
 
+	if len(asc.AllowedEnvironmentVariables) > 0 {
+		features = append(features, "allowed-environment-variables")
+	}
+
```
triarius, line 220: This requires some changes to the backend, so it won't work for now. It's only used internally so shouldn't be a blocker to you, so we recommend deleting it.
```suggestion
```

agent/job_runner.go
```diff
@@ -392,6 +394,33 @@ func (r *JobRunner) normalizeVerificationBehavior(behavior string) (string, erro
...
 }
 
+func filterEnv(env map[string]string, allowedVariables []string) []string {
+	var deniedVariables []string
+
+	// Filter out if the user has unlisted environment variables
+	for k := range env {
+		isDenied := true
+
+		for _, v := range allowedVariables { // Loop through allowed environment variables list
```
jakubm-canva, line 404: Remove that comment. Feels redundant as the code is clean.
```suggestion
		for _, v := range allowedVariables {
```

jakubm-canva, line 404:
```
// Add every environment variable to denied list and prune them if match allowed list.
for v := range env {
		deniedVariables = append(deniedVariables, v)
    for _, v := range allowedVariables {
				if isMatch, _ := regexp.MatchString(v, k); isMatch {
            deniedVariables[:len(deniedVariables)-1]
        }
    }
}
```

jakubm-canva (approved):

agent/job_runner.go
```diff
@@ -435,6 +464,24 @@ func (r *JobRunner) createEnvironment(ctx context.Context) ([]string, error) {
...
+	if len(r.conf.AgentConfiguration.AllowedEnvironmentVariables) > 0 {
+		// Fetch environment variables that are allowed by default
+		var allowedEnv = make([]string, len(DefaultAllowedEnv))
+		copy(allowedEnv, DefaultAllowedEnv)
+
+		// And add allowed environment variables from agent configuration
+		allowedEnv = append(allowedEnv, r.conf.AgentConfiguration.AllowedEnvironmentVariables...)
+
+		// Filter to get denied envs
+		deniedEnv := filterEnv(env, allowedEnv)
```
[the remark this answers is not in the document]
CheeseStick: I just updated the function to return `env` and also changed some codes to make it a pure function.

agent/job_runner.go
```diff
@@ -392,6 +394,33 @@ func (r *JobRunner) normalizeVerificationBehavior(behavior string) (string, erro
...
+		isDenied := true
+
+		for _, v := range allowedVariables { // Loop through allowed environment variables list
+			if isMatch, _ := regexp.MatchString(v, k); isMatch { // And find if any matches the regex
+				isDenied = false
+				break
+			}
+		}
+
+		// If nothing got matched,
```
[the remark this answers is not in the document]
CheeseStick: I deleted the comment. 😁

agent/job_runner.go
CheeseStick: Not a bad idea! Updated the code!

agent/job_runner.go
```diff
@@ -392,6 +394,36 @@ func (r *JobRunner) normalizeVerificationBehavior(behavior string) (string, erro
...
 
+func filterEnv(env map[string]string, allowedVariables []string) (map[string]string, []string) {
+	// Copy all environment variables to the new map
+	// So it won't modify the original list
+	var environmentVariables = make(map[string]string, len(env))
+	for k := range env {
+		environmentVariables[k] = env[k]
+	}
+
+	var deniedVariables []string
```
jakubm-canva, line 405: nit: keep it tight (var deniedVariables). no need for additional whitespaces as this parts of code a logically together.
```
// Filter out if the user has unlisted environment variables
// It adds environment variable to denied list and removes if it finds matching regex
var deniedVariables []string
for k := range environmentVariables {
		deniedVariables = append(deniedVariables, k)
		for _, v := range allowedVariables {
			if isMatch, _ := regexp.MatchString(v, k); isMatch {
				deniedVariables = deniedVariables[:len(deniedVariables)-1]
				break
			}
		}
	}
```

agent/job_runner.go
```diff
@@ -392,6 +394,36 @@ func (r *JobRunner) normalizeVerificationBehavior(behavior string) (string, erro
 	}
 }
 
+func filterEnv(env map[string]string, allowedVariables []string) (map[string]string, []string) {
+	// Copy all environment variables to the new map
+	// So it won't modify the original list
+	var environmentVariables = make(map[string]string, len(env))
```
jakubm-canva, line 400: I find it confusing that similar naming pattern is for a structure that is a map.
```
func filterEnv(env map[string]string, allowedVariables []string)
  var environmentVariables = make(map[string]string, len(env))
```

```
env -> envToValues
environmentVariables -> allowedEnvToValues
```

agent/job_runner.go
```diff
@@ -435,6 +467,26 @@ func (r *JobRunner) createEnvironment(ctx context.Context) ([]string, error) {
...
+		var allowedEnv = make([]string, len(DefaultAllowedEnv))
+		copy(allowedEnv, DefaultAllowedEnv)
+
+		var deniedEnv []string
+
+		// And add allowed environment variables from agent configuration
+		allowedEnv = append(allowedEnv, r.conf.AgentConfiguration.AllowedEnvironmentVariables...)
+
+		// Filter to get denied envs
+		env, deniedEnv = filterEnv(env, allowedEnv)
```
jakubm-canva, line 482: Are you purposely overwriting that env?
```suggestion
		allowedEnvToValues, deniedEnv := filterEnv(env, allowedEnv)
```

clicommand/agent_start.go
```diff
@@ -965,6 +973,20 @@ var AgentStartCommand = cli.Command{
...
+		// Show allowed environment variables if it is set
+		if len(agentConf.AllowedEnvironmentVariables) > 0 {
+			l.Info("Environment variable allow list is enabled. Only allow-listed environment variables can be set.")
+			l.Debug("Allowed environment variables:")
+
+			for _, v := range agentConf.AllowedEnvironmentVariables {
+				l.Debug(" - %s", v)
+
+				if v[0] != '^' || v[len(v)-1] != '$' {
+					l.Warn(" - %s is not anchored! It can result environment variable leak with partial match.", v)
```
jakubm-canva, line 985:
```suggestion
					l.Warn(" - %s is not anchored! It can result in environment variable leak with partial match.", v)
```

agent/job_runner.go
CheeseStick: Yup, it needs to be overwritten as the function doesn't edit `env` but creates a new `map` with filtered environment variables.

agent/job_runner.go
CheeseStick: Just updated the code to make it more explicit which will remove the naming confusion.

agent/job_runner.go
```diff
@@ -392,6 +394,28 @@ func (r *JobRunner) normalizeVerificationBehavior(behavior string) (string, erro
 	}
 }
 
+func filterEnv(environmentVariables map[string]string, allowedVariables []string) []string {
```
jakubm-canva, line 397: nit: I would either call it env (like across entire BK/agent repo, my pref) or envToValues (to emphasis its a map not slice)

agent/job_runner.go
```diff
@@ -435,6 +459,26 @@ func (r *JobRunner) createEnvironment(ctx context.Context) ([]string, error) {
...
+		var allowedEnv = make([]string, len(DefaultAllowedEnv))
+		copy(allowedEnv, DefaultAllowedEnv)
+
+		var deniedEnv []string
+
+		// And add allowed environment variables from agent configuration
+		allowedEnv = append(allowedEnv, r.conf.AgentConfiguration.AllowedEnvironmentVariables...)
+
+		// Filter to get denied envs
+		deniedEnv = filterEnv(env, allowedEnv)
```
jakubm-canva, line 474: remove
`var deniedEnv []string`
and make in-line:
`deniedEnv := filterEnv(env, allowedEnv)`

I would merge this together again:
```

// Set BUILDKITE_DENIED_ENV so the bootstrap can show warnings
deniedEnv := filterEnv(env, allowedEnv)
if len(deniedEnv) > 0 {
   	env["BUILDKITE_DENIED_ENV"] = strings.Join(deniedEnv, ",")
}
```

agent/job_runner.go
```diff
@@ -392,6 +394,28 @@ func (r *JobRunner) normalizeVerificationBehavior(behavior string) (string, erro
 	}
 }
 
+func filterEnv(environmentVariables map[string]string, allowedVariables []string) []string {
+	// Filter out if the user has unlisted environment variables
+	// It adds environment variable to denied list and removes if it finds matching regex
+	var deniedVariables []string
```
jakubm-canva, line 400: I would like to see the consistency, than mixing. Either we call things:
allowedEnv / deniedEnv OR allowedVariables / deniedvariables

agent/job_runner.go
jakubm-canva: Or even:
if deniedEnv := filterEnv(env, allowedEnv); len(deniedEnv) > 0 {
   	env["BUILDKITE_DENIED_ENV"] = strings.Join(deniedEnv, ",")
}

agent/job_runner.go
jakubm-canva, line 400: nit: a small performance improvement would be to use make
deniedVariables := make([]string, len(environmentVariables))
Then you avoid resizing it.

CheeseStick: I tried `make` and it failed when removing the environment variable from the `deniedVariables` list when it got a match, which I think, is because the control logic depends on the list size.

agent/job_runner.go
```diff
@@ -429,6 +454,21 @@ func (r *JobRunner) createEnvironment(ctx context.Context) ([]string, error) {
 		env["BUILDKITE_IGNORED_ENV"] = strings.Join(ignoredEnv, ",")
 	}
 
+	// Check if allowed environment variables are set
+	if len(r.conf.AgentConfiguration.AllowedEnvironmentVariables) > 0 {
+		// Fetch environment variables that are allowed by default
+		var allowedVariables = make([]string, len(DefaultAllowedEnv))
```
moskyb, line 460:
```suggestion
		allowedVariables := make([]string, len(DefaultAllowedEnv))
```

agent/job_runner.go
```diff
@@ -386,6 +388,29 @@ func (r *JobRunner) normalizeVerificationBehavior(behavior string) (string, erro
...
+	}
+
+	// Delete denied variables from environment variables list
+	for _, v := range deniedVariables {
+		delete(env, v)
+	}
+
+	return deniedVariables
+}
+
```
moskyb, line 413: this approach of stripping disallowed environment variables out is quite different to the ones in https://github.com/buildkite/agent/pull/2361 and https://github.com/buildkite/agent/pull/2394 - both of them refuse the job, instead of stripping the disallowed values out. is there a reason that you decided to do this?

moskyb (changes requested):

agent/job_runner.go
jakubm-canva: I suggested this approach rather than keeping a boolean that is checked after iterating finishes in L#398, but I do see your point in the consistency of building the set.

agent/job_runner_test.go
```diff
@@ -37,3 +37,79 @@ func TestValidateRepository(t *testing.T) {
...
+		allowedEnvironmentVariables: []string{"SOME_EXPLICITLY_ALLOWED_VARIABLE", "SOME_ALLOWED_.*", "^ANOTHER_ALLOWED_.*$"},
+		expectedEnvironmentVariables: map[string]string{
+			"SOME_EXPLICITLY_ALLOWED_VARIABLE": "I_SHOULD_BE_ALLOWED",
+			"SOME_ALLOWED_VARIABLE":            "I_ALSO_SHOULD_BE_ALLOWED",
+			"SOME_ALLOWED_ANOTHER_VARIABLE":    "I_ALSO_ALSO_SHOULD_BE_ALLOWED",
+			"ANOTHER_ALLOWED_VARIABLE":         "I_ALSO_ALSO_ALSO_SHOULD_BE_ALLOWED",
+			"ANOTHER_ALLOWED_ANOTHER_VARIABLE": "I_ALSO_ALSO_ALSO_ALSO_SHOULD_BE_ALLOWED",
+		},
+		expectedDeniedVariables: []string{"SOME_NOT_ALLOWED_VARIABLE"},
+	}}
```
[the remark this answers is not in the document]
CheeseStick: That totally makes sense! I just updated the test to include some actual attack factors using environment variables and also replaced the allowed one with the Buildkite one for better readability.

agent/job_runner.go
```diff
@@ -75,6 +75,8 @@ var ProtectedEnv = map[string]struct{}{
 	"BUILDKITE_SHELL":                    {},
 }
 
+var DefaultAllowedEnv = []string{"^BUILDKITE$", "^BUILDKITE_.*$", "^CI$"}
+
```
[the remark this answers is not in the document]
CheeseStick: I first thought about using a protected environment variables list, as filtering out all the environment variables BK set can break the Buildkite agent's behaviour. However, I soon found that the protected list doesn't include all the environment variables that the agent sets and added that default allowed environment variables list there.
I don't think removing the default list is a good idea, but as it can break some workflows, maybe some warnings can also be added to the doc. Something like "not including `"^BUILDKITE$", "^BUILDKITE_.*$", "^CI$"` can break the buildkite agent."
WDYT?

clicommand/agent_start.go
```diff
@@ -545,6 +546,12 @@ var AgentStartCommand = cli.Command{
 			Usage:  "A comma-separated list of regular expressions representing repositories the agent is allowed to clone (for example, \"^git@github.com:buildkite/.*\" or \"^https://github.com/buildkite/.*\")",
 			EnvVar: "BUILDKITE_ALLOWED_REPOSITORIES",
 		},
+		cli.StringSliceFlag{
+			Name:   "allowed-environment-variables",
+			Value:  &cli.StringSlice{},
+			Usage:  "A comma-separated list of regular expressions representing environment variables the agent is allowed to use (e.g. \"^BUILDKITE_.*$\" or \"BUILDKITE_.*\")",
```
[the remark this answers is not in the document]
CheeseStick: Good point, But as you mentioned, commas are not allowed in environment variable names, so it would be okay!

clicommand/agent_start.go
```diff
@@ -938,6 +946,20 @@ var AgentStartCommand = cli.Command{
...
+		if len(agentConf.AllowedEnvironmentVariables) > 0 {
+			l.Info("Environment variable allow list is enabled. Only allow-listed environment variables can be set.")
+			l.Debug("Allowed environment variables:")
+
+			for _, v := range agentConf.AllowedEnvironmentVariables {
+				l.Debug(" - %s", v)
+
+				if v[0] != '^' || v[len(v)-1] != '$' {
+					l.Warn(" - %s is not anchored! It can result in environment variable leak with partial match.", v)
+				}
```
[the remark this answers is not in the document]
CheeseStick: Thanks for the suggestion! I added the regex validation step to the agent start-up process.

And with the warning about unanchored regex, I agree that it can be okay if the user knows what they are doing but wants to warn as it can be a mistake, and it's always good to have some warnings if it's related to security. 😁 
But it also can get the point as it can be... too much, spamming the terminal.

Also, maybe my wording must not be clear with the message, `environment variable leak with partial match`. I will try to come up with something more explicit. What I meant there is that unanchored regex can end up allowing everything, like `PATH` can end up allowing `LD_LIBRARY_PATH` and any other that includes `PATH` as a partial match is also a regex match. (As it's an allow-list, not deny-list, bad configuration can end up leaking bad env vars to the actual environment.)

agent/job_runner.go
CheeseStick: Ah, I think @moskyb meant why it's not rejecting the job if it has bad variables set, unlike other PRs.

And that's actually really a good point. The reason for just filtering out is that an unknown repo or plugin can actually run some malicious code. However, with environment variables, I thought it's quite limited what attackers can do with it. So I just made it to filter the unknown values.

clicommand/agent_start.go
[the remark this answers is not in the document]
CheeseStick: I think the test should be located where it parses the command line arguments, But I couldn't find them. 🤔

clicommand/agent_start.go
```diff
@@ -939,6 +948,23 @@ var AgentStartCommand = cli.Command{
 			l.Info("Agents will disconnect after %d seconds of inactivity", agentConf.DisconnectAfterIdleTimeout)
 		}
 
+		// Show allowed environment variables if it is set
+		if len(agentConf.AllowedEnvironmentVariables) > 0 {
+			l.Info("Environment variable allow list is enabled. Only allow-listed environment variables can be set.")
```
moskyb, line 953:
```suggestion
```

none of the other config options emit a log when they get turned on, i think it's safe to get rid of this one. we can leave the debug logs though, it's okay to be a bit verbose with those

clicommand/agent_start.go
```diff
@@ -939,6 +948,23 @@ var AgentStartCommand = cli.Command{
...
+			l.Debug("Allowed environment variables:")
+
+			for _, v := range agentConf.AllowedEnvironmentVariables {
+				if _, err := regexp.Compile(v); err != nil {
+					l.Fatal("Regex %s in allowed-environment-variables failed to compile: %v", v, err)
+				} else {
+					l.Debug(" - %s", v)
+					if v[0] != '^' || v[len(v)-1] != '$' {
+						l.Warn(" - %s is not anchored! It can result in environment variable leak with partial match.", v)
+					}
```
moskyb, line 963: i've come around on this one a bit, but my opinion has evolved to the point that i think that we should either:
- allow whatever, and let customers shoot themselves in the foot if they want - we trust that they know how to write a regex, or:
- disallow any and all regexes not anchored at the start and end, forcing them to fully anchor all regexes

i think the middle ground of warning on what could well be a normal use-case is a bit of a half measure - i'm happy with either of the above options

moskyb: oh also! we can get rid of the `else` above, the program will have exited if there was an error, so being past that line means that there's implicitly no error - see [this passage in Effective Go](https://go.dev/doc/effective_go#if:~:text=In%20the%20Go%20libraries%2C%20you%27ll%20find%20that%20when%20an%20if%20statement%20doesn%27t%20flow%20into%20the%20next%20statement%E2%80%94that%20is%2C%20the%20body%20ends%20in%20break%2C%20continue%2C%20goto%2C%20or%20return%E2%80%94the%20unnecessary%20else%20is%20omitted.)

moskyb (changes requested):
hey, sorry about the wait on this! it's getting really close, thanks for sticking with it.

CheeseStick:
@moskyb Can you PTAL 🙏

moskyb:
hey @CheeseStick - i've opened a new PR #2539 that has the code from this PR + some other changes for usability and tests. i'm going to close this PR, but the one linked above should be merged soon and made a part of agent v3.61. thank you os much for your contribution!!!!

CheeseStick:
Thanks!!! 😁

PR #2407, opened 2023-10-03
Fix parsing pipelines what use a string as the skip key in a matrix adjustment

moskyb:
Also, demote SoftFail on matrix adjustments to being a RemainingField, as it's unused by the agent

### DIFF
--- a/internal/pipeline/step_command_matrix.go
+++ b/internal/pipeline/step_command_matrix.go
@@ -176,11 +176,23 @@ type MatrixAdjustments []*MatrixAdjustment
 // MatrixAdjustment models an adjustment - a combination of (possibly new)
 // matrix values, and skip/soft fail configuration.
 type MatrixAdjustment struct {
-	With     MatrixAdjustmentWith `yaml:"with"`
-	Skip     bool                 `yaml:"skip,omitempty"`
-	SoftFail bool                 `yaml:"soft_fail,omitempty"`
+	With MatrixAdjustmentWith `yaml:"with"`
+	Skip any                  `yaml:"skip,omitempty"`
 
-	RemainingFields map[string]any `yaml:",inline"`
+	RemainingFields map[string]any `yaml:",inline"` // NB: soft_fail is in the remaining fields
+}
+
+func (ma *MatrixAdjustment) ShouldSkip() bool {
+	switch s := ma.Skip.(type) {
+	case bool:
+		return s
+
+	case nil:
+		return false
+
+	default:
+		return true
+	}
 }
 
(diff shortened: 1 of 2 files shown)


internal/pipeline/step_command_matrix.go
```diff
@@ -176,11 +176,23 @@ type MatrixAdjustments []*MatrixAdjustment
...
+	RemainingFields map[string]any `yaml:",inline"` // NB: soft_fail is in the remaining fields
+}
+
+func (ma *MatrixAdjustment) ShouldSkip() bool {
+	switch s := ma.Skip.(type) {
+	case bool:
+		return s
+
+	case string:
+		return s != ""
```
triarius, line 191: Pretty sure the empty string is not falsy in ruby. So if the backend sees
```yaml
skip: ""
```
it will adjust to skip.

Unless, that is, if the backend is using `.emtpy?` or the like. But I don't think it is:
```ruby
      jobs.reject! do |job|
        if job.matrix >= adjustment.with
          job.soft_fail = adjustment.soft_fail

          # Return the skip value so `reject!` will remove matching elements
          adjustment.skip
        end
      end
```

internal/pipeline/steps.go
```diff
@@ -142,11 +142,6 @@ func (s Steps) sign(env map[string]string, key jwk.Key) error {
 	for _, step := range s {
 		switch step := step.(type) {
 		case *CommandStep:
-			if step.Matrix != nil {
-				// Don't sign matrix steps... yet
-				continue
-			}
-
```
triarius, line 149: Maybe do this in another PR?

internal/pipeline/step_command_matrix.go
```diff
@@ -176,11 +176,23 @@ type MatrixAdjustments []*MatrixAdjustment
...
+
+	default:
+		return false
```
triarius, line 194: Does this mean things like
```
skip: 1
```
are equivalent to
```
skip: false
```

Not sure if that's what we want

triarius (commented):
I've assumed `ShouldSkip` should have the same logic as what removes adjustments from the jobs array. I don't think it currently does.

internal/pipeline/steps.go
moskyb: oh yes, definitely, that slipped in. will remove.

internal/pipeline/step_command_matrix.go
moskyb: yeah, i think you're right. in ruby, only `nil` and `false` are truthy, so i guess we should just check for those

triarius (approved):
LGTM!

PR #2409, opened 2023-10-05
Permute matrices

moskyb:
**This PR requires #2395**

This PR: Adds the ability for the agent to be able to "permute" a matrix - ie, take a matrix and explode out every combination of it's dimension values. This capability will be necessary when verifying the signatures of matrix steps, as we will need to verify that the step job is a valid construction of the matrix step. to do this, we need to be able to generate every possible output of a matrix, including skips and additions created by the matrix's `adjustments`.

This PR also introduces [kr/pretty](https://github.com/kr/pretty) as a test dependency; it's used to output nicely formatted output when tests fail. In this case, we can't use the `cmp.Diff` package that we normally do, as it's tricky to get it to work when we don't care about ordering. While `cmp` does offer the [`cmpopts.SortSlices`](https://pkg.go.dev/github.com/google/go-cmp/cmp/cmpopts#SortSlices) function, it's tricky to get it to work with the `PermutationSet` type, as it's a doubly-nested slice where we don't care about ordering at either level.

### DIFF
--- a/internal/pipeline/step_command_matrix_permute.go
+++ b/internal/pipeline/step_command_matrix_permute.go
@@ -1,9 +1,8 @@
 package pipeline
 
 import (
-	"golang.org/x/exp/slices"
-
 	"golang.org/x/exp/maps"
+	"golang.org/x/exp/slices"
 )
 
 type PermutationSet []MatrixPermutation
@@ -25,7 +24,6 @@ func (m *Matrix) Permute() PermutationSet {
 		for _, value := range v {
 			valuesByDimension[k] = append(valuesByDimension[k], SelectedDimension{Dimension: k, Value: value})
 		}
-
 	}
 
 	p := cartesianProduct(maps.Values(valuesByDimension)...)
@@ -37,21 +35,21 @@ func (m *Matrix) Permute() PermutationSet {
 	return adjustPermutations(permutations, m.Adjustments)
 }
 
-func cartesianProduct[T any](slices ...[]T) [][]T {
-	if len(slices) == 0 {
+func cartesianProduct[T any](slcs ...[]T) [][]T {
+	if len(slcs) == 0 {
 		return [][]T{}
 	}
 
 	// Initialize the result with the first slice
 	result := [][]T{}
-	for _, element := range slices[0] {
+	for _, element := range slcs[0] {
 		result = append(result, []T{element})
 	}
 
 	// Iterate through the remaining slices and combine with the existing result
-	for i := 1; i < len(slices); i++ {
+	for i := 1; i < len(slcs); i++ {
 		tempResult := [][]T{}
-		for _, element := range slices[i] {
+		for _, element := range slcs[i] {
 			for _, existing := range result {
 				// Create a copy of the existing slice and add the new element
 				newSlice := make([]T, len(existing))
--- a/internal/pipeline/step_command_matrix_permute.go
+++ b/internal/pipeline/step_command_matrix_permute.go
@@ -1,0 +1,146 @@
+package pipeline
+
+import (
+	"golang.org/x/exp/slices"
+
+	"golang.org/x/exp/maps"
+)
+
+type PermutationSet []MatrixPermutation
+
+func (ps PermutationSet) Has(p MatrixPermutation) bool {
+	for _, pp := range ps {
+		if pp.Equals(p) {
+			return true
+		}
+	}
+
+	return false
+}
+
+func (m *Matrix) Permute() PermutationSet {
+	valuesByDimension := make(map[string][]SelectedDimension, len(m.Setup))
+	for k, v := range m.Setup {
+		valuesByDimension[k] = make([]SelectedDimension, 0, len(v))
+		for _, value := range v {
+			valuesByDimension[k] = append(valuesByDimension[k], SelectedDimension{Dimension: k, Value: value})
+		}
+
+	}
+
+	p := cartesianProduct(maps.Values(valuesByDimension)...)
+	permutations := make(PermutationSet, 0, len(p))
+	for _, variant := range p {
+		permutations = append(permutations, MatrixPermutation(variant))
+	}
+
+	return adjustPermutations(permutations, m.Adjustments)
+}
+
+func cartesianProduct[T any](slices ...[]T) [][]T {
+	if len(slices) == 0 {
+		return [][]T{}
+	}
+
+	// Initialize the result with the first slice
+	result := [][]T{}
+	for _, element := range slices[0] {
+		result = append(result, []T{element})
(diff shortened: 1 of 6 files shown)


internal/pipeline/step_command_matrix_permute.go
```diff
@@ -0,0 +1,144 @@
...
+
+func isSkippable(perm MatrixPermutation, skips []MatrixPermutation) bool {
+	for _, skip := range skips {
+		if perm.Equals(skip) {
+			return true
+		}
+	}
+
+	return false
+}
```
moskyb, line 144: this implementation makes adjusting the matrix O(n²), as for each permutation, we have to check all the skips to make sure that that permutation shouldn't be skipped - it's a bit inefficient, but because `MatrixPermutation`s can't be keys in maps, it's hard to make it nicer. 

I suppose i could make a tree out of dimension values and store the skip status at the root of that, or something? but it seems like a lot.

internal/pipeline/step_command_matrix_permute.go
```diff
@@ -0,0 +1,144 @@
...
+			for _, existing := range result {
+				// Create a copy of the existing slice and add the new element
+				newSlice := make([]T, len(existing))
+				copy(newSlice, existing)
+				newSlice = append(newSlice, element)
+				tempResult = append(tempResult, newSlice)
+			}
+		}
+		result = tempResult
+	}
```
moskyb, line 62: this implementation of `cartesianProduct` allocates a lot of slices, though they should all get GC'd at the end. chatGPT suggested this alternative that'll allocate fewer slices:
```Go
func cartesianProduct(slices [][]interface{}) [][]interface{} {
	if len(slices) == 0 {
		return [][]interface{}{}
	}

	// Calculate the total number of combinations
	totalCombinations := 1
	for _, slice := range slices {
		totalCombinations *= len(slice)
	}

	// Initialize the result with the correct capacity
	result := make([][]interface{}, 0, totalCombinations)

	// Initialize indices for each slice
	indices := make([]int, len(slices))

	for {
		// Generate a combination using the current indices
		combination := make([]interface{}, len(slices))
		for i, index := range indices {
			combination[i] = slices[i][index]
		}
		result = append(result, combination)

		// Increment the last index and propagate the carry
		lastIndex := len(indices) - 1
		indices[lastIndex]++
		for i := lastIndex; i > 0 && indices[i] == len(slices[i]); i-- {
			indices[i] = 0
			indices[i-1]++
		}

		// Check if all combinations have been generated
		if indices[0] == len(slices[0]) {
			break
		}
	}

	return result
}
```

but imo it's not as easy to read, and n is likely to be relatively small for build matrices(???)

triarius: I think separating the incrementing of the indices out makes this a bit more readable. Here's my attempt. It's essentially the same as the ChatGPT one, but allocates an extra O(dim) memory to keep track of the lengths of each of the input slices. We can get rid of that too, but that would require manually in lining the increment method, which may reduce readability.

```go
type counters struct {
	counts []int
	bounds []int
}

// newCounters returns a counters struct were all counters are zero.
// the bounds are the lengths of the slices in xss.
func newCounters[T any](xss [][]T) *counters {
	cs := &counters{
		counts: make([]int, len(xss)),
		bounds: make([]int, len(xss)),
	}
	for i, xs := range xss {
		cs.bounds[i] = len(xs)
	}
	return cs
}

// increment adds one to the last counter. If the counter overflows,
// it resets to zero and increments the previous counter, and so on until there
// are no more counters to increment.
func (cs *counters) increment() {
	for i := len(cs.counts) - 1; i >= 0; i-- {
		cs.counts[i]++
		if cs.counts[i] < cs.bounds[i] {
			return
		}
		cs.counts[i] = 0
	}
}

func (cs *counters) countAt(i int) int {
	return cs.counts[i]
}

// cartesianProduct returns the Cartesian product of the input slices.
func cartesianProduct[T any](xss ...[]T) [][]T {
	dim := len(xss)
	if dim == 0 {
		return [][]T{}
	}

	// O(N) allocation of counters
	counters := newCounters(xss)

	// preallocation of output
	outputSize := 1
	for _, xs := range xss {
		outputSize *= len(xs)
	}

	if outputSize == 0 {
		return [][]T{}
	}

	// each slice in the output has the same dimension
	output := make([][]T, outputSize)
	for i := 0; i < outputSize; i++ {
		output[i] = make([]T, dim)
	}

	// counters stores the position in each slice of the elements in the current tuple.
	// So we can use it to index into the slices and fill the output.
	// Then increment the counters to get the positions for the next tuple.
	for i := 0; i < outputSize; i++ {
		for j := 0; j < dim; j++ {
			output[i][j] = xss[j][counters.countAt(j)]
		}
		counters.increment()
	}

	return output
}
```

triarius: Ooo, we can use a closure to do instead:

```go
// cartesianProduct returns the Cartesian product of the input slices.
func cartesianProduct[T any](xss ...[]T) [][]T {
	dim := len(xss)
	if dim == 0 {
		return [][]T{}
	}

	counters := make([]int, len(xss))

	// increment adds one to the last counter. If the counter overflows,
	// it resets to zero and increments the previous counter, and so on until there
	// are no more counters to increment.
	increment := func(counters []int) {
		for i := len(counters) - 1; i >= 0; i-- {
			counters[i]++
			if counters[i] < len(xss[i]) {
				return
			}
			counters[i] = 0
		}
	}

	// preallocation of output
	outputSize := 1
	for _, xs := range xss {
		outputSize *= len(xs)
	}

	if outputSize == 0 {
		return [][]T{}
	}

	// each slice in the output has the same dimension
	output := make([][]T, outputSize)
	for i := 0; i < outputSize; i++ {
		output[i] = make([]T, dim)
	}

	// counters stores the position in each slice of the elements in the current tuple.
	// So we can use it to index into the slices and fill the output.
	// Then increment the counters to get the positions for the next tuple.
	for i := 0; i < outputSize; i++ {
		for j := 0; j < dim; j++ {
			output[i][j] = xss[j][counters[j]]
		}
		increment(counters)
	}

	return output
}
```

DrJosh9000:
I don't understand why generating the Cartesian product is needed, if we're provided with the choice of matrix parameters when verifying.

In the basic case (no adjustments), if a matrix is made up of dimensions `A, B, C`, then surely we can test that a selection of parameters `(a, b, c)` is valid by scanning through `A`, `B`, and `C` individually (in `O(|A|+|B|+|C|)`?

In the case with adjustments, can we not just do the same thing, but then scan through the adjustments to check for skips and additions?

moskyb:
😶

you're totally right.

Issue #2408, opened 2023-10-05
Patch agent OCI image for cve-2023-4911


andy-v-h:
**Is your feature request related to a problem? Please describe.**

:wave: I have a build that relies on the buildkite/agent:3-ubuntu image. It is currently failing my CVE scanner for [cve-2023-4911](https://avd.aquasec.com/nvd/cve-2023-4911)

**Describe the solution you'd like**
Rebuilding and pushing new images for the agent that are ubuntu based would be very helpful 

**Describe alternatives you've considered**
Waiting for the next agent release. 

**Additional context**

andy-v-h:
I did attempt to just upgrade `libc-bin`, no dice. Might just have to wait for Canonical on this one.

triarius:
@andy-v-h, FYI, the missing link may have been `libc6`, which also needed upgrading.

PR #2410, opened 2023-10-05
Upgrade pre-installed packages in docker images

triarius:
Sometimes, the upstream distro will release a package but we had to wait
until they made a new image before our images would get the new package.
This means the security scanners will report fixable vulnerabilities that a
rebuild would not fix. With this change, we just have to make a new
build to pull in the patched packages.

Fixes: \#2408

### DIFF
--- a/packaging/docker/alpine-k8s/Dockerfile
+++ b/packaging/docker/alpine-k8s/Dockerfile
@@ -2,7 +2,7 @@
 
 FROM alpine:3.18.3 AS base
 
-RUN apk update && apk add --no-cache \
+RUN apk update && apk upgrade && apk add --no-cache \
     bash \
     curl \
     docker-cli \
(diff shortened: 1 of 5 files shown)


moskyb (approved):

moskyb:
approved, though i wonder how much longer it'll make our builds take...

triarius:
> approved, though i wonder how much longer it'll make our builds take...

Yeah, if it becomes a problem, we can try and turn it off for feature branches. There will be some environment variable plumbing involved, but I think it should be possible.

moskyb:
> if it becomes a problem, we can try and turn it off for feature branches

seems like a good plan to me. i wonder if we need to be docker/binary building at all on non-main builds....

triarius:
> seems like a good plan to me. i wonder if we need to be docker/binary building at all on non-main builds....

So sometime a feature branch will be change something in the docker images, so you would want to build them there. So the problem reduces to that of monorepos.

PR #2414, opened 2023-10-06
Test clicommand config completeness

moskyb:
**This PR:** Adds some tests to the clicommand package that ensure that there's a one-to-one mapping between the command config structs and their `cli` struct tags, and the flags that we define on the `cli.Command`. Additionally, it adds a ratchet(? kinda) test that ensures that any new commands added will be added to the test above.

Additionally, it removes a CLI flag on the lock commands that was unused. Tests! Yay!


triarius (approved):
This makes me so happy! 💜

PR #2415, opened 2023-10-06
Add utility for generating signing and verification keys

moskyb:
This should make it much easier for users to generate signing and verification keys - from what i've found, there isn't much in the way of nice tooling if you don't build it yourself.


clicommand/util_keygen.go
```diff
@@ -0,0 +1,110 @@
...
+			Name:   "public-keyset-filename",
+			EnvVar: "BUILDKITE_AGENT_KEYGEN_PUBLIC_KEYSET_FILENAME",
+			Usage:  "The filename to write the public keyset to",
+		},
+		DebugFlag,
+		LogLevelFlag,
+	},
+	Action: func(c *cli.Context) {
+		ctx := context.Background()
+		_, cfg, l, _, done := setupLoggerAndConfig[KeygenConfig](ctx, c)
```
triarius, line 55:
```suggestion
		ctx, cfg, l, _, done := setupLoggerAndConfig[KeygenConfig](context.Backgroud(), c)
```

We want people to use the context returned by the `setupLoggerAndConfig` method, as that has the experiments in it. When we check if an experiment is enabled, we return `false` if the context had no mention of experiments: https://github.com/buildkite/agent/blob/f69df3866b3b34078df053b29bcd83d20e203c7e/internal/experiments/experiments.go#L123-L127

This means we could mistakenly believe an experiment is disabled if we check the wrong context. To prevent such bugs, we should not expose the original context at all until it has been populated with experiments.

clicommand/util_keygen.go
triarius, line 1: Just call this file `keygen.go`?

clicommand/util_keygen.go
```diff
@@ -0,0 +1,111 @@
...
+		cli.StringFlag{
+			Name:   "private-key-filename",
+			EnvVar: "BUILDKITE_AGENT_KEYGEN_PRIVATE_KEY_FILENAME",
+			Usage:  "The filename to write the private key to. Defaults to a name based on the key id in the current directory",
+		},
+		cli.StringFlag{
+			Name:   "public-keyset-filename",
+			EnvVar: "BUILDKITE_AGENT_KEYGEN_PUBLIC_KEYSET_FILENAME",
+			Usage:  "The filename to write the public keyset to. Defaults to a name based on the key id in the current directory",
+		},
```
triarius, line 48: I think we need defaults for these. It will be a barrier to first use otherwise.

clicommand/util_keygen.go
```diff
@@ -0,0 +1,111 @@
...
+)
+
+type KeygenConfig struct {
+	Alg                  string `cli:"alg" validate:"required"`
+	KeyID                string `cli:"key-id" validate:"required"`
+	PrivateKeyFilename   string `cli:"private-key-filename" normalize:"filepath"`
+	PublicKeysetFilename string `cli:"public-keyset-filename" normalize:"filepath"`
+
+	LogLevel string `cli:"log-level"`
+	Debug    bool   `cli:"debug"`
```
triarius, line 22: I think the usual set of
```
		// Global flags
		NoColorFlag,
		DebugFlag,
		LogLevelFlag,
		ExperimentsFlag,
		ProfileFlag,
```
should be copy pasted here. I know I said I have a way of using type constraints to eliminate this boilerplate. I haven't gotten around to implement it. Feel free to take it off my hands.

triarius: I know most of them don't make sense yet, but it will be easier to remove the boilerplate if every command has the same boilerplate.

clicommand/util_keygen.go
triarius: This is just a particular example. In general, I think users will want to use the returned context in most situations.

clicommand/util_keygen.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+var KeygenCommand = cli.Command{
+	Name:  "keygen",
+	Usage: "Generate a new key pair, used to sign and verify jobs",
```
triarius, line 27: Maybe a link to some docs on what the key and keyset formats are would be useful.

clicommand/util_keygen.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+
+		priv, pub, err := jwkutil.NewKeyPair(cfg.KeyID, sigAlg)
+		if err != nil {
+			l.Fatal("Failed to generate key pair: %v", err)
+		}
+
+		if cfg.PrivateKeyFilename == "" {
+			cfg.PrivateKeyFilename = fmt.Sprintf("./%s-%s-private.json", cfg.Alg, cfg.KeyID)
+		}
```
triarius, line 71: We should check that the private key file does not exist to prevent users overwriting a previously generated private key, and maybe have a flag to disable the check.

clicommand/util_keygen.go
```diff
@@ -0,0 +1,111 @@
...
+
+		if cfg.PublicKeysetFilename == "" {
+			cfg.PublicKeysetFilename = fmt.Sprintf("./%s-%s-public.json", cfg.Alg, cfg.KeyID)
+		}
+
+		privFile, err := os.Create(cfg.PrivateKeyFilename)
```
triarius, line 77: We should create the file with `0o600` permissions, I think you need `os.OpenFile` to do that.

triarius:
Looks great! Just a few suggestions. I'll try to use it before giving another review.

clicommand/util_keygen.go
moskyb: ah, it's following the established precedent of calling files `$SUPERCOMMAND_$SUBCOMMAND.go`, ie `meta_data_get.go`, `env_dump.go` etc

triarius: Oh, then the config name should change as well?

I also think `buildkite-agent util kegen` is not great. What's wrong with `buildkite-agent keygen`?

triarius: Because by this definition:
`"Utility commands, intended for users and operators of the agent to run directly"`, everything except `start` and `bootstrap` is a utility?

moskyb: where i'm coming from here is that none of the commands on the agent - save maybe `start` - are ever run directly by users on their machines. artifact uploads, metadata, annotations etc will only work if they're run by a buildkite agent, because they all rely on certain stuff (tokens, job IDs etc). the point of the `util` namespace (or `tools`, or something) is to separate the `keygen` command, as well as future operator tools, out from the rest of the agent codebase, which largely can't be easily run by end users

clicommand/util_keygen.go
moskyb: ooohhhhh yep, definitely need to add a usage string to this, that information can go there.

clicommand/util_keygen.go
moskyb: we had a chat about this over zoom. the TL;DR:
- "helpers for operators" isn't something that we think _should_ come in the agent codebase, but there isn't really a better place for it right now
- "`util` is the name you use when you can't think of anything else" - @triarius. we're going to rename the parent command to `tool`, which is slightly better

triarius (approved):
🎉

clicommand/tool_keygen.go
```diff
@@ -0,0 +1,134 @@
...
+
+    buildkite-agent tool keygen [options...]
+
+Description:
+
+This (experimental!) command generates a new JWS key pair, used for signing and verifying jobs in Buildkite.
+The key pair is written to two files, a private keyset and a public keyset. The private keyset should be used
+as for signing, and the public for verification. The keysets are written in JWKS format.
+
+For information about signing jobs in buildkite, see: <PUT A LINK HERE YA DINGO>
```
moskyb, line 41: @triarius how do you feel about a <PUT A LINK HERE YA DINGO> going into main? we don't have docs sorted for signed pipelines just yet...

moskyb: maybe i'll update it to say "`<PUT A DOC LINK HERE WHEN THERE IS ONE>`"

moskyb: maybe i'll update it to say "`<PUT A DOC LINK HERE WHEN THERE IS ONE>`"

triarius: Delete the line an put a TODO comment above the varaible?

moskyb: much more sensible.

PR #2420, opened 2023-10-10
Signing config cleanup

moskyb:
Closes: PDP-1764

**This PR:** Makes three changes to the way pipeline signing is configured:
- It removes the `--signing-algorithm` option, and the way that it would inject the algorithm into keys. As such, the `alg` key is now mandatory in signing JWKs
- I~t makes `--signing-key-id` optional. If one is provided, then the agent will use the signing key referred to by that ID, if it's present. If no signing key id is provided, the agent will use the first key in the JWKS. When verifyinh, the jwx library will just try every key in the verification key set~ This has been reverted, key IDs are now mandatory again
- It collapses the two flags `--job-verification-no-signature-behavior` and `--job-verification-invalid-signature-behavior` into one combined flag, `--job-verification-failure-behaviour`. This is simpler and a lot less wordy.


clicommand/agent_start.go
```diff
@@ -1106,6 +1099,37 @@ var AgentStartCommand = cli.Command{
...
+		return nil, fmt.Errorf("Job %s keyset is empty", keysetType)
+	}
+
+	iter := jwks.Keys(ctx)
+	for iter.Next(ctx) {
+		keyI := iter.Pair().Value
+		key, ok := keyI.(jwk.Key)
+		if !ok {
+			return nil, fmt.Errorf("Job %s keyset contains a non-key at index %d", keysetType, iter.Pair().Index)
+		}
```
moskyb, line 1108: it's deeply odd to me that the iterator for the jwk.Set type (which can only contain jwk.Keys) returns an int index and an any key, but ¯\\\_(ツ)_/¯

clicommand/pipeline_upload.go
```diff
@@ -422,56 +416,13 @@ func loadSigningKey(cfg PipelineUploadConfig) (jwk.Key, error) {
...
-		return fmt.Errorf("algorithm %s (%s) cannot be used for signing/verification", alg, from)
-	}
-
-	err := key.Set(jwk.AlgorithmKey, signingAlg)
-	if err != nil {
-		return fmt.Errorf("setting signing algorithm on key: %v", err)
-	}
-
-	return nil
-}
```
moskyb: goodbye, ugly terrible function, you have served your purpose

internal/pipeline/sign.go
```diff
@@ -111,7 +111,7 @@ func (s *Signature) Verify(env map[string]string, sf SignedFielder, keySet jwk.S
 	}
 
 	_, err = jws.Verify([]byte(s.Value),
-		jws.WithKeySet(keySet),
+		jws.WithKeySet(keySet, jws.WithRequireKid(false)),
```
moskyb, line 114: without this change, all keys would have to have key ids. while the [keygen utility](https://github.com/buildkite/agent/pull/2415) we wrote will always output keys with key ids, we don't enforce that people use our tooling, and should let users not use key ids if they don't want to.

moskyb: hmm, one downside of this change is that it makes the error messaging for "you don't have the right keys for verification" significantly worse. When we required key IDs, the message was something like
```
WARNING: Job verification failed: invalid signature: key provider 0 failed: failed to find key with key ID "my-cool-key" in key set
```

which tells you a fair bit about the problem.

with `jws.WithRequireKid(false)`, the error message is:
```
⚠️ WARNING: Job verification failed: invalid signature: could not verify message using any of the signatures or keys
```

which isn't too too much worse, but it's the same as if the job couldn't be verified (ie someone did crimes on the job).

i'm in two minds as to whether we should require/not require key IDs anyway, would love to hear some thoughts on this

triarius: I'm happy to require kids because:
1. You're going to need them if you want to rotate keys and maintain your sanity
2. Our tool keygen requires them

So if we enforce that the key always has an ID for both signing and verification, users who use the 'blessed path' aren't going to fail.

One thing we could do is generate a kid if user's can't be bothered to supply one. I should have thought of that when reviewing this: https://github.com/buildkite/agent/pull/2415

If we do that then user's don't have think about it

moskyb: 👀 https://github.com/buildkite/agent/pull/2422

moskyb:
i've just rebased away the commit that makes key IDs optional

agent/verify_job.go
```diff
@@ -40,7 +40,7 @@ func (r *JobRunner) verificationFailureLogs(err error, behavior string) {
 	r.logStreamer.Process([]byte(fmt.Sprintf("⚠️ %s: Job verification failed: %s\n", label, err.Error())))
 
 	if behavior == VerificationBehaviourWarn {
-		r.logger.Warn("Job will be run without verification - this is not recommended. You can change this behavior with the `job-verification-no-signature-behavior` agent configuration option.")
+		r.logger.Warn("Job will be run without verification - this is not recommended. You can change this behavior with the `job-verification-failure-behavior` agent configuration option.")
```
DrJosh9000, line 43: Just thinking about this (existing) wording...

```suggestion
		r.logger.Warn("Job will be run whether or not it can be verified - this is not recommended. You can change this behavior with the `job-verification-failure-behavior` agent configuration option.")
```

DrJosh9000 (approved):
Awesome!

agent/verify_job.go
moskyb: ooooh, i like it

triarius (approved):

PR #2422, opened 2023-10-10
Randomly generate key ID if none is provided

moskyb:
**This PR:** Changes the keygen tool such that inputting a key id is no longer required. If none is passed in, we'll generate a two-word pet name for that key using [dustinkirkland/golang-petname](https://github.com/dustinkirkland/golang-petname)


go.mod
```diff
@@ -51,6 +51,8 @@ require (
 	gotest.tools/v3 v3.5.1
 )
 
+require github.com/dustinkirkland/golang-petname v0.0.0-20231002161417-6a283f1aaaf2
+
```
moskyb, line 55: why is this in its own section? no idea, but i ran `go mod tidy` and it didn't change, so 🤷

triarius (approved):
🎉

go.mod
triarius: We need to contribute a cannoicaliser for go.mod

PR #2425, opened 2023-10-10
Make interpolation helpers transform-agnostic

DrJosh9000:
The current approach to interpolation assumes `${VAR}`-style interpolation, but we also want to do basically the same thing with the template-like `{{matrix}}` tokens. 

I considered using a func type like in #2395, but if we want to exclude matrix from interpolating into itself, it might make sense to do that by introspecting the type of the string transformer.


triarius (approved):
LGTM

PR #2428, opened 2023-10-12
Separate step/pipeline env vars for job validation

DrJosh9000:
Matrix interpolation is going to thwart the existing `env::` scheme. Splitting pipeline and step env checks should get ahead of the problem.

**The problem:** in the existing scheme, all vars (step and pipeline) are combined into the one "namespace" of signed fields. This happens "pre-matrix". Pipeline env vars should make it through unscathed, but step env vars will have matrix interpolation applied in the backend. So any step env var (~~not overridden by the pipeline env~~ lol turns out step vars should take precedence over pipeline vars) containing a matrix token is going to have a different value in the job env than it had when it was signed, and will fail to verify.

**The solution:** Preserve the step env as its own field, and keep it separate from pipeline env during signing.

**Also:** This switches the precedence of env vars to match https://buildkite.com/docs/tutorials/pipeline-upgrade#what-is-the-yaml-steps-editor-compatibility-issues and also removes interpolation of matrix tokens into env var names.

Step env vars cause pipeline vars with the same names to be excluded from `env::`, and hence from the signature, because at verify time the job env var will have the step var value, not the pipeline var value. If the pipeline var was still present under `env::`, then it would need to match the pipeline var value during signing, which the job won't have.

Interpolating matrix tokens into env var names would cause further problems: no pipeline env vars could be excluded from the signature, because matrix interpolation may or may not cause a corresponding step env var to exist. Thankfully, the backend doesn't interpolate matrix tokens in env var names. I think I was under the impression that it would do the same thing as env var interpolation.

### DIFF
--- a/agent/verify_job.go
+++ b/agent/verify_job.go
@@ -73,30 +73,74 @@ func (r *JobRunner) verifyJob(keySet jwk.Set) error {
 	// Note that each field needs a different consistency check:
 	//
 	// - command should match BUILDKITE_COMMAND exactly
-	// - env vars not listed in the signature as signed fields are allowed
+	// - env vars are complicated
 	// - plugins should match BUILDKITE_PLUGINS semantically
-	// - matrix interpolates into the others, and does not itself compare
+	// - matrix interpolates into the other fields, and does not itself compare
 	//   (not yet implemented).
 	//
 	// We can't check that the job is consistent with fields we don't know
 	// about yet, so these are rejected.
 	//
-	// More notes on `env::`:
-	// 1. Signature.Verify ensures every signed env var has the right value, and
-	// 2. step env can be overridden by the pipeline env, but each step only
-	//    knows about its own env. So the job env and step env can disagree
-	//    under normal circumstances.
-	// We still have to catch when a signature validates only because the step
-	// has an env var (not used to run the job), that is not present in the
-	// job env (is actually used).
+	// Notes on env vars:
+	// 1. Pipeline env values are included individually in the "env::" namespace
+	//    of fields, and step env vars are a single map under the "env" field.
+	// 2. When producing the job env, pipeline env overrides step env. This may
+	//    seem backwards but it is our documented behaviour:
+	//    https://buildkite.com/docs/pipelines/environment-variables#defining-your-own
+	// 3. The backend also adds env vars that can't be known in advance for
+	//    signing.
+	// 4. Step env vars can have matrix tokens (in both names and values).
+	// 5. Pipeline env is uploaded as a distinct map, but is not fully available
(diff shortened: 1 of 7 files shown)


agent/verify_job.go
```diff
@@ -99,6 +116,19 @@ func (r *JobRunner) verifyJob(keySet jwk.Set) error {
...
+		case "env":
+			// Everything in the step env (post-matrix interpolation) must be
+			// present in the job env, and have equal values.
+			for name, stepEnvValue := range step.Env {
+				jobEnvValue, has := r.conf.Job.Env[name]
+				if !has {
+					return newInvalidSignatureError(fmt.Errorf("job %q was signed with signature %q, but step.env defines %s which is missing from the job environment", r.conf.Job.ID, step.Signature.Value, name))
+				}
+				if jobEnvValue != stepEnvValue {
+					return newInvalidSignatureError(fmt.Errorf("job %q was signed with signature %q, but the value of %s (%q) does not match the value of step.env[%s] (%q)", r.conf.Job.ID, step.Signature.Value, name, jobEnvValue, name, stepEnvValue))
```
moskyb, line 128: re the `job %q was signed with signature %q` prefix on both of these errors: do we need to output the value of the signature? i realise this is based on precedent that i started...

also, re: our chat with ashwin the other day, how much information do we want to/should we disclose when the signatures match, but something in the job has been messed with? would it be worth just saying something to the effect off "step signatures matched, but the job was different from what we expected"?

internal/pipeline/interpolate.go
```diff
@@ -102,8 +102,22 @@ func interpolateSlice[E any, S ~[]E](tf stringTransformer, s S) error {
...
+		m[k] = intv
+	}
+	return nil
+}
+
+// interpolateMap applies interpolateAny over both keys and values of any type
+// of map. The map is altered in-place.
 func interpolateMap[K comparable, V any, M ~map[K]V](tf stringTransformer, m M) error {
 	for k, v := range m {
 		// We interpolate both keys and values.
```
moskyb, line 123: this is now only used for variable interpolation, right?

moskyb (approved):
LGTM! a couple of questions and notes, but this is looking really nice

internal/pipeline/interpolate.go
DrJosh9000: Deep matrix interpolation still applies to plugins, AFAICT

agent/verify_job.go
DrJosh9000: How about I merge this, then we can follow up with a PR comprehensively changing these errors?

moskyb: a followup sounds like a good idea

PR #2429, opened 2023-10-12
Enable matrixes for step signing

DrJosh9000:
- [x] Add MatrixPermutation to job response
- [x] Parse MatrixPermutation in a way compatible with backend
- [x] Use matrix interpolator on step after verifying signature
- [x] Disable checks for matrixes that disable signing and verifying
- [x] Integration test

### DIFF
--- a/internal/pipeline/step_command_matrix.go
+++ b/internal/pipeline/step_command_matrix.go
@@ -35,6 +35,8 @@ var (
 		ordered.Unmarshaler
 		selfInterpolater
 	} = (*MatrixScalars)(nil)
+
+	_ json.Unmarshaler = (*MatrixPermutation)(nil)
 )
 
 var (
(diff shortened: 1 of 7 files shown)


internal/pipeline/step_command_matrix.go
```diff
@@ -224,11 +226,38 @@ func (m *Matrix) validatePermutation(p MatrixPermutation) error {
...
+				Dimension: dim,
+				Value:     val,
+			})
+
+		default:
+			return fmt.Errorf("unsupported item type %T for key %q; want one of bool, int, or string", val, dim)
+		}
+	}
+	return nil
+}
```
moskyb, line 254: could we make the backend format matrix permutations in the way that we have them on the agent and avoid the extra unmarshalling logic here?

moskyb (approved):
awesome work! it's a testament to the work that we (but mostly you) have done that this PR is as simple and easy to read as it is.

internal/pipeline/step_command_matrix.go
DrJosh9000: That's reasonable. The backend already has to do similar work to massage the matrix back into the expected form. 

It wouldn't be enough to eliminate the method, because encoding/json will pick float64 for unmarshalling integers into any.

I was thinking, for simplicity, we could give up on preserving the type, and format ints and bools into strings as early as possible?

moskyb: i quite like that - i'm trying to think of a situation where munging stuff that's eventually going into a CLI into strings could be a problem, but i'm not coming up with much? that said, i'm sure there's something... 

it'd certainly be nice if the `Value` field of the `SelectedDimension` type was string.

DrJosh9000: I think I'll merge this, then do a follow up.

PR #2430, opened 2023-10-16
Revert "Upgrade pre-installed packages in docker images" and Pin docker images by digest

triarius:
In #2410 I responded to a report that the system packages in our docker images were out of date by upgrading the system packages at build time. However, this has a notable disadvantage - the version in an image is now determined both by the commit of the build and the time that the build ran. So we have lost some reproducibility. So in this PR, we revert that.

However, we can do a bit better than the situation before. Previously, we only pinned the tags of the base image, for example we pinned the `ubuntu` image to the tag `22.04`. Dependabot would only update the image when a newer tag emerged, which is very infrequent in the case of `ubuntu`. However, maintainers of the base image regularly push newer images to the tag that we have pinned. This mutates the "digest" of the image, and dependabot is [capable of detecting updates to the digest](https://github.com/buildkite/agent/pull/2179). So in this PR, we pin all docker base images by digest.

Notably, this is MORE reproducible than the original situation, as it now two builds at the same commit will always use the same base image.

Reverts buildkite/agent#2410

### DIFF
--- a/.buildkite/Dockerfile-compile
+++ b/.buildkite/Dockerfile-compile
@@ -1,4 +1,4 @@
-FROM golang:1.20.10@sha256:098d628490c97d4419ed44a23d893f37b764f3bea06e0827183e8af4120e19be
+FROM public.ecr.aws/docker/library/golang:1.20.10@sha256:098d628490c97d4419ed44a23d893f37b764f3bea06e0827183e8af4120e19be
 COPY build/ssh.conf /etc/ssh/ssh_config.d/
 RUN go install github.com/google/go-licenses@latest
 
--- a/.buildkite/Dockerfile-github-release
+++ b/.buildkite/Dockerfile-github-release
@@ -1,4 +1,4 @@
-FROM public.ecr.aws/docker/library/alpine:3.18.4
+FROM public.ecr.aws/docker/library/alpine:3.18.4@sha256:eece025e432126ce23f223450a0326fbebde39cdf496a85d8c016293fc851978
 
 # We need buildkite-agent to download artifacts, and zip for Windows zipping
 RUN apk --no-cache add bash zip curl \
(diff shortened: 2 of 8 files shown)


.buildkite/Dockerfile-compile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.20.10@sha256:098d628490c97d4419ed44a23d893f37b764f3bea06e0827183e8af4120e19be
+FROM public.ecr.aws/docker/library/golang:1.20.10@sha256:098d628490c97d4419ed44a23d893f37b764f3bea06e0827183e8af4120e19be
```
DrJosh9000, line 1: Good catch here

DrJosh9000 (approved):
LGTM 🥦

I gather that pinning versions of the packages installed with `apk`, `apt`, etc and then having Dependabot update those is not possible, otherwise we could do the same and close another hole in reproducibility.

triarius:
> LGTM 🥦
> 
> I gather that pinning versions of the packages installed with `apk`, `apt`, etc and then having Dependabot update those is not possible, otherwise we could do the same and close another hole in reproducibility.

Yeah, the story is a bit murky here. As noted in: https://github.com/dependabot/dependabot-core/issues/2129#issuecomment-1112715022, if we pinned the versions of apk/apt packages, the build will break after the package is upgraded in the repo as these repos typically only hold one version of each package. There may be some ways around this with caching docker layers, but that won't work consistently either with the way we currently keep the docker build cache local to each instance the builder agents run on.

In theory, dependabot could detect the update and create a PR that builds successfully. But everyone's branches will be broken until they incorporate the changes in the dependabot PR. And dependabot will never create the new PR immediately, there is always going to be some lag between when packages are updated in the repos and dependabot runs. I think this will be a bad development experience.

PR #2436, opened 2023-10-16
Update JCS to v1.0.1

moskyb:
This fixes the issue with bare json nulls

### DIFF
--- a/agent/verify_job.go
+++ b/agent/verify_job.go
@@ -105,7 +105,7 @@ func (r *JobRunner) verifyJob(keySet jwk.Set) error {
 			// jcs.Transform chokes on "" and "null", and json.Marshal encodes
 			// nil slice as "null", but zero-length slice as "[]".
 			emptyStepPlugins := len(step.Plugins) == 0
-			emptyJobPlugins := (jobPluginsJSON == "" || jobPluginsJSON == "null" || jobPluginsJSON == "[]")
+			emptyJobPlugins := jobPluginsJSON == ""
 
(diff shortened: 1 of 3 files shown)


agent/verify_job.go
```diff
@@ -105,7 +105,7 @@ func (r *JobRunner) verifyJob(keySet jwk.Set) error {
 			// jcs.Transform chokes on "" and "null", and json.Marshal encodes
```
DrJosh9000, line 105: The comment should be updated too

agent/verify_job.go
```diff
@@ -105,7 +105,7 @@ func (r *JobRunner) verifyJob(keySet jwk.Set) error {
...
 			// nil slice as "null", but zero-length slice as "[]".
 			emptyStepPlugins := len(step.Plugins) == 0
-			emptyJobPlugins := (jobPluginsJSON == "" || jobPluginsJSON == "null" || jobPluginsJSON == "[]")
```
DrJosh9000, line 108: Hmm... I don't know if this can be removed, even with the updated JCS lib. If `jobPluginsJSON == "null"`, but `step.Plugins` was an empty slice, then `emptyStepPlugins` would be true but `emptyJobPlugins` would be false, and then after marshalling and canonicalisation, would compare `null` against `[]`.

But maybe it doesn't matter? If there are no plugins then the backend should not even give us a `BUILDKITE_PLUGINS` env var.

DrJosh9000 (approved):

moskyb:
fixed by https://github.com/buildkite/agent/pull/2435

PR #2438, opened 2023-10-17
Add packagecloud upload step for rpms

sj26:
I'd like to try uploading new agent packages to packagecloud repositories. I'd also like to get it working using OIDC tokens, so we can avoid storing or transmitting long-lived secrets during the exchange.

This introduces a step which uploads new rpms to [a new agent-experimental repository](https://packagecloud.io/buildkite/agent-experimental). It's set to soft fail because I expect this to fail at first, but I'll be pushing through [the OIDC support](https://github.com/computology/packagecloud.io/pull/1815) to make this work.

Fixes PKG-5934.

### DIFF
--- a/.buildkite/pipeline.release-experimental.yml
+++ b/.buildkite/pipeline.release-experimental.yml
@@ -32,6 +32,21 @@ steps:
           volumes:
             - "/yum.buildkite.com"
 
+  - name: ":redhat: Publish Edge RPM Package to Packagecloud"
+    command: ".buildkite/steps/publish-rpm-packagecloud.sh"
+    env:
+      REPOSITORY: "buildkite/agent-experimental"
+      DISTRO_VERSION: rpm_any/rpm_any
+    agents:
+      queue: "deploy"
+    plugins:
+      - docker#v5.8.0:
+          image: "buildkite/agent:3.55.0-ubuntu"
+          entrypoint: bash
+          propagate-environment: true
+          mount-buildkite-agent: true
+    soft_fail: true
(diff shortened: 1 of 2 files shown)


tommeier (approved):
<its_happening.gif>

.buildkite/pipeline.release-experimental.yml
```diff
@@ -32,6 +32,21 @@ steps:
...
+      REPOSITORY: "buildkite/agent-experimental"
+      DISTRO_VERSION: rpm_any/rpm_any
+    agents:
+      queue: "deploy"
+    plugins:
+      - docker#v5.8.0:
+          image: "buildkite/agent:3.55.0-ubuntu"
+          entrypoint: bash
+          propagate-environment: true
+          mount-buildkite-agent: true
```
moskyb, line 47: could this just be a `ruby` container with `mount-buildkite-agent` turned on?

moskyb: or just an ubuntu container (which have ruby installed by default i guess?)

sj26: Maybe! I've picked the same as the other step because it's already known to work, and because it's likely to already be in cache. Feel free to modify it, if you modify the other steps :pray:

sj26: (Eventually I'd like us to have a Golang single-binary CLI for packages, and to be able to distribute that as part of a binary plugin instead of using docker!)

PR #2439, opened 2023-10-18
Add debug logs in verifyJob

DrJosh9000:
Spare a thought for the poor saps that actually _do_ need to debug the job verification logic.


moskyb (approved):
good call sending it to the agent logs rather than the job logs 👍

PR #2440, opened 2023-10-18
All scalars in matrixes are now strings

DrJosh9000:
This PR does two things:

1. (Fixes a verification bug) All scalars inside `Matrix` are stored as strings. During parsing they are all `fmt.Sprint`-ed. `MatrixScalars` is gone. `MatrixSetup` is now `map[string][]string` and `MatrixAdjustmentWith` is now `map[string]string`. This should fix verification when non-string scalars are used, because the backend at some point stringifies them all, before they are reflected back in the job payload, meaning that the JSONified `Matrix`es are different during verification, manifesting as a signature verification failure. Stringifying everything in advance should be no issue, because there is no difference between the results of interpolating `47` and `"47"` into a matrix token.
2. (Willing to go back on this) `MatrixPermutation` is now `map[string]string`, removing `SelectedDimension`. The alternative is to make the backend output be an array, but using a map removes one of the error cases in permutation validation (dupes) and also lets us remove `UnmarshalJSON` (without altering the current backend output - either way it's a PR here or there). But I didn't mind `MatrixPermutation` so much really.

### DIFF
--- a/agent/integration/job_verification_integration_test.go
+++ b/agent/integration/job_verification_integration_test.go
@@ -170,8 +170,8 @@ var (
 			"MESSAGE":           "hello world",
 		},
 		MatrixPermutation: pipeline.MatrixPermutation{
-			{Dimension: "greeting", Value: "hello"},
-			{Dimension: "object", Value: "world"},
+			"greeting": "hello",
+			"object":   "world",
 		},
 	}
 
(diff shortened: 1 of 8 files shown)


moskyb (approved):
LGTM on both fronts. 

i think there was originally a reason for `MatrixPermutation` being a `[]SelectedDimension`, but i can't remember what it was, and having it be a `map[string]string` certainly makes things easier to read

moskyb:
i do wanna call out the assumption here:

> Stringifying everything in advance should be no issue, because there is no difference between the results of interpolating 47 and "47" into a matrix token.

... that we know of. i think this is definitely fine for now though.

PR #2453, opened 2023-10-24
experiment: pty-raw avoids LF→CRLF mapping by setting PTY to raw mode

### LINKED ISSUE #736: Downloaded output file has extra line breaks
I can see extra CR invisible characters in the downloaded output files, the line endings become CRCRLF instead of just CRLF, and my browser and editors display an extra line break between each line as a result. For example:

https://builds.ninenines.eu/logs/cowlib/141/centos/output.txt

I'm not sure it's the agent's fault but considering the extra CRs occur only on lines of outputs from the configured commands, and not on the lines from the agent itself, I think it's likely to be something the agent is doing.

If it's inserted by the agent on purpose then an option to disable this behavior would be nice to have.

pda:
The command inside a build job is allocated and attached to a [Pseudoterminal](https://en.wikipedia.org/wiki/Pseudoterminal) so that programs detect that their output will be interpreted by a terminal emulator, and so they default to emitting colours, cursor movement, and other escape sequences.

This pull request sets that PTY into raw mode, disabling any default processing of the input/output streams, most specifically the [`ONLCR` flag](https://man.archlinux.org/man/termios.h.0p#ONLCR): “Map NL to CR-NL on output”.

Otherwise the PTY driver in the kernel will map `LF` (`\n`) to `CR` `LF` (`\r\n`), and possibly do other unwanted processing of the output stream. This is particularly vexing when programs intentionally (misguidedly?) emit `CR LF` newlines which get mapped to `CR CR LF`.

This is behind an experiment because it's _possible_ that by entering raw mode we're turning off some default PTY input or output processing that some programs actually rely on in some situations.

- Fixes #736

### DIFF
--- a/EXPERIMENTS.md
+++ b/EXPERIMENTS.md
@@ -117,3 +117,15 @@ The new glob library should resolve a few issues experienced with the old librar
 The new library should handle all syntax supported by the old library, but because of the chance of incompatibilities and bugs, we're providing it via experiment only for now.
 
 **Status:** Since using the old library causes problems, we hope to promote this to be the default soon™️.
+
+### `pty-raw`
+
+Set PTY to raw mode, to avoid mapping LF (\n) to CR,LF (\r\n) in job command output.
+These extra newline characters are normally not noticed, but can make raw logs appear double-spaced
+in some circumstances.
+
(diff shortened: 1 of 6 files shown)


DrJosh9000 (approved):
I have one comment, otherwise LGTM

PR #2454, opened 2023-10-25
Log warning about short vars once

DrJosh9000:
As opposed to a log line per short variable.

### DIFF
--- a/internal/redact/redact.go
+++ b/internal/redact/redact.go
@@ -3,8 +3,11 @@ package redact
 
 import (
 	"path"
+	"sort"
+	"strings"
 
 	"github.com/buildkite/agent/v3/internal/job/shell"
+	"golang.org/x/exp/maps"
 )
 
 // LengthMin is the shortest string length that will be considered a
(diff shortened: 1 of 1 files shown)


internal/redact/redact.go
```diff
@@ -67,5 +71,11 @@ func Vars(logger shell.Logger, patterns []string, environment map[string]string)
 		}
 	}
 
+	if len(shortVars) > 0 {
+		// TODO: Use stdlib maps when it gets a Keys function (Go 1.22?)
+		vars := maps.Keys(shortVars)
+		sort.Strings(vars)
+		logger.Warningf("Some variables have values below minimum length (%d bytes) and will not be redacted: %s", LengthMin, strings.Join(vars, ","))
```
moskyb, line 78:
```suggestion
		logger.Warningf("Some variables have values below minimum length (%d bytes) and will not be redacted: %s", LengthMin, strings.Join(vars, ", "))
```

moskyb (approved):
LGTM! what a turnaround time 🥳

PR #2456, opened 2023-10-25
Parallelise artifact collection

DrJosh9000:
It's possible to configure `Paths` with a lot of separate globs. It's not unreasonable that each of these could be of the form `**/something`, and each one requires walking the entire directory tree. `go-zglob` is not terribly slow because it uses a parallel walker under the hood (`fastwalk`). While I think about how to speed up `zzglob`, this is a reasonable first step.

This also fixes a bug where `wd` is replaced with `"/"` (or `filepath.VolumeName(absolutePath) + "/"` on Windows) on the first glob path in the list that is is absolute, changing `wd` for later glob paths that might not be absolute.

### DIFF
--- a/agent/artifact_uploader.go
+++ b/agent/artifact_uploader.go
@@ -116,7 +116,6 @@ func (a *ArtifactUploader) Collect(ctx context.Context) ([]*api.Artifact, error)
 		return nil, fmt.Errorf("getting working directory: %w", err)
 	}
 
-	workCh := make(chan string)
 	var wg sync.WaitGroup
 
 	ac := &artifactCollector{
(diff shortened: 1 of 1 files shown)


inez:
Very interested in giving it a try!

agent/artifact_uploader.go
```diff
@@ -110,87 +110,161 @@ func isDir(path string) bool {
...
 
+	ac := &artifactCollector{
+		ArtifactUploader: a,
+		wd:               wd,
+		seenPaths:        make(map[string]bool),
+	}
+
+	// Start N workers in a new context
+	wctx, cancel := context.WithCancelCause(ctx)
+	for i := 0; i < runtime.NumCPU(); i++ {
```
moskyb, line 130: is `runtime.NumCPU()` being used here as "an arbitrary smallish number to parallelise by, which will be roughly proportional to the size of the box we're running on"? that's a totally valid use case, i just wanna be sure i'm reading right

moskyb (approved):
LGTM. happy to see us getting more `chan`-y with it, for a go project, the agent has surprisingly few goroutines.

agent/artifact_uploader.go
DrJosh9000: Yes. It's not a great heuristic though. We could create N goroutines and let the runtime figure it out. The code would be simpler, and unlike a server, the amount of work is fixed ahead of time, so it's not unreasonable. Actually, that idea is growing on me...

PR #2457, opened 2023-10-25
Add some pipeline invariants to the signature and create a cli subcommand to sign a pipeline

triarius:
The pipeline invariants are:
- pipeline slug
- organization slug
- repository

An agent configured to verify signatures will now check that the pipeline slug, organization slug and repository in the accepted job definition are the same as the ones in the signature.

The cli subcommand
```
buildkite-agnet tool sign --jwks-file-path <keyfile> --signing-key-id <key-id> --pipeline-slug <pipeline-slug> --organization-slug <organization-slug> --repo <repository> <pipeline-yaml-file>
```
can be used to sign the initial pipeline. Its output may be pasted into the Steps Editor in the Buildkite Web UI so that the initial pipeline may be verified as well.

## Implementation details
I created a new struct `CommandStepWithInvariants` and moved the `SignedFielder` implementation from `CommandStep` to it. I went back and forth on a few ways to do this, but I think it's easier to reason about if the marshalling of fields to sign happens in one place.

### DIFF
--- a/clicommand/commands.go
+++ b/clicommand/commands.go
@@ -82,6 +82,7 @@ var BuildkiteAgentCommands = []cli.Command{
 		Usage: "Utility commands, intended for users and operators of the agent to run directly on their machines, and not as part of a Buildkite job",
 		Subcommands: []cli.Command{
 			KeygenCommand,
+			SignCommand,
 		},
 	},
 }
--- a/clicommand/commands.go
+++ b/clicommand/commands.go
@@ -82,7 +82,7 @@ var BuildkiteAgentCommands = []cli.Command{
 		Usage: "Utility commands, intended for users and operators of the agent to run directly on their machines, and not as part of a Buildkite job",
 		Subcommands: []cli.Command{
 			KeygenCommand,
-			SignCommand,
+			ToolSignCommand,
 		},
 	},
 }
--- a/clicommand/commands.go
+++ b/clicommand/commands.go
@@ -81,7 +81,7 @@ var BuildkiteAgentCommands = []cli.Command{
 		Name:  "tool",
 		Usage: "Utility commands, intended for users and operators of the agent to run directly on their machines, and not as part of a Buildkite job",
 		Subcommands: []cli.Command{
-			KeygenCommand,
+			ToolKeygenCommand,
 			ToolSignCommand,
 		},
 	},
--- a/clicommand/pipeline_upload.go
+++ b/clicommand/pipeline_upload.go
@@ -72,6 +72,7 @@ type PipelineUploadConfig struct {
 
 	JWKSFilePath string `cli:"jwks-file-path"`
 	SigningKeyID string `cli:"signing-key-id"`
+	ExpireAfter  string `cli:"expire-after"`
 
 	// Global flags
 	Debug       bool     `cli:"debug"`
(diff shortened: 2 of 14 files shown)


clicommand/commands.go
```diff
@@ -82,6 +82,7 @@ var BuildkiteAgentCommands = []cli.Command{
 		Usage: "Utility commands, intended for users and operators of the agent to run directly on their machines, and not as part of a Buildkite job",
 		Subcommands: []cli.Command{
 			KeygenCommand,
+			ToolSignCommand,
```
moskyb, line 85: should we either rename `KeyGen` to `ToolKeygen` or rename `ToolSign` to `Sign`?

clicommand/pipeline_upload.go
```diff
@@ -116,7 +116,7 @@ var PipelineUploadCommand = cli.Command{
 		},
 		cli.BoolFlag{
 			Name:   "no-interpolation",
-			Usage:  "Skip variable interpolation the pipeline when uploaded",
+			Usage:  "Skip variable interpolation into the pipeline prior to upload",
```
moskyb, line 119: 👍

internal/pipeline/steps.go
```diff
@@ -137,18 +137,23 @@ func stepByKeyInference(o *ordered.MapSA) (Step, error) {
 // sign adds signatures to each command step (and recursively to any command
 // steps that are within group steps. The steps are mutated directly, so an
 // error part-way through may leave some steps un-signed.
-func (s Steps) sign(env map[string]string, key jwk.Key) error {
+func (s Steps) sign(key jwk.Key, env map[string]string, pInv *PipelineInvariants) error {
```
moskyb, line 140: 👍

moskyb (changes requested):
looks good! there's one that i'm not sure about on re: using `environ`, but other than that this looks geat

clicommand/pipeline_upload.go
```diff
@@ -277,12 +277,18 @@ var PipelineUploadCommand = cli.Command{
 		if cfg.JWKSFilePath != "" {
 			l.Warn("Pipeline signing is experimental and the user interface might change! Also it might not work, it might sign the pipeline only partially, or it might eat your pet dog. You have been warned!")
 
-			key, err := loadSigningKey(cfg)
+			pInv := &pipeline.PipelineInvariants{
+				OrganizationSlug: environ.GetStringDefaultEmpty("BUILDKITE_ORGANIZATION_SLUG"),
+				PipelineSlug:     environ.GetStringDefaultEmpty("BUILDKITE_PIPELINE_SLUG"),
+				Repository:       environ.GetStringDefaultEmpty("BUILDKITE_REPO"),
```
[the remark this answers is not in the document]
triarius: Nope, it's me that's missing the context. Good catch!

clicommand/commands.go
triarius: I think you know which one I will choose 😂

internal/pipeline/steps.go
triarius: I feel like there's a another place that's like this. I'll see if I can put the key consistently as the first arg everywhere.

moskyb (approved):

PR #2461, opened 2023-10-30
Add log group headers and timestamps to job verification success and failure logs

triarius:
I had to also add a header to the Job API logs as otherwise they would appear to be part of the verification log group.
There are probably other such logs like this, we need to weed them out eventually.

I also simplified some of the logging, in particular, the signature is not printed as part of the log group header, but in a subsequent line.
As the log group will be collapsed unless there is a failure and the agent is configured to block, the signature will not be immediately visible.

Also, for the agent logs, I tried to add fields using the `logger.WithFields` method. These will be displayed in the logfmt format (or json if configured to do so). This will make the logs more structured.

Finally, I've removed the use of `(*JobRunner).logStreamer.Process` to send logs directly to the API as it bypasses various transformations that happen before (like timestamping). Users will still be able to call `logStreamer.Process`, but should use `(*JobRunner).jobLogs`, which exposes an `io.Writer` interface, instead.

#### Before
2023-10-30T15:10:53,358662464+11:00
#### After
2023-10-30T14:53:59,108749453+11:00
2023-10-30T14:54:06,655403708+11:00
#### Before
2023-10-30T15:13:58,323915394+11:00
#### After
2023-10-30T15:14:31,964587642+11:00

#### Before
2023-10-30T15:17:01,810886339+11:00
#### After
2023-10-30T15:16:09,340588835+11:00
2023-10-30T15:16:17,748683530+11:00

### DIFF
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -118,8 +118,8 @@ type JobRunner struct {
 	// How the JobRunner should respond when job verification fails (one of `block` or `warn`)
 	VerificationFailureBehavior string
 
-	// The logger to use
-	logger logger.Logger
+	// agentLogger is a agentLogger that outputs to the agent logs
+	agentLogger logger.Logger
 
 	// The APIClient that will be used when updating the job
 	apiClient APIClient
@@ -133,8 +133,11 @@ type JobRunner struct {
 	// The internal header time streamer
 	headerTimesStreamer *headerTimesStreamer
 
-	// The internal log streamer
+	// The internal log streamer. Don't write to this directly, use `jobLogs` instead
 	logStreamer *LogStreamer
+
+	// jobLogs is a logger that outputs to the job logs
+	jobLogs io.Writer
 
 	// If the job is being cancelled
 	cancelled bool
@@ -166,9 +169,9 @@ var _ jobRunner = (*JobRunner)(nil)
 // Initializes the job runner
 func NewJobRunner(ctx context.Context, l logger.Logger, apiClient APIClient, conf JobRunnerConfig) (jobRunner, error) {
 	r := &JobRunner{
-		logger:    l,
-		conf:      conf,
-		apiClient: apiClient,
+		agentLogger: l,
+		conf:        conf,
+		apiClient:   apiClient,
 	}
 
 	var err error
@@ -186,15 +189,15 @@ func NewJobRunner(ctx context.Context, l logger.Logger, apiClient APIClient, con
 	if r.conf.Job.Token != "" {
 		clientConf := r.apiClient.Config()
 		clientConf.Token = r.conf.Job.Token
-		r.apiClient = api.NewClient(r.logger, clientConf)
+		r.apiClient = api.NewClient(r.agentLogger, clientConf)
 	}
 
 	// Create our header times struct
-	r.headerTimesStreamer = newHeaderTimesStreamer(r.logger, r.onUploadHeaderTime)
+	r.headerTimesStreamer = newHeaderTimesStreamer(r.agentLogger, r.onUploadHeaderTime)
 
 	// The log streamer that will take the output chunks, and send them to
 	// the Buildkite Agent API
-	r.logStreamer = NewLogStreamer(r.logger, r.onUploadChunk, LogStreamerConfig{
+	r.logStreamer = NewLogStreamer(r.agentLogger, r.onUploadChunk, LogStreamerConfig{
 		Concurrency:       3,
 		MaxChunkSizeBytes: r.conf.Job.ChunksMaxSizeBytes,
 		MaxSizeBytes:      r.conf.Job.LogMaxSizeBytes,
@@ -213,7 +216,7 @@ func NewJobRunner(ctx context.Context, l logger.Logger, apiClient APIClient, con
 	if file, err := os.CreateTemp(tempDir, fmt.Sprintf("job-env-%s", r.conf.Job.ID)); err != nil {
 		return r, err
 	} else {
-		r.logger.Debug("[JobRunner] Created env file: %s", file.Name())
+		r.agentLogger.Debug("[JobRunner] Created env file: %s", file.Name())
 		r.envFile = file
 	}
 
@@ -237,7 +240,7 @@ func NewJobRunner(ctx context.Context, l logger.Logger, apiClient APIClient, con
 		jobLogDir := ""
 		if conf.AgentConfiguration.JobLogPath != "" {
 			jobLogDir = conf.AgentConfiguration.JobLogPath
-			r.logger.Debug("[JobRunner] Job Log Path: %s", jobLogDir)
+			r.agentLogger.Debug("[JobRunner] Job Log Path: %s", jobLogDir)
 		}
 		tmpFile, err = os.CreateTemp(jobLogDir, "buildkite_job_log")
 		if err != nil {
@@ -270,7 +273,7 @@ func NewJobRunner(ctx context.Context, l logger.Logger, apiClient APIClient, con
 
 		go func() {
 			// Use a scanner to process output line by line
-			err := process.NewScanner(r.logger).ScanLines(pr, func(line string) {
+			err := process.NewScanner(r.agentLogger).ScanLines(pr, func(line string) {
 				// Send to our header streamer and determine if it's a header
 				// or header expansion.
 				isHeaderOrExpansion := r.headerTimesStreamer.Scan(line)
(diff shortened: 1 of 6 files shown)


agent/run_job.go
```diff
@@ -148,6 +157,50 @@ func (r *JobRunner) Run(ctx context.Context) error {
...
+			"[%s] %s",
+			time.Now().UTC().Format(time.RFC3339),
+			fmt.Sprintf(s, args...),
+		))
+	default:
+		return []byte(fmt.Sprintf(s, args...))
+	}
+}
+
+func (r *JobRunner) verificationFailureLogs(behavior string, err error) {
```
moskyb, line 181: for readability in this func it might be nice to add something like 
```Go
jobLog := func(s string, args ...any) {
  l := fmt.Sprintf(s, args)
  r.logStreamer.Process([]byte(r.prependTimestampForLogs(l)))
}
```
and then use it like

```Go
jobLog("%s Job verification failed\n", prefix)
```

agent/run_job.go
```diff
@@ -148,6 +157,50 @@ func (r *JobRunner) Run(ctx context.Context) error {
...
+	l := r.logger.WithFields(logger.StringField("jobID", r.conf.Job.ID), logger.StringField("error", err.Error()))
+	prefix := "~~~ ⚠️"
```
moskyb, line 183: i think we should probably autoexpand warnings for a little bit more 🌶 - warning on verification failure it's not a state that pipelines should be in for long, and it's the kind of thing that, ideally, pipeline users are asking their infra teams about
```suggestion
	prefix := "+++ ⚠️"
```

agent/run_job.go
```diff
@@ -148,6 +157,50 @@ func (r *JobRunner) Run(ctx context.Context) error {
 	return nil
 }
 
+func (r *JobRunner) prependTimestampForLogs(s string, args ...any) []byte {
```
moskyb, line 162: right now this func both prepending stuff and also sprintf-ing its args... is there a reason that it has to happen all in one go? or can we sprintf prior to passing something to this func?

agent/run_job.go
```diff
@@ -148,6 +159,50 @@ func (r *JobRunner) Run(ctx context.Context) error {
...
+
+func (r *JobRunner) verificationFailureLogs(behavior string, err error) {
+	l := r.logger.WithFields(logger.StringField("jobID", r.conf.Job.ID), logger.StringField("error", err.Error()))
+	prefix := "~~~ ⚠️"
+	if behavior == VerificationBehaviourBlock {
+		prefix = "+++ ⛔"
+	}
+
+	l.Warn("Job verification failed")
+	r.logStreamer.Process([]byte(r.prependTimestampForLogs("%s Job verification failed\n", prefix)))
```
moskyb, line 189: as above, so below:

```suggestion
	r.logStreamer.Process([]byte(r.prependTimestampForLogs("%s Job signature verification failed\n", prefix)))
```

jobapi/server.go
```diff
@@ -57,7 +57,8 @@ func (s *Server) Start() error {
 		return fmt.Errorf("starting socket server: %w", err)
 	}
 
-	s.Logger.Commentf("Job API server listening on %s", s.SocketPath)
+	s.Logger.Printf("~~~ Job API")
+	s.Logger.Printf("Server listening on %s", s.SocketPath)
```
moskyb, line 61: 👩‍🍳💋

agent/run_job.go
```diff
@@ -89,22 +98,24 @@ func (r *JobRunner) Run(ctx context.Context) error {
...
 			exit.Status = -1
 			exit.SignalReason = SignalReasonSignatureRejected
 			return nil
 
 		default: // no error, all good, keep going
-			r.logger.Info("Successfully verified job %s with signature %s", job.ID, job.Step.Signature.Value)
-			r.logStreamer.Process([]byte(fmt.Sprintf("✅ Verified job with signature %s\n", job.Step.Signature.Value)))
+			l := r.logger.WithFields(logger.StringField("jobID", job.ID), logger.StringField("signature", job.Step.Signature.Value))
+			l.Info("Successfully verified job")
+			r.logStreamer.Process(r.prependTimestampForLogs("~~~ ✅ Job verified\n"))
```
moskyb, line 117: i know i'm bikeshedding my own wording here, but.... maybe 

```suggestion
			r.logStreamer.Process(r.prependTimestampForLogs("~~~ ✅ Job signature verified\n"))
```

tells people more about what's actually happening? i can see an inexperienced user saying "job verified to what"

moskyb (changes requested):

agent/run_job.go
```diff
@@ -148,6 +159,50 @@ func (r *JobRunner) Run(ctx context.Context) error {
...
 
+func (r *JobRunner) prependTimestampForLogs(s string, args ...any) []byte {
+	switch {
+	case r.conf.AgentConfiguration.ANSITimestamps:
+		return []byte(fmt.Sprintf(
+			"\x1b_bk;t=%d\x07%s",
+			time.Now().UnixNano()/int64(time.Millisecond),
+			fmt.Sprintf(s, args...),
+		))
+	case r.conf.AgentConfiguration.TimestampLines:
```
DrJosh9000, line 170: Can we have a single stream for processing job output + stuff the job runner wants to pretend is job output? Aside from duplicating the timestamping logic, this could mess with the header times streamer, and might also skip the local log output file?

triarius: I don't think we can have a single stream. The verification logs happen in the `agent start` process, while most of the other logs are streamed from the `agent bootstrap` process. So I think the best we can do is have two streams that are created from a common method.

We could have something in the local API so that the bootstrap process can stream logs over the local API to the remote agent API, but creating that for this task seems like a massive yak shave.

> pushing logs directly into r.logStreamer could make header times incorrect with non-ANSI timestamps
I don't really want to carry around the legacy of non-ANSI timestamps. Are you happy for me to remove support for these from this PR?

> might also skip the local log output file?
Yeah, they probably will. I'll look into something for this, and a way to not duplicate the timestamp logic, but if the answer to the question above is in the affirmative, then only about 1 line will be duplicated.

agent/run_job.go
triarius: It just seems easier to call one method rather than 2. If you want to sprintf prior to this method, then this interface does not prevent that. But if you don't want to, the suggested interface does.

But I should rename the method, its name only suggests the prepending, not the sprintfing.

agent/run_job.go
triarius: > I don't think we can have a single stream. The verification logs happen in the agent start process, while most of the other logs are streamed from the agent bootstrap process. So I think the best we can do is have two streams that are created from a common method.

Looks like I was wrong about this. I'll do the work to make a single stream.

agent/job_runner.go
```diff
@@ -131,9 +131,12 @@ type JobRunner struct {
 	// The internal header time streamer
 	headerTimesStreamer *headerTimesStreamer
 
-	// The internal log streamer
+	// The internal log streamer. Don't write to this directly, use `jobLogs` instead
 	logStreamer *LogStreamer
 
+	// jobLogs is a logger that outputs to the job logs
```
triarius, line 137:
```suggestion
	// jobLogs is a i.Writer that sends data to the job logs
```

agent/job_runner.go
```diff
@@ -131,9 +131,12 @@ type JobRunner struct {
 	// The internal header time streamer
 	headerTimesStreamer *headerTimesStreamer
 
-	// The internal log streamer
+	// The internal log streamer. Don't write to this directly, use `jobLogs` instead
 	logStreamer *LogStreamer
 
+	// jobLogs is an io.Writer that sends data to the job logs
+	jobLogs io.Writer
+
```
moskyb, line 139: 🥳

moskyb (approved):
awesome work!

agent/run_job.go
```diff
@@ -309,17 +351,8 @@ func (r *JobRunner) jobLogStreamer(ctx context.Context, wg *sync.WaitGroup) {
...
-		if err := r.logStreamer.Process(chunk); err != nil {
-			r.logger.Error("Could not stream the log output: %v", err)
-			// So far, the only error from logStreamer.Process is if the log has
-			// reached the limit.
-			// Since we're not writing any more, close the buffer, which will
-			// cause future Writes to return an error.
-			r.output.Close()
-		}
+		// Send the output of the process to the log streamer for processing
+		r.logStreamer.Process(r.output.ReadAndTruncate())
```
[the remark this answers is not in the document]
DrJosh9000: We can easily resurrect the error return once we start enforcing job log limits.

DrJosh9000 (approved):
☘️

PR #2465, opened 2023-10-31
Fix checkout of short commit hashes

triarius:
When `BUILDKITE_COMMIT` is set to a short commit hash, the agent will first attempt to just fetch that hash, but this will not work and git will return an error. Recently we changed the behaviour so that this error would delete the checkout directory, and try again, but it would still attempt to fetch the short hash. After a few attempts, the checkout would fail.

Instead, the old behaviour allowed the agent to fetch all heads and tags, and then checkout (instead of fetch) the short commit hash. This successfully sets the state of the build directory to the commit that was intended to be checked out.

In this PR, we use the `olefactor` to read the error message that git prints when the `fetch` fails and fall through from retrying the checkout to fetching all the heads and tags. I've also added a test that recreates this behaviour.

That this was broken was reported to us in #2437, but unfortunately the solution suggested there - to use `git rev-parse` to resolve the short commit hash to a long commit hash would not work because the checkout directory will generally not know about a new commit until a `git fetch` is run, and the issue here is that such a `git fetch` fails. It was also independent of BitBucket - this was broken for all providers when `BUILDKITE_COMMIT` was a short commit hash, and that's easily reproducible by creating a commit from the UI.

Closes: #2437

### DIFF
--- a/internal/job/integration/checkout_integration_test.go
+++ b/internal/job/integration/checkout_integration_test.go
@@ -321,6 +321,13 @@ func TestCheckingOutLocalGitProjectWithShortCommitHash(t *testing.T) {
 		fmt.Sprintf("BUILDKITE_COMMIT=%s", shortCommitHash),
 	}
 	tester.RunAndCheck(t, env...)
+
+	// Check state of the build directory
+	checkoutRepo := &gitRepository{Path: tester.CheckoutDir()}
+	checkoutRepoCommit, err := checkoutRepo.RevParse("HEAD")
+	assert.NilError(t, err)
+
+	assert.Equal(t, checkoutRepoCommit, commitHash)
 }
 
 func TestCheckoutErrorIsRetried(t *testing.T) {
--- a/internal/job/integration/checkout_integration_test.go
+++ b/internal/job/integration/checkout_integration_test.go
@@ -322,7 +322,7 @@ func TestCheckingOutLocalGitProjectWithShortCommitHash(t *testing.T) {
 	}
 	tester.RunAndCheck(t, env...)
 
-	// Check state of the build directory
+	// Check state of the checkout directory
 	checkoutRepo := &gitRepository{Path: tester.CheckoutDir()}
 	checkoutRepoCommit, err := checkoutRepo.RevParse("HEAD")
 	assert.NilError(t, err)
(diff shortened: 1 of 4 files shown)


internal/job/integration/checkout_integration_test.go
```diff
@@ -274,6 +275,61 @@ func TestCheckingOutShallowCloneOfLocalGitProject(t *testing.T) {
 	tester.RunAndCheck(t, env...)
 }
 
+func TestCheckingOutLocalGitProjectWithShortCommitHash(t *testing.T) {
+	t.Parallel()
+
+	tester, err := NewBootstrapTester(mainCtx)
+	assert.NilError(t, err)
```
DrJosh9000, line 282: (I don't mean to relitigate this here - but I am going to dump some state from my brain because I had to hit the docs for what `assert.NilError` did exactly)

I've been thinking about tests and assert libraries. We now have two (3? 4?) different assert libs in the agent:

* gotest.tools/v3/assert, where the functions call t.FailNow and log maybe too sparsely
* github.com/stretchr/testify/assert, whose functions use t.Errorf and don't terminate the test early, but dump some extra (sometimes redundant) info
* and we also import testify/require in places - the t.Fatalf variation on testify/assert
* bintest and the bootstrap tester 

https://go.dev/play/p/ff8l-kNBibK

Thoughts: 

* The usual argument against assert libraries in general is that it forces us to understand a small extra language with its own nuances in order to work with the assert lib. That we import two different packages in different places called "assert" that have different fail behaviours kinda demonstrates this.
* gotest.tools/v3/assert's functions by default don't include the name of the function that was called that returned the error and what its args were (if not too huge). Putting that info directly in the test log reduces cycle time. But it at least lets us provide that info: `assert.NilError(t, err, "NewBootstrapTester(mainCtx)")`)

DrJosh9000 (approved):
Congrats! You have won a green checkmark ✅ and a rant about tests

internal/job/integration/checkout_integration_test.go
triarius: > Putting that info directly in the test log reduces cycle time.

What happens when the name or arguments to the method changes? I'm pretty sure our code base is littered with examples where that's happened, and now someone reading the test output will be confused. For this reason, I prefer that we don't use this pattern. The file and line of the test failure is always printed, I prefer to start from there.

DrJosh9000: The info being logged is very close to the function call, so we should be better at keeping it updated.

mohsenamini128 (approved):

PR #2472, opened 2023-11-01
Use zzglob.MultiGlob, drop GlobPath field

DrJosh9000:
`MultiGlob` can take care of coalescing multiple patterns that have the same root into a single directory walk, as well as parallelising walks.

Since `GlobPath` isn't used in the backend, and it is unclear which pattern a file could have matched, drop `GlobPath` from the `Artifact`.

A small tweak is to grab the file size from the result of `io.Copy` (instead of doing another `Stat`), and also check its error.

### DIFF
--- a/agent/artifact_uploader.go
+++ b/agent/artifact_uploader.go
@@ -130,7 +130,7 @@ func (a *ArtifactUploader) Collect(ctx context.Context) ([]*api.Artifact, error)
 	// goroutine per file).
 	wctx, cancel := context.WithCancelCause(ctx)
 	var wg sync.WaitGroup
-	for i := 0; i < runtime.NumCPU(); i++ {
+	for i := 0; i < runtime.GOMAXPROCS(0); i++ {
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
--- a/agent/artifact_uploader.go
+++ b/agent/artifact_uploader.go
@@ -122,23 +122,30 @@ func (a *ArtifactUploader) Collect(ctx context.Context) ([]*api.Artifact, error)
 		seenPaths:        make(map[string]bool),
 	}
 
+	filesCh := make(chan string)
+
+	// Create a few workers to process files as they are found.
+	// Because a single glob could match many many files, a fixed number of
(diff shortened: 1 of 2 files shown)


agent/artifact_uploader.go
```diff
@@ -124,29 +124,36 @@ func (a *ArtifactUploader) Collect(ctx context.Context) ([]*api.Artifact, error)
...
-			continue
-		}
+	filesCh := make(chan string)
 
+	// Create a few workers to process files as they are found.
+	// Because a single glob could match many many files, a fixed number of
+	// workers will avoid slamming the runtime (as could happen with a
+	// goroutine per file).
+	wctx, cancel := context.WithCancelCause(ctx)
+	for i := 0; i < runtime.NumCPU(); i++ {
```
triarius, line 134: I think this can cause problems when the agent is run in a container. `runtime.NumCPU` may not respect the limits set by the container orchestration framework about the number of CPUs. At least if you do this
```suggestion
	for i := 0; i < runtime.GOMAXPROCS(0); i++ {
```
then this will be configurable and users can make their configuration match between the agent and the container orchestration framework.

To test this, I ran this program:

```go
package main

import (
	"fmt"
	"runtime"
)

func main() {
	fmt.Printf(
		"NumCPU=%d GOMAXPROCS=%d\n",
		runtime.NumCPU(),
		runtime.GOMAXPROCS(0),
	)
}
```
in docker:

```shell
❯ docker run --rm -it -v "$PWD:/app" -w /app --env GOMAXPROCS=1 --cpus=1 golang:1.21 go run .
NumCPU=20 GOMAXPROCS=1
❯ docker run --rm -it -v "$PWD:/app" -w /app --cpus=1 golang:1.21 go run .
NumCPU=20 GOMAXPROCS=20
```

There's also this: https://github.com/uber-go/automaxprocs, which I don't think we need to use, though one day a customer may ask us to.

triarius: Everything else look great!

triarius (approved):
🎉

PR #2479, opened 2023-11-02
Remove Org and Pipeline slugs from pipeline invariants and update the signing tool to use the GraphQL API

triarius:
We are going to make repo the only thing in the signature that's not in the pipeline YAML. If customers want to isolate their agents to particular pipelines, they need to isolate their signing and verification keys. 

Also, we add a tool that uses the GraphQL API to read the pipeline YAML and repository URL from the backend and update it after it is signed. This is much more user-friendly than requiring the user to copy the YAML for the initial pipeline into a file and signing that.

The GraphQL is completely optional. Users who don't want to use it will have to find the repository URL themselves and pass it in. This means signing will be completely offline, which has the advantage that the user does not have to trust that the GraphQL API is not deceiving them about the Pipeline YAML and the Repository URL.

There is a remaining UX issue in that the signing tool does not preserve comments or the order of keys in the signed pipeline, this might cause some confusion for users. We should fix this in a future PR.


moskyb (commented):
this all LGTM with a few quibbles, and the experience of using it is very slick - great work!

one thing i'm a little concerned about is using slugs at all ion invariants - it feels like it could lead to a situation where a user renames an org/pipeline and all of their pipelines break, because they're not using the right slug. would it be worth only using IDs as invariants?

triarius:
> one thing i'm a little concerned about is using slugs at all ion invariants - it feels like it could lead to a situation where a user renames an org/pipeline and all of their pipelines break, because they're not using the right slug. would it be worth only using IDs as invariants?

Good point, I'll remove the slugs.

clicommand/tool_sign.go
```diff
@@ -88,70 +135,199 @@ editor in the Buildkite UI so that the agents running these steps can verify the
...
+
+	l.Info("Signing pipeline with invariants:\n%s", prettyPipelineInvariants)
+
+	pipelineYaml := strings.NewReader(resp.Pipeline.Steps.Yaml)
+	parsedPipeline, err := pipeline.Parse(pipelineYaml)
+	if err != nil {
+		return fmt.Errorf("pipeline parsing failed: %v", err)
+	}
+
+	debugL.Debug("Pipeline parsed successfully: %v", parsedPipeline)
```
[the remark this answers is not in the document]
triarius: This should be done now.

clicommand/tool_sign.go
```diff
@@ -59,24 +99,26 @@ editor in the Buildkite UI so that the agents running these steps can verify the
 			Required: true,
 		},
 
-		// Pipeline invariants
+		// These are required to use the GraphQL API
 		cli.StringFlag{
 			Name:     "organization-slug",
-			Usage:    "The organization slug to use when signing the pipeline.",
+			Usage:    "The organization slug to use to talk to the GraphQL API.",
```
triarius, line 105:
```suggestion
			Usage:    "The organization slug. Used to connect to the GraphQL API.",
```

clicommand/tool_sign.go
```diff
@@ -59,24 +99,26 @@ editor in the Buildkite UI so that the agents running these steps can verify the
...
 			EnvVar:   "BUILDKITE_ORGANIZATION_SLUG",
-			Required: true,
+			Required: false,
 		},
 		cli.StringFlag{
 			Name:     "pipeline-slug",
-			Usage:    "The pipeline slug to use when signing the pipeline.",
+			Usage:    "The pipeline slug to use when to talk to the GraphQL API.",
```
triarius, line 111:
```suggestion
			Usage:    "The pipeline slug. Used to connect to the GraphQL API.",
```

clicommand/tool_sign.go
```diff
@@ -88,70 +130,201 @@ editor in the Buildkite UI so that the agents running these steps can verify the
...
+			ErrNotFound,
+			cfg.OrganizationSlug,
+			cfg.PipelineSlug,
+		)
+	}
+
+	debugL.Debug("Pipeline retrieved successfully: %#v", resp)
+	l.Info("Signing pipeline with the repository URL:\n%s", resp.Pipeline.Repository.Url)
+
+	pipelineYaml := strings.NewReader(resp.Pipeline.Steps.Yaml)
```
DrJosh9000, line 252: YAML is an initialism, so I'd write it as `pipelineYAML`, even if the generated code has other ideas. Repeat below (`signedPipelineYAML`, `signedPipelineYAMLBuilder`)

```suggestion
	pipelineYAML := strings.NewReader(resp.Pipeline.Steps.Yaml)
```

internal/pipeline/pipeline_invariants.go
```diff
@@ -5,40 +5,35 @@ import (
...
-		"command":             c.Command,
-		"env":                 c.Env,
-		"plugins":             c.Plugins,
-		"matrix":              c.Matrix,
-		"pipeline_invariants": c.PipelineInvariants,
+		"command":        c.Command,
+		"env":            c.Env,
+		"plugins":        c.Plugins,
+		"matrix":         c.Matrix,
+		"repository_url": c.RepositoryURL,
```
DrJosh9000, line 23: FWIW, putting each invariant into its own signed field is closer to what I had in mind with signed-fields originally (if/when we add more of them, we'll be backwards compatible).

DrJosh9000 (commented):
Generally pretty good!

internal/pipeline/pipeline_invariants.go
triarius: Yeah, I agree. One "pipeline_invariants" field was a mistake. I like the ergonomics of passing them all around in a single struct, but for the signed fields, they should be flattened out.

You may have seen me talk about a way to "compose signedFielders" in the chats somewhere. The idea that you could pass the invariants around in a struct, but the signedFielder implementation would flatten and prefix them appropriately. It's not necessary to do this right now, though, so I didn't.

moskyb (approved):
LGTM - awesoem work getting this done 🎉

clicommand/tool_sign.go
triarius: I've side stepped this by eliminating the variable. I've never liked names like `pipelineYAML`. It's bordering on Hungarian notation.

clicommand/tool_sign.go
```diff
@@ -59,24 +99,26 @@ editor in the Buildkite UI so that the agents running these steps can verify the
 			Required: true,
 		},
 
-		// Pipeline invariants
+		// These are required to use the GraphQL API
```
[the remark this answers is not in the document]
triarius: It used to say "used in" in the usage. It says "required" now. I've added some more copy about the difference between graphql and non graphql modes as well.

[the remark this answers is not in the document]
triarius: Whoops, I did that in the other PR: https://github.com/buildkite/agent/pull/2481

clicommand/tool_sign.go
triarius: Whoops, I did that in the other PR: https://github.com/buildkite/agent/pull/2481

triarius:
I'm going to merge this now and address some of the feedback in https://github.com/buildkite/agent/pull/2481

PR #2481, opened 2023-11-06
Allow omitting the key ID when signing pipelines

triarius:
It will be inferred from the key set if the key set has only one key. I decided to move most of the common parts of key signing shared between `pipeline upload` and `tool sign` to the `jwkutils` module.

Also, I made some of the arguments a bit shorter.

In `pipeline upload` and `tool sign`, the following arguments were changed:
- `jwks-file-path` -> `jwks-file`
- `signing-key-id` -> `jwks-key-id`

I've preferred `file` to `path` because `path` suggests that a directory may be given, but that's not the case. It's also obvious that the key id is used for signing in these sub-commands.

In `start`, there remains a need to distinguish the signing key and verification key to discourage users from distributing both keys everywhere. However, I make the follow changes:
- `job-signing-jwks-path` -> `signing-jwks-file`
- `job-signing-key-id` -> `signing-key-id`
- `job-verification-jwks-path` -> `verification-jwks-path`
- `job-verification-failure-behaviour` -> `verification-jwks-path` 

The reason I've removed `job` from these is
1. jobs aren't signed, steps are.
2. while jobs are verified in some sense, it is more accurate to say that the step signature is cryptographically verified and the job is checked for consistency with the step
3. In any case, I think this type of verification is important enough that we don't need to add the "job" qualifier to it.

In `tool keygen`:
- `private-keyset-filename` -> `private-jwks-file`
- `public-keyset-filename` -> `public-jwks-file`
This is to make it consistent with the other commands, which just specify `file`.

### DIFF
--- a/clicommand/tool_sign.go
+++ b/clicommand/tool_sign.go
@@ -242,8 +242,7 @@ func signWithGraphQL(
 	debugL.Debug("Pipeline retrieved successfully: %#v", resp)
 	l.Info("Signing pipeline with the repository URL:\n%s", resp.Pipeline.Repository.Url)
 
-	pipelineYaml := strings.NewReader(resp.Pipeline.Steps.Yaml)
-	parsedPipeline, err := pipeline.Parse(pipelineYaml)
+	parsedPipeline, err := pipeline.Parse(strings.NewReader(resp.Pipeline.Steps.Yaml))
 	if err != nil {
 		return fmt.Errorf("pipeline parsing failed: %w", err)
 	}
--- a/clicommand/tool_sign.go
+++ b/clicommand/tool_sign.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/buildkite/agent/v3/internal/bkgql"
+	"github.com/buildkite/agent/v3/internal/jwkutil"
 	"github.com/buildkite/agent/v3/internal/pipeline"
 	"github.com/buildkite/agent/v3/internal/stdin"
 	"github.com/buildkite/agent/v3/logger"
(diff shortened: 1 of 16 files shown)


internal/jwkutil/key_in_jwks_file.go
DrJosh9000, line 1: At some point we should remember to add a package comment and doc comments on all the exported things, even though this is in internal, because that'd be helpful to ourselves.

DrJosh9000 (approved):

clicommand/tool_sign.go
```diff
@@ -60,65 +61,63 @@ var ToolSignCommand = cli.Command{
...
+pipeline definition and repo using the Buildkite GraphQL API. If ′update′ is also set, it will
+update the pipeline definition with the signed version using the GraphQL API too.`,
 	Flags: []cli.Flag{
 		cli.StringFlag{
-			Name:     "graphql-token",
-			Usage:    "A token for the buildkite graphql API. This will be used to populate the pipeline invariants if they are not provided.",
-			EnvVar:   "BUILDKITE_GRAPHQL_TOKEN",
-			Required: false,
+			Name:   "graphql-token",
+			Usage:  "A token for the buildkite graphql API. This will be used to populate the value of the repository URL, and download the pipeline definition. Both ′repo′ and ′pipeline-file′ will be ignored in preferance of values from the GraphQL API if the token in provided.",
```
moskyb, line 78: tyop:

```suggestion
			Usage:  "A token for the buildkite graphql API. This will be used to populate the value of the repository URL, and download the pipeline definition. Both ′repo′ and ′pipeline-file′ will be ignored in preference of values from the GraphQL API if the token in provided.",
```

internal/jwkutil/key_in_jwks_file.go
triarius: That point is now! I've added some package and doc comments in https://github.com/buildkite/agent/pull/2481/commits/7bd89c39d36812202728f52d40d62813f1cbaafd, https://github.com/buildkite/agent/pull/2481/commits/0a311fa47d17e4cea795515d8d894c1dd3ec54bb, and https://github.com/buildkite/agent/pull/2481/commits/f2dcf2515faf8e2e7ff07450db0af6931a6ee300.

We can adjust them in another PR.

PR #2485, opened 2023-11-08
Fix race in header times streamer

DrJosh9000:
`sync.WaitGroup`'s Add and Wait methods are racy when called concurrently. There's a better way: channels.


triarius (approved):
LGTM, just a quibble

PR #2487, opened 2023-11-08
Add more comment about channel capacity

DrJosh9000:
(Auto-merge and I raced; auto-merge won.)

Also, 100 should be plenty, but why not 1000? That much memory doesn't cost much.


triarius (approved):
💚

PR #2491, opened 2023-11-09
Fix build number incorporation

DrJosh9000:
scripts/build-binary.sh was setting buildVersion, but the variable is now buildNumber.

Plus an opportunistic shell style cleanup.


triarius (approved):
Whoops, looks like something I changed 😅

Should we add a test that downloads the built agent and run `--version` to ensure it has the build number in it?

DrJosh9000:
Yeah, a small update to .buildkite/steps/check-version-string.sh ?

Issue #50, opened 2014-09-04
Beta agent needs new bootstrap.sh


sj26:
I know it's probably on your TODO, but as a reminder the calls to buildbox artifact need to be updated:

https://github.com/buildbox/agent/blob/v1.0-beta.1/templates/bootstrap.sh#L62-L82

Just fixed these in our bootstrap. :-)

PR #2505, opened 2023-11-19
Refactor hook wrapper writing to enable writing in subprocesses

triarius:
I want to write what used to be known as the script wrapper in a subprocess, so that we can avoid ETXTBSY errors on linux that a customer has reported.

To enable this, I've renamed the script wrapper to the hook wrapper to give it a more specific name. Then, I moved the writing of the template into a separate method, so that it can be called from a subprocess. While doing this, I decided to refactor how tempfiles are written, I noticed the old code did some unnecessary things like return file handles, which may have exacerbated the ETXTBSY problem. Now, temp file creation has its own package.


hook/wrapper.go
```diff
@@ -93,33 +94,33 @@ type scriptWrapperOpt func(*ScriptWrapper)
...
 		wrap.hookPath = path
 	}
 }
 
-func WithOS(os string) scriptWrapperOpt {
-	return func(wrap *ScriptWrapper) {
-		wrap.os = os
+func WithOS(o string) WrapperOpt {
+	return func(wrap *Wrapper) {
+		wrap.os = o
```
triarius, line 116: To avoid shadowing the package name.

internal/job/shell/tempfile.go
```diff
@@ -1,50 +0,0 @@
...
-
-		// Rename it
-		newTempFileName := tempFile.Name() + extension
-		err = os.Rename(tempFile.Name(), newTempFileName)
-		if err != nil {
-			return nil, fmt.Errorf("Failed to rename \"%s\" to \"%s\" (%s)", tempFile.Name(), newTempFileName, err)
-		}
-
-		// Open it again
-		tempFile, err = os.OpenFile(newTempFileName, os.O_RDWR|os.O_EXCL, 0600)
```
triarius, line 43: It seems strange to use `O_EXCL` here, as the file was created by the earlier call to `os.CreateTemp` and then (maybe) renamed to its current location. So the file should already exist, and this may be expected to error.

However, turns [out](https://man7.org/linux/man-pages/man2/open.2.html#DESCRIPTION):
> In general, the behavior of O_EXCL is undefined if it is used without O_CREAT. There is one exception: on Linux 2.6 and later, O_EXCL can be used without O_CREAT if pathname refers to a block device. If the block device is in use by the system (e.g., mounted), open() fails with the error EBUSY.

So this may have had no effect.

triarius: We avoid all this in the replacement code in `internal/tmpfile/tmpfile.go`.

moskyb:
extremely 👀

hook/wrapper.go
```diff
@@ -169,110 +170,94 @@ func NewScriptWrapper(opts ...scriptWrapperOpt) (*ScriptWrapper, error) {
...
+	return w.wrapperPath
 }
 
-// Close cleans up the wrapper script and the environment files
-func (wrap *ScriptWrapper) Close() {
-	os.Remove(wrap.scriptFile.Name())
-	os.Remove(wrap.beforeEnvFile.Name())
-	os.Remove(wrap.afterEnvFile.Name())
+// Close cleans up the wrapper script and the environment files. Ignores errors, in
+// particular the error from os.Remove if the file doesn't exist.
```
DrJosh9000, line 237: Do we know why it ignored errors before? just a "best effort" cleanup?

hook/wrapper.go
```diff
@@ -169,110 +170,94 @@ func NewScriptWrapper(opts ...scriptWrapperOpt) (*ScriptWrapper, error) {
...
+func (w *Wrapper) Close() {
+	_ = os.Remove(w.wrapperPath)
```
DrJosh9000, line 239: Is there a linter that suggests adding these?

hook/wrapper.go
```diff
@@ -169,110 +170,94 @@ func NewScriptWrapper(opts ...scriptWrapperOpt) (*ScriptWrapper, error) {
...
+	tmpl *template.Template,
+	input WrapperTemplateInput,
+	hookWrapperName string,
+) (name string, err error) {
+	f, err := tmpfile.KeepExtensionWithMode(hookWrapperDir, hookWrapperName, 0o700)
+	if err != nil {
+		return "", err
 	}
-	return wrap, nil
+	defer func() {
```
DrJosh9000, line 223: I think deferring the call to close that checks the error, and returning at `tmpl.Execute`, obscures the main flow, which is:

1. create the file
2. write to the file
3. close the file

Deferring `Close` as cleanup that happens in error flows makes sense. If `Close` did something like panic on a second call, then I can see this being useful.

Also "other: <nil>" would look weird...

So I think this should look more like:

```go
f, err := tmpfile.KeepExtensionWithMode(...)
if err != nil {
    return "", err
}
defer f.Close()
if err := tmpl.Execute(f, input); err != nil {
    return "", err
}
if err := f.Close(); err != nil {
    return "", err
}
return f.Name(), nil
```

internal/tmpfile/tmpfile.go
```diff
@@ -0,0 +1,66 @@
...
+// KeepExtensionWithMode creates a temporary file with the same extension of the
+// provided `filename` and sets its permissions to `perm`. The file will be created in the
+// subdirectory `dir` inside the system temp directory.
+func KeepExtensionWithMode(dir, filename string, perm fs.FileMode) (*os.File, error) {
+	extension := filepath.Ext(filename)
+	basename := strings.TrimSuffix(filename, extension)
+	tempDir := filepath.Join(os.TempDir(), dir)
+
+	// Best effort ensure tempDir exists
+	if _, err := os.Stat(tempDir); os.IsNotExist(err) {
```
DrJosh9000, line 38: Why does it need to check existence of the dir beforehand? `MkdirAll` shouldn't error if it does.

triarius: Good point, I just ported it over from the old code.

internal/tmpfile/tmpfile.go
moskyb, line 1: the mix of `Mode`, `Close`, `ModeAndClose`, as well as the somewhat-hidden option of `dir` makes this seem like a good place to use function opts, eg.

```Go
tmpfile.KeepExtension("my-cool-file", 
  tmpfile.WithDir("buildkite-agent"),
  tmpfile.WithMode(0o600),
  tmpfile.AndClose(),
)
```

i wrote up a patch to see if this is possible - happy to push it if you like, but this is all basically just style stuff - happy to keep it as is as well.

hook/wrapper.go
triarius: Yes, pretty neat huh?

internal/tmpfile/tmpfile.go
triarius: 🤩 Do you mind if I reimplant it so that it gets in my muscle memory?

hook/wrapper.go
triarius: IDK, I belive we started to remove them here in d062287f7a7074ccb3d269f3994a4a437283e811, but that looks like a marathon commit so I have no I idea why.

I think it might be a good idea to put all these in the same directory and clean the entire directory. Happy to do that here. I can change it to pass up the error then too. LMK what you think.

internal/tmpfile/tmpfile.go
moskyb: please!

hook/wrapper.go
DrJosh9000: Oh dear, history homework. This is fine.

hook/wrapper.go
triarius: Ok, I've done this.

internal/tmpfile/tmpfile.go
triarius: I've done this now.

internal/tmpfile/tmpfile.go
triarius: @moskyb 
Because the `AndClose` option requires a different return type, I've implemented it as a separate method. Not sure if there's a simple way to achieve a different return type with the options pattern, but if you've found one, can you share it?

triarius: Otherwise, everything is options now.

moskyb: when i did it, i made the `KeepExtensions` method return a closed file handle and have users call `f.Name()`, but i suspect that that's not the best way to do things

moskyb: i also think that `New` and `NewClosed` are better user experience, though

triarius: Kk, I think not returning the file handle is part of the point of the method, maybe I should call it `NewReturningName` or something like that. The fact that was open and closed is an implementation detail, but a pretty important one in the context of ETXTBSY errors! So I think the name `NewClosed` has less friction.

moskyb: yeah, +1 on `NewClosed`. sounds good.

internal/tempfile/tmpfile_test.go
```diff
@@ -0,0 +1,173 @@
...
+	"gotest.tools/v3/assert"
+)
+
+func TestNew(t *testing.T) {
+	t.Parallel()
+
+	for _, test := range []struct {
+		name  string
+		opts  []tempfile.TempFileOpts
+		check func(t *testing.T, f *os.File)
```
DrJosh9000, line 20: This is an interesting approach to the problem of different test cases requiring different logic to check. I'm not super opposed to it, because it has the interesting property that test-case-specific checking logic is _right there_ in the test case. But then... would it read better if done as separate tests or non-table-driven subtests?

https://google.github.io/styleguide/go/decisions#table-driven-tests says:

> When some test cases need to be checked using different logic from other test cases, it is more appropriate to write multiple test functions, as explained in [GoTip #50: Disjoint Table Tests](https://google.github.io/styleguide/go/index.html#gotip). The logic of your test code can get difficult to understand when each entry in a table has its own different conditional logic to check each output for its inputs. If test cases have different logic but identical setup, a sequence of [subtests](https://google.github.io/styleguide/go/decisions#subtests) within a single test function might make sense.

I will also point out that, if each case was separated out (i.e. not table driven), there'd be no need for `opts`, which appears to be the only use of `TempFileOpts` outside of `tempfile`, so you could avoid exporting it.

internal/tempfile/tmpfile_test.go
```diff
@@ -0,0 +1,173 @@
...
+		test := test
+
+		t.Run(test.name, func(t *testing.T) {
+			t.Parallel()
+
+			f, err := tempfile.New(test.opts...)
+			assert.NilError(t, err, `New(%v) = %v`, test.opts, err)
+
+			t.Cleanup(func() {
+				assert.NilError(t, f.Close(), "failed to close file: %s", f.Name())
```
DrJosh9000, line 85: `assert.NilError` uses `t.FailNow`, which stops the test immediately (https://go.dev/play/p/4QnMdl90ZUn). 

But since this is cleanup code - we should probably try to `os.Remove` the file whether or not `f.Close` succeeded.

DrJosh9000 (commented):
This is really very good! As a reward, you get another pass from me

internal/tempfile/tmpfile_test.go
triarius: They are `assert.Check`s now. We could also use multiple `t.Cleanup`s, but it gets a bit confusing b/c they're LIFO. It's no more complicated than `defer` thought, but in my experience, it's more common to write multi-statement `defer`s than to write multiple consecutive `defer` statements.

internal/tempfile/tempfile_test.go
```diff
@@ -0,0 +1,180 @@
...
+
+func TestNewWithDir(t *testing.T) {
+	t.Parallel()
+
+	f, err := tempfile.New(tempfile.WithDir("foo"))
+	assert.NilError(t, err, `New(WithDir("foo")) = %v`, err)
+
+	t.Cleanup(func() {
+		assert.Check(t, f.Close() == nil, "failed to close file: %s", f.Name())
+		assert.Check(t, os.Remove(f.Name()) == nil, "failed to remove file: %s", f.Name())
```
DrJosh9000, line 50: Since `WithDir` creates `foo`, this cleanup ought to `os.RemoveAll(filepath.Join(os.TempDir(), "foo"))`

(Don't forget `TestNewClosedWithDir`)

DrJosh9000 (approved):
Nice! Here's one final thing.

Issue #2511, opened 2023-11-22
Set EnvVar for git submodule mirrors to support volume mounting git mirrors in plugins


tmcg-gusto:
**Is your feature request related to a problem? Please describe.**
We have a project that uses a git submodule for some external data, and runs its CI steps using the `docker-buildkite-plugin`. We're using the latest `buildkite-agent` so git-mirrors are enabled everywhere.

The agent provides the `BUILDKITE_REPO_MIRROR` env var so that the plugins can volume mount the mirror directory into the containers, but there is no variable provided for any submodules, resulting in errors attempting to do git operations on the submodule inside the CI container.

**Describe the solution you'd like**
The default checkout phase should set environment variable(s) for all git submodule mirrors used for the checkout so that plugin authors can add mirror volume mounts. I imagine it'd need to look something like `BUILDKITE_REPO_SUBMODULE_MIRROR_<x>` with x being an integer, similar to how the other plugin variables work for arrays.

**Describe alternatives you've considered**
If I could disable git-mirrors just for this pipeline I would, since this repo is relatively small (please let me know if I missed an env var that'd do that for me). That'd be a fine solution for us, but having the mirrors mount properly would be preferable long term.

Our current workaround is to hardcode a volume mount for the submodule mirror in our docker plugin config.

**Additional context**
The operations that are failing in our containers are `go build`s (with buildvcs enabled, buildvcs=false works around the issue for the build steps) and a golang linter that needs to run `go list all` which apparently needs the submodule's git dir.

PR #2512, opened 2023-11-22
Set EnvVar for Submodule mirrors

tmcg-gusto:
A quick attempt at solving this issue. https://github.com/buildkite/agent/issues/2511

TL;DR: The agent doesn't provide an env var containing the directories of submodule git mirrors, so plugins (docker, in my case) can't mount those directories, causing git issues in containers.

Let me know if I missed anything obvious.

For testing, I've been struggling to get tests to pass locally (both on main, and on my branch). Do I need to be using a valid buildkite agent token in order to run tests? It looks like it's failing to send real data off to buildkite's API. Once I get the tests working I can add an expectation in the checkout integration test.

### DIFF
--- a/internal/job/checkout.go
+++ b/internal/job/checkout.go
@@ -600,6 +600,7 @@ func (e *Executor) defaultCheckoutPhase(ctx context.Context) error {
 			e.shell.Warningf("Failed to enumerate git submodules: %v", err)
 		} else {
 			mirrorSubmodules := e.ExecutorConfig.GitMirrorsPath != ""
+			submoduleMirrorDirs := make([]string, 0)
 			for _, repository := range submoduleRepos {
 				submoduleArgs := append([]string(nil), args...)
 				// submodules might need their fingerprints verified too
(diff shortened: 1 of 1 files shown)


internal/job/checkout.go
```diff
@@ -642,6 +645,10 @@ func (e *Executor) defaultCheckoutPhase(ctx context.Context) error {
 				}
 			}
 
+			for i, dir := range submoduleMirrorDirs {
+				e.shell.Env.Set(fmt.Sprintf("BUILDKITE_REPO_SUBMODULE_MIRROR_%d", i), dir)
+			}
```
triarius, line 650: Would it be better to have 1 env var and a comma separated list as its value? I suspect its easier to get one env variable and parse the comma separated list than to search for all the the env variables containing the submodule mirror dirs. I know the docker plugin is in bash, which makes parsing the value hard, but it should be possible: https://stackoverflow.com/questions/27702452/loop-through-a-comma-separated-shell-variable

triarius (commented):
Thanks for adding this @tmcg-gusto. I have some opinions about the design. LMK what you think.

internal/job/checkout.go
tmcg-gusto: @triarius Yeah we could definitely do that, and that'd likely be easier to work with. 

My goal for using the integer suffix was to keep things consistent with how it's done for arrays of configs in plugins. I imagine many plugins already have patterns for consuming those sort of environment variables (I know I've written it in a few plugins myself).

If y'all would prefer the comma separated list since we can safely assume there won't be commas in the mirror paths I'll make the adjustments.

dabarrell: Hey @tmcg-gusto - apologies it's taken some time to get a reply to you! I'd agree with @triarius that a comma-separated list in a single env var would be appropriate here. Would you mind making the adjustment then we can get this merged in?

DrJosh9000:
Closing due to staleness - if this is something that you're still interested in having, just shout.

PR #2513, opened 2023-11-23
Humanize bytes to IEC (1024 → KiB etc) not SI (1000 → KB etc)

pda:
When displaying byte quantities, use IEC / base-2 (1024 → KiB etc) rather than SI / base-10 (1000 → KB etc) units.

Using IEC / base-2 units matches our server-side display; consistency reduces confusion.
(Rails defaults to IEC / base-2, although it labels them e.g. “KB” not “KiB”).

Related:
- https://en.wikipedia.org/wiki/Byte#Multiple-byte_units
- https://pkg.go.dev/github.com/dustin/go-humanize#IBytes


mitchbne (approved):
LGTM 🚀

DrJosh9000 (approved):
⛳️

PR #2520, opened 2023-11-27
Add a helpful note to security researchers

DrJosh9000:


triarius (approved):
😂 though we should perhaps make a more general comment about this on hacker one

PR #2524, opened 2023-11-29
Fix data race in log_streamer_test

DrJosh9000:
Fortunately this seems to be limited to the test.


triarius (approved):
Thanks for fixing this 🙏

How did it ever work w/o the sort?

DrJosh9000:
> How did it ever work w/o the sort?

I suspect the test callback is fast enough that the log streamer almost always naturally sequentialises.

PR #2525, opened 2023-11-30
Fix data race on Process.pid

DrJosh9000:
The process can start, run, and output "Ready" before `p.pid` is set. Use of `p.pid` in `Interrupt` and `Terminate` is guarded by `p.mu`, but setting `p.pid` was not guarded. Even with the guard, it is still possible that `Interrupt` is called before `p.pid` is set, so I restored `<-p.Started()` (that I removed in #1852).

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -89,13 +89,14 @@ type Config struct {
 
 // Process is an operating system level process
 type Process struct {
-	waitResult    error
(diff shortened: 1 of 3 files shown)


moskyb (approved):
man, you're chunking through the data races @DrJosh9000, awesome work

PR #2526, opened 2023-12-01
Fix typo in environment variable name for allowed-plugins

moskyb:
i came across this when reviewing https://github.com/buildkite/agent/pull/2406 - i'm kind of shocked that this made it through review. i blame the reviewer (me).

sigh. this is technically a breaking change. what do we do about that?

cc/ @jakubm-canva @david-poirier


jakubm-canva:
😨 I don't even know how that got through 🤦 AFAIK, we (canva) are not using it yet.

triarius (approved):

moskyb:
i'm going to explain away the breaking-ness of this change thusly:

no one who used this environment variables (if indeed anyone has) will have used this and not said "wait, what? it's called `BUILDKITE_PLUGINSS`? what on earth?" so i'm going to call this a bug fix.

PR #2538, opened 2023-12-04
Use os.RemoveAll in cleanup

DrJosh9000:
Because the tests are parallel, using the same directory would cause flakes. So also use a different directory for each test.

### DIFF
--- a/internal/tempfile/tempfile_test.go
+++ b/internal/tempfile/tempfile_test.go
@@ -42,15 +42,17 @@ func TestNewWithFilename(t *testing.T) {
 func TestNewWithDir(t *testing.T) {
 	t.Parallel()
 
-	f, err := tempfile.New(tempfile.WithDir("foo"))
-	assert.NilError(t, err, `New(WithDir("foo")) = %v`, err)
+	f, err := tempfile.New(tempfile.WithDir("TestNewWithDir"))
+	assert.NilError(t, err, `New(WithDir("TestNewWithDir")) = %v`, err)
+
+	dir := filepath.Join(os.TempDir(), "TestNewWithDir")
 
 	t.Cleanup(func() {
 		assert.Check(t, f.Close() == nil, "failed to close file: %s", f.Name())
-		assert.Check(t, os.Remove(f.Name()) == nil, "failed to remove file: %s", f.Name())
+		assert.Check(t, os.RemoveAll(dir) == nil, "failed to remove file: %s", f.Name())
 	})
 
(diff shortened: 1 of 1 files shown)


internal/tempfile/tempfile_test.go
```diff
@@ -42,15 +42,17 @@ func TestNewWithFilename(t *testing.T) {
...
-	assert.NilError(t, err, `New(WithDir("foo")) = %v`, err)
+	f, err := tempfile.New(tempfile.WithDir("TestNewWithDir"))
+	assert.NilError(t, err, `New(WithDir("TestNewWithDir")) = %v`, err)
+
+	dir := filepath.Join(os.TempDir(), "TestNewWithDir")
 
 	t.Cleanup(func() {
 		assert.Check(t, f.Close() == nil, "failed to close file: %s", f.Name())
-		assert.Check(t, os.Remove(f.Name()) == nil, "failed to remove file: %s", f.Name())
+		assert.Check(t, os.RemoveAll(dir) == nil, "failed to remove file: %s", f.Name())
```
triarius, line 52:
```suggestion
		assert.Check(t, os.RemoveAll(dir) == nil, "failed to remove dir: %s", dir)
```

triarius: Maybe it should be more like this

internal/tempfile/tempfile_test.go
```diff
@@ -127,14 +129,16 @@ func TestNewClosedWithFilename(t *testing.T) {
...
-	filename, err := tempfile.NewClosed(tempfile.WithDir("foo"))
-	assert.NilError(t, err, `NewClosed(WithDir("foo")) = %v`, err)
+	filename, err := tempfile.NewClosed(tempfile.WithDir("TestNewClosedWithDir"))
+	assert.NilError(t, err, `NewClosed(WithDir("TestNewClosedWithDir")) = %v`, err)
+
+	dir := filepath.Join(os.TempDir(), "TestNewClosedWithDir")
 
 	t.Cleanup(func() {
-		assert.Check(t, os.Remove(filename) == nil, "failed to remove file: %s", filename)
+		assert.Check(t, os.RemoveAll(dir) == nil, "failed to remove file: %s", filename)
```
triarius, line 138:
```suggestion
		assert.Check(t, os.RemoveAll(dir) == nil, "failed to remove dir: %s", dir)
```

triarius (approved):
Thanks for cleaning up (even more) after me!

PR #2539, opened 2023-12-05
Environment Variable allowlisting

moskyb:
**This work is heavily based off of the great work of @CheeseStick in https://github.com/buildkite/agent/pull/2406. Thanks so much for doing the majority of this work @CheeseStick!!** i've forked your branch so that we can make a couple of changes easier, and so i'm not pushing to your fork all the time.

**This PR:** Adds a new feature to the agent, environment variable allowlisting. This should allow operators of the agent to prevent jobs that have unknown environment variables from being run on their agents.

This behavior is controlled by two flags on the agent: `--enable-environment-variable-allowlist`, which engages the feature, and adds any buildkite-sourced environment variables to the allowlist, and `--allowed-environment-variables`, which allows users to define a list of patterns of envars they want passed into their jobs. For example:
```Sh
buildkite-agent start /* ... */ \
  --enable-environment-variable-allowlist \
  --allowed-environment-variables "^DATADOG_.*$,^MY_APP_SPECIFIC_VARIABLE$"
```

Should a job definition contain any environment variables that don't match the allowed patterns, the agent will refuse the job, and put a message into both the job log and the agent log.

Using the above config as an example, if i tried to add the envar `CRIMES=yes` into my job via the buildkite web console, it would output an error:

CleanShot 2023-12-05 at 13 40 19@2x


agent/run_job.go
```diff
@@ -166,6 +161,87 @@ func (r *JobRunner) Run(ctx context.Context) error {
...
+		"plugins": func() error {
+			return r.validatePlugins()
+		},
+	}
+
+	for name, validation := range validations {
+		if err := validation(); err != nil {
+			return fmt.Errorf("failed to validate %s: %w", name, err)
+		}
+	}
```
DrJosh9000, line 181: A map of validations and a loop feels like an excessive amount of mechanism when there's only three of them...

```suggestion
	if err := validateJobValue(r.conf.AgentConfiguration.AllowedRepositories, job.Env["BUILDKITE_REPO"]); err != nil {
		return fmt.Errorf("failed to validate BUILDKITE_REPO: %w", err)
	}
	if err := validateEnv(job.Env, r.conf.AgentConfiguration.AllowedEnvironmentVariables); err != nil {
		return fmt.Errorf("failed to validate environment variables: %w", err)
	}
	if err := r.validatePlugins(); err != nil {
		return fmt.Errorf("failed to validate plugins: %w", err)
	}
```

agent/run_job.go
```diff
@@ -166,6 +161,87 @@ func (r *JobRunner) Run(ctx context.Context) error {
...
+		"repo": func() error {
+			return validateJobValue(r.conf.AgentConfiguration.AllowedRepositories, job.Env["BUILDKITE_REPO"])
+		},
+		"environment variables": func() error {
+			return validateEnv(job.Env, r.conf.AgentConfiguration.AllowedEnvironmentVariables)
+		},
+		"plugins": func() error {
+			return r.validatePlugins()
+		},
+	}
```
triarius, line 175: 🤩

agent/run_job.go
```diff
@@ -166,6 +161,87 @@ func (r *JobRunner) Run(ctx context.Context) error {
...
+		if err := validation(); err != nil {
+			return fmt.Errorf("failed to validate %s: %w", name, err)
+		}
+	}
+
+	return nil
+}
+
+// validateEnv checks if the environment variables are allowed by ensuring their names match at least one of the regular
+// expressions in the allowedVariablePatterns list. The error message will contain all of the variables that aren't are
```
DrJosh9000, line 187: "The error message will contain all of the variables that aren't are invalid." 

Grammar-o aside, is that true?

triarius (approved):
Looks great! Thank you, @CheeseStick and @moskyb!

agent/run_job.go
triarius: It's consistent with WET vs DRY. I'll allow it.

DrJosh9000: Then at least:
```go
"plugins": r.validatePlugins,
```

triarius:
BTW you can add `Closes: #2406` to the PR description to automate the closing of the other PR.

agent/run_job.go
moskyb: ahhh not any more! i will make it true again

PR #2543, opened 2023-12-06
Add more debug logging and error wrapping for running processes

triarius:
A customer is seeing errors trying to start a process on an exotic platform that we currently don't support. They are getting an error from `(*Process).Run`. Let's add some debug logging and error wrapping to see what's going on.

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -170,12 +170,14 @@ func (p *Process) Run(ctx context.Context) error {
 
 	// Toggle between running in a pty
 	if p.conf.PTY {
+		p.logger.Debug("[Process] Running with a PTY")
+
 		// Commands like tput expect a TERM value for a PTY
 		p.command.Env = append(p.command.Env, "TERM="+termType)
 
 		pty, err := StartPTY(p.command)
 		if err != nil {
-			return err
+			return fmt.Errorf("error starting pty: %w", err)
 		}
 
 		// Make sure to close the pty at the end.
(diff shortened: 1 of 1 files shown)


moskyb (approved):

process/process.go
```diff
@@ -202,32 +204,37 @@ func (p *Process) Run(ctx context.Context) error {
...
 			_, err = io.Copy(p.conf.Stdout, pty)
-			if e, ok := err.(*os.PathError); ok && e.Err == syscall.EIO {
-				// We can safely ignore this error, because
-				// it's just the PTY telling us that it closed
-				// successfully.  See:
-				// https://github.com/buildkite/agent/pull/34#issuecomment-46080419
-				err = nil
-			}
+			switch {
+			case errors.Is(err, syscall.EIO):
```
triarius, line 210: I guess technically this catches error that are not `os.PathError` that also happen to wrap `syscall.EIO`. I think it's unlikely that `io.Copy` will create such a thing, but I'm happy to do a more specific unwrap if we think otherwise.

v1gnesh:
@triarius Hey, I now see this in the log in the Pipelines page on the site.

```
2023-12-08 14:33:54 UTC	Error running job: error starting pty: unsupported
```

The build script I'm using is [this](https://github.com/zosopentools/buildkiteport)

v1gnesh:
Oh that's a nice option.
`GOOS=zos` and `GOARCH=s390x`.

Although the upstream Go compiler has these keywords reserved, the OS's bits haven't fully been upstreamed yet.
But when running the code/build on this platform, looking for these values, it'll work.

v1gnesh:
@triarius I've successfully run the sample pipeline and all green on the BK site :heart: 
Thank you! I'll continue investigating pty and try more BK stuff.

PR #2569, opened 2023-12-22
PDP-2032 when api returns a 401 stop retrying and bail out

SorchaAbel:
https://linear.app/buildkite/issue/PDP-2032/on-the-agent-side-we-need-to-not-retry-on-a-401

This PR is related to job tokens

When the agent receives a 401 from the api due to invalid tokens we don't want to keep retrying. Below is a screen shot of what happens when the token expires

This PR is my very poor attempt to stop the agent retrying when it gets a 401. The screenshot below gives a bit more detail

Screenshot 2023-12-22 at 3 58 30 pm

Testing on Localhost
Now when a token expires (buildkite returns a 401) the agent will eventually stop trying however it does seem to repeatedly loop thru the whole process (assigned..started..process is running..job token error..finished..and then back to assigned again). In my test I have a job that has timed out but I'm worried if a job doesn't have `timeout_in_minutes:` would it just run for 30 days or 24 hours (lost grace period exceeded) 😬 
Screenshot 2024-01-05 at 11 13 25 am

### DIFF
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -645,15 +645,17 @@ func (r *JobRunner) jobCancellationChecker(ctx context.Context, wg *sync.WaitGro
 		jobState, response, err := r.apiClient.GetJobState(ctx, r.conf.Job.ID)
 
 		if err != nil {
-			// if the job token is invalid there is no point in retrying
-			if jobState.State == "canceling" || jobState.State == "canceled" || (response != nil && response.StatusCode == 401) {
+			if response != nil && response.StatusCode == 401 {
 				if err := r.Cancel(); err != nil {
-					r.agentLogger.Error("Unexpected error canceling process as requested by server (job: %s) (err: %s)", r.conf.Job.ID, err)
+					r.agentLogger.Error("Invalid access token: an error occurred in the canceling process (job: %s) (err: %s)", r.conf.Job.ID, err)
 				}
 			} else {
-				// We don't really care if it fails, we'll just
-				// try again soon anyway
+				// We don't really care if it fails, we'll just try again soon anyway
 				r.agentLogger.Warn("Problem with getting job state %s (%s)", r.conf.Job.ID, err)
+			}
+		} else if jobState.State == "canceling" || jobState.State == "canceled" {
+			if err := r.Cancel(); err != nil {
+				r.agentLogger.Error("Unexpected error canceling process as requested by server (job: %s) (err: %s)", r.conf.Job.ID, err)
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -107,8 +107,13 @@ cd agent
...
 # Build an agent binary and start the agent.
-go build -o /usr/local/bin/buildkite-agent .
-buildkite-agent start --debug --build-path=/tmp/buildkite-builds --token "abc"
+go build -o /tmp/buildkite-builds/ .
+
+# Check the version of the build 
+/tmp/buildkite-builds/agent --version
+
+# To run the agent locally
+/tmp/buildkite-builds/agent start --debug --build-path=/tmp/buildkite-builds --token "abc"  --endpoint http://agent.buildkite.localhost/v3
```
dabarrell, line 116: Ohhhh I see where you were coming from. The README (without this change) was suggesting creating a temp folder (`/tmp/buildkite-builds`) to use as the working directory for the jobs the agent is running, not to build the agent in that folder. `go build -o /usr/local/bin/buildkite-agent .` would build the agent into `/usr/local/bin/`, which is on $PATH by default, meaning when you then run `buildkite-agent start` it uses the one you've built locally. So the instructions there would work unless you already have a `buildkite-agent` in your path (i.e. because you installed it using homebrew etc)

SorchaAbel: ah lol mystery solved. Thanks @dabarrell

dabarrell: I'd suggest leaving it closer to the original (not building the binary into a temp folder), but making it not clash with an install of the published version. Maybe @moskyb , @DrJosh9000 or @triarius could suggest the workflow they use

dabarrell: Could do something like

```
go build -o /usr/local/bin/buildkite-agent-local .
buildkite-agent-local start --debug --build-path=/tmp/buildkite-builds --token "abc"
```

so you could run both the local and published versions as needed (`buildkite-agent` vs `buildkite-agent-local`), perhaps?

SorchaAbel: @dabarrell I dropped the commit for the README, I think my changes are unnecessary and strictly speaking they don't belong in this PR :) But thanks for all your comments

agent/job_runner.go
```diff
@@ -642,11 +642,17 @@ func (r *JobRunner) jobCancellationChecker(ctx context.Context, wg *sync.WaitGro
...
-		jobState, _, err := r.apiClient.GetJobState(ctx, r.conf.Job.ID)
+		jobState, response, err := r.apiClient.GetJobState(ctx, r.conf.Job.ID)
+
 		if err != nil {
-			// We don't really care if it fails, we'll just
-			// try again soon anyway
-			r.agentLogger.Warn("Problem with getting job state %s (%s)", r.conf.Job.ID, err)
+			if response != nil && response.StatusCode == 401 {
+				if err := r.Cancel(); err != nil {
+					r.agentLogger.Error("Invalid access token: an error occurred in the canceling process (job: %s) (err: %s)", r.conf.Job.ID, err)
```
DrJosh9000, line 650: I think we should have two separate log lines, one for the 401 status ("the token was invalid") and another for then failing to cancel. Both are errors, but have different vibes:

```suggestion
			if response != nil && response.StatusCode == 401 {
				r.agentLogger.Error("Invalid access token, cancelling job %s", r.conf.Job.ID)
				if err := r.Cancel(); err != nil {
					r.agentLogger.Error("Failed to cancel the process (job: %s): %v", r.conf.Job.ID, err)
```

(Aside: I've also tweaked the format token for err to `%v`, which is different from the existing code nearby; although the underlying `fmt` is flexible enough to accept a variety of things for `%s`, `%s` looks like the corresponding arg should be typed as string, and all example code in https://google.github.io/styleguide/go uses `%v` for errors.)

PR #2589, opened 2024-01-17
Add readme note about go module versioning

moskyb:
we change public method signatures all the time, and we should call it out to users in the readme. i more or less crammed this in the first place in the readme i thought would fit it, but i'm happy for it to go elsewhere

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,10 @@ our Go dependencies. Dependencies are not
 [vendored](https://go.dev/ref/mod#go-mod-vendor) into the repository unless
 necessary.
 
+The go module published by this repo (ie the one you could use by adding `import "github.com/buildkite/agent/v3"` to your code)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -126,6 +126,10 @@ our Go dependencies. Dependencies are not
 [vendored](https://go.dev/ref/mod#go-mod-vendor) into the repository unless
 necessary.
 
+The go module published by this repo (ie the one you could use by adding `import "github.com/buildkite/agent/v3"` to your code)
```
DrJosh9000, line 129:
```suggestion
The Go module published by this repo (i.e. the one you could use by adding `import "github.com/buildkite/agent/v3"` to your code)
```

DrJosh9000 (approved):
Agreed, not a thing we want to guarantee.

PR #2613, opened 2024-01-31
Fix powershell hooks on windows

triarius:
### Description

Powershell pre-bootstrap hooks on Windows were broken because the hook was being run as an executable, but that does not seem to always work on Windows. This did not happen for other hook types as they are run through a mechanism that checks if the hook is a script an invokes powershell.exe if appropriate. I've changed it to use that mechanism for pre-bootstrap hooks too.

### Context

https://coda.io/d/Escalations-Feedback_dHnUHNps1YO/Pipelines-Escalations_su7FT#Pipelines-Escalations-Board_tu__K/r245

### Changes
- Added tests that hooks and pre-bootstrap hooks run on Windows. Previously, we did not test PowerShell hooks on Windows, which is why these bugs crept in. We did a test that the hook wrapper template rendered as expected, but I think this test is of limited value, so I deleted it. Now there are tests that use real hook and test that they run and (if applicable) that the hook wrappers behave as expected.
- Pre-bootstrap hooks will be run as scripts.
- The hook wrappers invoke the same binary that created them, rather than a hard-coded `buildkite-agent`. This is generally a good idea, but it also allows running the hook wrappers in tests where there is no `buildkite-agent` in the PATH. That's the case for the Linux tests which are run in docker.
- I moved the `hook` package to `internal/job/hook` and changed a method name in it. Nobody is (publicly) importing it: https://pkg.go.dev/github.com/buildkite/agent/v3@v3.62.0/hook?tab=importedby

### Testing


agent/job_runner.go
```diff
@@ -582,9 +582,8 @@ func (r *JobRunner) executePreBootstrapHook(ctx context.Context, hook string) (b
 		l: r.agentLogger,
 	}
 
-	if err := sh.RunWithoutPrompt(ctx, hook); err != nil {
-		fmt.Printf("err: %s\n", err)
-		r.agentLogger.Error("Finished pre-bootstrap hook %q: job rejected", hook)
+	if err := sh.RunScript(ctx, hook, nil); err != nil {
```
triarius, line 585: This can actually run binary hooks as well, it falls back to just trying to execute the file if it can't detect the script type: https://github.com/buildkite/agent/blob/4a98b41cac8dc9e9b811c89c51565459dba1b757/internal/job/shell/shell.go#L434-L437

agent/integration/job_runner_integration_test.go
```diff
@@ -4,14 +4,117 @@ import (
...
+				allowed:  false,
+			},
+		)
+	}
+
+	for _, tc := range testCases {
+		tc := tc
+
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
```
DrJosh9000, line 77: I think `RunScript` also runs scripts with shebang lines directly, so might this occasionally flake with the ETXTBSY bug?

triarius: Good point, I'll write the hook in a subprocess.

triarius: It's written by a subprocess now. PTAL

agent/integration/main_test.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/urfave/cli"
+)
+
+var WriteExecutableCommand = cli.Command{
+	Name:  "write-exec",
+	Usage: "Write STDIN to an executable file",
+	Action: func(c *cli.Context) error {
+		path := c.Args().First()
+
+		f, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0o755)
```
triarius, line 21: I wanted to use `os.Create` here, but it didn't take an argument to set the file permissions, which I need to be executble.

DrJosh9000: That's reasonable. You could also call `f.Chmod(0o755)` separately.

agent/integration/job_runner_integration_test.go
DrJosh9000: Hmmm... hang on a sec...

The scripts are all very short, and don't change. Wouldn't it be less work to make them test fixtures?

triarius: Considering I've already done the work, is this moot?

DrJosh9000: I suppose so. If anything I just regret not having the realisation 5 days ago that it could have been implemented with a handful of files checked into the repo. 

Though I suppose this also has to be maintained... no, not going to argue further. It's fine.

DrJosh9000 (approved):
It's a clear improvement in multiple ways, so let's :shipit:

PR #2614, opened 2024-02-02
Fix bug where unauthorised register was retrying erroneously

moskyb:
### Description

A delicious, delicious single-character typo has caused unauthorised checkouts to retry for 5 minutes. There's no recovering from this error, so it shouldn't be retried.

### Testing


agent/register.go
```diff
@@ -53,15 +56,10 @@ func Register(ctx context.Context, l logger.Logger, ac APIClient, req api.AgentR
...
-		registered, resp = reg, rsp
-		return nil
-	}
 
-	// Try to register, retrying every 10 seconds for a maximum of 30 attempts (5 minutes)
-	err := roko.NewRetrier(
-		roko.WithMaxAttempts(30),
-		roko.WithStrategy(roko.Constant(10*time.Second)),
-	).DoWithContext(ctx, register)
+		registered = reg
```
triarius, line 60: Does `registered` need to be different from `reg`?

triarius (approved):
Ouch, LGTM

agent/register.go
moskyb: i don't think it does! will change

triarius (approved):

DrJosh9000 (commented):
Oh no! Thanks for fixing my mistake!

PR #2617, opened 2024-02-06
Move check from upload-release-steps.sh to pipeline.yml

DrJosh9000:
### Description

Remove the first `$DRY_RUN` and branch check from the script, and make it a pipeline conditional instead.

This means one less script for an agent to run, when cycling on a PR.

### Context

None, it's just something I spotted.

I did some archaeology: the `$DRY_RUN` check was added to the script (and `branches` removed from the pipeline step) in 341b1f4 (March 2018), but pipeline conditionals were introduced later.

### Changes

See Description.

### Testing

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -255,3 +255,4 @@ steps:
       - build-debian-packages
       - build-github-release
     command: ".buildkite/steps/upload-release-steps.sh"
+    if: 'build.env("DRY_RUN") == "false" && build.branch =~ /^(main|.*-.*-stable)$$/'
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -255,4 +255,4 @@ steps:
       - build-debian-packages
       - build-github-release
     command: ".buildkite/steps/upload-release-steps.sh"
-    if: 'build.env("DRY_RUN") == "false" && build.branch =~ /^(main|.*-.*-stable)$$/'
+    if: build.env("DRY_RUN") == "true" || build.branch =~ /^(main|.*-.*-stable)$$/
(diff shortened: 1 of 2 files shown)


.buildkite/pipeline.yml
```diff
@@ -255,3 +255,4 @@ steps:
       - build-debian-packages
       - build-github-release
     command: ".buildkite/steps/upload-release-steps.sh"
+    if: 'build.env("DRY_RUN") == "false" && build.branch =~ /^(main|.*-.*-stable)$$/'
```
triarius, line 258: Does anything set the `DRY_RUN` environment variable to `"false"`? I think `build.env("DRY_RUN")` will evaluate to `null` if the environment variable is not set. So, even if nothing sets it and the build is on the main branch, this step won't run, which I don't think is the intention.

```suggestion
    if: build.env("DRY_RUN") == "true" || build.branch =~ /^(main|.*-.*-stable)$$/
```
Is this what was intended? Then this step will run on the main or stable branches, or if `DRY_RUN` is set to `true` in the build env.

It's also clearer to me if there are no unnecessary quotes around the YAML.

triarius (commented):
🤔

.buildkite/pipeline.yml
DrJosh9000: Argh, too many negations. Yes, you're right - the negation of

```shell
[[ ${BUILDKITE_BRANCH}" != "main" && "${DRY_RUN}" != "true" && "${BUILDKITE_BRANCH}" != *-*-stable ]]
```
is 

```shell
[[ ${BUILDKITE_BRANCH}" == "main" || "${DRY_RUN}" == "true" || "${BUILDKITE_BRANCH}" == *-*-stable ]]
```

triarius (approved):
🎉

Issue #2612, opened 2024-01-30
Buildkite agent for MacOS errors out on extraction


rogerhu:
```
wget https://github.com/buildkite/agent/releases/download/v3.62.0/buildkite-agent-darwin-arm64-3.62.0.tar.gz
tar xvfpz buildkite-agent-darwin-arm64-3.62.0.tar.gz 
x ./: Can't restore time
x ./buildkite-agent
x ./bootstrap.sh
x ./buildkite-agent.cfg
x ./hooks/
x ./hooks/checkout.sample
x ./hooks/command.sample
x ./hooks/environment.sample
x ./hooks/post-artifact.sample
x ./hooks/post-checkout.sample
x ./hooks/post-command.sample
x ./hooks/pre-artifact.sample
x ./hooks/pre-checkout.sample
x ./hooks/pre-command.sample
x ./hooks/pre-exit.sample
tar: Error exit delayed from previous errors.
```

The tar version:

```
tar --version
bsdtar 3.5.3 - libarchive 3.5.3 zlib/1.2.12 liblzma/5.0.5 bz2lib/1.0.8 
```

triarius:
Hi @rogerhu. We've tested that this archive can be extracted. We suspect that the output line
```
x ./: Can't restore time
```
indicates some permissions issue with the location you're extracting the archive to. Perhaps try to run `tar xvfpz buildkite-agent-darwin-arm64-3.62.0.tar.gz` in a different directory.

rogerhu:
It seems to bail when unpackaging it inside `/tmp`!

PR #2619, opened 2024-02-07
build-github-release.sh tidyups

DrJosh9000:
### Description

Opportunistic improvements to this one file.

### Context

Makes the tarball less susceptible to generating reports like #2612

### Changes

* Shell style tidyups, mostly double-quoting and curly braces
* Avoiding tarballing `.`

### Testing


triarius (approved):
✨

PR #2621, opened 2024-02-07
Fix environment variable changes in hooks logged incorrectly

triarius:
### Description

When a hook changes an environment variable, and the environment
variable is on an allow list of known non-secrets, the change is logged
as
```
<env var> is now <new value>
```
At least that was the intention. However, what was actually logged was
```
<env var> is now {<old value> <new value>}
```

This changes it to
```
<env var> was <old value> and is now <new value>
```

### Context

https://coda.io/d/Escalations-Feedback_dHnUHNps1YO/Pipelines-Escalations_su7FT#Pipelines-Escalations-Board_tu__K/r252&view=modal

### Testing

### DIFF
--- a/internal/job/executor.go
+++ b/internal/job/executor.go
@@ -521,7 +521,7 @@ func (e *Executor) applyEnvironmentChanges(changes hook.EnvChanges, redactors re
 	}
 	for k, v := range changes.Diff.Changed {
 		if _, ok := executorConfigEnvChanges[k]; ok {
-			e.shell.Commentf("%s is now %q", k, v)
+			e.shell.Commentf("%s was %q and is now %q", k, v.Old, v.New)
 		} else {
 			e.shell.Commentf("%s changed", k)
 		}


internal/job/executor.go
```diff
@@ -521,7 +521,7 @@ func (e *Executor) applyEnvironmentChanges(changes hook.EnvChanges, redactors re
 	}
 	for k, v := range changes.Diff.Changed {
 		if _, ok := executorConfigEnvChanges[k]; ok {
-			e.shell.Commentf("%s is now %q", k, v)
+			e.shell.Commentf("%s was %q and is now %q", k, v.Old, v.New)
```
triarius, line 524: `v` is an [`env.DiffPair`](https://github.com/buildkite/agent/blob/20f215b5bc69ec968d08cd0cabd49ddb50e0cf67/env/environment.go#L284-L287)

DrJosh9000 (approved):

moskyb (approved):

PR #2624, opened 2024-02-09
Fix pipeline interpolation case sensitivity on Windows, and runtime environment variable precedence

triarius:
### Description

On windows, we should allow interpolation of runtime environment variables into pipeline YAML to be case insensitive, as that's what would happen if the commands ruan in a Windows shell.

### Context

https://coda.io/d/Escalations-Feedback_dHnUHNps1YO/Pipelines-Escalations_su7FT#Pipelines-Escalations-Board_tu__K/r251&view=modal

### Changes

Use a [v0.4.0](https://github.com/buildkite/go-pipeline/releases/tag/v0.4.0) of `go-pipeline`, which has a fix for case sensitivity on Windows. This version also rectifies the [precedence](https://buildkite.com/docs/pipelines/environment-variables#environment-variable-precedence) of runtime environment variables over pipeline environment variable when interpolating a pipeline.

### Testing

### DIFF
--- a/env/environment.go
+++ b/env/environment.go
@@ -105,9 +105,8 @@ func (e *Environment) Exists(key string) bool {
 }
 
 // Set sets a key in the environment
-func (e *Environment) Set(key string, value string) string {
+func (e *Environment) Set(key string, value string) {
 	e.underlying.Store(normalizeKeyName(key), value)
-	return value
 }
 
 // Remove a key from the Environment and return its value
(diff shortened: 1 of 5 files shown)


env/environment.go
```diff
@@ -105,9 +105,8 @@ func (e *Environment) Exists(key string) bool {
 }
 
 // Set sets a key in the environment
-func (e *Environment) Set(key string, value string) string {
+func (e *Environment) Set(key string, value string) {
 	e.underlying.Store(normalizeKeyName(key), value)
-	return value
 }
```
triarius, line 110: I need it to fit the `pipeline.InterpolationEnv` interface. It turns out nobody was using the return value anyway.

moskyb (approved):
LGTM!

PR #2631, opened 2024-02-14
Go 1.22

moskyb:
### Description

Updates Golang to 1.22, and makes use of the easy-to-convert features it brings. Most notably:
- Makes use of the new [range-over-int](https://tip.golang.org/doc/go1.22#language) technology
  - i used the regex `for (a-z) := 0; (.*); \1\+\+` to find, and then `for $1 := range $2` to replace. Golang then complained that some of the loop variables weren't being used, so i went through those and manually removed the variables that it was complaining about
- Removes redeclarations of loop variables, as they're now scoped to the body of the loop. hooray!
  - i used the regex `(\w+) := \1\n` to find, and replaced with an empty string

### Context

https://tip.golang.org/doc/go1.22

### Testing

### DIFF
--- a/.buildkite/Dockerfile-compile
+++ b/.buildkite/Dockerfile-compile
@@ -1,4 +1,4 @@
-FROM public.ecr.aws/docker/library/golang:1.21.6@sha256:76aadd914a29a2ee7a6b0f3389bb2fdb87727291d688e1d972abe6c0fa6f2ee0
+FROM public.ecr.aws/docker/library/golang:1.22.0@sha256:ef61a20960397f4d44b0e729298bf02327ca94f1519239ddc6d91689615b1367
 COPY build/ssh.conf /etc/ssh/ssh_config.d/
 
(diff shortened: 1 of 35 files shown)


DrJosh9000 (approved):
💪💪

triarius (approved):
🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉

PR #2599, opened 2024-01-23
BK github app git credentials helper

moskyb:
### Description

this PR adds a couple of pieces of new functionality to the agent, largely centred around an upcoming feature in Buildkite, where in certain situations, the agent will be able to request git pull credentials from the Buildkite Github App, rather than relying on credentials pre-existing on the agent.

the PR adds two main things:
- methods to interact with the new Agent API endpoint that will return git credentials for the agent to use
- a new command in the agent, implemented as a [git credential helper](https://git-scm.com/docs/gitcredentials#_custom_helpers) that, when a clone is run, will request a token from the above endpoint, and inject it into the clone process

Note that the API endpoint for github code access tokens doesn't just blindly pass them out - there are multiple layers of security, the app has to be enabled, and it's all being feature flags right now anyway.

_This PR is about 80% done at time of writing, and still needs to have tests added, and to be tested manually a bit._

### Context

[COMP-14](https://linear.app/buildkite/issue/COMP-14/implement-git-credential-helper)

### Changes

```
$ buildkite-agent git-credentials-helper -h
Usage:

    buildkite-agent git-credential-helper [options...]

Description:

[EXPERIMENTAL]

Ask buildkite for credentials to use to authenticate with Github when cloning via HTTPS.
The credentials are returned in the git-credential format, and are intended to be used as a git credential helper,
which is automatically configured by the agent when it runs a job that qualifies for this feature.

This command will only work if the organization running the job has connected a Github app with Code Access enabled, and
if the pipeline has this feature enabled. All hosted compute jobs automatically qualify for this feature.

This command is intended to be used as a git credential helper, and not called directly.

Options:

  --job-id value              The job id to get credentials for [$BUILDKITE_JOB_ID]
  --agent-access-token value  The access token used to identify the agent [$BUILDKITE_AGENT_ACCESS_TOKEN]
  --endpoint value            The Agent API endpoint (default: "https://agent.buildkite.com/v3") [$BUILDKITE_AGENT_ENDPOINT]
  --no-http2                  Disable HTTP2 when communicating with the Agent API. [$BUILDKITE_NO_HTTP2]
  --debug-http                Enable HTTP debug mode, which dumps all request and response bodies to the log [$BUILDKITE_AGENT_DEBUG_HTTP]
  --no-color                  Don't show colors in logging [$BUILDKITE_AGENT_NO_COLOR]
[...]

PR #2634, opened 2024-02-16
Fix NPE when decoding token response

moskyb:
### Description

a minor refactoring slipup in #2599 meant that the json package was complaining when trying to decode into a `GithubCodeAccessTokenResponse` struct. 

this PR fixes that issue.

### Context

#2599 

### Testing

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -292,9 +292,11 @@ func (c *Client) doRequest(req *http.Request, v any) (*Response, error) {
 			io.Copy(w, resp.Body)
 		} else {
 			if strings.Contains(req.Header.Get("Content-Type"), "application/msgpack") {
-				err = errors.New("Msgpack not supported")
-			} else {
-				err = json.NewDecoder(resp.Body).Decode(v)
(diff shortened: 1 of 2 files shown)


api/client.go
```diff
@@ -292,9 +292,12 @@ func (c *Client) doRequest(req *http.Request, v any) (*Response, error) {
...
 		} else {
 			if strings.Contains(req.Header.Get("Content-Type"), "application/msgpack") {
-				err = errors.New("Msgpack not supported")
+				return response, errors.New("Msgpack not supported")
 			} else {
 				err = json.NewDecoder(resp.Body).Decode(v)
+				if err != nil {
+					return response, fmt.Errorf("failed to decode JSON response: %v", err)
+				}
 			}
```
triarius, line 301:
```suggestion
		err = json.NewDecoder(resp.Body).Decode(v)
		if err != nil {
			return response, fmt.Errorf("failed to decode JSON response: %v", err)
		}
```
If you're returning, you may as well drop the else clause.

triarius (approved):

PR #2636, opened 2024-02-19
Use separate repos for each package type

sj26:
@moskyb pointed out that the packagecloud steps have been unhappy for a while:

image

We made a change a while back that repos are now restricted to one package type, so they need to be split apart. This should help.

I'm a bit unsure exactly where the failure is coming from, though. It's just saying "failed." Let's see if this much fixes it, or if we need to dig a bit deeper.

### DIFF
--- a/.buildkite/pipeline.release-experimental.yml
+++ b/.buildkite/pipeline.release-experimental.yml
@@ -35,7 +35,7 @@ steps:
   - name: ":redhat: Publish Edge RPM Package to Packagecloud"
     command: ".buildkite/steps/publish-rpm-packagecloud.sh"
     env:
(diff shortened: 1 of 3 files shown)


moskyb (approved):

sj26:
I've already made these repos and with aligned package types, and I'm pretty sure all the names match:

image

sj26:
This change will definitely fix [this error on the edge release pipeline](https://buildkite.com/buildkite/agent-release-edge/builds/1129):

image

sj26:
Yeah, the edge pipeline is looking good again:

image

We'll have to wait and see on the beta and mainline release pipelines.

PR #2645, opened 2024-02-21
Add an info log of which experiments are known and enabled on agent start

triarius:
### Description
I often find myself wanting to know which experiments are known and enabled on an agent. While this information is in the debug logs, I think having it also available in the info logs would be useful.

### Context
None

### Changes
The agent startup now looks like this:

### Testing

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -281,7 +281,7 @@ func defaultConfigFilePaths() (paths []string) {
 		}
 	}
 
-	return
+	return paths
 }
 
 var AgentStartCommand = cli.Command{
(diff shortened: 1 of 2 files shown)


moskyb (approved):
love it 🎉

clicommand/agent_start.go
```diff
@@ -281,7 +281,7 @@ func defaultConfigFilePaths() (paths []string) {
 		}
 	}
 
-	return
+	return paths
```
DrJosh9000, line 284: ✅

DrJosh9000 (approved):

PR #2646, opened 2024-02-21
De-experimentify Job API

triarius:
### Description
The Job API experiment has been running for a while and we have decided to promote it to a stable feature. This PR removes the experiment flag from the Job API and updates the `EXPERIMENTS.md` file to promote the JobAPI.
I've also added a toggle on `buildkite-agent bootstrap` to enable/disable the Job API. Let's dicuss the merits of this in this PR.

### Context

A meeting we had

### Changes
- Remove the `job-api` experiment
- Add a flag to enable/disable the Job API on `buildkite-agent bootstrap`

### Testing

### DIFF
--- a/clicommand/bootstrap.go
+++ b/clicommand/bootstrap.go
@@ -95,6 +95,7 @@ type BootstrapConfig struct {
 	RedactedVars                 []string `cli:"redacted-vars" normalize:"list"`
 	TracingBackend               string   `cli:"tracing-backend"`
 	TracingServiceName           string   `cli:"tracing-service-name"`
+	JobAPI                       bool     `cli:"job-api"`
 }
 
 var BootstrapCommand = cli.Command{
@@ -356,6 +357,11 @@ var BootstrapCommand = cli.Command{
 			EnvVar: "BUILDKITE_TRACING_SERVICE_NAME",
 			Value:  "buildkite-agent",
 		},
+		cli.BoolTFlag{
+			Name:   "job-api",
+			Usage:  "Starts the job-api, giving the commands in the job some abilities to intropect and mutate the state of the job.",
+			EnvVar: "BUILDKITE_AGENT_JOB_API",
+		},
 		DebugFlag,
 		LogLevelFlag,
 		ExperimentsFlag,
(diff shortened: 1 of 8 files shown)


clicommand/bootstrap.go
```diff
@@ -356,6 +357,11 @@ var BootstrapCommand = cli.Command{
 			EnvVar: "BUILDKITE_TRACING_SERVICE_NAME",
 			Value:  "buildkite-agent",
 		},
+		cli.BoolTFlag{
+			Name:   "job-api",
+			Usage:  "Starts the job-api, giving the commands in the job some abilities to intropect and mutate the state of the job.",
+			EnvVar: "BUILDKITE_AGENT_JOB_API",
+		},
```
moskyb, line 364: i think we should probably make this `--no-job-api` and make it a `BoolFlag` - the default state should be having the job API enabled, and then allowing people to disable it, rather than having the UX be `--job-api false`, which reads a little confusingly, i think. 

there's a fair bit of prior art [here](https://github.com/buildkite/agent/blob/main/clicommand/agent_start.go#L519-L538)

moskyb (commented):
one quibble about flag naming, but other than that this looks great! i love promoting experiments

clicommand/bootstrap.go
triarius: Yeah, that's much better

triarius:
@moskyb I've also added a warning if the Job API is disabled to the job logs. It could end up being log spam, but if we are going to rely on the JobAPI in things like the redactor, it's important to know if the user has explicitly disabled it.

internal/job/executor.go
```diff
@@ -121,14 +121,17 @@ func (e *Executor) Run(ctx context.Context) (exitCode int) {
...
-		return 1
+	if e.JobAPI {
+		cleanup, err := e.startJobAPI()
+		if err != nil {
+			e.shell.Errorf("Error setting up Job API: %v", err)
+			return 1
+		}
+		defer cleanup()
+	} else {
+		e.shell.Warningf("The Job API has been disabled.")
```
moskyb, line 132:
```suggestion
		e.shell.Warningf("The Job API has been disabled. Features like automatic redaction of secrets and polyglot hooks will either not work or have degraded functionality")
```

moskyb (approved):
lgtm! one note on wording, but everything looks good

DrJosh9000 (approved):
🥳

PR #2648, opened 2024-02-21
Add explicit queue flag to the agent

moskyb:
### Description

which queue the agent listens for work on is an important part of the agent's lifecycle, especially with queues being made more explicit as part of our work to make clusters the default in buildkite. however, getting the agent to listen on a particular queue is a bit of an odd experience, it's a super special option in the `--tags` flag
```
buildkite-agent start /* ... */ --tags "queue=deploy"
```

this PR brings queue up to being a first-class citizen in the agent's CLI, giving it its own flag, so now it's `buildkite-agent start --queue deploy`. much nicer!

behind the scenes, we're still munging the queue into the tags that the agent sends to the buildkite mothership when it registers, as that's what the backend expects, but the CLI flag makes things nicer to use.

### Testing

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -112,6 +112,7 @@ type AgentStartConfig struct {
 	NoANSITimestamps bool `cli:"no-ansi-timestamps"`
 	TimestampLines   bool `cli:"timestamp-lines"`
 
+	Queue                     string   `cli:"queue"`
 	Tags                      []string `cli:"tags" normalize:"list"`
 	TagsFromEC2MetaData       bool     `cli:"tags-from-ec2-meta-data"`
 	TagsFromEC2MetaDataPaths  []string `cli:"tags-from-ec2-meta-data-paths" normalize:"list"`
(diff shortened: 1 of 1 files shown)


DrJosh9000:
What should people who are used to setting [multiple queues](https://buildkite.com/docs/agent/v3/queues#setting-an-agents-queue) do?

If someone passes both `--queue` and queue tags, this will override the first `queue=` tag, but I think any further queue tags will not be overridden?

moskyb:
yeah, i wondered about that when writing it... maybe we should just exit if both `--queue` and `--tags "queue=..."` are set?

triarius:
> What should people who are used to setting [multiple queues](https://buildkite.com/docs/agent/v3/queues#setting-an-agents-queue) do?

That whole doc page is highly misleading when it comes to cluster queues.
Things like:

> If you don't configure a queue for your agent by setting an [agent tag](https://buildkite.com/docs/agent/v3/cli-start#setting-tags) of queue=my_example_queue it will accept jobs from the default queue as if you had set queue=default.

are flat out not true for cluster queues.

IIRC we plan to deprecate it soon. Should we until then to merge this?

moskyb:
> Should we [wait] until then to merge this?

i think this is probably fine to merge now, and then later on down the track we can crack down on multiple queues. i think it'd probably have to be a major version, and we'd have to have everybody on clusters, so i'm keen to bias for action here

triarius (approved):

DrJosh9000 (approved):

triarius (approved):

PR #2650, opened 2024-02-22
Chill out credential helper logging

moskyb:
### Description

This PR makes two changes to the git credential helper:
- It makes it only respond to the `get` git credential helper action, reducing the number of calls we have to do to the backend - previously, we were still pinging the buildkite backend when git asked us to store and erase credentials, which this helper isn't able to do. now, it'll just exit 0 when git sends us those actions. [this is what git recommends you do](https://git-scm.com/docs/gitcredentials#Documentation/gitcredentials.txt-codegetcode:~:text=If%20it%20does%20not%20support%20the%20requested%20operation%20(e.g.%2C%20a%20read%2Donly%20store%20or%20generator)%2C%20it%20should%20silently%20ignore%20the%20request.).
- It chills out a lot of the logging, bumping everything down to `debug` level. we also include a little more detail in the debug logs, adding the credential input and the action type to the debug logs

### Context

COMP-211

### Testing

### DIFF
--- a/clicommand/git_credentials_helper.go
+++ b/clicommand/git_credentials_helper.go
@@ -31,7 +31,8 @@ if the pipeline has this feature enabled. All hosted compute jobs automatically 
 This command is intended to be used as a git credential helper, and not called directly.`
 
 type GitCredentialsHelperConfig struct {
-	JobID string `cli:"job-id" validate:"required"`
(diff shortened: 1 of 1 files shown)


triarius (approved):
LGTM ⛔🎉

PR #2654, opened 2024-02-25
Extend graceful cancellation to all job phases

david-poirier:
### Description

We want to be able to gracefully cancel all phases of a job.

### Context

The graceful job cancellation feature introduced [here](https://github.com/buildkite/agent/pull/2250) is great but only applies to the checkout phase of a job.

### Changes

- Use a cancellable context in all phases where processes are being run
- Pass `BUILDKITE_SIGNAL_GRACE_PERIOD_SECONDS` between the agent and executor

### Testing

### DIFF
--- a/internal/job/executor.go
+++ b/internal/job/executor.go
@@ -70,6 +70,10 @@ func New(conf ExecutorConfig) *Executor {
 
 // Run the job and return the exit code
 func (e *Executor) Run(ctx context.Context) (exitCode int) {
+	// Create a context to use for cancelation of the job
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
 	// Check if not nil to allow for tests to overwrite shell
 	if e.shell == nil {
 		var err error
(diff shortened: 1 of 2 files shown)


internal/job/executor.go
```diff
@@ -100,7 +100,7 @@ func (e *Executor) Run(ctx context.Context) (exitCode int) {
 	defer stopper()
 	defer func() { span.FinishWithError(err) }()
 
-	// Create a context to use for cancelation of the job
+	// Create a context to use for cancellation of the job
 	cancelCtx, cancel := context.WithCancel(ctx)
```
moskyb, line 104: given that this is now the default context for the executor, i think we should just call it `ctx` - it encourages people to use this one as it seems like the default

internal/job/executor.go
```diff
@@ -134,7 +133,7 @@ func (e *Executor) Run(ctx context.Context) (exitCode int) {
 
 	// Tear down the environment (and fire pre-exit hook) before we exit
 	defer func() {
-		if err = e.tearDown(ctx); err != nil {
+		if err = e.tearDown(cancelCtx); err != nil {
```
moskyb, line 136: this is the exception to my comment above - `tearDown` still happens after a job is cancelled, and so shouldn't be called with a cancellable context.

i think we'll need to rename the context before the call to `context.WithCancel` to something like `nonCancelContext`, and then pass the non-cancel context into `tearDown`

moskyb (changes requested):
hey! i just wanna say this is awesome work. as i mentioned on slack, cancellation is super complicated - much more complicated than the diff here might suggest, to anyone passing by. this looks really good.

i have a couple of suggested changes, but after those are in i'd be happy to merge this - one of the changes should fix your test issue.

internal/job/executor.go
```diff
@@ -96,23 +96,22 @@ func (e *Executor) Run(ctx context.Context) (exitCode int) {
...
 	// Create a context to use for cancelation of the job
-	cancelCtx, cancel := context.WithCancel(ctx)
+	cancelCtx, cancel := context.WithCancel(nonCancelCtx)
 	defer cancel()
 
 	// Listen for cancellation
 	go func() {
 		select {
-		case <-ctx.Done():
+		case <-nonCancelCtx.Done():
```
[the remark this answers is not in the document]
DrJosh9000: I mentioned to @moskyb that the `select` could be simplified away, since `<-nonCancelCtx.Done()` won't ever happen, looks like they did that

DrJosh9000 (approved):
Makes sense to me! 🥬

moskyb (approved):

PR #2655, opened 2024-02-26
Force GitHub URLs to use HTTPS if the agent's git-credential-helper if it is enabled

triarius:
### Description
When using buildkite's git-credential-helper, the agent may still attempt to checkout with ssh. But users who use the git-credential-helper would have set up the Github App that provides code access. This only works with https URLs. So, here, we force git to use https URLs instead of the ssh ones. We also disable the ssh key scan in this case.

I've limited this to all of GitHub. We may expand it for other SCM providers in the future.

There's a risk that this will break existing setups of self-hosted agents that already have a ssh key that works for some repos that the GitHub App with code access has not been set up for. As the git-credential-helper is not supported on self-hosted agents at the moment, and this circumstance is expected to be rare, we think that risk is acceptable. Users that do encounter this should just esure that the code access app has been configured with ALL the repositories that need to be checked out, including submodules.

### Context
https://linear.app/buildkite/issue/COMP-216/rewrite-ssh-to-https-on-agent
https://buildkite-corp.slack.com/archives/C06D36S59N1/p1708899990413789

### Changes
When the git credential helper is enabled, we run
```
git config --global url."https://github.com/".insteadOf git@github.com:
```

### Testing

### DIFF
--- a/internal/job/checkout.go
+++ b/internal/job/checkout.go
@@ -42,6 +42,12 @@ func (e *Executor) configureGitCredentialHelper(ctx context.Context) error {
 		return fmt.Errorf("configuring git credential.helper: %v", err)
 	}
 
+	// so that the new credential helper will be used for all github urls
+	err = e.shell.RunWithoutPrompt(ctx, "git", "config", "--global", "url.https://github.com/.insteadOf", "git@github.com:")
+	if err != nil {
+		return fmt.Errorf("enabling git rewrite: %v", err)
+	}
+
 	return nil
 }
 
--- a/internal/job/checkout.go
+++ b/internal/job/checkout.go
@@ -42,13 +42,15 @@ func (e *Executor) configureGitCredentialHelper(ctx context.Context) error {
 		return fmt.Errorf("configuring git credential.helper: %v", err)
 	}
 
-	// so that the new credential helper will be used for all github urls
-	err = e.shell.RunWithoutPrompt(ctx, "git", "config", "--global", "url.https://github.com/.insteadOf", "git@github.com:")
-	if err != nil {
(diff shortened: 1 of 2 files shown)


internal/job/checkout.go
```diff
@@ -42,6 +42,12 @@ func (e *Executor) configureGitCredentialHelper(ctx context.Context) error {
 		return fmt.Errorf("configuring git credential.helper: %v", err)
 	}
 
+	// so that the new credential helper will be used for all github urls
+	err = e.shell.RunWithoutPrompt(ctx, "git", "config", "--global", "url.https://github.com/.insteadOf", "git@github.com:")
+	if err != nil {
+		return fmt.Errorf("enabling git rewrite: %v", err)
+	}
+
```
moskyb, line 50: Should we add something like

```suggestion
	// Don't ssh keyscan - we won't be using ssh anyway.
	e.SSHKeyscan = false
	
	// so that the new credential helper will be used for all github urls
	err = e.shell.RunWithoutPrompt(ctx, "git", "config", "--global", "url.https://github.com/.insteadOf", "git@github.com:")
	if err != nil {
		return fmt.Errorf("enabling git rewrite: %v", err)
	}

```

it means that users running on an instance without ssh installed (which might be compute instances?) won't see errors, and we know that YAGNI at this point

triarius: Maybe, but then on self-hosted, if there is an SSH key for a non-GitHub SCM, and this is the first checkout of that repo on that machine, then we would fail with an unauthorized host.

I think the self-hosted situation will require significantly more thought. So i'm going to scatter some to-do comments throughout.

triarius: Ok done.

moskyb (approved):

PR #2669, opened 2024-03-05
Promote `avoid-recursive-trap` experiment

triarius:
### Description

This has been operating for some time in our dogfood stacks and we have not
noticed any issues. It is time to promote experiment. I have not added any
code to invert the experiment, as I believe it is safe to promote without
that.

### Context

https://3.basecamp.com/3453178/buckets/27608512/messages/6299400220

### Testing

### DIFF
--- a/internal/experiments/experiments.go
+++ b/internal/experiments/experiments.go
@@ -30,37 +30,37 @@ const (
 	PTYRaw                     = "pty-raw"
 	PolyglotHooks              = "polyglot-hooks"
 	ResolveCommitAfterCheckout = "resolve-commit-after-checkout"
-	AvoidRecursiveTrap         = "avoid-recursive-trap"
 	IsolatedPluginCheckout     = "isolated-plugin-checkout"
 	UseZZGlob                  = "use-zzglob"
 
 	// Promoted experiments
-	ANSITimestamps    = "ansi-timestamps"
-	FlockFileLocks    = "flock-file-locks"
-	GitMirrors        = "git-mirrors"
-	InbuiltStatusPage = "inbuilt-status-page"
-	JobAPI            = "job-api"
+	ANSITimestamps     = "ansi-timestamps"
+	FlockFileLocks     = "flock-file-locks"
+	GitMirrors         = "git-mirrors"
+	InbuiltStatusPage  = "inbuilt-status-page"
+	JobAPI             = "job-api"
+	AvoidRecursiveTrap = "avoid-recursive-trap"
 )
 
 var (
 	Available = map[string]struct{}{
 		AgentAPI:                   {},
 		DescendingSpawnPrioity:     {},
+		IsolatedPluginCheckout:     {},
 		KubernetesExec:             {},
 		NormalisedUploadPaths:      {},
 		PolyglotHooks:              {},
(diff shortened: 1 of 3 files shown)


internal/experiments/experiments.go
```diff
@@ -30,16 +30,16 @@ const (
...
-	FlockFileLocks    = "flock-file-locks"
-	GitMirrors        = "git-mirrors"
-	InbuiltStatusPage = "inbuilt-status-page"
-	JobAPI            = "job-api"
+	ANSITimestamps     = "ansi-timestamps"
+	FlockFileLocks     = "flock-file-locks"
+	GitMirrors         = "git-mirrors"
+	InbuiltStatusPage  = "inbuilt-status-page"
+	JobAPI             = "job-api"
+	AvoidRecursiveTrap = "avoid-recursive-trap"
```
DrJosh9000, line 42: why no lexicographic ordering

internal/job/executor.go
```diff
@@ -1032,8 +1007,7 @@ func (e *Executor) defaultCommandPhase(ctx context.Context) error {
 		e.shell.Promptf("%s", cmdToExec)
 	}
 
-	err = e.shell.RunWithoutPrompt(ctx, cmd[0], cmd[1:]...)
-	return err
+	return e.shell.RunWithoutPrompt(ctx, cmd[0], cmd[1:]...)
```
DrJosh9000, line 1010: This change breaks the `defer` near the start of the function (introduced in 3eaf791):

```go
	var err error
```

If we want to do this cleanup, we'd need to make `err` a named return instead.

triarius: I think the same thing will happen here: https://github.com/buildkite/agent/blob/745645930b7d322eb9fc3660a0bb52b9edc945cd/internal/job/executor.go#L683 and possibly several other places yeah?

I think using named return values is an option to fix it: https://goplay.tools/snippet/yHRT8QgN8BG. Then this code can stay as is.

internal/experiments/experiments.go
```diff
@@ -25,42 +25,42 @@ const (
 	// Available experiments
 	AgentAPI                   = "agent-api"
 	DescendingSpawnPrioity     = "descending-spawn-priority"
+	IsolatedPluginCheckout     = "isolated-plugin-checkout"
```
DrJosh9000, line 28: thank you 🙏

internal/job/executor.go
triarius: And I think this is the relevant part of the spec:

> A "return" statement that specifies results sets the result parameters before any deferred functions are executed.

which appears under: https://go.dev/ref/spec#ReturnStmt

DrJosh9000 (approved):
LGTM 🌿

internal/job/executor.go
DrJosh9000: Yep, I think we're on the same page there.

torarnv:
Has the experiment been promoted? I still see the `trap` unless I run with `avoid-recursive-trap` ?

torarnv:
Never mind, I upgraded the agent 3.59.0 -> 3.69.0 and now it's on by default. Sorry :)

PR #2670, opened 2024-03-05
Change annotation priority default from 3 to 1

matthewborden:
Currently annotation priority defaults are set server side, this means that agent version doesn't currently control the default (even if they mention it in the help text). This means that we're not able to pin a default annotation priority to an agent version.

This updates the annotation priority to be set on the agent, not the server. It also changes the default from 3 to 1 because priority of 3 has been considered to be confusing and not intuitive. This now matches agent priority.

Agent version v3.65.0 will continue to use the server side default (which will be updated to 1).

### DIFF
--- a/clicommand/annotate.go
+++ b/clicommand/annotate.go
@@ -97,8 +97,9 @@ var AnnotateCommand = cli.Command{
 		},
 		cli.IntFlag{
 			Name:   "priority",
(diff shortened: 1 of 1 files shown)


matthewborden:
Closing in favour of https://github.com/buildkite/agent/pull/2934#pullrequestreview-2244500639

PR #2671, opened 2024-03-05
Set a user agent when downloading most artifacts

yob:
This downloader tooling was using the default golang user agent which works fine, but it's nice to indicate to the servers what the client is with a little more detail.

We already have a helper for generating a user agent string on API calls to Buildkite, so I've used it here too. This helper is used for artifatc downloads from:

* S3 (AWS)
* GCS (Google)
* Artifactory

It is not used for artifact downloads from Azure blob store.

I'm only setting the user agent if none is provided to the downloader, but currently none of the places its used pass one in.

### Testing


moskyb (approved):
LGTM! 🎉

PR #2672, opened 2024-03-05
Add a User-Agent header when uploading artifacts to Buildkite's default location

yob:
If the agent isn't configured with a custom artifact storage location, it defaults to uploading them to an S3 bucket provided by Buildkite. The upload is performed by FormUploader, which wasn't setting a User-Agent header.

This PR adds the header, using the same tooling we have for setting User-Agent on requests to the Buildkite API.

The header is primarily for debugging. It is occasionally helpful for us to see which version of the agent was used for particular uploaded artifacts.

I have not attempted to set the User-Agent header on uploads when a custom artifact storage location (S3, GCS, Artifactory, etc) is in use. That might be a useful thing to do, but we can consider it separately.

Fixes PLT-2222

### Testing

### DIFF
--- a/agent/form_uploader.go
+++ b/agent/form_uploader.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/buildkite/agent/v3/api"
 	"github.com/buildkite/agent/v3/logger"
+	"github.com/buildkite/agent/v3/version"
 )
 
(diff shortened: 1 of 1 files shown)


triarius (approved):
LGTM

PR #2673, opened 2024-03-05
Downgrade "submodules disabled" message to info

bpoland:
### Description

- When submodules have been disabled intentionally, we don't want the group to auto-expand in the Buildkite UI (which happens when there is a warning logged)

### Context

- As discussed on Slack


bpoland:
Replaced by https://github.com/buildkite/agent/pull/2674

Issue #1773, opened 2022-09-28
disabling submodules should not print a warning


zach-sherman:
**Is your feature request related to a problem? Please describe.**
We have submodules disabled at the agent level, so on every build it prints a warning causing the "Preparing working directory" drawer to be expanded. This is unnecessarily noisy and causes extra mental load.

**Describe the solution you'd like**
Don't print a warning if submodules are disabled, or provide a config option to disable the warning. If users are explicitly disabling submodules, then they shouldn't need to be warned that, in fact, submodules _are_ disabled. The current behavior seems to presume that the agent-configuror has disabled submodules, but a someone working on a pipeline might expect them to be present? Either way I think that situation is not worth warning over since it is totally valid and intended behavior. Seems like it should be info-level.

**Describe alternatives you've considered**
Nothing we can really do about this output since it's hard-coded in the agent.

**Additional context**

pda:
Hey @zach-sherman 👋🏼

So you're running `buildkite-agent start` with [`--no-git-submodules`](https://github.com/buildkite/agent/blob/6fec074dce224d039821b3d650eb3f4062e69b24/clicommand/agent_start.go#L458-L462) to prevent the agent from [attempting to clone submodules when it finds them](https://github.com/buildkite/agent/blob/6fec074dce224d039821b3d650eb3f4062e69b24/bootstrap/bootstrap.go#L1376-L1415). I assume your repository has submodules, and you can't or don't want to remove them. So the repo is the way you want it, and the agent is configured the way you want it, but this zero-value-to-you warning is making the job log view much less usable by expanding the rather noisy “Preparing working directory” section.

That sounds frustrating.

That said, I think this warning:

> This repository has submodules, but submodules are disabled at an agent level

… is actually quite important to visibly surface. I can imagine somebody just spent however long creating a new pipeline, writing their YAML, getting their first build running, and then wondering why it's failing because half the files are missing. This warning helps them realise that the other team running the agents has configured them to disallow submodules, and that's the clue they need to get unblocked; speaking to that team, or removing submodules, or working around the submodules not being present in CI.

I wonder if there's a way we could keep the warning visible without expanding out the entire “Preparing working directory” section. Maybe giving it its own one-line section after that? Would that solve your problem (giant noisy section expanding), or would it still be frustrating (warning still visible)?

[...]

PR #2674, opened 2024-03-06
Configurably optional warnings

moskyb:
### Description

there's a subset of warnings in the agent that are essentially us trying to clippy the users, saying "Hmm, something about this setup looks a little bit weird, but we'll proceed anyway" - notably a message about when [submodules are disabled, but the repo the agent is cloning containing submodules](https://github.com/buildkite/agent/issues/1773).

he somewhat frequently get feedback from users that when you know that you're doing the right thing - for example, if you know you're not going to need the submodules that aren't being cloned - these warnings are annoying, and pop open a large log group in the job log that isn't useful.

however, we're usually a bit reluctant to add flags to disable individual warnings - for example, a hypothetical `--disable-warning-about-submodules`, because the agent arguably has too many flags already.

this PR is a bit of a middle ground - certain warnings will become optional by using the `--no-warn-for` flag. Each optional warning is given an ID, and when not disabled, will have  "You can disable this warning by passing the `--no-warn-for <some warning name>`" appended to the end.

If users want to disable that warning - for example, the submodule warning mentioned above, `submodule-conflict` - they can pass `--no-warn-for submodule-conflict` to their `buildkite-agent start` command.

Thanks so much to @amu-g for pairing with me on this!

### Testing

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -99,7 +99,7 @@ type AgentStartConfig struct {
 
 	LogFormat            string   `cli:"log-format"`
 	WriteJobLogsToStdout bool     `cli:"write-job-logs-to-stdout"`
-	NoWarnFor            []string `cli:"no-warn-for" normalize:"list"`
+	DisableWarningsFor   []string `cli:"disable-warnings-for" normalize:"list"`
 
 	BuildPath   string `cli:"build-path" normalize:"filepath" validate:"required"`
 	HooksPath   string `cli:"hooks-path" normalize:"filepath"`
@@ -655,9 +655,9 @@ var AgentStartCommand = cli.Command{
 			EnvVar: "BUILDKITE_AGENT_JOB_VERIFICATION_NO_SIGNATURE_BEHAVIOR",
 		},
 		cli.StringSliceFlag{
-			Name:   "no-warn-for",
+			Name:   "disable-warnings-for",
 			Usage:  "A list of warning IDs to disable",
-			EnvVar: "BUILDKITE_AGENT_NO_WARN_FOR",
+			EnvVar: "BUILDKITE_AGENT_DISABLE_WARNINGS_FOR",
 		},
 
 		// API Flags
@@ -939,7 +939,7 @@ var AgentStartCommand = cli.Command{
 
 			VerificationJWKS: verificationJWKS,
 
-			NoWarnFor: cfg.NoWarnFor,
+			DisableWarningsFor: cfg.DisableWarningsFor,
 		}
 
 		if configFile != nil {
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -97,8 +97,9 @@ type AgentStartConfig struct {
 	EnableJobLogTmpfile bool   `cli:"enable-job-log-tmpfile"`
 	JobLogPath          string `cli:"job-log-path" normalize:"filepath"`
 
-	LogFormat            string `cli:"log-format"`
-	WriteJobLogsToStdout bool   `cli:"write-job-logs-to-stdout"`
+	LogFormat            string   `cli:"log-format"`
+	WriteJobLogsToStdout bool     `cli:"write-job-logs-to-stdout"`
+	NoWarnFor            []string `cli:"no-warn-for" normalize:"list"`
 
 	BuildPath   string `cli:"build-path" normalize:"filepath" validate:"required"`
 	HooksPath   string `cli:"hooks-path" normalize:"filepath"`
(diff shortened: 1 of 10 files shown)


internal/job/shell/logger.go
```diff
@@ -91,6 +96,15 @@ func (wl *WriterLogger) Warningf(format string, v ...any) {
...
 
+func (wl *WriterLogger) OptionalWarningf(id, format string, v ...any) {
+	if slices.Contains(wl.DisabledWarningIDs, id) {
+		return
+	}
+
+	warningFormatWithDisable := format + fmt.Sprintf(". You can disable this warning by passing the `--no-warn-for %s` flag", id)
+	wl.Warningf(warningFormatWithDisable, v...)
+}
+
```
moskyb, line 107: one weakness of this approach is that the values of optional warnings aren't easily discoverable, and could potentially intersect.

one idea we had for this was to make a warning id type and use that instead, a la
```Go
type optionalWarningID int
const (
  JobAPIDisabled optionalWarningID = iota
  SubmoduleConflict
  // ...etc
)
```

and then make this function take an `optionalWarningID` for the `id` param rather than a string. the reason that we ended up going with this approach is that _which_ warnings are ignored are more a concern of the job executor than they are of this module, but that decision isn't without its drawbacks.

WDYT?

clicommand/agent_start.go
```diff
@@ -653,6 +654,11 @@ var AgentStartCommand = cli.Command{
 			Usage:  fmt.Sprintf("The behavior when a job is received without a signature. One of: %v. Defaults to %s", verificationFailureBehaviors, agent.VerificationBehaviourBlock),
 			EnvVar: "BUILDKITE_AGENT_JOB_VERIFICATION_NO_SIGNATURE_BEHAVIOR",
 		},
+		cli.StringSliceFlag{
+			Name:   "no-warn-for",
```
dabarrell, line 658: I can't review the go, but `no-warn-for` feels weird - grammatically incorrect etc. Seems we're okay with slightly longer flags if it means they read more easily. How about `disable-warnings-for`? to match the language you're using in the usage string, and the code (e.g. `DisabledWarningIDs`)

internal/job/shell/logger.go
```diff
@@ -91,6 +96,15 @@ func (wl *WriterLogger) Warningf(format string, v ...any) {
 	wl.Printf("^^^ +++")
 }
 
+func (wl *WriterLogger) OptionalWarningf(id, format string, v ...any) {
+	if slices.Contains(wl.DisabledWarningIDs, id) {
+		return
```
dabarrell, line 101: This is what I came to check - whether or not you downgrade to `info` level log line, or drop it entirely. Perhaps a `debug` line is a good middle ground? discoverable if debug is enabled, otherwise not?

internal/job/shell/logger.go
DrJosh9000: How likely is it that someone adding a new optional warning is going to accidentally re-use an existing ID?

clicommand/agent_start.go
moskyb: yeah, that feels way better

internal/job/shell/logger.go
moskyb: ah, logs don't quite work that way (with levels etc) in the shell, they're either there or they're not - i think we should probably downgrade it to just a call to `wl.Printf` though.

dabarrell: Ah, okay. Yeah, I think printing something is better than nothing, and if that itself proves to be a pain point for people, we can evaluate dropping them entirely then

internal/job/shell/logger.go
moskyb: relatively unlikely, i think, and i think the impact would be relatively low? ie they'd hide more warnings than they were intending to

clicommand/bootstrap.go
```diff
@@ -362,6 +363,11 @@ var BootstrapCommand = cli.Command{
 			Usage:  "Disables the Job API, which gives commands in jobs some abilities to introspect and mutate the state of the job.",
 			EnvVar: "BUILDKITE_AGENT_NO_JOB_API",
 		},
+		cli.StringSliceFlag{
+			Name:   "disable-warnings-for",
+			Usage:  "A list of warning IDs to disable",
+			EnvVar: "BUILDKITE_AGENT_NO_WARN_FOR",
```
bpoland, line 369: Should the env var be renamed too? It's certainly longer but hat seems less confusing as a user if the CLI arg and env var match
```suggestion
			EnvVar: "BUILDKITE_AGENT_DISABLE_WARNINGS_FOR",
```

moskyb: yes def! good point, forgot that in my find-and-replace spree 😅

bpoland (approved):
This looks great to me! Do the docs on your site get generated automatically after this is merged?

HugeIRL:
> This looks great to me! Do the docs on your site get generated automatically after this is merged?

Hey @bpoland,

The docs may pick up some of the changes, however us from Support will make sure that if anything is missing/not clear, we can jump on that to fix it.

DrJosh9000 (approved):
Inventory Get: Green Tick of Approval ✅

PR #2675, opened 2024-03-07
Handle warnings from go-pipeline `Parse`

DrJosh9000:
### Description

* Update go-pipeline to v0.5.0
* Display new parser warnings produced by go-pipeline

### Context

https://linear.app/buildkite/issue/PS-46/json-valued-env-vars-in-pipelineyml-silently-dropped

### Testing

### DIFF
--- a/clicommand/pipeline_upload.go
+++ b/clicommand/pipeline_upload.go
@@ -264,7 +264,7 @@ var PipelineUploadCommand = cli.Command{
 
 		result, err := cfg.parseAndInterpolate(src, input, environ)
 		if w := warning.As(err); w != nil {
-			l.Warn("There were some issues with the pipeline input:\n%v", w)
+			l.Warn("There were some issues with the pipeline input - pipeline upload will proceed, but might not succeed:\n%v", w)
 		} else if err != nil {
 			return err
 		}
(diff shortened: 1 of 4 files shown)


clicommand/pipeline_upload.go
```diff
@@ -262,7 +263,9 @@ var PipelineUploadCommand = cli.Command{
 		}
 
 		result, err := cfg.parseAndInterpolate(src, input, environ)
-		if err != nil {
+		if w := warning.As(err); w != nil {
+			l.Warn("There were some issues with the pipeline input:\n%v", w)
```
moskyb, line 267: is it worth adding something like `pipeline upload will proceed, but may not succeed`, or something?

clicommand/tool_sign.go
```diff
@@ -186,7 +187,9 @@ func signOffline(
 	}
 
 	parsedPipeline, err := pipeline.Parse(input)
-	if err != nil {
+	if w := warning.As(err); w != nil {
+		l.Warn("There were some issues with the pipeline input:\n%v", w)
```
moskyb, line 191: as above

tessereth:
See https://github.com/buildkite/go-pipeline/pull/27

tessereth (approved):
It would be neat to have a test for this. But maybe that's hard. So LGTM regardless.

PR #2676, opened 2024-03-08
Always set git commit

moskyb:
### Description

As a part of the checkout process, the agent sets a special metadata key called `buildkite:git:commit` to a string containing information about the commit that the agent has checked out. This metadata is used by the buildkite backend to populate VCS-related information on the build, like the commit message, author and commit hash - highlighted below:
CleanShot 2024-03-08 at 13 53 57@2x

(note that in some cases this information can come from webhooks when builds are triggered that way, but not all builds are triggered by webhooks)

this information is also used set the commit used for the rest of the build; often a build will get triggered at the HEAD of a specific branch, but the HEAD of a branch will change if extra commits are added over time, so the agent needs to know the "canonical" commit for the build, which is stored in this metadata value.

However, this metadata is only set during the default checkout phase - if a user overrides the checkout phase using agent hooks or plugins, this special metadata value will never get set, and it can lead to a build running across multiple commits by accident if the checkout hook is overridden.

**This PR** hoists the commit metadata setting process up a level in the checkout process, and it will now happen after _any_ checkout hook, not just the default one.

### Context

[Coda link](https://coda.io/d/Escalations-Feedback_dHnUHNps1YO/Pipelines-Escalations_su7FT#Pipelines-Escalations-Board_tu__K/r297)

### Testing

### DIFF
--- a/internal/job/checkout.go
+++ b/internal/job/checkout.go
@@ -219,7 +219,7 @@ func (e *Executor) CheckoutPhase(ctx context.Context) error {
 
 	err = e.sendCommitToBuildkite(ctx)
 	if err != nil {
-		return err
+		e.shell.OptionalWarningf("commit-metadata", err.Error())
 	}
 
 	// Store the current value of BUILDKITE_BUILD_CHECKOUT_PATH, so we can detect if
--- a/internal/job/checkout.go
+++ b/internal/job/checkout.go
@@ -219,7 +219,7 @@ func (e *Executor) CheckoutPhase(ctx context.Context) error {
 
 	err = e.sendCommitToBuildkite(ctx)
 	if err != nil {
-		e.shell.OptionalWarningf("commit-metadata", err.Error())
+		e.shell.OptionalWarningf("git-commit-resolution-failed", err.Error())
 	}
 
 	// Store the current value of BUILDKITE_BUILD_CHECKOUT_PATH, so we can detect if
--- a/internal/job/checkout.go
+++ b/internal/job/checkout.go
@@ -726,7 +726,7 @@ func (e *Executor) defaultCheckoutPhase(ctx context.Context) error {
 	return nil
 }
 
-const commitMetadataKey = "buildkite:git:commit"
+const CommitMetadataKey = "buildkite:git:commit"
 
 // sendCommitToBuildkite sends commit information (commit, author, subject, body) to Buildkite, as the BK backend doesn't
 // have access to user's VCSes. To do this, we set a special meta-data key in the build, but only if it isn't already present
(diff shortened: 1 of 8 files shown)


internal/job/checkout.go
```diff
@@ -217,6 +217,11 @@ func (e *Executor) CheckoutPhase(ctx context.Context) error {
 		}
 	}
 
+	err = e.sendCommitToBuildkite(ctx)
+	if err != nil {
+		return err
```
dabarrell, line 222: Are you satisfied that this won't be a breaking change for anyone? I wonder if there are customers out there who have modified their checkout phase to such an extent that this won't resolve – in which case logging this as a warning rather than erroring out might be a good idea? I'm not sure - just want to make sure we cover the bases

triarius (approved):

internal/job/checkout.go
```diff
@@ -217,6 +217,11 @@ func (e *Executor) CheckoutPhase(ctx context.Context) error {
 		}
 	}
 
+	err = e.sendCommitToBuildkite(ctx)
+	if err != nil {
+		e.shell.OptionalWarningf("commit-metadata", err.Error())
```
DrJosh9000, line 222: Maybe don't use an error string in the format arg (e.g. imagine if `err.Error()` returns a string with `%s` in it)...

```suggestion
		  e.shell.OptionalWarningf("commit-metadata", "Couldn't %v", err)
```

(Adjust format string or error strings as necessary.)

moskyb: oooh yeah good call. it felt yucky doing it, i'll fix it up

dabarrell, line 222: `commit-metadata` doesn't really feel like it represents what this is - perhaps. Can we fit the word `resolve` or something in there? `resolve-git-commit` maybe?

dabarrell: To create an actual suggestion - had a look at some other equivalent strings, and they describe the thing that happened (`-unavailable`, `-disabled` etc)
```suggestion
		e.shell.OptionalWarningf("git-commit-resolution-failed", err.Error())
```

DrJosh9000: ?

moskyb: ah shoot, i had this committed locally, but forgot to push it 🤦 i'll make a new PR.

moskyb: https://github.com/buildkite/agent/pull/2699

jradtilbrook:
FYI there is a forum post about this functionality that has broken it. I believe its probably broken for the customer in a way that highlights the exact point this is trying to fix. But looping it in here in case other things come out as well https://forum.buildkite.community/t/agent-3-67-0-has-incorrect-pipeline-headings/3651

PR #2677, opened 2024-03-10
[Update] Updating `tool sign` usage description to match actual command

CheeseStick:
### Description

For the pipeline signing, `buildkite-agent` uses `buildkite-agent tool sign.` However, the command's usage description prints `buildkite-agent tool sign-pipeline,` which can confuse people when there is an error.

### Context

Screenshot 2024-03-11 at 11 40 46 AM

Like the above screenshot, it's something minor but can be pretty confusing. 😅 

### Changes

This PR updates the usage text in the code to match the actual command.


triarius (approved):
Whoops! Thanks for catching this.

PR #2688, opened 2024-03-12
Add entry in PR description asking Pull Requesters to document any new commands they add

moskyb:
### Description

The current (internal) guidance is that when a buildkite employee makes a new release of the agent, they open a PR against the [docs](https://github.com/buildkite/docs) that includes updates in CLI reference there, which is kinda-sorta autogenerated from the helptexts that we write for commands in this repo.

the issue with this process is that the process for adding a new page for a new agent cli command is a bit involved, and often it leaves the person who made the release in charge of adding docs for a command they didn't write, which is a minor pain.

this PR adds another couple of checkboxes to the PR description (the one i'm typing in right now!) asking that if the user has added a command, that they open a PR against the docs repo. i also added a checkbox for "updated the readme as necessary", because the other checkbox was looking lonely on its own

### Context

https://github.com/buildkite/docs/pull/2710

### Testing

### DIFF
--- a/docs/updating-docs.md
+++ b/docs/updating-docs.md
@@ -1,0 +1,64 @@
+<!-- This doc is copied from the Buildkite engineering handbook, and is copied here so that it can be consumed by the public -->
+
+# Updating Agent Docs
+
+The public documentation for the agent is hosted at https://buildkite.com/docs/agent/v3. The source code is in https://github.com/buildkite/docs. This document is our guide on how to update the public docs as we develop the agent.
+
+## New sub command
+
+Creating a new sub command for the agent requires touching a few different files. In this example, our new sub command will be called `foo` and will be invoked as
+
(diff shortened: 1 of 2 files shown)


docs/updating-docs.md
triarius, line 1: The docs repo is public too yeah? I think this document belongs there as most of the commands assume the working directory is a docs repo.

```
1. Clone or change direcotry to the docs repo: https://github.com/buildkite/docs
2. Follow the instructions in https://github.com/buildkite/docs?tab=readme-ov-file#agent
```

I suggested putting it in an agent section in the README as there is no `docs` directory in the `docs` repo for documentation on the documentation! You could create one if you want. I think it's a good precedent to set.

DrJosh9000:
Closing due to staleness

PR #2689, opened 2024-03-12
Use roko.DoFunc

DrJosh9000:
### Description

Take advantage of recently-shipped improvement in roko.

### Context

None, this is an opportunistic improvement.

### Changes

Replace `Do` and `DoWithContext` with `DoFunc` or `DoFunc2`, where the intent of the code being retried is to produce some value.

### Testing


moskyb (approved):
yessssssssss 👏 it's so good

PR #2693, opened 2024-03-19
Update outdated option name

fruechel-canva:
### Description

The `job-` prefix was used during development/beta but when it was made stable it was dropped and the name changed. This message was just not updated then.

### Testing


DrJosh9000 (approved):

DrJosh9000:
Good catch, thanks for the fix!

PR #2694, opened 2024-03-20
De-experiment isolated plugin checkout

triarius:
### Description
Isolated plugin checkout was an experiment that ensures each agent worker checks out plugins into a separate directory.
The experiment was made because the previous behavior of sharing the a plugin's directory between agents works was not locking the checkouts correctly.
If two agent workers run two different version of a plugin concurrenlty, the worker that checked out last would control the version of the plugin used by both agent workers.

A solution involving shared and exclusive locks was proposed at the time, but as plugins are usually a small collection of bash scripts, the complexity of the solution was not worth the benefit.

### Context
https://github.com/buildkite/agent/pull/2301
https://coda.io/d/Escalations-Feedback_dHnUHNps1YO/Escalations_sudOj#Escalations-Table_tuS42/r22&view=modal

### Changes
Promote the isolated plugin checkout experiment so that the only behavior is that each work has an independent copy of every plugin.

### Testing


internal/experiments/experiments.go
```diff
@@ -55,12 +54,13 @@ var (
...
-		GitMirrors:         standardPromotionMsg(GitMirrors, "v3.47.0"),
-		InbuiltStatusPage:  standardPromotionMsg(InbuiltStatusPage, "v3.48.0"),
-		JobAPI:             standardPromotionMsg(JobAPI, "v3.64.0"),
+		ANSITimestamps:         standardPromotionMsg(ANSITimestamps, "v3.48.0"),
+		AvoidRecursiveTrap:     standardPromotionMsg(AvoidRecursiveTrap, "v3.66.0"),
+		FlockFileLocks:         standardPromotionMsg(FlockFileLocks, "v3.48.0"),
+		GitMirrors:             standardPromotionMsg(GitMirrors, "v3.47.0"),
+		InbuiltStatusPage:      standardPromotionMsg(InbuiltStatusPage, "v3.48.0"),
+		IsolatedPluginCheckout: standardPromotionMsg(IsolatedPluginCheckout, "v3.67.0"),
+		JobAPI:                 standardPromotionMsg(JobAPI, "v3.64.0"),
```
moskyb, line 63: should we sort these by version number instead of alphabetically?

triarius: I think that's a good idea, but enforcing it will not be straight forward at the moment. Always happy to review a PR by you that does it.

DrJosh9000: Why by version number? Not really seeing a reason

DrJosh9000 (approved):

internal/experiments/experiments.go
moskyb: eh, i was thinking to sort by recency? but also it really doesn't matter that much, i think

PR #2695, opened 2024-03-20
Even more silent Job API

triarius:
### Description
I've made a few PRs to reduce the amount of logging that the job api does. But I kept missing a few places. This PR should fix that, and for good measure I've added a test to make sure that the job api doesn't log anything unless in debug mode. To test that the tests work, the commit that introduces the tests should fail and that failure should be fixed by the next commit.

In particular, it should stop printing the message that the Job API server has shut down. This was difficult to see if log headers (e.g. `+++`, `~~~`, etc) are being used in a job as it's often hidden in the last log group. But it's still an annoyance if log groups are not being used.

### Context
https://github.com/buildkite/agent/pull/2690
https://github.com/buildkite/agent/pull/2662

### Changes
- The Job API should no longer print *anything* to the job logs unless in debug mode.

### Testing

### DIFF
--- a/jobapi/server_test.go
+++ b/jobapi/server_test.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


jobapi/server_test.go
```diff
@@ -454,3 +455,92 @@ func TestCreateRedaction(t *testing.T) {
...
+	req, err := http.NewRequest(http.MethodGet, "http://job/api/current-job/v0/env", buf)
+	assert.NilError(t, err)
+
+	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", token))
+	testAPI(t, env, req, client, tc)
+
+	assert.NilError(t, srv.Stop())
+
+	logs := logBuf.String()
+	assert.Check(t, strings.Contains(logs, "~~~ Job API"), "logs: %q", logs)
```
DrJosh9000, line 499: I think the test names need swapping?

DrJosh9000 (approved):

PR #2697, opened 2024-03-22
core-103 break from retrying if we get a 422

SorchaAbel:
### Description

If the server returns an error when uploading artifacts (i.e. 400 Bad Request) then the agent will retry 10 times. 
If the artifacts uploaded exceed the service quota the server will return a 403 and we don't want the agent to retry (link to server [PR](https://github.com/buildkite/buildkite/pull/16071))

The screenshot below is from local, the service quota for max artifacts per job is not really 2 :)

**Before changes**
Screenshot 2024-03-22 at 1 55 26 pm

**After changes**
Screenshot 2024-03-22 at 2 01 12 pm

### Context

(https://linear.app/buildkite/issue/CORE-103/artifacts-per-job-communicate-when-limit-is-exceeded)

### Testing

### DIFF
--- a/agent/artifact_batch_creator.go
+++ b/agent/artifact_batch_creator.go
@@ -87,7 +87,8 @@ func (a *ArtifactBatchCreator) Create(ctx context.Context) ([]*api.Artifact, err
 
 			creation, resp, err := a.apiClient.CreateArtifacts(ctxTimeout, a.conf.JobID, batch)
 			// the server returns a 403 code if the artifact has exceeded the service quota
-			if resp != nil && (resp.StatusCode >= 400 && resp.StatusCode <= 499) {
+			// Break the retry on any 4xx code except for 429 Too Many Requests.
+			if resp != nil && (resp.StatusCode != 429 && resp.StatusCode >= 400 && resp.StatusCode <= 499) {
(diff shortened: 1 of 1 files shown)


agent/artifact_batch_creator.go
```diff
@@ -86,7 +86,8 @@ func (a *ArtifactBatchCreator) Create(ctx context.Context) ([]*api.Artifact, err
 			}
 
 			creation, resp, err := a.apiClient.CreateArtifacts(ctxTimeout, a.conf.JobID, batch)
-			if resp != nil && (resp.StatusCode == 401 || resp.StatusCode == 404) {
+			// the server returns a 403 code if the artifact has exceeded the service quota
+			if resp != nil && (resp.StatusCode == 401 || resp.StatusCode == 403 || resp.StatusCode == 404) {
```
triarius, line 90: It might be worth doing some debug printing to see what the value of `resp` is.

Something like
```suggestion
      a.logger.Info("response: %#v", resp)
			if resp != nil && (resp.StatusCode == 401 || resp.StatusCode == 403 || resp.StatusCode == 404) {
```

will give you a lot of information.

triarius (approved):
Great work! LGTM!

PR #2698, opened 2024-03-22
Ensure that disabled warnings get passed to the logger in kubernetes agents

moskyb:
### Description

In #2674, we added a feature to the agent where some warnings would be "dismissable" through CLI flags - for example, one could disable this warning:
CleanShot 2024-03-22 at 15 20 58@2x

by passing `--disable-warnings-for "submodules-disabled"`.

However, due to an oversight, this feature wasn't working on Kubernetes agents. This PR fixes that issue by correctly passing the config down the the customised logger that K8s pods use, and making it harder for developers to miss this kind of thing in the future.

### Context

#2674

### Testing

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -1296,8 +1296,8 @@ func agentLifecycleHook(hookName string, log logger.Logger, cfg AgentStartConfig
 
 	// pipe from hook output to logger
 	r, w := io.Pipe()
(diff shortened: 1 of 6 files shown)


triarius (approved):
LGTM ⚙️

DrJosh9000 (approved):
✅

PR #2703, opened 2024-03-26
Artifact creation: Break retry on any 4xx code except 429

DrJosh9000:
There's a debate around which 4xx code we're going to use. #2697 mentions 422 in the title but adds handling for 403. Some people think 429 is better.

Rather than playing whack-a-mole with all the 4xx codes across multiple releases, I think we should do the same thing as the log uploader and break on any of them.

### DIFF
--- a/agent/artifact_batch_creator.go
+++ b/agent/artifact_batch_creator.go
@@ -87,7 +87,8 @@ func (a *ArtifactBatchCreator) Create(ctx context.Context) ([]*api.Artifact, err
 
 			creation, resp, err := a.apiClient.CreateArtifacts(ctxTimeout, a.conf.JobID, batch)
 			// the server returns a 403 code if the artifact has exceeded the service quota
-			if resp != nil && (resp.StatusCode >= 400 && resp.StatusCode <= 499) {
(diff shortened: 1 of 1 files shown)


triarius (commented):
I don't think it's appropriate to break retry on 429. I expect hat if the response is `429 Too Many Requests`, it DOES exhaust the retry loop (but the retry should exponentially back off). Most of the retry loops in the agent don't exponentially back off (I don't really know why) but this one does, so I think we should preserve this even if rate limiting has not been implemented server side yet.

Perhaps you should change it to exit the loop on all 4xx except 429.

The title of https://github.com/buildkite/agent/pull/2697 is clearly a mistake because the linked server side PR clearly returns `Forbidden`. I think 403 Forbidden makes sense for the purpose stated because (I assumed) this is a limit on the number of resources that can exist, not a limit on the number of requests. 422 does not make sense to me because there's no way for the user to modify the request to work FOR THE SAME RESOURCE.

DrJosh9000:
Thanks, that makes sense. Changed it.

tessereth (approved):

PR #2707, opened 2024-03-28
Don't run pre-exit hooks without command phase

DrJosh9000:
### Description

The k8s stack uses phases to split job execution between multiple containers. In particular, the checkout phase happens in a separate container to the command phase. This makes pre-exit hook execute twice, once in each container - this is usually not what is intended by plugin authors, and often breaks when the different phase have different configs or environment variables.

Changing it so pre-exit hooks only run if the command phase is enabled makes it work, and is unlikely to break anybody (who isn't deep in the agent fiddling with phases).

### Context

https://linear.app/buildkite/issue/PENT-10/make-hooks-configurable

### Changes

* Refactor `includePhase` from inline closure to method
* Change pre-exit hooks to only run if phases includes `command`
* Add an integration test

### Testing

### DIFF
--- a/internal/job/executor.go
+++ b/internal/job/executor.go
@@ -13,6 +13,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"runtime"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


internal/job/executor.go
```diff
@@ -180,30 +181,18 @@ func (e *Executor) Run(ctx context.Context) (exitCode int) {
...
-		if len(e.Phases) == 0 {
-			return true
-		}
-		for _, include := range e.Phases {
-			if include == phase {
-				return true
-			}
-		}
-		return false
-	}
```
moskyb, line 193: i have looked at this code and thought "why is this an inline closure? i should make it a method" and then... not for the entire time i've worked here. thank you for making it happen!!!

moskyb (approved):

PR #2720, opened 2024-04-09
Cache paths env var

moskyb:
### Description

Buildkite's upcoming hosted compute product will include a feature to mount cache volumes when instructed to do so by the buildkite backend. Though the mounting itself happens outside of the agent, the agent needs to tell the mounter where to mount the cache volumes.

This PR adds the ability for the agent to consume the cache paths from the accept job response, and pipes those paths through to an envar `BUILDKITE_AGENT_CACHE_PATHS`.

### Context

COMP-321 

### Testing

### DIFF
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -430,6 +430,11 @@ func (r *JobRunner) createEnvironment(ctx context.Context) ([]string, error) {
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


DrJosh9000 (approved):

matthewborden (approved):
Would be good to have a test to prevent regressions.

Looks good 👍🏻

PR #2724, opened 2024-04-10
Upload agent images to GHCR

DrJosh9000:
### Description

Make agent images accessible directly from GHCR.

This relates to pinning agent versions used by the k8s stack, which currently publishes images and helm charts to GHCR. Some customers might also be interested in pulling from GHCR instead of Docker Hub.

### Context

https://linear.app/buildkite/issue/PENT-9/pin-the-version-of-the-buildkite-agent-used-by-agent-stack-k8s

### Changes

- `docker login` to GitHub using the existing credential
- `skopeo copy` the image to ghcr.io/buildkite/agent

### Testing

I also ran a modified release pipeline to verify that it works.

### DIFF
--- a/.buildkite/steps/publish-docker-image.sh
+++ b/.buildkite/steps/publish-docker-image.sh
@@ -36,8 +36,10 @@ parse_version() {
 
 release_image() {
   local tag="$1"
-  echo "--- :docker: Tagging ${target_image}:${tag}"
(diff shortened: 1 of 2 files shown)


moskyb (approved):
this looks great! very happy to ship it. 

when i was working on the k8s stack, one thing that would have been useful would have been publishing per-branch docker images to ghcr, so that we can test development branches of the agent in the kubernetes stack. i don't think we should do this in this PR (unless it's really easy?), but something to keep in mind for the future.

PR #74, opened 2015-01-15
Bootstrap Upgrades

keithpitt:
- [x] Report `bootstrap.sh` version to Buildkite
- [x] Handle Forked GItHub Pull Requests
- [x] Handle clean checkouts

PR #2735, opened 2024-04-19
Fix checkout race condition on GitHub PR builds

rianmcguire:
### Description

The agent has a [special case for GitHub PRs](https://github.com/buildkite/agent/blob/3ebb9d6ff79364ba4329713f1616520022898de8/internal/job/checkout.go#L545-L558), where it fetches `refs/pull/{id}/head`, instead of [directly fetching the commit](https://github.com/buildkite/agent/blob/3ebb9d6ff79364ba4329713f1616520022898de8/internal/job/checkout.go#L570) we want to build.

This can cause the build to fail if there is a force push to the PR branch, as the commit that we wanted to build will no longer be a parent of the current `refs/pull/{id}/head` and so won't be fetched. That causes the checkout to fail with "reference is not a tree".

#### Reproduction steps
1. Set up a Buildkite pipeline that triggers on GitHub pull requests
2. Ensure no agents are connected
3. Create a pull request (a Buildkite build will be created, but will be waiting for agent)
4. Amend the latest commit in the pull request and force push to the PR's branch (a second Buildkite build will be created, but will be waiting for the agent)
5. Start an agent

#### Expected behaviour

Both builds succeed

#### Actual behaviour

The first build fails to checkout its commit with:
```
reference is not a tree: <commit hash>
```

#### Use cases we need to cover

* Pull request build with commit = (full commit hash)
* Pull request build after a force push, with commit = (full commit hash) (this is what we want to fix)
* Pull request build with commit = "HEAD"
* Pull request build with commit = (short commit hash)
* Pull request build from a fork with commit = (full commit hash)
* Pull request build from a fork with commit = "HEAD"
  * This is rare, but has happened. It does mean we can't completely remove the special case, as the only way to find the latest commit for a fork PR is to fetch `refs/pull/{id}/head`, because we can't fetch the fork's branch directly.
* ~~Pull request build from a fork with commit = (short commit hash)~~
  * This hasn't occurred in the last 12 months
* Repos configured with [custom remote tracking](https://git-scm.com/docs/git-fetch#_configured_remote_tracking_branches) that includes `refs/pull/*`
  * This doesn't seem super likely, but better to be backwards compatible where we can
  * We need to ensure we always fetch `refs/pull/{id}/head` for PRs, so the local ref is updated
* ...did I miss anything?

### Context

[Coda](https://coda.io/d/Escalations-Feedback_dHnUHNps1YO/Escalations_sudOj#escalations-board_tu9Bj/r365&view=modal)

This behaviour was introduced in https://github.com/buildkite/agent/pull/74, where it related to a fix for building PRs from forks.

### Changes

For GitHub pull request builds where the commit is known (ie. not `HEAD`), fetch the commit directly.

### Testing

### DIFF
--- a/internal/job/checkout.go
+++ b/internal/job/checkout.go
@@ -708,9 +708,11 @@ func (e *Executor) defaultCheckoutPhase(ctx context.Context) error {
 }
 
 func gitFetchCommitWithFallback(ctx context.Context, shell *shell.Shell, gitFetchFlags, commit string) error {
-	if err := gitFetch(ctx, shell, gitFetchFlags, "origin", commit); err == nil {
+	err := gitFetch(ctx, shell, gitFetchFlags, "origin", commit)
+	if err == nil {
 		return nil // it worked
-	} else if gerr := new(gitError); errors.As(err, &gerr) {
+	}
+	if gerr := new(gitError); errors.As(err, &gerr) {
 		// if we fail in a way that means the repository is corrupt, we should bail
 		switch gerr.Type {
 		case gitErrorFetchRetryClean, gitErrorFetchBadObject:
--- a/internal/job/checkout.go
+++ b/internal/job/checkout.go
@@ -720,8 +720,9 @@ func gitFetchCommitWithFallback(ctx context.Context, shell *shell.Shell, gitFetc
 		}
 	}
 
-	// Some repositories don't support fetching a specific commit so we fall
-	// back to fetching all heads and tags, hoping that the commit is included.
+	// The commit might be something that's not possible to fetch directly
+	// (eg. a short commit hash), so we fall back to fetching all heads and tags,
+	// hoping that the commit is included.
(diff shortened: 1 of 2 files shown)


internal/job/checkout.go
```diff
@@ -726,6 +707,41 @@ func (e *Executor) defaultCheckoutPhase(ctx context.Context) error {
 	return nil
 }
 
+func gitFetchCommitWithFallback(ctx context.Context, shell *shell.Shell, gitFetchFlags, commit string) error {
+	if err := gitFetch(ctx, shell, gitFetchFlags, "origin", commit); err == nil {
+		return nil // it worked
+	} else if gerr := new(gitError); errors.As(err, &gerr) {
```
DrJosh9000, line 713: I've wanted to clean up this particular nit for a while. Because the previous case returns:
```suggestion
	}
	if gerr := new(gitError); errors.As(err, &gerr) {
```

DrJosh9000 (approved):
I like the fairly thorough testing!

PR #2736, opened 2024-04-22
Tweak ETXTBSY retry, and be helpful for ENOENT

### LINKED ISSUE #2734: hooks failed to run because it was already open (on `NixOS`)
I tried using some plugins (`docker`, `docker-compose`, etc) in my pipeline as such:
```yaml
steps:
  - label: ":docker: :rust: build"
    command: RUSTFLAGS="-D warnings" cargo build
    plugins:
      - docker-compose#v3.0.3:
          run: node
    agents:
      queue: nixos
```
Which is executed by the agent on a `nixos` host, which successfully runs all my other pipelines, but this setup fails to run with the following message:

```shell
$ /var/buildkite/plugins/github-com-buildkite-plugins-docker-compose-buildkite-plugin-v3-0-3/hooks/pre-checkout
🚨 Error: The plugin docker-compose pre-checkout hook failed to run because it was already open. We couldn't find out what process had the hook open
🚨 Error: error running "/tmp/buildkite-agent-bootstrap-hook-runner-1100140013": fork/exec /tmp/buildkite-agent-bootstrap-hook-runner-1100140013: no such file or directory
```

@tomowatt gave insight into why this is happening here: https://github.com/buildkite-plugins/docker-compose-buildkite-plugin/issues/434#issuecomment-2063904376

TLDR: The `/bin/bash` executable exists at a different path in `NixOS` as it departs from the `FHS` (File hierarchy standard) to ensure reproducibility. The plugin scripts rely on `bash` being there and it thus fails.  

The only real issue here is that the above **error message is cryptic** and could be improved.
Fixing the path in the nix package is possible to support existing scripts / plugins.

DrJosh9000:
### Description

Since #2325, we learned that the "text file busy" error reported by the customer was probably https://github.com/golang/go/issues/22315.

This change makes the retry and log pathway more specific to that error, and retries more times more frequently. It also adds a new log pathway for ENOENT, which can happen for reasons that the existing log messages obscure (e.g. the error log says something like `fork/exec /tmp/buildkite-agent-hook-wrapper/hook-script-wrapper-XYZ: no such file or directory`, but the _real_ cause is that the script starts with `#!/bin/bash` but Bash isn't there).

### Context

* Fixes #2734

### Testing

### DIFF
--- a/internal/job/executor.go
+++ b/internal/job/executor.go
@@ -422,6 +422,31 @@ func logOpenedHookInfo(l shell.Logger, debug bool, hookName, path string) {
 	}
 }
 
+func logMissingHookInfo(l shell.Logger, wrapperPath string) {
+	// It's unlikely, but possible, that the script wrapper was spontaneously
+	// deleted or corrupted (it's usually in /tmp, which is fair game).
+	// A common setup error is to try to run a Bash hook in a container or other
+	// environment without Bash (or Bash is not in the expected location).
+	shebang, err := shellscript.ShebangLine(wrapperPath)
+	if err != nil {
+		// It's reasonable to assume the script wrapper was spontaneously
+		// deleted, or had something equally horrible happen to it.
+		l.Errorf("The %s hook failed to run - perhaps the wrapper script %q was spontaneously deleted", wrapperPath)
+		return
+	}
+	interpreter := strings.TrimPrefix(shebang, "#!")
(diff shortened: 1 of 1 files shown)


tessereth (approved):
Given the bug ("feature"?) you're trying to work around, this change seems reasonable.

PR #2741, opened 2024-04-23
Experiment: override zero exit code on cancel

### LINKED ISSUE #1489: Cancelled jobs report an exit code of 0 on Windows
This is inconsistent with Linux (haven’t tested on macOS). We should make this behaviour the same across all operating systems.

DrJosh9000:
### Description

Fixes #1489

### Changes

* Add a new experiment, `override-zero-exit-on-cancel`
* If the job is cancelled, and the exit status of the process is 0, override it to be 1 instead.

### Testing


moskyb (approved):
🎉🪟

Issue #2744, opened 2024-04-23
Race condition with socket on startup of kubernetes agent


bpoland:
Hi there, I'm running into an intermittent issue which I think is a race condition when running the agent with https://github.com/buildkite/agent-stack-k8s

In the Buildkite UI the job shows in gray. It looks like it's trying to start the job on a pod but no log output ever comes through.

In kubernetes I can see the pod gets started but the command container (e.g. `container-0`) exits immediately with this error message:
```
# Using experimental Kubernetes support
🚨 Error: Failed to start kubernetes client: error connecting to kubernetes runner: dial unix /workspace/buildkite.sock: connect: no such file or directory
^^^ +++
```

Working with support, I believe there is a race condition where the command container is starting slightly ahead of the `agent` container which would normally create the `/workspace/buildkite.sock`.  I have verified that when a pod gets in this state, I can `kubectl exec` into the agent container and the `/workspace/buildkite.sock` is created by that time.

This seems to have been triggered by us using the new option to skip checkout (https://github.com/buildkite/agent-stack-k8s/pull/303) -- I think previously the checkout container/process was taking long enough that we didn't hit this issue because the agent had time to create the socket, but now it seems to happen 50% of the time or more :(

Since there's no way to ensure kubernetes containers start in a specific order, perhaps all that's needed is a wait during the command container startup until the socket is there?

bpoland:
I think this might fix it but I don't have a Go dev environment so I just wrote it in GitHub :)

https://github.com/buildkite/agent/pull/2745

PR #2745, opened 2024-04-23
Add retry on socket connection issues

bpoland:
### Description

- Wait up to 30 seconds for agent container to start and create socket, to avoid race condition

### Context

Fixes https://github.com/buildkite/agent/issues/2744

### Testing
Not tested, just a suggestion :)


DrJosh9000:
Thanks @bpoland! I might run with this in another PR

bpoland:
> Thanks @bpoland! I might run with this in another PR

Your solution definitely looks more legit. Thanks for the quick turnaround :)

PR #2747, opened 2024-04-26
Make preferring runtime env default off

patrobinson:
### Description

In v3.63.0 we made a breaking change to how pipelines were parsed, specifically the precedence of environment variables in a pipeline upload step.

This undoes that breaking change to restore the original behaviour (see https://github.com/buildkite/go-pipeline/pull/35)

### Changes

The following pipeline example exhibits the breaking change (Credit to @lizrabuya for the reproducible test case).

Pipeline Steps:

```yaml
env:
  HEADER_ENV: "123"
​
steps:
  - command:
      - buildkite-agent pipeline upload pipeline.yml
```

pipeline.yml:

```yaml
env:
  HEADER_ENV: "789" 
  NEW_ENV: "NEW-${HEADER_ENV:-}"
steps: 
- label: "Fixed Values ENV: ${NEW_ENV}" 
  commands: echo "Checking fixed env values ${NEW_ENV} "
- label: "Fixed Values Runtime ENV"
  commands: echo "Checking fixed env values $${NEW_ENV} "
```

Prior to agent version v3.63.0 this pipeline printed Checking fixed env values NEW-789 for both the first and second command jobs.
As of v3.63.0 and above the steps now prints NEW-123.

This restores the original behaviour.

This PR also adds an experiment flag `interpolation-prefers-runtime-env` which enables the behaviour introduced in v3.63.0.

### Testing

### DIFF
--- a/clicommand/pipeline_upload.go
+++ b/clicommand/pipeline_upload.go
@@ -263,7 +263,7 @@ var PipelineUploadCommand = cli.Command{
 			src = "(stdin)"
 		}
 
-		result, err := cfg.parseAndInterpolate(src, input, environ, ctx)
+		result, err := cfg.parseAndInterpolate(ctx, src, input, environ)
 		if w := warning.As(err); w != nil {
 			l.Warn("There were some issues with the pipeline input - pipeline upload will proceed, but might not succeed:\n%v", w)
 		} else if err != nil {
@@ -390,7 +390,7 @@ func searchForSecrets(
 	return nil
 }
 
-func (cfg *PipelineUploadConfig) parseAndInterpolate(src string, input io.Reader, environ *env.Environment, ctx context.Context) (*pipeline.Pipeline, error) {
+func (cfg *PipelineUploadConfig) parseAndInterpolate(ctx context.Context, src string, input io.Reader, environ *env.Environment) (*pipeline.Pipeline, error) {
 	result, err := pipeline.Parse(input)
 	if err != nil && !warning.Is(err) {
 		return nil, fmt.Errorf("pipeline parsing of %q failed: %w", src, err)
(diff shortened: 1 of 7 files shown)


clicommand/pipeline_upload.go
```diff
@@ -401,7 +402,11 @@ func (cfg *PipelineUploadConfig) parseAndInterpolate(src string, input io.Reader
 			}
 			result.Env.Set(tracetools.EnvVarTraceContextKey, tracing)
 		}
-		if err := result.Interpolate(environ); err != nil {
+		preferRuntimeEnv := false
+		if experiments.IsEnabled(ctx, experiments.InterpolationPrefersRuntimeEnv) {
+			preferRuntimeEnv = true
+		}
```
DrJosh9000, line 408: This ought to be equivalent:

```suggestion
		preferRuntimeEnv := experiments.IsEnabled(ctx, experiments.InterpolationPrefersRuntimeEnv)
```

clicommand/pipeline_upload.go
```diff
@@ -389,7 +390,7 @@ func searchForSecrets(
 	return nil
 }
 
-func (cfg *PipelineUploadConfig) parseAndInterpolate(src string, input io.Reader, environ *env.Environment) (*pipeline.Pipeline, error) {
+func (cfg *PipelineUploadConfig) parseAndInterpolate(src string, input io.Reader, environ *env.Environment, ctx context.Context) (*pipeline.Pipeline, error) {
```
DrJosh9000, line 393: When I tidied up context usage a while ago, I stuck to the convention that `ctx` is the first parameter (see https://pkg.go.dev/context and https://google.github.io/styleguide/go/decisions#contexts)

```suggestion
func (cfg *PipelineUploadConfig) parseAndInterpolate(ctx context.Context, src string, input io.Reader, environ *env.Environment) (*pipeline.Pipeline, error) {
```

DrJosh9000 (changes requested):

DrJosh9000 (approved):
✅✅

PR #2759, opened 2024-04-30
Add status.json endpoint to health check endpoints

moskyb:
### Description

This PR adds a new endpoint to the healthcheck server, `/status.json`. the idea is that this endpoint will be a very cut-down version of the information available at `/status`, but in a format (JSON) that's easily machine-readable.

A sample of this output is:
```JSONC
{
  "health": "ok",
  "aggregate_status": "busy", # One of the workers below is busy, so the "aggregate status" of the agent is busy
  "worker_statuses": [
    {
      "status": "idle",
      "id": "018f2d68-1d0c-42cd-be6c-c1a89bda6453",
      "spawn_index": 1
    },
    {
      "status": "idle",
      "id": "018f2d68-1e33-4f09-ae1e-8b962f4e4e77",
      "spawn_index": 2
    },
    {
      "status": "busy",
      "current_job_id": "018f2d68-5e54-4255-a99e-829aa4691dc1", # only busy agents have a current job ID
      "id": "018f2d68-1f57-4374-9dbe-ec033f915c75",
      "spawn_index": 3
    },
    {
      "status": "idle",
      "id": "018f2d68-2080-4eaf-a33d-e1c3d0521206",
      "spawn_index": 4
    }
  ]
}
```

This PR also refactors the healthcheck server to be fully owned by the `AgentPool`, and replaces the use of the default mux (used when `http.HandleFunc` is called) with a dedicated mux for just the healthcheck server.

### Testing

### DIFF
--- a/agent/agent_pool.go
+++ b/agent/agent_pool.go
@@ -2,21 +2,52 @@ package agent
 
 import (
 	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strconv"
 	"sync"
 
+	"github.com/buildkite/agent/v3/logger"
 	"github.com/buildkite/agent/v3/status"
 )
 
 // AgentPool manages multiple parallel AgentWorkers
 type AgentPool struct {
-	workers []*AgentWorker
+	workers     []*AgentWorker
+	idleMonitor *IdleMonitor
 }
 
 // NewAgentPool returns a new AgentPool
 func NewAgentPool(workers []*AgentWorker) *AgentPool {
 	return &AgentPool{
-		workers: workers,
+		workers:     workers,
+		idleMonitor: NewIdleMonitor(len(workers)),
 	}
+}
+
+func (ap *AgentPool) StartStatusServer(ctx context.Context, l logger.Logger, addr string) {
+	mux := http.NewServeMux()
+
+	mux.HandleFunc("/", healthHandler(l))
+	mux.HandleFunc("/status", status.Handle)
+	mux.HandleFunc("/status.json", ap.statusJSONHandler(l))
+
+	for _, worker := range ap.workers {
+		mux.HandleFunc("/agent/"+strconv.Itoa(worker.spawnIndex), worker.healthHandler())
+	}
+
+	go func() {
+		_, setStatus, done := status.AddSimpleItem(ctx, "Health check server")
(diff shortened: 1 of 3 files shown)


agent/agent_pool.go
```diff
@@ -64,11 +93,81 @@ func (r *AgentPool) runWorker(ctx context.Context, worker *AgentWorker, im *Idle
...
+				Status:       worker.getState(),
+				CurrentJobID: worker.getCurrentJobID(),
+				SpawnIndex:   worker.spawnIndex,
+			})
+		}
+
+		err := json.NewEncoder(w).Encode(struct {
+			Health          string              `json:"health"`
+			AggregateStatus agentWorkerState    `json:"aggregate_status"`
+			WorkerStatuses  []agentWorkerStatus `json:"worker_statuses"`
```
dabarrell, line 151: Could this just be `workers`? it's a little simpler, and the child elements are more than just status (also id, job id, etc)

agent/agent_pool.go
```diff
@@ -64,11 +93,81 @@ func (r *AgentPool) runWorker(ctx context.Context, worker *AgentWorker, im *Idle
...
+		}
+	}
+}
+
+func (ap *AgentPool) statusJSONHandler(l logger.Logger) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		type agentWorkerStatus struct {
+			Status       agentWorkerState `json:"status"`
+			CurrentJobID string           `json:"current_job_id,omitempty"`
+			ID           string           `json:"id"`
```
dabarrell, line 128: Noting these are both uuids – did you check what the pattern (naming the attribute id or uuid) was elsewhere in the agent? and in env vars etc?

DrJosh9000 (commented):
Looks pretty good

agent/agent_pool.go
moskyb: we generally call these IDs in the agent, even though they're almost always UUIDs - i suspect the logic is that all UUIDs are IDs, and the scheme might change on the buildkite end at some point? regardless, i suspect the fact that these are UUIDs (rather than some other ID scheme) is of little interest to the users of this endpoint

DrJosh9000 (approved):
🥒

PR #2762, opened 2024-05-02
Make failed job acquisitions return a specific exit code (27)

moskyb:
### Description

When the agent is running in acquire mode - that is, by specifying a job UUID for the agent to pick up, "inverting" the normal dispatch process - sometimes, the job it's been instructed to pick up in unavailable. Perhaps it's already running on another agent, or maybe the job was cancelled, but made its way to the agent anyway.

In these situations, the agent prints the error to stdout, and returns an exit code of `1`, indicating failure.

However, acquisition can fail for both recoverable- and non-recoverable reasons, and the agent doesn't differentiate between these in its status code. Users can grep through the logs for things unrecoverable-sounding things, but this is way more work than it needs to be.

This PR makes it so that if acquiring the job fails in such a way that the error is unrecoverable, the agent will return a status code of `27` (chosen arbitrarily, because it's a nice number). This allows consumers of the agent to not try to pick up that job again. 

### Context

COMP-332
https://docs.google.com/document/d/1qjIXw2gm88iiDggQAKVMQx0qwKntPjtlA4pW84om3xw/edit
[Slack convo w/ Namespace](https://buildkite-corp.slack.com/archives/C05MQ1W710X/p1714457049899449)

### Testing

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -1182,7 +1182,16 @@ var AgentStartCommand = cli.Command{
 			pool.StartStatusServer(ctx, l, cfg.HealthCheckAddr)
 		}
 
-		return pool.Start(ctx)
+		err = pool.Start(ctx)
+		if errors.Is(err, agent.ErrJobAcquisitionFailure) {
+			// If the agent tried to acquire a job, but it couldn't because the job was already taken, we should exit with a
+			// specific exit code so that the caller can know that this job can't be acquired.
+
+			const acquisitionFailedExitCode = 27 // chosen by fair dice roll
+			return cli.NewExitError(err, acquisitionFailedExitCode)
+		}
+
+		return err
 	},
 }
 
--- a/agent/agent_worker.go
+++ b/agent/agent_worker.go
@@ -551,6 +551,8 @@ func (a *AgentWorker) Ping(ctx context.Context) (*api.Job, error) {
 	return ping.Job, nil
 }
 
+var ErrJobAcquisitionFailure = errors.New("failed to acquire job")
+
 // AcquireAndRunJob attempts to acquire a job an run it. It will retry at after the
 // server determined interval (from the Retry-After response header) if the job is in the waiting
 // state. If the job is in an unassignable state, it will return an error immediately.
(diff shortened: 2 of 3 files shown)


clicommand/agent_start.go
```diff
@@ -1182,7 +1182,16 @@ var AgentStartCommand = cli.Command{
...
 		}
 
-		return pool.Start(ctx)
+		err = pool.Start(ctx)
+		if errors.Is(err, agent.ErrJobAcquisitionFailure) {
+			// If the agent tried to acquire a job, but it couldn't because the job was already taken, we should exit with a
+			// specific exit code so that the caller can know that this job can't be acquired.
+
+			const acquisitionFailedExitCode = 77 // chosen by fair dice roll
+			return cli.NewExitError(err, acquisitionFailedExitCode)
```
moskyb, line 1191: happy to replace this number with something chosen with a bit more rigour if necessary

clicommand/agent_start.go
```diff
@@ -1182,7 +1182,16 @@ var AgentStartCommand = cli.Command{
...
 		}
 
-		return pool.Start(ctx)
+		err = pool.Start(ctx)
+		if errors.Is(err, agent.ErrJobAcquisitionFailure) {
+			// If the agent tried to acquire a job, but it couldn't because the job was already taken, we should exit with a
+			// specific exit code so that the caller can know that this job can't be acquired.
+
+			const acquisitionFailedExitCode = 27 // chosen by fair dice roll
+			return cli.NewExitError(err, acquisitionFailedExitCode)
```
patrobinson, line 1191: I wish I could find a better way to do this, It feels like the job of looking up an error and translating it to an exit code should be simpler.

patrobinson (approved):

clicommand/agent_start.go
```diff
@@ -1182,7 +1182,16 @@ var AgentStartCommand = cli.Command{
 			pool.StartStatusServer(ctx, l, cfg.HealthCheckAddr)
 		}
 
-		return pool.Start(ctx)
+		err = pool.Start(ctx)
+		if errors.Is(err, agent.ErrJobAcquisitionFailure) {
+			// If the agent tried to acquire a job, but it couldn't because the job was already taken, we should exit with a
+			// specific exit code so that the caller can know that this job can't be acquired.
+
+			const acquisitionFailedExitCode = 27 // chosen by fair dice roll
```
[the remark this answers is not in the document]
moskyb: nope, AFAICT, we only ever return 0 or 1. the reason i didn't go with `2` is that it implies that this is the second most important exit code. because it's one of two non-zero exit codes, it is the second-most important exit code by definition, but were we to add a third, this would likely not be the case.

regardless, potentially it's a good idea to hoist this const definition out from this function and put it somewhere central-ish in the `clicommand` package?

clicommand/agent_start.go
moskyb: i mean, if the return value of a `cli.Command` is an `ExitCoder` (ie, fulfils `interface { ExitCode() int }`, this would happen automatically, but i actually think that that would be worse than what we have here - it would mean pulling a concern of the CLI into the `agent` package. i think this is fine as is for now, and if we end up checking a whole bunch of error types, we can refactor as necessary

patrobinson: 100% agreed, that's the conclusion I came to, but it was niggling at me

clicommand/agent_start.go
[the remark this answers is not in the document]
keithpitt: Sounds good to me

PR #2764, opened 2024-05-02
Refactor agent integration test API

moskyb:
### Description

The agent integration tests use a mock server to pretend to be the Buildkite Agent API, but it's extremely brittle, and quite frankly, some of the worst code i've ever written. it's difficult to add new endpoints, and callers of the server can't modify its behaviour in any way, which is frustrating when you need to test against specific behaviour.

This PR refactors that server to use Go 1.22's new `http.HandleFunc("GET /some/path/{with}/path/params")` gear, cleans up the handler code to not be godawful, and allows callers to add endpoints, and override the existing ones.

It also adds an error return to the `runJob` helper that gets used in agent integration tests, as we might in future want to test failures of this code

### Testing

### DIFF
--- a/agent/integration/test_helpers.go
+++ b/agent/integration/test_helpers.go
@@ -70,6 +70,7 @@ func runJob(t *testing.T, ctx context.Context, cfg testRunJobConfig) error {
 		AgentConfiguration: cfg.agentCfg,
 		MetricsScope:       scope,
 	})
+
 	if err != nil {
 		t.Fatalf("agent.NewJobRunner() error = %v", err)
 	}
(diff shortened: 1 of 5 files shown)


agent/integration/test_helpers.go
```diff
@@ -125,33 +128,109 @@ func (tae *testAgentEndpoint) logsFor(t *testing.T, jobID string) string {
...
+	// extra routes overwrite default routes if they conflict
+	for _, r := range extraRoutes {
+		routesUniq[fmt.Sprintf("%s %s", r.Method, r.Path)] = r.HandlerFunc
+	}
+
+	wrapRecordRequest := func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+			b, _ := io.ReadAll(req.Body)
+			req.Body.Close()
+			req.Body = io.NopCloser(bytes.NewBuffer(b))
```
patrobinson, line 219: What's the purpose of this?

patrobinson (approved):

agent/integration/test_helpers.go
moskyb: right... this function records the contents of the body of the request so that consumers can look through it later on. unfortunately, the body of an HTTP request is an io.ReadCloser, which is forward-read-only. this means that if we read the body, then tried to read it again, it wouldn't work (it would either return an EOF error, or it would return an empty read body, i can't remember which.

regardless, this little dance is basically:
- read the body into a `[]byte` caled `b`
- close the body (we're done with it)
- replace the body with a copy of `b` that is also an `io.ReadCloser`

i'll add a comment explaining this, it's not super clear just from reading it

PR #2773, opened 2024-05-08
Exit with code 28 when job acquisition is locked

moskyb:
### Description

In https://github.com/buildkite/agent/pull/2762, we made it so that failed job acquisitions would return an exit code `27`, so that consumers could know that acquiring that job would never be viable.

This PR adds similar handling for jobs that are waiting on dependencies to run, and returns an exit code `28` when this is the case. The reason that this exit code differs is that jobs that are waiting on dependencies will eventually (probably, some jobs depend on things that don't exist) become available for acquisition

### Context

[Slack convo](https://buildkite-corp.slack.com/archives/C05MQ1W710X/p1715093549414119)
COMP-400

### Testing

### DIFF
--- a/agent/agent_worker.go
+++ b/agent/agent_worker.go
@@ -551,7 +551,10 @@ func (a *AgentWorker) Ping(ctx context.Context) (*api.Job, error) {
 	return ping.Job, nil
 }
 
-var ErrJobAcquisitionFailure = errors.New("failed to acquire job")
+var (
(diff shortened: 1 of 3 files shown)


moskyb:
this turned out to be the wrong solution for this issue - this was intended to solve some issues our hosted compute platform was seeing, but it turns out that these were actually caused by a bug in our backend. i'm going close out this issue.

PR #2775, opened 2024-05-09
Expand buildkite-agent secret command with a more useful description.

gilesgas:
### Description

This PR expands the `buildkite-agent secret get` command with a more useful description for both the output and the docs.

### Context

The aim is to flesh out this description so that the current extremely brief one we have here:
https://buildkite.com/docs/agent/v3/cli-secret

Contains more info (i.e. separate **Usage**, **Description** and **Example** sections) like the one here:
https://buildkite.com/docs/agent/v3/cli-step#updating-a-step

This edited description (in this PR) affects the wording on https://buildkite.com/docs/agent/v3/cli-secret, which is auto-generated from this content.

### Changes

Just the description, which will be surfaced through the CLI command: 
`buildkite-agent secret get --help`.

### Testing

I have tested the output locally with the command:
`go run *.go secret get --help --debug`
and the output is as intended.

### DIFF
--- a/clicommand/secret_get.go
+++ b/clicommand/secret_get.go
@@ -29,9 +29,23 @@ type SecretGetConfig struct {
 }
 
 var SecretGetCommand = cli.Command{
-	Name:        "get",
-	Usage:       "Get a secret by its key",
-	Description: "Get a secret by key from Buildkite Pipelines Secrets and print it to stdout.",
+	Name:  "get",
+	Usage: "Get a secret by its key",
+	Description: `Usage:
+
+    buildkite-agent secret get [key] [options...]
+
+Description:
+
+Gets a secret from Buildkite secrets and prints it to stdout. The ′key′
+specified in this command is key's name defined for the secret in its cluster.
+The key's name is case-insensitive in this command, and the key's value is
+automatically redacted in the build logs.
+
+Example:
+
+    $ buildkite-agent secret get deploy_key
+	`,
 	Flags: []cli.Flag{
 		cli.StringFlag{
 			Name:   "job",
--- a/clicommand/secret_get.go
+++ b/clicommand/secret_get.go
@@ -30,7 +30,7 @@ type SecretGetConfig struct {
 
 var SecretGetCommand = cli.Command{
 	Name:  "get",
-	Usage: "Get a secret by its key",
+	Usage: "Get a secret by its key and print it to stdout",
 	Description: `Usage:
 
     buildkite-agent secret get [key] [options...]
--- a/clicommand/secret_get.go
+++ b/clicommand/secret_get.go
@@ -38,13 +38,16 @@ var SecretGetCommand = cli.Command{
 Description:
 
 Gets a secret from Buildkite secrets and prints it to stdout. The ′key′
-specified in this command is key's name defined for the secret in its cluster.
-The key's name is case-insensitive in this command, and the key's value is
-automatically redacted in the build logs.
+specified in this command is the key's name defined for the secret in its
+cluster. The key's name is case insensitive in this command, and the
+key's value is automatically redacted in the build logs.
 
-Example:
+Examples:
+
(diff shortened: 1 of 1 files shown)


clicommand/secret_get.go
```diff
@@ -29,9 +29,25 @@ type SecretGetConfig struct {
...
+Description:
+
+Gets a secret from Buildkite secrets and prints it to stdout. The ′key′
+specified in this command is the key's name defined for the secret in its
+cluster. The key's name is case insensitive in this command, and the
+key's value is automatically redacted in the build logs.
+
+Examples:
+
+The following examples reference the same Buildkite secret ′key′:
```
gilesgas, line 47: Would the fact that this line ends in a colon (`:`) cause this line to be rendered as a heading in the docs?

For example, see **Usage**, **Description** and **Example** at the [top of this page of the docs](https://buildkite.com/docs/agent/v3/cli-step#updating-a-step). I suspect not (based on the `To avoid orphaned status updates, in your Pipeline Settings > GitHub:` line on this page) but just wanted to double-check.

moskyb: nope, you're good. i checked this by running the [`scripts/update-agent-help.sh`](https://github.com/buildkite/docs/blob/main/scripts/update-agent-help.sh) script in the docs repo, which produced this output for the `secret get` command:

---
### Usage

`buildkite-agent secret get [key] [options...]`

### Description

Gets a secret from Buildkite secrets and prints it to stdout. The `key`
specified in this command is the key&#39;s name defined for the secret in its
cluster. The key&#39;s name is case insensitive in this command, and the
key&#39;s value is automatically redacted in the build logs.

### Examples

The following examples reference the same Buildkite secret `key`:

```shell
$ buildkite-agent secret get deploy_key
$ buildkite-agent secret get DEPLOY_KEY
```
---

ironically, in typing out my reply, initially the first line was a header. i think so long as there's a newline after the thing that ends with `:`, it won't render it as a header?

clicommand/secret_get.go
```diff
@@ -29,9 +29,25 @@ type SecretGetConfig struct {
 }
 
 var SecretGetCommand = cli.Command{
-	Name:        "get",
-	Usage:       "Get a secret by its key",
-	Description: "Get a secret by key from Buildkite Pipelines Secrets and print it to stdout.",
+	Name:  "get",
+	Usage: "Get a secret by its key",
```
moskyb, line 33: maybe we should keep the `and print it to stdout` bit in the usage - otherwise it could be a bit unclear where the secret goes? ie is it crammed directly into env? (no, but the user wouldn't know that)

moskyb (approved):
one i have one quibble, but it's not blocking, and other than that LGTM! thanks for keeping our docs spic and span ^_^

clicommand/secret_get.go
gilesgas: This wasn't originally in the 'Usage' field. It was in the 'Description', which I've now expanded upon, and I've retained this piece of info.

TBH, I'm not sure where the 'Usage' field value is actually surfaced, but I'll add it nonetheless.

moskyb: `Usage` is surfaced in the parent command's help - ie, currently, calling `buildkite-agent secret --help` returns 
```
Usage:

  buildkite-agent secret <command> [options...]

Available commands are:

  get   Get a secret by its key
  help  Shows a list of commands or help for one command


Options:

  --help, -h  show help
```

moskyb: but now that you mention it, it's a relatively obscure place to be putting info

clicommand/secret_get.go
gilesgas: Thanks for checking this :-)

clicommand/secret_get.go
gilesgas: Aha, of course! I'll take note of this for next time :-)

PR #2776, opened 2024-05-10
Return earlier from non-get credential actions

moskyb:
### Description

The git-credential helper only supports the `get` action, so we should exit from any other action as soon as we can.

### Context

[Slack convo](https://buildkite-corp.slack.com/archives/C071B119REG/p1715307176654929)

### Testing


DrJosh9000 (approved):
🌳

PR #2778, opened 2024-05-13
Fix poorly-timed timestamp insertions

DrJosh9000:
### Description

Prevent the insertion of BK timestamp APCs in the middle of other ANSI sequences.

### Context

https://github.com/buildkite/terminal-to-html/issues/143 among other examples

### The bug

Prior to #2447, BK APCs were only added immediately after new lines or "line clear" codes. In #2447 I made it insert extra timestamps, based on time-since-last timestamp. This made it possible to insert a timestamp within another ANSI sequence, but it seemed unlikely to happen, because a process typically outputs an entire ANSI sequence in one write call (or so I must have thought). This ignores buffers in the middle. If a process writes ANSI sequences into a buffer, and the buffer is written out in multiple write calls, then two write calls could occur on either side of the last-timestamp check. While this still seems unlikely, I've since spotted multiple cases of timestamps being inserted into existing ANSI sequences.

### Changes

- Add a minimal ANSI sequence parser
- Use it to ensure sequences are terminated before inserting a timestamp

### Testing

### DIFF
--- a/process/ansi.go
+++ b/process/ansi.go
@@ -64,11 +64,12 @@ func init() {
 	// and any other byte is the "final byte" which terminates the sequence.
 	// Here's an ASCII-art state diagram:
 	//
-	//         0x30-0x3F             0x20-0x2F
-	//             ^|                    ^|
-	//             |v                    |v
-	// () --'['--> () --anything else--> () --anything else--> (nil)
-	//          parameter           intermediate               final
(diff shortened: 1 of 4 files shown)


tessereth (approved):
I can't say I understand all of it. ANSI escape codes are confusing. But the tests look good and it seems reasonable.

Issue #2748, opened 2024-04-26
Newer versions don't respect `BUILDKITE_CANCEL_GRACE_PERIOD`


etodd:
Somewhere in the last few releases, the Buildkite agent stopped respecting `BUILDKITE_CANCEL_GRACE_PERIOD`, which we currently set to `600`. Instead it kills the job after the default 10 second grace period.

I rolled back our agent to 3.65 and that version seems to respect it properly.

tessereth:
Thanks for reporting this! It looks like this changed in https://github.com/buildkite/agent/pull/2654 in v3.66 when we extended signal grace period handling to all phases of a build. I'll look into the best way to make this more intuitive, but in the mean time, you can set `BUILDKITE_SIGNAL_GRACE_PERIOD_SECONDS` to slightly less than `BUILDKITE_CANCEL_GRACE_PERIOD` and you should get the behaviour you're after.

tessereth:
This should be fixed in the next agent release (v3.74)

PR #2788, opened 2024-05-21
Default signal grace period to slightly less than cancel grace period

tessereth:
When a job is canceled, parent processes send a SIGTERM (by default) to child processes to try and gracefully fail. If a child doesn't finish after a period of time, we SIGKILL it. The grace periods are configured by two options:

* `--cancel-grace-period` - used by the `start` process when canceling the `bootstrap`
* `--signal-grace-period-seconds` - used by the `bootstrap` process when canceling its children (command/hooks/plugins etc)

In https://github.com/buildkite/agent/pull/2654, the signal grace period was extended to cover all job phases. Which means the effective grace period for a job changed from the cancel grace period to the minimum of the two periods.

In https://github.com/buildkite/agent/pull/2696, we changed the default signal grace period to be one second less than the default cancel grace period. But as reported in https://github.com/buildkite/agent/issues/2748, that doesn't help if you're trying to increase the cancel grace period.

This PR instead defaults `--signal-grace-period-seconds` to one less than `--cancel-grace-period`, whatever that happens to be set to. I've manually tested it and it works how you'd expect.

Fixes https://github.com/buildkite/agent/pull/2696.

### Testing

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -836,8 +836,8 @@ var AgentStartCommand = cli.Command{
 		}
 
 		// default signal grace period to slightly less than the cancel grace period
-		if cfg.SignalGracePeriodSeconds == -1 {
-			cfg.SignalGracePeriodSeconds = cfg.CancelGracePeriod - 1
+		if cfg.SignalGracePeriodSeconds < 0 {
+			cfg.SignalGracePeriodSeconds = cfg.CancelGracePeriod + cfg.SignalGracePeriodSeconds
 		}
 
 		if cfg.CancelGracePeriod <= cfg.SignalGracePeriodSeconds {
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -837,11 +837,11 @@ var AgentStartCommand = cli.Command{
 
 		// Treat a negative signal grace period as relative to the cancel grace period
 		if cfg.SignalGracePeriodSeconds < 0 {
-			if cfg.SignalGracePeriodSeconds < cfg.CancelGracePeriod*-1 {
+			if cfg.CancelGracePeriod < -cfg.SignalGracePeriodSeconds {
 				return fmt.Errorf(
-					"signal-grace-period-seconds (%d) must be greater than the negative cancel-grace-period (%d)",
+					"cancel-grace-period (%d) must be at least as big as signal-grace-period-seconds (%d)",
+					cfg.CancelGracePeriod,
 					cfg.SignalGracePeriodSeconds,
-					cfg.CancelGracePeriod*-1,
 				)
 			}
 			cfg.SignalGracePeriodSeconds = cfg.CancelGracePeriod + cfg.SignalGracePeriodSeconds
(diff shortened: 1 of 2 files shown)


clicommand/agent_start.go
```diff
@@ -835,6 +835,11 @@ var AgentStartCommand = cli.Command{
 			}
 		}
 
+		// default signal grace period to slightly less than the cancel grace period
+		if cfg.SignalGracePeriodSeconds == -1 {
+			cfg.SignalGracePeriodSeconds = cfg.CancelGracePeriod - 1
+		}
```
DrJosh9000, line 841: Is it perhaps worth generalising this? e.g.

```suggestion
		if cfg.SignalGracePeriodSeconds < 0 {
			cfg.SignalGracePeriodSeconds = cfg.CancelGracePeriod + cfg.SignalGracePeriodSeconds
		}
```

clicommand/cancel_signal.go
```diff
@@ -29,6 +25,6 @@ var (
 		Usage: "The number of seconds given to a subprocess to handle being sent ′cancel-signal′. " +
 			"After this period has elapsed, SIGKILL will be sent.",
```
DrJosh9000, line 26: The Usage string for this flag should be updated to describe the new interpretation of -1 (or negative values generally).

DrJosh9000 (commented):
This seems sensible

clicommand/agent_start.go
tessereth: I guess? I was just using `-1` as "unset" in true awful linux fashion but I don't have strong opinions

clicommand/agent_start.go
```diff
@@ -835,6 +835,18 @@ var AgentStartCommand = cli.Command{
...
 
+		// Treat a negative signal grace period as relative to the cancel grace period
+		if cfg.SignalGracePeriodSeconds < 0 {
+			if cfg.SignalGracePeriodSeconds < cfg.CancelGracePeriod*-1 {
+				return fmt.Errorf(
+					"signal-grace-period-seconds (%d) must be greater than the negative cancel-grace-period (%d)",
+					cfg.SignalGracePeriodSeconds,
+					cfg.CancelGracePeriod*-1,
+				)
+			}
```
DrJosh9000, line 846: Hmmm. Comparisons with negatives get a bit harder to understand. At least, I don't think it's natural to negate `CancelGracePeriod` when the already-negated quantity is `SignalGracePeriod`. Phrase in terms of positive values?

```go
			if cfg.CancelGracePeriod < -cfg.SignalGracePeriodSeconds {
				return fmt.Errorf(
					"cancel-grace-period (%d) must be at least as big as signal-grace-period-seconds (%d)",
					cfg.CancelGracePeriod,
					cfg.SignalGracePeriodSeconds,
				)
			}
```

...or...

```go
			if cfg.CancelGracePeriod + cfg.SignalGracePeriodSeconds < 0 {
				return fmt.Errorf(
					"cancel-grace-period (%d) + signal-grace-period-seconds (%d) must be non-negative",
					cfg.CancelGracePeriod,
					cfg.SignalGracePeriodSeconds,
				)
			}
```

DrJosh9000 (approved):

PR #2794, opened 2024-05-24
Fix redaction when no initial redactors are present

moskyb:
**This PR is best reviewed commit-by-commit**

### Problem

There's currently an edge case with with the agent's sensitive value redaction, where if the agent is started with an empty value for the `--redact-vars` flag (which in most cases has a sensible default), no redactions would happen. this is to be expected - it's what the user explicitly asked for, after all! however, doing this also prevents the user from dynamically adding values to redact using the `buildkite-agent redactor add` command, or automatically via the `buildkite-agent secret get` command. 

This behaviour was due to a bug where if upon startup, the job executor [found no values to redact](https://github.com/buildkite/agent/blob/0aa38aa4bd5e780296df293d220f78a796351958/internal/job/executor.go#L1130-L1133) - ie if the value of the `--redact-vars` flag was empty - it would not create any redactors. This behaviour was based on the assumption that vars to redact wouldn't change, but this is no longer the case since the introduction of the `buildkite-agent redactor add` command.

On its own this isn't much of an issue, as overriding the `--redact-vars` flag to not redact anything is a bit of a weird choice, but this bug interacted with the the [agent-stack-k8s](https://github.com/buildkite/agent-stack-k8s) in an interesting way.

In most modes of use, the `buildkite-agent start` process directly starts the `buildkite-agent bootstrap` process, which is the thing that actually runs the buildkite jobs. In these cases, the `start` process (the parent) will pass down the value of `--redact-vars` to its chile processes.

In the k8s-stack, however, the bootstrap container is started directly by the kubernetes controller, and only the command container (but not the checkout container) were given values for the `--redact-vars` flag.

All of this came together to mean that in jobs running on the kubernetes stack, in the command container, **no redaction at all would occur**.

### Solution

This PR fixes the above error in two ways:
- It fixes the underlying error of not creating redactors when there are initially no values to redact - values can be added by the `redactor add` command, so the job executor should be created with an empty redactor
- It updates the `buildkite-agent bootstrap` command to have the [same sensible default](https://github.com/buildkite/agent/blob/main/clicommand/global.go#L99-L109) as the `buildkite-agent start` command, such that if it's called by kubernetes, it will still have sensible defaults for redaction

### Context

[Support Escalation](https://coda.io/d/Escalations-Feedback_dHnUHNps1YO/Pipelines-Escalations_su7FT#Pipelines-Escalations-Board_tu__K/r453&view=modal)
[Slack convo w/ customer](https://buildkite.slack.com/archives/C074YF917A5/p1716424614877029)

### Testing

### DIFF
--- a/internal/job/integration/redaction_integration_test.go
+++ b/internal/job/integration/redaction_integration_test.go
@@ -59,10 +59,7 @@ func TestRedactorDoesNotRedactAgentToken_WhenNotInRedactedVars(t *testing.T) {
 }
 
 func TestRedactorAdd_RedactsVarsAfterUse(t *testing.T) {
-	// when this test is run inside an agent (ie in CI), we don't want to connect to the job API of the executor running this test
-	// we want to instead connect to the job API of the executor process we're testing
-	t.Setenv("BUILDKITE_AGENT_JOB_API_TOKEN", "")
-	t.Setenv("BUILDKITE_AGENT_JOB_API_SOCKET", "")
+	t.Parallel()
 
 	tester, err := NewExecutorTester(mainCtx)
 	if err != nil {
@@ -70,7 +67,7 @@ func TestRedactorAdd_RedactsVarsAfterUse(t *testing.T) {
 	}
 	defer tester.Close()
 
-	secret := "huunter2" // secrets 6 chars or shorter don't get redacted, hence the extra u
+	secret := "hunter2"
 	tester.ExpectGlobalHook("command").AndCallFunc(redactionTestCommandHook(t, secret))
 
 	err = tester.Run(t)
@@ -93,10 +90,7 @@ func TestRedactorAdd_RedactsVarsAfterUse(t *testing.T) {
 // an empty list of redacted vars, it would not initialise the redactors at all, meaning that later redactions (ie with
 // the `buildkite-agent redactor add` command, or automatically when using `buildkite-agent secret get`) would not occur
 func TestRegression_TestRedactorAdd_StillWorksWhenNoInitialRedactedVarsAreProvided(t *testing.T) {
-	// when this test is run inside an agent (ie in CI), we don't want to connect to the job API of the executor running this test
-	// we want to instead connect to the job API of the executor process we're testing
-	t.Setenv("BUILDKITE_AGENT_JOB_API_SOCKET", "")
-	t.Setenv("BUILDKITE_AGENT_JOB_API_TOKEN", "")
+	t.Parallel()
 
 	tester, err := NewExecutorTester(mainCtx)
 	if err != nil {
@@ -104,7 +98,7 @@ func TestRegression_TestRedactorAdd_StillWorksWhenNoInitialRedactedVarsAreProvid
 	}
 	defer tester.Close()
 
-	secret := "huunter2" // secrets 6 chars or shorter don't get redacted, hence the extra u
+	secret := "hunter2"
 	tester.ExpectGlobalHook("command").AndCallFunc(redactionTestCommandHook(t, secret))
 
 	err = tester.Run(t, `BUILDKITE_REDACTED_VARS=""`)
--- a/internal/job/integration/redaction_integration_test.go
+++ b/internal/job/integration/redaction_integration_test.go
@@ -1,0 +1,161 @@
+package integration
+
+import (
+	"fmt"
+	"strings"
+	"testing"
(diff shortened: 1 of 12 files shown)


internal/job/integration/redaction_integration_test.go
```diff
@@ -0,0 +1,137 @@
...
+		t.Fatalf("running executor tester: %v", err)
+	}
+
+	// verify that the secret was echoed prior to calling RedactionCreate
+	if !strings.Contains(tester.Output, fmt.Sprintf("The secret is: %s", secret)) {
+		t.Fatalf("expected secret to be echoed prior to redaction, but it wasn't. Full output: %s", tester.Output)
+	}
+
+	// now check that the string is still unredacted after the call to RedactionCreate
+	if !strings.Contains(tester.Output, fmt.Sprintf("There should be a redacted here: %s", secret)) {
```
moskyb: i wonder if it's clearer (if less idiomatic) to just have this be
```suggestion
	if !strings.Contains(tester.Output, "There should be a redacted here: huunter2" {

```

internal/job/integration/redaction_integration_test.go
```diff
@@ -0,0 +1,140 @@
...
+	t.Setenv("BUILDKITE_AGENT_JOB_API_SOCKET", "")
+	t.Setenv("BUILDKITE_AGENT_JOB_API_TOKEN", "")
+
+	tester, err := NewExecutorTester(mainCtx)
+	if err != nil {
+		t.Fatalf("setting up executor tester: %v", err)
+	}
+	defer tester.Close()
+
+	secret := "huunter2" // secrets 6 chars or shorter don't get redacted, hence the extra u
```
DrJosh9000, line 107: Isn't `hunter2` 7 chars long?

internal/job/integration/redaction_integration_test.go
DrJosh9000: It's not less idiomatic, if anything I think it's more idiomatic for a test on `strings.Contains` to have the full string being searched for plainly (e.g. you can grep the repo for the whole string to find the test).

internal/job/integration/redaction_integration_test.go
```diff
@@ -0,0 +1,164 @@
...
+
+		client, err := jobapi.NewClient(mainCtx, socketPath, socketToken)
+		if err != nil {
+			t.Errorf("creating Job API client: %v", err)
+			c.Exit(1)
+			return
+		}
+
+		fmt.Fprintf(c.Stderr, "The secret is: %s\n", secret)
+		time.Sleep(time.Second) // let the log line be written before we add it to the redactor
```
DrJosh9000, line 152: This could be a source of flakes in future - something to fix later.
(There's no guarantee that the log line will make it through in any particular amount of time.)

internal/job/integration/redaction_integration_test.go
```diff
@@ -0,0 +1,164 @@
...
+	if !strings.Contains(tester.Output, "The agent token is: test-token-please-ignore") {
+		t.Fatalf("expected agent token to be printed in full, but it wasn't. Full output: %s", tester.Output)
+	}
+}
+
+func TestRedactorAdd_RedactsVarsAfterUse(t *testing.T) {
+	// when this test is run inside an agent (ie in CI), we don't want to connect to the job API of the executor running this test
+	// we want to instead connect to the job API of the executor process we're testing
+	t.Setenv("BUILDKITE_AGENT_JOB_API_TOKEN", "")
+	t.Setenv("BUILDKITE_AGENT_JOB_API_SOCKET", "")
```
DrJosh9000, line 65: I'm not sure why this needs to be set here and not passed through `tester.Run(t, ...)` or even set as defaults in `NewExecutorTester` (should _any_ tester be connecting to the executor in CI?) Or maybe I'm forgetting some key detail about how the executor works. 

Either way this can be cleaned up later.

DrJosh9000 (approved):
LGTM

internal/job/integration/redaction_integration_test.go
moskyb: 😶

internal/job/integration/redaction_integration_test.go
moskyb: hmm, i think you're right. i think i misdiagnosed an issue in CI, it seems to run fine without these stanzas

PR #2795, opened 2024-05-27
Use Rand per-test in agent/plugin/error_test.go

moskyb:
rand.Rands are not threadsafe and only one should be used per goroutine, so one should be used per parallel tests, and shouldn't be shared between them


CerealBoy (approved):
LGTM!

DrJosh9000 (approved):
Very true!

PR #2812, opened 2024-06-05
Revert "Dont sign initial steps with interpolations"

moskyb:
Reverts buildkite/agent#2807

This PR had an issue where escaped dollar signs that aren't interpolations (now i know that those exist! 🫠) would fail to parse as pipeline uploads, as interpolate would try to parse them as an expansion, which they were not.

we were alerted to this by the agent pipeline, which contains as its final line:
```
if: build.env("DRY_RUN") == "true" || build.branch =~ /^(main|.*-.*-stable)$$/
```

there's a double-dollar there being used as an escaped literal dollar sign for the end of a regex, but the change to the interpolate library was attempting to parse it as an expansion `$$/`, but `/` is not a valid identifier, so the interpolate library was getting confused. the initial error message can be found [here](https://buildkite.com/buildkite/agent/builds/8536#018fe212-fba5-4c36-8c62-f551438be49a):
```
$ buildkite-agent pipeline upload
2024-06-04 07:07:29 INFO   Searching for pipeline config...
2024-06-04 07:07:29 INFO   Found config file ".buildkite/pipeline.yml"
2024-06-04 07:07:29 INFO   Updating BUILDKITE_COMMIT to "d36f21b28f507cb9e36c6fdb25ed449b0343867f"
fatal: pipeline interpolation of "pipeline.yml" failed: interpolating remaining fields: Expected identifier to start with a letter, got /
```

### DIFF
--- a/clicommand/tool_sign.go
+++ b/clicommand/tool_sign.go
@@ -1,7 +1,6 @@
 package clicommand
 
 import (
-	"bytes"
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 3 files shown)


DrJosh9000 (approved):
😔

PR #2814, opened 2024-06-06
kubernetes-exec is now a flag

DrJosh9000:
### Description

Replaces the `kubernetes-exec` "experiment" with a proper flag, environment variable, etc.

### Context

The `kubernetes-exec` "experiment" was not really an experiment, but rather misused the experiments system as a way to pass a Boolean flag into various parts of the agent without setting up a flag, environment variable, etc.

### Testing

### DIFF
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -50,28 +50,30 @@ const (
 // Certain env can only be set by agent configuration.
 // We show the user a warning in the bootstrap if they use any of these at a job level.
 var ProtectedEnv = map[string]struct{}{
-	"BUILDKITE_AGENT_ENDPOINT":           {},
 	"BUILDKITE_AGENT_ACCESS_TOKEN":       {},
 	"BUILDKITE_AGENT_DEBUG":              {},
+	"BUILDKITE_AGENT_ENDPOINT":           {},
 	"BUILDKITE_AGENT_PID":                {},
 	"BUILDKITE_BIN_PATH":                 {},
-	"BUILDKITE_CONFIG_PATH":              {},
 	"BUILDKITE_BUILD_PATH":               {},
+	"BUILDKITE_COMMAND_EVAL":             {},
+	"BUILDKITE_CONFIG_PATH":              {},
+	"BUILDKITE_CONTAINER_COUNT":          {},
+	"BUILDKITE_GIT_CLEAN_FLAGS":          {},
+	"BUILDKITE_GIT_CLONE_FLAGS":          {},
+	"BUILDKITE_GIT_CLONE_MIRROR_FLAGS":   {},
+	"BUILDKITE_GIT_FETCH_FLAGS":          {},
+	"BUILDKITE_GIT_MIRRORS_LOCK_TIMEOUT": {},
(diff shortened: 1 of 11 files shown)


agent/tags.go
```diff
@@ -78,7 +77,7 @@ type tagFetcher struct {
 func (t *tagFetcher) Fetch(ctx context.Context, l logger.Logger, conf FetchTagsConfig) []string {
 	tags := conf.Tags
 
-	if experiments.IsEnabled(ctx, experiments.KubernetesExec) {
+	if t.k8s != nil {
```
zhming0, line 80: Is this line right? the `t.k8s` seems to be set always?

agent/job_runner.go
```diff
@@ -488,6 +494,10 @@ func (r *JobRunner) createEnvironment(ctx context.Context) ([]string, error) {
 	env["BUILDKITE_STRICT_SINGLE_HOOKS"] = fmt.Sprintf("%t", r.conf.AgentConfiguration.StrictSingleHooks)
 	env["BUILDKITE_SIGNAL_GRACE_PERIOD_SECONDS"] = fmt.Sprintf("%d", int(r.conf.AgentConfiguration.SignalGracePeriod/time.Second))
 
+	if r.conf.KubernetesExec {
+		env["BUILDKITE_KUBERNETES_EXEC"] = "true"
```
zhming0, line 498: This seems to be new. Just for my curiosity, what is this impacting?

zhming0 (commented):
I have one question. But the rest looks good 👍🏿 .

moskyb (approved):
it's so good. it's so so good.

agent/job_runner.go
DrJosh9000: Previously, if `kubernetes-exec` were enabled (as an experiment), it would be passed down to the subprocess via the `BUILDKITE_AGENT_EXPERIMENTS` variable. Now it could be set in the parent process via a flag, we should ensure that if enabled it should be passed to the subprocess. 

But we'll make the k8s controller set this anyway, so it's probably unnecessary.

agent/tags.go
DrJosh9000: Good point, on one hand this is defensive in case we set `t.k8s optionally, on the other YAGNI.

DrJosh9000: I changed it so that it used the same logic as before (only run `t.k8s` to create tags if `kubernetes-exec` is enabled).

clicommand/agent_start.go
```diff
@@ -165,6 +165,7 @@ type AgentStartConfig struct {
 	Experiments       []string `cli:"experiment" normalize:"list"`
 	Profile           string   `cli:"profile"`
 	StrictSingleHooks bool     `cli:"strict-single-hooks"`
+	KuberentesExec    bool     `cli:"kubernetes-exec"`
```
bpoland, line 168: Heads up there's a typo here Kuber`en`tesExec

moskyb: lol, whoops. i've just put a PR in to fix it, but there should be no impact

PR #2815, opened 2024-06-06
Disallow artifact path traversal, escape-hatch experiment

DrJosh9000:
### Description

Prevent `..` path components in artifact records from causing `artifact download` to traverse outside the destination path, unless a new experiment is enabled.

A secondary change is the removal of the loop from target path calculation that compared every 100th download path component against the final destination path component; the probability anyone has exercised the code path is effectively zero, and the behaviour would be incredibly surprising if it were hit.

### Context

https://linear.app/buildkite/issue/PS-68/fix-surprising-agent-behaviour-in-artifact-download

### Changes

* New code to trim `..` path components
* Add the experiment
* The removal of the astonishing `for ... i += 100` loop
* A bunch of cleanup

### Testing


agent/download.go
```diff
@@ -73,38 +74,43 @@ func (d Download) Start(ctx context.Context) error {
...
-			break
-		}
+	destPathComponents := strings.Split(destPath, string(os.PathSeparator))
+	dlPathComponents := strings.Split(dlPath, string(os.PathSeparator))
+
+	// If the last component of the destination path matches the first component
+	// of the download path, then trim the last component of the destination.
+	lastIndex := len(destPathComponents) - 1
+	lastDestComponent := destPathComponents[lastIndex]
+	if lastDestComponent == dlPathComponents[0] {
```
zhming0, line 91: not relevant to this PR, but this isn't mentioned anywhere in the doc: https://buildkite.com/docs/agent/v3/cli-artifact#downloading-artifacts-artifact-download-examples

😮

agent/download.go
```diff
@@ -73,38 +74,43 @@ func (d Download) Start(ctx context.Context) error {
...
-		if lastPathInDestination == downloadPaths[i] {
-			destinationPaths = destinationPaths[:lastIndex]
+	if experiments.IsEnabled(ctx, experiments.DisallowArtifactPathTraversal) {
+		// Trim empty or ".." components from the prefix of dlPath; it should not be
+		// permitted to walk up the directory tree from destPath.
+		for len(dlPathComponents) > 0 {
+			if c := dlPathComponents[0]; c != "" && c != ".." {
+				break
+			}
+			dlPathComponents = dlPathComponents[1:]
```
zhming0, line 103: I feel mutating users' intention is a bit questionable. It's likely to cause some support tickets. 🤔 

And given they can add relative path in both download path and destination path, is this fix sufficient?

DrJosh9000: I answered this out of band, but basically, it's less surprising if a user has deliberately written the command:

```shell
buildkite-agent artifact download "*foo" ../../dest
```

than if they wrote the following (and "*foo" happened to match an artifact with path "../../dest"):

```shell
buildkite-agent artifact download "*foo" .
```

zhming0 (approved):

DrJosh9000:
I'm strongly considering inverting the experiment, as there's little evidence I can find that anyone is relying on the old behaviour.

PR #2824, opened 2024-06-12
Publish debian and rpm packages to Buildkite Packages

sj26:
We've added support for OIDC tokens to the package create endpoint built into Buildkite Packages now, so let's try it out! This adds another step similar to the Packagecloud task which uses an OIDC token to upload the agent packages to Buildkite Packages.

Testing this is a little tricky. I think we kinda just need to try it. All the changes are isolated to steps which are allowed to soft fail. The scripts are mostly copy/pasted from the proven Packagecloud scripts and then the differences I've tested in my terminal (e.g. https://github.com/buildkite/buildkite/pull/17083#issuecomment-2161873567).

Fixes [PKG-7120](https://linear.app/buildkite/issue/PKG-7120/use-oidc-from-the-buildkite-agent-release-pipeline).


.buildkite/pipeline.release-experimental.yml
```diff
@@ -40,6 +40,20 @@ steps:
...
 
+  - name: ":redhat: Publish Edge RPM Package to Buildkite Packages"
+    command: ".buildkite/steps/publish-rpm-buildkite-packages.sh"
+    env:
+      REPOSITORY: "buildkite/agent-rpm-experimental"
+    agents:
+      queue: "deploy"
+    plugins:
+      - docker#v5.8.0:
+          image: "public.ecr.aws/docker/library/ruby:3.0"
```
moskyb, line 51: afaict we aren't actually using ruby at all - can we make this `alpine` or `debian:$WHATEVER_DEBIAN_VERSION_PEOPLE_LIKE_USING_THESE_DAYS` or something?

.buildkite/steps/publish-debian-buildkite-packages.sh
moskyb, line 1: these two files (`publish-debian-buildkite-packages.sh` and `publish-rpm-buildkite-packages.sh`) seem to have almost the same functionality - are you able to combine them into one and parameterise them?

moskyb (changes requested):
yo! this looks good and i'm psyched to be publishing things to packages. i have a couple notes, and would also be keen for you to run shellcheck across the newly-added script files just to catch any low-hanging fruit

.buildkite/pipeline.release-experimental.yml
sj26: Yup, happy to use whatever. I was using the same image so that it might already be warm in the cache, and it has curl installed. Neither alpine nor debian have curl installed.

`032379705303.dkr.ecr.us-east-1.amazonaws.com/deploytools:2022.07` seems to be used by other steps and has curl, I'll swap to that.

.buildkite/steps/publish-debian-buildkite-packages.sh
sj26: Sure

sj26:
Cheers for the feedback @moskyb. `shellcheck` runs clean. Mind taking another look?

.buildkite/pipeline.release-experimental.yml
sj26: Worth nothing, what these script are currently doing should be folded into `bk` or a plugin over time, hopefully without needing docker involved at all.

moskyb: cool, looking forward to it!

moskyb (approved):
🎉

sj26:
Looks like the deploy tools container is unhappy with me, not sure why yet

image

https://buildkite.com/buildkite/agent-release-edge/builds/1275#01900ad3-e46c-4f19-8301-8b9f8f62f0c5

sj26:
(I was using the wrong queue!)

yob:
@sj26 buildkite/agent isn't on the allowlist for the deploy queue any more - hopefully `elastic-runners` works great though

sj26:
@yob yeah thanks, I think I originally copy/pasted the Packagecloud steps from an older branch which was still using the deploy queue — PEBKAC.

yob:
change change, everywhere, :cheese:

sj26:
https://buildkite.com/buildkite/agent-release-edge/builds/1286#01900f34-e625-4497-895d-b3315aefb8ed/197

Screenshot 2024-06-13 at 11 51 41 am

nichsolson-yes

PR #2825, opened 2024-06-12
Shellcheck

moskyb:
### Testing


DrJosh9000:
Closing due to staleness ☹️ but I like it and we should do it

PR #2830, opened 2024-06-12
Publish Buildkite Packages without Docker

sj26:
Following on from #2824 and #2824, it looks like the deploytools image is only available when authenticated to ECR:

image

I started by copy/pasting the auth gear from the other steps which use deploytools .. but then realised all these steps need are bash, buildkite-agent, and curl .. and those are all available on the host. So just remove all the docker plugin stuff and let it run on the host.


DrJosh9000 (approved):
🦖

sj26:
(rebased because GitHub wanted it updated)

PR #2834, opened 2024-06-13
Redact Job API token like other env vars

DrJosh9000:
### Description

Reduce confusion around having a `_TOKEN` env var around that should be redacted as configured, even though the risks associated with it leaking should be minimal.

### Context

https://linear.app/buildkite/issue/PS-71/job-api-token-not-redacted-is-confusing-customers

### Changes

* Add the Job API token to the redactors if the name would have matched the configured redacted vars patterns
* Move tests

### Testing

### DIFF
--- a/internal/job/api.go
+++ b/internal/job/api.go
@@ -3,6 +3,7 @@ package job
 import (
 	"fmt"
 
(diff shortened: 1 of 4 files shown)


wolfeidau (approved):
Looks great 🚀

PR #2837, opened 2024-06-17
Release Docker images on Buildkite Packages

swebb:
### Description

Pushing our images to Buildkite Packages to a great example of dogfooding out own product. However, while we're building confidence that everything works correctly we don't want the entire build to fail if there is an issue with Buildkite Packages so I've set it to soft fail.

### Context

[PKG-7137](https://github.com/buildkite/agent/pull/2837)

### Testing

I ran `checkshell` locally. The only error was unrelated to these changes.

### DIFF
--- a/.buildkite/steps/publish-docker-image.sh
+++ b/.buildkite/steps/publish-docker-image.sh
@@ -41,8 +41,7 @@ release_image() {
   echo "--- :github: Copying ${target_image}:${tag} to GHCR"
   dry_run skopeo copy --multi-arch all "docker://${source_image}" "docker://ghcr.io/${target_image}:${tag}"
   echo "--- :buildkite: Copying ${target_image}:${tag} to Buildkite Packages"
-  dry_run skopeo copy --multi-arch all "docker://${source_image}" "docker://packages.buildkite.com/buildkite/agent-docker/agent/${target_image}:${tag}" \
-    || true # Soft fail if we can't upload to Packages
+  dry_run skopeo copy --multi-arch all "docker://${source_image}" "docker://packages.buildkite.com/buildkite/agent-docker/agent/${target_image}:${tag}"
 }
 
 variant="${1:-}"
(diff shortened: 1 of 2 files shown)


.buildkite/steps/publish-docker-image.sh
```diff
@@ -40,6 +40,9 @@ release_image() {
   dry_run skopeo copy --multi-arch all "docker://${source_image}" "docker://docker.io/${target_image}:${tag}"
   echo "--- :github: Copying ${target_image}:${tag} to GHCR"
   dry_run skopeo copy --multi-arch all "docker://${source_image}" "docker://ghcr.io/${target_image}:${tag}"
+  echo "--- :buildkite: Copying ${target_image}:${tag} to Buildkite Packages"
+  dry_run skopeo copy --multi-arch all "docker://${source_image}" "docker://packages.buildkite.com/buildkite/agent-docker/agent/${target_image}:${tag}" \
+    || true # Soft fail if we can't upload to Packages
```
moskyb, line 45:
```suggestion
  dry_run skopeo copy --multi-arch all "docker://${source_image}" "docker://packages.buildkite.com/buildkite/agent-docker/agent/${target_image}:${tag}"
```

i think that soft-failing the step when pushing to packages is a good call, but my worry is that will just silently fail, and we don't notice. i reckon we should either: 
- make it so that (as above) this hard-fails the build - and we can go and whinge to packages that docker registries are broken - or, 
- have a separate step that pushes docker images to packages, and allow that step to soft fail

with both of the above strategies, there's strong signals that pushing docker images to BK packages has failed, and we can come and talk to y'all about it

moskyb (changes requested):

.buildkite/steps/publish-docker-image.sh
swebb: I'd like to build a little confidence it in before switching to hard fail. How about we leave it as soft-fail for a week and get a few builds pushing through successfully and then switch to hard fail?

DrJosh9000: I completely agree with @moskyb that this could just start failing one day and nobody would notice. I doubt we need to make it soft-fail either. If it doesn't work, it's not difficult to revert and try again.

swebb: I've changed this so it will hard fail.

moskyb (approved):
🤙 send it

PR #2851, opened 2024-06-25
 Convey env vars between k8s containers

DrJosh9000:
### Description

Pass environment vars through the k8s socket

### Context

https://linear.app/buildkite/issue/PS-72/make-buildkite-agent-id-available-within-k8s-stack-containers

### Changes

* In kubernetes/kubernetes.go: Replace `AccessToken string` with `Env []string`
* In agent/job_runner.go: Provide `processEnv` to the Kubernetes socket server
* In clicommand/bootstrap.go: Connect the socket as the very first thing it does, to get env vars in time for configuring the executor
* Some cleanup

### Testing

### DIFF
--- a/clicommand/bootstrap.go
+++ b/clicommand/bootstrap.go
@@ -10,8 +10,11 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/buildkite/agent/v3/env"
 	"github.com/buildkite/agent/v3/internal/job"
+	"github.com/buildkite/agent/v3/kubernetes"
 	"github.com/buildkite/agent/v3/process"
+	"github.com/buildkite/roko"
 	"github.com/urfave/cli"
 )
 
@@ -98,6 +101,7 @@ type BootstrapConfig struct {
 	NoJobAPI                     bool     `cli:"no-job-api"`
 	DisableWarningsFor           []string `cli:"disable-warnings-for" normalize:"list"`
 	KubernetesExec               bool     `cli:"kubernetes-exec"`
+	KubernetesContainerID        int      `cli:"kubernetes-container-id"`
 }
 
 var BootstrapCommand = cli.Command{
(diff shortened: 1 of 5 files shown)


clicommand/bootstrap.go
```diff
@@ -376,6 +387,34 @@ var BootstrapCommand = cli.Command{
...
+				return k8sClient.Connect(ctx)
+			})
+			if err != nil {
+				return fmt.Errorf("error connecting to kubernetes runner: %w", err)
+			}
+			for n, v := range env.FromSlice(regResp.Env).Dump() {
+				if err := os.Setenv(n, v); err != nil {
+					return err
+				}
+			}
```
zhming0, line 415: The only issues I can think of for this are: 

1. now many environment variables set by our k8s controller as part of the pod spec will be entirely useless. We might want to eliminate those logic in k8s stack.
2. Is there any environment variables that agent container have access but bootstrap containers don't/shouldn't?

clicommand/bootstrap.go
```diff
@@ -454,7 +493,7 @@ var BootstrapCommand = cli.Command{
 			TracingServiceName:           cfg.TracingServiceName,
 			JobAPI:                       !cfg.NoJobAPI,
 			DisabledWarnings:             cfg.DisableWarningsFor,
-			KubernetesExec:               cfg.KubernetesExec,
+			KubernetesClient:             k8sClient,
```
zhming0, line 496: [no action required] Not relevant to this PR, but I got tripped by this "k8s client" naming a few times. To my brain it always means the "k8s api client". 

Agent socket client might be a better name?

zhming0 (approved):
No test coverage on this code path makes me a bit nervous but the change looks very reasonable and solid. 

It's good to put more controls in agent.

clicommand/bootstrap.go
DrJosh9000: 1. Probably!
2. In theory, everything the agent container passes this way to the bootstrap container would, if not using the k8s stack, be set specifically on the bootstrap subprocess forked by the agent. So I don't imagine there's anything the bootstrap shouldn't have access to. Definitely something to think about.

clicommand/bootstrap.go
DrJosh9000: Great point, I'll come up with a better name in a minute...

clicommand/bootstrap.go
DrJosh9000: Point 2 was a good callout - after thinking on it for a bit, some of the vars normally set by the agent would interfere with the variables set by the k8s stack if they were set after the bootstrap is forked.

DrJosh9000:
I did some testing after merge (it was slightly easier to grab an edge build from Docker Hub last night) and I discovered that despite setting env vars as the first thing inside the bootstrap command, it's still not early enough. urfave/cli uses the `EnvVar` field on each flag as the flag default, so the env vars would need to be set even earlier in the process (before urfave/cli processes the flag definitions).

PR #2852, opened 2024-06-27
Don't try to early-set env vars

DrJosh9000:
### Description

After some testing, I discovered that the env vars can't be set early enough to affect the bootstrap config, so this moves back to doing most of the socket setup in executor.go

### Changes

* Undo most of 36940e5

### Testing

### DIFF
--- a/internal/job/executor.go
+++ b/internal/job/executor.go
@@ -101,13 +101,14 @@ func (e *Executor) Run(ctx context.Context) (exitCode int) {
 		e.shell.SignalGracePeriod = e.ExecutorConfig.SignalGracePeriod
 	}
 
-	if e.K8sAgentSocket != nil {
-		if err := e.kubernetesSetup(ctx, e.K8sAgentSocket); err != nil {
+	if e.KubernetesExec {
+		socket := &kubernetes.Client{ID: e.KubernetesContainerID}
+		if err := e.kubernetesSetup(ctx, socket); err != nil {
 			e.shell.Errorf("Failed to start kubernetes socket client: %v", err)
 			return 1
 		}
 		defer func() {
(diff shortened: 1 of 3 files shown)


internal/job/executor.go
```diff
@@ -1185,6 +1186,57 @@ func (e *Executor) setupRedactors() {
...
+	// So we should skip setting any vars that are already set, and
+	// specifically any that could be deliberately *unset* by the
+	// k8s stack (BUILDKITE_PLUGINS could be unset if kubernetes is
+	// the only "plugin" in the step).
+	// (Maybe we could move some of the k8s stack processing in here?)
+	//
+	// To think about: how to obtain the env vars early enough to set
+	// them in ExecutorConfig (because of how urfave/cli works, it
+	// must happen before App.Run, which is before the program even knows
+	// which subcommand is running).
```
zhming0, line 1222: Just for my knowledges, what environment variables were we trying to set "early"? 

For our original goal of BUILDKITE_AGEND_ID, does this setup work?

zhming0 (approved):
The change looks good. Hopefully the comprehensive code comment can help future traveller to understand 🤞🏿 .

internal/job/executor.go
DrJosh9000: Anything that normally configures the bootstrap subcommand, but especially the required flags/vars: `BUILDKITE_JOB_ID`, `BUILDKITE_REPO`, etc. I would have expected the command container to use these without having to set the env vars for them at schedule time. But that didn't work.

For the original goal of agent env vars that aren't known when the pod is scheduled, yes, this should still work. Commands, plugins etc would be run with the `e.shell.Env`, but for completeness this PR still inserts them into the current process env (`os.Setenv`).

PR #2878, opened 2024-07-08
Fix some common artifact download bugs

### LINKED ISSUE #2774: Downloading multiple artifacts with the same file name corrupts them.
We've recently diagnosed a situation where artifacts appeared to be getting corrupted and discovered it's when multiple are created with the same name in different steps, then downloaded to a final step.

eg. 

Step 1:    Uploads artifacts:
 *   package_amd64.deb
 *   package_all.deb

Step 2:    Uploads artifacts:
 *   package_arm64.deb
 *   package_all.deb

Step 3:   Downloads all artifacts from step 1 and 2: 
 *   package_amd64.deb
 *   package_arm64.deb
 *   package_all.deb    <-  will be corrupted combination of the two.

This took a while to diagnose because the corrupted file is the correct size and because it's a mixing of two similar files (in our case they were supposedly identical but because some file date stamps were different, the compressed content was unusable).

The buildkite agent code that downloads artifacts from S3 appears to grab a list of all artifacts and downloads them in parallel, it looks likely that this is where the problem happens.

I realise creating multiple artifacts with the same name is not recommended, but it's a very easy mistake to make.

Is it possible to either fix this behaviour or at least make it clearer in the documentation that this is an easy to make mistake.

DrJosh9000:
### Description

While investigating the artifact download process a bit more, I discovered some fairly low-hanging fruit in the existing downloader.

### Context

Fixes #2774 by making one or another artifact the "winner", and printing a SHA256 sum of the content that it downloaded.

https://coda.io/d/_dHnUHNps1YO#View-3-of-Escalations-Table_tu3KF/r437&view=modal

### Changes

- Write to a temp file and move it to the destination name when ready
- Check the error return from Close
- Print a SHA256 sum of the content
- Add ability to verify SHA256 checksum (not yet used)
- Make error strings more Gothic

### Testing

### DIFF
--- a/agent/download.go
+++ b/agent/download.go
@@ -2,6 +2,8 @@ package agent
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"fmt"
 	"io"
 	"net/http"
@@ -37,6 +39,9 @@ type DownloadConfig struct {
 
 	// How many times should it retry the download before giving up
 	Retries int
+
+	// Hexadecimal(SHA256(content)) used to verify the downloaded contents, if not empty
+	WantSHA256 string
 
 	// If failed responses should be dumped to the log
 	DebugHTTP bool
@@ -114,11 +119,11 @@ func targetPath(ctx context.Context, dlPath, destPath string) string {
 }
 
 func (d Download) try(ctx context.Context) error {
-	targetFile := targetPath(ctx, d.conf.Path, d.conf.Destination)
-	targetDirectory, _ := filepath.Split(targetFile)
+	targetPath := targetPath(ctx, d.conf.Path, d.conf.Destination)
+	targetDirectory, targetFile := filepath.Split(targetPath)
 
 	// Show a nice message that we're starting to download the file
-	d.logger.Debug("Downloading %s to %s", d.conf.URL, targetFile)
+	d.logger.Debug("Downloading %s to %s", d.conf.URL, targetPath)
 
 	request, err := http.NewRequestWithContext(ctx, "GET", d.conf.URL, nil)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


agent/download.go
```diff
@@ -158,23 +163,47 @@ func (d Download) try(ctx context.Context) error {
...
+	// On Unix-like platforms this is generally an "atomic replace".
+	// Caveats: https://pkg.go.dev/os#Rename
+	if err := os.Rename(temp.Name(), targetPath); err != nil {
+		return fmt.Errorf("renaming temp file to target (%T: %w)", err, err)
+	}
+
+	gotSHA256 := hex.EncodeToString(hash.Sum(nil))
+
+	// If the downloader was configured with a checksum to check, check it
+	if d.conf.WantSHA256 != "" && gotSHA256 != d.conf.WantSHA256 {
```
wolfeidau, line 202: Should the verification happen before you move the file?

wolfeidau (changes requested):
Just one query.

agent/download.go
DrJosh9000: Yes, it should! Good pickup.

wolfeidau (approved):

jim-barber-he:
This change has resulted in permissions changes of the artifacts where group and world permissions bits are lost/gone.
It used to use `os.Create` with combine 0666 with the UMASK, while `os.CreateTemp` will lock permissions down since temporary files are only accessible by the user creating them.

This caused an outage for us because our application build doesn't make sure the files have appropriate permissions on them.
See:
https://forum.buildkite.community/t/permissions-on-artifacts-changed-with-version-3-75-0-of-buildkite-agent/3803

DrJosh9000:
Thanks for pointing this out @jim-barber-he, the permissions change was definitely unintentional. #2894 to fix it should be in the next release.

PR #2884, opened 2024-07-09
Reject more secrets

DrJosh9000:
### Description

Reject secrets originating from more places (e.g. within the pipeline YAML being uploaded)

### Context

With the flag enabled, env vars containing secrets are already rejected, but only if they are in the process env of the agent uploading a pipeline.

A pipeline generator might accidentally insert secrets as env vars within the pipeline env block or a step env block. These are still secrets, so they should be found and rejected.

### Changes

Each commit should be fairly self contained. 

Main changes that ought to be reasoned about:

- Considering all env maps for secrets. This is the main change, and involves recursively inspecting the pipeline for things with `env` blocks (pipeline or command), then adding the vars and values found to the input to `redact.Vars`.
- Not scanning for pipeline / step env vars. By definition pipeline and step env vars are already in the pipeline. So we don't have to scan the JSON version of the pipeline for them.
- Skipping pure variable substitutions. Some secrets aren't actually secrets until runtime interpolation happens, so any env var that consists entirely of one or more variable substitutions without defaults can be skipped from consideration as a secret.

Refactors / tidyups:

- Preferring `if err != nil { /* check for warning */ }` over `if warning { } else if err != nil { }`
- Moving the resolution of git commit to its own function
- Making the redact package not responsible for logging (about malformed patterns or short variables), and simplifying away `Values`.
- Adding `env.Pair` and `(*env.Environment).DumpPairs` as an alternative to dumping as a map.  Representing env vars as a slice of pairs is necessary in case a user defines the same env var with different values in multiple `env` blocks.

### Testing

### DIFF
--- a/clicommand/pipeline_upload.go
+++ b/clicommand/pipeline_upload.go
@@ -10,6 +10,7 @@ import (
 	"os/exec"
 	"path"
 	"path/filepath"
+	"slices"
 	"strings"
 	"time"
 
@@ -29,6 +30,7 @@ import (
 	"github.com/buildkite/go-pipeline/signature"
 	"github.com/buildkite/go-pipeline/warning"
 	"github.com/urfave/cli"
+	"golang.org/x/exp/maps"
 	"gopkg.in/yaml.v3"
 )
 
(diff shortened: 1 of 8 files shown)


wolfeidau (approved):
Looks good, great to see the extraction of that logic and error tidy.

PR #2892, opened 2024-07-18
Log in to buildkite packages right before pushing images

moskyb:
### Description

Currently, we `docker login` to the buildkite packages container registry in the same place as all the other docker logins, in [`.buildkite/steps/publish_docker_images.sh`](https://github.com/buildkite/agent/blob/main/.buildkite/steps/publish-docker-images.sh). However, because this docker login is done using an OIDC token, it's timed, and only lasts five minutes.

If pushing the docker images to registries before BK packages takes longer than five minutes, the push to BK packages will fail, as the token will expire - see [here](https://buildkite.com/buildkite/agent-release-edge/builds/1315#0190c3aa-4ada-4828-ba5e-46765578cb1b).

This PR updates the docker push process so that we log in to buildkite packages as close as possible to push-time.


swebb (approved):
🙌🏻

PR #2894, opened 2024-07-22
Fix downloaded artifact permissions

DrJosh9000:
### Description

The change from `os.Create` to `os.CreateTemp` changed the file permissions of the created file (from 0o666 to 0o600). This was unintentional. 

### Context

#2878 

### Testing


wolfeidau (approved):
👍🏻 looks good

PR #2908, opened 2024-07-30
fix enable http/2 by default as intended by flags

wolfeidau:
### Description

At some point the agent HTTP client configuration was updated however this resulted in it falling back to HTTP/1.1. Ideally we would take advantage of HTTP/2 to improve overall performance, especially for agents who are connecting from locations outside the US.

This will also lead to less TLS handshakes, and connections overall, which in turn should reduce the likelyhood of stalls uploading artifacts.

### Context

The defaulting to HTTP 1.1 was discovered while debugging higher than expected latency while uploading artifacts.

### Changes

Use the default transport to ensure that the `ForceAttemptHTTP2` is set and we use HTTP/2.0. Cloning the default transport is the simplest approach to ensure you get the most up to date configuration. 

See https://github.com/golang/go/issues/26013 for some background on the `Clone` method in the default transport.

From https://github.com/golang/go/blob/master/src/net/http/transport.go#L50.
```
	// ForceAttemptHTTP2 controls whether HTTP/2 is enabled when a non-zero
	// Dial, DialTLS, or DialContext func or TLSClientConfig is provided.
	// By default, use of any those fields conservatively disables HTTP/2.
	// To use a custom dialer or TLS config and still attempt HTTP/2
	// upgrades, set this to true.
	ForceAttemptHTTP2 bool
```

To disable HTTP/2.0 we need to update a few things, these are documented in this change as well.

### Testing

Before.

```
2024-07-30 13:25:03 DEBUG  POST https://agent.buildkite.com/v3/connect command=start agent_version=3.75.1+x..dirty
2024-07-30 13:25:04 DEBUG  ↳ POST https://agent.buildkite.com/v3/connect command=start proto=HTTP/1.1 status=200 Δ=743.567041ms agent_version=3.75.1+x..dirty
```

After.

```
2024-07-30 13:24:44 DEBUG  POST https://agent.buildkite.com/v3/connect command=start agent_version=3.75.1+x..dirty
2024-07-30 13:24:44 DEBUG  ↳ POST https://agent.buildkite.com/v3/connect command=start proto=HTTP/2.0 status=200 Δ=733.139667ms agent_version=3.75.1+x..dirty
```

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -5,6 +5,7 @@ package api
 import (
 	"bytes"
 	"context"
+	"crypto/tls"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -47,6 +48,9 @@ type Config struct {
 
 	// The http client used, leave nil for the default
 	HTTPClient *http.Client
+
+	// optional TLS configuration primarily used for testing
+	TLSConfig *tls.Config
 }
 
 // A Client manages communication with the Buildkite Agent API.
@@ -85,6 +89,7 @@ func NewClient(l logger.Logger, conf Config) *Client {
 			IdleConnTimeout:     90 * time.Second,
 			TLSHandshakeTimeout: 30 * time.Second,
 			ForceAttemptHTTP2:   !conf.DisableHTTP2, // HTTP2 is enabled by default
+			TLSClientConfig:     conf.TLSConfig,
 		}
 
 		httpClient = &http.Client{
--- a/api/client.go
+++ b/api/client.go
@@ -86,7 +86,7 @@ func NewClient(l logger.Logger, conf Config) *Client {
 			tr.TLSNextProto = make(map[string]func(authority string, c *tls.Conn) http.RoundTripper)
 			// The default TLSClientConfig has h2 in NextProtos, so the negotiated TLS connection will assume h2 support.
 			// see https://github.com/golang/go/issues/50571
-			tr.TLSClientConfig = nil
+			tr.TLSClientConfig.NextProtos = []string{"http/1.1"}
 		}
 
 		tr.TLSClientConfig = conf.TLSConfig
--- a/api/client.go
+++ b/api/client.go
@@ -5,7 +5,6 @@ package api
 import (
 	"bytes"
 	"context"
-	"crypto/tls"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


SorchaAbel (approved):
nice one!!

DrJosh9000 (approved):
Nice find!

yob:
Fascinating! Do we know what version we regressed back to HTTP 1.1 by default? Might be a good detail to flag with our support friends so they can keep an eye out for opportunities to suggest upgrading.

> discovered while debugging higher than expected latency while uploading artifacts

Do you suspect this is a cause of slow artifact uploads? Or is this a fix the you noticed while debugging, but slow artifact uploads still possible?

DrJosh9000:
Appears to have been this way since https://github.com/buildkite/agent/pull/1020

api/client.go
```diff
@@ -74,28 +76,26 @@ func NewClient(l logger.Logger, conf Config) *Client {
...
-				KeepAlive: 30 * time.Second,
-			}).DialContext,
-			MaxIdleConns:        100,
-			IdleConnTimeout:     90 * time.Second,
-			TLSHandshakeTimeout: 30 * time.Second,
-		}
+
+		// use the default transport as it is optimized and configured for http2
+		// and will avoid accidents in the future
+		tr := http.DefaultTransport.(*http.Transport).Clone()
```
DrJosh9000, line 82: Making a note in case this is important later on - no changes needed.

The diff from the old transport to the default is:

```diff
- TLSHandshakeTimeout:   30 * time.Second,
+ ForceAttemptHTTP2:     true,
+	TLSHandshakeTimeout:   10 * time.Second,
+	ExpectContinueTimeout: 1 * time.Second,
```

and the use of  `defaultTransportDialContext` wrapping `net.Dialer`, but that is equivalent to `.DialContext`.

DrJosh9000 (approved):
🌴😎🍹

api/client.go
```diff
@@ -74,28 +76,26 @@ func NewClient(l logger.Logger, conf Config) *Client {
...
 
 		if conf.DisableHTTP2 {
-			t.TLSNextProto = make(map[string]func(string, *tls.Conn) http.RoundTripper)
+			tr.ForceAttemptHTTP2 = false
+			tr.TLSNextProto = make(map[string]func(authority string, c *tls.Conn) http.RoundTripper)
+			// The default TLSClientConfig has h2 in NextProtos, so the negotiated TLS connection will assume h2 support.
+			// see https://github.com/golang/go/issues/50571
+			tr.TLSClientConfig = nil
```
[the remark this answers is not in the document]
wolfeidau: Good pickup thanks!

api/client.go
wolfeidau: Good callout, thanks!

PR #2910, opened 2024-07-31
Pass cancel grace period to bootstrap

DrJosh9000:
### Description

The `buildkite-agent bootstrap` forked from `buildkite-agent start` needs both cancel-grace-period and signal-grace-period.

### Context

In #2899 the cancel grace period was added as a flag and env var read by the bootstrap subcommand, but I forgot to set it from the job runner.

### Changes

* Pass the value through an env var just like `BUILDKITE_SIGNAL_GRACE_PERIOD_SECONDS`
* Also a small cleanup across the whole agent codebase to replace `fmt.Sprintf("%d", ...)` with `strconv.Itoa(...)` (does the same thing, but it doesn't need the `fmt` package to parse a format string).

Note: the two env vars are different due to historical reasons. See https://github.com/buildkite/agent/blob/main/clicommand/cancel_signal.go

### Testing
- [x] Manually checked that it works

### DIFF
--- a/agent/agent_worker.go
+++ b/agent/agent_worker.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"math/rand/v2"
 	"net/http"
+	"strconv"
 	"sync"
 	"time"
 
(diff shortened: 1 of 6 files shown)


wolfeidau (approved):
👍🏻 much better using `strconv.Itoa` 🚀

PR #2913, opened 2024-08-01
Only override TLSClientConfig if set

DrJosh9000:
A hopefully minor bug: `TLSClientConfig` is overridden with `conf.TLSConfig` in all cases, when we should only really override it if it is set.

If `DisableHTTP2` is enabled, we want the default `TLSClientConfig` with `NextProtos` set in a particular way. But the unconditional override with `conf.TLSConfig` nukes it.


wolfeidau (approved):
👍🏻

PR #2915, opened 2024-08-01
agent/core refactor

DrJosh9000:
### Description

Refactor various core parts into a new package called `core` (which can be imported with an alias to disambiguate as needed, i.e. `import agentcore "github.com/buildkite/agent/v3/core"`). Expose a new `Controller` that can handle registering, connecting, disconnnecting, acquiring, and starting, and `JobController` that can handle uploading logs and finishing.

### Context

agent-stack-k8s needs to spin up an entire pod, with an agent container, command container, and init container to copy the agent binary... just to post an error back to Buildkite under the job.

That's too much mechanism. All it really needs is a small sequence of REST API calls:

* register an agent (with the token, that it has)
* connect to the API
* acquire a job (it has the ID)
* start the job
* upload 1 (one) log chunk (containing the error message)
* finish the job
* disconnect

This could be wrapped up into a single function, `RegisterAndAcquireJobAndPostALogAndExitWithStatus`. Or...

### Changes

* Create `core`
* Move various parts of `agent_worker.go`, `job_runner.go`, `log_streamer.go`, and `run_job.go` into methods of `core.Client`.
* Add `Controller` and `JobController` for managing the lifetime of an acquired job.
* Oh look, turns out `LogStreamerChunk` is redundant - `api.Chunk` is practically identical.

### Testing


agentlib/miniagent.go
dabarrell, line 1: Is there a convention around filenames in go -> should this be `mini_agent.go` instead? Just noticing that `JobHandler` is `job_handler.go`, and similar for `AgentWorker` etc

DrJosh9000: I thought about it a bit, and decided that these new types are more "controller"-like, so `MiniAgent` is now `Controller` and `JobHandler` is now `JobController`. File names updated.

wolfeidau (approved):
👍🏻 great work breaking this up

PR #2925, opened 2024-08-07
Write hooks into new tempdir

### LINKED ISSUE #2916: `/tmp/buildkite-agent-hook-wrapper` permission problems
I bumped buildkite-agent from 3.59.0 to 3.76.1, and suddenly jobs started failing with the following error message:

```
🚨 Error: Error creating hook script: failed to create temporary file "hook-before-env": open /tmp/buildkite-agent-hook-wrapper/hook-before-env678448514: permission denied
^^^ +++
🚨 Error: Error setting up job executor: failed to create temporary file "hook-before-env": open /tmp/buildkite-agent-hook-wrapper/hook-before-env678448514: permission denied
^^^ +++
```

Upon further inspection, the issue is due to `/tmp/buildkite-agent-hook-wrapper` being owned by the wrong user. 

We have multiple buildkite-agent instances running on this host, each as its own linux user. So the first one started up and ceated `/tmp/buildkite-agent-hook-wrapper` as its user, and then the next step run by another agent fails to write its hook in there.

Looking in the history a bit, this might have been a regression introduced by a798bab35764f3f5ea423061db3dda2373fc2522, essentially not using a unique tempdir for each invocation anymore.

Use [os.MkdirTemp](https://pkg.go.dev/os#MkdirTemp) to create a temporary dir, and drop the hooks in there, which shouldn't conflict with each other.

For now, we resorted to giving each buildkite-agent systemd service its own private `/tmp`, by setting `PrivateTmp=yes` in the systemd unit, though this might not work for everyone.

DrJosh9000:
### Description

Using a fixed path to contain all hook wrappers makes life hard for some users.

### Context

Fixes #2916 

### Changes

* Make `tempfile.WithDir` accept paths that are absolute or relative (to `os.TempDir`).
* Create one temporary directory with `os.MkdirTemp` inside `NewWrapper`.
* Clean up by removing the whole temporary directory, rather than individual files.

### Testing

### DIFF
--- a/internal/job/hook/wrapper.go
+++ b/internal/job/hook/wrapper.go
@@ -121,6 +121,7 @@ type WrapperOpt func(*Wrapper)
 type Wrapper struct {
 	hookPath      string
 	os            string
+	tempDir       string
 	wrapperPath   string
 	beforeEnvPath string
 	afterEnvPath  string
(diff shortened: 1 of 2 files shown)


internal/job/hook/wrapper.go
```diff
@@ -202,19 +203,33 @@ func NewWrapper(opts ...WrapperOpt) (*Wrapper, error) {
...
 
-	if wrap.beforeEnvPath, err = tempfile.NewClosed(
-		tempfile.WithDir(hookWrapperDir),
+	// On systems where multiple buildkite-agents are running under different
+	// users, a shared path could be owned by a different user.
+	// Creating a new temporary directory to hold the temporary files avoids
+	// this issue and makes cleanup easier.
+	tempDir, err := os.MkdirTemp(os.TempDir(), hookWrapperDir)
+	if err != nil {
+		return nil, err
```
wolfeidau, line 212: Should we `fmt.Errorf` this errors from this package can be very terse?

wolfeidau (changes requested):
Just a question about error wrapping.

internal/job/hook/wrapper.go
DrJosh9000: Done.

wolfeidau (approved):
👍🏻

PR #2926, opened 2024-08-08
Add a bit more context to the debugging for failing signature verify

wolfeidau:
### Description

Working with customers to debug signature verification and one issue found was it could be hard to locate debug messages which highlight specific failure cases. 

### Context

This is security code, so we need to be suitably vague when reporting errors while in production, however for those looking to enable debugging and troubleshoot issues it would be good to provide some consistent patterns to look for in logs.

### Changes

Following existing log messages within the agent I have added a bit more context to the debug logging prefix making it easier to filter and locate in logs.

### Testing


DrJosh9000 (approved):
✅

PR #2933, opened 2024-08-15
fix for layout issues with log messages

wolfeidau:
### Description

When the group expansion characters are printed `^^^ +++`  whilst also using ansi colours, our log rendering in the UI does not expand the group, so the message is hidden.

### Context

Changes in https://github.com/buildkite/agent/pull/2805 a while back adjusted the structure of the messages slightly when using ansi color, this resulted in some formatting issues.

With @tessereth expert guidance we found the issue and added a test to validate it.

### Changes

Just moved the `^^^ +++` sequence out of the ansi formatting.

### Testing


jradtilbrook (approved):
I was hoping/expecting it might be something simple like this 🙏

DrJosh9000 (approved):
LGTM

PR #2936, opened 2024-08-19
Enable configuration for no local hook or plugin failure behavours

wolfeidau:
### Description

Currently if you block local hooks or plugins the build fails, however there are cases during migration where you may want to just print a warning for the developer while still not executing hooks or plugins.

### Context

This change enables teams to block local hooks without breaking the build.

### Changes

This change adds a couple of new flags which allow configuration of the failure behaviour to either "error" or "warn".

### Testing


DrJosh9000:
This is a nice idea, but the PR is stale. Please resuscitate if you think this branch still has legs @wolfeidau

PR #2942, opened 2024-08-20
k8s: Unconditionally set BUILDKITE_AGENT_ACCESS_TOKEN

DrJosh9000:
### Description

Unconditionally set `BUILDKITE_AGENT_ACCESS_TOKEN`, regardless of whether it is already set.
This matches pre-#2851 behaviour.

### Context

Manually setting `BUILDKITE_AGENT_ACCESS_TOKEN` within the command container was found to break uses of many agent subcommands (e.g. `meta-data set`) within jobs run on newer agents (v3.74.1 and later) in the k8s stack.

https://coda.io/d/Escalations-Feedback_dHnUHNps1YO/Pipelines-Escalations_su7FT#Pipelines-Escalations-Board_tu__K/r597&view=modal

### Testing


CerealBoy (approved):
🚀 LGTM

PR #2945, opened 2024-08-21
Allow `buildkite-agent` to run a job when JWK is unavailable but failure behaviour is set to `warn`

CheeseStick:
### Context

Recently, while working on an agent migration plan, I noticed that the agent without JWK configuration is not compatible with an agent with JWK, even when the failure behaviour is set to `warn`.

Because the agent that started with the JWK configuration adds a job signature to the pipeline's job when it uploads, a requested job can fail if it lands on the agent that doesn't have a JWK key configuration.

### Changes

This PR updates a job runner not to reject a signed job when JWK is not configured and `verification-failure-behaviour` is set to `warn`. It also adds a new error, `SignalReasonUnableToVerifySignature`, to the job runner so the misconfiguration can be monitored.

### Testing
image


wolfeidau:
@CheeseStick thanks for posting this PR, we have just merged some changes which overlap with this code.

The fix for missing JWKS and warnings is a good pickup, we need to impove how we handle this, however the change to SignalReasonUnableToVerifySignature is something I need to dig into further as there are some security concerns.

That said having some specific log messages which are logged in the agent is what we are leaning towards.

Let me pull this down and do some testing.

CheeseStick:
> @CheeseStick thanks for posting this PR, we have just merged some changes which overlap with this code.
> 
> 
> 
> The fix for missing JWKS and warnings is a good pickup, we need to impove how we handle this, however the change to SignalReasonUnableToVerifySignature is something I need to dig into further as there are some security concerns.
> 
> 
> 
> That said having some specific log messages which are logged in the agent is what we are leaning towards.
> 
> 
> 
> Let me pull this down and do some testing. 

Thanks!!! 🙇

wolfeidau:
We have reviewed this soluton, and based on recent exprience, this change also makes it clearer when builds fail because agents aren't configured with a key, overall this is a plus.

wolfeidau (approved):
Looks good 👍🏻

Issue #2941, opened 2024-08-19
Add executable permissions to directories when attempting to clean up build directory


staticfloat:
## Problem

If a build does something silly like `chmod 0444 ./subdir`, attempts to clean up the build directory [will fail with messages like](https://buildkite.com/julialang/julia-master/builds/39322#01916b11-6bff-4a23-a6e8-dac68df78347/48-113):

```
# Removing /Users/julia/buildkite-agent/build/julia-master
🚨 Error: Failed to remove "/Users/julia/buildkite-agent/build/julia-master" (fstatat /Users/julia/buildkite-agent/build/julia-master/subdir/Manifest.toml: permission denied)
```

Then, subsequent git checkouts will fail (because that build directory is not empty) and the cleanup process will fail all over again.  In our case, this happened due to a misconfigured `find . -name \*.jl -exec chmod 0444 {} \;` which found directories with names that ended in `.jl` instead of source code files like we intended.  Unfortunately, this then required manual intervention on the build agent machines to clear out those bad permissions to allow the build to successfully go through.

## Potential solution

I propose that when buildkite is attempting to clear out the build directory, that it should add execute (and write) permissions to all subdirectories, e.g. the equivalent of:
```
find . -type d -exec chmod u+wx {} \;
```

This will allow normal directory removal code to then process the repository.  I don't think there's any valid usecase for allowing user code to create things that cannot be deleted like this, as it badly breaks the ability of buildkite to deal with the build directory.

CerealBoy:
Hi @staticfloat, this sounds like a reasonable idea to an interesting problem, though we would be cautious to add this type of change to the agent as it could easily have a negative impact on other customers unique setups.

Rather, this could be resolved using a [hook](https://buildkite.com/docs/agent/v3/hooks) to make the permission changes required for this specific scenario. It could be part of the `pre-checkout` or `post-command` hook to make the permission changes required. Would this be suitable for your situation?

staticfloat:
> though we would be cautious to add this type of change to the agent as it could easily have a negative impact on other customers unique setups.

Can you explain how this could have a negative impact?  I don't quite see how this could have a negative impact, because as far as I can tell, if buildkite attempts to remove the build directory (due to a failed checkout or any other reason) and fails, the agent then enters a state that it cannot recover from.  If buildkite is already attempting to delete the entire build directory, I can't see why it would be harmful to modify permissions of the files about to be deleted.

[...]

PR #2946, opened 2024-08-22
Improvements to removeCheckoutDir

DrJosh9000:
### Description

Ensure that, when removing all files and directories for a cleanup, that directories with weird permissions (e.g. lacking execute bit) are cleaned up.

### Context

#2941

### Changes

- Add `hardRemoveAll` helper function that tests for permissions problems after failing to `os.RemoveAll`, in a loop
- Replace some calls to `os.RemoveAll` with a call to `hardRemoveAll`

### Testing

### DIFF
--- a/internal/job/checkout.go
+++ b/internal/job/checkout.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io/fs"
 	"os"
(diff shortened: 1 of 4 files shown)


DrJosh9000:
Closing due to staleness - I'll rewrite it better if the need arises.

PR #2948, opened 2024-08-26
 Fix error when hook cannot be run due to missing interpreter

mcncl:
### Description
There is a missing value in the formatted string for the error, it is passing the interpreter as the string value for the hook name, and is then printing an empty `%q` for the interpreter value.

### Changes
There is an addition of the `hookName` passed as a string to the `logMissingHookInfo` function. This value is passed from the `runWrappedShellScriptHook` function on [534](https://github.com/buildkite/agent/blob/65ed6aaf3dd3e50f88dee3fbd27714120e558965/internal/job/executor.go#L534).

### Testing

### DIFF
--- a/internal/job/executor.go
+++ b/internal/job/executor.go
@@ -436,7 +436,7 @@ func logOpenedHookInfo(l shell.Logger, debug bool, hookName, path string) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


DrJosh9000 (approved):
Thanks! Good catch!

PR #2958, opened 2024-08-28
Add command for canceling a running a build

dannymidnight:
### Description
This PR drops in a new CLI command for canceling a running build. 

**Note: Currently blocked on agent endpoint being added**

Example usage:

```sh
BUILDKITE_BUILD_ID=1234 buildkite-agent build cancel
``` 

### Testing

### DIFF
--- a/clicommand/build_cancel.go
+++ b/clicommand/build_cancel.go
@@ -104,7 +104,7 @@ func cancelBuild(ctx context.Context, cfg BuildCancelConfig, l logger.Logger) er
 			return err
 		}
 
-		l.Info("Successfully cancelled build %s", build.ID)
+		l.Info("Successfully cancelled build %s", build.UUID)
 		return nil
 	}); err != nil {
 		return fmt.Errorf("failed to cancel build: %w", err)
--- a/clicommand/build_cancel.go
+++ b/clicommand/build_cancel.go
@@ -1,0 +1,114 @@
+package clicommand
+
+import (
+	"context"
+	"fmt"
+
+	"time"
+
+	"github.com/buildkite/agent/v3/api"
+	"github.com/buildkite/agent/v3/logger"
+	"github.com/buildkite/roko"
+	"github.com/urfave/cli"
+)
+
+const buildCancelDescription = `Usage:
+
+    buildkite-agent build cancel [options...]
+
+Description:
+
+Cancel a running build.
+
+Example:
+
+    # Cancels the current build
+    $ buildkite-agent build cancel
+
+    # Cancel a different build
+    $ buildkite-agent build cancel --build "01234567-89ab-cdef-0123-456789abcdef"`
+
+type BuildCancelConfig struct {
+	Build string `cli:"build" validate:"required"`
+
+	// Global flags
+	Debug       bool     `cli:"debug"`
+	LogLevel    string   `cli:"log-level"`
+	NoColor     bool     `cli:"no-color"`
+	Experiments []string `cli:"experiment" normalize:"list"`
+	Profile     string   `cli:"profile"`
+
+	// API config
+	DebugHTTP        bool   `cli:"debug-http"`
+	AgentAccessToken string `cli:"agent-access-token" validate:"required"`
+	Endpoint         string `cli:"endpoint" validate:"required"`
+	NoHTTP2          bool   `cli:"no-http2"`
+}
+
+var BuildCancelCommand = cli.Command{
+	Name:        "cancel",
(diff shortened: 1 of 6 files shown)


dabarrell:
> BUILDKITE_BUILD_ID=1234 buildkite-agent build cancel

Noting that accepting `BUILDKITE_BUILD_ID` will mean that `buildkite-agent build cancel` will cancel the current build if no build ID is specified

dabarrell:
CleanShot 2024-08-28 at 12 18 40@2x

Oh, `BUILDKITE_BUILD_ID` is a uuid, not an ID

clicommand/build_cancel.go
```diff
@@ -0,0 +1,114 @@
...
+
+    buildkite-agent build cancel [options...]
+
+Description:
+
+Cancel a running build.
+
+Example:
+
+    $ buildkite-agent build cancel --build "1234"`
```
DrJosh9000, line 25: Maybe make it clear this should work within the current build (if there is one)?

```suggestion
    # Cancels the current build
    $ buildkite-agent build cancel
    # Cancel a different build
    $ buildkite-agent build cancel --build "01234567-89ab-cdef-0123-456789abcdef"`
```

clicommand/build_cancel.go
```diff
@@ -0,0 +1,114 @@
...
+		LogLevelFlag,
+		ExperimentsFlag,
+		ProfileFlag,
+	},
+	Action: func(c *cli.Context) error {
+		ctx := context.Background()
+		ctx, cfg, l, _, done := setupLoggerAndConfig[BuildCancelConfig](ctx, c)
+		defer done()
+
+		if err := cancelBuild(ctx, cfg, l); err != nil {
```
DrJosh9000, line 74: Well that's just

```suggestion
		return cancelBuild(ctx, cfg, l)
```

You could consider lifting the whole action out of the struct literal, e.g.

```go
    Action: cancelBuildAction,
}

func cancelBuildAction(c *cli.Context) error {
   ctx := context.Background()
   ctx, cfg, l, _, done := setupLoggerAndConfig[BuildCancelConfig](ctx, c)
   defer done()
   
   // body of cancelBuild here
}
```

dannymidnight: > You could consider lifting the whole action out of the struct literal, e.g.

I had a quick play but couldn't find a nice way to inject test config into the cli context. I think for now I'll leave them as two seperate functions if you're alright with that?

clicommand/build_cancel.go
```diff
@@ -0,0 +1,114 @@
...
+
+Cancel a running build.
+
+Example:
+
+    # Cancels the current build
+    $ buildkite-agent build cancel
+
+    # Cancel a different build
+    $ buildkite-agent build cancel --build "01234567-89ab-cdef-0123-456789abcdef"`
```
dannymidnight, line 29: @Based on @DrJosh9000's suggestion, is this clear enough for you @dabarrell that we require a UUID?

thejcannon:
🎵  Hello from the other side! 🎵 

How does this relate to https://buildkite.com/docs/apis/rest-api/builds#cancel-a-build? I'm guessing this will eventually be hooking itself into (via the agent endpoints?) that eventual server codepath?

If so, might be nice to put in PR description to disambiguate (which may be a "me" problem because I was linked here and mildly was expecting something of the same shape as "cancel as failing", which this may be a precursor to, but isn't 1:1 that feature just yet)

dannymidnight:
> How does this relate to https://buildkite.com/docs/apis/rest-api/builds#cancel-a-build? I'm guessing this will eventually be hooking itself into (via the agent endpoints?) that eventual server codepath?

@thejcannon Yep, exactly. It'll be using an agent endpoint, authenticating with an agent token rather than hitting up the bk user REST API. 

> was expecting something of the same shape as "cancel as failing"

Yeah we did discuss briefly whether or not it would make sense adding a "fail" flag as part of this change but landed on a build "failing" being a symptom of having a failed job. I'll have a bit more of a discussion with the team around that specific change, but I suspect it'll be tackled in addition to this change.

clicommand/build_cancel.go
dabarrell: Missed this, sorry! yep!

dannymidnight:
Here's an example of the flow:


https://github.com/user-attachments/assets/81526742-3737-41dd-b9d0-a583098d6686

DrJosh9000 (approved):
LGTM! 🍏

PR #2973, opened 2024-09-05
Updated to support only ECC_NIST_P256 keyspec for initial release

wolfeidau:
### Description

This reduces the support for key types to just `ECC_NIST_P256` for the initial release.

### Context

The reason I removed the other key types are:

1. The RSA types are more complex to configure and use as they support multiple signing algorythms, some of which have issues as identified in https://buildkite.com/docs/agent/v3/signed-pipelines regarding RSASSA-PKCS1 v1.5 signatures.
2. Currently the `ECC_NIST_P256` is the only ECDSA KMS key type tested for `crypto.Signing` with KMS in our JWS library, I will need to raise this upstream to figure out why.

So given the `ECC_NIST_P256` is widely used I think it is fine as a starting point for KMS signing support.


CerealBoy (approved):
LGTM

PR #2974, opened 2024-09-06
Add sha256 checksum output to the formatting options

patrobinson:
### Description

When running `artifact shasum` you can optionally print the SHA256 checksum for a single artifact.
But if you're trying to list multiple artifacts using the `search` subcommand, we only support printing the SHA1 checksum

### Changes

Add new formatting option, `%T`, for SHA256. I chose t because it's the next letter in the alphabet and both `%s` and `%S` are taken. Open to suggestions!

### Testing

image

### DIFF
--- a/clicommand/artifact_search.go
+++ b/clicommand/artifact_search.go
@@ -64,7 +64,7 @@ Format specifiers:
 
   %S    SHA1 checksum of the artifact
 
-	%t    SHA256 checksum of the artifact
+  %T    SHA256 checksum of the artifact
 
   %u    Download URL for the artifact, though consider using 'buildkite-agent artifact download' instead`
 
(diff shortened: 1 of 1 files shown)


clicommand/artifact_search.go
```diff
@@ -64,6 +64,8 @@ Format specifiers:
 
   %S    SHA1 checksum of the artifact
 
+	%t    SHA256 checksum of the artifact
```
isaacsu, line 67: spaces rather than tab to match the other lines

```suggestion
  %t    SHA256 checksum of the artifact
```

clicommand/artifact_search.go
```diff
@@ -162,6 +164,7 @@ var ArtifactSearchCommand = cli.Command{
 				"%j", artifact.JobID,
 				"%s", strconv.FormatInt(artifact.FileSize, 10),
 				"%S", artifact.Sha1Sum,
+				"%t", artifact.Sha256Sum,
```
isaacsu, line 167: Personally I'd go uppercase to make it nearer to `%S` and further from `%s` but 🤷 
```suggestion
				"%T", artifact.Sha256Sum,
```

wolfeidau (approved):
👍🏻 looks good

PR #2977, opened 2024-09-09
Fix for region discovery issue with aws sdkv2 when running in ec2

wolfeidau:
### Description

The new KMS signing support uses aws sdkv2, this automatically discovers region from environment or local configuration files, however it doesn't automatically use imds. This code replicates the v1 local configuration preference with fall back to imds service.

### Testing

### DIFF
--- a/agent/awsv2.go
+++ b/agent/awsv2.go
@@ -1,0 +1,40 @@
+package agent
+
+import (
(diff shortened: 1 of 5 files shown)


DrJosh9000 (approved):
✅

PR #2984, opened 2024-09-10
Explain verification-failure-behavior in more detail

DrJosh9000:
### Description

Fix the usage string to list more cases where "warn" mode continues with the job.

### Context

Got side-tracked and ended up looking at what the behaviour actually was. 

### Testing


wolfeidau (approved):
Good pickup 👍🏻 🚀

PR #2985, opened 2024-09-10
Consolidate artifact functionality in internal package

DrJosh9000:
### Description

#2204 lives!

Artifact-related library functionality isn't intended to be imported and used outside the agent. This moves all the artifact stuff out of `agent` into its own internal package.

Since there are some common AWS things needed from the new artifact package, these are also factored into their own internal package.

### Context

Preparing to work on more artifact things, maybe.

### Changes

* Move artifact code into `internal/artifact`
* Move AWS library code into `internal/awslib`
* Rename artifact package identifiers (generally dropping `Artifact`)
* Fix tests...
  * ...by moving their fixtures, eliminating the need to `os.Chdir`, making the tests parallelizable.

### Testing


CerealBoy (approved):
Nice one!

wolfeidau:
@DrJosh9000 great work!

PR #2986, opened 2024-09-11
Use job tokens for log chunk uploads

tessereth:
In https://github.com/buildkite/agent/pull/2915 we refactored a bunch of this code. In the process, we now take a copy of the `apiClient` object in `NewJobRunner` _before_ we replace the `apiClient` with a new one that uses the job token rather than the session token. We then use the copy for log streaming. This means agents since v3.77.0 use session tokens rather than job tokens for chunk uploads.

~I've tested it locally and this change fixes the problem. I'm not sure how to add regression tests for this though.~ This is tested by updating the fake bk server in the integration tests to require job tokens for everything.

### DIFF
--- a/agent/integration/test_helpers.go
+++ b/agent/integration/test_helpers.go
@@ -230,9 +230,11 @@ func (t *testAgentEndpoint) server(extraRoutes ...route) *httptest.Server {
 			t.mtx.Unlock()
 
 			// We also require job tokens are used for authentication
-			if !strings.HasPrefix(req.Header["Authorization"][0], "Token bkaj_") {
-				fmt.Printf("Invalid authorization header, expected job token, got: %q\n", req.Header["Authorization"][0])
-				rw.WriteHeader(http.StatusUnauthorized)
+			authzHeader := req.Header.Get("Authorization")
+			if !strings.HasPrefix(authzHeader, "Token bkaj_") {
+				msg := fmt.Sprintf("Authorization header = %q, want job token prefix 'Token bkaj_'", authzHeader)
+				fmt.Fprintln(os.Stderr, msg)
+				http.Error(rw, msg, http.StatusUnauthorized)
 				return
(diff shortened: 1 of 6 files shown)


DrJosh9000 (approved):
Whoops! LGTM

agent/integration/test_helpers.go
```diff
@@ -229,6 +229,13 @@ func (t *testAgentEndpoint) server(extraRoutes ...route) *httptest.Server {
 			t.calls[req.URL.Path] = append(t.calls[req.URL.Path], b)
 			t.mtx.Unlock()
 
+			// We also require job tokens are used for authentication
+			if !strings.HasPrefix(req.Header["Authorization"][0], "Token bkaj_") {
+				fmt.Printf("Invalid authorization header, expected job token, got: %q\n", req.Header["Authorization"][0])
+				rw.WriteHeader(http.StatusUnauthorized)
```
DrJosh9000, line 235: Suggestion 1: pull the authz header out of `req.Header` with the `Get` method, which should handle missing headers and canonicalisation

Suggestion 2: I'd serve the error message as the HTTP response body, and prefer stderr to stdout - `http.Error` and `fmt.Fprintf(os.Stderr, ...)` or `log.Printf` functions can help

Suggestion 3: some [tweaks](https://google.github.io/styleguide/go/decisions#got-before-want) to the message formatting

```suggestion
			authzHeader := req.Header.Get("Authorization")
			if !strings.HasPrefix(authzHeader, "Token bkaj_") {
				msg := fmt.Sprintf("Authorization header = %q, want job token prefix 'Token bkaj_'", authzHeader)
				fmt.Fprintln(os.Stderr, msg)
				http.Error(w, msg, http.StatusUnauthorized)
```

DrJosh9000 (approved):
LGTM again, one comment for you

PR #2988, opened 2024-09-12
Add JSON-format env file, allow annotations from pre-bootstrap

### LINKED ISSUE #2962: pre-bootstrap hooks can't annotate builds

A failure in a `pre-bootstrap` script results only in the message "pre-bootstrap hook rejected this job, see the buildkite-agent logs for more details". For our users who don't have access to these logs this means that they are not able to further diagnose their issue, leading to increased support load.

I'd like to write a build annotation (and also something in the user-facing log) describing what failed and how to go about fixing the issue. Trying this though, it appears that the `BUILDKITE_AGENT_ACCESS_TOKEN` is not supplied to the `pre-bootstrap` hook.

Is there an alternative apart from `buildkite-agent annotate` to provide feedback to the user?

I'd like to be able to:

1. Write a message to the log on failure that that the user can see
2. Write a build annotation to give users some context on their issue and help them self-diagnose their issue.

Without this, we either need to execute checks in the `environment` hook, give access to the agent logs (and train engineers to access them), or swallow an increased support burden.

We can do this if we use the `environment` hook to perform checks. This seems to miss the point of the `pre-checkout` hook though.

### LINKED ISSUE #2966: Add JSON equivalent for `BUILDKITE_ENV_FILE` in the `pre-bootstrap` hook

When writing a security-focused script in the `pre-bootstrap` hook, I'm uncomfortable with the potential for naive Bash parsing of a key value file to introduce security issues through mismatched escaping bugs. For example: unicode character handling, newline handling and quote handling all pose the potential to inject new key value pairs into the evaluation through an escaping exploit.

Provide the key/value pairs in JSON as well, using `BUILDKITE_ENV_JSON`. This is an additional option for the script writer, allowing the use of tools like `jq` to confidently parse the data. It also makes it easier for those that use non-script executables for their hooks, as most languages have solid JSON support.

I am using the following (dodgey) JQ transform to create JSON from the env file myself. It works, but I trust it about as far as I can kick it.

```jq
.
| split("\n")
| map(
  .
  | select(. != "")
  | split("=")
  | { key:(.[0]), value:(.[1] | fromjson)}
)
| from_entries
```

**Additional context**

DrJosh9000:
### Description

Two changes, since they touch similar code, done in one:
- Alongside `BUILDKITE_ENV_FILE`, write a `BUILDKITE_ENV_JSON_FILE` which has better-understood escaping and canonicalisation qualities.
- Set enough environment variables in `pre-bootstrap` hook to enable `buildkite-agent annotate` to work.

Opportunistic fixes:

- Turns out `BUILDKITE_NO_HTTP2` wasn't being passed to the bootstrap subcommand. That's fixed too.
- Globally replace `fmt.Sprintf("%t", someBool)` with the equivalent `fmt.Sprint(someBool)`.
- Tweaks to `job_runner_test.go`, including regexes and getting rid of assert/require.

### Context

Fixes #2962
Fixes #2966

### Testing

I've also manually tested with a pre-bootstrap hook that reads `$BUILDKITE_ENV_JSON_FILE` with `jq .`, and annotates the build with it. :+1:

### DIFF
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -2,6 +2,7 @@ package agent
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"os"
@@ -156,8 +157,9 @@ type JobRunner struct {
 	// A lock to protect concurrent calls to cancel
 	cancelLock sync.Mutex
 
-	// File containing a copy of the job env
-	envFile *os.File
+	// Files containing a copy of the job env
+	envShellFile *os.File
+	envJSONFile  *os.File
 }
 
 type jobAPI interface {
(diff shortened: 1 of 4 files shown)


wolfeidau (approved):
👍🏻 🚀 looks good

PR #2990, opened 2024-09-16
Remove mitchellh/go-homedir; it's archived

mckern:
### Description

[Go has caught up to the functionality](https://github.com/mitchellh/go-homedir/issues/34#issuecomment-2011915588) of [mitchellh/go-homedir](https://github.com/mitchellh/go-homedir) since 1.20, and go-homedir has been archived since June. No time like the present to drop a dependency!

### Testing

### DIFF
--- a/internal/job/integration/executor_tester.go
+++ b/internal/job/integration/executor_tester.go
@@ -135,6 +135,7 @@ func NewExecutorTester(ctx context.Context) (*ExecutorTester, error) {
 			"PATHEXT="+os.Getenv("PATHEXT"),
 			"TMP="+os.Getenv("TMP"),
 			"TEMP="+os.Getenv("TEMP"),
+			"USERPROFILE="+os.Getenv("USERPROFILE"),
 		)
 	} else {
 		bt.Env = append(bt.Env, "PATH="+pathDir+":"+os.Getenv("PATH"))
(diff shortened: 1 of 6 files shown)


DrJosh9000 (approved):
Nice! :+1:

DrJosh9000:
Looks like the tests are failing on Windows - I think it's because [mitchellh/go-homedir](https://github.com/mitchellh/go-homedir/blob/af06845cf3004701891bf4fdb884bfe4920b3727/homedir.go#L148-L152) tries to use `HOME` first, but [`os.UserHomeDir`](https://pkg.go.dev/os#UserHomeDir) only consults `USERPROFILE`. The env maps created in internal/job/shell/test.go and internal/job/integration/executor_tester.go would need `USERPROFILE` added.

internal/job/integration/executor_tester.go
```diff
@@ -135,6 +135,7 @@ func NewExecutorTester(ctx context.Context) (*ExecutorTester, error) {
 			"PATHEXT="+os.Getenv("PATHEXT"),
 			"TMP="+os.Getenv("TMP"),
 			"TEMP="+os.Getenv("TEMP"),
+			"USERPROFILE="+os.Getenv("USERPROFILE"),
```
DrJosh9000, line 138: I spent a bit of time scratching my head over why this still didn't work. Turns out the tester creates a temp `homeDir` and sets `HOME` above to that. We should use the same path for `USERPROFILE`.

```suggestion
			"USERPROFILE="+homeDir,
```

PR #2991, opened 2024-09-16
Multipart artifact upload

DrJosh9000:
### Description

Add a multipart-upload mechanism for Buildkite-hosted artifacts, and upload parts in parallel (with each other, and with other artifacts).

### Context

https://linear.app/buildkite/issue/PIPE-180/implement-agent-half-of-multipart-uploads

### Changes

The changes to artifact uploading is substantial, but basically:

* Some tidyups to the existing code, such as context propagation, un-exporting the `Collect` method
* `FormUploader` is renamed to `BKUploader`, since it will manage both form uploads and multipart uploads to BK
* Changing all the existing uploaders to return one or more units of work, which are ultimately responsible for performing an upload
* Running work units concurrently on a pool of worker goroutines, and communicating between goroutines using channels
* Actually implementing multipart uploads:
  * Alter the batch artifact creation API to signal that multipart is supported and enabled
  * Alter the creation response to handle per-artifact instructions and multiple actions per instruction
  * Alter the batch update API to accept part number-ETag mappings
  * Implement multipart upload work units that are responsible for uploading a single part each

### Testing

### DIFF
--- a/agent/agent_configuration.go
+++ b/agent/agent_configuration.go
@@ -43,23 +43,24 @@ type AgentConfiguration struct {
 	VerificationJWKS             any    // The set of keys to verify jobs with
 	VerificationFailureBehaviour string // What to do if job verification fails (one of `block` or `warn`)
 
-	ANSITimestamps             bool
-	TimestampLines             bool
-	HealthCheckAddr            string
-	DisconnectAfterJob         bool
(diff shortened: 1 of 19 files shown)


CerealBoy (approved):
LGTM!

PR #3000, opened 2024-09-17
Prefer $HOME on all platforms

DrJosh9000:
### Description

#2990 is good, but is likely to be a breaking change on Windows particularly. For instance the Elastic CI Stack runs the agent with [`HOME=C:\buildkite-agent`](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/41363708a4231dd801e4c30d4ae2b4b5a6c8ff7c/packer/windows/conf/bin/bk-install-elastic-stack.ps1#L231).

Rather than revert #2990, in the spirit of Go Proverbs I propose we do [a little copying instead of a little dependency](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=9m28s).

### Changes

* Rename internal/utils to internal/osutil, which suits it a little better.
* Add `osutil.UserHomeDir`, which wraps `os.UserHomeDir` and prefers `$HOME` when available.

### Testing

### DIFF
--- a/internal/osutil/homedir.go
+++ b/internal/osutil/homedir.go
@@ -1,0 +1,12 @@
+package osutil
+
+import "os"
+
+// UserHomeDir is similar to os.UserHomeDir, but prefers $HOME when available
+// over other options (such as USERPROFILE on Windows).
+func UserHomeDir() (string, error) {
+	if h := os.Getenv("HOME"); h != "" {
+		return h, nil
+	}
+	return os.UserHomeDir()
+}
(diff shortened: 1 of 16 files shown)


internal/osutil/homedir.go
```diff
@@ -0,0 +1,13 @@
+package osutil
+
+import "os"
+
+// UserHomeDir is similar to os.UserHomeDir, but prefers $HOME when available
+// over other options (such as USERPROFILE on Windows).
+func UserHomeDir() (string, error) {
+	h := os.Getenv("HOME")
+	if h != "" {
```
patrobinson, line 9: Is there any preference of this style over?:
```go
	if h := os.Getenv("HOME"); h != "" {
```

patrobinson (approved):

moskyb (approved):

internal/osutil/homedir.go
DrJosh9000: No, in fact I think we _should_ prefer it that way. Just hammered out the code quickly.

PR #3017, opened 2024-09-25
Refactor the various agent HTTP clients

DrJosh9000:
### Description

It'd sure be be nice to do HTTP clients consistently the same way in the agent. Then we could reuse Pat's trace work, disable HTTP/2 consistently, ...

### Context

Noticed while working on artifacts, which create blank `&http.Client{}`s or use `http.DefaultClient`.

### Changes

* Moves client and transport creation into a new internal package with functional options
* Provide a single wrapper for `Do`, again with functional options
* Use it everywhere
* Some opportunistic removals of `context.Background` and a deprecated method
* Removed some unused `DebugHTTP` fields

Not changed:

* Some artifact upload/download code uses specific clients, e.g. `GSUploader` and `GSDownloader` make use of `google.DefaultClient`. These remain.

### Testing

### DIFF
--- a/internal/agenthttp/client.go
+++ b/internal/agenthttp/client.go
@@ -1,0 +1,136 @@
+// Package agenthttp creates standard Go [net/http.Client]s with common config
+// options.
+package agenthttp
+
+import (
+	"crypto/tls"
+	"net/http"
+	"sync"
+	"time"
+
+	"golang.org/x/net/http2"
+)
+
+// NewClient creates a HTTP client.
+func NewClient(opts ...ClientOption) *http.Client {
+	conf := clientConfig{
+		// This spells out the defaults, even if some of them are zero values.
+		Bearer:     "",
+		Token:      "",
+		AllowHTTP2: true,
+		Timeout:    60 * time.Second,
+		TLSConfig:  nil,
+	}
+	for _, opt := range opts {
+		opt(&conf)
+	}
+
+	cacheKey := transportCacheKey{
+		AllowHTTP2: conf.AllowHTTP2,
+		TLSConfig:  conf.TLSConfig,
+	}
+
+	transportCacheMu.Lock()
+	transport := transportCache[cacheKey]
+	if transport == nil {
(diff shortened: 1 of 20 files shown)


internal/agenthttp/client.go
```diff
@@ -0,0 +1,136 @@
...
+		TLSConfig:  conf.TLSConfig,
+	}
+
+	transportCacheMu.Lock()
+	transport := transportCache[cacheKey]
+	if transport == nil {
+		transport = newTransport(&conf)
+		transportCache[cacheKey] = transport
+	}
+	transportCacheMu.Unlock()
```
patrobinson, line 39: Is the goal by sharing the transports here to also share the connection pooling?

Once side effect is when we call `CloseIdleConnections` that impacts all clients using this transport, even if they have a different http client. I think that's ok?

internal/artifact/uploader.go
```diff
@@ -371,28 +373,30 @@ func (a *Uploader) createUploader() (_ uploader, err error) {
...
+			DebugHTTP:    a.conf.DebugHTTP,
+			TraceHTTP:    a.conf.TraceHTTP,
+			DisableHTTP2: a.conf.DisableHTTP2,
 		}), nil
 
 	case strings.HasPrefix(a.conf.Destination, "s3://"):
 		dest = "Amazon S3"
 		return NewS3Uploader(a.logger, S3UploaderConfig{
 			Destination: a.conf.Destination,
-			DebugHTTP:   a.conf.DebugHTTP,
```
patrobinson, line 381: It seems like this option never did anything?

patrobinson (approved):

internal/agenthttp/client.go
DrJosh9000: Yes, that's the idea, but I expect it will have very limited impact.

internal/artifact/uploader.go
DrJosh9000: Yep, set and never used.

PR #3018, opened 2024-09-25
Migrate Agent Pipeline to Agent Cluster

matthewborden:
### Description

Internally at Buildkite we're refactoring our clusters and will be moving the agent pipeline to a new cluster. This PR makes the queue names override-able to allow running the pipeline in both the previous and new cluster. 

### Context

https://3.basecamp.com/3453178/buckets/1933834/messages/7835428487

### Changes

Queues can be overridden using environment variables from the Pipeline Steps configuration. This allows running the agent pipeline simultaneously in the Agent and existing cluster.

Assumes an OIDC role to publish docker images to internal registry.

### Testing
Successful builds in both the current and new cluster.


DrJosh9000 (approved):
✅

PR #3021, opened 2024-10-02
Ensure all string slice args have whitespace cleaned off of each element

moskyb:
This allows users to use:
```
buildkite-agent oidc request-token --claim 'this, that, the-other'
```
where previously, even if `that` and `the-other` were allowed optional claims, they'd be rejected, as they'd be sent to buildkite with their preceeding whitespace

### Testing

### DIFF
--- a/cliconfig/loader.go
+++ b/cliconfig/loader.go
@@ -421,6 +421,8 @@ func (l Loader) normalizeField(fieldName string, normalization string) error {
 						continue
 					}
 
+					normalized = strings.TrimSpace(normalized)
+
 					normalizedSlice = append(normalizedSlice, normalized)
 				}
 			}
--- a/clicommand/oidc_request_token.go
+++ b/clicommand/oidc_request_token.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"net/http"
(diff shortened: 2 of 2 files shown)


sj26 (approved):
Nice idea!

Want to do the same to `AWSSessionTags`?

DrJosh9000 (approved):
✅

cliconfig/loader.go
```diff
@@ -225,7 +225,11 @@ func (l Loader) setFieldValueFromCLI(fieldName string, cliName string) error {
 				case reflect.String:
 					value = configFileValue
 				case reflect.Slice:
-					value = strings.Split(configFileValue, ",")
+					valueSlice := strings.Split(configFileValue, ",")
```
DrJosh9000, line 228: Good move, it seems very unlikely that any string list will be dependent on whitespace.

moskyb:
@DrJosh9000 i've updated this so that it's the `normalize:"list"` struct tag that does this, rather than the CLI parsing.

i would love to have a couple of days free to unwind `loader.go` and give it some proper tests....

DrJosh9000 (approved):
Even better!

PR #3023, opened 2024-10-03
Exit 94 if a mirror lock times out

DrJosh9000:
### Description

Lockfiles (used for serialising access to git mirrors) aren't totally reliable in the face of agents dying without releasing them. As a stopgap, if a lockfile can't be acquired before the timeout, exit with a distinctive code (94, chosen randomly).

### Context

https://coda.io/d/_dHnUHNps1YO#Escalations-Pipelines_tuQbCBf6/r711

### Changes

* Create a new error type, `ErrTimedOutAcquiringLock`.
* Return it if the git mirror lockfiles fail for `context.DeadlineExceeded`
* Check for it in the checkout phase, and wrap it in a `shell.ExitError`
* Update the one other use of `shell.ExitError`

### Testing

### DIFF
--- a/internal/job/checkout.go
+++ b/internal/job/checkout.go
@@ -165,10 +165,17 @@ func (e *Executor) CheckoutPhase(ctx context.Context) error {
 				return nil
 			}
 
+			var errLockTimeout ErrTimedOutAcquiringLock
 			switch {
 			case shell.IsExitError(err) && shell.GetExitCode(err) == -1:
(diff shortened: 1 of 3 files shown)


wolfeidau (approved):
Great work, should we add something to the docs for agent error codes?

DrJosh9000:
> Great work, should we add something to the docs for agent error codes?

Yes we should!

PR #3032, opened 2024-10-09
install.sh tidyups

DrJosh9000:
### Description

install.sh is in need of some love. I borrowed some ideas from https://getmic.ro

### Context

https://coda.io/d/Escalations-Feedback_dHnUHNps1YO/Pipelines-In-Progress-Board_suIFKYEb#Escalations-Pipelines_tuQbCBf6/r732&view=modal

### Changes

* More quotes and handlebars
* Refactor picking the HTTP client program (wget/curl)
* Use `mktemp -d`, and a cleanup trap
* Heredoc
* Use the new-ish SHA256SUMS file to checksum the file, with an escape hatch

### Testing
- [x] Works on macOS
- [x] Works on Ubuntu

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,12 +1,12 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # This is the installer for the Buildkite Agent.
 #
 # For more information, see: https://github.com/buildkite/agent
 
-set -e
+set -eu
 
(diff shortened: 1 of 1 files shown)


swaller-bk (approved):
Tested on darwin amd64, I was able to use the releases dir to trigger the checksum check, nice work!

PR #3043, opened 2024-10-17
Implement several documentation improvements to the Agent (for the Buildkite Docs).

gilesgas:
### Description

This PR addresses several docs improvements:

- Adds more info regarding 'environment propagation' for JWKS files.
- Improve the description for the 'redactor' command to make it behave like the others for the docs.

### Context

The purpose of this PR is to address some long-standing issues with the Buildkite Docs that can only be addressed through the Agent codebase.

### Testing


clicommand/agent_start.go
```diff
@@ -658,7 +658,7 @@ var AgentStartCommand = cli.Command{
 		},
 		cli.StringFlag{
 			Name:   "signing-jwks-file",
-			Usage:  "Path to a file containing a signing key. Passing this flag enables pipeline signing for all pipelines uploaded by this agent. For hmac-sha256, the raw file content is used as the shared key",
+			Usage:  "Path to a file containing a signing key. Passing this flag enables pipeline signing for all pipelines uploaded by this agent. For hmac-sha256, the raw file content is used as the shared key. If using Docker containers to upload pipeline steps dynamically, turn on environment propagation to allow signing for the new steps generated.",
```
gilesgas, line 661: Hi @PriyaSudip , I had a couple of simple queries about this addition of yours:

- Whereabouts is 'environment propagation' turn on? Is this somewhere in the Buildkite interface. If not, whereabouts exactly is this done?
- Would it be more correct to refer to 'environment propagation' as 'environment variable propagation' (and whereabouts to is the propagation happening)? Otherwise, if it's not environment variables which are being propagated, then what aspect of 'environment' is being propagated?
- Last, what exactly is being signed (a pipeline or a step of a pipeline)?

clicommand/redactor_add.go
```diff
@@ -52,9 +52,26 @@ type RedactorAddConfig struct {
...
+	it is recommended that you use this command to ensure they will be
+	redacted from subsequent logs. Secrets fetched with the builtin
+	′secret get′ command do not require the use of this command, they will
+	be redacted automatically.
+
+  Example:
+
+    $ buildkite-agent redactor add ...
+
+	Redacts something.`,
```
gilesgas, line 74: Hello @wolfeidau ,

Could you (or would you know anyone who could) please provide an example of how this command is used? Just a typical example using some of the options would be great, and then I can provide a better explanation than what's currently provided in the agent.

clicommand/agent_start.go
DrJosh9000: I can take a stab at answering these:

* "Environment variable propagation" (which I agree is better than "environment propagation") refers to the `docker` CLI `-e` flag (or equivalent configuration in `docker compose` or the docker/docker compose plugins, or whatever the user is using), i.e. the user must ensure Docker/their container system is configured to set the signing-related environment variables to the same values _within_ the container that runs `buildkite-agent pipeline upload` as the agent has set _outside_ the container.
* All the command steps within the pipeline are signed when `buildkite-agent pipeline upload` is run with this configuration (https://buildkite.com/docs/agent/v3/signed-pipelines).

gilesgas: Thanks very much for clarifying this @DrJosh9000 ! I'll revise this wording accordingly.

clicommand/redactor_add.go
gilesgas: FYI, I fixed the spacing in the Description section only, which is why this comment appears outdated, but it's actually not.

gilesgas:
The redactor docs will look like this after the merge and release of this PR:

Screenshot 2024-10-29 at 2 48 10 pm

clicommand/agent_start.go
```diff
@@ -658,7 +658,7 @@ var AgentStartCommand = cli.Command{
 		},
 		cli.StringFlag{
 			Name:   "signing-jwks-file",
-			Usage:  "Path to a file containing a signing key. Passing this flag enables pipeline signing for all pipelines uploaded by this agent. For hmac-sha256, the raw file content is used as the shared key",
+			Usage:  `Path to a file containing a signing key. Passing this flag enables pipeline signing for all pipelines uploaded by this agent. For hmac-sha256, the raw file content is used as the shared key. When using Docker containers to upload pipeline steps dynamically, use environment variable propagation (for example, "docker run -e MYVAR") to allow all steps within the pipeline to be signed.`,
```
DrJosh9000, line 661: The reason env var propagation is needed is specifically so that the agent can pass the file on to the `pipeline upload` command within the container, so I think this ought to use that particular variable in the example:

```suggestion
			Usage:  `Path to a file containing a signing key. Passing this flag enables pipeline signing for all pipelines uploaded by this agent. For hmac-sha256, the raw file content is used as the shared key. When using Docker containers to upload pipeline steps dynamically, use environment variable propagation (for example, "docker run -e BUILDKITE_AGENT_JWKS_FILE") to allow all steps within the pipeline to be signed.`,
```

DrJosh9000 (approved):
LGTM with one suggestion ✅

clicommand/agent_start.go
gilesgas: Aha, thank you for that, I'll merge in this PR now.

PR #3045, opened 2024-10-21
k8s exec: Perform liveness check of clients

DrJosh9000:
### Description

Add a mechanism for terminating if a client stops communicating.

### Context

Sometimes, a container will just die (e.g. it is OOM-killed). If that happens, the agent can sit around happily until the heat death of the universe waiting for `Exit` to be called by something that isn't running any more.

### Changes

* Some code rearrangement
* Store clients in a slice instead of a map, since they're keyed by sequential int
* Finer-grained locking
  * Locking around channel close inside a `sync.Once.Do` is pointless, given nothing else writes to the channels. The existing locking only prevented data races on the clients, and the clients map/slice never changes, so the mutexes can be made client-specific.
* Add a goroutine that inspects client states and terminates the job if a client that should be connected goes silent for too long.

### Testing

### DIFF
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -362,7 +362,7 @@ func NewJobRunner(ctx context.Context, l logger.Logger, apiClient APIClient, con
 		if err != nil {
 			return nil, fmt.Errorf("failed to parse BUILDKITE_CONTAINER_COUNT: %w", err)
 		}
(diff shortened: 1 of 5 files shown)


wolfeidau (approved):
seem solid, supervisor is 👍🏻 🔍

PR #3051, opened 2024-10-24
Allow token to be empty if graphql-token is provided

jordandcarter:
### Description
Fixes `tool sign` when using `--graphql-token`.

### Context
Running `tool sign --graphql-endpoint http://graphql.buildkite.localhost/v1 --graphql-token this_is_not_the_actual_token` kept returning an error `fatal: Error signing pipeline: couldn't retrieve pipeline: Post "http://graphql.buildkite.localhost/v1": Invalid token, empty string supplied`

### Testing


DrJosh9000 (approved):
Whoops. LGTM

PR #3054, opened 2024-10-28
Add tini-static to alpine images

DrJosh9000:
### Description

Install both `tini` and `tini-static` in `alpine` and `alpine-k8s`.

### Context

This is the easiest way to make `tini-static` available to `agent-stack-k8s`, which will be needed for defunct-process-reaping within arbitrary container images. `agent-stack-k8s` by default uses the agent `alpine` image, not `alpine-k8s`, but I think they should be somewhat consistent.

### Testing


wolfeidau (approved):
👍🏻

PR #3060, opened 2024-10-30
Minor update to the 'redactor' CLI command examples.

gilesgas:
In the 'redactor' CLI command examples, this PR removes the equal signs from the `--format` options, to match options table in docs, where the option is specified as `--format value`.

It's better to keep examples and options tables consistent, to avoid customer/reader confusion.

Nevertheless, using an equal sign in this context, or not, should make no functional difference to the way this option is used.

### DIFF
--- a/clicommand/redactor_add.go
+++ b/clicommand/redactor_add.go
@@ -88,11 +88,8 @@ Or
     $ buildkite-agent redactor add --format json my-secrets.json`,
 	Flags: []cli.Flag{
 		cli.StringFlag{
-			Name: "format",
-			Usage: fmt.Sprintf(
-				"The format for the input, whose value is either ′json′ or ′none′. ′none′ adds the entire input's content to the redactor, with the exception of leading and trailing space. ′json′ parses the input's content as a JSON object, where each value of each key is added to the redactor.",
-				secretsFormats,
-			),
+			Name:   "format",
+			Usage:  "The format for the input, whose value is either ′json′ or ′none′. ′none′ adds the entire input's content to the redactor, with the exception of leading and trailing space. ′json′ parses the input's content as a JSON object, where each value of each key is added to the redactor.",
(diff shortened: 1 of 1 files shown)


DrJosh9000 (approved):
LGTM ✅

gilesgas:
Thanks @DrJosh9000 !

Just want to add a couple more tweaks to at least one of the related option descriptions, now that I've read them in a tad more detail ;-)

clicommand/redactor_add.go
```diff
@@ -81,16 +81,16 @@ To redact multiple secrets from future logs in one command, create a flat
...
 
-    $ buildkite-agent redactor add --format=json my-secrets.json`,
+    $ buildkite-agent redactor add --format json my-secrets.json`,
 	Flags: []cli.Flag{
 		cli.StringFlag{
 			Name: "format",
 			Usage: fmt.Sprintf(
-				"The format for the input, one of: %s. ′none′ will add the entire input as a to the redactor, save for leading and trailing whitespace, ′json′ will parse it a string valued JSON Object, where each value of each key will be added to the redactor.",
+				"The format for the input, whose value is either ′json′ or ′none′. ′none′ adds the entire input's content to the redactor, with the exception of leading and trailing space. ′json′ parses the input's content as a JSON object, where each value of each key is added to the redactor.",
 				secretsFormats,
```
DrJosh9000, line 94: If we're moving away from `%s` to embed some values into the string, to it merely being a fixed string, then we should get rid of the `fmt.Sprintf` call:

```suggestion
			Usage: "The format for the input, whose value is either ′json′ or ′none′. ′none′ adds the entire input's content to the redactor, with the exception of leading and trailing space. ′json′ parses the input's content as a JSON object, where each value of each key is added to the redactor.",
```

gilesgas: Ah! TIL!

PR #3062, opened 2024-10-30
Use sha256 in the checksum verification

esenmarti:
### Description

Older version of openssl (e.g. 1.0.2) don't use sha256 as default
option for the openssl dgst so make it explicit.

### Testing


DrJosh9000 (approved):
Thanks! LGTM

PR #3064, opened 2024-10-31
Remove .editorconfig

DrJosh9000:
### Description

I think that Go code should _not_ use 2 space tabs, especially if it's just to be consistent with Ruby backend code. Personally I am used to 4 or 8 space tabs. But I also think that `.editorconfig` is a matter of personal preference: if you want to use 2 space tabs you go ahead and configure your editor that way. We shouldn't be supplying an `.editorconfig` for you.

### Context

The editor I've been using started respecting this file ahead of my other config, so I'm inclined to remove the file.

### Testing

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,2 +1,0 @@
-[*.go]
-indent_size=2
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,6 @@ packaging/docker/*/hooks/
 
 .idea
 .vscode
+
+# Editor config boils down to personal preference
+.editorconfig


pda (approved):
Makes sense to me.

Presumably we're using actual `0x09` tab characters, making this is purely a display option, which doesn't really help drive consistency. Arguably it might help align ASCII diagrams etc maybe? 🤷🏼‍♂️

And there's no other config in this file… 🚮

> `indent_size`: a whole number defining the number of columns used for each indentation level and the width of soft tabs (when supported). When set to tab, the value of `tab_width` (if specified) will be used.

— https://editorconfig.org/#supported-properties

The documentation at https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#indent_size is super confusing:

> If indent_size equals to tab, the indent_size will be set to the tab size, which should be tab_width if tab_width is specified, or the tab size set by editor if tab_width is not specified.

DrJosh9000:
> Presumably we're using actual `0x09` tab characters, making this is purely a display option, which doesn't really help drive consistency. Arguably it might help align ASCII diagrams etc maybe? 🤷🏼‍♂️

Indeed, thanks to `gofmt`. I feel that using "hard" tabs within ASCII art is asking for trouble. `gofmt` will also sometimes mess with comments (usually with positive results).

wolfeidau:
@DrJosh9000 @pda this is interesting, I normally add an editor configuration to repos to help contributors with preferences for non Go code, YAML, JSON and markdown.

TBH I have never seen it just target Go.

This is what I would consider normal https://github.com/editorconfig/editorconfig-core-go/blob/master/.editorconfig mostly to assist contributors in not spending ages aligning their code with the repos standards.

PR #3070, opened 2024-11-08
Create `buildkite-agent step cancel` subcommand

mitchbne:
### Description

Creates a new subcommand `buildkite-agent step cancel --step "key"`. This subcommand can be used to cancel all unfinished jobs for a command step. We optionally allow a `--force` flag which will mean that we don't wait for an agent running a job to finish it before cancelling the job in Buildkite.

### Testing


wolfeidau (approved):
👍🏻 🚢

PR #3074, opened 2024-11-08
Fix nil pointer panic in k8s mode

DrJosh9000:
### Description

`kubernetesSetup` used `e.shell` to log before `e.shell` was created. Oops.

### Context

https://coda.io/d/Escalations-Feedback_dHnUHNps1YO/Pipelines-Escalations_suQ4B7FT#Pipelines-Escalations-Board_tu66S__K/r822&view=center

### Changes

* Don't log anything in `kubernetesSetup`
* Move the temporary logger up from `setupRedactors` and use that.

### Testing


catkins (approved):
Thanks @DrJosh9000

PR #3077, opened 2024-11-10
dependabot: Group Dockerfiles

DrJosh9000:
### Description

Docker rebuilds take a while, and Dependabot has been failing to pull images, so try to group updates.

### Context

Recent build times.

### Testing


wolfeidau (approved):
👍🏻 🤖

PR #3078, opened 2024-11-11
Drop Ubuntu 18.04, add Ubuntu 24.04

DrJosh9000:
### Description

It's about time.

### Context

None

### Testing


wolfeidau (approved):
👍🏻 🤖 

We should do a change log for this just so customers are aware?

PR #3082, opened 2024-11-11
Handle older version of remote ref error message

steveh:
### Description

We have code to handle the case where we can't fetch a short remote ref and fall back to fetching all tags. However the case of this error changed in git v2.21.0 (2019) and we only caught the new lowercase version. Handle the old title case version to support customers running old versions of git.

### Context

Linear PS-179

### Testing


SorchaAbel (approved):
Thank you Steve!

DrJosh9000 (approved):
LGTM ✅

PR #3083, opened 2024-11-11
Wrap S3 client test request in retry

DrJosh9000:
### Description
Test the hypothesis that the S3 client test request can fail with `ErrNoValidProvidersFoundInChain` because of transient network issues, by wrapping in a retry.

Also: update a deprecated call.

### Testing


wolfeidau:
@DrJosh9000 doesn't the AWS client have retries already? https://docs.aws.amazon.com/sdk-for-go/api/aws/client/#DefaultRetryer 

Why add one on top?

DrJosh9000:
I agree @wolfeidau, this was just the first thought I had from the meeting this came up in.

PR #3087, opened 2024-11-12
Rename env var to `BUILDKITE_STEP_CANCEL_FORCE_GRACE_PERIOD_SECONDS`

mitchbne:
### Description

Meant to include  https://github.com/buildkite/agent/commit/a3b8339605d826de15f8ca5ce85e148c6c313b19 in https://github.com/buildkite/agent/pull/3084 but messed up the rebase. `BUILDKITE_STEP_CANCEL_FORCE_GRACE_PERIOD_SECONDS ` feels like a better name for this env variable.

### Testing

### DIFF
--- a/clicommand/step_cancel.go
+++ b/clicommand/step_cancel.go
@@ -74,7 +74,7 @@ var StepCancelCommand = cli.Command{
 			Name:   "force-grace-period-seconds",
 			Value:  defaultCancelGracePeriod,
 			Usage:  "The number of seconds to wait for agents to finish uploading artifacts before transitioning unfinished jobs to a canceled state. ′--force′ must also be supplied for this to take affect",
-			EnvVar: "BUILDKITE_FORCE_GRACE_PERIOD_SECONDS,BUILDKITE_CANCEL_GRACE_PERIOD",
+			EnvVar: "BUILDKITE_STEP_CANCEL_FORCE_GRACE_PERIOD_SECONDS,BUILDKITE_CANCEL_GRACE_PERIOD",
 		},
 
 		// API Flags


clicommand/step_cancel.go
```diff
@@ -74,7 +74,7 @@ var StepCancelCommand = cli.Command{
 			Name:   "force-grace-period-seconds",
 			Value:  defaultCancelGracePeriod,
 			Usage:  "The number of seconds to wait for agents to finish uploading artifacts before transitioning unfinished jobs to a canceled state. ′--force′ must also be supplied for this to take affect",
-			EnvVar: "BUILDKITE_FORCE_GRACE_PERIOD_SECONDS,BUILDKITE_CANCEL_GRACE_PERIOD",
+			EnvVar: "BUILDKITE_FORCE_CANCEL_GRACE_PERIOD_SECONDS,BUILDKITE_CANCEL_GRACE_PERIOD",
```
mhaylock, line 77: @DrJosh9000, @wolfeidau is it ok that this environment variable doesn't 1:1 match the flag? It feels a bit redundant to include the work "cancel" in the flag name - but fairly important to have it in the global ENV var name.

mhaylock (approved):
I think this is a good improvement (I did suggest it 😅).

My only hesitation is around whether its aligned with the rest of the agent that the ENV var name doesn't match the flag name 1:1.

clicommand/step_cancel.go
wolfeidau: We have some similar differences in other flags such as https://github.com/buildkite/agent/blob/main/clicommand/tool_sign.go#L123

Env vars are a global space so prefixing seems legit to me.

mhaylock: I guess a legitimate prefix would technically be `BUILDKITE_STEP_CANCEL_FORCE_GRACE_PERIOD_SECONDS`.

DrJosh9000: I think it's good - it's clear that this env var goes with this flag, and the flag belongs in the context of the command but the env var stands alone.

clicommand/step_cancel.go
```diff
@@ -74,7 +74,7 @@ var StepCancelCommand = cli.Command{
 			Name:   "force-grace-period-seconds",
 			Value:  defaultCancelGracePeriod,
 			Usage:  "The number of seconds to wait for agents to finish uploading artifacts before transitioning unfinished jobs to a canceled state. ′--force′ must also be supplied for this to take affect",
-			EnvVar: "BUILDKITE_FORCE_GRACE_PERIOD_SECONDS,BUILDKITE_CANCEL_GRACE_PERIOD",
+			EnvVar: "BUILDKITE_STEP_CANCEL_FORCE_GRACE_PERIOD_SECONDS,BUILDKITE_CANCEL_GRACE_PERIOD",
```
DrJosh9000, line 77: Even better 👍

wolfeidau (approved):
👍🏻

DrJosh9000 (approved):

PR #3091, opened 2024-11-14
Switch to agent-base images

DrJosh9000:
### Description

Speed up agent builds. A large amount of the pipeline is spent in `apt-get install ...` for the Ubuntu images. I've split that part out into https://github.com/buildkite/agent-base-images which push to Public ECR - so we can do a scheduled build of that, and use Dependabot to update the pins here.

### Changes

* Point to public.ecr.aws/buildkite/agent-base
* Remove the parts that agent-base-images now builds into the new base images.

### Testing


moskyb (approved):
😍

PR #3095, opened 2024-11-18
Don't surface expected stderr output from git rev-parse

CerealBoy:
### Description

When using git-mirrors, a rev-parse call is made to look for the commit within the current mirror, which will fatal if not found. As this is expected behaviour, capturing and displaying this error is misleading and not helpful. We can instead ignore it, as the mirror will be updated in this circumstance.

### Context

https://coda.io/d/Escalations-Feedback_dHnUHNps1YO/Pipelines-In-Progress-Board_suIFKYEb#Escalations-Pipelines_tuQbCBf6/r850&view=modal

### Changes

When doing the rev-parse check for the commit, include a `ShowStderr(false)` to the shell call.

### Testing


DrJosh9000 (approved):
LGTM ✅ 

Sorry to cause the confusion 🙇

PR #3106, opened 2024-11-20
OLDMK-2904 uodaet annotation

PriyaSudip:
Discussed with @gilesgas just to enable emoji name as an example

### DIFF
--- a/clicommand/annotate.go
+++ b/clicommand/annotate.go
@@ -48,7 +48,7 @@ body entirely and providing a new style value.
 
 Example:
 
-    $ buildkite-agent annotate "All tests passed! :rocket:"
+    $ buildkite-agent annotate "All tests passed! :your-emoji: like :rocket:"
     $ cat annotation.md | buildkite-agent annotate --style "warning"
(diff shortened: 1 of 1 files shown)


gilesgas:
Thanks @PriyaSudip ! 🎉

Just for a little more context, our Buildkite Docs site automatically renders emojis like `:wave:` into their emoji output like 👋 , unless the emoji syntax has been explicitly escaped like `\:wave\:`.

Therefore, when https://github.com/buildkite/docs/blob/main/scripts/update-agent-help.sh runs https://github.com/buildkite/docs/blob/main/scripts/cli2md.rb (which are the scripts/programs that generate some of the Buildkite Agent CLI content for the Buildkite Docs from the `buildkite-agent $command -help` output, the `cli2md.rb` program doesn't automatically escape emoji syntax, and so the command line examples end up being a little confusing.

Therefore, this change is meant to make these docs a little clearer, since we don't necessarily want to escape all emoji syntax in these scripts/programs.

gilesgas (approved):
Ready for you to merge this one in whenever you're ready @PriyaSudip

PR #3108, opened 2024-11-24
Fix duplicated output when debug is enabled

### LINKED ISSUE #3107: Command output is printed twice in debug mode
Hi,

Somewhere between 3.65.0 and 3.87.0, a change caused logs to be printed twice when running in debug mode (`BUILDKITE_AGENT_DEBUG=true`).

3.65.0:
image

3.87.0 with `BUILDKITE_AGENT_DEBUG=true`:
image

3.87.0 with `BUILDKITE_AGENT_DEBUG=false`:
image

Notably, this doesn't affect log output coming from the agent itself:
image

We're running the agent from the `buildkite/agent:3.87.0-ubuntu-22.04` Docker image.

DrJosh9000:
### Description

Fix duplicate command output when debug mode is enabled.

Fixes #3107

### Context

https://linear.app/buildkite/issue/PS-200/duplicate-logs-in-agent-with-debug-mode-enabled

The bug was added in #3068 - at the time, I didn't think through all the consequences of changing to "tee-ing" command output.

### Changes

* Instead of tee-ing output to the log streamer when debug is enabled, substitute stdout or stderr (or both) if set to `io.Discard`.

### Testing

### DIFF
--- a/internal/shell/shell.go
+++ b/internal/shell/shell.go
@@ -566,14 +566,18 @@ func (s *Shell) executeCommand(ctx context.Context, cmdCfg process.Config, stdou
 	}
 
 	if s.debug {
-		// Tee output streams to debug logger.
-		stdOutStreamer := NewLoggerStreamer(s.Logger)
-		defer stdOutStreamer.Close()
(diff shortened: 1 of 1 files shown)


wolfeidau (approved):
👍🏻

PR #3117, opened 2024-12-03
fix: allow for empty files on hook check

nzspambot:
### Description

Empty hook will fail on agent on run

### Context

We produce an empty checkout hook file :

```
checkout_override_file=$(mktemp "${rw_root_dir}/tmp/checkout-override.XXXXXXXX")
```

```
    # We need to have a custom override for the checkout hook. This copies an
    # empty checkout file so buildkite-agents can still run the pre-checkout and post-checkout
    # hooks. This is very much necessary as a several plugins seem to rely on these
    # hooks to run their own logic like secrets-manager
    # We're also copying the git repo structure as well. This preserves
    # the current file structure, mainly due to developers using
    # this exact path for their pipeline scripts. Without this file path structure
    # pipelines end up breaking.
```

This currently doesn't work as the file is empty as the hook check will fail (this is your extracted function):

```
bash-5.2$ go run /tmp/main.go -file /tmp/checkout-override.kFBBE4Ia 
Error: reading first four bytes of file "/tmp/checkout-override.kFBBE4Ia": EOF
```

image

### Changes

Adds a check to see if the file has 4 bytes if not then return nil (this is your updated function):

```
bash-5.2$ go run /tmp/main.go -file /tmp/checkout-override.kFBBE4Ia 
The file /tmp/checkout-override.kFBBE4Ia is not a binary executable.
```

### Testing
I would like to test this but see no tests around this? unless I am missing them, also it builds fine locally


internal/job/hook/binary.go
```diff
@@ -60,6 +60,16 @@ func isBinaryExecutable(path string) (bool, error) {
...
 
+	fileInfo, err := f.Stat()
+	if err != nil {
+		return false, fmt.Errorf("stat file %q: %w", path, err)
+	}
+
+	if fileInfo.Size() < 4 {
+		// there are less than four bytes in the file, there's nothing that we can do with it
+		return false, nil
+	}
```
DrJosh9000, line 71: I think we should move this in between line 56 `defer f.Close()` and line 57 `r := bufio.NewReader(f)`, so that it checks the size before trying to `r.Peek(4)`.

DrJosh9000 (commented):
Thanks @nzspambot! Looking good, I have one comment. 

Re the lack of tests, I agree it would be good to have a couple of fixtures to test this function. I might add something. I don't see it as a blocker to this PR though.

internal/job/hook/binary.go
nzspambot: 👍🏻 I will fix that right now

nzspambot: all done 🥇

DrJosh9000 (approved):
LGTM!

PR #3122, opened 2024-12-10
Ignore empty submodule clone configs

DrJosh9000:
### Description

A small bit of hardening - if `BUILDKITE_GIT_SUBMODULE_CLONE_CONFIG` is set but empty, then `GitSubmoduleCloneConfig` can be a slice with one element `""`, causing an invalid Git flag to be appended.

### Context

https://coda.io/d/_dHnUHNps1YO#Escalations-by-Assignee_tugT4T3x/r883&view=center

### Testing


rianmcguire (approved):
Good idea

PR #3123, opened 2024-12-11
Notify when host and bootstrap agent paths mismatch

jordandcarter:
### Description
During agent startup check if the bootstrap agent path matches the startup agent path. This is to avoid confusion during development when changes are made to the agent but the bootstrap agent still uses the default agent on the `$PATH`. Usually this results in changes being made to the agent, but then those changes not being observed when running the updated agent locally.

For example if I 

1. `go build -o buildkite-agent .` then
2. `./buildkite-agent start --spawn 2 --endpoint http://agent.buildkite.localhost/v3 --token buildkite`

By default the above will result in `./buildkite-agent` starting up and waiting for jobs, when a job is run, if that job uploads a pipeline it will use (in my case) `/opt/homebrew/bin/buildkite-agent`, not `./buildkite-agent`, so any changes I've made to `./buildkite-agent` will not be observed for the agent pipeline upload command.

This PR updates the agent to return a error `fatal: check binary paths: mismatched buildkite-agent paths: host="/Users/jordan/src/agent/buildkite-agent" bootstrap="/opt/homebrew/bin/buildkite-agent"` on boot. If the host agent is a development build then the error is fatal, otherwise it is just a warning.

### Testing

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"os/exec"
 	"os/signal"
 	"path/filepath"
 	"regexp"
@@ -758,6 +759,18 @@ var AgentStartCommand = cli.Command{
 		))
 		defer done()
 
+		// Verify that the bootstrap buildkite-agent executable matches the current host agent
+		// executable. In development builds, it exits on mismatch; otherwise it only logs a warning.
+		// This is to avoid confusion when making changes to the buildkite-agent but forgetting to
+		// update the buildkite-agent binary available from $PATH
+		if err := checkBinaryPaths(); err != nil {
+			if version.IsDevelopmentBuild() {
+				return fmt.Errorf("check binary paths: %s", err)
+			} else {
+				l.Warn("check binary paths: %s", err)
+			}
+		}
+
 		// Remove any config env from the environment to prevent them propagating to bootstrap
 		if err := UnsetConfigFromEnvironment(c); err != nil {
 			return fmt.Errorf("failed to unset config from environment: %w", err)
@@ -1255,6 +1268,28 @@ var AgentStartCommand = cli.Command{
 
 		return err
 	},
+}
+
+// checkBinaryPaths looks up both the bootstrap and host buildkite-agent paths,
+// and returns an error if they do not match or if either cannot be determined.
+func checkBinaryPaths() error {
+	bootstrapPath, err := exec.LookPath("buildkite-agent")
+	if err != nil {
+		return fmt.Errorf("failed to locate bootstrap buildkite-agent: %w", err)
+	}
+
+	hostPath, err := os.Executable()
+	if err != nil {
+		return fmt.Errorf("failed to determine host buildkite-agent executable: %w", err)
+	}
+
+	if hostPath != bootstrapPath {
+		return fmt.Errorf(
+			"mismatched buildkite-agent paths: host=%q bootstrap=%q",
+			hostPath, bootstrapPath,
+		)
+	}
+	return nil
 }
 
 func parseAndValidateJWKS(ctx context.Context, keysetType, path string) (jwk.Set, error) {
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -1278,17 +1278,28 @@ func checkBinaryPaths() error {
 		return fmt.Errorf("failed to locate bootstrap buildkite-agent: %w", err)
 	}
 
+	evalBootstrapPath, err := filepath.EvalSymlinks(bootstrapPath)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


clicommand/agent_start.go
```diff
@@ -758,6 +759,18 @@ var AgentStartCommand = cli.Command{
 		))
 		defer done()
 
+		// Verify that the bootstrap buildkite-agent executable matches the current host agent
+		// executable. In development builds, it exits on mismatch; otherwise it only logs a warning.
+		// This is to avoid confusion when making changes to the buildkite-agent but forgetting to
+		// update the buildkite-agent binary available from $PATH
+		if err := checkBinaryPaths(); err != nil {
+			if version.IsDevelopmentBuild() {
+				return fmt.Errorf("check binary paths: %s", err)
```
patrobinson, line 768: There might be situations where we want to allow this. I think it's good to err on the side of blowing up, rather than behaving unexpectedly, but I wonder if we should have a parameter to allow it?

clicommand/agent_start.go
```diff
@@ -1257,6 +1270,28 @@ var AgentStartCommand = cli.Command{
...
 
+// checkBinaryPaths looks up both the bootstrap and host buildkite-agent paths,
+// and returns an error if they do not match or if either cannot be determined.
+func checkBinaryPaths() error {
+	bootstrapPath, err := exec.LookPath("buildkite-agent")
+	if err != nil {
+		return fmt.Errorf("failed to locate bootstrap buildkite-agent: %w", err)
+	}
+
+	hostPath, err := os.Executable()
```
patrobinson, line 1281: The documentation points out this may return a symlink, which could cause inconsistencies. It recommends evaluating the symlink:

> If a symlink was used to start the process, depending on the operating system, the result might be the symlink or the path it pointed to. If a stable result is needed, [path/filepath.EvalSymlinks](https://pkg.go.dev/path/filepath#EvalSymlinks) might help.

clicommand/agent_start.go
```diff
@@ -758,6 +759,18 @@ var AgentStartCommand = cli.Command{
...
 
+		// Verify that the bootstrap buildkite-agent executable matches the current host agent
+		// executable. In development builds, it exits on mismatch; otherwise it only logs a warning.
+		// This is to avoid confusion when making changes to the buildkite-agent but forgetting to
+		// update the buildkite-agent binary available from $PATH
+		if err := checkBinaryPaths(); err != nil {
+			if version.IsDevelopmentBuild() {
+				return fmt.Errorf("check binary paths: %s", err)
+			} else {
+				l.Warn("check binary paths: %s", err)
```
patrobinson, line 770: This could create a lot of logs for customers. I wonder if it should be debug level?

patrobinson (approved):
I'm not sure if we should generate a warning for production environments, although maybe it'll be useful?

I'm pretty happy with this as an approach, and will save us so much time in the future.

clicommand/agent_start.go
```diff
@@ -1257,6 +1270,28 @@ var AgentStartCommand = cli.Command{
 	},
 }
 
+// checkBinaryPaths looks up both the bootstrap and host buildkite-agent paths,
+// and returns an error if they do not match or if either cannot be determined.
+func checkBinaryPaths() error {
+	bootstrapPath, err := exec.LookPath("buildkite-agent")
```
patrobinson, line 1276: We may also want to resolve this to ensure it's not a symlink, to ensure we're comparing like for like

clicommand/agent_start.go
moskyb: it should only be once at the start of the agent logs (not the job logs), so i think that this is probably fine.

clicommand/agent_start.go
moskyb: given that it's only used for development builds, i think if we need to turn it off for dev builds we can just comment out the code?

moskyb (commented):
looks awesome, such a good change.

moskyb (approved):
whoops! wrong review

jordandcarter:
> I'm not sure if we should generate a warning for production environments, although maybe it'll be useful?
Yeah I wasn't sure either but was thinking if it ever did log the warning in production that would be super informative to the user, then I thought of the following scenario too.

A production environment in this context includes a user downloading a new agent binary directly to test out a new feature locally. They forgot they installed the agent via brew 12 months ago and during their testing they'll run into the same confusion as us. In this case it will only raise the warning, which is the least disruptive approach while still giving them a clue to whats wrong.

DrJosh9000:
I think the overall approach is good. 

Though at some point I want to get #1816 on the frontburner again...

clicommand/agent_start.go
jordandcarter: Thanks for looking that up, good catch! I've updated it to handle symlinks.

patrobinson (approved):

DrJosh9000 (approved):
LGTM!

PR #3124, opened 2024-12-12
Adding support for Additional Hooks Paths

CerealBoy:
### Description

We would like to provide the ability to place hooks in multiple locations, meaning there can be multiple hooks of the same type, and they can be provided under different circumstances. These additional hooks are comma separated list, and ordered as they are provided in terms of how multiple hooks are executed. Those in `hooks-path` are always first, and then sequentially through `additional-hooks-paths`.

There are 2 places where we pull in these hooks:
1. During the agent startup phase
2. Through the job execution phases

We don't currently support the `pre-bootstrap` hook with additional paths.

#### Note

There is no change to any default behaviour included, nor is there any change in existing functionality when configuration is not changed. The only way to have any impact from this change is to set the new option with some value (which should be a comma separated list of paths).

A test is included here to validate the ordering of the hooks is predictable and preserved, however there is a weird timing issue during the test that prevents the additional hooks from being executed successfully. All manual usage of the changes has yielded expected results, so the test is left with `t.SkipNow()` to highlight that it _should_ be possible to test this, and perhaps there's something I've overlooked here.

#### Screenshots

**Before any additional hook paths are defined:**
Screenshot 2024-12-30 at 12 32 11

**Example trace showing additional hooks being run for a job:**
image

**Build output of the above within Buildkite**
image

### Changes

- Adding `BUILDKITE_ADDITIONAL_HOOKS_PATHS` configuration option
- Find and run any hooks found _after_ `hooks-path`
  - For the `agent-startup` hook
  - For hooks within job execution

### Testing

- [x] Additional tests covering new functionality (although `t.SkipNow()` is used)

### DIFF
--- a/agent/agent_configuration.go
+++ b/agent/agent_configuration.go
@@ -12,6 +12,7 @@ type AgentConfiguration struct {
 	BootstrapScript             string
 	BuildPath                   string
 	HooksPath                   string
+	AdditionalHooksPaths        []string
 	SocketsPath                 string
 	GitMirrorsPath              string
 	GitMirrorsLockTimeout       int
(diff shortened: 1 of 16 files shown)


wolfeidau (approved):
Looks great 🚀 ready for some testing.

PR #3134, opened 2024-12-17
Enable process debug logging

patrobinson:
### Description

A long long time ago, in a galaxy far far away, the process debug logging was disabled 
https://github.com/buildkite/agent/commit/146c68e652445acd75ca4026f897b3ec59f11035

We currently don't pass the agentLogger object into Shell, so I can't easily use the existing logger. So we create one in using plain text mode. Suggestions welcome on how to do that.

### Changes

Enable process logging in debug mode.

### Testing


internal/shell/shell.go
```diff
@@ -578,6 +581,9 @@ func (s *Shell) executeCommand(ctx context.Context, cmdCfg process.Config, stdou
 			defer stdErrStreamer.Close()
 			cmdCfg.Stderr = stdErrStreamer
 		}
+
+		// This should respect the log format we set for the agent
+		processLogger = logger.NewConsoleLogger(logger.NewTextPrinter(os.Stderr), os.Exit)
```
DrJosh9000, line 586: Actually there is a `Logger` embedded in `Shell`, so you can do this:

```suggestion
		processLogger = s.Logger
```

DrJosh9000 (commented):
One change, otherwise LGTM

internal/shell/shell.go
patrobinson: This doesn't work because `s.Logger` is a `shell.Logger` interface not a `logger.Logger` interface, which are completely different. Why do we have multiple logger interfaces?

DrJosh9000: Oh what? argh

DrJosh9000: At least don't write to `os.Stderr` - that would bypass the secret redactor. Use `cmdCfg.Stderr` I guess?

DrJosh9000 (approved):

PR #3135, opened 2024-12-17
Prefix fatal error message with 'buildkite-agent:'

jordandcarter:
### Description
When the agent exits and prints an error message prefix the message with `buildkite-agent: ` to provide clarity as to where the error came from. Our buildkite-agent cli is often used in scripts and when it exits with an error it isn't always obvious to the user what printed the error. It is common practice for cli tools to prefix error messages with the tool name.

### Context
https://linear.app/buildkite/issue/PS-217/prefix-fatal-error-messages-with-buildkite-agent

### Testing


DrJosh9000 (approved):
LGTM nice 👍

PR #3140, opened 2024-12-19
Update golang.org/x/net

yob:
v0.32.0 was vulnerable to CVE-2024-45338, although we weren't using it in a way that was a problem. Still worth bumping the version to avoid v0.32.0 being flagged in this repo or our binaries.


wolfeidau (approved):
👍🏻 🔒 ⭐

PR #3152, opened 2025-01-08
Skip pushing the git commit metadata if BUILDKITE_COMMIT_RESOLVED=true

CerealBoy:
### Description

During the checkout phase the agent will validate the `BUILDKITE_COMMIT` environment variable has been set knowingly within the build already, or push the data within this job. The first job in a build is likely to be the only time this needs to actually happen, as once it is and `BUILDKITE_COMMIT` has been set it will be available to remaining jobs within the build. We use the meta-data API to track this, which means we're making many API calls back to `buildkite.com` when really, we shouldn't need to do this by checking the value ourselves.

**Disclaimer:** This is quite naive in how the value of `BUILDKITE_COMMIT` is trusted, where we assume that if the value is 40 characters long and the characters match the accepted values of a SHA then we use it.

The value for `BUILDKITE_COMMIT` should only be **HEAD** or a git SHA, so we shouldn't have to worry too much. If there's a branch that was set as the value instead, it should fail the validation and then be set anyway (as in, the previous flow is preserved). In that event anyway, or in any where the validation fails, subsequent jobs will go via the happy path as the existing flow sets `BUILDKITE_COMMIT` to the determined value. I don't think there will be a problem by having this change made to shortcut the process.

### Screenshots

I did some additional testing to see if I could get some better data to back-up the intended claims from what this PR will accomplish. The idea is that only the first job in a build would need to go down the path of checking and setting the build metadata, if the `BUILDKITE_COMMIT` value is not _acceptable_, and in doing so sets the value so all remaining jobs in the build will skip that work. This tangibly means that we should see a performance improvement with job runs, as there are less calls back to Buildkite.

Screenshot 2025-01-09 at 14 40 49

This shows the log output for the first job, where the metadata pathway is getting triggered as the job sees `BUILDKITE_COMMIT=head` being the first to run.

Screenshot 2025-01-09 at 14 41 02

This shows the log output for the second job, where the `BUILDKITE_COMMIT` value matches the regex and is _assumed_ to be a valid commit. This is highlighted by the comment in the output.

Screenshot 2025-01-09 at 14 41 10

This is a screenshot of the trace for the first job, highlighting the duration of the `checkout` phase consuming most of the time (as the command is super basic), and having a length of just under `7` seconds.

Screenshot 2025-01-09 at 14 41 16

This is a screenshot of the trace for the second job, where the spans included are the same as the first except the timing is drastically reduced without the overhead of the metadata API call(s).

These traces show a _rough_ reduction of `2.5x` for the checkout alone.

### Testing

### DIFF
--- a/internal/job/checkout.go
+++ b/internal/job/checkout.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"regexp"
 	"strings"
 	"time"
 
@@ -781,22 +780,21 @@ func gitFetchCommitWithFallback(ctx context.Context, shell *shell.Shell, gitFetc
 
 const CommitMetadataKey = "buildkite:git:commit"
 
-var potentiallyValidGitHashRE = regexp.MustCompile(`^[a-f0-9]{40}|[a-f0-9]{64}$`)
-
 // sendCommitToBuildkite sends commit information (commit, author, subject, body) to Buildkite, as the BK backend doesn't
 // have access to user's VCSes. To do this, we set a special meta-data key in the build, but only if it isn't already present
 // Functionally, this means that the first job in a build (usually a pipeline upload or similar) will push the commit info
 // to buildkite, which uses this info to display commit info in the UI eg in the title for the build
 // note that we bail early if the key already exists, as we don't want to overwrite it
 func (e *Executor) sendCommitToBuildkite(ctx context.Context) error {
-	commitRef, _ := e.shell.Env.Get("BUILDKITE_COMMIT")
-	if potentiallyValidGitHashRE.MatchString(commitRef) {
+	e.shell.Commentf("Checking to see if git commit information needs to be sent to Buildkite...")
+
+	commitResolved, _ := e.shell.Env.Get("BUILDKITE_COMMIT_RESOLVED")
+	if commitResolved == "true" {
 		// we can skip the metadata shenanigans here and push straight through
-		e.shell.Commentf("Skipping the meta-data git commit steps and assuming %q is correct...", commitRef)
+		e.shell.Commentf("BUILDKITE_COMMIT is already resolved and meta-data populated, skipping...")
 		return nil
 	}
 
-	e.shell.Commentf("Checking to see if git commit information needs to be sent to Buildkite...")
 	cmd := e.shell.Command("buildkite-agent", "meta-data", "exists", CommitMetadataKey)
 	if err := cmd.Run(ctx); err == nil {
 		// Command exited 0, ie the key exists, so we don't need to send it again
--- a/internal/job/checkout.go
+++ b/internal/job/checkout.go
@@ -791,7 +791,7 @@ func (e *Executor) sendCommitToBuildkite(ctx context.Context) error {
 	commitResolved, _ := e.shell.Env.Get("BUILDKITE_COMMIT_RESOLVED")
 	if commitResolved == "true" {
 		// we can skip the metadata shenanigans here and push straight through
-		e.shell.Commentf("BUILDKITE_COMMIT is already resolved and meta-data populated, skipping...")
+		e.shell.Commentf("BUILDKITE_COMMIT is already resolved and meta-data populated, skipping")
 		return nil
 	}
 
--- a/internal/job/checkout.go
+++ b/internal/job/checkout.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"time"
 
@@ -780,12 +781,21 @@ func gitFetchCommitWithFallback(ctx context.Context, shell *shell.Shell, gitFetc
 
 const CommitMetadataKey = "buildkite:git:commit"
 
+var potentiallyValidGitSHA = regexp.MustCompile(`^[a-f0-9]{40}$`)
+
 // sendCommitToBuildkite sends commit information (commit, author, subject, body) to Buildkite, as the BK backend doesn't
 // have access to user's VCSes. To do this, we set a special meta-data key in the build, but only if it isn't already present
 // Functionally, this means that the first job in a build (usually a pipeline upload or similar) will push the commit info
 // to buildkite, which uses this info to display commit info in the UI eg in the title for the build
 // note that we bail early if the key already exists, as we don't want to overwrite it
 func (e *Executor) sendCommitToBuildkite(ctx context.Context) error {
+	commitRef, _ := e.shell.Env.Get("BUILDKITE_COMMIT")
+	if commitRef != "HEAD" && potentiallyValidGitSHA.MatchString(commitRef) {
+		// we can skip the metadata shenanigans here and push straight through
+		e.shell.Commentf("Skipping the meta-data git commit steps and assuming %q is correct...", commitRef)
+		return nil
+	}
+
 	e.shell.Commentf("Checking to see if git commit information needs to be sent to Buildkite...")
 	cmd := e.shell.Command("buildkite-agent", "meta-data", "exists", CommitMetadataKey)
 	if err := cmd.Run(ctx); err == nil {
--- a/internal/job/checkout.go
+++ b/internal/job/checkout.go
@@ -781,7 +781,7 @@ func gitFetchCommitWithFallback(ctx context.Context, shell *shell.Shell, gitFetc
 
 const CommitMetadataKey = "buildkite:git:commit"
 
-var potentiallyValidGitSHA = regexp.MustCompile(`^[a-f0-9]{40}$`)
+var potentiallyValidGitSHA = regexp.MustCompile(`^[a-f0-9]{40}|[a-f0-9]{64}$`)
 
 // sendCommitToBuildkite sends commit information (commit, author, subject, body) to Buildkite, as the BK backend doesn't
 // have access to user's VCSes. To do this, we set a special meta-data key in the build, but only if it isn't already present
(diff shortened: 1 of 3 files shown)


internal/job/checkout.go
```diff
@@ -780,12 +781,21 @@ func gitFetchCommitWithFallback(ctx context.Context, shell *shell.Shell, gitFetc
 
 const CommitMetadataKey = "buildkite:git:commit"
 
+var potentiallyValidGitSHA = regexp.MustCompile(`^[a-f0-9]{40}$`)
```
patrobinson, line 784: sha256 checksums are 64 characters long. Should we allow for those as well?

patrobinson (commented):
Can you test what happens when you trigger a build for a specific SHA?
In that case we would `BUILDKITE_COMMIT` gets populated automatically, so then the metadata won't get updated... I wonder if that has any flow on effects?

CerealBoy:
> Can you test what happens when you trigger a build for a specific SHA?

I really should've validated that when I was in the middle of doing the testing, will do so and report back any findings.

> In that case we would `BUILDKITE_COMMIT` gets populated automatically, so then the metadata won't get updated... I wonder if that has any flow on effects?

This is part of why I opened the PR, aiming to prompt conversation about this (and any other topics that I may simply not know). Will also make note to validate what happens from that perspective, if anything 🤞🏼

internal/job/checkout.go
```diff
@@ -780,12 +781,21 @@ func gitFetchCommitWithFallback(ctx context.Context, shell *shell.Shell, gitFetc
...
+
 // sendCommitToBuildkite sends commit information (commit, author, subject, body) to Buildkite, as the BK backend doesn't
 // have access to user's VCSes. To do this, we set a special meta-data key in the build, but only if it isn't already present
 // Functionally, this means that the first job in a build (usually a pipeline upload or similar) will push the commit info
 // to buildkite, which uses this info to display commit info in the UI eg in the title for the build
 // note that we bail early if the key already exists, as we don't want to overwrite it
 func (e *Executor) sendCommitToBuildkite(ctx context.Context) error {
+	commitRef, _ := e.shell.Env.Get("BUILDKITE_COMMIT")
+	if commitRef != "HEAD" && potentiallyValidGitSHA.MatchString(commitRef) {
+		// we can skip the metadata shenanigans here and push straight through
```
DrJosh9000, line 794: If `commitRef` matches the regex, then it can't possibly equal `"HEAD"`.

internal/job/checkout.go
```diff
@@ -780,12 +781,21 @@ func gitFetchCommitWithFallback(ctx context.Context, shell *shell.Shell, gitFetc
 
 const CommitMetadataKey = "buildkite:git:commit"
 
+var potentiallyValidGitSHA = regexp.MustCompile(`^[a-f0-9]{40}|[a-f0-9]{64}$`)
```
DrJosh9000, line 784: If I suggest my usual hobby-horse of naming regexps ending in `RE`, then we get:

```go
var potentiallyValidGitSHARE = regexp.MustCompile(`^[a-f0-9]{40}|[a-f0-9]{64}$`)
```

Hmm 🤔  What about:

```suggestion
var potentiallyValidGitHashRE = regexp.MustCompile(`^[a-f0-9]{40}|[a-f0-9]{64}$`)
```

DrJosh9000 (approved):
Overall LGTM, two optional nits

DrJosh9000:
Assuming the testing works out.

internal/job/checkout.go
CerealBoy: Classic "writing the if statement as my train of thought was going". I'll remove the check.

internal/job/checkout.go
CerealBoy: Makes sense to match the naming, have updated from your suggestion!

CerealBoy:
> Assuming the testing works out.

Good point @DrJosh9000, I forgot to share my results here.

Short answer is that everything works fine, the meta-data isn't _necessary_ to be set for builds / steps / jobs to succeed. I didn't try anything terribly complex though, so there's perhaps nuances I didn't find. For the most part where it's needed `BUIDLKITE_COMMIT` is used as the reference.

I did find 1 use internally, though I think with the current logic to set the meta data when `BUILDKITE_COMMIT` doesn't match will still satisfy this use. I need to raise this with the team to get further feedback on the viability of the change.

pda (commented):
Rather than regexp, could we use `git rev-parse` to see if `BUILDKITE_COMMIT` is already resolved? 🤔

Here's a PoC in shell:

```sh
$ check() { if [[ $(git rev-parse "$1") == $1 ]]; then echo "already resolved: $1"; else echo "$1 -> $(git rev-parse "$1")"; fi }

$ check c1bb97fb56f448b91fb913d9f9e4d304607191b0
already resolved: c1bb97fb56f448b91fb913d9f9e4d304607191b0

$ check HEAD
HEAD -> c1bb97fb56f448b91fb913d9f9e4d304607191b0
```

internal/job/checkout.go
```diff
@@ -780,12 +781,21 @@ func gitFetchCommitWithFallback(ctx context.Context, shell *shell.Shell, gitFetc
...
 // sendCommitToBuildkite sends commit information (commit, author, subject, body) to Buildkite, as the BK backend doesn't
 // have access to user's VCSes. To do this, we set a special meta-data key in the build, but only if it isn't already present
 // Functionally, this means that the first job in a build (usually a pipeline upload or similar) will push the commit info
 // to buildkite, which uses this info to display commit info in the UI eg in the title for the build
 // note that we bail early if the key already exists, as we don't want to overwrite it
 func (e *Executor) sendCommitToBuildkite(ctx context.Context) error {
+	commitRef, _ := e.shell.Env.Get("BUILDKITE_COMMIT")
+	if potentiallyValidGitHashRE.MatchString(commitRef) {
+		// we can skip the metadata shenanigans here and push straight through
+		e.shell.Commentf("Skipping the meta-data git commit steps and assuming %q is correct...", commitRef)
```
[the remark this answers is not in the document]
CerealBoy: Good call, I've moved the original message to the start of the function and updated the skipping message to better reflect what's happening

internal/job/checkout.go
```diff
@@ -780,12 +781,21 @@ func gitFetchCommitWithFallback(ctx context.Context, shell *shell.Shell, gitFetc
 
 const CommitMetadataKey = "buildkite:git:commit"
 
+var potentiallyValidGitHashRE = regexp.MustCompile(`^[a-f0-9]{40}|[a-f0-9]{64}$`)
```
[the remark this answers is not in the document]
CerealBoy: We've moved away from the regexp method now, though you had a good point about validating the case of the value being checked. Regardless, the 64 is indeed for the SHA256 move, which I'm sure we'll all do soon enough with each of our respective git repositories (😂).

pda (approved):

internal/job/checkout.go
```diff
@@ -787,6 +787,14 @@ const CommitMetadataKey = "buildkite:git:commit"
 // note that we bail early if the key already exists, as we don't want to overwrite it
 func (e *Executor) sendCommitToBuildkite(ctx context.Context) error {
 	e.shell.Commentf("Checking to see if git commit information needs to be sent to Buildkite...")
+
+	commitResolved, _ := e.shell.Env.Get("BUILDKITE_COMMIT_RESOLVED")
+	if commitResolved == "true" {
+		// we can skip the metadata shenanigans here and push straight through
+		e.shell.Commentf("BUILDKITE_COMMIT is already resolved and meta-data populated, skipping...")
```
sj26, line 794: This doesn't have a follow-on comment, so I'd avoid the trailing dots:

```suggestion
		e.shell.Commentf("BUILDKITE_COMMIT is already resolved and meta-data populated, skipping")
```

sj26:
Can we add a test?

https://github.com/buildkite/agent/compare/chore/skip-git-meta-data...chore/skip-git-meta-data-test

internal/job/checkout.go
CerealBoy: Good point! Will update alongside adding tests

CerealBoy:
> Can we add a test?
> 
> [chore/skip-git-meta-data...chore/skip-git-meta-data-test](https://github.com/buildkite/agent/compare/chore/skip-git-meta-data...chore/skip-git-meta-data-test)
> 
> I notice some of the other tests have been modified too in a way that looks unrelated - what's going on there?

The existing tests tracked the progress of the various `git` commands that are executed, and removing the generating of the meta-data call means there's 1 less `git` call being made. I'll improve the tests though to cater for the differing cases, with and without the ENV being set.

internal/job/integration/checkout_integration_test.go
```diff
@@ -956,6 +950,89 @@ func TestRepositorylessCheckout(t *testing.T) {
...
+	git.ExpectAll([][]any{
+		{"clone", "-v", "--", tester.Repo.Path, "."},
+		{"clean", "-ffxdq"},
+		{"fetch", "--", "origin", "main"},
+		{"checkout", "-f", "FETCH_HEAD"},
+		{"clean", "-ffxdq"},
+	})
+
+	agent := tester.MockAgent(t)
+	agent.Expect("meta-data", "exists", job.CommitMetadataKey).Exactly(0)
```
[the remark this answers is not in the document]
CerealBoy: Yes, that's why there's `.Exactly(0)` to specifically highlight it. I could remove these 2 lines completely and it would work as we expect, however I wanted to leave it here with the `0` to be pedantic about it.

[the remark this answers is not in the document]
sj26: Oh I missed the end bit! Nice.

sj26 (approved):
Nice 🙌🏻

PR #3153, opened 2025-01-09
PIPE-631 Backoff exponentially when job acquisition fails

moskyb:
### Description

During recent incidents, we've run into situations where when in acquire-job mode, the agent can tend to be a bit of a thundering herd - we retry every 3 seconds, and don't jitter. This means that if a lot of jobs are acquired all at once, and Buildkite is having a bad time, it'll tell them all to go away all at once, and then they'll all come back 3 seconds later, with the load all happening all at once.

This PR updates the retry strategy the agent uses when acquiring jobs, making the following changes:
- The basic strategy used in `roko` has been updated from 
  ```Go
  roko.WithMaxAttempts(9)
  roko.WithStrategy(roko.Constant(3*time.Second))
  ```
  to
  ```Go
  roko.WithMaxAttempts(7)
  roko.WithStrategy(roko.Exponential(2*time.Second, 0))
  ```
  
  effectively, this means that we'll make fewer requests over a longer period of time, with the delay growing exponentially with each successive attempt. Incidentally, the retrier was already documented to use this approach, but it wasn't actually using it.
- We've added jitter to the retry strategy using roko's new [`WithJitterRange` feature](https://github.com/buildkite/roko/pull/17) - we jitter on a range of `[-1s, 5s]`. This means that the wait between each outgoing request will be adjusted by a random amount between -1 and 5 seconds.
- The retry strategy will now respect server-defined backoff periods, defined in the `Retry-After` response header. These headers will only be respected for status codes `423 Locked`, `429 Too Many Requests` and any 5xx status. If the `Retry-After` header is absent from the repsonse, the agent will retry with the default exponential backoff strategy
- The agent will no longer retry responses with a 4xx status code that isn't `423 Locked` or `429 Too Many Requests`

As an ancillary change, the agent wasn't respecting SIGTERMs while in the backoff loop attempting to acquire a job. This is a bug that's unlikely to occur in customer environments, but it was sure annoying when testing these changes locally, so i've added a little listener that cancels the context if it detects that the `AgentWorker` is stopping.

### Context

PIPE-631

### Testing

### DIFF
--- a/core/client.go
+++ b/core/client.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"math/rand/v2"
 	"net/http"
 	"os"
 	"runtime"
@@ -48,7 +47,7 @@ type Client struct {
 
 // AcquireJob acquires a specific job from Buildkite.
 // It doesn't interpret or run the job - the caller is responsible for that.
-// It contains a builtin timeout of 270 seconds and makes up to 10 attempts.
+// It contains a builtin timeout of 330 seconds and makes up to 7 attempts, backing off exponentially.
 func (c *Client) AcquireJob(ctx context.Context, jobID string) (*api.Job, error) {
 	c.Logger.Info("Attempting to acquire job %s...", jobID)
 
@@ -63,13 +62,15 @@ func (c *Client) AcquireJob(ctx context.Context, jobID string) (*api.Job, error)
 	timeoutCtx, cancel := context.WithTimeout(ctx, 330*time.Second)
 	defer cancel()
 
-	// Acquire the job using the ID we were provided. We'll retry as best we can on non 422 error.
-	// Except for 423 errors, in which we exponentially back off under the direction of the API
-	// setting the Retry-After header
+	// Acquire the job using the ID we were provided.
+	// We'll retry as best we can on non 5xx errors, as well as 423 Locked and 429 Too Many Requests.
+	// For retryable errors, if available, we'll consume the value of the server-defined `Retry-After` response header
+	// to determine our next retry interval.
+	// 4xx errors that are not 423 or 429 will not be retried.
 	r := roko.NewRetrier(
 		roko.WithMaxAttempts(7),
 		roko.WithStrategy(roko.Exponential(2*time.Second, 0)),
-		roko.WithJitterRange(-time.Second, 5*time.Second),
+		roko.WithJitterRange(-1*time.Second, 5*time.Second),
 		roko.WithSleepFunc(c.RetrySleepFunc),
 	)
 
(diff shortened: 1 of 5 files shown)


core/client.go
```diff
@@ -57,18 +57,19 @@ func (c *Client) AcquireJob(ctx context.Context, jobID string) (*api.Job, error)
...
 
 	// Acquire the job using the ID we were provided. We'll retry as best we can on non 422 error.
 	// Except for 423 errors, in which we exponentially back off under the direction of the API
 	// setting the Retry-After header
 	r := roko.NewRetrier(
-		roko.WithMaxAttempts(10),
-		roko.WithStrategy(roko.Constant(3*time.Second)),
+		roko.WithMaxAttempts(9),
+		roko.WithStrategy(roko.Exponential(2*time.Second, 0)),
+		roko.WithJitterRange(-time.Second, 5*time.Second),
```
moskyb, line 72: source: i made it up. happy to change this to something more considered.

core/client.go
```diff
@@ -56,19 +55,20 @@ func (c *Client) AcquireJob(ctx context.Context, jobID string) (*api.Job, error)
...
+	timeoutCtx, cancel := context.WithTimeout(ctx, 330*time.Second)
 	defer cancel()
 
 	// Acquire the job using the ID we were provided. We'll retry as best we can on non 422 error.
 	// Except for 423 errors, in which we exponentially back off under the direction of the API
 	// setting the Retry-After header
 	r := roko.NewRetrier(
-		roko.WithMaxAttempts(10),
-		roko.WithStrategy(roko.Constant(3*time.Second)),
+		roko.WithMaxAttempts(7),
```
keithpitt, line 69: Drive by question: do we want any runtime operational control of this number? i.e. maybe something that's return in the agent registration endpoint?

I know we could build that gear — but the real question is: _**do we want it?**_

moskyb: side note: this comment made me realise that that comment is now a lie. i'll fix that.

to your actual question, for this particular value (the max retry count for acquire job), i don't really think so. acquire mode is generally intended for customers running their own customer schedulers, and if their acquire fails after 7 (in this case) attempts, they can probably call `buildkite-agent start --acquire-job xxx` again, so i think the utility of being able to change it serverside would be limited.

in a more general sense, being able to remote-customise certain agent behaviour would be super useful - doing stuff like dynamically slowing down ping intervals or setting an agent-wide rate limit ("you in particular are only allowed to make 1 request per second now", or something) would be pretty cool - i think we'd need a pretty useful initial thing to do that for though.

CerealBoy (approved):
LGTM 👍🏼 

I think we could look at having further customisation options in the future, perhaps not for right now though.

Splitting out more of the response codes and handling them more pedantically is a great add here, especially with the comments. that provide the guidance around why these are handled the way they will be.

PR #3157, opened 2025-01-14
Roko v1.3.1

moskyb:
### Description

This PR fixes two issues introduced in https://github.com/buildkite/agent/pull/3153:
- The roko retrier would lie about the time it was going to retry in when jitter was enabled - see https://github.com/buildkite/roko/pull/18. This problem is solved by updating to roko v1.3.1, which includes the PR linked.
- In the acquire-job retry loop, in some cases, the server can set the amount of time that the retrier can retry in. However, because the `Retrying in...` message was logged prior to setting the next interval, the retrier would log the interval it calculated instead of the interval that the backend told it to use. This issue is resolved by simply logging after we set the next retry interval.

### Testing


CerealBoy (approved):
LGTM

PR #3164, opened 2025-01-22
Jitter within ping, status, log loops

DrJosh9000:
### Description

More techniques to avoid thundering herds - introduce jittering inside fixed interval loops.

### Context

This is a change I've wanted to make for a while, but never really got around to it.

The job status checker loop (used to stop a job cancelled server-side) uses a sleep on a fixed interval between requests, meaning it is susceptible to spontaneous synchronisation between agents. 

The log processing loop is similar to the job status loop, with fixed 1 second wait. It's not directly connected to a network request, but the pipeline is tight.

The ping loop uses a ticker, which has internal mechanisms for keeping a fixed pace. But this therefore preserves initial synchronisation between agents (e.g. multiple machines come up at once due to, say, a power outage, and they proceed to start the agent at the same time).

### Changes

* Make the job status loop use a ticker
* Make the log processing loop use a ticker
* Make all three loops wait a random duration before continuing
* Tweak the implementation of the goroutine that waits for stopping within AcquireAndRunJob.

### Testing

### DIFF
--- a/agent/agent_worker.go
+++ b/agent/agent_worker.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math/rand/v2"
 	"net/http"
 	"sync"
 	"time"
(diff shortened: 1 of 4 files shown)


CerealBoy (approved):
LGTM! Brilliant use of 🫨 too

agent/agent_worker.go
```diff
@@ -291,85 +292,102 @@ func (a *AgentWorker) runPingLoop(ctx context.Context, idleMonitor *IdleMonitor)
...
+		// Within the interval, wait a random amount of time to avoid
+		// spontaneous synchronisation across agents.
+		jitter := rand.N(pingInterval)
+		setStat(fmt.Sprintf("🫨 Jittering for %v", jitter))
+		select {
+		case <-time.After(jitter):
+			// continue below
+		case <-a.stop:
+			return nil
+		}
```
pda, line 322: @DrJosh9000 I might be misunderstanding some aspect of this, but I wonder if `pingInterval` (often 10 seconds) is too much jitter?

My understanding is that `pingTicker` will trigger every (e.g.) 10 seconds, adjusting/dropping ticks to stay aligned with its start time when the channel is blocked.

So we're waiting 10 seconds for the tick, and then up to 10 seconds for the jitter? And then the remainder of that 10 seconds before the next tick.

I think that means the actual time between two pings (and therefore time waiting for a job to be accepted) can be up to double the server-specified PingInterval?

Also, is it possible that a maximum-duration jitter will cause the next tick to be dropped?

I wonder if the following timeline is possible?

```
T=0 <-pingTicker.C
T=1
T=2 <-time.After(jitter) … send ping
T=3 (waiting for ping response)
T=4 process ping response
T=5
T=6
T=7
T=8
T=9
T=10 <-pingTicker.C
T=11
T=12
T=13
T=14
T=15
T=16
T=17
T=18
T=19 <-time.After(jitter) … send ping (17 seconds after previous)
T=20 … (waiting for ping response; <-pingTicker.C DROPPED)
T=21 process ping response
T=22
T=23
T=24
T=25
T=26
T=27
T=28
T=29
T=30 <-pingTicker.C
T=31
T=32
T=33
T=34
T=35
T=36
T=37 <-time.After(jitter) … send ping (18 seconds after previous)
T=38 (waiting for ping response)
T=39 process ping response
…
```

Might reducing the jitter to be e.g. half the ping duration improve things?

DrJosh9000: It's true that the time between successive pings can be double the interval, but it should not happen multiple times in a row.

One aspect of `time.Ticker` that is not in [its doc comment](https://cs.opensource.google/go/go/+/go1.24.2:src/time/tick.go;l=36), but would be incredibly difficult for the Go team to change, is that the ticker channel is buffered (with a size of 1):

https://cs.opensource.google/go/go/+/refs/tags/go1.24.2:src/time/tick.go;l=40-42

So suppose we have the situation where the jitter waits a full 10 seconds after the start of the loop (the last receive from the ticker channel). The ping takes some non-zero amount of time. The ticker will have sent on the channel during the ping, and that would not have blocked because of the channel buffer. When the loop begins again it can receive from the channel immediately since that there is a value already in the buffer. In other words, ticks won't be dropped (unless the ping itself takes over 10 seconds, so causing the ticker to block sending on a full channel).

The ticker _might_ send the next tick slightly early to account for the 10 seconds plus the time to ping causing the receive to be slower than normal, but that should be temporary.

pda: Ah right, I had assumed it's unbuffered 👍🏼

I still wonder if a smaller jitter value e.g. 20% or 50% of PingInterval would give similar benefit while reducing the extra wait time for agents to pick up jobs 🤔

DrJosh9000: In the worst case, the tickers of a large group of agents would all be synchronised. If the jitter was 20% of the ticker interval, then the backend would receive 5x the current traffic for 2 seconds, followed by no traffic for 8 seconds - in order to reduce the maximum possible inter-ping time from 20 to 12 seconds. 

If we wanted to focus on the maximum inter-ping time of a single agent (and I believe that the 10 second interval happens when there are at least so many agents, which is why I wasn't so worried when I wrote this PR), then perhaps we should consider a scheme like this:

1. Ping immediately at the start of the loop
2. Jitter the full interval
3. Run the loop with the fixed interval and no jitter within it
4. After the ping, roll a 10-sided die; if it rolls a 1, jitter the full interval and then reset the ticker

This would ensure the inter-ping time is 10 seconds most of the time and some random smaller duration 10% of the time, but smooths the traffic and would avoid spontaneous synchronisation.

PR #3165, opened 2025-01-22
Introduce http client profile option for custom configuration

wolfeidau:
### Description

We have customers using the agent in a number of different environments, across different cloud services all over the world, so different configuration profiles for the HTTP client would be valueble.

### Context

The Go HTTP client has a log of configurable timeouts, passing all these through is not sustainable so we need a way to bless a subset of configuration and provide this with a generic name, hence the idea of configuration profiles.

On top of this different version of Go introduce new features, and new timeouts, so we also need a way to introduce new profiles as the HTTP Client evolves.

### Changes

This change introduces a flag which allows us to add HTTP client configurations which allows us to provide new profiles for customers, while also providing an unconfigured standard library profile to take advantage of the current defaults.

To get the ball rolling I have created a stdlib profile which applies the absolulte minimum configuration, this includes tlsconfig overrides, and the timeout value. Disabling H2 is not provided as this is avaliable via the `GODEBUG` environment variable supported by the Go Team. https://pkg.go.dev/net/http#hdr-HTTP_2

### Testing

### DIFF
--- a/internal/agenthttp/client.go
+++ b/internal/agenthttp/client.go
@@ -26,6 +26,18 @@ func NewClient(opts ...ClientOption) *http.Client {
 		opt(&conf)
 	}
 
+	switch conf.HTTPClientProfile {
+	case "stdlib":
+		return &http.Client{
+			Timeout: conf.Timeout,
+			Transport: &authenticatedTransport{
+				Bearer:   conf.Bearer,
+				Token:    conf.Token,
+				Delegate: http.DefaultTransport,
+			},
+		}
+	}
+
 	cacheKey := transportCacheKey{
 		AllowHTTP2: conf.AllowHTTP2,
 		TLSConfig:  conf.TLSConfig,
@@ -65,6 +77,9 @@ func WithAllowHTTP2(a bool) ClientOption       { return func(c *clientConfig) { 
 func WithTimeout(d time.Duration) ClientOption { return func(c *clientConfig) { c.Timeout = d } }
 func WithNoTimeout(c *clientConfig)            { c.Timeout = 0 }
 func WithTLSConfig(t *tls.Config) ClientOption { return func(c *clientConfig) { c.TLSConfig = t } }
+func WithHTTPClientProfile(p string) ClientOption {
+	return func(c *clientConfig) { c.HTTPClientProfile = p }
+}
 
 type ClientOption = func(*clientConfig)
 
@@ -122,6 +137,9 @@ type clientConfig struct {
 
 	// optional TLS configuration primarily used for testing
 	TLSConfig *tls.Config
+
+	// HTTPClientProfile profile to use for the http client
+	HTTPClientProfile string
 }
 
 // The underlying http.Transport is cached, mainly so that multiple clients with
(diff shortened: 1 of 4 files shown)


internal/agenthttp/client.go
```diff
@@ -26,6 +30,29 @@ func NewClient(opts ...ClientOption) *http.Client {
 		opt(&conf)
 	}
 
+	// http client profile is used to switch between different http client implementations
+	// - stdlib: uses the standard library http client
+	switch conf.HTTPClientProfile {
+	case "stdlib":
```
DrJosh9000, line 36: I think the one thing missing here is support for `AllowHTTP2: false`.

DrJosh9000 (commented):
I like the concept!

internal/agenthttp/client.go
wolfeidau: @DrJosh9000 I put a note in the PR about this, as it is quite complex and already a feature of the STDLIB maybe we do it the standard library way?

> Disabling H2 is not provided as this is avaliable via the GODEBUG environment variable supported by the Go Team. https://pkg.go.dev/net/http#hdr-HTTP_2

DrJosh9000: /me looks again

Ah, yep, carry on.

DrJosh9000 (approved):
LGTM 👍 
On the HTTP/2 issue, I think it might lack visibility if someone is trying to use a profile and disable it the legacy way. May be worth checking and logging or even fatal-ing?

wolfeidau:
@DrJosh9000 great point, the agent now reports a conflict for that particular setting.

```
2025-01-31 14:07:04 FATAL  NoHTTP2 is not supported with the standard library (stdlib) HTTP client profile, use GODEBUG see https://pkg.go.dev/net/http#hdr-HTTP_2
```

internal/agenthttp/client.go
```diff
@@ -26,6 +34,29 @@ func NewClient(opts ...ClientOption) *http.Client {
 		opt(&conf)
 	}
 
+	// http client profile is used to switch between different http client implementations
+	// - stdlib: uses the standard library http client
+	switch conf.HTTPClientProfile {
+	case ClientProfileStdlib:
+		// Base any modifications on the default transport.
+		transport := http.DefaultTransport.(*http.Transport).Clone()
```
patrobinson, line 42: As far as I can tell https://go-review.googlesource.com/c/net/+/617655 is merged but not released. So using the default option would re-introduce the bug we fixed in #3005?

patrobinson (commented):
I like the idea, I have a question about if that patch has been released?

internal/agenthttp/client.go
wolfeidau: @patrobinson Given 1.24 is coming out soon, with 1.24rc3 coming out a few days ago, I think I am going to wait rather than add this workaround.

I have confirmed this fix is bundled with 1.24 in https://github.com/golang/go/blob/release-branch.go1.24/src/net/http/h2_bundle.go#L7834-L7840 🤞🏻.

DrJosh9000:
Closing due to staleness, but please resuscitate if this is still something you'd like to see happen, @wolfeidau

PR #3166, opened 2025-01-23
Better plugin checkout logging

### LINKED ISSUE #3090: Bug: plugin checkout dir printed incorrectly
I'm debugging a plugin. I have a terminal open into my agent mid-job.

Earlier, it printed this:

image

I copied the path from line 36 but it doesn't exist:

image

Then I noticed there's a different path on line 34 - that one worked:

image

Is this a bug?

DrJosh9000:
### Description

Fixes #3090 

### Changes

- Change the temp directory pattern to be less reminiscent of a plugin directory
- Log the `mktemp` and `mv` operations similarly to the `cd` operation

To think about: these new logged operations should probably appear different on Windows 🤔 

### Testing

### DIFF
--- a/internal/job/plugin.go
+++ b/internal/job/plugin.go
@@ -15,6 +15,7 @@ import (
 	"github.com/buildkite/agent/v3/internal/job/hook"
 	"github.com/buildkite/agent/v3/internal/osutil"
 	"github.com/buildkite/roko"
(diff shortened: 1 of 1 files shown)


wolfeidau (approved):
👍🏻 NICE

PR #3171, opened 2025-01-30
Clean up worker pool implementation

DrJosh9000:
### Description

Simplify.

* It's not important that we only return the first error - we can return all errors joined and `errors.Is` (used for changing the exit code in acquire mode, a few levels up the call stack) will continue to work - but it's better to report all errors.
* Conditionally sending on a channel makes balancing receives harder, which the old implementation worked around with a waitgroup and yet another goroutine that closed the channel after all workers finished. Always sending one value per worker means receiving one value per worker, which eliminates the waitgroup and extra goroutine.
* `errors.Join` does the Obviously Correct Thing with empty slices and nil errors.

### Context

Saw this again while investigating a bug, and I decided that I've had enough.

### Testing

### DIFF
--- a/agent/agent_pool.go
+++ b/agent/agent_pool.go
@@ -3,10 +3,10 @@ package agent
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
 	"strconv"
-	"sync"
 
 	"github.com/buildkite/agent/v3/logger"
 	"github.com/buildkite/agent/v3/status"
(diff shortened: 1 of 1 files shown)


agent/agent_pool.go
```diff
@@ -56,31 +56,23 @@ func (r *AgentPool) Start(ctx context.Context) error {
...
-	}()
-
-	return <-errs
+	// Number of receives = number of sends
+	errs := make([]error, 0, len(r.workers))
+	for range r.workers {
+		errs = append(errs, <-errCh)
+	}
+	return errors.Join(errs...) // nil if all errs are nil
 }
```
wolfeidau, line 76: Nice use of this new feature!

wolfeidau (approved):
👍🏻

PR #3172, opened 2025-01-30
Timeout waiting for client containers

DrJosh9000:
### Description

Ensure the agent process does not wait forever for the client containers to connect.

### Context

Addresses one cause of agent containers continuing to run, as reported in buildkite/agent-stack-k8s#479

### Changes

Adds a `startupCheck` that waits for all client containers to connect. If one or more fail to connect within a timeout, an error is returned.

### Testing

### DIFF
--- a/kubernetes/client.go
+++ b/kubernetes/client.go
@@ -68,26 +68,37 @@ func (c *Client) Write(p []byte) (int, error) {
 var ErrInterrupt = errors.New("interrupt signal received")
 
 func (c *Client) Await(ctx context.Context, desiredState RunState) error {
+	ticker := time.NewTicker(time.Second)
+	defer ticker.Stop()
+
+	// Because [time.Ticker] doesn't tick until after the duration (time.Second),
+	// but we want to call the RPC method in the first loop iteration without
+	// waiting, here's a channel (first) that can be received from once and then
+	// never again, providing a non-blocking path through the select on the
+	// first iteration.
+	first := make(chan struct{}, 1)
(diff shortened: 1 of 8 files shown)


kubernetes/client.go
```diff
@@ -68,26 +68,32 @@ func (c *Client) Write(p []byte) (int, error) {
...
-			if current == RunStateInterrupt {
-				return ErrInterrupt
-			}
-			select {
-			case <-ctx.Done():
-				return ctx.Err()
-			case <-time.After(time.Second):
-			}
+
+		case <-first:
```
wolfeidau, line 81: whats with the first thing?

wolfeidau (commented):
Looks good just one question.

kubernetes/client.go
DrJosh9000: `time.Ticker` doesn't tick until after the duration, but we want to call the RPC immediately. `first` provides a path through the select that isn't blocked on the first iteration, but is blocked forever afterwards.

wolfeidau: Nifty, trick, so it is like "once" or "immediate", I get flash back to `do while` 😂

wolfeidau: @DrJosh9000 Maybe add a comment just to make this obvious?

DrJosh9000: Done.

wolfeidau (approved):
👍🏻 🎉 thanks!

PR #2588, opened 2024-01-16
Replace roko.TryForever with exponential backoff for ~24 hours

tessereth:
Occasionally something goes wrong and the agent gets stuck retrying job log uploads or job finish calls forever (eg because the agent's access token is revoked). Forever is a really long time. Instead just retry for ~24 hours. I also changed these to exponential backoff rather than constant time because if a request fails, it's likely to either be fixed very quickly (network blip) or take a long time (incident). So no point retrying every second if it's already failed a hundred times. I also made them both the same and 2 seconds initial interval to match other parts of the code.

Using the `ExponentialSubsecond` algorithm with a 2 second initial interval, this is what the retry intervals looks like:

Retry attempt | Delay | Cumulative delay
-- | -- | --
0 | 0:00:02 | 0:00:02
1 | 0:00:03 | 0:00:05
2 | 0:00:05 | 0:00:10
3 | 0:00:08 | 0:00:19
4 | 0:00:13 | 0:00:32
5 | 0:00:22 | 0:00:54
6 | 0:00:35 | 0:01:28
7 | 0:00:56 | 0:02:24
8 | 0:01:29 | 0:03:53
9 | 0:02:24 | 0:06:17
10 | 0:03:51 | 0:10:08
11 | 0:06:12 | 0:16:20
12 | 0:09:58 | 0:26:18
13 | 0:16:02 | 0:42:20
14 | 0:25:47 | 1:08:07
15 | 0:41:27 | 1:49:35
16 | 1:06:40 | 2:56:15
17 | 1:47:12 | 4:43:27
18 | 2:52:24 | 7:35:51
19 | 4:37:14 | 12:13:05
20 | 7:25:50 | 19:38:55
21 | 11:56:56 | 31:35:51

PR #3177, opened 2025-02-04
Reduce timeout for these two operations to avoid holding up compute

wolfeidau:
### Description

When there is a failure between a customer and buildkite we "could" hold onto compute for up to 48 hours.

I am starting with these two calls as they are important calls which can block agent shutting down.

### Context

If the buildkite API, or intermediate network devices, such as nat GW trigger drop outs and timeouts of requests over an extended period of time we can tie up compute resources, this isn't ideal as it can hold onto expensive resources.

### Changes

Reducing these two values avoids surprising outcomes for customers when things go wrong.

### Testing

### DIFF
--- a/core/client.go
+++ b/core/client.go
@@ -219,7 +219,7 @@ func (c *Client) FinishJob(ctx context.Context, job *api.Job, finishedAt time.Ti
 	return roko.NewRetrier(
 		// retry for ~a day with exponential backoff
 		roko.WithStrategy(roko.ExponentialSubsecond(2*time.Second)),
-		roko.WithMaxAttempts(20),
+		roko.WithMaxAttempts(12), // 12 attempts will take 26 minutes
 		roko.WithJitter(),
 		roko.WithSleepFunc(c.RetrySleepFunc),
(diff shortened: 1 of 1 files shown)


yob:
I think historically we'd had these high to ensure completed jobs will eventually be marked as finished (and have all job logs flushed) through long network outages (on either the customer side or our side). I can definitely see how holding an expensive compute resource for 48 hours would be not ideal in some cases though 🤔

DrJosh9000:
These retry loops were last changed in #2588. Limiting to 1 hour will implicitly limit the number of retries to 13 (according to the table from that previous PR), so maybe we should reflect that in the `WithMaxAttempts` options too. 

I think I'm OK with shipping this without an escape hatch, because even 24 hours seems excessive, but I know we're thinking about how to better configure all these retry loops. :thinking:

DrJosh9000 (approved):

wolfeidau:
@DrJosh9000 I have reduced the retry count to 12, allowing for connection timeouts which will introduce further delays.

DrJosh9000 (approved):
LGTM

PR #3179, opened 2025-02-06
Speed up needlessly slow tests

DrJosh9000:
### Description

These two tests take 40 - 50 seconds each. They can be short-circuited a bit (without affecting their fidelity).

### Context

Noticed by @pda here: https://buildkite-corp.slack.com/archives/C061U5PG9K5/p1738728903252739

### Changes

* Instead of trying to send traces to a non-existent DD agent (and timing out), use `mocktracer`.
* Instead of trying to cancel a job against a fake server that always returns HTTP 500, return HTTP 400 which aborts the retry quickly.

### Testing


moskyb (approved):
🎉

PR #3180, opened 2025-02-07
Ensure the log streamer respects forced shutdown of the agent

wolfeidau:
Currently if the agent is sent a sigterm the log streamer will wait for all chunks to be uploaded, and even if a second sigterm is sent we still wait.

Other components of the agent honor the a graceful=false flag which allows operators to force the agent to stop.

This change adds that graceful option to the log streamer stop.

### Description

This change introduces immediate cancellation to the log streamer when the agent is forced to stop.

### Context

If customers egress is overloaded, or there is a large backlog of chunks to upload, and that is blocked due to internet connectivity issues, the agent can block for extended periods waiting to upload chuncks even if has been "forced" to quit.

### Changes

Adds a simple way to signal to the uploader workers to stop work and quite immediately when the agent has been told to exit and cancel.

### Testing

### DIFF
--- a/agent/log_streamer.go
+++ b/agent/log_streamer.go
@@ -66,6 +66,9 @@ type LogStreamer struct {
 
 	// Have we stopped?
 	stopped bool
+
+	// have we been asked to exit immediately?
+	exitImmediately bool
 }
 
 // NewLogStreamer creates a new instance of the log streamer.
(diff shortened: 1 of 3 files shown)


CerealBoy (approved):
LGTM!

PR #3187, opened 2025-02-11
Fix data race on exitImmediately

DrJosh9000:
### Description

* What it says on the title
* Also switches to `atomic.Int32` for `chunksFailedCount`.

### Testing


wolfeidau (approved):
Thanks, 🤦🏻 didn't consider an atomic. 🎉

PR #3193, opened 2025-02-13
Capture some HTTP client details from registering agents

yob:
Agent registration requests include an array of "features" that are enabled on the agent. To assist with debugging, it would be helpful to know if a particular agent:

* has opted into HTTP 1.1 only
* is using a HTTP proxy
* has set GODEBUG, which [can be used to change the behaviour of net/http](https://pkg.go.dev/net/http)

We're submitting a boolean only, to indicate the feature is active. We don't submit values, such as the address of the HTTP proxy.

It would also be quite helpful to have the agent log out the state of its connection pool to agent.buildkite.com, but that's quite a different change to this so I've left it out and kept this PR simple.

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -236,6 +236,10 @@ func (asc AgentStartConfig) Features(ctx context.Context) []string {
 		features = append(features, "no-script-eval")
 	}
 
+	if asc.NoHTTP2 {
+		features = append(features, "no-http2")
+	}
+
 	if len(asc.AllowedRepositories) > 0 {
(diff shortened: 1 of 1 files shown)


wolfeidau:
@yob we should check if `GODEBUG` is set as well, so we know if customers are using some of the internal runtime flags, see https://pkg.go.dev/net/http.

yob:
Oh yer, good one. We wouldn't know the value, but would still be helpful to know when GODEBUG is active and might be changing agent behaviour

wolfeidau (approved):
👍🏻 Looks good

PR #3195, opened 2025-02-19
Capture datadog metrics usage from registering agents

wolfeidau:
As an extension to https://github.com/buildkite/agent/pull/3193 would be good to capture the use of datadog metrics.

As with the previous PR, we're submitting a boolean only, to indicate the feature is active.

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -162,11 +162,14 @@ type AgentStartConfig struct {
 
 	HealthCheckAddr string `cli:"health-check-addr"`
 
+	// Datadog statsd metrics config
 	MetricsDatadog              bool   `cli:"metrics-datadog"`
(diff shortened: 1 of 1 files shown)


DrJosh9000 (approved):
Makes sense, LGTM

pda:
Is this PR redundant with this existing code?
https://github.com/buildkite/agent/blob/main/clicommand/agent_start.go#L215-L217

wolfeidau:
@pda yeah this is a seperate feature which sends metrics to dogstatsd which is used by datadog. [metrics/metrics.go](https://github.com/buildkite/agent/blob/main/metrics/metrics.go) as posed to tracing.

wolfeidau:
I split these flags and added a comment for future adventures into this area. cc @DrJosh9000 @pda

PR #3197, opened 2025-02-19
Update cancel_signal.go

karensawrey:
Update to fix an error in the description to fix an error in the Buildkite Docs.

### Description

Fixes an error in https://buildkite.com/docs/agent/v3/cli-bootstrap#signal-grace-period-seconds as this page in the Buildkite Docs is being generated from a script and the script is dependent on the content of this file the PR is updating in https://github.com/buildkite/agent

### Context

https://linear.app/buildkite/issue/SUP-3416/update-the-docs-for-signal-grace-period-seconds-value

### Changes

A two-word change, please see the diff.

### Testing
The change does not touch any functionality, only a description.

### DIFF
--- a/clicommand/cancel_signal.go
+++ b/clicommand/cancel_signal.go
@@ -30,7 +30,7 @@ var (
 		Usage: "The number of seconds given to a subprocess to handle being sent ′cancel-signal′. " +
 			"After this period has elapsed, SIGKILL will be sent. " +
 			"Negative values are taken relative to ′cancel-grace-period′. " +
-			"The default is ′signal-grace-period-seconds′ - 1.",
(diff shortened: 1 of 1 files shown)


DrJosh9000 (commented):
Thanks! I agree it could be worded better - I'm not sure the change makes the situation clearer. 

The default value of the `signal-grace-period-seconds` flag (`-1`) is interpreted to be relative to the value for `cancel-grace-period-seconds`, so perhaps this sentence should say:

"The default value (-1) means the effective signal grace period is equal to ′cancel-grace-period-seconds′ minus 1."

karensawrey:
Thanks, @DrJosh9000! I'll revisit and update this during my day!

DrJosh9000 (approved):

DrJosh9000:
Thanks for this @karensawrey !

PR #3198, opened 2025-02-20
Add agent stop command

sj26:
Add a `buildkite-agent stop [--force]` command.

The main use case is to allow customers to write "health checks" in agent and job hooks which can tell the agent to stop before or after the current job. For example, a `pre-command` agent hook could detect a spot instance has been asked to terminate and forcefully stop the agent, and allowing the use of an `AGENT_STOP` signal reason to automatically retry in that case. Or a `post-command` agent hook could detect a 95% full disk and tell the agent to stop after the current job completes.

Implementation is mostly extrapolated from what exists, I'm not here to set up new patterns 🙈

This PR only adds a basic test to make sure the plumbing works. There's not a lot of maturity in the testing, so this felt like a nice add without going full ham on figuring out how to do more comprehensive testing. We could also test the error response (a 422) and the force parameter, but manual testing shows they work.


DrJosh9000 (approved):
LGTM ✅ 
(pending the server side, obviously)

sj26:
The Agent API stop endpoint has shipped, so this is good to go

PR #3199, opened 2025-02-21
Don't fail if the interrupt fails when the PID is already exited

wolfeidau:
While testing a number of failure scenarios which simulate network failures I can across this error, which occurs if an agent is sent a `SIGTERM` and tries to kill a process which is already exited.

This sort of race condition is caused by outages of the buildkite api, or network interuptions.

One open question I have is the switch linked below in the correct order? Given the error being raised skips the timeout, and the done check being after this timeout which seems odd.

https://github.com/buildkite/agent/blob/fix_dont_fail_interrupt_if_pid_exited/agent/run_job.go#L504-L520

### DIFF
--- a/process/process.go
+++ b/process/process.go
@@ -361,6 +361,12 @@ func (p *Process) Interrupt() error {
 
 	// interrupt the process (ctrl-c or SIGINT)
 	if err := p.interruptProcessGroup(); err != nil {
+		//  No process or process group can be found corresponding to that specified by pid.
+		if errors.Is(err, syscall.ESRCH) {
+			p.logger.Warn("[Process] Process %d has already exited", p.pid)
+			return nil
+		}
+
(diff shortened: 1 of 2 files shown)


DrJosh9000:
> One open question I have is the switch linked below in the correct order? Given the error being raised skips the timeout, and the done check being after this timeout which seems odd.
> 
> https://github.com/buildkite/agent/blob/fix_dont_fail_interrupt_if_pid_exited/agent/run_job.go#L504-L520

I may be misunderstanding, but I'm fairly sure it's immaterial what order the case statements are within a `select` - it blocks until any channel operation can proceed, then proceeds that case. So either `<-r.process.Done()` (the subprocess exited first) or `<-time.After(...)` (the timeout happened first).

DrJosh9000 (approved):
🍏

PR #3200, opened 2025-02-24
Change the signal handler to ensure the agent quits after the grace period

wolfeidau:
### Description

This uses two new calls to which are only triggered after the second quit attempt to ensure the agent exits:

1. After half the cancel grace period (defaults to at least 10s) the agent will cancel the context.
2. At the end of the cancel grace period, if the agent is stuck it will call os.Exit(1).

The pool stop with graceful false should, if at all possible, trigger disconnect calls for all the agents. 

> [!NOTE]
> This code logic is only invoked after a second exit signal.

### Context

If the agent is unable to commicate with the agent API it can take almost an hour to stop.

### Changes

Adds more logic to the signal handler to ensure the agent exits.

### Testing

To test this logic I used https://github.com/smarty/cproxy as a HTTP proxy with the agent configured to use it, then once a job was started and the process was running I killed the proxy, interupted the agent twice and waited for the service to exit.

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -63,6 +63,8 @@ Example:
     $ buildkite-agent start --token xxx`
 
 var (
+	minGracePeriod = 10
+
 	verificationFailureBehaviors = []string{agent.VerificationBehaviourBlock, agent.VerificationBehaviourWarn}
 
 	buildkiteSetEnvironmentVariables = []*regexp.Regexp{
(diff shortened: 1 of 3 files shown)


internal/job/grace.go
DrJosh9000, line 1: 💭 This little helper be moved to a different package, given it will be no longer job-specific.

clicommand/agent_start.go
```diff
@@ -1566,3 +1588,11 @@ func envHasKey(key string) bool {
 	return ok
 
 }
+
+func cancelGracePeriodDuration(cancelGracePeriod int) time.Duration {
+	if cancelGracePeriod < 10 {
+		return 10 // minimum 10 seconds for forceful shutdown
```
DrJosh9000, line 1594: The function's return type is `time.Duration`, but `time.Duration(10)` means 10 nanoseconds. If we keep this function it ought to return a `time.Duration` that represents the intended duration (as in, multiply by `time.Second` within).

clicommand/agent_start.go
```diff
@@ -1392,8 +1396,26 @@ func handlePoolSignals(ctx context.Context, l logger.Logger, pool *agent.AgentPo
...
-					pool.Stop(false)
+					l.Info("Forcefully stopping running jobs and stopping the agent(s) in %d seconds", cancelGracePeriod)
+
+					gracefulContext, _ := job.WithGracePeriod(ctx, cancelGracePeriodDuration(cancelGracePeriod)*time.Second)
+
+					go func() {
+						l.Info("Forced agent(s) to stop")
+						pool.Stop(false) // one last chance to stop
+
+						time.Sleep(cancelGracePeriodDuration(cancelGracePeriod/2) * time.Second)
```
DrJosh9000, line 1407: `cancelGracePeriodDuration` _could_ be eliminated with the new `max` builtin:

```suggestion
						time.Sleep(max(cancelGracePeriod/2, 10) * time.Second)
```

clicommand/agent_start.go
```diff
@@ -1392,8 +1396,26 @@ func handlePoolSignals(ctx context.Context, l logger.Logger, pool *agent.AgentPo
...
+
+						l.Info("Cancelling all internal tasks and API requests")
+						cancel() // cancel the context to stop all network operations
+					}()
+
+					<-gracefulContext.Done()
```
DrJosh9000, line 1413: This section is unusual since it's what happens if the program hasn't already exited, but looks like part of the "normal flow" once the second interrupt is received. May be worth a comment explaining why "forceful stop" within the grace period doesn't lead to `os.Exit(1)`: 

```suggestion
					// Once pending retries and requests are cancelled, 
					// the main goroutine should exit before this grace period expires, ending the program.
					// If that doesn't happen, exit 1 below.
					<-gracefulContext.Done()
```

DrJosh9000 (approved):
Mostly LGTM, just some small things.

wolfeidau:
@DrJosh9000 I added your suggestions and tweaked it to remove the helper function, see what you think.

internal/job/grace.go
wolfeidau: I am open to suggestions here, agree it should move out, just not sure were best to put it.

DrJosh9000: Not urgent, we can come back to it later.

DrJosh9000 (approved):
Still LGTM 👍

PR #3201, opened 2025-02-24
Set BUILDKITE_SIGNAL_REASON in time for post-command hook

DrJosh9000:
### Description

Make the signal reason available to post-command hooks.

### Context

https://linear.app/buildkite/issue/PIPE-481/add-signal-reason-to-the-post-command-hook

### TODO

Most of the Kubernetes side (establish an RPC method to poll for stdin? or something tailored for env vars specifically?)

### Changes

* process: Copy data from stdin to the pty
* agent: Set up a pipe for the process to read and Cancel to write
* agent: On cancellation, write an environment variable into stdin indicating either that the job is cancelled or the agent is stopping
* internal/job: Read stdin for env vars to set

### Why stdin 🤔 

The "bootstrap" has a Job API, and even has methods for manipulating env vars. Accessing this needs the socket path and token, which are only available to the bootstrap process and its subprocesses. I guess we could sneak that info back to the agent through the bootstrap's stdout or stderr...

### Testing

### DIFF
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -135,6 +135,9 @@ type JobRunner struct {
 	// The internal process of the job
 	process jobAPI
 
+	// Writes input to the internal process of the job
+	processStdin io.Writer
+
 	// The internal buffer of the process output
 	output *process.Buffer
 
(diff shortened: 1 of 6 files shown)


DrJosh9000:
#3213 is better.

PR #3210, opened 2025-02-25
Kill exp/maps and replace with stdlib maps

moskyb:
### Description

I, a bad and evil person, introduced the `exp/maps` library into this project before it was introduced to the stdlib. As penance, i give to you this gift: `exp` has been excised from this repo and cast into the fires that birthed it.

### Testing


DrJosh9000 (approved):
Code: LGTM
Admitting your sins: Approved
Don't beat yourself up too hard!

PR #3211, opened 2025-02-26
Handle pause actions

DrJosh9000:
### Description

Handle "pause" as a ping action.

### Context

https://linear.app/buildkite/issue/PIPE-747/agent-prevent-paused-agent-terminating-in-one-shot-modes

### Changes

Important bits:
* Don't return early when using acquire-job, so that there's a ping loop after it, but exit from the ping loop instead
* Add "pause" as an action returned from ping, use it to `continue` the ping loop (avoiding either running a job or exiting the loop)

Refactoring/cleanup:
* Rearrange Ping so that it returns the action, and handle the action in the ping loop
* Move idle monitor to its own file
* Replace sync.Mutex+bool+channel with sync.Once+channel for tracking stopping

### Testing

### DIFF
--- a/agent/agent_worker.go
+++ b/agent/agent_worker.go
@@ -87,9 +87,8 @@ type AgentWorker struct {
 	cancelSig process.Signal
 
 	// Stop controls
-	stopMutex sync.Mutex
-	stop      chan struct{}
-	stopping  bool
+	stopOnce sync.Once // prevents double-closing the channel
+	stop     chan struct{}
 
 	// The index of this agent worker
 	spawnIndex int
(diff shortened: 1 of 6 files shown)


sj26:
Nice.

Have you managed to test this gear pretty thoroughly? Can we write a test for it?

DrJosh9000:
I'll write some tests shortly.

sj26:
I got this running locally, and I can set the pause message - nice

image

But there was no indication that the agent was unpaused. Should that log as well?

sj26 (approved):
Nice! Works as advertised locally, and the test looks good to me - although I'm not a Go expert!

DrJosh9000:
> Should that log as well?
You're right, I noticed that all messages from `/ping` are logged, but when it's not paused any more there's no message. I've added something that logs once on the transition from paused back to normal.

PR #3213, opened 2025-02-27
Set env when job cancelled for hooks

sj26:
So that hook behaviour can change when the job has cancelled, either from the platform or because the agent is being stopped, and so its result is likely inconsequential. We have some customers who would like to use this to avoid sending a notification to a system about a result that doesn't matter because it is either incomplete or will be retried.

Alternative to #3201 / PIPE-481.

### Testing

This could use an additional automated test but I'm not sure how complicated that'll be. But this seems a pretty low-risk change in any case.

I manually tested by adding a post-command hook like this:

```bash
if [[ "$BUILDKITE_JOB_CANCELLED" == "true" ]]; then
  echo "Job was cancelled, skipping notifications!"
  exit
fi

echo "Sending notifications ..."
```

and it worked a treat:

image

That should also cover when the agent is stopped forcefully which also cancels the job.

### DIFF
--- a/internal/job/executor.go
+++ b/internal/job/executor.go
@@ -309,6 +309,7 @@ func (e *Executor) Cancel() error {
 		return errors.New("already cancelled")
 	}
 	e.cancelled = true
+	e.shell.Env.Set("BUILDKITE_JOB_CANCELLED", "true")
 	close(e.cancelCh)
 	return nil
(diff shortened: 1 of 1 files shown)


DrJosh9000 (approved):
Nice, LGTM!

PR #3215, opened 2025-02-27
Record build URL in the buildkite-agent log for easier traceability

mkrapivner-zipline:
### Description

As Developer Productivity engineers, we are sometimes tasked with troubleshooting misbehaving agents. To do this, we collect agent metrics and the logs in our o11y platform. This PR is a QoL improvement to record build URL in the buildkite-agent log in addition to the job ID. If the agent has been around for a while and has executed several jobs, it would be great to have easier traceability to see what has been executed on the agent. Having the build URL will allow to quickly locate the executed job in Buildkite UI.

### Testing

### DIFF
--- a/agent/run_job.go
+++ b/agent/run_job.go
@@ -54,7 +54,7 @@ func (e *missingKeyError) Error() string {
 
 // Runs the job
 func (r *JobRunner) Run(ctx context.Context) error {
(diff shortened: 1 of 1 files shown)


wolfeidau (approved):
👍🏻 thanks for the contribution.

PR #3226, opened 2025-03-06
[WIP] Skip if unchanged

DrJosh9000:
### Description

Copies an idea of @lox : annotate steps with `skip_if_unchanged`, and if the pattern doesn't match any of the file paths in `git diff`, convert it into a `skip`.

### Context

TODO

### Changes

TODO

### Testing

### DIFF
--- a/clicommand/pipeline_upload.go
+++ b/clicommand/pipeline_upload.go
@@ -15,6 +15,7 @@ import (
 	"strings"
 	"time"
 
+	"drjosh.dev/zzglob"
 	"github.com/aws/aws-sdk-go-v2/service/kms"
 	"github.com/buildkite/agent/v3/agent"
 	"github.com/buildkite/agent/v3/api"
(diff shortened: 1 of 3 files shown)


clicommand/pipeline_upload.go
```diff
@@ -301,6 +316,24 @@ var PipelineUploadCommand = cli.Command{
 			}
 		}
 
+		if cfg.ApplyIfChanged {
+			func() {
```
wolfeidau: should this be lifted out, rather than inline?

wolfeidau (changes requested):
Just one observation/question.

clicommand/pipeline_upload.go
DrJosh9000: Have done that now.

wolfeidau (approved):
Looks good to me. 🚀

PR #3233, opened 2025-03-12
Fix redaction of multiline secrets printed as single lines

DrJosh9000:
### Description

Broaden the set of strings matched by `'\n'` to include any number of `' '` (not necessarily followed by `'\n'`).

### Context

It's easy to accidentally leak multiline secrets with the shell option `set -x`, or even just `echo ${MY_MULTILINE_SECRET}`.

https://linear.app/buildkite/issue/PS-446

### Changes

* Handle `' '` in the input stream when matching `'\n'` by splitting into two match states at different positions. 
* Change a map `map[string]unit` to a string slice to restore scanning performance

### Testing

### DIFF
--- a/internal/replacer/replacer.go
+++ b/internal/replacer/replacer.go
@@ -4,11 +4,10 @@ package replacer
 import (
 	"fmt"
 	"io"
+	"slices"
 	"strings"
 	"sync"
 )
-
-type unit struct{}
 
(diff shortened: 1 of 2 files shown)


wolfeidau (approved):
👍🏻

jordandcarter (approved):
LGTM

PR #3237, opened 2025-03-13
ANSI parser speedup

DrJosh9000:
### Description

Significant speed savings:

```
❯ benchstat ansi-parser-{maps,arrays}.txt
goos: darwin
goarch: arm64
pkg: github.com/buildkite/agent/v3/process
cpu: Apple M1 Max
              │ ansi-parser-maps.txt │       ansi-parser-arrays.txt        │
              │        sec/op        │   sec/op     vs base                │
ANSIParser-10           2322.3µ ± 0%   565.3µ ± 1%  -75.66% (p=0.000 n=30)
```

### Context

The secret redactor was sped up a bit by switching away from a map, and I remembered this used maps too.

### Changes

* Add a benchmark
* Rename `feed` to `Write` in case we want to e.g. `io.Copy` or `MultiWriter` into it
* Use `[256]*ansiParserState` instead of `map[byte]ansiParserState`

### Testing

### DIFF
--- a/process/ansi.go
+++ b/process/ansi.go
@@ -8,7 +8,7 @@ package process
 // including mention of some of the necessary deviations from the standards
 // (such as allowing some sequences to terminate with BEL instead of ESC '\').
 type ansiParser struct {
-	state ansiParserState
+	state *ansiParserState
 }
 
 // Write passes more bytes through the parser.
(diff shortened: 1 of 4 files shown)


pda (approved):
Nice!

PR #3242, opened 2025-03-19
Redact annotations

DrJosh9000:
### Description

Annotations are as visible as job logs, and can just as easily have secrets leaked into them accidentally, so apply redaction to them.

### Context

https://3.basecamp.com/3453178/buckets/23112918/messages/8443739637

### Changes

* internal/redact: Some helper refactoring 
* clicommand: Pass the annotation body through `redact.String`.

### Testing


CerealBoy (approved):
LGTM! 🚀

PR #3243, opened 2025-03-20
Redact meta-data values and step attribute updates with warnings

DrJosh9000:
### Description

Apply redaction to meta-data and step updates. 
When redaction happens in these cases, log a warning (also for annotations).

### Context

https://3.basecamp.com/3453178/buckets/23112918/messages/8443739637

### Testing

### DIFF
--- a/clicommand/annotate.go
+++ b/clicommand/annotate.go
@@ -160,7 +160,10 @@ func annotate(ctx context.Context, cfg AnnotateConfig, l logger.Logger) error {
 	if err != nil {
 		return err
 	}
-	body = redact.String(body, needles)
+	if redactedBody := redact.String(body, needles); redactedBody != body {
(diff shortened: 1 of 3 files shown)


clicommand/annotate.go
```diff
@@ -160,7 +160,10 @@ func annotate(ctx context.Context, cfg AnnotateConfig, l logger.Logger) error {
 	if err != nil {
 		return err
 	}
-	body = redact.String(body, needles)
+	if redactedBody := redact.String(body, needles); redactedBody != body {
+		l.Warn("Annotation body contained one or more secrets from environment variables that have been redacted. If this is deliberate, pass --redacted-vars='' or a list of patterns that does not match the variable containing the secret")
```
catkins, line 164: 👏 nice and clear

catkins (approved):
👏 Looks good to me. Having the _clear_, and _actionable_ note in the warnings will offset any surprises.

Thanks @DrJosh9000 🚀

PR #3245, opened 2025-03-20
added explanation how job annotations works, has no impact on UI

ivannalisetska:
### Description
Added more explanation how job annotations works with "--job"

### Testing

### DIFF
--- a/clicommand/annotate.go
+++ b/clicommand/annotate.go
@@ -106,7 +106,7 @@ var AnnotateCommand = cli.Command{
 		cli.StringFlag{
 			Name:   "job",
 			Value:  "",
-			Usage:  "Which job should the annotation come from",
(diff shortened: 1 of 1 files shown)


clicommand/annotate.go
```diff
@@ -106,7 +106,7 @@ var AnnotateCommand = cli.Command{
 		cli.StringFlag{
 			Name:   "job",
 			Value:  "",
-			Usage:  "Which job should the annotation come from",
+			Usage:  "Which job should the annotation come from. Effectively, specifying the current job ID has no impact on UI placement — it will execute, but it would not produce any visual changes in the Buildkite interface.",
```
gilesgas, line 109: Just a few points to clarify:

> ... specifying the current job ID has no impact on UI placement.

What do you mean specifically by UI placement? Is this within the annotation on the current build or on the build that triggered this current build?

> it will execute, but it would not produce any visual changes in the Buildkite interface.

What will execute (the current job?) and what exactly are you referring to by 'visual changes in the Buildkite interface'?

PR #3246, opened 2025-03-23
Fix incomplete processing in newly-redacted operations

DrJosh9000:
### Description

Fixes the operations that recently had redaction added (annotate, meta-data set, step update) where the final character(s) would be missing.

The cause was a missing call to `Flush` on the string replacer, combined with one or more secrets that started with the missing character.

### Context

https://linear.app/buildkite/issue/PS-485

### Testing


CerealBoy (approved):

catkins (approved):
👏 thanks @DrJosh9000

PR #3260, opened 2025-04-03
Add otel metrics export to the agent

wolfeidau:
### Description

This is a bit of an experiment looking at what it would look like to build a set of metrics from the spans generated during execution of a job, currently these are written out to the logs of the agent and sent to buildkite.

I need to explore where to write this data, and what options there are to integrate it into otel tracing already present.

### Testing


DrJosh9000:
Closing due to staleness

PR #3261, opened 2025-04-03
Adding a reference  in our docs the limit of an annotation's contexts

lizrabuya:
### Description

We had a customer hitting the limit of the annotation's context. So adding the limit in our docs would help future travellers be aware that such limit exits.

### Testing

### DIFF
--- a/clicommand/annotate.go
+++ b/clicommand/annotate.go
@@ -84,7 +84,7 @@ var AnnotateCommand = cli.Command{
 	Flags: []cli.Flag{
 		cli.StringFlag{
 			Name:   "context",
-			Usage:  "The context of the annotation used to differentiate this annotation from others. The value has a limit of 100 characters",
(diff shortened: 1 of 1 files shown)


clicommand/annotate.go
```diff
@@ -84,7 +84,7 @@ var AnnotateCommand = cli.Command{
 	Flags: []cli.Flag{
 		cli.StringFlag{
 			Name:   "context",
-			Usage:  "The context of the annotation used to differentiate this annotation from others",
+			Usage:  "The context of the annotation used to differentiate this annotation from others. The value has a limit of 100 characters",
```
gilesgas, line 87:
```suggestion
			Usage:  "The context of the annotation used to differentiate this annotation from others. This value has a limit of 100 characters.",
```

gilesgas (approved):
Minor tweak and then you can merge this in!

Thanks very much for this @lizrabuya ! 🎉

DrJosh9000 (approved):

PR #3262, opened 2025-04-04
`AgentWorker` has `noWaitBetweenPingsForTesting` field

pda:
The minimum valid server-specified ping interval is 1 second, which means tests that run the ping loop through multiple pings spend multiple seconds waiting.

This introduces a private `AgentWorker.noWaitBetweenPingsForTesting bool` field that tests within package agent can set, which will skip the delay between pings by immediately returning from the `select { … }` due to a closed channel.

This speeds up the existing agent tests by ~5 seconds when run with single parallelism. The benefits are presumably less when running the entire suite in parallel, but this remains useful when iterating on a single test (milliseconds instead of seconds).

Context: I'm working on [some changes](https://github.com/buildkite/agent/pull/3263) that introduce new tests which use the ping loop.

### DIFF
--- a/agent/agent_worker.go
+++ b/agent/agent_worker.go
@@ -104,6 +104,9 @@ type AgentWorker struct {
 	stateMtx     sync.Mutex
 	state        agentWorkerState
 	currentJobID string
+
+	// force a sub-second ping interval for testing
+	pingIntervalForTesting time.Duration
 }
 
 type agentWorkerState string
@@ -291,6 +294,9 @@ func (a *AgentWorker) runPingLoop(ctx context.Context, idleMonitor *IdleMonitor)
 
 	// Create the ticker
 	pingInterval := time.Second * time.Duration(a.agent.PingInterval)
+	if a.pingIntervalForTesting != 0 {
+		pingInterval = a.pingIntervalForTesting
+	}
 	pingTicker := time.NewTicker(pingInterval)
 	defer pingTicker.Stop()
 
--- a/agent/agent_worker.go
+++ b/agent/agent_worker.go
@@ -105,8 +105,8 @@ type AgentWorker struct {
 	state        agentWorkerState
 	currentJobID string
 
-	// force a sub-second ping interval for testing
-	pingIntervalForTesting time.Duration
+	// disable the delay between pings, to speed up certain testing scenarios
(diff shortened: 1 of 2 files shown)


agent/agent_worker.go
```diff
@@ -291,6 +294,9 @@ func (a *AgentWorker) runPingLoop(ctx context.Context, idleMonitor *IdleMonitor)
 
 	// Create the ticker
 	pingInterval := time.Second * time.Duration(a.agent.PingInterval)
+	if a.pingIntervalForTesting != 0 {
```
DrJosh9000, line 297: I think this is fine, and would be happy to ship it. But I like to come up with suggestions. :smile:

If the only non-zero value the new field will be set to is "negligible amount of time" (1ms), and it is a proxy for 0 wait time, then maybe we should consider a bool (e.g. `noWaitBetweenPings`) and implement an actual 0 wait time. Something like:

```go
// Operations on a nil channel block forever
var testTriggerCh chan struct{}
if a.noWaitBetweenPings { 
    // Receive from a closed channel unblocks immediately, infinitely
    testTriggerCh = make(chan struct{})
    close(testTriggerCh)
}

...

for {
    select {
    case <-testTriggerCh:
        // Only happens if a.noWaitBetweenPings is true, continue below.
    ...
    }
    // and again in the jitter select
```

DrJosh9000 (approved):

agent/agent_worker.go
pda: Ooh interesting.

Yeah I can't immediately see any use-case for values other than basically-zero. I was _slightly vaguely_ worried that _zero wait_ would somehow behave differently to _some wait_, regarding… I dunno race conditions or something. But I'm sure it's fine, especially if it doesn't trigger your spider senses.

What do you think about the `…ForTesting` suffix on private fields that are nonsense/dangerous outside of testing?

DrJosh9000: Sounds reasonable.

pda: Done!
And verified against the existing tests 👍🏼

pda: I like that this is orthogonal to the actual ping delay, interacting only via the `select { … }` which is exactly where such interactions belong 👌🏼

PR #3264, opened 2025-04-07
Refactor fake API server

DrJosh9000:
### Description

Follow through on an earlier ~~threat~~ TODO comment.

### Context

Reminded of it while discussing #3263 

### Changes

- Refactor the `httptest.Server` used in the two recently added agent tests into its own thing, `FakeAPIServer`
- Update the tests to use the new fake

### Testing

### DIFF
--- a/agent/fake_api_server_test.go
+++ b/agent/fake_api_server_test.go
@@ -1,0 +1,372 @@
+package agent
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/buildkite/agent/v3/api"
+	"github.com/google/uuid"
+)
+
+// This file implements a flexible fake testing server for the agent, including
+// registration through to pinging and running jobs, but not including many of
+// the other things the agent REST API does.
+//
+// Why not fake the client? Because there's a lot of value in testing that
+// real requests round-trip through a network connection, even if both endpoints
+// are the same process.
+
+const (
+	JobStateScheduled = "scheduled"
+	JobStateAssigned  = "assigned"
+	JobStateAccepted  = "accepted"
+	JobStateRunning   = "running"
+	JobStateFinished  = "finished"
+)
+
+type FakeJob struct {
+	Job   *api.Job
+	Auth  string
+	State string
+	Logs  strings.Builder
+}
+
+type FakeAgent struct {
+	Assigned   *FakeJob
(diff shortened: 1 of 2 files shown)


pda (approved):
Ooh looks very nice 👌🏼

agent/fake_api_server_test.go
```diff
@@ -0,0 +1,372 @@
...
+	}
+
+	// job is assigned to this agent, accepted, and is now accessible using a
+	// job token.
+	fs.Assign(agent, job)
+	job.State = JobStateAccepted
+
+	out, err := json.Marshal(job.Job)
+	if err != nil {
+		http.Error(rw, encodeMsgf("json.NewEncoder(http.ResponseWriter).Encode(%v) = %v", job.Job, err), http.StatusInternalServerError)
```
pda, line 157: Should these all be more like this?

```suggestion
		http.Error(rw, encodeMsgf("json.Marshal(%v) = %v", job.Job, err), http.StatusInternalServerError)
```

DrJosh9000: Ja, die copypasta ist schlect.

agent/fake_api_server_test.go
```diff
@@ -0,0 +1,372 @@
...
+			ChunksMaxSizeBytes: 1024,
+			Env:                env,
+		},
+		Auth: "Token " + jobToken,
+	}
+	fs.jobs[j.Job.ID] = j
+	return j
+}
+
+func (fs *FakeAPIServer) Assign(agent *FakeAgent, job *FakeJob) {
```
pda, line 111: Is this missing `fs.mu.Lock()` for potentially-concurrent `fs.agentJobs` map access?

DrJosh9000: Yes it is!

pda: Would we perhaps use [`sync/v2`'s `Map`](https://github.com/golang/go/issues/71076) if/when that becomes a thing? Or is this mutex serving other purposes too?

DrJosh9000: We could, which would potentially be a win from the perspectives of finer-grained locking and not having to remember to lock/unlock a mutex. The only downside would be potentially inconsistent views of the data, but this is for testing, so that's probably fine. (It might even be fine without a mutex at all.)

pda: > and not having to remember to lock/unlock a mutex

Yeah that's the bit I like, I don't have a strong instinct for mutexing things 😅

PR #3268, opened 2025-04-08
`api.Client` sends request headers specified by server in register & ping

pda:
Allow Buildkite server to add `Buildkite-*` request headers to an agent's session, and then use those subsequent request headers to e.g. improve routing, performance, and reliability of requests. Specifically, we intend to use these headers to extend the resilience of our sharded architecture.

The request headers would be included in an optional `request_headers` field of /v3/register and /v3/ping responses:

    {
      "…": "…",
      "request_headers": {
        "Buildkite-Hello": "world!"
      }
    }

## Summary of changes:

- The API client has a new internal map of server-specified headers to include in every request (`api.Client.requestHeaders`).
- Only headers beginning with `Buildkite-` are accepted by the agent, so general HTTP behaviour cannot be impacted by this mechanism.
- The presence of `request_headers` in register/ping _replaces_ the client's request headers.
- Thus, the presence of an _empty_ object i.e. `"request_headers": {}` clears `Client.requestHeaders` and stops sending any server-specified headers.
- Whereas the absence of that field, or a value of `null`, is a no-op; the client is not updated and any existing headers continue to be sent. This allows the majority of ping responses to omit the `request_headers` data.

## Header Propagation

To include these server-specified headers in all subsequent Agent API requests, there's a few scenarios:

### Requests from the same `api.Client`

This is straightforward; the server-specified headers from ping & register responses are stored in a private field of the `api.Client` struct, and added to subsequent requests.

### Requests from new/reconfigured `api.Client`

There's a few places where an existing `api.Client` is exchanged for a reconfigured one.
- [`(*api.Client).FromAgentRegisterResponse()`](https://github.com/buildkite/agent/blob/88969df5795566d13d2caeb5e1bc17078698a754/api/client.go#L117) (after register, before ping/heartbeat/etc)
- [`(*api.Client).FromPing()`](https://github.com/buildkite/agent/blob/88969df5795566d13d2caeb5e1bc17078698a754/api/client.go#L129) (after a ping _only_ if the endpoint has changed)
- [`agent.NewJobRunner()`](https://github.com/buildkite/agent/blob/88969df5795566d13d2caeb5e1bc17078698a754/agent/job_runner.go#L202) (if the accept response has a _job token_)

These scenarios previously used `api.NewClient(logger, modifiedCopyOfConfig)` which lost any internal state from the client. I've instead introduced `(*api.Client).New(config)` which passes non-config state (currently: `logger` & `requestHeaders`) from the old to the new client.

### Requests from `buildkite-agent …` calls inside jobs

Commands like `buildkite-agent annotate …` and `buildkite-agent artifact …` are generally run from scripts running inside the command of a job running in the “bootstrap” (`job.Executor`), so they're several layers of process removed from the process with the `api.Client` that has the request headers.

As with all other information passed to the bootstrap / job command, this PR maps the necessary request headers to environment variables with `BUILDKITE_REQUEST_HEADER_` prefix, e.g. `Buildkite-Hello: world` becomes `BUILDKITE_REQUEST_HEADER_BUILDKITE_HELLO=world`.

Inside the `buildkite-agent …` sub-process within the job command, the `api.NewClient()` method discovers these environment variables and maps them back to its internal store of request headers (but only for headers beginning with `Buildkite-`).

Related:
- Closes #3263; this is an alternative (and more complete) implementation approach.
- extracted https://github.com/buildkite/agent/pull/3269 from this PR, which will make this PR a bit smaller and more focused.

### DIFF
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -491,7 +491,7 @@ func (r *JobRunner) createEnvironment(ctx context.Context) ([]string, error) {
 
 	// ... including any server-specified request headers, so that sub-processes such as
 	// buildkite-agent annotate etc can respect them.
-	for header, values := range r.apiClient.GetServerSpecifiedRequestHeaders() {
+	for header, values := range r.apiClient.ServerSpecifiedRequestHeaders() {
 		k := fmt.Sprintf(
 			"BUILDKITE_REQUEST_HEADER_%s",
 			strings.ToUpper(strings.ReplaceAll(header, "-", "_")),
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -178,10 +178,19 @@ var _ jobRunner = (*JobRunner)(nil)
 
 // Initializes the job runner
 func NewJobRunner(ctx context.Context, l logger.Logger, apiClient APIClient, conf JobRunnerConfig) (jobRunner, error) {
+	// If the accept response has a token attached, we should use that instead of the Agent Access Token that
+	// our current apiClient is using
+	if conf.Job.Token != "" {
+		clientConf := apiClient.Config()
+		clientConf.Token = conf.Job.Token
+		apiClient = apiClient.New(clientConf)
+	}
+
 	r := &JobRunner{
 		agentLogger: l,
 		conf:        conf,
 		apiClient:   apiClient,
+		client:      &core.Client{APIClient: apiClient, Logger: l},
 	}
 
 	var err error
@@ -193,15 +202,6 @@ func NewJobRunner(ctx context.Context, l logger.Logger, apiClient APIClient, con
 	if conf.JobStatusInterval == 0 {
 		conf.JobStatusInterval = 1 * time.Second
 	}
-
-	// If the accept response has a token attached, we should use that instead of the Agent Access Token that
-	// our current apiClient is using
-	if r.conf.Job.Token != "" {
-		clientConf := r.apiClient.Config()
-		clientConf.Token = r.conf.Job.Token
-		r.apiClient = api.NewClient(r.agentLogger, clientConf)
-	}
-	r.client = &core.Client{APIClient: r.apiClient, Logger: l}
 
 	// Create our header times struct
 	r.headerTimesStreamer = newHeaderTimesStreamer(r.agentLogger, r.onUploadHeaderTime)
@@ -488,6 +488,18 @@ func (r *JobRunner) createEnvironment(ctx context.Context) ([]string, error) {
 	env["BUILDKITE_AGENT_ENDPOINT"] = apiConfig.Endpoint
 	env["BUILDKITE_AGENT_ACCESS_TOKEN"] = apiConfig.Token
 	env["BUILDKITE_NO_HTTP2"] = fmt.Sprint(apiConfig.DisableHTTP2)
+
+	// ... including any server-specified request headers, so that sub-processes such as
+	// buildkite-agent annotate etc can respect them.
+	for header, values := range r.apiClient.GetServerSpecifiedRequestHeaders() {
+		k := fmt.Sprintf(
+			"BUILDKITE_REQUEST_HEADER_%s",
+			strings.ToUpper(strings.ReplaceAll(header, "-", "_")),
+		)
+		for _, v := range values {
+			env[k] = v
+		}
+	}
 
 	// Add agent environment variables
 	env["BUILDKITE_AGENT_DEBUG"] = fmt.Sprint(r.conf.Debug)
--- a/api/agents.go
+++ b/api/agents.go
@@ -47,7 +47,7 @@ func (c *Client) Register(ctx context.Context, regReq *AgentRegisterRequest) (*A
 	}
 
 	// If Buildkite told us to use Buildkite-* request headers, store those
-	c.setRequestHeadersFromMap(a.RequestHeaders)
+	c.setRequestHeaders(a.RequestHeaders)
 
 	return a, resp, err
 }
--- a/api/agents.go
+++ b/api/agents.go
@@ -21,14 +21,15 @@ type AgentRegisterRequest struct {
 
 // AgentRegisterResponse is the response from the Buildkite Agent API
 type AgentRegisterResponse struct {
-	UUID              string   `json:"id"`
-	Name              string   `json:"name"`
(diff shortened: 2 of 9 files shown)


agent/job_runner.go
```diff
@@ -489,6 +489,18 @@ func (r *JobRunner) createEnvironment(ctx context.Context) ([]string, error) {
...
+	// buildkite-agent annotate etc can respect them.
+	for header, values := range r.apiClient.GetServerSpecifiedRequestHeaders() {
+		k := fmt.Sprintf(
+			"BUILDKITE_REQUEST_HEADER_%s",
+			strings.ToUpper(strings.ReplaceAll(header, "-", "_")),
+		)
+		for _, v := range values {
+			env[k] = v
+		}
+	}
```
pda, line 502: Here's where we map request headers to job environment variables, so that sub-processes like `buildkite-agent annotate` etc can pick them up.

agent/job_runner.go
```diff
@@ -178,10 +178,19 @@ var _ jobRunner = (*JobRunner)(nil)
 
 // Initializes the job runner
 func NewJobRunner(ctx context.Context, l logger.Logger, apiClient APIClient, conf JobRunnerConfig) (jobRunner, error) {
+	// If the accept response has a token attached, we should use that instead of the Agent Access Token that
+	// our current apiClient is using
+	if conf.Job.Token != "" {
+		clientConf := apiClient.Config()
+		clientConf.Token = conf.Job.Token
+		apiClient = apiClient.New(clientConf)
```
pda, line 186: Here's where we ensure request headers are propagated to the new client when switching from agent session token to agent job token. The previous `api.NewClient(…)` is on old line 202; I did a bit of refactoring so that we're just mutating/replacing _local vars_ before constructing any structs.

api/agents.go
```diff
@@ -45,6 +46,9 @@ func (c *Client) Register(ctx context.Context, regReq *AgentRegisterRequest) (*A
 		return nil, resp, err
 	}
 
+	// If Buildkite told us to use Buildkite-* request headers, store those
+	c.setRequestHeadersFromMap(a.RequestHeaders)
```
pda, line 50: Grabbing server-specified request headers from `AgentRegisterResponse`

api/client.go
```diff
@@ -68,6 +69,9 @@ type Client struct {
 
 	// The logger used
 	logger logger.Logger
+
+	// server-specified HTTP request headers to include in all requests
+	requestHeaders http.Header
```
pda, line 74: Here's the internal state where the server-specified headers live inside the API client.

api/client.go
```diff
@@ -99,17 +103,51 @@ func NewClient(l logger.Logger, conf Config) *Client {
 	}
 
 	return &Client{
-		logger: l,
-		client: agenthttp.NewClient(clientOptions...),
-		conf:   conf,
+		logger:         l,
+		client:         agenthttp.NewClient(clientOptions...),
+		conf:           conf,
+		requestHeaders: requestHeadersFromEnv(os.Environ()),
```
pda, line 109: For sub-processes running inside job commands like `buildkite-agent annotate`, this is where we grab `BUILDKITE_REQUEST_HEADER_BUILDKITE_…` from environment into `Buildkite-*` request headers.

api/client.go
```diff
@@ -180,6 +232,13 @@ func (c *Client) newRequest(
 		}
 	}
 
+	// add any request headers specified by the server during register/ping
+	for k, values := range c.requestHeaders {
+		for _, v := range values {
+			req.Header.Add(k, v)
+		}
+	}
```
pda, line 240: Adding the server-specified headers to outbound requests.

api/pings.go
```diff
@@ -23,5 +24,8 @@ func (c *Client) Ping(ctx context.Context) (*Ping, *Response, error) {
 		return nil, resp, err
 	}
 
+	// If Buildkite told us to use Buildkite-* request headers, store those
+	c.setRequestHeadersFromMap(ping.RequestHeaders)
```
pda, line 28: Grabbing server-specified request headers from Ping responses.

pda (commented):
A quick tour of key changes…

api/client.go
```diff
@@ -123,7 +161,21 @@ func (c *Client) FromAgentRegisterResponse(reg *AgentRegisterResponse) *Client {
 		conf.Endpoint = reg.Endpoint
 	}
 
-	return NewClient(c.logger, conf)
+	return c.New(conf)
+}
+
+func (c *Client) setRequestHeadersFromMap(headers map[string]string) {
```
DrJosh9000, line 167: "FromMap" feels redundant with the arg type. If there were several ways to set the request headers with different types, "FromMap" would be reasonable since there's no overloading.

DrJosh9000 (commented):
This is great!

api/client.go
pda: Yeah I was reserving the shorter name for `setRequestHeaders(http.Header)` but I didn't end up needing to implement that. If it were public, I'd be inclined to keep that reservation, but being private I'd be happy to shorten the name now, and change it in future if needed.

(Writing some Java recently, I have quite enjoyed its signature overloading e.g. `setRequestHeaders(map[string]string)` and `setRequestHeaders(http.Header)` being two different methods)

api/client.go
```diff
@@ -99,17 +103,51 @@ func NewClient(l logger.Logger, conf Config) *Client {
...
+func requestHeadersFromEnv(environ []string) http.Header {
+	prefix := "BUILDKITE_REQUEST_HEADER_"
+	headers := make(http.Header)
+	for _, line := range environ {
+		parts := strings.SplitN(line, "=", 2)
+		if len(parts) != 2 || !strings.HasPrefix(parts[0], prefix) {
+			// not a BUILDKITE_REQUEST_HEADER_... environment variable
+			continue
+		}
+		key := http.CanonicalHeaderKey(strings.ReplaceAll(parts[0][len(prefix):], "_", "-"))
```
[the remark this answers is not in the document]
pda: Nice 👍🏼

I also just noticed that the unit test for this method has gone missing via some git mistake or other.
I'll see if I can resurrect it or rewrite it.

DrJosh9000 (approved):
LGTM ✅ :shipit:

PR #3269, opened 2025-04-08
agent_worker_test: tests for endpoint switching during register/ping

pda:
This functionality has existed for a long time, but hasn't had test coverage.

Uses the nice new `FakeAPIServer` from https://github.com/buildkite/agent/pull/3264

This also introduces `(api.Config).Timeout time.Duration` so that the timeout on a known-broken endpoint in tests can be reduced to milliseconds instead of the default 60 seconds.

These tests are extracted and adapted from ~#3263~ #3268, so that/those PRs can be rebased after this is merged to shrink them to their actual intent.

### DIFF
--- a/agent/agent_worker_test.go
+++ b/agent/agent_worker_test.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"net"
 	"net/http"
 	"net/http/httptest"
 	"os"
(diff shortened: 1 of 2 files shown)


agent/agent_worker_test.go
```diff
@@ -399,3 +401,240 @@ func TestAgentWorker_DisconnectAfterJob_Start_Pause_Unpause(t *testing.T) {
...
+	}
+
+	// Create a listener without an HTTP server; hitting this would cause a network error/timeout.
+	// This is a bit neater than using a random host/port and hoping nothing is listening on it.
+	registrationServer, err := net.Listen("tcp", ":0")
+	if err != nil {
+		t.Fatalf("creating broken endpoint listener: %v", err)
+	}
+	defer registrationServer.Close()
+	registrationEndpoint := fmt.Sprintf("http://%s/", registrationServer.Addr().String())
```
DrJosh9000, line 436: `fmt.Sprintf` understands the `String` method, and should call it automatically:

```suggestion
	registrationEndpoint := fmt.Sprintf("http://%s/", registrationServer.Addr())
```

DrJosh9000 (approved):
Nice!

PR #3271, opened 2025-04-09
`FakeAPIServer`'s `PingHandler` is passed the `*http.Request`

pda:
This allows `PingHandler`s to make decisions/assertions on HTTP headers etc, which I need in #3268.

Related:
- #3268
- #3264

### DIFF
--- a/agent/agent_worker_test.go
+++ b/agent/agent_worker_test.go
@@ -255,7 +255,7 @@ func TestAgentWorker_Start_AcquireJob_Pause_Unpause(t *testing.T) {
 	// Pre-register the agent.
 	const agentSessionToken = "alpacas"
 	agent := server.AddAgent(agentSessionToken)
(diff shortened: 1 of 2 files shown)


agent/agent_worker_test.go
```diff
@@ -255,7 +255,7 @@ func TestAgentWorker_Start_AcquireJob_Pause_Unpause(t *testing.T) {
 	// Pre-register the agent.
 	const agentSessionToken = "alpacas"
 	agent := server.AddAgent(agentSessionToken)
-	agent.PingHandler = func() (api.Ping, error) {
+	agent.PingHandler = func(req *http.Request) (api.Ping, error) {
```
DrJosh9000, line 258: For implementations that don't look at the arg, the name can be omitted

```suggestion
	agent.PingHandler = func(*http.Request) (api.Ping, error) {
```

DrJosh9000 (approved):

PR #3272, opened 2025-04-09
`FakeAPIServer` handles agent registration: `AddRegistration(tok, resp)`

pda:
I intend to use this in #3268 to have the `api.Client` perform a registration request, with server-specified `request_headers` in the response, and ensure those headers are passed in subsequent requests.

Related:
- #3268
- #3264


DrJosh9000 (approved):
LGTM

PR #3274, opened 2025-04-10
chore: Use golangci-lint for code checks

DrJosh9000:
### Description

Use a more powerful lint tool to find issues.
Since it finds numerous violations in the existing codebase, ignore its exit code (for now).

### Context

Noticed during #3273 

### Testing

### DIFF
--- a/.buildkite/steps/check-code-committed.sh
+++ b/.buildkite/steps/check-code-committed.sh
@@ -1,6 +1,6 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
 
-set -Eeufo pipefail
+set -euf
 
 echo --- :go: Checking go mod tidyness
 go mod tidy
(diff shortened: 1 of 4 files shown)


.buildkite/steps/check-code-committed.sh
```diff
@@ -34,4 +34,9 @@ if ! git diff --no-ext-diff --exit-code; then
   exit 1
 fi
 
+# While we're cleaning up things found by golangci-lint, don't fail if it finds
+# things.
+echo +++ :go: Running golangci-lint...
+golangci-lint run || true
```
catkins, line 40: ooh nice, lets see how noisy it is with the default linters!

catkins: Okay, thats good - gives us some stuff to work through. Having a scan of the erros, a lot of the staticcheckones are 🤷 and we can probably disable the rules. The errcheck ones are useful though!

```
106 issues:
--
  | * errcheck: 50
  | * govet: 4
  | * ineffassign: 1
  | * staticcheck: 39
  | * unused: 12

```

catkins (approved):
I was going to suggest plonking the output into an annotation if we're not failing on it, but we can do that later!

Thanks for adding this @DrJosh9000 💚

PR #3276, opened 2025-04-10
chore: flag uniformity through embedding

DrJosh9000:
### Description

Replace repeated struct fields and flags with embedded structs and helper funcs.

### Context

Noticed by @catkins in #3273 

### Changes

- Introduce `GlobalConfig` and `APIConfig` structs for embedding. Nearly everything uses `GlobalConfig`, most things also use `APIConfig`. 
- Introduce func `apiFlags` for defining the API flags (`globalFlags` already exists). Then combine with per-command flags with `slices.Concat`.
- Update everything to use.

### Testing


clicommand/agent_pause.go
```diff
@@ -29,28 +29,18 @@ Example:
...
-	Endpoint         string `cli:"endpoint" validate:"required"`
-	NoHTTP2          bool   `cli:"no-http2"`
 }
 
 var AgentPauseCommand = cli.Command{
 	Name:        "pause",
 	Usage:       "Pause the agent",
 	Description: pauseDescription,
-	Flags: []cli.Flag{
+	Flags: append(globalAndAPIFlags(),
```
catkins, line 43: This is a bit arbitrary and what you've done here works too, but another approach here could be to have the `globalAndAPIFlags()` function accept the flags as a variadic function?

```go
func withGlobalAndAPIFlags(additionalFlags ...[]cli.Flag) []cli.Flag {
  return append(globalAndAPIFlags(), additionalFlags)
}

// ... 
var AgentPauseCommand = cli.Command{
	Name:        "pause",
	Usage:       "Pause the agent",
	Description: pauseDescription,
	Flags: withGlobalAndAPIFlags(
    cli.StringFlag{
      // ...
    },
    cli.StringFlag{
      // ...
    },
    // ... etc
  ),

```

cliconfig/loader.go
```diff
@@ -77,7 +77,7 @@ func (l *Loader) Load() (warnings []string, err error) {
 	// Now it's onto actually setting the fields. We start by getting all
 	// the fields from the configuration interface
 	var fields []string
-	fields, _ = reflections.Fields(l.Config)
+	fields, _ = reflections.FieldsDeep(l.Config)
```
catkins, line 80: 👏

catkins (approved):
:+1: nice work as always @DrJosh9000 

I left one 💭 for consideration, but happy for you to go either way with it!

clicommand/agent_pause.go
DrJosh9000: What I was going for was expressing a command's flags as a composition of multiple groups of flags and its own specific flags, but the multiple `append`s was getting a bit ugly:

```go
append(append(globalFlags(), apiFlags()...), /* custom flags */ )
```

hence `globalAndAPIFlags`. A small helper function could replace `append` though, at the cost of wrapping the command-specific flags in a `[]cli.Flag{...}`, which doesn't seem all that bad actually.

catkins: Nice. I like what you did with the `slices.Concat(globalFlags(), apiFlags(), []cli.Flag{ ...` 👏

PR #3282, opened 2025-04-16
Ignore the ping interval if agent will disconnect after job

patrobinson:
### Description

Since a change in 3.95.0 (#3238) to ping after a job is finished we've had reports that jobs are assigned to agents that are about to be disconnected. While this could happen before this change, we believe waiting for the ping interval has made it more frequent and it's likely to happen multiple times for the same job.

This change ensures we don't wait up to 10s before disconnecting, making it as likely as before

### Changes

Reset the ping interval if disconnectAfterJob is set.

### Testing

### DIFF
--- a/agent/agent_worker.go
+++ b/agent/agent_worker.go
@@ -436,6 +436,8 @@ func (a *AgentWorker) runPingLoop(ctx context.Context, idleMonitor *IdleMonitor)
 		ranJob = true
 		if a.agentConfiguration.DisconnectAfterJob {
 			// Unless paused, this agent disconnects after the next ping.
+			// Do the ping immediately so we reduce the chances our agent is assigned a job
+			pingTicker.Reset(pingInterval)
 			continue
 		}
 		lastActionTime = time.Now()


agent/agent_worker.go
```diff
@@ -436,6 +436,8 @@ func (a *AgentWorker) runPingLoop(ctx context.Context, idleMonitor *IdleMonitor)
 		ranJob = true
 		if a.agentConfiguration.DisconnectAfterJob {
 			// Unless paused, this agent disconnects after the next ping.
+			// Do the ping immediately so we reduce the chances our agent is assigned a job
+			pingTicker.Reset(pingInterval)
```
patrobinson, line 440: This same code is called immediately below, I thought about wrapping the `lastActionTime` in a `if` block instead, but it felt like that could cause issues if someone tries to add code that shouldn't happen here? 🤷

DrJosh9000 (approved):
LGTM, worth a shot

agent/agent_worker.go
DrJosh9000: I just thought of something that affects this (and the below) - we probably want to also bypass the jitter (select on line 344)

patrobinson:
Hmm, this won't completely eliminate the interval, since we'll still have the jitter? It should at least improve the situation though?

agent/agent_worker.go
patrobinson: yeah I realised this too, not sure of a good way to do that?

DrJosh9000: Maybe:

1. above the loop `bypassJitter := make(chan struct{}, 1)`
2. in the jitter select, add a `case <-bypassJitter:`
3. when resetting the interval:

```go 
// Send on bypassJitter unless it's already got a value
select {
case bypassJitter <- struct{}{}:
default:
}
```

PR #3285, opened 2025-04-17
Fix unusable `BUILDKITE_AGENT_TAGS_FROM_EC2_TAGS` env var

shanesmith:
The `BUILDKITE_AGENT_TAGS_FROM_EC2_TAGS` env var is assigned to set both `tags-from-ec2-tags` and its deprecated duplicate `meta-data-ec2-tags`. The agent [errors][1] when both the deprecated and new version of an option are set, which makes `BUILDKITE_AGENT_TAGS_FROM_EC2_TAGS` unusable.

Following the naming of other deprecated `meta-data-*` it seems `BUILDKITE_AGENT_META_DATA_EC2_TAGS` was meant to be used for `meta-data-ec2-tags`.

[1]: https://github.com/buildkite/agent/blob/fd239e03db6abffdb562ef0963459ab90203b763/cliconfig/loader.go#L120

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -744,7 +744,7 @@ var AgentStartCommand = cli.Command{
 		cli.BoolFlag{
 			Name:   "meta-data-ec2-tags",
 			Hidden: true,
-			EnvVar: "BUILDKITE_AGENT_TAGS_FROM_EC2_TAGS",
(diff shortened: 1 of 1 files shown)


DrJosh9000 (approved):
Good find, easy to replicate, and mostly: thanks for sending in a fix! LGTM 🌈✅

PR #3289, opened 2025-04-24
Ensure git submodules aren't interacted with if no-submodules is configured

### LINKED ISSUE #3286: Submodules not disabled for git mirrors
Hi, I think I've hit an edge case with git mirrors and submodules disabled with agent-stack-k8s.

It seems that `git submodule foreach --recursive "git clean -ffxdq"` gets run during git mirror updates, even when we have disabled submodules (using the agent-stack-k8s noSubmodules option which just sets the corresponding env var under the hood).

This doesn't seem to cause an issue normally, I guess because the submodules weren't initialized, but we recently had a change committed by accident which added an unused submodule without a corresponding `.gitmodules` entry. This caused the above command to complain `fatal: No url found for submodule path 'xyz' in .gitmodules` which completely broke all our builds once the commit got to main :(

CerealBoy:
### Description

When git submodules have been deactivated, do not run any submodule related commands even if submodules are potentially detected. There's a specific choice here for them to not be used, so even for cleaning operations this should be respected.

### Context

Fixes #3286 

### Changes

Validate that `--no-submodules` is not active when cleaning the checkout path, other locations respect this flag however this did location did not.

### Testing


internal/job/checkout.go
```diff
@@ -549,9 +549,18 @@ func (e *Executor) defaultCheckoutPhase(ctx context.Context) error {
 
 	// Git clean prior to checkout, we do this even if submodules have been
 	// disabled to ensure previous submodules are cleaned up
+	gitSubmodules := false
```
patrobinson, line 552: The comment above this is no longer true. I can see it was added in https://github.com/buildkite/agent/pull/1054 but there's no more information about why it was added.

internal/job/checkout.go
```diff
@@ -549,9 +549,18 @@ func (e *Executor) defaultCheckoutPhase(ctx context.Context) error {
...
 	if hasGitSubmodules(e.shell) {
-		if err := gitCleanSubmodules(ctx, e.shell, e.GitCleanFlags); err != nil {
-			return fmt.Errorf("cleaning git submodules: %w", err)
+		// Only clean if submodules have not been explicitly deactivated
```
patrobinson, line 554:
```suggestion
		// Skip cleaning up if submodules have been deactivated
```

The double negative always turns my brain inside out

patrobinson (commented):
Some minor feedback, I think this is good to go otherwise

internal/job/integration/checkout_git_mirrors_integration_test.go
```diff
@@ -266,7 +266,6 @@ func TestCheckingOutLocalGitProjectWithSubmodulesDisabled_WithGitMirrors(t *test
 		{"clone", "--mirror", "-v", "--", tester.Repo.Path, matchSubDir(tester.GitMirrorsDir)},
 		{"clone", "-v", "--reference", matchSubDir(tester.GitMirrorsDir), "--", tester.Repo.Path, "."},
 		{"clean", "-fdq"},
-		{"submodule", "foreach", "--recursive", "git clean -fdq"},
```
patrobinson, line 269: Same comment as the other spec, should we just remove it?

internal/job/integration/checkout_integration_test.go
```diff
@@ -216,7 +216,6 @@ func TestCheckingOutLocalGitProjectWithSubmodulesDisabled(t *testing.T) {
 	git.ExpectAll([][]any{
 		{"clone", "-v", "--", tester.Repo.Path, "."},
 		{"clean", "-fdq"},
-		{"submodule", "foreach", "--recursive", "git clean -fdq"},
```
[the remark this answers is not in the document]
CerealBoy: If anything I was interpreting this test as being incorrect from when it was added, there shouldn't be any submodule interaction if `--no-submodules` was specified right?

Happy to remove if the other tests continue covering the various scenarios we're interested in here.

internal/job/checkout.go
CerealBoy: 🤦🏼 I just read the comment and realised why this happened in the first place now.

Now I'm wondering what circumstances lead to this being the case, and how we could possibly cater for both situations.

patrobinson (approved):
I'm happy with this, but it'd be good to get @DrJosh9000's feedback. I can't think of any negative consequences to not cleaning up old submodules but 🤷

internal/job/checkout.go
DrJosh9000: AIUI, unless it's a clean checkout (directory doesn't exist, it was empty, or the agent is configured to always do clean checkouts), the agent is running the checkout against a directory containing code that was checked out in a previous job. So it might contain submodules, that have since been removed in whatever ref is being checked out now.

So I wonder if we can harden this in some other way? Such as making `hasGitSubmodules` more robust, or skipping submodule cleaning if the checkout itself is clean?

CerealBoy:
There's some additional context here that we've been missing, and chatting through it a bit more this is not the best approach to take. Closing this PR for the time being.

Issue #3254, opened 2025-03-28
`refs/pull/%s/head` might not always be there, but if it is missing, buildkite fails the checkout


romanoid-openai:
Based on our recent observations, github creates `refs/pull/%s/head` asynchronously and sometimes these creations queue up enough that the ref will not be there by the time buildkite receives the webhook and proceeds. 

Unfortunately if the provider is github and we're building the PR, buildkite agent completely fails the job if it is unable to fetch the `refs/pull/%s/head` even the refspec is not even guaranteed to match the commit we're building (I can update between the time original webhook is issued and when the job runs, for example when retrying old job, or even when developer pushes new commit and the build is not configured to cancel) 

The main fetch:
https://github.com/buildkite/agent/blob/e55e764ab2ada4b730faa8acd68defc1f3044253/internal/job/checkout.go#L578

Mirror update logic: 
https://github.com/buildkite/agent/blob/e55e764ab2ada4b730faa8acd68defc1f3044253/internal/job/checkout.go#L386

Proposed solution: allow this fetch to fail and proceed as long as we still can fetch the specific refspec here: 
https://github.com/buildkite/agent/blob/e55e764ab2ada4b730faa8acd68defc1f3044253/internal/job/checkout.go#L580

Note that mirror logic does not mirror the same dual-fetch, but it also can fail the build if the `refs/pull/%s/head` is not there yet, so it needs to be expecting this corner case too.

PR #3293, opened 2025-04-30
Gracefully Handle Missing GitHub PR `refs/pull/%s/head` in Checkout

123sarahj123:
### Description

GitHub creates `refs/pull/%s/head` asynchronously, and we’ve seen cases where this ref isn't available yet when the Buildkite agent begins the checkout process. Currently, this causes the entire job to fail, even though the specific commit we want to build is still fetchable.

This PR updates the agent’s checkout logic to:
- Attempt to fetch refs/pull/%s/head (for GitHub PRs) as usual
- Log a warning instead of failing the job if the pull ref is missing
- Proceed to fetch the specific commit SHA via gitFetchCommitWithFallback(...)

This ensures that builds can proceed as long as the required commit exists, even if the PR ref is temporarily unavailable, which will improve reliability/flakey failures for Github PRs

This change also applies the same behavior to the mirror update logic in updateGitMirror. Previously, missing GitHub PR refs would cause builds to fail during mirrored checkouts. The mirror fetch will now continue with a warning if the ref is unavailable.

### Context

Issue raised:
https://github.com/buildkite/agent/issues/3254

### Changes

- tolerate missing refs/pull/%s/head during checkout
- extend the same handling to mirrored GitHub checkouts (#updateGitMirror)

### Testing
  - [x] `go test -v github.com/buildkite/agent/v3/internal/job/integration`

### DIFF
--- a/internal/job/checkout.go
+++ b/internal/job/checkout.go
@@ -572,6 +572,23 @@ func (e *Executor) defaultCheckoutPhase(ctx context.Context) error {
 			return fmt.Errorf("fetching branch %q: %w", e.Branch, err)
 		}
 
+	case e.PullRequest != "false" && strings.Contains(e.PipelineProvider, "github"):
+		e.shell.Commentf("Fetch and checkout pull request head from GitHub")
+		refspec := fmt.Sprintf("refs/pull/%s/head", e.PullRequest)
+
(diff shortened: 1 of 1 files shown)


123sarahj123:
Used wrong (stale) repo. 

See https://github.com/buildkite/agent/pull/3294 instead

PR #3294, opened 2025-04-30
Gracefully Handle Missing GitHub PR refs/pull/%s/head in Checkout

123sarahj123:
### Description

GitHub creates `refs/pull/%s/head` asynchronously, and we’ve seen cases where this ref isn't available yet when the Buildkite agent begins the checkout process. Currently, this causes the entire job to fail, even though the specific commit we want to build is still fetchable.

This PR updates the agent’s checkout logic to:
- Attempt to fetch refs/pull/%s/head (for GitHub PRs) as usual
- Log a warning instead of failing the job if the pull ref is missing
- Proceed to fetch the specific commit SHA via gitFetchCommitWithFallback(...)

This ensures that builds can proceed as long as the required commit exists, even if the PR ref is temporarily unavailable, which will improve reliability/flakey failures for Github PRs

This change also applies the same behavior to the mirror update logic in updateGitMirror. Previously, missing GitHub PR refs would cause builds to fail during mirrored checkouts. The mirror fetch will now continue with a warning if the ref is unavailable.

### Context

Issue raised:
https://github.com/buildkite/agent/issues/3254

### Changes

- tolerate missing refs/pull/%s/head during checkout
- extend the same handling to mirrored GitHub checkouts (#updateGitMirror)

### Testing
  - [x] `go test -v github.com/buildkite/agent/v3/internal/job/integration`


DrJosh9000 (commented):
The integration tests are failing because they are mock-based (expect certain number of calls)

Since the `git fetch`es are now retried multiple times, it should be a matter of duplicating the expected calls enough times.

internal/job/git.go
```diff
@@ -158,32 +164,40 @@ func gitFetch(
...
-			return &gitError{error: err, Type: gitErrorFetchBadReference}
-		}
-
-		// 128 is extremely broad, but it seems permissions errors, network unreachable errors etc,
-		// don't result in it
-		if exitErr := new(exec.ExitError); errors.As(err, &exitErr) && exitErr.ExitCode() == 128 {
-			return &gitError{error: err, Type: gitErrorFetchRetryClean}
+	return roko.NewRetrier(
+		roko.WithStrategy(roko.ExponentialSubsecond(1*time.Second)),
+		roko.WithMaxAttempts(10), // 10 attempts will take  minutes
```
123sarahj123, line 169: todo update comment

internal/job/git.go
```diff
@@ -158,32 +164,40 @@ func gitFetch(
...
-		}
-
-		// "fatal: [Cc]ouldn't find remote ref" can happen when just the short commit hash is given.
-		if smelt[badReference] || smelt[badReferencePreGit221] {
-			return &gitError{error: err, Type: gitErrorFetchBadReference}
+	return roko.NewRetrier(
+		roko.WithStrategy(roko.ExponentialSubsecond(1*time.Second)),
+		roko.WithMaxAttempts(3), // 10 attempts will take <> minutes
+		roko.WithJitter(),
+	).DoWithContext(ctx, func(retrier *roko.Retrier) error {
```
zhming0: It's interesting to think about the implication of retrying here. 
There are primary a few cases for retry: 

- Async ref creation (which I TIL 😮 ), in this case, refetching the same refspec make sense. 
- Short/ambiguous commit SHA, in this case refetching the same refspec wouldn't help. but I think it's edge case, so it shouldn't hurt. 
- genuinely wrong refspec, which is also edge case, retrying won't help but won't hurt either.

zhming0: Another considertation is that the `gitFetch` method is used in those existing retry mechanism too. Adding a retry on `gitFetch` level will cause the existing retry mechanism to retry for 6 times. It may not be an issue, but combining with the fact this is covering limited error scenario, I wonder if making this retry an opt-in thing will be safer? 


Also I think it might be necessary to document why we added a retry here for the next traveller. 🙏🏿

zhming0 (commented):
Overall I think the retry mechanism on `gitFetch` is a net positive change, but my suggestion is making it opt-in.

internal/job/git.go
zhming0: in case you found adding a optional flag in golang difficult, we could offer 2 methods instead: `gitFetchWithRetry` and `gitFetch`.

internal/job/git.go
```diff
@@ -158,31 +165,61 @@ func gitFetch(
...
+	// remotely (e.g. async ref creation), and retrying `git fetch` can resolve it.
+	// This is *not* always desirable—some call sites have their own retry mechanisms,
+	// so the retry must be opt-in to avoid nested retries and increased test flakiness.
+	if withRetry {
+		return roko.NewRetrier(
+			roko.WithStrategy(roko.ExponentialSubsecond(1*time.Second)),
+			roko.WithMaxAttempts(10), // 10 attempts will take ~8.5 minutes total (first attempt has no wait, backoff adds up to ~511s)
+
+			roko.WithJitter(),
+		).DoWithContext(ctx, func(retrier *roko.Retrier) error {
```
DrJosh9000:
```suggestion
	var retrier *roko.Retrier
	if withRetry {
		retrier = roko.NewRetrier(
			roko.WithStrategy(roko.ExponentialSubsecond(1*time.Second)),
			roko.WithMaxAttempts(10), // 10 attempts will take ~8.5 minutes total (first attempt has no wait, backoff adds up to ~511s)

			roko.WithJitter(),
		)
	} else {
		retrier = roko.NewRetrier(roko.WithMaxAttempts(1))
	}
	return retrier.DoWithContext(ctx, func(*roko.Retrier) error {
```

internal/job/git.go
```diff
@@ -129,18 +131,23 @@ func gitFetch(
 	ctx context.Context,
 	sh *shell.Shell,
 	gitFetchFlags, repository string,
```
DrJosh9000:
```suggestion
	gitFlags, gitFetchFlags, repository string,
```

internal/job/git.go
```diff
@@ -129,18 +131,23 @@ func gitFetch(
...
+	withRetry bool,
 	refSpec ...string,
 ) error {
-	individualFetchFlags, err := shellwords.Split(gitFetchFlags)
-	if err != nil {
-		return err
+	// Build the initial part of the command: git [flags]
+	commandArgs := []string{"fetch"}
+	if gitFetchFlags != "" {
+		parts, err := shellwords.Split(gitFetchFlags)
```
DrJosh9000:
```suggestion
	// Build the initial part of the command: git [gitFlags] fetch [gitFetchFlags]
	if gitFlags != "" {
		parts, err := shellwords.Split(gitFlags)
		if err != nil {
			return fmt.Errorf("failed to parse gitFlags: %w", err)
		}
		commandArgs = append(commandArgs, parts...)
	}
	commandArgs = append(commandArgs, "fetch")
	if gitFetchFlags != "" {
		parts, err := shellwords.Split(gitFetchFlags)
```

DrJosh9000 (approved):
LGTM!

PR #3295, opened 2025-05-01
TE-3708-follow-up: Use go test -cover to generate coverage report

zhming0:
### Description

As we adopted `bktec` auto retry, test splitting for agent, the old single file coverage reporting mechanism will report incorrect data.

This PR:

* use go 1.20's `go test -cover` to generate test coverage binary data, submit to artifact. 
* add a test coverage step in CI, it collects all submitted coverage binary data, merge them into a single report. 
* Only make the ARM64 test step generating coverage, this accelerate test runs in all other platforms. 

### Context

part of TE-3708

### DIFF
--- a/.buildkite/steps/tests.sh
+++ b/.buildkite/steps/tests.sh
@@ -8,14 +8,18 @@ go install gotest.tools/gotestsum@v1.8.0
 
 go install github.com/buildkite/test-engine-client@v1.5.0-rc.3
 
-echo '+++ Running tests'
 export BUILDKITE_TEST_ENGINE_SUITE_SLUG=buildkite-agent
 export BUILDKITE_TEST_ENGINE_TEST_RUNNER=gotest
 export BUILDKITE_TEST_ENGINE_RESULT_PATH="junit-${BUILDKITE_JOB_ID}.xml"
 export BUILDKITE_TEST_ENGINE_RETRY_COUNT=1
-export BUILDKITE_TEST_ENGINE_TEST_CMD="gotestsum --junitfile={{resultPath}} -- -count=1 -coverprofile=cover.out $@ {{packages}}"
+if [[ "$(go env GOOS)" == "windows" ]]; then
+  # I can't get windows to work with the $COVERAGE_DIR, I tried cygpath but no luck.
+  # need a Windows VM to debug.
+  export BUILDKITE_TEST_ENGINE_TEST_CMD="gotestsum --junitfile={{resultPath}} -- -count=1 $* {{packages}}"
+else
+  mkdir -p coverage
+  COVERAGE_DIR="$PWD/coverage"
+  export BUILDKITE_TEST_ENGINE_TEST_CMD="gotestsum --junitfile={{resultPath}} -- -count=1 -cover $* {{packages}} -test.gocoverdir=${COVERAGE_DIR}"
+fi
 
 test-engine-client
-
-echo 'Producing coverage report'
-go tool cover -html cover.out -o cover.html
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -23,7 +23,9 @@ steps:
     key: test-linux-amd64
     command: ".buildkite/steps/tests.sh"
     parallelism: 2
-    artifact_paths: junit-*.xml
+    artifact_paths:
+      - junit-*.xml
+      - "coverage/**/*"
     plugins:
       - docker-compose#v4.14.0:
           config: .buildkite/docker-compose.yml
@@ -33,14 +35,14 @@ steps:
       - test-collector#v1.2.0:
           files: "junit-*.xml"
           format: "junit"
-      - artifacts#v1.9.0:
-          upload: "cover.{html,out}"
 
   - name: ":linux: Linux ARM64 Tests"
     key: test-linux-arm64
     command: ".buildkite/steps/tests.sh"
     parallelism: 2
-    artifact_paths: junit-*.xml
+    artifact_paths:
+      - junit-*.xml
+      - "coverage/**/*"
     agents:
       queue: $AGENT_RUNNERS_LINUX_ARM64_QUEUE
     plugins:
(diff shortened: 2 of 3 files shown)


.buildkite/steps/tests.sh
```diff
@@ -8,14 +8,18 @@ go install gotest.tools/gotestsum@v1.8.0
...
 export BUILDKITE_TEST_ENGINE_TEST_RUNNER=gotest
 export BUILDKITE_TEST_ENGINE_RESULT_PATH="junit-${BUILDKITE_JOB_ID}.xml"
 export BUILDKITE_TEST_ENGINE_RETRY_COUNT=1
-export BUILDKITE_TEST_ENGINE_TEST_CMD="gotestsum --junitfile={{resultPath}} -- -count=1 -coverprofile=cover.out $@ {{packages}}"
+export BUILDKITE_TEST_ENGINE_TEST_CMD="gotestsum --junitfile={{resultPath}} -- -count=1 -cover -test.gocoverdir=$COVERDIR $@ {{packages}}"
 
 test-engine-client
 
 echo 'Producing coverage report'
+go tool covdata textfmt -i "$COVERDIR" -o cover.out
```
catkins, line 24: so as we're parallising we'll need to merge it for the parallel runs too using artifacts and a dependant step I think?

zhming0: ahhh true! let me pull this back to draft.

.buildkite/steps/tests.sh
```diff
@@ -8,14 +8,15 @@ go install gotest.tools/gotestsum@v1.8.0
...
 export BUILDKITE_TEST_ENGINE_TEST_RUNNER=gotest
 export BUILDKITE_TEST_ENGINE_RESULT_PATH="junit-${BUILDKITE_JOB_ID}.xml"
 export BUILDKITE_TEST_ENGINE_RETRY_COUNT=1
-export BUILDKITE_TEST_ENGINE_TEST_CMD="gotestsum --junitfile={{resultPath}} -- -count=1 -coverprofile=cover.out $@ {{packages}}"
+if [[ "${TEST_COVERAGE:-false}" == "true" ]]; then
+  mkdir -p coverage
+  export BUILDKITE_TEST_ENGINE_TEST_CMD="gotestsum --junitfile={{resultPath}} -- -count=1 -cover $* {{packages}} -test.gocoverdir=$PWD/coverage"
+else
+  export BUILDKITE_TEST_ENGINE_TEST_CMD="gotestsum --junitfile={{resultPath}} -- -count=1 $* {{packages}}"
+fi
```
DrJosh9000, line 20: Can we continue having coverage for Windows too?

I think it's a matter of translating Unix paths back into Windows paths...

Does this work?

```suggestion
mkdir -p coverage
COVERAGE_DIR="$PWD/coverage"
if [[ "$(uname)" == *"windows"* ]]; then
  COVERAGE_DIR="$(cygpath -w "${COVERAGE_DIR}")"
fi
export BUILDKITE_TEST_ENGINE_TEST_CMD="gotestsum --junitfile={{resultPath}} -- -count=1 -cover $* {{packages}} -test.gocoverdir=${COVERAGE_DIR}"
```

zhming0: I tried a number of ways to get it works on Windows, none of them works including your suggestion 😬 .

They all failed on: https://buildkite.com/buildkite/agent/builds/10110/steps?jid=01968a3e-200d-4ff3-a51e-7d5098c9a90c

So I gave up and make this opt-in.

DrJosh9000: I suggested it without trying it, but I think the problem may be that `uname` doesn't actually return a string containing "windows" on Windows:

Screenshot 2025-05-01 at 3 29 41 pm

Doing some [reading](https://stackoverflow.com/questions/3466166/how-to-check-if-running-in-cygwin-mac-or-linux) it could be `CYGWIN*`, `MINGW*`, or `MSYS*`?

zhming0: Another complication of including windows and other platform tests is that: there seems to be some form of interference among them: 

`error: counter mode clash while reading meta-data file coverage/covmeta.15abce4185690bdced71054bf6cf38cd: previous file had atomic, new file has set`

DrJosh9000: I wonder why it's failing that way, when the Windows tests in that build all failed to write coverage profiles? ☹️

zhming0: I am not sure either. 
I don't know what the "previous file" was referring to. I have a guess, it was referring to other files about the same test records? and somehow, some are using "set" mode, others are using "atomic" mode.

But I don't know if I should drill too deep. I wonder if it would be possible to kick this can down the road?

DrJosh9000: What if we generated separate coverage reports for each platform? That should correspond to how things used to be, and hopefully avoid this apparent counter mode clash error?

At this point I don't mind if you do it with or without Windows.

zhming0: can do, but the amd64, arm64, and race are going to have basically the same data?

zhming0: cool, I created one CI step each for every test step (except for Windows). 

I gave Windows another stab (I fixed the uname check) and still no luck, the feedback cycle is bit too long for me, I don't have a windows env to test this properly. 😢

.buildkite/steps/tests.sh
```diff
@@ -8,14 +8,18 @@ go install gotest.tools/gotestsum@v1.8.0
 
 go install github.com/buildkite/test-engine-client@v1.5.0-rc.3
 
-echo '+++ Running tests'
 export BUILDKITE_TEST_ENGINE_SUITE_SLUG=buildkite-agent
 export BUILDKITE_TEST_ENGINE_TEST_RUNNER=gotest
 export BUILDKITE_TEST_ENGINE_RESULT_PATH="junit-${BUILDKITE_JOB_ID}.xml"
 export BUILDKITE_TEST_ENGINE_RETRY_COUNT=1
-export BUILDKITE_TEST_ENGINE_TEST_CMD="gotestsum --junitfile={{resultPath}} -- -count=1 -coverprofile=cover.out $@ {{packages}}"
+if [[ "$(go version)" == *"windows"* ]]; then
```
DrJosh9000, line 15: If we're using Go we can ask for the relevant info specifically:

```suggestion
if [[ "$(go env GOOS)" == "windows" ]]; then
```

zhming0: Done ✅

DrJosh9000 (approved):
Thanks for following up! LGTM ✅

PR #3296, opened 2025-05-02
Introduce a structure where coverage can increase on githttp checkout code

wolfeidau:
### Description

This PR introduces a githttptest package which provides a [Git smart http server](https://git-scm.com/book/en/v2/Git-on-the-Server-Smart-HTTP) used for testing, along with an array of utility methods to manipulate a git repository.

### Context

The checkout logic is a bit of a blind spot at the moment as there isn't any test coverage for this code. The goal of this PR is to change that by adding some tools which verify behavour, testing both the git command execution and the go code.

This will allow us to test some of the changes in #3294 

### Changes

This increases the test coverage for executor targeting the `defaultCheckoutPhase` method to start with.

### Testing

### DIFF
--- a/internal/job/checkout_test.go
+++ b/internal/job/checkout_test.go
@@ -1,0 +1,95 @@
+package job
+
+import (
+	"context"
+	"os"
+	"testing"
+
+	"github.com/buildkite/agent/v3/internal/job/githttptest"
+	"github.com/buildkite/agent/v3/internal/shell"
+	"github.com/stretchr/testify/require"
+)
+
+func TestDefaultCheckoutPhase(t *testing.T) {
+	assert := require.New(t)
+	ctx := context.Background()
+
+	shell, err := shell.New()
+	assert.NoError(err)
+
+	tests := []struct {
+		name        string
+		executor    *Executor
+		projectName string
+		checkoutDir string
+	}{
+		{
+			name: "Default checkout phase with HEAD commit",
+			executor: &Executor{
+				shell: shell,
+				ExecutorConfig: ExecutorConfig{
+					Commit:        "HEAD",
+					Branch:        "main",
+					CleanCheckout: false,
+					GitCleanFlags: "-f -d -x",
+				},
+			},
+			projectName: "project-name-head",
+		},
+		{
+			name: "Default checkout phase with custom refspec",
+			executor: &Executor{
+				shell: shell,
+				ExecutorConfig: ExecutorConfig{
+					Commit:        "HEAD",
+					Branch:        "main",
+					CleanCheckout: false,
+					GitCleanFlags: "-f -d -x",
+					RefSpec:       "refs/pull/124/head",
+				},
+			},
+			projectName: "project-name-refspec",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert := require.New(t)
+
+			t.Setenv("GIT_CONFIG_GLOBAL", "~/.gitconfig.empty")
+
+			s := githttptest.NewServer()
+			s.Start()
+			defer s.Close()
+
+			err = s.CreateRepository(tt.projectName)
+			assert.NoError(err)
+
+			err = s.InitRepository(tt.projectName)
+			assert.NoError(err)
+
+			commit, err := s.PushBranch(tt.projectName, "feature-branch")
+			assert.NoError(err)
+
+			// create a custom ref to test
+			err = s.CreateRef(tt.projectName, "refs/pull/124/head", commit)
+			assert.NoError(err)
+
+			buildDir, err := os.MkdirTemp("", "build-path-")
+			assert.NoError(err)
+			defer os.RemoveAll(buildDir)
+
+			tt.executor.BuildPath = buildDir
+			tt.executor.Repository = s.RepoURL(tt.projectName)
+
+			checkoutDir, err := os.MkdirTemp("", "checkout-path-")
+			assert.NoError(err)
+			defer os.RemoveAll(checkoutDir)
+
+			shell.Env.Set("BUILDKITE_BUILD_CHECKOUT_PATH", checkoutDir)
+
+			err = tt.executor.defaultCheckoutPhase(ctx)
+			assert.NoError(err)
+		})
+	}
+}
--- a/internal/job/checkout_test.go
+++ b/internal/job/checkout_test.go
@@ -23,7 +23,6 @@ func TestDefaultCheckoutPhase(t *testing.T) {
 		projectName string
 		checkoutDir string
 		refSpec     string
-		expectErr   bool
 	}{
 		{
 			name: "Default checkout phase with HEAD commit",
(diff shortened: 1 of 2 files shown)


internal/job/githttptest/githttptest.go
```diff
@@ -0,0 +1,371 @@
...
+	return l
+}
+
+func (s *Server) goServe() {
+	s.wg.Add(1)
+	go func() {
+		defer s.wg.Done()
+		s.Config.Serve(s.Listener)
+	}()
+}
```
DrJosh9000, line 73: Since the goal of the `githttptest` server is to provide something for tests to test against, and not for general use, it seems to me like it having to manually call `Start` and have special code to set up the listener and create a goroutine is a bunch of extra work compared with piggy-backing off of `httptest` via embedding... consider:

```go
import "net/http/httptest"

type Server struct {
	*httptest.Server
	repositories string
}

func NewServer() *Server {
	repositories, err := os.MkdirTemp("", "githttptest")
	if err != nil {
		panic(fmt.Sprintf("githttptest: failed to create temp dir: %v", err))
	}

	s := &Server{
		repositories: repositories,
	}

	mux := http.NewServeMux()
	mux.HandleFunc("/{repository}/git-upload-pack", s.handleGitUploadPack)
	mux.HandleFunc("/{repository}/git-receive-pack", s.handleGitReceivePack)
	mux.HandleFunc("/{repository}/info/refs", s.handleGitInfoRefs)

	s.Server = httptest.NewServer(mux)
	return s
}

func (s *Server) Close() {
	s.Server.Close()
	os.RemoveAll(s.repositories)
}

func (s *Server) RepoURL(repoName string) string {
	return path.Join(s.Server.URL, repoName+".git")
}
```

- The server comes pre-`Start`-ed
- `Close` is practically free, just needs that `os.RemoveAll`
- No hand-holding of `net.Listener`, goroutine, waitgroup, etc

internal/job/githttptest/githttptest.go
```diff
@@ -0,0 +1,371 @@
...
+	// For a production server, you might want to add authentication here
+	w.Header().Set("Content-Type", "application/x-git-receive-pack-result")
+	w.WriteHeader(http.StatusOK)
+
+	cmd := exec.Command("git", "receive-pack", "--stateless-rpc", repoPath)
+	cmd.Stdin = r.Body
+	cmd.Stdout = w
+	cmd.Stderr = os.Stderr
+
+	if err := cmd.Run(); err != nil {
```
DrJosh9000, line 330: Empty here too

internal/job/githttptest/githttptest.go
```diff
@@ -0,0 +1,371 @@
...
+}
+
+// handleGitUploadPack handles the git-upload-pack endpoint (used for git fetch/clone)
+func (s *Server) handleGitUploadPack(w http.ResponseWriter, r *http.Request) {
+	repoName := r.PathValue("repository")
+	repoName = strings.TrimSuffix(repoName, ".git")
+
+	if !isValidRepoName(repoName) {
+		w.WriteHeader(http.StatusBadRequest)
+		fmt.Fprintf(w, "Invalid repository name")
```
DrJosh9000, line 273: This looks like a job for `http.Error`

```suggestion
		http.Error(w, "Invalid repository name", http.StatusBadRequest)
```

internal/job/githttptest/githttptest.go
```diff
@@ -0,0 +1,371 @@
...
+		return
+	}
+
+	repoPath := filepath.Join(s.repositories, repoName)
+
+	// Check if repository exists
+	if _, err := os.Stat(repoPath); os.IsNotExist(err) {
+		w.WriteHeader(http.StatusNotFound)
+		fmt.Fprintf(w, "Repository not found")
```
DrJosh9000, line 282:
```suggestion
		http.Error(w, "Repository not found", http.StatusNotFound)
```

internal/job/githttptest/githttptest.go
```diff
@@ -0,0 +1,371 @@
...
+	// Smart HTTP protocol requires a specific preamble
+	pktLine := fmt.Sprintf("# service=%s\n", service)
+	fmt.Fprintf(w, "%04x%s0000", len(pktLine)+4, pktLine)
+
+	// Execute the corresponding Git command
+	cmd := exec.Command("git", strings.TrimPrefix(service, "git-"), "--stateless-rpc", "--advertise-refs", repoPath)
+	cmd.Stdout = w
+	cmd.Stderr = os.Stderr
+
+	if err := cmd.Run(); err != nil {
```
DrJosh9000, line 369: And here

internal/job/githttptest/githttptest.go
```diff
@@ -0,0 +1,371 @@
...
+	w.Header().Set("Content-Type", "application/x-git-upload-pack-result")
+	w.WriteHeader(http.StatusOK)
+
+	cmd := exec.Command("git", "upload-pack", "--stateless-rpc", repoPath)
+	cmd.Stdin = r.Body
+	cmd.Stdout = w
+	cmd.Stderr = os.Stderr
+
+	if err := cmd.Run(); err != nil {
+	}
```
DrJosh9000, line 295: Empty `if` body here - perhaps don't stream the output from `git upload-pack` directly into `w`, and instead buffer it, so that the error can be reported as say `http.StatusInternalServerError` and the error message?

```suggestion
	buf := bytes.NewBuffer(nil)

	cmd := exec.Command("git", "upload-pack", "--stateless-rpc", repoPath)
	cmd.Stdin = r.Body
	cmd.Stdout = buf
	cmd.Stderr = os.Stderr

	if err := cmd.Run(); err != nil {
		http.Error(w, err.Error(), http.StatusInternalServerError)
		return
	}
	
	w.Header().Set("Content-Type", "application/x-git-upload-pack-result")
	w.WriteHeader(http.StatusOK)
	io.Copy(w, buf)
```

DrJosh9000 (commented):
It's good to finally improve the coverage here!

internal/job/githttptest/githttptest.go
wolfeidau: Great suggestion!

internal/job/githttptest/githttptest.go
```diff
@@ -0,0 +1,333 @@
...
+
+	// Check if repository exists
+	if _, err := os.Stat(repoPath); os.IsNotExist(err) {
+		w.WriteHeader(http.StatusNotFound)
+		fmt.Fprintf(w, "Repository not found")
+		return
+	}
+
+	// Check if we're allowing pushes
+	// For a production server, you might want to add authentication here
```
catkins, line 272: is this comment a copypasta from somewhere?

internal/job/checkout_test.go
```diff
@@ -0,0 +1,126 @@
...
+	shell, err := shell.New()
+	assert.NoError(err)
+
+	tests := []struct {
+		name        string
+		executor    *Executor
+		projectName string
+		checkoutDir string
+		refSpec     string
+		expectErr   bool
```
catkins, line 26: I think this is currently unused?

catkins (approved):
:+1: Great work @wolfeidau! I can tell you've put in some hours on reading the docs for the git porcelain commands 🔍

internal/job/checkout_test.go
wolfeidau: @catkins yeah I want to add some error cases, can remove for now.

PR #3297, opened 2025-05-05
Set ignore_agent_in_dispatches when finishing with disconnect-after-job

DrJosh9000:
### Description

Prevent jobs being dispatched to agents that are finishing up by passing a newly-available flag in the finish call.

### Context

https://linear.app/buildkite/issue/PS-554

### Changes

- Thread the new flag through the call stack
- Pass it when disconnect-after-job is enabled

### Testing


wolfeidau (approved):
👍🏻 happy to see this merged as I use this feature 😂

Issue #3303, opened 2025-05-07
Add config option to fail pipeline upload on parsing warnings as well as errors


nic-littlepay:
**Is your feature request related to a problem? Please describe.**
Sometimes our pipeline contains bad yaml that means the pipeline won't run or result in the way we expect. [The agent will log warnings relating to parse issues](https://github.com/buildkite/agent/blob/64ac08b6916a66a6801c45a4782d240d0cfe5239/clicommand/pipeline_upload.go#L255) but it will continue to upload the pipeline.

**Describe the solution you'd like**
Add a option to configure the agent to bail out on any warnings or errors returned from parsing.

**Describe alternatives you've considered**
We've tired no real alternatives at this point although we did suggest adding some sort of separate pipeline yaml linting, but not exactly sure what that could look like or if it's the best solution.

**Additional context**
 - Re linking [this area of the agent code](https://github.com/buildkite/agent/blob/64ac08b6916a66a6801c45a4782d240d0cfe5239/clicommand/pipeline_upload.go#L255) that I think would be where the agent bails on parse warnings.

dabarrell:
Hi @nic-littlepay - thanks for raising this! I'm going to double check internally to make sure I understand the reasoning behind it being a warning (not an error) in the first place, then I'll get back to you. Thanks for your patience!

nic-littlepay:
@dabarrell , have you had any update on this?

nic-littlepay:
Gday @dabarrell , dno't suppose there's ben any movement on this one?

DrJosh9000:
Hi, I believe this has been implemented in #4031 and was released in v3.130.0. Just pass the `--reject-parse-warnings` flag.

PR #3304, opened 2025-05-08
Add config option to fail pipeline upload on parsing warnings as well as errors

nic-littlepay:
### Description

Attempts to provide fixe for issue #3303 

### Context

There's slightly more context in the issue but broadly we wanted to be able to configure the agent to fail pipeline if it finds parse warnings in pipeline upload.

### Changes

 - Adds a new config bool flag to pipeline upload command.
 - If that flag is set `true`, returns error on parse warnings.

### Testing


DrJosh9000:
I believe this has been accomplished in #4031. Please let us know if there's anything else we can do!

PR #3306, opened 2025-05-12
Add kubernetes-bootstrap

DrJosh9000:
### Description

Reduce Kubernetes resource size (when using agent-stack-k8s) by applying more environment variables from the socket.

### Context

A customer had feedback that having a lot of env vars repeated across containers in the pod spec can consume a lot of space in etcd.

### Changes

- Add a `kubernetes-bootstrap` subcommand. This allows fetching the env vars from the socket before the `bootstrap` subcommand runs, which was the fatal flaw in commit 36940e5
- Filter out the kubernetes plugin when creating the bootstrap env
- Some small-scale refactoring of flags and env helpers

### Testing

### DIFF
--- a/clicommand/cancel_signal.go
+++ b/clicommand/cancel_signal.go
@@ -10,7 +10,7 @@ import (
 const (
 	defaultCancelGracePeriodSecs = 10
 	defaultSignalGracePeriodSecs = -1
-	defaultSignalGracePeriod     = (defaultCancelGracePeriodSecs + defaultCancelGracePeriodSecs) * time.Second
+	defaultSignalGracePeriod     = (defaultCancelGracePeriodSecs + defaultSignalGracePeriodSecs) * time.Second
 )
 
 var (
--- a/clicommand/cancel_signal.go
+++ b/clicommand/cancel_signal.go
@@ -8,31 +8,33 @@ import (
 )
 
 const (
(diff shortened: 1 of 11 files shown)


clicommand/cancel_signal.go
```diff
@@ -8,31 +8,33 @@ import (
 )
 
 const (
-	defaultCancelGracePeriod = 10
+	defaultCancelGracePeriodSecs = 10
+	defaultSignalGracePeriodSecs = -1
+	defaultSignalGracePeriod     = (defaultCancelGracePeriodSecs + defaultCancelGracePeriodSecs) * time.Second
```
CerealBoy, line 13:
```suggestion
	defaultSignalGracePeriod     = (defaultCancelGracePeriodSecs + defaultSignalGracePeriodSecs) * time.Second
```

Unless I'm misunderstanding this, we want `defaultSignalGracePeriod` to be the combination of both?

CerealBoy (approved):
I had a small question about the default config setup, otherwise LGTM!

clicommand/cancel_signal.go
DrJosh9000: Good catch. This has often been confusing, and I probably didn't strictly need to try to improve it.

PR #3312, opened 2025-05-13
fix: Don't disconnect-after-idle when just given a job

DrJosh9000:
### Description

Fixes bug introduced in #3238.

### Context

https://linear.app/buildkite/issue/PS-554

### Changes

- Only apply disconnect-after-idle-timeout check if the ping did not yield a job.
- Log if a job was yielded in states where it is unexpected (acquire-job, disconnect-after-job after job)

### Testing


wolfeidau (approved):
👍🏻 great work spotting the issue.

PR #3314, opened 2025-05-14
Fix env loading order in kubernetes-bootstrap

DrJosh9000:
### Description

The registration response env vars are meant to be read to find out `BUILDKITE_PTY`, etc, but these were being added to `environ` after. 

### Context

Found during testing.

### Changes

- Merge the registration response vars into `environ` before reading them.
- A tiny simplification to `$PATH` computation found when trying to understand it.

### Testing


CerealBoy (approved):
🚀

PR #3322, opened 2025-05-23
Fix `redactor add --format json` help string

francoiscampbell:
### Description
The help string for `buildkite-agent redactor add --format json` has a bug; the example given uses the JSON object `{"key":"secret1","key":"secret2"}`, which is invalid because it contains duplicate keys. At runtime, this value is unmarshaled and the second value takes precedence over the first, which prevents redaction of all secrets but the last.

A test using version 3.88:
image

The proper syntax is to use unique names for each key:
image

### Context
Current help string for `buildkite-agent redactor add --format json`

### Changes
Updated the help string with unique key names and specified that the key names have no effect on redaction based on https://github.com/buildkite/agent/blob/8f29ee7bf1e739a2664a7b887428c39d298a6fe3/clicommand/redactor_add.go#L150-L152 ignoring the key.

### Testing
This is a documentation-only change, the output when using the example in the help string in a test job is above.

### DIFF
--- a/clicommand/redactor_add.go
+++ b/clicommand/redactor_add.go
@@ -68,10 +68,10 @@ To redact the string 'llamasecret' from future logs:
 		$ echo llamasecret | buildkite-agent redactor add
 
 To redact multiple secrets from future logs in one command, create a flat
-JSON object file (for example, 'my-secrets.json'), with multiple "key" values,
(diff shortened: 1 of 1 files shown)


DrJosh9000 (approved):
Thanks, much appreciated!

PR #3326, opened 2025-05-30
Tag tests with os / arch

catkins:
### Description

I was looking at the beatiful @buildkite/test-engine Tests tab in an agent build today, and realised that it would be awesome to have tags (and then labels) on each of the executions so that they would show up in the UI more clearly as from separate platforms as opposed to retries.

CleanShot 2025-05-30 at 14 34 29@2x

### Context

https://github.com/buildkite-plugins/test-collector-buildkite-plugin/releases/tag/v1.11.0

### Changes

- Add test tagging using the bk test collector plugin

### Testing

N/A


catkins:
(I just amended my commit to pick up the label config in Test Engine by running tests again)

CleanShot 2025-05-30 at 14 43 21@2x

PR #3327, opened 2025-06-02
Reduce git fetch from twice to once for typical Github PR build

zhming0:
### Description

Currently, when a PR triggers a build, we trigger two `git fetch`. This PR merges these two `git fetch` into one.

### Context

PS-624

### Testing

### DIFF
--- a/internal/job/checkout.go
+++ b/internal/job/checkout.go
@@ -576,21 +576,22 @@ func (e *Executor) defaultCheckoutPhase(ctx context.Context) error {
 		// https://help.github.com/articles/checking-out-pull-requests-locally/#modifying-an-inactive-pull-request-locally
 		e.shell.Commentf("Fetch and checkout pull request head from GitHub")
 		refspec := fmt.Sprintf("refs/pull/%s/head", e.PullRequest)
+		refspecs := []string{refspec}
 
-		if err := gitFetch(ctx, e.shell, gitFetchFlags, "origin", refspec); err != nil {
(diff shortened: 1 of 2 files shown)


internal/job/checkout.go
```diff
@@ -576,22 +576,23 @@ func (e *Executor) defaultCheckoutPhase(ctx context.Context) error {
...
-			return fmt.Errorf("fetching PR refspec %q: %w", refspec, err)
+		// If we know the commit, also fetch it directly. The commit might not be in the history of `refspec` if there
+		// have been force pushes to the pull request, so this ensures we have it.
+		// Note: this is the typical case e.Commit != HEAD.
+		if e.Commit != "HEAD" {
+			refspecs = append(refspecs, e.Commit)
+		}
+
+		// We aim to eliminate network round-trip as much as possible so we use a single git fetch here.
+		if err := gitFetchWithFallback(ctx, e.shell, gitFetchFlags, refspecs...); err != nil {
```
zhming0, line 589: It's worthnoting that this change will make `git fetch refs/pull/<pr>/head` auto retry when remote ref is not found, this retry is unnecessary technically, but the likelihood of this happening is really low so the retry isn't going to be harmful.

catkins: cc/ @123sarahj123 who'd been working on some git fetch retry stuff for an OpenAI ticket

123sarahj123: https://github.com/buildkite/agent/pull/3294

PR #3330, opened 2025-06-03
Set job log tempfile permissions to 644 (was 600)

moskyb:
- **Make job log tempfile world-readable**

### Description

Currently, when users use the `--job-log-tempfile` flag on the agent, the tempfile is created with `600` permissions (ie, readable and writable by the current user, unavailable to everyone else). This can be frustrating, as a primary use case for this file is ingestion by a log ingestion framework a la Cloudwatch/Stackdriver/Loki, which often run as different users.

To fix this, this PR sets the permissions on the job log tempfile to `644` (readable and writable by us, readable by everyone else). This is relatively in line with the unix permissions policy for files created by the agent, which are generally created with relatively permissive permissions (usually 777), to be reduced by the umask. Given that job logs can be somewhat sensitive, I've chosen to have an explicitly lower permission set than usual for this specific file.

### Context

[PS-276](https://linear.app/buildkite/issue/PS-726/2-zipline-logfile-permission-needs-update-from-600-to-644#comment-5b396b68)

### Testing

### DIFF
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -265,6 +265,12 @@ func NewJobRunner(ctx context.Context, l logger.Logger, apiClient APIClient, con
 		if err != nil {
 			return nil, err
 		}
+
+		err := os.Chmod(tmpFile.Name(), 0o644) // Make it world-readable - useful for log collection etc
+		if err != nil {
+			return nil, fmt.Errorf("failed to set permissions on job log tmpfile %s: %w", tmpFile.Name(), err)
+		}
+
 		os.Setenv("BUILDKITE_JOB_LOG_TMPFILE", tmpFile.Name())
 		outputWriter = io.MultiWriter(outputWriter, tmpFile)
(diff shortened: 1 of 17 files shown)


DrJosh9000 (approved):
Thanks so much for updating all the other octal literals!

agent/job_runner.go
```diff
@@ -265,6 +265,12 @@ func NewJobRunner(ctx context.Context, l logger.Logger, apiClient APIClient, con
 		if err != nil {
 			return nil, err
 		}
+
```
catkins, line 268: [`os.CreateTemp`](https://pkg.go.dev/os#CreateTemp) ah cool, I see in the godocs the default perms
> The file is created with mode 0o600 (before umask)

catkins (approved):
:+1: Thanks @moskyb.

Also, good opportunistic tidyup on the modern octal notation 👏

PR #3331, opened 2025-06-04
Fix bootstrap subprocess handling

DrJosh9000:
### Description

Fixes the handling of the bootstrap subprocess from `buildkite-agent kubernetes-bootstrap`.

### Context

Signal propagation (from Kubernetes, in this case) was not working, because the `Run` method of `process` blocks, and signal forwarding was set up afterwards.

https://linear.app/buildkite/issue/PIPE-1120/delays-in-k8s-job-deletion-reflected-in-buildkite-ui

### Changes

- Put signal forwarding in a goroutine
- Rearrange existing code
- Use -1 exit code for failure to start subprocess (like JobRunner)

### Testing


zhming0 (approved):
👍🏿

PR #3339, opened 2025-06-13
Fix PR ref retry logic

moskyb:
### Description

In #3294, we introduced behaviour to retry `git fetch`es during the default checkout hook for checkouts using git mirrors. This is because during the checkout process for a PR build we fetch the `refs/pull/<some number>/head` ref that GitHub creates. The creation of this ref happens asynchronously to the PR being opened though, so in some cases a build will have started before the ref has been created. In this situation we're trying to fetch a ref that doesn't (yet) exist, which will fail the fetch, so we wait around and retry for a bit until either the ref exists, or we run out our retry loop, which tries 10 times over about 2 minutes.

However, #3294 also introduced a couple of bugs and oddities in this flow, primarily that:
1. The retry loop for fetching PR refs was nested inside [another retry loop](https://github.com/buildkite/agent/blob/main/internal/job/checkout.go#L160-L163) encapsulating the whole checkout phase. This meant that if there were a non-transient error in fetching the ref, both the inner (just retrying the fetch) and the outer (retrying the entire checkout phase) retry loop would run, meaning that the error would be retried 30 times instead of 10, and take ~7 minutes instead of ~2
1. Instead of just retrying the fetch for the PR head ref (`refs/pull/123/head`), all fetches in the git mirror update would be retried 30 times, not just the PR head ref. This means that if, for example, someone were to create a build against a branch that doesn't exist, the agent would try to check out that branch for 7 minutes before finally giving up, even though that branch is never going to exist

This PR fixes the above two issues, and makes a couple of cleanups along the way. It:
- Makes it so that only the PR ref fetch is retried, so bug/oddity number 2 doesn't happen. Fetches for refs that aren't the PR head won't have an inner retry loop, and will only retried as part of the outer (entire checkout phase) retry loop
- Makes it so that if the inner retry loop happens, the outer one is skipped -- so if fetching the PR head ref fails 10 times, it won't get kicked back out to the outer retry loop

All of this is to say that users will generally be waiting around on git fetches a whole bunch less now.

Additionally, I've made a couple of non-functional changes:
- The `job.gitFetch` function's signature was getting super long and hard to read, so I've updated it to use an args struct
- Make the error handling in the checkout phase a little more straightforward, with slightly fewer nested switches
- I've spruced up the logs for retrying PR head ref fetches a wee bit:
<details>
<summary>Before</summary>

CleanShot 2025-06-13 at 15 05 04@2x
</details>

<details>
<summary>After</summary>

CleanShot 2025-06-13 at 15 05 52@2x
</details>

### Context

[PS-755](https://linear.app/buildkite/issue/PS-755)

### Testing

### DIFF
--- a/internal/job/checkout.go
+++ b/internal/job/checkout.go
@@ -167,6 +167,8 @@ func (e *Executor) CheckoutPhase(ctx context.Context) error {
 			}
 
 			var errLockTimeout ErrTimedOutAcquiringLock
+			var errGit *gitError
+
 			switch {
 			case shell.IsExitError(err) && shell.ExitCode(err) == -1:
 				e.shell.Warningf("Checkout was interrupted by a signal")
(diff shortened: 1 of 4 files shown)


internal/job/checkout.go
```diff
@@ -186,34 +188,49 @@ func (e *Executor) CheckoutPhase(ctx context.Context) error {
...
-					default:
+				switch errGit.Type {
+				case gitErrorClean, gitErrorCleanSubmodules, gitErrorClone,
+					gitErrorCheckoutRetryClean, gitErrorFetchRetryClean,
+					gitErrorFetchBadObject:
+					// Checkout can fail because of corrupted files in the checkout which can leave the agent in a state where it
+					// keeps failing. This removes the checkout dir, which means the next checkout will be a lot slower (clone vs
+					// fetch), but hopefully will allow the agent to self-heal
+					if err := e.removeCheckoutDir(); err != nil {
+						e.shell.Printf("Failed to remove checkout dir while cleaning up after a checkout error.")
```
wolfeidau: Should this be a warning as well?

internal/job/git.go
```diff
@@ -34,14 +34,21 @@ const (
 	gitErrorCleanSubmodules
 )
 
+const (
+	gitErrStrbadObject             = "fatal: bad object"
```
wolfeidau: Capital B for bad in this error string?

wolfeidau (changes requested):
Couple of small questions / nit picks.

internal/job/git.go
moskyb: ooh good catch

DrJosh9000: Yeah I don't think so - this is for matching the string output by git.

DrJosh9000: Unless you mean the identifier? `gitErrStrbadObject` -> `gitErrStrBadObject` definitely.

internal/job/checkout.go
```diff
@@ -380,17 +397,28 @@ func (e *Executor) updateGitMirror(ctx context.Context, repository string) (stri
 
 	if isMainRepository {
 		if e.PullRequest != "false" && strings.Contains(e.PipelineProvider, "github") {
-			e.shell.Commentf("Fetch and mirror pull request head from GitHub")
+			e.shell.Commentf("Fetch and mirror pull request head from GitHub. This will be retried if it fails, as the pull request head might not be available yet — GitHub creates them asynchronously")
```
moskyb: is it too wordy?

internal/job/checkout.go
```diff
@@ -380,17 +397,28 @@ func (e *Executor) updateGitMirror(ctx context.Context, repository string) (stri
...
 
 		} else {
 			// Fetch the build branch from the upstream repository into the mirror.
-			if err := gitFetch(ctx, e.shell, fmt.Sprintf("--git-dir=%s", mirrorDir), "", "origin", true, e.Branch); err != nil {
+			if err := gitFetch(ctx, gitFetchArgs{
+				Shell:      e.shell,
+				GitFlags:   fmt.Sprintf("--git-dir=%s", mirrorDir),
+				Repository: "origin",
+				RefSpecs:   []string{e.Branch},
+			}); err != nil {
```
moskyb, line 421: note that we're no longer retrying in this stanza

internal/job/git.go
```diff
@@ -130,18 +137,21 @@ func gitCleanSubmodules(ctx context.Context, sh *shell.Shell, gitCleanFlags stri
...
-	sh *shell.Shell,
-	gitFlags, gitFetchFlags, repository string,
-	withRetry bool,
-	refSpec ...string,
-) error {
+type gitFetchArgs struct {
+	Shell         *shell.Shell // The shell to run the command in
+	GitFlags      string       // Global git flags to pass to the command
+	GitFetchFlags string       // Flags to pass to the fetch command
+	Repository    string       // The remote to fetch from
```
moskyb, line 144: i was a little surprised to see that in the `git fetch` command, this argument is indeed called `repository`, when we almost exclusively call it with `origin` (a remote), but that is apparently real.

internal/job/git.go
wolfeidau: Yeah just the Go variable names 👍🏻

internal/job/checkout.go
wolfeidau: @moskyb This is should be rare so 🤷🏻 more is more, we can address feedback when we get it IMO

wolfeidau (approved):
Looks good to me 👍🏻

internal/job/git.go
```diff
@@ -34,14 +34,21 @@ const (
 	gitErrorCleanSubmodules
 )
 
+const (
+	gitErrStrBadObject             = "fatal: bad object"
+	gitErrStrbadReference          = "fatal: couldn't find remote ref"
+	gitErrStrbadReferencePreGit221 = "fatal: Couldn't find remote ref"
```
DrJosh9000:
```suggestion
	gitErrStrBadReference          = "fatal: couldn't find remote ref"
	gitErrStrBadReferencePreGit221 = "fatal: Couldn't find remote ref"
```

PR #3346, opened 2025-06-19
PS-794: fix vendored plugin path ending with slash breaking envvar names

zhming0:
### Description

Fixes an issue when vendored plugin path ending with `/`, the plugin parameter env var becomes: `BUILDKITE_PLUGINS__{var name}` instead of `BUILDKITE_PLUGINS_{plugin name}_{var name}`.

### Context

PS-794

### Testing


DrJosh9000 (approved):
LGTM! ✅

PR #3347, opened 2025-06-19
Upload all pipelines present in the input

DrJosh9000:
### Description

Allow pipeline YAML containing more than one document. When using `buildkite-agent pipeline upload` with such an input, upload all of them.

### Context

https://linear.app/buildkite/issue/PS-795

### Changes

- Update go-pipeline to v0.14.0
- Change `parseAndInterpolate` to use `pipeline.ParseAll` and return an iterator
- Change the action to put most of the work in a loop
- Move dry-run output setup outside of the loop so that the output YAML will consist of multiple correctly-separated documents

### Testing

### DIFF
--- a/clicommand/pipeline_upload.go
+++ b/clicommand/pipeline_upload.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"iter"
 	"maps"
 	"os"
 	"os/exec"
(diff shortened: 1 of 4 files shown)


wolfeidau (approved):
Nice work, looks really clean with the iterator 👍🏻 🚀

PR #3348, opened 2025-06-20
[PIPE-1021] Propagate parent OTel trace/span from backend if provided

catkins:
### Description

As part of the OpenTelemetry Tracing Notification Service work, we provide the `traceparent` from the job acquire / accept endpoints. When present, and the user has opted-in, we set the root span for the job to use the provided trace/parent span id.

This is based on the W3C Tracecontext

### Testing

### DIFF
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -617,6 +617,10 @@ func (r *JobRunner) createEnvironment(ctx context.Context) ([]string, error) {
 	if r.conf.AgentConfiguration.TracingBackend != "" {
 		env["BUILDKITE_TRACING_BACKEND"] = r.conf.AgentConfiguration.TracingBackend
 		env["BUILDKITE_TRACING_SERVICE_NAME"] = r.conf.AgentConfiguration.TracingServiceName
+
+		if r.conf.Job.TraceParent != "" {
+			env["BUILDKITE_TRACING_TRACEPARENT"] = r.conf.Job.TraceParent
+		}
 	}
 
 	env["BUILDKITE_AGENT_DISABLE_WARNINGS_FOR"] = strings.Join(r.conf.AgentConfiguration.DisableWarningsFor, ",")
--- a/internal/job/tracing.go
+++ b/internal/job/tracing.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"slices"
 	"strconv"
+	"strings"
 
 	"github.com/buildkite/agent/v3/env"
 	"github.com/buildkite/agent/v3/tracetools"
@@ -168,6 +169,21 @@ func (e *Executor) startTracingOpenTelemetry(ctx context.Context) (tracetools.Sp
 		trace.WithSchemaURL(semconv.SchemaURL),
 	)
 
+	// if we have server side tracing enabled, we provide the parent tracing context
+	if e.ExecutorConfig.TracingTraceParent != "" {
+		traceID, spanID, err := extractTraceParent(e.ExecutorConfig.TracingTraceParent)
+		if err != nil {
+			e.shell.Warningf("error extracting tracing context: %v", err)
+		} else {
+			spanContext := trace.NewSpanContext(trace.SpanContextConfig{
+				TraceID:    traceID,
+				SpanID:     spanID,
+				TraceFlags: trace.FlagsSampled,
+			})
+			ctx = trace.ContextWithRemoteSpanContext(ctx, spanContext)
+		}
+	}
+
 	ctx, span := tracer.Start(ctx, e.otRootSpanName(),
 		trace.WithAttributes(
 			attribute.String("analytics.event", "true"),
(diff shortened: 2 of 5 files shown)


agent/job_runner.go
```diff
@@ -617,6 +617,13 @@ func (r *JobRunner) createEnvironment(ctx context.Context) ([]string, error) {
 	if r.conf.AgentConfiguration.TracingBackend != "" {
 		env["BUILDKITE_TRACING_BACKEND"] = r.conf.AgentConfiguration.TracingBackend
 		env["BUILDKITE_TRACING_SERVICE_NAME"] = r.conf.AgentConfiguration.TracingServiceName
+
+		if r.conf.Job.TraceParent != "" {
```
wolfeidau, line 625: maybe a comment to explain what one mite find in this value.

clicommand/agent_start.go
```diff
@@ -173,8 +173,9 @@ type AgentStartConfig struct {
 	MetricsDatadogDistributions bool   `cli:"metrics-datadog-distributions"`
 
 	// Tracing config
-	TracingBackend     string `cli:"tracing-backend"`
-	TracingServiceName string `cli:"tracing-service-name"`
+	TracingBackend           string `cli:"tracing-backend"`
+	TracingServiceName       string `cli:"tracing-service-name"`
+	TracingAcceptTraceparent bool   `cli:"tracing-accept-traceparent"`
```
wolfeidau: Is ther a reason why you wouldn't want to propegate this value? Is propegate a better word?

internal/job/tracing.go
```diff
@@ -183,6 +186,69 @@ func (e *Executor) startTracingOpenTelemetry(ctx context.Context) (tracetools.Sp
 	return tracetools.NewOpenTelemetrySpan(span), ctx, stop
 }
 
+// accepting traceparent from Buildkite control plane is an opt-in feature as its technically
+// a breaking change to the behaviour, and if the server-side tracing isn't set up
+// correctly, agent traces may end up without root spans to link to
```
wolfeidau: OK I see now, is this actualy the case though based on your code it looks like it skips this issue?

internal/job/tracing.go
```diff
@@ -183,6 +186,69 @@ func (e *Executor) startTracingOpenTelemetry(ctx context.Context) (tracetools.Sp
...
+	return trace.ContextWithRemoteSpanContext(ctx, trace.NewSpanContext(trace.SpanContextConfig{
+		TraceID:    traceID,
+		SpanID:     spanID,
+		TraceFlags: traceFlags,
+	}))
+}
+
+// parse W3C trace context traceparent header format
+// https://www.w3.org/TR/trace-context/#traceparent-header
+func extractTraceParent(traceParent string) (trace.TraceID, trace.SpanID, trace.TraceFlags, error) {
```
wolfeidau: Can you just use the inbuild parser? 

```go
        otel.GetTextMapPropagator().Inject(r.Context(), propagation.HeaderCarrier(w.Header()))
```

clicommand/agent_start.go
catkins: The rationale for making it opt-in is in the comment in `internal/job/tracing.go` you commented on. Propagate is a better word 🧠  I'll change accept => propagate

internal/job/tracing.go
catkins: Ooh I was trying to find an implementation of that in the otel sdk but was obviously looking in the wrong spots! I'll have a dig and get rid of this handcrafted logic!

internal/job/tracing.go
catkins: It's mostly the breaking change thing in behavior. In some tools like Grafana it looks very clunky if you're missing the root span from the trace.

The backend will add a `traceparent` property in the agent jobs accept / acquire APIs (behind a FF currently) but may not have the opentelemetry tracing notification service set up, so we only want to propagate it when they opt-in for now.

internal/job/tracing.go
catkins: Okay, sorted in [4963913](https://github.com/buildkite/agent/pull/3348/commits/4963913280a5f65d8f62f696528f99d43a05fbcc)

- I needed use `Extract` rather than `Inject`, to decorate the `context.Context`
- I couldn't use headers, because (a) its not passed as a header from the backend and (b) the whole `buildkite-agent bootstrap` dance that passes off the parameters and config to the child process, so I just used a `propagation.MapCarrier` and stuffed the value where I needed it.


Refs:
- https://github.com/open-telemetry/opentelemetry-go/blob/v1.36.0/propagation/trace_context.go#L67-L128
- https://github.com/open-telemetry/opentelemetry-go/blob/v1.36.0/propagation/propagation.go#L45-L69

clicommand/agent_start.go
catkins: Updated the terminology to propagate in [d6721c9](https://github.com/buildkite/agent/pull/3348/commits/d6721c9eb1dde064b50b5b2c865178e9b6f00412)

agent/job_runner.go
catkins: 👍 1b6afee2

catkins:
With the `--tracing-propagate-traceparent` flag provided, and otel backend pointing to grafana

CleanShot 2025-06-20 at 14 45 58@2x

With otel backend pointing to grafana but without the flag: Maintains the current behaviour where the root span is the job in the `buildkite-agent` service.

CleanShot 2025-06-20 at 14 46 06@2x

wolfeidau (approved):
Looks good 👍🏻 from me, be great to try this out 🚀 📈

PR #3349, opened 2025-06-20
Fix AquireJob to return early and trigger a sentinal error for rejection

wolfeidau:
### Description

This change https://github.com/buildkite/agent/commit/77fc7dc04462183d29d44217a18d13fa710fabd7 introduced a regression which resulted in acquire job not triggering a exit code 27 for jobs when acquire job API calls are rejected by buildkite.

### Context

https://linear.app/buildkite/issue/MDC-493/regression-in-agent-behavior-not-reporting-non-eligible-jobs

### Changes

Do the same sentinel error check on the return from AcquireAndRunJob and returning early to restore the previous exit code behaviour.

> [!WARNING]  
> This change resolves a regression, if you encounter issues with this release rollback raise an issue or contact support@buildkite.com

### Testing


CerealBoy (approved):
LGTM! 🚀

PR #3359, opened 2025-06-25
Fix regression in pipeline upload with no-interpolation

### LINKED ISSUE #3358: Regression in `buildkite-agent` v3.99.0 with `--no-interpolation` and step `key`
## Summary

Starting from Buildkite Agent version **3.99.0**, using `--no-interpolation` with nested pipeline uploads fails when a step defines a `key`, even though the parent pipeline does not reuse the key itself.

### Error

```bash
buildkite-agent: fatal: Failed to upload and process pipeline: Pipeline upload rejected: The key "test-key" has already been used by another step in this build
```

This appears to be a regression, as previous versions allowed such nested uploads without complaint.

---

## 🔁 Reproduction Steps

Given the following pipeline structure:

### 1. Initial pipeline uploaded via UI

```yaml
steps:
  - label: "testing"
    command:
      - buildkite-agent pipeline upload test-pipeline.yml
```

### 2. `test-pipeline.yml`

```yaml
steps:
  - label: "testing"
    command: buildkite-agent pipeline upload test-pipeline-uploaded.yml --no-interpolation
```

### 3. `test-pipeline-uploaded.yml`

```yaml
steps:
  - label: "testing"
    key: "test-key"
    command:
      - echo hello world
```

This results in the following failure in v3.99.0:

Error Screenshot

---

## ✅ Expected Behavior

The nested `pipeline upload` should succeed as long as no two steps within the *same final flattened pipeline* reuse the same key. In this case, only one step uses the `key: test-key`.

This behavior worked in previous versions of the agent, and pipelines depending on this now fail after upgrading to v3.99.0.

---

## 📌 Notes

- The issue only happens when using `--no-interpolation`.

DrJosh9000:
### Description

When interpolation was disabled, a pipeline would be uploaded twice.

### Context

Fixes #3358 

### Changes

- Fix the bug
- Add new test
- Make all the other tests check outputs as a whole

### Testing

### DIFF
--- a/clicommand/pipeline_upload.go
+++ b/clicommand/pipeline_upload.go
@@ -548,32 +548,36 @@ func searchForSecrets(
 func (cfg *PipelineUploadConfig) parseAndInterpolate(ctx context.Context, src string, input io.Reader, environ *env.Environment) iter.Seq2[*pipeline.Pipeline, error] {
 	return func(yield func(*pipeline.Pipeline, error) bool) {
 		for result, err := range pipeline.ParseAll(input) {
-			if err != nil && !warning.Is(err) {
-				if !yield(nil, fmt.Errorf("pipeline parsing of %q failed: %w", src, err)) {
-					return
-				}
-			}
-			if cfg.NoInterpolation {
+			// Check the error, apply interpolation if needed.
+			switch {
+			case err != nil && !warning.Is(err):
+				err = fmt.Errorf("pipeline parsing of %q failed: %w", src, err)
+				// yield below
+
+			case cfg.NoInterpolation:
 				// Note that err may be nil or a non-nil warning from pipeline.Parse
-				if !yield(result, err) {
-					return
+				// yield below
+
+			default: // yes, interpolation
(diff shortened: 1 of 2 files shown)


clicommand/pipeline_upload.go
```diff
@@ -552,12 +552,14 @@ func (cfg *PipelineUploadConfig) parseAndInterpolate(ctx context.Context, src st
 				if !yield(nil, fmt.Errorf("pipeline parsing of %q failed: %w", src, err)) {
 					return
 				}
+				continue
```
wolfeidau: should this refactored into a switch?

wolfeidau (changes requested):
Just one suggestion, i find all the continues are a sign of maybe needing a switch statement?

clicommand/pipeline_upload.go
DrJosh9000: Done.

wolfeidau (approved):
👍🏻 🚀

PR #3366, opened 2025-07-01
[PB-116] Add support for http/protobuf transport for OTLP

catkins:
### Description

The OpenTelemetry integration currently, only supports sending tracing with the grpc transport, but with a small amount of juggling it can support http/protobuf transport also.

### Context

PB-116

### Changes

- Add support for http/protobuf transport for OpenTelemetry

### Testing

### DIFF
--- a/internal/job/tracing.go
+++ b/internal/job/tracing.go
@@ -18,8 +18,8 @@ import (
 	"go.opentelemetry.io/contrib/propagators/ot"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
-	"go.opentelemetry.io/otel/exporters/otlp/otlptrace"
 	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc"
+	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp"
 	"go.opentelemetry.io/otel/propagation"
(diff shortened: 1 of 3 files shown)


zhming0 (approved):

internal/job/tracing.go
```diff
@@ -125,8 +125,24 @@ func (e *Executor) otRootSpanName() string {
...
+	// Set up trace exporter based on protocol
+	var exporter sdktrace.SpanExporter
+	var err error
+	protocol := os.Getenv("OTEL_EXPORTER_OTLP_PROTOCOL")
+	// default to grpc to avoid breaking change
+	if protocol == "" {
+		protocol = "grpc"
+	}
+
+	switch protocol {
```
DrJosh9000: [Suggestion] Declare the vars closer to where they are set, because they are irrelevant to the code in between:

```suggestion
	// Set up trace exporter based on protocol
	protocol := os.Getenv("OTEL_EXPORTER_OTLP_PROTOCOL")
	// default to grpc to avoid breaking change
	if protocol == "" {
		protocol = "grpc"
	}
	
	var exporter sdktrace.SpanExporter
	var err error
	switch protocol {
```

catkins: 👌 good call! thanks 😁

DrJosh9000 (approved):
Love it!

PR #3372, opened 2025-07-03
if_changed fixes

DrJosh9000:
### Description

- Support older Git versions by invoking `git merge-base` separately rather than using `--merge-base`.
- Including the glob pattern in the skip message would be nice for debugging, but unfortunately it easily exceeds the 70 character limit on the skip message.

### Context

Some additional ad-hoc testing.

### Testing


moskyb (approved):
LGTM!

PR #3381, opened 2025-07-09
Fix git mirrors + refspec

sj26:
### Description

This adjusts the git mirror implementation to respect refspecs.

### Context

We have a customer using refspec with git mirrors and it's not working as expected.

### Changes

Amp gave me a hand making an AGENT.md and then making the change:

https://ampcode.com/threads/T-a695e9fa-78aa-4f81-b488-b9f4f6162a35

Reading through the commits, it seems to have done a pretty reasonable job :+1:

### Testing

Fixes PIPE-1366.

### DIFF
--- a/internal/job/checkout.go
+++ b/internal/job/checkout.go
@@ -393,6 +393,7 @@ func (e *Executor) updateGitMirror(ctx context.Context, repository string) (stri
 
 	if isMainRepository {
 		var refspecs []string
+		var retry bool
 
 		switch {
 		case e.RefSpec != "":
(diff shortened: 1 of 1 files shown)


internal/job/checkout.go
```diff
@@ -392,31 +392,31 @@ func (e *Executor) updateGitMirror(ctx context.Context, repository string) (stri
...
+			refspecs = []string{e.Branch}
+		}
+
+		// Fetch the refspecs from the upstream repository into the mirror.
+		if err := gitFetch(ctx, gitFetchArgs{
+			Shell:      e.shell,
+			GitFlags:   fmt.Sprintf("--git-dir=%s", mirrorDir),
+			Repository: "origin",
+			RefSpecs:   refspecs,
+			Retry:      e.PullRequest != "false" && strings.Contains(e.PipelineProvider, "github"),
```
DrJosh9000: Rather than recomputing whether this is a PR head here, perhaps we can do the same thing as it does with `refspecs` and pull out a `retry` bool?

```go
var retry bool
...

case e.PullRequest != "false" && strings.Contains(e.PipelineProvider, "github"):
	retry = true
	...
```

```suggestion
			Retry:      retry,
```

DrJosh9000 (commented):
LGTM, one small suggestion and be sure to run go fmt

internal/job/checkout.go
sj26: Yeah, nice idea

DrJosh9000 (approved):
✅

sj26:
Thanks @DrJosh9000!

sj26:
I always forget to run `go fmt`. I've added a [lefthook](https://github.com/evilmartians/lefthook) so I won't forget again.

PR #3387, opened 2025-07-14
if_changed improvements

DrJosh9000:
### Description

Make improvements to `if_changed`

- Enable `apply-if-changed` by default
- Lazy-load changed paths when `if_changed` is detected
- Log more error output when problems happen
- Log the diff base, merge-base hash, and the number of changed files on the happy path

### Context

https://linear.app/buildkite/issue/PIPE-1385

### Testing
- [x] Tests have run locally (with `go test ./...`). Buildkite employees may check this if the pipeline has run automatically.
- [x] Code is formatted (with `go fmt ./...`)

PR #3396, opened 2025-07-15
Use BUILDKITE_PIPELINE_DEFAULT_BRANCH as a default git diff base

DrJosh9000:
### Description

Make the defaults more useful. `BUILDKITE_PIPELINE_DEFAULT_BRANCH` is usually set, and `BUILDKITE_PULL_REQUEST_BASE_BRANCH` is usually empty.
But when `BUILDKITE_PULL_REQUEST_BASE_BRANCH` is non-empty we probably want that instead. 

### Context

From this comment: https://github.com/buildkite/agent/pull/3387#discussion_r2204211893

### Changes

Use `BUILDKITE_PIPELINE_DEFAULT_BRANCH` as the last of 3 env vars to use for `--git-diff-base`. 

### Testing

### DIFF
--- a/clicommand/pipeline_upload.go
+++ b/clicommand/pipeline_upload.go
@@ -73,6 +73,8 @@ Example:
 
 const ifChangedSkippedMsg = "if_changed pattern did not match any paths changed in this build"
 
+const defaultGitDiffBase = "origin/main"
+
 type PipelineUploadConfig struct {
 	GlobalConfig
 	APIConfig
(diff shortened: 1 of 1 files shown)


sj26 (approved):
Nice.

DrJosh9000:
I've just tweaked it based on experiments. Because of local repo caching/reuse, `main` can be a bit older than expected - `origin/main` works better. I think the same prefixing should apply to `BUILDKITE_PULL_REQUEST_BASE_BRANCH` and `BUILDKITE_PIPELINE_DEFAULT_BRANCH`. But then the urfave/cli env-var defaulting just doesn't work.

I also realised the log got diff-base and merge-base backwards ("abc123", the merge-base, is a commit that is an ancestor of the diff-base "origin/foo" and HEAD), so fixed that.

PR #3398, opened 2025-07-17
PS-906: configurable kubernetes-bootstrap connection timeout

zhming0:
### Description

Allow `kubernetes-bootstrap`'s connection step timeout to be configurable via a `--kubernetes-bootstrap-connection-timeout` integer flag.

This is helpful in network unstable environment. 

### Context

part of PS-906

### Testing

### DIFF
--- a/clicommand/kubernetes_bootstrap.go
+++ b/clicommand/kubernetes_bootstrap.go
@@ -28,7 +28,8 @@ This command is used internally by Buildkite Kubernetes jobs. It is not
 intended to be used directly.`
 
 type KubernetesBootstrapConfig struct {
-	KubernetesContainerID int `cli:"kubernetes-container-id"`
+	KubernetesContainerID                int           `cli:"kubernetes-container-id"`
+	KubernetesBootstrapConnectionTimeout time.Duration `cli:"kubernetes-bootstrap-connection-timeout"`
 
 	// Global flags for debugging, etc
 	LogLevel    string   `cli:"log-level"`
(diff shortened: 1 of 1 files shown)


clicommand/global.go
```diff
@@ -115,6 +115,14 @@ var (
...
 	}
 
+	KubernetesBootstrapConnectionTimeoutFlag = cli.IntFlag{
+		Name: "kubernetes-bootstrap-connection-timeout",
+		Usage: "This is intended to be used only by the Buildkite k8s stack " +
+			"(github.com/buildkite/agent-stack-k8s); it set the max time a container will wait " +
+			"to connect Agent.",
+		EnvVar: "BUILDKITE_KUBERNETES_BOOTSTRAP_CONNECTION_TIMEOUT",
+	}
+
```
DrJosh9000: We've generally not put command-specific flags in `global.go`. `KubernetesContainerIDFlag` is currently shared with `bootstrap.go` and `kubernetes_bootstrap.go` which is why it's here. 

(At some point we should remove the legacy kubernetes-exec support from within bootstrap.go.)

clicommand/global.go
```diff
@@ -115,6 +115,14 @@ var (
 		EnvVar: "BUILDKITE_CONTAINER_ID",
 	}
 
+	KubernetesBootstrapConnectionTimeoutFlag = cli.IntFlag{
+		Name: "kubernetes-bootstrap-connection-timeout",
```
DrJosh9000: Is this in seconds?

Should this be a `cli.DurationFlag`?

duckalini:
CI is failing because the [updated pipeline.yml](https://github.com/buildkite/agent/blob/dcd1e6af5381427bc4627a3615cafb6a1e0b74be/.buildkite/pipeline.yml#L252-L257) from `main` is missing - merge in main and CI should pass again.

PR #3401, opened 2025-07-18
Rename local -> repository hooks, global -> agent hooks

moskyb:
### Description

In the docs, we say:
image

but then in job logs, we say:
image

also, `local` and `global` as names are pretty meaningless. `agent` and `repository` are much clearer, so this PR updates the the hook scopes to be formally named `agent` and `repository`.

There's a wee punchout for the datadog tracing implementation, which expects hook scopes to be called the old names in span attributes, so we've retained them there and there only.

**Before:**
CleanShot 2025-07-18 at 11 22 18@2x

**After**
CleanShot 2025-07-18 at 11 22 45@2x

### Context

https://buildkite-corp.slack.com/archives/C07LSRYCPSR/p1752753127654479

### Testing

### DIFF
--- a/internal/job/executor.go
+++ b/internal/job/executor.go
@@ -314,6 +314,12 @@ func (e *Executor) Cancel() error {
 	return nil
 }
 
+const (
+	HookScopeAgent      = "agent"
+	HookScopeRepository = "repository"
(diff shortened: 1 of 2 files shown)


zhming0 (approved):
Looks good.

PR #3402, opened 2025-07-18
Support ignoring env vars in job signatures

moskyb:
### Description

In some cases, when using signed pipelines, the values of some environment variables can't necessarily be controlled by the user to the degree that steps will always pass verification -- for example, successive pipeline uploads can change build-level environment such that steps will fail to verify.

To allow users to still use pipeline signing in these situations, we've added a pair of flags to the agent `--signing-ignored-env-vars` and `--verification-ignored-env-vars`. As their names imply, they remove environment variables from the step payloads that we sign and verify. 

This means that if we knew that environment variables called `MOUNTAIN` and `RIVER` were going to change between signing and verification, we could start the signing agents with `--signing-ignored-env-vars="MOUNTAIN,RIVER"` and the verifying agents with `--verification-ignored-env-vars="MOUNTAIN,RIVER"`. 

Setting these flags reduces the integrity of the signature, and thus weakens the security guarantees that signed pipelines provides. As such, when in use, signing pipelines with ignored env vars will always emit a warning. The intent is that the use of these flags will be a stopgap until agent environments can be better controlled.

This PR relies on the associated one in [go-pipeline](https://github.com/buildkite/go-pipeline/pull/67).

### Testing

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -94,12 +94,14 @@ type AgentStartConfig struct {
 
 	SigningJWKSKeyID string `cli:"signing-jwks-key-id"`
 
-	SigningJWKSFile  string `cli:"signing-jwks-file" normalize:"filepath"`
-	SigningAWSKMSKey string `cli:"signing-aws-kms-key"`
-	DebugSigning     bool   `cli:"debug-signing"`
+	SigningJWKSFile       string   `cli:"signing-jwks-file" normalize:"filepath"`
+	SigningAWSKMSKey      string   `cli:"signing-aws-kms-key"`
+	DebugSigning          bool     `cli:"debug-signing"`
+	SigningIgnoredEnvVars []string `cli:"signing-ignored-env-vars" normalize:"list"`
 
-	VerificationJWKSFile        string `cli:"verification-jwks-file" normalize:"filepath"`
-	VerificationFailureBehavior string `cli:"verification-failure-behavior"`
+	VerificationIgnoredEnvVars  []string `cli:"verification-ignored-env-vars" normalize:"list"`
+	VerificationJWKSFile        string   `cli:"verification-jwks-file" normalize:"filepath"`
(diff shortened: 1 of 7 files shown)


go.mod
```diff
@@ -16,7 +16,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/kms v1.41.2
 	github.com/brunoscheufler/aws-ecs-metadata-go v0.0.0-20220812150832-b6b31c6eeeaf
 	github.com/buildkite/bintest/v3 v3.3.0
-	github.com/buildkite/go-pipeline v0.14.0
+	github.com/buildkite/go-pipeline v0.14.1-0.20250718044939-1ded6233e93a
```
DrJosh9000, line 19: I assume you intend to do a go-pipeline release before merging?

moskyb: yes! very much the plan

clicommand/agent_start.go
```diff
@@ -724,6 +726,16 @@ var AgentStartCommand = cli.Command{
...
 			EnvVar: "BUILDKITE_AGENT_DEBUG_SIGNING",
 		},
+		cli.StringSliceFlag{
+			Name:   "signing-ignored-env-vars",
+			Usage:  "A list of environment variable names to ignore when signing the pipeline. These variables (and their values) will not be included in the signature. For signatures to match, the corresponding flag must be set on the agent that verifies the signature and runs the job",
+			EnvVar: "BUILDKITE_AGENT_SIGNING_IGNORED_ENV_VARS",
+		},
+		cli.StringSliceFlag{
+			Name:   "verification-ignored-env-vars",
+			Usage:  "A list of environment variable names to ignore when signing the pipeline. These variables (and their values) will not be included in the signature. For signatures to match, the corresponding flag must be set on the agent that verifies the signature and runs the job",
```
DrJosh9000, line 736: This is the same usage string as the other new flag? Should it match the comment in agent_configuration.go?

PR #3403, opened 2025-07-18
Exit with a specific code when the Job being Acquired is Locked

CerealBoy:
### Description

When a Job fails to be Acquired, if that Job is locked due to dependencies then retry for a short time. If this wait continues for too long, exit with a unique code that will allow a scheduler to take specific behaviour for that scenario. There will be another webhook once the Job has moved back to Scheduled so the Job won't be lost.

### Context

At present a scheduler is given no indication on whether the failed acquisition was for any particular purpose, and in this scenario where the job is locked the behaviour can be better handled with a new code.

### Changes

- Exit with a unique code (`28`) when the Job being Acquired is Locked.
- Pepper the error message generated with the Locked detail to surface this in the Job log.
- Add test cases to handle the expected error codes for different Job statuses.

### Testing

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -1307,6 +1307,13 @@ var AgentStartCommand = cli.Command{
 			const acquisitionFailedExitCode = 27 // chosen by fair dice roll
 			return cli.NewExitError(err, acquisitionFailedExitCode)
 		}
+		if errors.Is(err, core.ErrJobLocked) {
+			// If the agent tried to acquire a job, but it couldn't because the job is locked (waiting for dependencies),
+			// we should exit with a specific exit code so that the caller can know that this job is locked.
+
+			const jobLockedExitCode = 28
+			return cli.NewExitError(err, jobLockedExitCode)
+		}
 		if exit := new(core.ProcessExit); errors.As(err, exit) {
 			if cfg.ReflectExitStatus {
 				// If the agent acquired a job and it failed or was cancelled,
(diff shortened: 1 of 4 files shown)


core/client.go
```diff
@@ -155,7 +159,7 @@ func handleRetriableJobAcquisitionError(warning string, resp *api.Response, r *r
 	r.SetNextInterval(duration)
 }
 
-// Connects the agent to the Buildkite Agent API, retrying up to 10 times with 5
+// Connect the agent to the Buildkite Agent API, retrying up to 10 times with 5
```
DrJosh9000:
```suggestion
// Connect connects the agent to the Buildkite Agent API, retrying up to 10 times with 5
```

clicommand/agent_start.go
```diff
@@ -1307,6 +1307,13 @@ var AgentStartCommand = cli.Command{
 			const acquisitionFailedExitCode = 27 // chosen by fair dice roll
 			return cli.NewExitError(err, acquisitionFailedExitCode)
 		}
+		if errors.Is(err, core.ErrJobLocked) {
```
DrJosh9000: Two sequential `if`s checking the same variable: fine, maybe even good
Three: time to consider `switch`. 

The main downside is creating the `core.ProcessExit` outside its case.

```go
		var exit core.ProcessExit
		switch err := pool.Start(ctx); {
		case errors.Is(err, core.ErrJobAcquisitionRejected):
			...
			return cli.NewExitError(err, acquisitionFailedExitCode)

		case errors.Is(err, core.ErrJobLocked): 
			...
			return cli.NewExitError(err, jobLockedExitCode)
		
		if errors.As(err, &exit) {
			...
			return nil

		default:
			return err
		}
		
```

DrJosh9000 (approved):
LGTM

CerealBoy:
Thank you very much @DrJosh9000! I've pushed a commit with the changes from your comments, do you have a moment to confirm this is suitable please?

PR #3407, opened 2025-07-22
Only run tests if code has changed

DrJosh9000:
### Description

Reduce the number of steps that run if no Go code (or their non-code dependencies) has changed.

### Context

Dogfood `if_changed` support

### Changes

- Group Tests and Coverage
- Make Lint step only run if code or the script that runs it has changed
- Make Tests and Coverage group only run if code or related scripts changed

### Testing

Pipeline change only - but particularly, these are _not_ run because the code didn't change!

- ~~[] Tests have run locally (with `go test ./...`). Buildkite employees may check this if the pipeline has run automatically.~~ 
- ~~[] Code is formatted (with `go fmt ./...`)~~

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -13,6 +13,7 @@ steps:
   - name: ":go::robot_face: Lint"
     key: check-code-committed
     command: .buildkite/steps/check-code-committed.sh
+    if_changed: "{go.mod,go.sum,**.go,.buildkite/steps/check-code-committed.sh}"
     plugins:
       - docker-compose#v4.14.0:
           config: .buildkite/docker-compose.yml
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline.yml
```diff
@@ -13,160 +13,166 @@ steps:
...
-      - test-linux-amd64
-      - test-race-linux-arm64
-      - test-linux-arm64
-      - test-windows
-    allow_dependency_failure: true
-    plugins:
-      - junit-annotate#v1.6.0:
-          artifacts: junit-*.xml
+  - group: ":go::scientist: Tests and Coverage"
+    if_changed: "{go.mod,go.sum,**.go,**/fixtures/**,.buildkite/steps/{tests,test-coverage-report}.sh}"
```
catkins, line 25: its great that you can use it at the group level 👨‍🍳 😚

catkins (approved):
Love to see it

.buildkite/pipeline.yml
DrJosh9000: Oh yeah it was my first thought when implementing: "this should totally work on groups and trigger steps too"

zhming0: should `.buildkite/pipeline.yml` be part of the dependencies?

PR #3420, opened 2025-08-07
PS-1000: ensure a static & short checkout path for k8s stack agent

zhming0:
### Description

In earlier k8s stack, we used to set a short `BUILDKITE_AGENT_NAME` variable, it was removed. 

We'd like to bring that behavior back so we always get a short and consistent checkout path. 

### Context

PS-1000

### Testing


DrJosh9000 (approved):
LGTM, nice comment

DrJosh9000 (approved):
Even better!

PR #3429, opened 2025-08-14
Add plugins-always-clone-fresh to config, CLI start

petetomasik:
### Description

While updating docs for the agent's configuration to align with `buildkite-agent start` options, I came across a broken link for a [config option mentioned](https://buildkite.com/docs/pipelines/integrations/plugins/writing#developing-a-plugin-with-a-feature-branch) (`plugins-always-clone-fresh`) that is available to the `bootstrap` command, but not the `start` command.

### Context

[Linear ticket](https://linear.app/buildkite/issue/SUP-3986/add-missing-agent-configuration-settings)

### Changes

This change adds the `plugins-always-clone-fresh` option to the agent configuration and allows configuring as CLI `start` option.

### Testing

Confirmed both `buildkite-agent start --plugins-always-clone-fresh` and `plugins-always-clone-fresh=true` (in `buildkite-agent.cfg`) both result in the expected behavior during `Preparing plugins`:
```
# BUILDKITE_PLUGINS_ALWAYS_CLONE_FRESH is true; removing previous checkout of plugin github.com/buildkite-plugins/docker-buildkite-plugin#v5.13.0
```


zhming0 (approved):
LGTM 👍🏿

moskyb (dismissed):
just a quick countermand to stop automerge, will write something up very soon

moskyb:
hello! quick "i was there when the deep magic was written" moment from me.

originally the purpose of adding this flag to `bootstrap` and not `start` was that always cloning plugins fresh can be and usually is pretty slow, and will generally slow down builds a lot. 

when it was originally added in https://github.com/buildkite/agent/pull/1636, its purpose was to make developing plugins easier, as the agent caching your local plugin checkout when you want it to go and get the latest version is pretty annoying, so we made it a not-especially-obvious escape hatch in `bootstrap` but not in `start`. a relatively advanced user would know to pass `BUILDKITE_PLUGINS_ALWAYS_CLONE_FRESH`.

these decisions were made years ago, and are probably no longer valid -- we're keeping our customers from shooting themselves in the foot, sure, but our general policy these days is to allow that sort of thing, but let people know the risks before doing so.

all this to say: i thought it would be worth putting some context as to why we originally made this decision, but i'm happy to proceed as planned if you like @petetomasik

Issue #213, opened 2015-11-19
Windows: Can't cancel tasks


essen:
When clicking Cancel in the UI, the Windows agent reports:

```
[Process] Failed to send signal: terminated to PID: 6540 (syscall.Errno: not supported by windows)
```

And the build subsequently continues to run.

essen:
How about a temporary solution using `TASKKILL /PID %d`?

keithpitt:
Hrm..the beta version of the agent (v3.0) _should_ be able to cancel tasks in Windows now. Are you using that version by chance?

essen:
I'm on v3.0 beta.12 and it does not cancel my tasks. I'll check if the error message changed since then.

essen:
I don't suppose the Windows build logs its output anywhere? I can't see it in the documentation.

keithpitt:
The build logs should be getting streamed to Buildkite - is that what you mean? Or the actual logs from the agent themselves. We don't log them anywhere, so they'll be coming from STDERR in the agent. How are you running the Agent on Windows?

essen:
The logs from the agent. I'm running it as a Task Scheduler task. Apparently I need to redirect it myself to a file, so I'll do that and try again, see what error I get.

essen:
Yep I still get the same error message as the OP.

kamalaknn:
Still seeing this in 3.0-beta.23. Any updates?

lox:
I believe this is related to https://github.com/buildkite/agent/issues/345. I'll see what I can do.

lox:
This is fixed in 2-6-stable, still need to port it to master / v3.

lox:
Fixed in https://github.com/buildkite/agent/commit/9b8a0b922f312ace8a2df1ef1cbae63e79ef494d

PR #3430, opened 2025-08-18
SUP-4325: Implement mixed-format env file for automatic environment variable propagation

Mykematt:
### Description

This PR implements automatic environment variable propagation for OTEL tracing variables and other agent configuration variables to **any plugin using `propagate-environment: true`**. Agent config variables are written as names-only (allowing containers to pull values from the host environment) and job-specific variables are written as key=value pairs.

**Problem:** OTEL tracing variables (`BUILDKITE_TRACING_*`) and other agent config variables were not automatically available to plugins using `propagate-environment: true`, requiring manual plugin configuration for each variable.

**Solution:** Modified the agent to include these variables in `BUILDKITE_ENV_FILE` before writing, enabling automatic propagation to any plugin that uses `propagate-environment: true` without requiring plugin changes.

**Alternatives considered:** Manual plugin-level propagation, but this requires maintaining variable lists in multiple plugins and doesn't scale well.

### Context

Linear ticket SUP-4325: "[Pinterest/Docker & Docker Compose] Propagate BUILDKITE_TRACING_* Env Vars"

### Changes

- **Modified [agent/job_runner.go](cci:7://file:///Users/olabuildkite/GitHub/agent/agent/job_runner.go:0:0-0:0)** to include 29 agent configuration variables in `BUILDKITE_ENV_FILE`
- **Implemented mixed-format env file:**
  - Agent config variables written as names-only (containers pull from host environment)
  - Job variables written as key=value pairs
- **Added OTEL tracing variables** to automatic propagation:
  - `BUILDKITE_TRACING_BACKEND`
  - `BUILDKITE_TRACING_SERVICE_NAME`
  - `BUILDKITE_TRACING_PROPAGATE_TRACEPARENT`
  - `BUILDKITE_TRACING_TRACEPARENT`
  - `BUILDKITE_TRACE_CONTEXT_ENCODING`
- **Maintained security boundaries** by excluding 23 sensitive variables (tokens, agent paths, internal state)
- **Backward compatible** - existing plugin configurations continue to work unchanged

**Impact:** Any plugin using `propagate-environment: true` now automatically receives 92 variables instead of requiring manual configuration for each one.

### Testing
- [x] Real-world validation with Docker plugin v5.11.0 and Docker Compose plugin v4.16.0
- [x] Verified 92 variables automatically propagate in live pipeline execution (build [#213](https://buildkite.com/olabuildkite/plugin/builds/213))


DrJosh9000:
Thanks @Mykematt. This sounds like a backwards-incompatible change for anyone parsing the env file with the expectation that every variable has a value. Is that right?

DrJosh9000 (changes requested):
I think that this cannot be merged as-is, for two reasons. 

One is the backwards compatibility with user scripts or programs that parse the env file. While they can now migrate to the better-defined JSON version of the file, this requires a recent agent version and work on their part to reimplement. They could just as easily adapt to the additional valueless env vars in the file, but either way we don't know whose builds we will break by doing this.

Two is the change to propagated environment variables itself. A user could have a pipeline set up with the Docker and Docker Compose plugins case (and any other plugin that might be using the env file), and, without making any change in version to their plugins, be surprised when they upgrade their agent version, with a bunch of extra env vars in their jobs. They may be relying on certain env vars _not_ existing. 

Both the Docker and Docker Compose plugin Readmes specifically call out:

> Important: only pipeline environment variables will be propagated

and these extra env vars aren't "pipeline environment variables" _per se_, and that may be something someone is relying on.

To merge this, I would suggest either:

- writing an additional file that could be read by a future release of the Docker and Docker Compose plugins, or
- writing them to the existing env file under an experiment (see `internal/experiments`).

Mykematt:
@DrJosh9000 thanks for the in-depth review. I have implemented the second approach you mentioned.

agent/job_runner.go
```diff
@@ -452,17 +452,121 @@ func (r *JobRunner) createEnvironment(ctx context.Context) ([]string, error) {
...
+				env["BUILDKITE_TRACING_TRACEPARENT"] = r.conf.Job.TraceParent
+			}
+			if r.conf.AgentConfiguration.TracingPropagateTraceparent {
+				env["BUILDKITE_TRACING_PROPAGATE_TRACEPARENT"] = "true"
+			}
+		}
+	}
+
+	// Pipeline signing variables - conditionally set based on experiment flag
+	if experiments.IsEnabled(ctx, experiments.PropagateAgentVars) {
```
DrJosh9000: This `if` condition is the same as on line 456, so they could be combined into a single `if`.

agent/job_runner.go
```diff
@@ -452,17 +452,121 @@ func (r *JobRunner) createEnvironment(ctx context.Context) ([]string, error) {
...
+			env["BUILDKITE_AGENT_JWKS_FILE"] = r.conf.AgentConfiguration.SigningJWKSFile
+		}
+		if r.conf.AgentConfiguration.SigningJWKSKeyID != "" {
+			env["BUILDKITE_AGENT_JWKS_KEY_ID"] = r.conf.AgentConfiguration.SigningJWKSKeyID
+		}
+	}
+
+	// Variables to write as names-only to env file (Docker will pull values from agent environment)
+	var envFileNameOnlyVars []string
+	if experiments.IsEnabled(ctx, experiments.PropagateAgentVars) {
```
DrJosh9000, line 504: Similarly, this is the same condition again.

agent/job_runner.go
```diff
@@ -539,29 +643,10 @@ func (r *JobRunner) createEnvironment(ctx context.Context) ([]string, error) {
...
-	env["BUILDKITE_GIT_CLONE_MIRROR_FLAGS"] = r.conf.AgentConfiguration.GitCloneMirrorFlags
-	env["BUILDKITE_GIT_CLEAN_FLAGS"] = r.conf.AgentConfiguration.GitCleanFlags
-	env["BUILDKITE_GIT_MIRRORS_LOCK_TIMEOUT"] = strconv.Itoa(r.conf.AgentConfiguration.GitMirrorsLockTimeout)
-	env["BUILDKITE_SHELL"] = r.conf.AgentConfiguration.Shell
 	env["BUILDKITE_AGENT_EXPERIMENT"] = strings.Join(experiments.Enabled(ctx), ",")
-	env["BUILDKITE_REDACTED_VARS"] = strings.Join(r.conf.AgentConfiguration.RedactedVars, ",")
-	env["BUILDKITE_STRICT_SINGLE_HOOKS"] = fmt.Sprint(r.conf.AgentConfiguration.StrictSingleHooks)
-	env["BUILDKITE_CANCEL_GRACE_PERIOD"] = strconv.Itoa(r.conf.AgentConfiguration.CancelGracePeriod)
-	env["BUILDKITE_SIGNAL_GRACE_PERIOD_SECONDS"] = strconv.Itoa(int(r.conf.AgentConfiguration.SignalGracePeriod / time.Second))
-	env["BUILDKITE_TRACE_CONTEXT_ENCODING"] = r.conf.AgentConfiguration.TraceContextEncoding
```
DrJosh9000, line 564: Env vars that existed prior to this PR must remain in the env - none of these should be deleted, they shouldn't be conditional on the experiment.

agent/job_runner.go
```diff
@@ -587,27 +672,12 @@ func (r *JobRunner) createEnvironment(ctx context.Context) ([]string, error) {
...
-
-	// Pass signing details through to the executor - any pipelines uploaded by this agent will be signed
-	if r.conf.AgentConfiguration.SigningJWKSFile != "" {
-		env["BUILDKITE_AGENT_JWKS_FILE"] = r.conf.AgentConfiguration.SigningJWKSFile
-	}
-
-	if r.conf.AgentConfiguration.SigningJWKSKeyID != "" {
-		env["BUILDKITE_AGENT_JWKS_KEY_ID"] = r.conf.AgentConfiguration.SigningJWKSKeyID
-	}
-
```
DrJosh9000, line 603: Similarly, these should remain

agent/job_runner.go
```diff
@@ -616,22 +686,6 @@ func (r *JobRunner) createEnvironment(ctx context.Context) ([]string, error) {
...
-		// which can be propagated to the job tracing if OpenTelemetry is used
-		//
-		// https://www.w3.org/TR/trace-context/#traceparent-header
-		if r.conf.Job.TraceParent != "" {
-			env["BUILDKITE_TRACING_TRACEPARENT"] = r.conf.Job.TraceParent
-		}
-		if r.conf.AgentConfiguration.TracingPropagateTraceparent {
-			env["BUILDKITE_TRACING_PROPAGATE_TRACEPARENT"] = "true"
-		}
-	}
```
DrJosh9000, line 633: And this too

DrJosh9000:
@Mykematt Thanks, sorry for the delay in responding. 

This still has problems. If the experiment is not enabled, then all the agent config vars that are not present in the env file today, but are unconditionally added to the map before writing the file, would be written to the file. Ideally the file would be 100% unchanged unless the experiment is active.

Rather than go back and forth again, I've taken a stab at implementing this myself here: #3471. Would you mind testing it to ensure it suits your purposes?

PR #3431, opened 2025-08-19
Allow multiple pipeline files for upload

DrJosh9000:
### Description

Like how we now accept multiple pipelines within a single file or stdin stream, let's accept multiple files passed as args.

### Context

https://linear.app/buildkite/issue/PS-978

### Changes

- Change the loader to look for `arg:*`
- Process all files passed as non-flag args

### Testing

### DIFF
--- a/clicommand/pipeline_upload.go
+++ b/clicommand/pipeline_upload.go
@@ -11,7 +11,6 @@ import (
 	"maps"
 	"os"
 	"os/exec"
-	"path"
 	"path/filepath"
 	"slices"
 	"strings"
@@ -79,7 +78,7 @@ type PipelineUploadConfig struct {
 	GlobalConfig
 	APIConfig
 
-	FilePath        string   `cli:"arg:0" label:"upload paths"`
+	FilePaths       []string `cli:"arg:*" label:"upload paths"`
 	Replace         bool     `cli:"replace"`
 	Job             string   `cli:"job"` // required, but not in dry-run mode
 	DryRun          bool     `cli:"dry-run"`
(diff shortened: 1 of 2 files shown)


clicommand/pipeline_upload.go
```diff
@@ -176,27 +175,31 @@ var PipelineUploadCommand = cli.Command{
...
-			filename = filepath.Base(cfg.FilePath)
-			file, err := os.Open(cfg.FilePath)
-			if err != nil {
-				return fmt.Errorf("failed to read file: %w", err)
+			for _, fn := range cfg.FilePaths {
+				file, err := os.Open(fn)
+				if err != nil {
+					return fmt.Errorf("failed to read file: %w", err)
+				}
+				defer file.Close()
```
CerealBoy, line 194: With the `defer` within the loop, should the loop content be wrapped within `func() {` to properly ensure each file handle is closed correctly?

DrJosh9000: Yes!

DrJosh9000: Wait, no! The files aren't actually used until a couple of loops further down.

CerealBoy: Hmmmm will the defer still then keep the reference correctly for each of the individual files that have been opened? For some reason I was thinking that only the last loop would have the correct defer retained.

DrJosh9000: `file` is declared within the loop body, and so is the defer, so there should be N `defer`s and each should capture a reference to a different variable. This should be easy to test on go.dev/play...

DrJosh9000: https://go.dev/play/p/tijE6PJW00m

CerealBoy: Perfect, nothing to see here then!

PR #3434, opened 2025-08-25
Update EC2 tags/metadata to use AWS Go SDK v2

DrJosh9000:
### Description

Upgrade

### Context

Noticed while looking at linter output that AWS Go SDK v1 is now no longer supported.

### Changes

Update to the v2 SDK. This only updates calls within the agent package, other packages to come soon.

### Testing

### Disclosures / Credits


moskyb (approved):
🤠

agent/ec2_tags.go
```diff
@@ -1,51 +1,59 @@
...
+	}
+
+	instanceID, err := io.ReadAll(mdOut.Content)
 	if err != nil {
-		return tags, err
+		return nil, fmt.Errorf("reading instance ID from metadata: %w", err)
 	}
 
-	svc := ec2.New(sess)
+	svc := ec2.New(ec2.Options{})
```
rahdjoudj, line 38: don't we need to pass the config here ? to it can access the instance tags ?
smth like svc := ec2.NewFromConfig(cfg)

PR #3444, opened 2025-08-27
Read env after running unwrapped hook

francoiscampbell:
### Description
Certain runtime agent configs are not updated after running non-shell hooks. This PR calls `ExecutorConfig.ReadFromEnvironment` after the hook to update from the latest environment.

### Context
When running a shell hook, `ExecutorConfig.ReadFromEnvironment` is called as part of `Executor.runWrappedShellScriptHook`, so changes to env vars that affect the agent's behaviour (e.g. `BUILDKITE_REPO`) are properly reflected in the agent's internal state.

When running non-shell hooks (Python scripts, compiled binaries, etc), the hook is run through `Executor.runUnwrappedHook`, which doesn't capture env changes after running the hook (the Job API is used to set environment variables).

### Changes
- Adds an explicit call to `ExecutorConfig.ReadFromEnvironment` in `Executor.runUnwrappedHook` after the hook exits successfully

### Testing

TODO: add a new test for this

### DIFF
--- a/internal/job/executor.go
+++ b/internal/job/executor.go
@@ -432,7 +432,16 @@ func (e *Executor) runUnwrappedHook(ctx context.Context, _ string, hookCfg HookC
 	environ.Set("BUILDKITE_HOOK_PATH", hookCfg.Path)
 	environ.Set("BUILDKITE_HOOK_SCOPE", hookCfg.Scope)
 
-	return e.shell.Command(hookCfg.Path).Run(ctx, shell.WithExtraEnv(environ))
+	if err := e.shell.Command(hookCfg.Path).Run(ctx, shell.WithExtraEnv(environ)); err != nil {
+		return err
+	}
+
+	if e.Debug {
+		e.shell.Commentf("Refreshing executor config to reflect changes made through the job API")
+	}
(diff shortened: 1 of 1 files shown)


zhming0:
Hi @francoiscampbell, I take this PR as both a bug report and a bug fix? 🙂 

The fact that we don't capture environment variable changes in non-shell script hook is an expected behavior, it's documented in [polyglot hook caveats](https://buildkite.com/docs/agent/v3/hooks#polyglot-hooks-caveats). 

The longer version is that: we can't do a `source foo.py` so we can't capture environment variables change in a non-shell script hook. 

In this case, you can use `buildkite-agent env set ...` to propagate environment variable changes.

francoiscampbell:
@zhming0 The issue isn't that env is not captured, it's that when you run a script hook like
```bash
#!/usr/bin/env bash
export BUILDKITE_REPO=
```
it correctly skips checkout but if you run an interpreted hook like
```python
#!/usr/bin/env python
import subprocess
subprocess.run(["buildkite-agent", "env", "set", "BUILDKITE_REPO="])
```
it doesn't skip the checkout.

@SorchaAbel Yeah I feel like this merits a test but I'm not sure where to add it yet.

PR #3445, opened 2025-08-28
Restrict plugin and hook symlinks

DrJosh9000:
### Description

Restrict symlinking and other path shenanigans from local hooks and plugins to within the relevant checkout directory.

### Context

https://linear.app/buildkite/issue/PS-1095

### Changes

- Add a `checkoutRoot` (of type `*os.Root`) to `Executor`, set during the checkout phase.
- Change how `pluginCheckout` works to include a `Root` (of type `*os.Root`), and refer to that root when finding hooks or validating the plugin.
- Make vendored plugins use the executor's `checkoutRoot`. Check that the vendored plugin directory and its hooks directory exist in the checkout root.
- Make non-vendored plugins use a `checkoutRoot` opened after the plugin is established to exist or is checked out. Check that the hooks directory exist in the checkout root.
- Change `hook.Find` to find hooks relative to a root, if one is provided.
- Provide the checkout root to `hook.Find` for local hooks, normal plugins, and vendored plugins.

### Testing

### Disclosures / Credits


internal/job/hook/hook.go
```diff
@@ -8,24 +8,45 @@ import (
...
+		// If we're not on Windows, then no matter what bits are set, we do
+		// further inspection later on to figure out how best to run it.
+
+		// Prepend the root's path to the relative hook dir path, if we have
+		// one.
+		if root != nil {
+			return filepath.Join(root.Name(), p), nil
+		}
 		return p, nil
 	}
```
zhming0, line 48: Are these conceptually equivalent to `shell.LookPath(name, root + hookDir, exts)`?

Or maybe because LookPath wouldn't respect the `root` so we don't use it?

internal/job/hook/hook.go
```diff
@@ -8,24 +8,45 @@ import (
...
+
+	// Check for a file named name+ext in hookDir.
+	for _, ext := range exts {
+		p := filepath.Join(hookDir, name+ext)
+		if fi, err := stat(p); err != nil || fi.IsDir() {
+			continue
+		}
+		// If we're on Windows, there are no executable perms bits to check.
+		// If we're not on Windows, then no matter what bits are set, we do
+		// further inspection later on to figure out how best to run it.
```
zhming0, line 40: What is this comment referring to? I don't see where we do the "further inspection"?

internal/job/checkout.go
```diff
@@ -74,17 +80,30 @@ func (e *Executor) removeCheckoutDir() error {
...
 		if err := os.MkdirAll(checkoutPath, 0o777); err != nil {
 			return err
 		}
 	}
 
+	root, err := os.OpenRoot(checkoutPath)
+	if err != nil {
+		return fmt.Errorf("opening checkout path as root: %w", err)
+	}
+	// This cleanup is largely ornamental, since the executor pointer only
```
zhming0, line 102: [non-blocking] I have two questions mainly for my curiosity: 

- If it's truly ornamental, maybe we can remove it? (but yeah it feels bad opening a file without closing it) 
- Is it truly ornamental though, could a Agent checkout multiple times with different executor pointers?

internal/job/executor.go
```diff
@@ -717,8 +720,9 @@ func (e *Executor) executeGlobalHook(ctx context.Context, name string) error {
 
 // Returns the absolute path to a local hook, or os.ErrNotExist if none is found
 func (e *Executor) localHookPath(name string) (string, error) {
-	dir := filepath.Join(e.shell.Getwd(), ".buildkite", "hooks")
-	return hook.Find(dir, name)
+	// The local hooks dir must exist within the checkout root.
+	dir := filepath.Join(".buildkite", "hooks")
+	return hook.Find(e.checkoutRoot, dir, name)
```
zhming0, line 751: I am not entirely following this logic, if our intention is to limit the `hook.Find` to work within a directory, doesn't changing `hook.Find` implementation suffice?

zhming0 (commented):
I see where this is going but I have some questions 🙏🏿

internal/job/hook/hook.go
DrJosh9000: Conceptually, yes.

`LookPath` isn't a great fit for this task. The primary function of `LookPath` is to look through multiple directories (i.e. the `$PATH` env var) to find the first executable with a given name. Here we only need to search for a file in a single path, on Windows with different possible extensions. As far as I can tell, it was only ever used here only for Windows in order to re-use the inner loop which loops over the extensions.

I'd rather not have to modify `LookPath` in order to take a root, and there's at least one concerning quirk in the implementation for Windows. The implementation here is fairly short and clear.

internal/job/hook/hook.go
DrJosh9000: This is referring to the polyglot hook/shebang line logic in the hook runner. The point here is just that the hook need not have the executable bit to be considered a hook.

internal/job/checkout.go
DrJosh9000: The bootstrap is never going to have more than one executor (pointer), but the tests might. So it's not actually ornamental. Perhaps this should be restructured so that `Executor` has a `Close` / `Cleanup` function that the caller can manage. 🤔

internal/job/executor.go
DrJosh9000: I'm not sure I understand... `hook.Find` is changed from looking for a file within an absolute path, to looking for a file in a relative path in a root. So I had to figure out what root to pass to it in place of `e.shell.Getwd()`. And at this point in the executor, `e.shell.Getwd()` should always be the checkout directory for the job. So if anything, the previous code was obscuring the fact that `.buildkite/hooks` is relative to the checkout directory.

internal/job/checkout.go
zhming0: ah good point, the caller probably shouldn't be concerned about Executor cleanup.

internal/job/executor.go
zhming0: I understand it now. Reading it again, I understand that this is to prevent: 

```yaml
plugins:
  - file:///../../../my-plugin.git#v1.0.0: ~
```

is it right?

DrJosh9000: Well...

In this particular function, we're dealing with "local hooks", which are hooks that may or may not exist in the repo being built. The agent has always looked for these at the relative path `.buildkite/hooks`. There's no user-provided path that could contain `..` syntax, but the potential escape is that `.buildkite` or `hooks` could be symlinks out of the repo.

For plugins, there are two cases. Plugins can exist in their own repo (see e.g. github.com/buildkite-plugins/..., which is what I consider the "regular" kind of plugin) or plugins can be _vendored_ into the repo being built. Vendored plugins are always prefixed with `.`, possibly also with `file://` because they are parsed first as a URL. Vendored plugin paths are relative to the repo being built. So in theory they could be escaped with `..` syntax, but the original report related to symlinks as well.

Finally, regular plugins should be confined to their own repo in the same way as local hooks and vendored plugins should be confined to the repo being built.

zhming0: Ahhh right, now I finally get it. Directory traversal via symlink 👍🏿

PR #3447, opened 2025-08-30
[PB-569] Expose TRACEPARENT to jobs

catkins:
### Testing


catkins:
@dannyfallon yep it can, thanks!

PR #3450, opened 2025-09-05
Automatically redact OIDC tokens

matthewborden:
OIDC tokens are now automatically redacted from build logs by default, with an optional --skip-redaction flag to disable this behavior when needed.

```
$ buildkite-agent oidc request-token
[REDACTED]
```

```
$ buildkite-agent oidc request-token --skip-redaction
eyJra....
```

This uses the same mechanisms as `buildkite-agent secret get` for redacting the token.

### Context

Fixes PS-1128

### Testing

### Disclosures / Credits

Amp wrote the change with this prompt:

We want to implement auto-redaction for OIDC tokens. `buildkite-agent secret get <SECRET_NAME>` already supports auto-redaction. We want to extend this to `buildkite-agent oidc request-token`. This should work the same as secret redaction where the token is automatically redacted by the CLI command and does not require manual redaction. Add a skip-redaction flag. Read agent/clicommand/secret_get.go for how secrets are redacted.

### DIFF
--- a/clicommand/oidc_request_token.go
+++ b/clicommand/oidc_request_token.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/buildkite/agent/v3/api"
+	"github.com/buildkite/agent/v3/jobapi"
 	"github.com/buildkite/roko"
 	"github.com/urfave/cli"
 )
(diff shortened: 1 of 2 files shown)


wolfeidau:
@matthewborden is there a reason to every skip redaction? I think we can cater for that if someone requests it.

wolfeidau (approved):
👍🏻 🚀

PR #3453, opened 2025-09-08
Support `secrets` pipeline syntax

matthewborden:
### Description

This pull request implements support for `secrets` pipeline syntax. Cluster secrets can be specified within pipeline.yml and will then be automatically fetched and injected into the job environment. This partially replaces the https://buildkite.com/resources/plugins/buildkite-plugins/secrets-buildkite-plugin/. 

Basic usage:
```
steps:
  - command: "deploy.sh"
     secrets:
       - DATABASE_URL
       - API_TOKEN
```

Pipeline and step secrets merge together:
```
secrets:
  - DATABASE_URL
  - REDIS_URL

steps:
  - command: "test.sh"
    secrets:
      - API_TOKEN        # Step-specific secret
      - DATABASE_URL

  - command: "deploy.sh" # Includes (DATABASE_URL, REDIS_URL, API_TOKEN)
 ```

For now this only supports the simple string array format where the secret key becomes the environment variable name. The Secret struct is designed to support future extensions like custom environment variable names, file secrets, SSH keys, etc.

Example with the amp-buildkite-plugin:

```
secrets:
  - AMP_API_KEY

steps:
 - plugins:
     sj26/amp:
       prompt:
        Tell me a story.
```

Adds the `AMP_API_KEY` to the shell environment (available to all hooks):

Screenshot 2025-09-08 at 2 25 26 pm

`AMP_API_KEY` is available within the environment hook:

Screenshot 2025-09-08 at 2 28 42 pm

`AMP_API_KEY` is available within the `post-checkout` hook:

Screenshot 2025-09-08 at 2 29 24 pm

### Context

Fixes PG-301

### Testing

### Disclosures / Credits

Amp was used heavily in the development of this feature, but was reviewed by a human.

### CODE REVIEW

File: agent/job_runner.go [hunk 1]
```diff
@@ -503,6 +503,19 @@ func (r *JobRunner) createEnvironment(ctx context.Context) ([]string, error) {
...
+	if len(r.conf.Job.Step.Secrets) > 0 {
+		if secretsJSON, err := json.Marshal(r.conf.Job.Step.Secrets); err != nil {
+			r.agentLogger.Error("Failed to marshal secrets configuration: %v", err)
+
+			if err != nil {
+				return nil, err
+			}
+		} else {
+			env["BUILDKITE_SECRETS_CONFIG"] = string(secretsJSON)
+		}
```
moskyb: we're double checking the error here and can probably just get rid of the `else`

```suggestion
		secretsJSON, err := json.Marshal(r.conf.Job.Step.Secrets)
		if err != nil {
			r.agentLogger.Error("Failed to marshal secrets configuration: %v", err)
			return nil, err
		}
		
		env["BUILDKITE_SECRETS_CONFIG"] = string(secretsJSON)
```

(one of the benefits of the `if blah, err := func(); err != nil` syntax is that it keeps the scope of both `blah` and `err` solely inside the if block, but they'll be garbage-collected at the end of the outer if block anyway, so there's not much benefit to this scoping)

agent/verify_job.go
```diff
@@ -177,6 +177,41 @@ func (r *JobRunner) verifyJob(ctx context.Context, keySet any) error {
...
+				r.agentLogger.Debug("failed to verifyJob: step.Secrets length %d != jobSecrets length %d", len(step.Secrets), len(jobSecrets))
+				return newInvalidSignatureError(ErrInvalidJob)
+			}
+
+			for i, stepSecret := range step.Secrets {
+				jobSecret := jobSecrets[i]
+
+				if stepSecret == nil && jobSecret == nil {
+					continue
+				}
```
moskyb: does this mean that yaml `null` is a valid secret? i think that's not necessarily a bad thing if so, but it is interesting.

agent/verify_job.go
```diff
@@ -177,6 +177,41 @@ func (r *JobRunner) verifyJob(ctx context.Context, keySet any) error {
...
+
+				if stepSecret.Key != jobSecret.Key {
+					r.agentLogger.Debug("failed to verifyJob: secret at index %d - Key %q != %q", i, stepSecret.Key, jobSecret.Key)
+					return newInvalidSignatureError(ErrInvalidJob)
+				}
+
+				if stepSecret.EnvironmentVariable != jobSecret.EnvironmentVariable {
+					r.agentLogger.Debug("failed to verifyJob: secret at index %d - EnvironmentVariable %q != %q", i, stepSecret.EnvironmentVariable, jobSecret.EnvironmentVariable)
+					return newInvalidSignatureError(ErrInvalidJob)
+				}
```
moskyb: for both of these, we say `Key "blah" != "blerg"`, but don't state the source of the secret (job vs step) — maybe something like `secret at index %d: Step environment variable %q did not match job environment variable %q` or something

internal/secrets/secret.go
```diff
@@ -0,0 +1,71 @@
...
+		var errorMsg string
+		for i, err := range errs {
+			if i == 0 {
+				errorMsg = fmt.Sprintf("\n   %s", err.Error())
+			} else {
+				errorMsg += fmt.Sprintf("\n   %s", err.Error())
+			}
+		}
+		return nil, fmt.Errorf("%s", errorMsg)
+	}
```
moskyb, line 67: this whole thing can be made a lot cleaner using [`errors.Join`](https://pkg.go.dev/errors#Join) — something like:

```Go
  var multiErr error
	for _, key := range keys {
		apiSecret, _, err := client.GetSecret(ctx, &api.GetSecretRequest{
			Key:   key,
			JobID: jobID,
		})
		if err != nil {
			// Decorate your errors as desired etc
			multiErr = errors.Join(multiErr, err)
			continue
		}
	
		secrets = append(secrets, Secret{
			Key:   key,
			Value: apiSecret.Value,
		})
	}
	if multiErr != nil {
		return nil, multiErr
	}
```

this only loses you the indentation in the printed error, but lets callers use `errors.Is`/`errors.As` on the errors to determine causes etc

internal/secrets/secret_test.go
```diff
@@ -0,0 +1,186 @@
...
+	"testing"
+
+	"github.com/buildkite/agent/v3/api"
+)
+
+// mockAPIClient implements the APIClient interface for testing
+type mockAPIClient struct {
+	secrets map[string]*api.Secret // map of secret key to secret response
+	errors  map[string]error       // map of secret key to error to return
+}
```
moskyb: instead of mocking the API client, i'd recommend using `httptest` to mock out a server for a real API client to talk to, as it puts more of our production code under test.  [`agent/agent_worker_test.go`](https://github.com/buildkite/agent/blob/main/agent/agent_worker_test.go#L37-L80) is some good prior art for this

internal/secrets/secret.go
```diff
@@ -0,0 +1,71 @@
...
+		})
+		if err != nil {
+			// Include secret key name (never values) in error messages for debugging
+			// For API errors, extract just the message for cleaner display, unless debug is enabled
+			var apiErr *api.ErrorResponse
+			if !debug && errors.As(err, &apiErr) && apiErr.Message != "" {
+				errs = append(errs, fmt.Errorf("secret %q: %s", key, apiErr.Message))
+			} else {
+				errs = append(errs, fmt.Errorf("secret %q: %w", key, err))
+			}
```
moskyb: i'd be inclined to just use the full error, if only to simplify the code and give customers more breadcrumbs to investigate failures:

```suggestion
			errs = append(errs, fmt.Errorf("secret %q: %w", key, err))
```

i might be missing something with the error display — if it's super ugly i could be convinced to leave it as is ahah

agent/integration/job_verification_integration_test.go
```diff
@@ -577,6 +660,65 @@ func TestJobVerification(t *testing.T) {
...
+			expectedExitStatus:       "0",
+			expectedSignalReason:     "",
+		},
+		{
+			name:                     "when the step signature matches, but the secrets don't match the signed step, it fails signature verification",
+			agentConf:                agent.AgentConfiguration{VerificationFailureBehaviour: agent.VerificationBehaviourBlock},
+			job:                      jobWithMismatchedSecrets,
+			repositoryURL:            defaultRepositoryURL,
+			signingKey:               symmetricJWKFor(t, signingKeyLlamas),
+			verificationJWKS:         jwksFromKeys(t, symmetricJWKFor(t, signingKeyAlpacas)), // Different key to cause mismatch
```
moskyb: ⛔ we should be using the same signing and verification key here, right?

moskyb (changes requested):

internal/secrets/secret.go
matthewborden: I found this more difficult to read:
Screenshot 2025-09-09 at 2 36 25 pm

It was also a regret from `buildkite-agent secret get <secret-name>` that it prints such a long error message with the status codes.

I find this much easier to read, giving each secret a line too, so make it clear which secrets are missing:

Screenshot 2025-09-09 at 2 35 57 pm

The request/status code is included with `--debug`.

internal/secrets/secret.go
matthewborden: Is there a way that I can do both?

I really like the multi-line errors for readability:

Screenshot 2025-09-09 at 2 35 57 pm

agent/verify_job.go
matthewborden: No `null` it shouldn't be valid

internal/secrets/secret.go
moskyb: using `errors.Join` should have each error on a new line, but i don't think they'll get indented

internal/secrets/secret.go
```diff
@@ -0,0 +1,71 @@
...
+		// Format errors with newlines and indentation for better readability
+		var errorMsg string
+		for i, err := range errs {
+			if i == 0 {
+				errorMsg = fmt.Sprintf("\n   %s", err.Error())
+			} else {
+				errorMsg += fmt.Sprintf("\n   %s", err.Error())
+			}
+		}
+		return nil, fmt.Errorf("%s", errorMsg)
```
DrJosh9000: Some implementation notes, aside from the `errors.Join` discussion:-

Firstly, `fmt` functions know how to call the `Error` method for you.

Secondly, `errorMsg` starts as an empty string, and the format string being appended each time is the same (`"\n   %s"`), so this could be simplified to:

```suggestion
		// Format errors with newlines and indentation for better readability
		var errorMsg string
		for _, err := range errs {
			errorMsg += fmt.Sprintf("\n   %s", err)
		}
		return nil, fmt.Errorf("%s", errorMsg)
```

But appending more than a few strings is the strength of `strings.Builder`, which could be used in conjunction with `fmt.Fprintf` like so:

```suggestion
		// Format errors with newlines and indentation for better readability
		var errorMsg strings.Builder
		for _, err := range errs {
			fmt.Fprintf(&errorMsg, "\n   %s", err)
		}
		return nil, fmt.Errorf("%s", errorMsg)
```

But if I'm being honest, passing multiple errors through a single `error` where the string is formatted a particular way is a bit fiddly.
Perhaps this function `FetchSecrets` could return a slice of errors instead (`[]error`), and make the caller responsible for formatting them appropriately? This would work nicer with the Go approach of designing error messages to be incorporated in other output (see https://google.github.io/styleguide/go/decisions#error-strings)

internal/secrets/secret_test.go
matthewborden: Actioned

internal/secrets/secret.go
matthewborden: I've actioned this with the longer errors for now:
Screenshot 2025-09-10 at 12 13 18 pm

internal/secrets/secret_test.go
```diff
@@ -0,0 +1,335 @@
...
+			default:
+				rw.WriteHeader(http.StatusNotFound)
+				fmt.Fprintf(rw, `{"message": "secret not found"}`)
+			}
+		default:
+			t.Errorf("Unknown endpoint %s %s", req.Method, req.URL.Path)
+			http.Error(rw, "Not found", http.StatusNotFound)
+		}
+	}))
+
```
moskyb, line 46: [non-blocking]

given that these are all using the same server, we could potentially pull this out into a function:

```Go
func setupTestServer(t *testing.T) httptest.Server {
  t.Helper()
  // ... actually set up the server
	t.Cleanup(func() { server.Close() }) 
  return server
}
```

internal/secrets/secret_test.go
```diff
@@ -0,0 +1,335 @@
...
+	server := httptest.NewServer(http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+		switch req.URL.Path {
+		case "/jobs/test-job-id/secrets":
+			switch req.URL.Query()["key"][0] {
+			case "DATABASE_URL":
+				rw.WriteHeader(http.StatusOK)
+				fmt.Fprintf(rw, `{"key": "DATABASE_URL", "value": "postgres://user:pass@host:5432/db"}`)
+			case "API_TOKEN":
+				rw.WriteHeader(http.StatusOK)
+				fmt.Fprintf(rw, `{"key": "DATABASE_URL", "value": "secret-token-123"}`)
```
matthewborden, line 83: That's wrong

buildkite-agent
DrJosh9000, line 1: Probably don't commit the agent binary

go.mod
```diff
@@ -16,7 +16,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/kms v1.45.2
 	github.com/brunoscheufler/aws-ecs-metadata-go v0.0.0-20220812150832-b6b31c6eeeaf
 	github.com/buildkite/bintest/v3 v3.3.0
-	github.com/buildkite/go-pipeline v0.15.0
+	github.com/buildkite/go-pipeline v0.15.1-0.20250911042023-a40d93b5ef09
```
DrJosh9000: Do a release of go-pipeline, then update this to remove the commit.

internal/secrets/secret_test.go
```diff
@@ -0,0 +1,335 @@
...
+		Token:    "llamas",
+	})
+
+	secrets, errs := FetchSecrets(context.Background(), apiClient, "test-job-id", []string{}, false)
+
+	if len(errs) > 0 {
+		t.Fatalf("expected no errors, got: %v", errs)
+	}
+
+	if secrets != nil {
```
DrJosh9000: Always use `len` to test for empty or nil slices:

```suggestion
	if len(secrets) > 0 {
```

internal/secrets/secret_test.go
```diff
@@ -0,0 +1,335 @@
...
+		t.Errorf("expected underlying network error, got: %v", secretErr.Err)
+		return
+	}
+
+	// Check for connection refused errors across platforms
+	isConnRefused := false
+	if runtime.GOOS == "windows" {
+		// On Windows, check for WSAECONNREFUSED (10061)
+		if errno, ok := netErr.Err.(syscall.Errno); ok && errno == 10061 {
+			isConnRefused = true
```
DrJosh9000, line 318: TIL

Issue #3224, opened 2025-03-05
please mention GO compiler version in changelog or release notes


dch:
I'm the maintainer of the FreeBSD port, so we build from source. The v3.93.0+ version now requires a newer version of the go toolchain, and tries to auto-download this from Go website.

- this is not ideal from a security perspective, even if you have a lot of trust for Google, it requires allowing *all* code to fetch additional data over the network
- generally we enforce that no additional code is fetched during builds, so users are 100% sure that what was shipped, is what was built, and the toolchain also has a secure provenance

Anyway! Can you mention any future changes in GOTOOLCHAIN version in release notes or changelog please? Ideally, this would be something that buildkite specifically validates in releases, so that a single new dependency doesn't require a new toolchain unless that is really required.

As go 1.24 is not yet available in FreeBSD, we are stuck on v3.92.1 which is not too bad, until that's available to us.

DrJosh9000:
Hi @dch, thanks for raising this issue! I think the Go toolchain increase slipped through as a consequence of me hastily fixing a Dependabot PR with `go mod tidy` without looking at the toolchain bump. So last week I undid the small number of changes that depended on stdlib features in 1.24, and rolled the versions in go.mod back, so you should be clear to build.

I agree that fetching another toolchain at build time is not good for a variety of reasons. (I believe setting the `GOTOOLCHAIN` env var yourself can override whatever go.mod says, but in this case the tests would have failed due to use of the newer features.) 

We could definitely call out Go and toolchain version changes in the release notes - that has sometimes happened in the past, especially if the version bump fixes some important security issue. The importance of it in this case was likely missed since I pushed the commit to fix a Dependabot PR, and they tend to be clustered together at the bottom of the release notes (given how many there usually are).

I think ideally we'd have testing able to guarantee it is safe to upgrade the Go version/toolchain. The effort seems substantial, though. Someone out there might care that [Go 1.24 now requires Linux kernel 3.2 or newer](https://go.dev/doc/go1.24#ports), for example. In the FreeBSD case, we might be misled by the fact that the Go team published releases of Go 1.24 for all their supported platforms (including FreeBSD) to go.dev/dl on Feb 11. 

[...]

PR #3464, opened 2025-09-09
Update shellwords to v1.0.1, relax Go version directive

moskyb:
### Description

We previously relied on a dependency, [`shellwords`](https://github.com/buildkite/shellwords), that required that the Agent to be on a very specific version of Go (1.24.3). In Shellwords [v1.0.1](https://github.com/buildkite/shellwords/releases/tag/v1.0.1), we relaxed this requirement to 1.24, so we should continue up the stream and make the agent as a go module a little more usable

> Note that in general, we don't recommend that people use the agent as a go module, and don't consider it to be versioned at all. However, people are going to be doing it anyway, and we should endeavour to not make their lives needlessly difficult.

This PR also adds a `toolchain` directive to the `go.mod`, which will do the work of actually selecting a go version to run commands like `go test` etc, where that heavy lifting was previously done by an overly-zealous `go` directive.

This work is pretty much taken wholesale from @haydentherapper's #3449, but with a couple tweaks.

### Context

#3449
https://github.com/buildkite/shellwords/pull/5

### Testing

### Disclosures / Credits

Thanks to @haydentherapper and @vpnachev for pushing this!

closes #3449

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,8 @@
 module github.com/buildkite/agent/v3
 
-go 1.24.3
+go 1.24.0
+
+toolchain go1.24.5
 
 require (
 	cloud.google.com/go/compute/metadata v0.8.0
@@ -19,7 +21,7 @@ require (
 	github.com/buildkite/go-pipeline v0.15.0
 	github.com/buildkite/interpolate v0.1.5
 	github.com/buildkite/roko v1.4.0
-	github.com/buildkite/shellwords v1.0.0
+	github.com/buildkite/shellwords v1.0.1
 	github.com/creack/pty v1.1.19
 	github.com/denisbrodbeck/machineid v1.0.1
 	github.com/dustin/go-humanize v1.0.1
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/buildkite/agent/v3
 
-go 1.24.3
+go 1.24
+
+toolchain go1.25.1
```
DrJosh9000: Hmm. Go 1.25 here could cause a repeat of #3224...

moskyb:
@DrJosh9000 i've bumped the `toolchain` directive down to latest 1.24

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/buildkite/agent/v3
 
-go 1.24.3
+go 1.24.0
+
+toolchain go1.24.5
```
DrJosh9000, line 5: It says 1.24.5, and if I'm understanding the issue correctly, we may not want to specify a toolchain line at all?

DrJosh9000: We looked at the docs again and this is probably fine :+1:

DrJosh9000 (approved):

PR #3466, opened 2025-09-09
PB-610: fix hook environment variable unable to propagate via bk-agent env set

zhming0:
### Description

If a hook mutate job environment variable via "buildkite-agent env set "bluhbluh", the Executor config will NOT get refreshed. 

This is because in code we have a check to see if script's environment variables are changed and will skip a bunch of codes if they do. 

This PR fixes that ☝🏿 .

NOTE: I tried as I might to add a test case, but it isn't ideal. I couldn't get the test tooling to work as I wanted, so I am open to feedback of improving.

### Context

PB-610

### Testing

### Disclosures / Credits

No AI for this one.


internal/job/executor.go
```diff
@@ -595,36 +595,22 @@ func (e *Executor) runWrappedShellScriptHook(ctx context.Context, hookName strin
...
 	if afterWd, err := changes.GetAfterWd(); err == nil {
 		if afterWd != e.shell.Getwd() {
 			_ = e.shell.Chdir(afterWd)
 		}
 	}
 
-	// Do we even have any environment variables to change?
-	if changes.Diff.Empty() {
-		return
-	}
```
zhming0, line 608: This is the actual fix.

internal/job/executor.go
```diff
@@ -636,27 +622,58 @@ func (e *Executor) applyEnvironmentChanges(changes hook.EnvChanges) {
...
+	toRedact, short, err := redact.Vars(e.RedactedVars, e.shell.Env.DumpPairs())
+	if err != nil {
+		e.shell.OptionalWarningf("bad-redacted-vars", "Couldn't match environment variable names against redacted-vars: %v", err)
+	}
+	if len(short) > 0 {
+		slices.Sort(short)
+		e.shell.OptionalWarningf("short-redacted-vars", "Some variables have values below minimum length (%d bytes) and will not be redacted: %s", redact.LengthMin, strings.Join(short, ", "))
+	}
+
+	// TODO: this should probably be a reset rather than a mutate.
```
DrJosh9000: Maybe, but does a particular string stop being a secret if we learn new secret strings?

internal/job/executor.go
```diff
@@ -636,27 +622,58 @@ func (e *Executor) applyEnvironmentChanges(changes hook.EnvChanges) {
...
 			e.shell.Commentf("%s is now %q", k, v)
 		} else {
 			e.shell.Commentf("%s removed", k)
 		}
 	}
+
+	// When an env var is changed via buildkite-agent env set instead,
+	// it might not appear in the script "changes".
+	for k, v := range executorConfigEnvChanges {
+		if ok := executorConfigEnvChangesLogged[k]; !ok {
```
DrJosh9000: [a debate topic with Go programmers more generally] `ok` is non-descriptive. What's OK or not-OK about it? Consider instead: `logged`... or leave off the variable declaration entirely:

```suggestion
		if !executorConfigEnvChangesLogged[k] {
```

internal/job/executor.go
zhming0: Ah fair point 👍🏿 , I added this comment to code comment and explained that we use `Add` for prudence.

internal/job/executor.go
zhming0: The `ok` bit is truly misleading, rectified!

PR #3467, opened 2025-09-09
PS-1101: refresh Executor config for Job API env change in polyglot hook

zhming0:
This PR is based on top of #3466.
Closes #3444.

### Description

When a polyglot hook uses `buildkite-agent env set` to update environment, executor config will not be refreshed. And this is against our expectation. 

This PR fixes it. 

A test case would be good but, however, after a few stab, I think a overall test setup refactor will be needed in order for that to happen. 

### Context

PS-1101

### Testing

### Disclosures / Credits

Peasant human coding


DrJosh9000 (approved):
LGTM

PR #3471, opened 2025-09-10
Experiment for propagating agent config env vars

DrJosh9000:
### Description

Add an experiment that enables various agent config vars to be named in the env file.

### Context

Closes #3430

### Changes

- Add a new experiment, `propagate-agent-config-vars`
- When the experiment is enabled, write an extra preamble to the env file, consisting of various env var names that could be usefully propagated.

### Testing

### Disclosures / Credits

I did not use AI tools at all. I copied the list of env vars from #3430.

### DIFF
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -453,11 +453,50 @@ func (r *JobRunner) createEnvironment(ctx context.Context) ([]string, error) {
 	}
 
 	// Write out the job environment to file:
-	// - envShellFile: in k="v" format, with newlines escaped
+	// - envShellFile: in k="v" format, with newlines escaped. If the
+	//   propagate-agent-vars experiment is enabled, the names of several agent
+	//   config variables are prepended at the top.
 	// - envJSONFile: as a single JSON object {"k":"v",...}, escaped appropriately for JSON.
 	// We present only the clean environment - i.e only variables configured
 	// on the job upstream - and expose the path in another environment variable.
 	if r.envShellFile != nil {
+		if experiments.IsEnabled(ctx, experiments.PropagateAgentConfigVars) {
+			// Note that some variables in this list might not be defined later,
+			// when something comes to read the file. See below where they are
+			// added conditionally, e.g. BUILDKITE_TRACING_BACKEND.
+			// Docker in particular tolerates undefined vars in an env file
+			// without complaints.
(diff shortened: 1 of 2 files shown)


DrJosh9000:
Thanks @Mykematt, but can you explain why those env vars are not in the environment readable by the plugin when similar lines of code already exist in `agent/job_runner.go` to add the environment variables? 

- BUILDKITE_GIT_CHECKOUT_FLAGS: https://github.com/buildkite/agent/blob/7d391fee83b39f353b080664bd205f03dad49836/agent/job_runner.go#L534C7-L534C35
- BUILDKITE_GIT_CLEAN_FLAGS: https://github.com/buildkite/agent/blob/7d391fee83b39f353b080664bd205f03dad49836/agent/job_runner.go#L538
- BUILDKITE_TRACE_CONTEXT_ENCODING: https://github.com/buildkite/agent/blob/7d391fee83b39f353b080664bd205f03dad49836/agent/job_runner.go#L546
- BUILDKITE_TRACING_BACKEND and BUILDKITE_TRACING_SERVICE_NAME: https://github.com/buildkite/agent/blob/7d391fee83b39f353b080664bd205f03dad49836/agent/job_runner.go#L602-L603
- BUILDKITE_AGENT_AWS_KMS_KEY: https://github.com/buildkite/agent/blob/7d391fee83b39f353b080664bd205f03dad49836/agent/job_runner.go#L574

Can you also explain why the two tests you linked used different versions of the docker-compose plugin? (v4.16.0 vs v5.3.0)?

DrJosh9000:
@zhming0:

> do you mind giving me some context on what use cases are this env file designed for?

The env file is used by pre-bootstrap hooks to allow custom job validation before running anything else. The env vars are passed via a file specifically to avoid shell injection. See e.g. https://buildkite.com/docs/agent/v3/securing#restrict-access-by-the-buildkite-agent-controller-strict-checks-using-a-pre-bootstrap-hook. Because the format of the file has some quirks, we also write a JSON version, but there will be a few uses of the original env file that I am wary of breaking, given how long it has been in the docs. Hence the need for an experiment (or a different file).

The env file is also used by the docker and docker-compose plugins to figure out what `--env` flags to pass to Docker, so that the corresponding env vars propagate into the container. The goal here is to propagate more of them this way.

PR #3472, opened 2025-09-11
Add GCP token format to OIDC token command

samgaudet:
### Description

This PR introduces functionality to return Buildkite OIDC tokens wrapped in standard JSON expected by GCP for use in Workload Identity Federation Credentials. With this change in place, using GCP Workload Identity Federation with Buildkite is greatly simplified.

### Context

When using Workload Identity Federation in GCP with Buildkite, Workload Identity JSON credentials can be structured in several ways (see: [_External Account Credentials (Workload Identity Federation)_](https://google.aip.dev/auth/4117)). Notably, the most streamlined and secure way of configuring these credentials is to provide a `credential_source` of a command that returns a token (this nullifies the requirement to save the token anywhere along the way, and instead allows the GCP library to execute the command to fetch the token at the time its needed to exchange the token for a GCP access token).

Using an executable `credential_source` with Buildkite's `oidc request-token` command is incompatible, as the output of the executable command is expected to be in a JSON format, not the token by itself.

Moreover, using some creative bash scripting to construct a "one-liner" is incompatible as the executable `credential_source`, as the underlying method of executing the command in Google's libraries runs a `.split()` (in Python) that breaks any one-liner apart in an unusable way. For instance, something like this will not work:

```bash
buildkite-agent oidc request-token | jq -R \"{\\\"id_token\\\": ., \\\"token_type\\\": \\\"urn:ietf:params:oauth:token-type:jwt\\\", \\\"version\\\": 1, \\\"success\\\": true}\"
```

(Additionally, wrapping that whole thing in `'` quotation marks and invoking `/bin/bash`, etc. won't work, either...)

### Changes

This PR adds a new CLI flag, `--gcp-format`, to the `buildkite-agent oidc request-token` command. When this flag is  `true`, the `buildkite-agent oidc request-token` command returns the token wrapped in JSON, as opposed to the default behavior of returning the token as a string.

### Testing

### Disclosures / Credits

This code was written with the help of Cursor and modified by a human (me) before being pushed to this branch.

### DIFF
--- a/clicommand/oidc_request_token.go
+++ b/clicommand/oidc_request_token.go
@@ -2,6 +2,7 @@ package clicommand
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"slices"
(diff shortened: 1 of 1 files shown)


samgaudet:
This is being handled in https://github.com/buildkite/agent/pull/3480

PR #3483, opened 2025-09-16
Batch get secrets

moskyb:
### Description

This PR updates the `buildkite-agent secret get` command to be able to concurrently fetch multiple secrets — eg, where previously you might've had to do something like:
```Bash
for key in "DATABASE_URL" "API_TOKEN" "SOME_OTHER_SECRET"; do
  val=$(buildkite-agent secret get key)
  declare "$key=$val"
done
```

and wait for each secret fetch to happen serially. Now, the `secret get` command can fetch the secrets concurrently, and output them in an easy-to-parse format like JSON or env:
```Bash
buildkite-agent secret get --skip-redaction "DATABASE_URL" "API_TOKEN" "SOME_OTHER_SECRET"
{"DATABASE_URL": "database.postgres:5432", "API_TOKEN": "hunter2", "SOME_OTHER_SECRET": "ssh!"}
# or
buildkite-agent secret get --skip-redaction --format env "DATABASE_URL" "API_TOKEN" "SOME_OTHER_SECRET"
DATABASE_URL="database.postgres:5432"
API_TOKEN="hunter2"
SOME_OTHER_SECRET="ssh!"
```

Note that under normal circumstances (where the `--skip-redaction` flag is omitted) the values of the secrets would be redacted from buildkite job logs.

Additionally, secrets embdedded in the buildkite pipeline YAML will also be fetched concurrently.

### Context

[PS-1064](https://linear.app/buildkite/issue/PS-1064/reddit-get-secrets-in-a-batch-with-buildkite-agent)
[Slack convo](https://buildkite-corp.slack.com/archives/C06DCKYRPL1/p1755819648063879)

### Changes

```
Usage:

    buildkite-agent secret get [options...] [key1] [key2] ...

Description:

Gets a list of secrets from Buildkite and prints them to stdout. Key names are case
insensitive in this command, and secret values are automatically redacted in the build logs
unless the ′skip-redaction′ flag is used.

If any request for a secret fails, the command will return a non-zero exit code and print details of all failed secrets.

If only a single key is provided, the secret value will be printed without any formatting.

If multiple keys are provided, the output format will be as defined by the ′format′ flag, which defaults to JSON.

Examples:

    # Secret keys are case insensitive
    $ buildkite-agent secret get deploy_key
    "..."
    $ buildkite-agent secret get DEPLOY_KEY
    "..."

    # Format is ignored when only a single key is provided
    $ buildkite-agent secret get --format env deploy_key
    "..."

    # JSON is the default format when multiple keys are provided
    $ buildkite-agent secret get deploy_key github_api_token
    {"deploy_key": "...", "github_api_token": "..."}

    # ... but you can also request env format, which can be piped into ′source′, ′declare -x′, and friends
    $ buildkite-agent secret get --format env deploy_key github_api_token
    DEPLOY_KEY="..."
    GITHUB_API_TOKEN="..."

Options:

  --no-color                  Don't show colors in logging [$BUILDKITE_AGENT_NO_COLOR]
  --debug                     Enable debug mode. Synonym for ′--log-level debug′. Takes precedence over ′--log-level′ [$BUILDKITE_AGENT_DEBUG]
  --log-level value           Set the log level for the agent, making logging more or less verbose. Defaults to notice. Allowed values are: debug, info, error, warn, fatal (default: "notice") [$BUILDKITE_AGENT_LOG_LEVEL]
  --experiment value          Enable experimental features within the buildkite-agent [$BUILDKITE_AGENT_EXPERIMENT]
  --profile value             Enable a profiling mode, either cpu, memory, mutex or block [$BUILDKITE_AGENT_PROFILE]
  --agent-access-token value  The access token used to identify the agent [$BUILDKITE_AGENT_ACCESS_TOKEN]
  --endpoint value            The Agent API endpoint (default: "https://agent.buildkite.com/v3") [$BUILDKITE_AGENT_ENDPOINT]
  --no-http2                  Disable HTTP2 when communicating with the Agent API. [$BUILDKITE_NO_HTTP2]
  --debug-http                Enable HTTP debug mode, which dumps all request and response bodies to the log [$BUILDKITE_AGENT_DEBUG_HTTP]
  --trace-http                Enable HTTP trace mode, which logs timings for each HTTP request. Timings are logged at the debug level unless a request fails at the network level in which case they are logged at the error level [$BUILDKITE_AGENT_TRACE_HTTP]
  --job value                 Which job should should the secret be for [$BUILDKITE_JOB_ID]
  --format value              The output format, either 'default', 'json', or 'env'. When 'default', a single secret will print just the value, while multiple secrets will print JSON. When 'json' or 'env', secrets will be printed as key-value pairs in the requested format (default: "default") [$BUILDKITE_AGENT_SECRET_GET_FORMAT]
  --skip-redaction            Skip redacting the retrieved secret from the logs. Then, the command will print the secret to the Job's logs if called directly. [$BUILDKITE_AGENT_SECRET_GET_SKIP_SECRET_REDACTION]
```

### Testing

### Disclosures / Credits

It's all artisanal, handcrafted slop from me.


internal/secrets/secret.go
```diff
@@ -34,31 +35,42 @@ func (e *SecretError) Unwrap() error {
...
 // If any secret fails, returns error with details of all failed secrets.
-func FetchSecrets(ctx context.Context, client APIClient, jobID string, keys []string, debug bool) ([]Secret, []error) {
+func FetchSecrets(ctx context.Context, client APIClient, jobID string, keys []string, concurrency int) ([]Secret, []error) {
 	if len(keys) == 0 {
 		return nil, nil
 	}
 
 	secrets := make([]Secret, 0, len(keys))
 	errs := make([]error, 0, len(keys))
+	sem := semaphore.NewWeighted(int64(concurrency))
```
DrJosh9000, line 46: Why not the stdlib `sync.WaitGroup` ?

DrJosh9000: Oh wait, you're limiting concurrency that way. Never mind

internal/secrets/secret_test.go
```diff
@@ -50,7 +50,7 @@ func TestFetchSecrets_Success(t *testing.T) {
 	})
 
 	keys := []string{"DATABASE_URL", "API_TOKEN"}
-	secrets, errs := FetchSecrets(context.Background(), apiClient, "test-job-id", keys, false)
+	secrets, errs := FetchSecrets(context.Background(), apiClient, "test-job-id", keys, 1)
```
DrJosh9000: While you're here, why not update all these:

```suggestion
	secrets, errs := FetchSecrets(t.Context(), apiClient, "test-job-id", keys, 1)
```

clicommand/secret_batch_read.go
```diff
@@ -0,0 +1,137 @@
...
+
+	Keys          []string `cli:"arg:*"`
+	Format        string   `cli:"format" validate:"required"`
+	Job           string   `cli:"job" validate:"required"`
+	SkipRedaction bool     `cli:"skip-redaction"`
+}
+
+var SecretBatchReadCommand = cli.Command{
+	Name:  "batch-read",
+	Usage: "Concurrently fetch a batch of secrets by their keys",
```
DrJosh9000: Mentioning "concurrently" here is implying something that might not be true in future (e.g. replacing all this with a single request). Maybe just
```suggestion
	Usage: "Fetch a batch of secrets by their keys",
```

clicommand/secret_batch_read.go
```diff
@@ -0,0 +1,137 @@
...
+
+		switch cfg.Format {
+		case "json":
+			if err := json.NewEncoder(c.App.Writer).Encode(secretsMap); err != nil {
+				return fmt.Errorf("failed to write JSON response: %w", err)
+			}
+
+		case "env":
+			sb := strings.Builder{}
+			for key, value := range secretsMap {
```
DrJosh9000: For consistency across multiple runs/tests, perhaps sort these by key?


```suggestion
			for _, key := range slices.Sorted(maps.Keys(secretsMap)) {
				sb.WriteString(fmt.Sprintf("%s=%q\n", strings.ToUpper(key), secretsMap[value]))
```

DrJosh9000 (commented):
Looks pretty good, I have some tweaks

clicommand/secret_batch_read.go
moskyb: ooh yep, good call

clicommand/secret_batch_read.go
moskyb: yeah, that's nice actually, good call

clicommand/secret_batch_read.go
```diff
@@ -0,0 +1,137 @@
...
+			return fmt.Errorf("invalid format %q: must be either 'json' or 'env'", cfg.Format)
+		}
+
+		agentClient := api.NewClient(l, loadAPIClientConfig(cfg, "AgentAccessToken"))
+		secrets, errs := secrets.FetchSecrets(ctx, agentClient, cfg.Job, cfg.Keys, 20)
+		if len(errs) > 0 {
+			sb := &strings.Builder{}
+			sb.WriteString("Failed to fetch some secrets:\n")
+			for _, err := range errs {
+				sb.WriteString(fmt.Sprintf(" - %s\n", err.Error()))
```
DrJosh9000: We can take advantage of `sb` being an `io.Writer`, and `fmt` functions understanding how to call `.Error()`:

```suggestion
				fmt.Fprintf(sb, " - %v\n", err)
```

internal/secrets/secret.go
```diff
@@ -34,31 +36,48 @@ func (e *SecretError) Unwrap() error {
...
+			errs = append(errs, fmt.Errorf("failed to acquire semaphore for key %q: %w", key, err))
+			errsMu.Unlock()
+			break
 		}
 
-		secrets = append(secrets, Secret{
-			Key:   key,
-			Value: apiSecret.Value,
-		})
+		go func(key string) {
```
DrJosh9000: The new loop variable semantics should mean you don't have to pass `key` through an arg.

clicommand/secret_batch_read.go
moskyb: oh how good, that's much nicer

internal/secrets/secret.go
moskyb: oh yay! we love go 1.23+

internal/secrets/secret_test.go
moskyb: i have another branch locally to clean up the tests in this file, will submit along with that

clicommand/secret_get.go
```diff
@@ -14,37 +19,63 @@ type SecretGetConfig struct {
...
+    GITHUB_API_TOKEN="..."`,
 	Flags: slices.Concat(globalFlags(), apiFlags(), []cli.Flag{
 		cli.StringFlag{
 			Name:   "job",
 			Usage:  "Which job should should the secret be for",
 			EnvVar: "BUILDKITE_JOB_ID",
 		},
+		cli.StringFlag{
+			Name:   "format",
+			Usage:  "The output format, either 'json' (default) or 'env'. Ignored when only a single key is provided",
```
DrJosh9000:
```suggestion
			Usage:  "The output format for multiple keys, either 'json' (default) or 'env'. Ignored when only a single key is provided",
```

clicommand/secret_get.go
```diff
@@ -56,28 +87,71 @@ The following examples reference the same Buildkite secret ′key′:
...
+		switch cfg.Format {
+		case "json":
+			if err := json.NewEncoder(c.App.Writer).Encode(secretsMap); err != nil {
+				return fmt.Errorf("failed to write JSON response: %w", err)
+			}
+
+		case "env":
+			sb := strings.Builder{}
+			for _, key := range slices.Sorted(maps.Keys(secretsMap)) {
+				sb.WriteString(fmt.Sprintf("%s=%q\n", strings.ToUpper(key), secretsMap[key]))
```
DrJosh9000: Similar thing here: 

```suggestion
				fmt.Fprintf(&sb, "%s=%q\n", strings.ToUpper(key), secretsMap[key])
```

Or... ditch the string builder and write directly to `c.App.Writer` ?

DrJosh9000 (approved):
✅

clicommand/secret_get.go
```diff
@@ -14,37 +19,63 @@ type SecretGetConfig struct {
...
+Examples:
 
+    # Secret keys are case insensitive
     $ buildkite-agent secret get deploy_key
-    $ buildkite-agent secret get DEPLOY_KEY`,
+    "..."
+    $ buildkite-agent secret get DEPLOY_KEY
+    "..."
+
+    # Format is ignored when only a single key is provided
```
dabarrell: I'd suggest we try to avoid ignoring params/magically changing them. No reason we can't return a single key in `env` or `json` format. Folks might want to build a script that dynamically generates a list of secrets – we'd want the output to be formatted the same whether or not that script generate one or multiple secrets

dabarrell (changes requested):

clicommand/secret_get.go
dabarrell: When no format is specified and multiple secrets are requested, I could go either way - defaulting to JSON or throwing an error and requiring the `format` param

moskyb: i think the issue there is that to do that, we'd need to make the default format `value` (or something, where we just print the value to stdout), but ignore (or disallow) that format when there are multiple secrets.

this was the original impetus behind making batch read its own command — format ergonomics are tricky when the current default format (`value`) is kinda meaningless when you fetch multiple secrets (doubly so because secrets don't necessarily come out in the order you fetched them)

DrJosh9000: Should the format flag have some default value that isn't a particular format? `"default"`, which is interpreted to mean a single raw value for a single key, and JSON for multiple?

clicommand/secret_get.go
moskyb: yeah, my thought was that writing to a stringbuilder a bunch of times and then writing to stdout once is going to be more efficient, but i don't think that's an optimisation that needs to be there. i'll just write straight to stdout, it's simpler

clicommand/secret_get.go
moskyb: yeah, i can live with that. i'll implement it

clicommand/secret_get.go
```diff
@@ -99,7 +99,7 @@ Examples:
 		}
 
 		agentClient := api.NewClient(l, loadAPIClientConfig(cfg, "AgentAccessToken"))
-		secrets, errs := secrets.FetchSecrets(ctx, agentClient, cfg.Job, cfg.Keys, 20)
+		secrets, errs := secrets.FetchSecrets(ctx, agentClient, cfg.Job, cfg.Keys, 10)
```
blaknite, line 102: 10 should be fine.

Here's reddit's concurrent jobs over the past 7 days. Not exactly massive load.

image

clicommand/secret_get.go
```diff
@@ -14,37 +19,66 @@ type SecretGetConfig struct {
...
+		{"deploy_key": "..."}
+
+    # JSON is the default format when multiple keys are provided...
+    $ buildkite-agent secret get deploy_key github_api_token
+    {"deploy_key": "...", "github_api_token": "..."}
+
+    # ... but you can also request env format, which can be piped into ′source′, ′declare -x′, and friends
+    $ buildkite-agent secret get --format env deploy_key github_api_token
+    DEPLOY_KEY="..."
+    GITHUB_API_TOKEN="..."`,
```
dabarrell: The help text feels like it's repeating itself a bit, showing both single and multiple keys for both env and json. I think it's self-explanatory that env/json can be used for single keys, so we probably don't need to call it out. Can I suggest something like this?
```suggestion
    "..."
    $ buildkite-agent secret get DEPLOY_KEY
    "..."
    
    # The return value can also be formatted using `env` (which can be piped
    # into e.g. ′source′, ′declare -x′), or `json`
    
    $ buildkite-agent secret get --format env deploy_key github_api_token
    DEPLOY_KEY="..."
    GITHUB_API_TOKEN="..."`

    $ buildkite-agent secret get --format json deploy_key github_api_token
    {"deploy_key": "...", "github_api_token": "..."}
    
    # The default format for a single secret is the raw value, and for multiple secrets is `json`
```

clicommand/secret_get.go
dabarrell: thanks @moskyb – LGTM

PR #3484, opened 2025-09-17
Simplify secret tests

moskyb:
### Description

There's a bunch of stuff going on in the `internal/secrets` tests. This PR removes a bunch of unused stuff, and generally makes them a wee bit easier to read.

Relies on #3483 

### Testing

### Disclosures / Credits

It's all me.

### DIFF
--- a/internal/secrets/secret_test.go
+++ b/internal/secrets/secret_test.go
@@ -37,6 +37,7 @@ func TestFetchSecrets_Success(t *testing.T) {
 			http.Error(rw, "Not found", http.StatusNotFound)
 		}
 	}))
+	t.Cleanup(server.Close)
 
 	apiClient := api.NewClient(logger.Discard, api.Config{
 		Endpoint: server.URL,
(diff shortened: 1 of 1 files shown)


internal/secrets/secret_test.go
```diff
@@ -74,77 +65,24 @@ func TestFetchSecrets_Success(t *testing.T) {
...
-		}
-	}))
-
-	apiClient := api.NewClient(logger.Discard, api.Config{
-		Endpoint: server.URL,
-		Token:    "llamas",
-	})
-
-	secrets, errs := FetchSecrets(context.Background(), apiClient, "test-job-id", []string{}, 10)
+	apiClient := api.NewClient(logger.Discard, api.Config{Token: "llamas"})
```
DrJosh9000: This client would send requests to the default (real) API, is that intentional? Shouldn't there still be a test server?

internal/secrets/secret_test.go
```diff
@@ -74,77 +65,24 @@ func TestFetchSecrets_Success(t *testing.T) {
...
-			default:
-				rw.WriteHeader(http.StatusNotFound)
-				fmt.Fprintf(rw, `{"message": "secret not found"}`)
-			}
-		default:
-			t.Errorf("Unknown endpoint %s %s", req.Method, req.URL.Path)
-			http.Error(rw, "Not found", http.StatusNotFound)
-		}
-	}))
+	apiClient := api.NewClient(logger.Discard, api.Config{Token: "llamas"})
```
DrJosh9000: This client too, shouldn't it use a test server as well?

internal/secrets/secret_test.go
moskyb: my thought was that because it short-circuits and just returns an empty slice when the  `keys` slice is empty, it doesn't need one, but that's testing internal behaviour implicitly. lemme re-add something to make it clear it doesn't hit the endpoint

DrJosh9000 (approved):
Nice

PR #3485, opened 2025-09-17
Update apply-if-changed information with agent minimum version

Damilola-obasa:
### Description

When reviewing this [document](https://buildkite.com/docs/agent/v3/cli-pipeline#apply-if-changed) the minimum agent version required is not displayed 

### Context

Users may get errors when trying to use the `--apply-if-changed` flag until they notice it is required to have their agent be on at least `v3.99` as described [ here](https://buildkite.com/docs/pipelines/configure/step-types/command-step#agent-applied-attributes)

### Changes

Added minimum agent version required information for `--apply-if-changed` flag as this will help users with older than  v3.99 version quickly identify and upgrade to the minimum version required for their agent before they can use this flag

### DIFF
--- a/clicommand/pipeline_upload.go
+++ b/clicommand/pipeline_upload.go
@@ -137,7 +137,7 @@ var PipelineUploadCommand = cli.Command{
 		},
 		cli.BoolTFlag{
 			Name:   "apply-if-changed",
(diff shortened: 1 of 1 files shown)


Damilola-obasa:
Added minimum agent version required information for "--apply-if-changed" flag as this will help users with older version quickly identify and upgrade to the minimum version required for their agent before they can use this flag

DrJosh9000 (approved):
Thanks, LGTM

PR #3488, opened 2025-09-18
Ability to checkout subdirectories of Plugins

tomowatt:
### Description

Currently it's not possible to specify subdirectories of Plugins to be used, meaning it's not possible for an user to have a monorepo with multiple Plugins inside and use them by referencing the subdirectory in e.g.,

```yml
 - label: "subdir plugin with .git"
    command: exit 0
    plugins:
      - ssh://git@github.com/tomowatt/test-plugin.git/subdir: ~
```

### Changes

Adds functionality to allow the Agent to clone Plugin repositories that specify a sub directory in their URL and then update the Plugin Dir and Hook paths so that the hooks from the sub directories can be used instead of those at root i.e., `./hooks`

### Testing

Testing Pipeline with Test Plugin Repository, testing out the following:
* `.git` suffix on repositories
* multiple plugins at same level
* plugin deeper within directories
 
```yml
 env:
   BUILDKITE_PLUGINS_ALWAYS_CLONE_FRESH: true
   
 steps:
  - label: "branch agent - subdir plugin"
    agents:
      test: true
    command: exit 0
    plugins:
       - ssh://git@github.com/tomowatt/the-repository-of-many-buildkite-plugins.git/plugin-1: ~
       - ssh://git@github.com/tomowatt/the-repository-of-many-buildkite-plugins/plugin-2: ~
       - ssh://git@github.com/tomowatt/the-repository-of-many-buildkite-plugins/plugin-3/plugin-4: ~

  - label: "v3.107.2 agent - subdir plugin"
    command: exit 0
    soft_fail: true
    plugins:
       - ssh://git@github.com/tomowatt/the-repository-of-many-buildkite-plugins.git/plugin-1: ~
       - ssh://git@github.com/tomowatt/the-repository-of-many-buildkite-plugins/plugin-2: ~
       - ssh://git@github.com/tomowatt/the-repository-of-many-buildkite-plugins/plugin-3/plugin-4: ~
```

with Job Log Output:

```
Preparing plugins
Running plugin the-repository-of-many-buildkite-plugins/plugin-1 environment hook
Plugin 1 Environment Hook
Preparing working directory
Running commands
Running plugin the-repository-of-many-buildkite-plugins/plugin-2 post-command hook
Plugin 2 Post-Command Hook
Running plugin the-repository-of-many-buildkite-plugins/plugin-3/plugin-4 pre-exit hook
Plugin 4 Pre-Exit
```

### Disclosures / Credits

This is AI created.

### DIFF
--- a/agent/plugin/plugin.go
+++ b/agent/plugin/plugin.go
@@ -319,9 +319,9 @@ func (p *Plugin) Label() string {
 
 // DisplayName returns a human-friendly name for the plugin suitable for logs.
 // Examples:
-//  - github.com/org/repo           => repo
-//  - github.com/org/repo/.buildkite => repo/.buildkite
-//  - file:///path/to/plugin         => plugin (last path element)
+//   - github.com/org/repo           => repo
+//   - github.com/org/repo/.buildkite => repo/.buildkite
+//   - file:///path/to/plugin         => plugin (last path element)
 func (p *Plugin) DisplayName() string {
 	// Filesystem paths: fall back to Name(), which returns the last segment normalized
 	if strings.HasPrefix(p.Location, "/") || strings.HasPrefix(p.Location, ".") || strings.Contains(p.Location, "\\") {
--- a/agent/plugin/plugin.go
+++ b/agent/plugin/plugin.go
@@ -147,6 +147,12 @@ func (p *Plugin) Name() string {
 	parts := strings.Split(location, "/")
 	name := parts[len(parts)-1]
 
+	// If the last path segment starts with a dot (e.g. ".buildkite"), trim leading dots
+	// so the human-friendly name isn't rendered as "-buildkite" after normalization.
+	if strings.HasPrefix(name, ".") {
+		name = strings.TrimLeft(name, ".")
+	}
+
 	// Clean up the name
 	name = strings.ToLower(name)
 	name = whitespaceRE.ReplaceAllString(name, " ")
@@ -167,7 +173,8 @@ func (p *Plugin) Identifier() (string, error) {
 	return id, nil
 }
 
-// Repository returns the repository host where the code is stored.
+// Repository returns the repository URL where the plugin code is stored, without any subdirectory path.
+// For example, for "github.com/buildkite/plugins/docker-compose/plugin", it returns "github.com/buildkite/plugins".
 func (p *Plugin) Repository() (string, error) {
 	s, err := p.constructRepositoryHost()
 	if err != nil {
@@ -192,6 +199,8 @@ func (p *Plugin) Repository() (string, error) {
 }
 
 // RepositorySubdirectory returns the subdirectory path that the plugin is in.
+// For example, for "github.com/buildkite/plugins/docker-compose/plugin", it returns "docker-compose/plugin".
+// If the plugin is in the root of the repository, it returns an empty string.
 func (p *Plugin) RepositorySubdirectory() (string, error) {
 	repository, err := p.constructRepositoryHost()
 	if err != nil {
@@ -199,8 +208,12 @@ func (p *Plugin) RepositorySubdirectory() (string, error) {
 	}
 
 	dir := strings.TrimPrefix(p.Location, repository)
+	dir = strings.TrimPrefix(dir, "/")
 
-	return strings.TrimPrefix(dir, "/"), nil
+	// Remove .git suffix if present as it's not part of the subdirectory
+	dir = strings.TrimSuffix(dir, ".git")
+
+	return dir, nil
 }
 
 // formatEnvKey converts strings into an ENV key friendly format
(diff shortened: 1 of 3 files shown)


internal/job/plugin.go
```diff
@@ -417,12 +439,38 @@ func (e *Executor) checkoutPlugin(ctx context.Context, p *plugin.Plugin) (*plugi
 		return nil, err
 	}
 
-	// Switch to the version if we need to
+	// If there's a subdirectory, configure sparse checkout regardless of version
+	if subdir != "" {
+		e.shell.Commentf("Configuring sparse checkout for subdirectory %q", subdir)
+		// Enable sparse checkout via modern porcelain
+		if err = e.shell.Command("git", "sparse-checkout", "set", subdir).Run(ctx); err != nil {
```
DrJosh9000: Although `git-sparse-checkout` has existed since at least v2.25, the documentation still has a shouty-case warning ("THIS COMMAND IS EXPERIMENTAL. ..."), so I'm wary.

internal/job/plugin.go
```diff
@@ -417,12 +439,38 @@ func (e *Executor) checkoutPlugin(ctx context.Context, p *plugin.Plugin) (*plugi
...
+			// Fallback for older Git versions
+			if err2 := e.shell.Command("git", "config", "core.sparseCheckout", "true").Run(ctx); err2 != nil {
+				return nil, fmt.Errorf("failed to enable sparse checkout: %w", err)
+			}
+			sparseCheckoutFile := filepath.Join(".git", "info", "sparse-checkout")
+			sparseCheckoutContent := fmt.Sprintf("/%s/**\n", subdir)
```
DrJosh9000: You can get a `[]byte` directly from `fmt.Appendf`:

```suggestion
			sparseCheckoutContent := fmt.Appendf(nil, "/%s/**\n", subdir)
```

What you're going for could be a "cone pattern", which has some optimisations: https://git-scm.com/docs/git-sparse-checkout/2.28.0#_cone_pattern_set

But I do think we should avoid sparse-checkout at least for the moment.

internal/job/plugin.go
```diff
@@ -399,11 +411,21 @@ func (e *Executor) checkoutPlugin(ctx context.Context, p *plugin.Plugin) (*plugi
...
 
 	args := []string{"clone", "-v"}
+
+	// If there's a subdirectory, prefer a sparse clone to reduce data transfer and avoid
+	// an initial checkout that would then be reconfigured. We'll checkout after configuring
+	// sparse-checkout below.
+	if subdir != "" {
+		e.shell.Commentf("Using sparse checkout for subdirectory %q", subdir)
+		args = append(args, "--filter=blob:none", "--sparse", "--no-checkout")
+	}
```
DrJosh9000: For the first cut (and possibly forever) I would not sparse-checkout the subdirectory. A use-case for multiple plugins in one repo would be to share some library of common code among them. That library is probably going to exist outside each plugin's subdirectory.

tomowatt: Cheer @DrJosh9000 that's it removed now, get where you are coming from on that point

internal/job/plugin.go
```diff
@@ -310,6 +310,15 @@ func (e *Executor) checkoutPlugin(ctx context.Context, p *plugin.Plugin) (*plugi
 		return nil, err
 	}
 
+	// Get the subdirectory path if specified in the plugin location
+	subdir, err := p.RepositorySubdirectory()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get repository subdirectory: %w", err)
+	}
+	if e.Debug {
+		e.shell.Commentf("Plugin subdirectory: %q", subdir)
```
DrJosh9000: This effectively repeats the debug log below, right?

internal/job/plugin.go
```diff
@@ -361,12 +379,15 @@ func (e *Executor) checkoutPlugin(ctx context.Context, p *plugin.Plugin) (*plugi
 
 		// Ensure hooks is a directory that exists within the checkout.
 		if fi, err := pluginRoot.Stat(checkout.HooksDir); err != nil || !fi.IsDir() {
+			if e.Debug {
+				e.shell.Commentf("Plugin paths - PluginDir: %q, HooksDir: %q", checkout.PluginDir, checkout.HooksDir)
+			}
```
DrJosh9000: Basically the same debug log again?

internal/job/plugin.go
```diff
@@ -319,6 +328,15 @@ func (e *Executor) checkoutPlugin(ctx context.Context, p *plugin.Plugin) (*plugi
...
 	}
 
+	// If there's a subdirectory, we'll adjust the paths accordingly
+	if subdir != "" {
+		checkout.PluginDir = subdir
+		checkout.HooksDir = filepath.Join(subdir, "hooks")
+		if e.Debug {
+			e.shell.Commentf("Subdirectory plugin paths - PluginDir: %q, HooksDir: %q", checkout.PluginDir, checkout.HooksDir)
+		}
+	}
```
DrJosh9000, line 335: What if this debug log was moved out of the `if subdir != ""` check? e.g.

```suggestion
	}
	if e.Debug {
		e.shell.Commentf("Plugin checkout paths - PluginDir: %q, HooksDir: %q", checkout.PluginDir, checkout.HooksDir)
	}
```

agent/plugin/plugin.go
```diff
@@ -304,6 +317,35 @@ func (p *Plugin) Label() string {
 	return p.Location + "#" + p.Version
 }
 
+// DisplayName returns a human-friendly name for the plugin suitable for logs.
```
DrJosh9000, line 320: How much does `DisplayName` differ from `Name`? Would the change from `Name` to `DisplayName` in the logs confuse anyone with existing plugins? Would `Name` be good enough for users of subdirectory plugins? (Is this method worth adding?)

Since subdirectory plugins are (effectively) new, we could change `Name` for them without breaking plugins depending on config env vars. Maybe the existing `Name` plus `-` plus the last element of the subdirectory path would be enough?

internal/job/plugin.go
tomowatt: Cheers @DrJosh9000 cleaned up now

agent/plugin/plugin.go
tomowatt: @DrJosh9000 In terms of differing it's purely cosmetic in understanding where the Plugin came from a repository perspective so it `DisplayName` could be dropped entirely. The `DisplayName` func is only used twice, so happy to drop if you think it's too much.

Here's the differences:
1. Using `DisplayName`
```
Preparing plugins
Running plugin the-repository-of-many-buildkite-plugins/plugin-1 environment hook
Plugin 1 Environment Hook
Preparing working directory
Running commands
Running plugin the-repository-of-many-buildkite-plugins/plugin-2 post-command hook
Plugin 2 Post-Command Hook
Running plugin the-repository-of-many-buildkite-plugins/plugin-3/plugin-4 pre-exit hook
Plugin 4 Pre-Exit
```
2. Using `Name`
```
Preparing plugins
Running plugin plugin-1 environment hook
Plugin 1 Environment Hook
Preparing working directory
Running commands
Running plugin plugin-2 post-command hook
Plugin 2 Post-Command Hook
Running plugin plugin-4 pre-exit hook
Plugin 4 Pre-Exit
```

DrJosh9000: Hmm. Yeah OK, it's an improvement.

PR #3493, opened 2025-09-24
Set e.checkoutRoot even if checkout phase is disabled

DrJosh9000:
### Description

Set `e.checkoutRoot` to the checkout path in the case where the checkout phase is disabled.

### Context

Fixes bug introduced in #3445 when used with agent-stack-k8s, which runs the checkout phase in a separate container, leading to `e.checkoutRoot` being nil within the command container(s).

### Changes

As described in the Description.

### Testing

### Disclosures / Credits

No AI needed or used.


moskyb (approved):
Nice

PR #3500, opened 2025-09-26
Fix log collection stopping too early on SIGTERM in Kubernetes

scadu:
### Description

agent-stack-k8s part: https://github.com/buildkite/agent-stack-k8s/pull/719

When a Kubernetes pod receives SIGTERM, the agent container immediately stops log collection, but command containers and post-command hooks continue running for up to 60 seconds (TerminationGracePeriodSeconds).

Wait up to 50 seconds for Kubernetes processes to complete before stopping log collection, ensuring post-command hook logs are captured during SIGTERM leaving time for log upload.
### Context

https://linear.app/buildkite/issue/PS-1220/

### Testing

### Disclosures / Credits

Kudos to @DrJosh9000 for educative comments in this PR that helped me orientate in the codebase ❤️

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -921,14 +921,7 @@ var AgentStartCommand = cli.Command{
 			return err
 		}
 
-		var kubernetesLogCollectionGracePeriod time.Duration
-		if t := cfg.KubernetesLogCollectionGracePeriod; t != "" {
-			var err error
-			kubernetesLogCollectionGracePeriod, err = time.ParseDuration(t)
-			if err != nil {
-				return fmt.Errorf("failed to parse kubernetes-log-collection-grace-period: %w", err)
-			}
-		}
+		kubernetesLogCollectionGracePeriod := cfg.KubernetesLogCollectionGracePeriod
 
 		if _, err := tracetools.ParseEncoding(cfg.TraceContextEncoding); err != nil {
 			return fmt.Errorf("while parsing trace context encoding: %v", err)


agent/run_job.go
```diff
@@ -428,6 +439,24 @@ func (r *JobRunner) cleanup(ctx context.Context, wg *sync.WaitGroup, exit core.P
...
+	defer cancel()
+
+	select {
+	case <-r.process.Done():
+		r.agentLogger.Debug("[JobRunner] Kubernetes processes completed, stopping log collection")
+	case <-timeoutCtx.Done():
+		r.agentLogger.Info("[JobRunner] Timeout waiting for Kubernetes processes, stopping log collection")
+	}
+
+	time.Sleep(1 * time.Second)
```
DrJosh9000: What's this sleep for?

scadu: @DrJosh9000 most probably unnecessary, or could be replaced with a better mechanism to allow k8s handle job cleanup gracefully.

agent/run_job.go
```diff
@@ -25,6 +25,12 @@ import (
 )
 
 const (
+	// defaultKubernetesLogCollectionGracePeriod defines how long to wait for Kubernetes
+	// processes to complete before stopping log collection during graceful termination.
+	// This should be less than agent-stack-k8s' defaultTermGracePeriodSeconds (60s) to
+	// allow time for final log upload before the pod is forcefully terminated.
+	defaultKubernetesLogCollectionGracePeriod = 50 * time.Second
```
DrJosh9000: [Something that could go in a ticket] Perhaps we should pass agent-stack-k8s term grace period into the agent (via env var, probably).

scadu: @DrJosh9000 from what I can see, we'd need to implement it similarly to what we do with `BUILDKITE_CONTAINER_COUNT` and read the `podSpec.Containers`. Will take a look!

scadu: agent-stack-k8s part: https://github.com/buildkite/agent-stack-k8s/pull/719

agent/run_job.go
```diff
@@ -428,6 +440,55 @@ func (r *JobRunner) cleanup(ctx context.Context, wg *sync.WaitGroup, exit core.P
...
+	case <-r.process.Started():
+		// Process has started, we can now safely wait for completion
+	case <-ctx.Done():
+		r.agentLogger.Debug("[JobRunner] Context cancelled before process started, skipping wait")
+		return
+	}
+
+	// Get grace period from environment variable, fallback to default
+	gracePeriod := defaultKubernetesLogCollectionGracePeriod
+	if envVal := os.Getenv("BUILDKITE_K8S_LOG_COLLECTION_GRACE_PERIOD_SECONDS"); envVal != "" {
```
DrJosh9000: The common pattern for configuration in the agent is to declare it as a flag+env var, since that way it could also be set via flag or config file  if we needed to, and the description flows through to the docs. This would involve:

- adding a new field in `AgentStartConfig` in `clicommand/agent_start.go` (don't forget the `cli:...` field tag)
- adding a new ~~`cli.IntFlag`~~ `cli.DurationFlag` in the `AgentStartCommand` in `clicommand/agent_start.go` (using a duration would let you drop the `_SECONDS`, but would have to be passed with a `s` suffix for seconds).
- adding a new field in `AgentConfiguration` in `agent/agent_configuration.go`
- passing the value between structs in the implementation
- using it here from `r.AgentConfiguration`

That would also let you avoid manually parsing a string here (`strconv.Atoi`), and would be a reasonable place to set (and document) the default.

DrJosh9000: The brain-wave I had after writing that comment was to use a duration instead of an int. Then it could be called `BUILDKITE_K8S_LOG_COLLECTION_GRACE_PERIOD` and given a value like `50s` instead of `50`.

agent/run_job.go
```diff
@@ -428,6 +440,55 @@ func (r *JobRunner) cleanup(ctx context.Context, wg *sync.WaitGroup, exit core.P
...
+		if seconds, err := strconv.Atoi(envVal); err == nil && seconds > 0 {
```
DrJosh9000: For future code: more idiomatic is to test `err != nil` and avoid testing `err == nil`. More here: https://google.github.io/styleguide/go/decisions#indent-error-flow

agent/run_job.go
scadu: > The common pattern for configuration in the agent is to declare it as a flag+env var,

Ahh, I missed that somehow. Updating it now!

agent/run_job.go
scadu: I removed the sleep in recent commit.

agent/run_job.go
```diff
@@ -428,6 +433,50 @@ func (r *JobRunner) cleanup(ctx context.Context, wg *sync.WaitGroup, exit core.P
...
+		// Process has started, we can now safely wait for completion
+	case <-ctx.Done():
+		r.agentLogger.Debug("[JobRunner] Context cancelled before process started, skipping wait")
+		return
+	}
+
+	gracePeriod := r.conf.AgentConfiguration.KubernetesLogCollectionGracePeriod
+
+	waitCtx := ctx
+	if gracePeriod > 0 {
```
DrJosh9000: What if, for some reason, someone wanted to disable the grace period? I think `0` should probably mean `0`. What about:


```suggestion
	if gracePeriod >= 0 {
```

clicommand/agent_start.go
```diff
@@ -180,10 +180,11 @@ type AgentStartConfig struct {
 	TracingPropagateTraceparent bool   `cli:"tracing-propagate-traceparent"`
 
 	// Other shared flags
-	StrictSingleHooks         bool   `cli:"strict-single-hooks"`
-	KubernetesExec            bool   `cli:"kubernetes-exec"`
-	TraceContextEncoding      string `cli:"trace-context-encoding"`
-	NoMultipartArtifactUpload bool   `cli:"no-multipart-artifact-upload"`
+	StrictSingleHooks                  bool   `cli:"strict-single-hooks"`
+	KubernetesExec                     bool   `cli:"kubernetes-exec"`
+	KubernetesLogCollectionGracePeriod string `cli:"kubernetes-log-collection-grace-period"`
```
DrJosh9000: This should "just work" if its type was `time.Duration`. I'm not sure why some other duration flags go via a string and then manually parse it. The config loader supports `time.Duration`. 

```suggestion
	KubernetesLogCollectionGracePeriod time.Duration `cli:"kubernetes-log-collection-grace-period"`
```

DrJosh9000 (approved):
Thanks for this!

DrJosh9000 (approved):
✅

PR #3507, opened 2025-10-02
fix: check for changes on main if committing to main

mcncl:
### Description

Currently we don't seem to check for a diff if committing directly to `main`; we compare the current commit to `HEAD` which may therefore evaluate to no changes.

### Context

Plain: https://app.plain.com/workspace/w_01J0T5SBKJHGBT6FZZHAMH3GRB/thread/th_01K6H416REM51XCQ4H40708308/

### Changes

Checks using `...HEAD` instead so that we can check when `HEAD` and `diffBase` point to the same branch

### Testing

### Disclosures / Credits

I used AI to point at whether or not our current testing would cover this; it concluded that it would but that it might be best to add a test, too. 

It pointed out that we would need a `.git` repository present for the test to work, and added comments covering that/added a `skip` if we don't have one present so that tests didn't fail due to that.

### DIFF
--- a/clicommand/pipeline_upload.go
+++ b/clicommand/pipeline_upload.go
@@ -613,7 +613,7 @@ func (cfg *PipelineUploadConfig) parseAndInterpolate(ctx context.Context, src st
 func gatherChangedFiles(diffBase string) (mergeBase string, changedPaths []string, err error) {
 	// Use three-dot syntax to find the merge base and diff in one command.
 	// This handles both feature branches and commits directly on main.
-	gitDiff, err := exec.Command("git", "diff", "--name-only", diffBase+"...HEAD").Output()
+	gitDiff, err := exec.Command("git", "diff", "--name-only", diffBase+"...HEAD~1").Output()
 	if err != nil {
 		return "", nil, gitDiffError{mergeBase: diffBase, wrapped: err}
 	}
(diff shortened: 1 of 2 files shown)


clicommand/pipeline_upload.go
```diff
@@ -613,17 +611,20 @@ func (cfg *PipelineUploadConfig) parseAndInterpolate(ctx context.Context, src st
 
 // gatherChangedFiles determines changed files in this build.
 func gatherChangedFiles(diffBase string) (mergeBase string, changedPaths []string, err error) {
-	// The --merge-base flag was only added to git-diff recently.
+	// Use three-dot syntax to find the merge base and diff in one command.
+	// This handles both feature branches and commits directly on main.
+	gitDiff, err := exec.Command("git", "diff", "--name-only", diffBase+"...HEAD").Output()
```
DrJosh9000: I'm curious to understand how exactly this addresses the problem (where the commit is directly to main). If `diffBase == "main"` and `HEAD` is the head of `main`, then `git diff --name-only main...HEAD` is still nothing, right?

nzspambot: @DrJosh9000 I agree, we had to use HEAD~1 for this to best on main and branches

mcncl: Whoops typo

DrJosh9000: Won't `diffBase+"...HEAD~1"` break the usual application (when run on a commit not yet merged)?

We may need to change the approach, e.g. make the default `diffBase` be `HEAD~1` when the usual default resolves to the same commit as `HEAD`.

DrJosh9000:
I took a crack at a different approach here: https://github.com/buildkite/agent/pull/3510

PR #3510, opened 2025-10-06
Improve if_changed when base=HEAD

DrJosh9000:
### Description

Make `if_changed` detect when the diff base is the same commit as `HEAD` (e.g. on a build on `main` after merging a PR, or when committing directly to `main`). When that happens, extract the list of changed files from the most recent commit, which requires some fun `git log` flags.

### Context

Closes #3507. There are also some other reports.

### Changes

`git rev-parse` both `diffBase` and `HEAD`, and if they are the same, use `git log --first-parent -1 --name-only --pretty='format:'` to get a list of changed files instead of `git diff`.

### Testing

### Disclosures / Credits

I did not use AI tools at all

### DIFF
--- a/clicommand/pipeline_upload.go
+++ b/clicommand/pipeline_upload.go
@@ -612,24 +612,88 @@ func (cfg *PipelineUploadConfig) parseAndInterpolate(ctx context.Context, src st
 }
 
 // gatherChangedFiles determines changed files in this build.
-func gatherChangedFiles(diffBase string) (mergeBase string, changedPaths []string, err error) {
-	// The --merge-base flag was only added to git-diff recently.
-	mergeBaseOut, err := exec.Command("git", "merge-base", diffBase, "HEAD").Output()
+func gatherChangedFiles(l logger.Logger, diffBase string) (changedPaths []string, err error) {
+	// Corporate needs you to find the differences between diffBase and HEAD.
+	diffBaseCommit, err := exec.Command("git", "rev-parse", diffBase).Output()
 	if err != nil {
-		return "", nil, gitMergeBaseError{diffBase: diffBase, wrapped: err}
+		return nil, gitRevParseError{arg: diffBase, wrapped: err}
 	}
-	mergeBase = strings.TrimSpace(string(mergeBaseOut))
+	headCommit, err := exec.Command("git", "rev-parse", "HEAD").Output()
+	if err != nil {
+		return nil, gitRevParseError{arg: "HEAD", wrapped: err}
+	}
+	if strings.TrimSpace(string(diffBaseCommit)) == strings.TrimSpace(string(headCommit)) {
+		// They're the same commit, so `git diff` will (correctly) report no
+		// changes between them.
+		// This often happens for builds triggered on main, when:
+		// * a PR is merged. HEAD is hopefully either a merge commit or a
+		//   squash commit.
+		// * the developers commit and push directly to main.
+		// As a heuristic, we look at changed files between the latest commit
+		// and its parent (the first parent, if it is a merge commit).
+		// If _multiple_ commits were pushed at once, then this approach will
+		// miss changes from earlier commits. Thus, log a warning.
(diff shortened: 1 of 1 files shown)


clicommand/pipeline_upload.go
```diff
@@ -613,24 +613,73 @@ func (cfg *PipelineUploadConfig) parseAndInterpolate(ctx context.Context, src st
...
-
-	gitDiff, err := exec.Command("git", "diff", "--name-only", mergeBase).Output()
+	headCommit, err := exec.Command("git", "rev-parse", "HEAD").Output()
 	if err != nil {
-		return mergeBase, nil, gitDiffError{mergeBase: mergeBase, wrapped: err}
+		return "", nil, gitRevParseError{arg: "HEAD", wrapped: err}
+	}
+	if strings.TrimSpace(string(diffBaseCommit)) == strings.TrimSpace(string(headCommit)) {
+		// They're the same commit, so `git diff` will (correctly) report no
+		// changes between them.
```
zhming0, line 627: It might be worth briefly document when this will happen.

clicommand/pipeline_upload.go
```diff
@@ -613,24 +613,73 @@ func (cfg *PipelineUploadConfig) parseAndInterpolate(ctx context.Context, src st
...
+		// Instead, we can use a git log invocation to get changed files from
+		// the commit, including merge commits. Flag Explainer:
+		// `--first-parent`: when reaching a merge commit, follow the first parent
+		// (the branch that was merged *into*). Implies --diff-merges=first-parent,
+		// which is important here.
+		// `-1`: show only the most recent entry from the log
+		// `--name-only`: show only the name of each changed file in the post-image tree
+		// `--pretty='format:'`:
+		gitLog, err := exec.Command("git", "log", "--first-parent", "-1", "--name-only", "--pretty='format:'").Output()
```
zhming0: `--first-parent` and `-1` is so obscure 😵 but it makes sense.

clicommand/pipeline_upload.go
```diff
@@ -613,24 +613,73 @@ func (cfg *PipelineUploadConfig) parseAndInterpolate(ctx context.Context, src st
...
+		if err != nil {
+			return "", nil, gitLogError{wrapped: err}
+		}
+		changedPaths = strings.Split(string(gitLog), "\n")
```
zhming0, line 650: Do we want to assign `mergeBase`? or it means nothing because the `HEAD` does not intend to be merged.

zhming0 (approved):
Looks good to me with a few questions.

clicommand/pipeline_upload.go
DrJosh9000: Good point. Rather than pretend there is a merge base, I've changed `gatherChangedFiles` to be responsible for logging what it is doing.

clicommand/pipeline_upload.go
DrJosh9000: Done!

clicommand/pipeline_upload.go
DrJosh9000: Git is complicated!

zhming0: I have one extra question that may not block this:

If people have a multi-commits and then eventually merged with fast forward without a merge commit. It feels that this will break? 

Or if people push 3 commits into `main` in a single push?

DrJosh9000: Yes, definitely. I don't think there's enough information to produce the right diff in those cases by default. If the merge/push is coming from a PR I wonder if we could find out the parent commit from the webhook. Since Git is so flexible in how it can be used, I think the user is responsible for setting a diff base when going off the beaten path.

DrJosh9000: I've upgraded the log in this case to be a warning, and added some more comments.

DrJosh9000:
Played with it for a bit, uncovered a bug, fixed it, adjusted the logging a little more. Happy.

PR #3518, opened 2025-10-08
if_changed: support lists, include/exclude

DrJosh9000:
### Description

Make `if_changed` more flexible:

- Accept either a single pattern or a list of patterns, so you don't have to wrap everything in `{,}`.
- Accept an exclusion/list of exclusions.

### Examples

```yaml
# Existing style
if_changed: "pattern"

# New style 1: list of patterns. If any match, the step is run.
if_changed:
  - "pattern1"
  - "pattern2"

# New style 2: include/exclude. Changed paths matching `exclude` are
# eliminated before checking `include`.
if_changed:
  include: "pattern1"
  exclude: "pattern2"

# New style 3: include and exclude can also be lists. Changed paths matching
# any `exclude` pattern are eliminated before checking `include`.
if_changed:
  include: 
    - "pattern1"
    - "pattern2"
  exclude: 
    - "pattern3"
    - "pattern4"
```

### Context

https://linear.app/buildkite/issue/PS-1254

### Changes

- Accept lists of glob patterns for `if_changed`. If any of them match a changed path, the step should run.
- Accept a mapping with keys `include` and `exclude`, which can be either a string or a list of strings. If any of `exclude` match a changed path, that path is then excluded from `if_changed`.

### Testing

### Disclosures / Credits

I did not use AI tools at all

### DIFF
--- a/clicommand/pipeline_upload.go
+++ b/clicommand/pipeline_upload.go
@@ -70,7 +70,7 @@ Example:
     $ buildkite-agent pipeline upload my-custom-pipeline.yml
     $ ./script/dynamic_step_generator | buildkite-agent pipeline upload`
 
-const ifChangedSkippedMsg = "if_changed pattern did not match any paths changed in this build"
+const ifChangedSkippedMsg = "if_changed matched no unexcluded changed files in this build"
 
 const defaultGitDiffBase = "origin/main"
 
@@ -748,13 +748,20 @@ stepsLoop:
 			content = st.Contents
 		}
 
-		// Retrieve the value then delete it.
-		// It's not yet understood by the backend.
-		ic := content["if_changed"]
-		if ic == nil {
+		// If content is nil then there's nothing containing an `if_changed`.
+		if content == nil {
 			continue
 		}
+
+		// Retrieve the if_changed value, then delete it.
+		// It is not yet understood by the backend.
+		ifChangedValue := content["if_changed"]
 		delete(content, "if_changed")
+
+		// If there's no if_changed, then the step is unconditional.
+		if ifChangedValue == nil {
+			continue
+		}
 
 		if !ica.enabled {
 			// Not applying the if_changed; leaving it deleted.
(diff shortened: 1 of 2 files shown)


clicommand/pipeline_upload.go
```diff
@@ -802,28 +811,81 @@ stepsLoop:
...
-		if err != nil {
-			l.Warn("if_changed value %q couldn't be parsed as a glob pattern: %v\n"+
-				"The step will not be skipped.", pattern, err)
-			continue
+		exclude := ifChangedPatterns(l, ifChangedExcludeValue, "if_changed_exclude")
+
+		changedPaths := ica.changedPaths
+		if len(exclude) > 0 {
+			// First, remove paths that are excluded.
+			changedPaths = slices.DeleteFunc(slices.Clone(ica.changedPaths), func(cp string) bool {
```
wolfeidau: Do you need to clone the list of changed paths every time?

clicommand/pipeline_upload.go
```diff
@@ -802,28 +811,81 @@ stepsLoop:
...
+	case string:
+		patterns = []string{x}
+
+	case []any:
+		for i, item := range x {
+			patt, ok := item.(string)
+			if !ok {
+				l.Warn("Each %s value must be a string containing a glob pattern, but the item at index %d was a %T\n"+
+					"The step will not be skipped.", name, i, item)
+				return nil
```
wolfeidau: what does `nil` result in? does it skip the step or run it?

wolfeidau (changes requested):
Just a couple of observations/questions.

clicommand/pipeline_upload.go
DrJosh9000: Only because `slices.DeleteFunc` alters its input. I'll rearrange it to avoid the clone.

clicommand/pipeline_upload.go
DrJosh9000: `nil` - it runs the step. This particular code path would happen if someone wrote:

```yaml
if_changed:
  - something
  - null
  - something/else
```

DrJosh9000: I realised that the effect would be different between `if_changed` and `if_changed_exclude`, so I changed it to report success to the caller so that the log is correct.

PR #3521, opened 2025-10-12
Configurable chunks interval

catkins:
### Description

To give us more control over log chunk, I've added an additional field to the agent job representation to allow us to remotely dictate the log chunk interval.

This PR accepts that field and uses it to determine the chunk ticker interval at the beginning of the job.

### Testing

### Disclosures / Credits

I wrote the app code, and had `amp` generate the test, which is kind of a crap test, but this codepath didn't have great test plumbing around it, and needs a bit of a refactor to make it easier to test I think. I'm keen to hear ideas from @buildkite/agent-stewards of how I can pin this down a bit more (mostly around the jitter, and first chunk plumbing) to make the tests more closer to the intent of the change.

### DIFF
--- a/agent/integration/job_runner_integration_test.go
+++ b/agent/integration/job_runner_integration_test.go
@@ -76,7 +76,6 @@ func TestPreBootstrapHookScripts(t *testing.T) {
 	}
 
 	for _, tc := range testCases {
-
 		t.Run(tc.name, func(t *testing.T) {
 			t.Parallel()
 			ctx := context.Background()
@@ -361,13 +360,13 @@ func TestJobRunnerIgnoresPipelineChangesToProtectedVars(t *testing.T) {
 func TestChunksIntervalSeconds_ControlsUploadTiming(t *testing.T) {
 	t.Parallel()
 
-	runTestWithInterval := func(t *testing.T, intervalSeconds int) time.Duration {
+	runTestWithInterval := func(t *testing.T, intervalSeconds int) int {
 		t.Helper()
 		ctx := context.Background()
 
 		var (
-			uploadTimes []time.Time
-			mu          sync.Mutex
+			chunkCount int
+			mu         sync.Mutex
 		)
 
 		e := createTestAgentEndpoint()
@@ -376,7 +375,7 @@ func TestChunksIntervalSeconds_ControlsUploadTiming(t *testing.T) {
 			Path:   "/jobs/{id}/chunks",
 			HandlerFunc: func(rw http.ResponseWriter, req *http.Request) {
 				mu.Lock()
-				uploadTimes = append(uploadTimes, time.Now())
+				chunkCount++
 				mu.Unlock()
 				e.chunksHandler()(rw, req)
 			},
@@ -385,7 +384,7 @@ func TestChunksIntervalSeconds_ControlsUploadTiming(t *testing.T) {
 
 		j := &api.Job{
 			ID:                    defaultJobID,
-			ChunksMaxSizeBytes:    10240,
+			ChunksMaxSizeBytes:    100_000, // large number that will never get divided into multiple chunks
 			ChunksIntervalSeconds: intervalSeconds,
 			Env:                   map[string]string{},
 			Token:                 "bkaj_job-token",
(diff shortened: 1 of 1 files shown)


agent/integration/job_runner_integration_test.go
```diff
@@ -354,3 +357,96 @@ func TestJobRunnerIgnoresPipelineChangesToProtectedVars(t *testing.T) {
...
+			ChunksMaxSizeBytes:    10240,
+			ChunksIntervalSeconds: intervalSeconds,
+			Env:                   map[string]string{},
+			Token:                 "bkaj_job-token",
+		}
+
+		mb := mockBootstrap(t)
+		mb.Expect().Once().AndCallFunc(func(c *bintest.Call) {
+			start := time.Now()
+			for time.Since(start) < 10*time.Second {
```
moskyb: TIL this construct. handy!

agent/integration/job_runner_integration_test.go
```diff
@@ -354,3 +357,96 @@ func TestJobRunnerIgnoresPipelineChangesToProtectedVars(t *testing.T) {
...
+		avg2s := runTestWithInterval(t, 2)
+
+		if avg2s <= avg1s {
+			t.Errorf("2s interval average (%v) should be longer than 1s interval average (%v)", avg2s, avg1s)
+		}
+
+		ratio := float64(avg2s) / float64(avg1s)
+		t.Logf("Ratio of 2s/1s intervals: %.2f", ratio)
+
+		if ratio < 1.4 {
```
moskyb: why 1.4? we're not jittering the interval internally, so shouldn't the interval on the 2s always be 2x longer than the interval on 1s? i suppose the chunk collector will sometimes collect more than its size limit and preempt the serverside timeout?

agent/integration/job_runner_integration_test.go
```diff
@@ -354,3 +357,96 @@ func TestJobRunnerIgnoresPipelineChangesToProtectedVars(t *testing.T) {
...
+		}
+		avgInterval := total / time.Duration(len(intervals))
+
+		t.Logf("Interval %ds: average upload interval = %v (from %d intervals)", intervalSeconds, avgInterval, len(intervals))
+		return avgInterval
+	}
+
+	t.Run("2s interval should be longer than 1s interval", func(t *testing.T) {
+		avg1s := runTestWithInterval(t, 1)
+		avg2s := runTestWithInterval(t, 2)
```
moskyb: this test suite is currently pretty quick — 1.6 seconds on my machine at current. This test needs(?) to wait for 10s to complete, but i think we should probably do both of these runs concurrently to minimise the amount of time we're lengthening this suite by. something like:
```Go
		times := map[int]time.Duration{}
		wg := sync.WaitGroup{}
		wg.Add(2)
		
		go func() {
			defer wg.Done()
			times[1] = runTestWithInterval(t, 1)
		}()
		go func() {
			defer wg.Done()
			times[2] = runTestWithInterval(t, 2)
		}()

		wg.Wait()

		avg1s := times[1]
		avg2s := times[2]
```

agent/integration/job_runner_integration_test.go
catkins: i reckon

agent/integration/job_runner_integration_test.go
catkins: yeah good idea

agent/integration/job_runner_integration_test.go
catkins: yeah 1.4 is a bit of a weird number... I had a bit of a struggle with testing this tbh.

that code in `run_job.go` `streamJobLogsAfterProcessStart` has a few funky interactions

- the jitter sleep is up-to the `processInterval`, using `rand.N`
- the jitter is applied after the ticker
- the ticker will keep ticking on its schedule irrespective of the jitter
- we also have this `first` channel to immediately try and push out a chunk without waiting

this all kinda means that the intervals themselves are kinda all over the joint depending on the dice-roll of the jitter itself which can be pathologically lumpy

eg. with 1s interval

| Ticker | Jitter | Final Time | Time since last |
| -     | -    | -    |     -|
| 1.0s  | 0.9s | 1.9s | N/A  |
| 2.0s  | 0.1s | 2.1s | 0.2s |
| 3.0s  | 0.9s | 3.9s | 1.8s |
| 4.0s  | 0.1s | 4.1s | 0.2s |
| 5.0s  | 0.5s | 5.5s | 1.4s |


Ideally it would be straightforward to control or disable the jitter in the tests too, but that also requires some surgery or exposing internals that aren't very integration test-ey which I don't love either. Sigh.

PR #3522, opened 2025-10-13
PS-1245: Fix artifact search output format escape sequence handling

zhming0:
### Description

Fix a long standing bug in `artifact search` command that this never worked 

```
buildkite-agent artifact search "..." -format "%p\n"
```

The output would be: 
```
artifact-1\nartifact-2
```
 instead of: 
```
artifact-1
artifact-2
```

It hasn't been an issue because our default format is hardcoded in golang using string literal: `%j %p %c\n`, so the `\n` is parsed and correctly registered as escape sequence by the compiler. 

This PR fixes all that by using `strconv.Unquote`.

### Context

PS-1245

### Testing

### Disclosures / Credits

Peasant hand written

### DIFF
--- a/clicommand/artifact_search.go
+++ b/clicommand/artifact_search.go
@@ -109,7 +109,7 @@ var ArtifactSearchCommand = cli.Command{
 		},
 		cli.StringFlag{
 			Name:  "format",
-			Value: "%j %p %c\n",
+			Value: "%j %p %c\\n", // Note: users supply \n in the flag value literally
 			Usage: "Output formatting of results. See below for listing of available format specifiers.",
 		},
 	}),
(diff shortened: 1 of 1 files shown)


clicommand/artifact_search.go
```diff
@@ -109,7 +109,7 @@ var ArtifactSearchCommand = cli.Command{
 		},
 		cli.StringFlag{
 			Name:  "format",
-			Value: "%j %p %c\n",
+			Value: `%j %p %c\n`, // use backtick as that is what user would use
```
DrJosh9000: I don't understand this comment, as written. I think you mean "backslash" (`\`) rather than "backtick" (`` ` ``). In this case to make the literalness of `\n` stand out, I think we should do this instead:

```suggestion
			Value: "%j %p %c\\n", // Note: users supply \n in the flag value literally
```

clicommand/artifact_search.go
```diff
@@ -118,6 +118,18 @@ var ArtifactSearchCommand = cli.Command{
 		ctx, cfg, l, _, done := setupLoggerAndConfig[ArtifactSearchConfig](ctx, c)
 		defer done()
 
+		printFormat := cfg.PrintFormat
+		if strings.Contains(printFormat, `"`) {
+			// Otherwise this would break the strconv.Unquote
+			printFormat = strings.ReplaceAll(printFormat, `"`, `\"`)
+		}
+		// Handling all escape sequences, like \n, \t etc
+		unquoted, err := strconv.Unquote("\"" + printFormat + "\"")
```
DrJosh9000: Use `` `"` `` same as above?

```suggestion
		unquoted, err := strconv.Unquote(`"` + printFormat + `"`)
```

zhming0: Good idea.

clicommand/artifact_search.go
zhming0: That's better 👍🏿

PR #3534, opened 2025-10-16
Review of signal handling and grace periods

DrJosh9000:
### Description

Various changes related to signals, signal handling, and grace periods.

### Context

https://linear.app/buildkite/issue/PS-1220

https://linear.app/buildkite/issue/PS-1069

#### Signal vs cancel grace period

When a job is cancelled (either through the UI, or when the agent is ctrl-C-ed twice), the timeline is meant to go like this (I think):

- `t0`: Cancellation begins. The job runner is told to `Cancel`, which sends an `Interrupt` to the job "process". Which is either a `process` (and the bootstrap is in a proper subprocess of the agent) or the Kubernetes runner (the bootstrap is in another container in the pod). 
  - When it's a subprocess, this is a `SIGTERM` to the process group of the subprocess.
  - When it's Kubernetes, it changes the state to `RunStateInterrupt` which the client side should notice in the next poll.
  Either way, the bootstrap should handle a `SIGTERM` and wrap itself up within `signalGracePeriod`.
- `t0 + x < signalGracePeriod`: Happy path. The subprocess/other container has wrapped up and we can mark the job finished (and disconnect if the agent is stopping).
- `t0 + signalGracePeriod`: Not happy path. Now we `Terminate` (send a `SIGKILL`) to the subprocess. There's no equivalent for Kubernetes, since we can't `SIGKILL` something in a different PID namespace, but we can pretend it has died and move on.
- `t0 + (signalGracePeriod + x) < cancelGracePeriod`: Bittersweet path - the subprocess was killed, but we uploaded the logs and marked it finished.
- `t0 + cancelGracePeriod`: Bad end - we need to wrap it up now. There's no more time to finish gracefully. Go to the next job if not stopping, or exit the agent if stopping.

The first problem is that the job runner's `Cancel` implementation diverged from `process`, and sent `SIGKILL` after `cancelGracePeriod`, not after `signalGracePeriod`. So I changed it to wait for `signalGracePeriod` instead.

#### Kubernetes

Kubernetes has the concept of `terminationGracePeriodSeconds`, which implements the same thing as `signalGracePeriod` above, but it happens when the pod is deleted. It also sends `SIGTERM` to every container in the pod. 

Currently, on `SIGTERM` and others, the `kubernetes-bootstrap` cancels the subprocess context, which _also_ runs through the `Interrupt`-`signalGracePeriod`-`Terminate` dance. Does this mean the bootstrap receives `SIGTERM` twice? If so, the `bootstrap` signal handler removes itself after the first signal, causing the second signal to be handled according to the normal Go signal handling, which kills it before the grace period expires.

We need to listen to one or the other, and I'm choosing to favour the signal passed over the agent socket. Thus `SIGTERM` and friends will be swallowed by `kubernetes-bootstrap`. 

#### Ctrl-C

The double-Ctrl-C behaviour is useful when running the agent in a local terminal. But the code path in the `start` signal handler is shared between `SIGINT` (default Ctrl-C signal) and `SIGTERM`.

This means that the Kubernetes-delivered `SIGTERM` causes the agent to enter graceful exit mode, waiting for the job to complete. So the `agent` container never starts ungracefully cancelling the job. We're back to 1 `SIGTERM` for the `bootstrap` container!

Thus on Kubernetes mode, I'm choosing to go straight to ungraceful cancellation on `SIGINT/SIGTERM`. 

Bonus 1: I've implemented a third Ctrl-C to really exit the agent immediately.
Bonus 2: I set up a struct to pass the configuration for the signal handler rather than add another arg.

#### Graceful contexts

But back to the `start` signal handler current implementation. On the second `SIGINT/SIGTERM` it does this:

1. Create a new graceful context whose parent is `ctx` and is cancelled 1 cancellation grace period after `ctx` is done.
2. Asynchronously: 
    1. send the worker pool a `Stop(false)`
    1. Sleep half the cancel grace period 
    2. and then call the cancel function for `ctx`
3. Wait for the graceful context and then exit 1.

I recall that it seemed reasonable at the time, but the effect is that halfway through the cancel grace period, which is usually only 1 second more than the signal grace period, the agent _loses the ability to upload logs or finish the job_.

It's a complicated but effective footgun. I took it out and replaced it with a much simpler dead-mans-switch that waits the full cancel grace period before exiting.

#### What about `KubernetesLogCollectionGracePeriod`?

Good question. I think #3500 didn't actually change what it intended to change.

`KubernetesLogCollectionGracePeriod` changes the time waited in the function `waitForKubernetesProcessesToComplete`, which is called in the job runner's `cleanup`. `cleanup` is deferred in the job runner's `Run`, which to cut a long story short, only starts returning (and running defers) after the "process" has already exited. So it should always see the process has finished, and never wait.

What we probably need to fix in the k8s case is the default `signalGracePeriod` (probably via the 1-second-before-`cancelGracePeriod` computation). If the default `terminationGracePeriodSeconds` is 60, then we should consider setting `BUILDKITE_CANCEL_GRACE_PERIOD` to around 60 too. This will mean changing agent-stack-k8s. But here, I've just ripped out the bulk of #3500, leaving the flag for backwards compatibility.

#### Bootstrap self-signaling

When the bootstrap receives one of the signals it handles in a special way, it registers a cancellation and removes the custom signal handler, then once the cancellation is complete, it re-signals itself with the signal so that the default signal handler handles the signal. The expected behaviour is that it exits with status 128+signal. Go's default behaviour on SIGQUIT is to panic and exit with status 2. Since there seems to be no way to "exit due to SIGQUIT but without a panic" in Go, I've turned it into a manual "exit 131".

### Testing

### Disclosures / Credits

I did not use AI tools at all

### DIFF
--- a/clicommand/bootstrap.go
+++ b/clicommand/bootstrap.go
@@ -568,6 +568,13 @@ var BootstrapCommand = cli.Command{
 		// If cancelled and our child process returns a non-zero, we should terminate
 		// ourselves with the same signal so that our caller can detect and handle appropriately
 		if cancelled && runtime.GOOS != "windows" {
+			// Per https://pkg.go.dev/os/signal:
+			// "A SIGQUIT, SIGILL, SIGTRAP, SIGABRT, SIGSTKFLT, SIGEMT, or
+			// SIGSYS signal causes the program to exit with a stack dump."
+			// Of these, `received` can only be SIGQUIT.
+			if received == syscall.SIGQUIT {
+				return &SilentExitError{code: 131} // 128 + 3 (SIGQUIT).
+			}
 			if err := signalSelf(l, received); err != nil {
 				l.Error("Failed to signal self: %v", err)
 			}


agent/run_job.go
```diff
@@ -571,13 +522,13 @@ func (r *JobRunner) Cancel() error {
 
 	reason := ""
 	if r.stopped {
-		reason = "(agent stopping)"
+		reason = " (agent stopping)"
```
wolfeidau: is this extra space intended?

agent/run_job.go
```diff
@@ -571,13 +522,13 @@ func (r *JobRunner) Cancel() error {
...
 	}
 
 	r.agentLogger.Info(
-		"Canceling job %s with a grace period of %ds %s",
+		"Canceling job %s with a signal grace period of %v%s",
```
wolfeidau: is this intential spacing wise?

clicommand/kubernetes_bootstrap.go
```diff
@@ -164,6 +164,9 @@ var KubernetesBootstrapCommand = cli.Command{
 			if err != nil {
 				l.Error("Error waiting for client interrupt: %v", err)
 			}
+			// The context cancellation handler in process.Run first calls
+			// Interrupt, waits for its signalGracePeriod, and then calls
+			// Terminate.
```
wolfeidau, line 172: should we log something for this cancel, we have a few things happening at the same time here with go funcs ect, would be good to log so we kind of know what happend when. 

Staring at this code I don't know what will happen when so maybe some log messages would help.

clicommand/kubernetes_bootstrap.go
```diff
@@ -189,29 +192,31 @@ var KubernetesBootstrapCommand = cli.Command{
...
-			// Forward signals to the subprocess.
 			for {
 				select {
 				case <-ctx.Done():
 					return
 				case <-proc.Done():
 					return
 				case <-signals:
-					proc.Interrupt()
+					// swallow the signal
```
wolfeidau: maybe log this, otherwise we may never know it happened?

wolfeidau (changes requested):
Some observations, not sure how helpful they are.

agent/run_job.go
DrJosh9000: I had intended to shift the space from the log format to the reason, so that if reason is empty, it doesn't log a line with a space on the end. It's overly pedantic so I reverted it.

agent/run_job.go
DrJosh9000: Per the above.

clicommand/kubernetes_bootstrap.go
DrJosh9000: Done, I've also added a sleep + exit similar to the non-k8s case.

clicommand/kubernetes_bootstrap.go
DrJosh9000: Done!

PR #3543, opened 2025-10-20
Add --literal and --delimiter flags to artifact upload

### LINKED ISSUE #3535: [zz-glob regression] No pattern option for buildkite-agent artifact upload
Following this [PR](https://github.com/buildkite/agent/pull/3428/files), which was later included in `3.104.0`, there looks to be some regression with glob pattern matching when using `buildkite-agent artifact upload` with special characters within filenames, more specifically in this case with `~`.

This has been specifically raised as an issue here: https://forum.buildkite.community/t/no-pattern-option-for-buildkite-agent-artifact-upload/4527

It would be nice if there was a --no-pattern or --literal option to make filenames be interpreted literally.

Original command in pipeline:

`(cd debian/build/ && find -type f -exec buildkite-agent artifact upload {} \;)`
New command in pipeline:

`(cd debian/build/ && find -type f -print0 | sed -Ee 's/([]?*{}~[])/\\\1/g' | xargs -0 -L 1 -- buildkite-agent artifact upload)`

**Additional context**

DrJosh9000:
### Description

Provide easier ways to specify file paths for uploading when they could be unintentionally parsed as globs, or they use the default splitting delimiter (`;`).

### Context

Fixes #3535

### Changes

- Add `--literal` flag; when set, send file paths directly to the next stage
- Add `--delimiter` flag; use its value to split the upload paths into multiple strings (or set to empty to disable splitting)

### Testing

### Disclosures / Credits

100% free range human

### DIFF
--- a/internal/artifact/uploader.go
+++ b/internal/artifact/uploader.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io"
 	"io/fs"
+	"iter"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -26,10 +27,7 @@ import (
 	"github.com/dustin/go-humanize"
 )
 
-const (
-	ArtifactPathDelimiter    = ";"
-	ArtifactFallbackMimeType = "binary/octet-stream"
-)
+const ArtifactFallbackMimeType = "binary/octet-stream"
 
 type UploaderConfig struct {
 	// The ID of the Job
@@ -49,8 +47,11 @@ type UploaderConfig struct {
 	TraceHTTP    bool
 	DisableHTTP2 bool
 
-	// When true, disables parsing Paths as globs.
+	// When true, disables parsing Paths as globs; treat each path literally.
 	Literal bool
+
+	// The delimiter used to split Paths into multiple paths/globs.
+	Delimiter string
 
 	// Whether to follow symbolic links when resolving globs
 	GlobResolveFollowSymlinks bool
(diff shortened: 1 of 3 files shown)


internal/artifact/uploader.go
```diff
@@ -195,17 +205,38 @@ type artifactCollector struct {
 	artifacts []*api.Artifact
 }
 
+func (a *Uploader) paths() iter.Seq[string] {
+	if a.conf.Delimiter == "" {
+		// Yield it once
```
wolfeidau: What does yeilding it once do?

internal/artifact/uploader_test.go
```diff
@@ -79,6 +79,7 @@ func TestCollect(t *testing.T) {
 			filepath.Join("fixtures", "**/*.jpg"),
 			filepath.Join(root, "fixtures", "**/*.gif"),
 		),
+		Delimiter: ";",
```
wolfeidau, line 84: Do we need a test for `Literal` to show how it effects things?

internal/artifact/uploader.go
DrJosh9000: I've changed it to be a bit more semantic (`slices.Values([]string{a.conf.Paths})`).

internal/artifact/uploader_test.go
DrJosh9000: Added!

wolfeidau (approved):
👍🏻 🚀

PR #3546, opened 2025-10-20
PS-1302: refresh checkout root file handle after checkout hook

### LINKED ISSUE #3530: Vendored plugin path cannot be found: no such file or directory
# Vendored Plugin Validation Fails Intermittently 

## Summary

After upgrading from buildkite-agent v3.103.0.10530 to newer versions (tried at 3.107.0 and 3.109.0), vendored plugins fail validation intermittently with the error:

```
Vendored plugin path "./buildkite/plugins/my-plugin" must be a directory within the checked-out repository: statat ./buildkite/plugins/my-plugin: no such file or directory
```

The directory **definitely exists** and the same job succeeds when retried immediately on the same agent.

## Example Logs

**Failing case** (checkout hook succeeds but plugin validation fails):
```
Running agent checkout hook
# ... checkout succeeds after retries ...
# ... 
Error: Vendored plugin path "./buildkite/plugins/my-plugin" must be a directory within the checked-out repository: statat ./buildkite/plugins/my-plugin: no such file or directory
```

**Passing case** (immediately after on same agent):
```
Running agent checkout hook  
# ... same checkout process ...
Running plugin my-plugin command hook
# Success
```

## Suspected Cause

The issue appears to be related to commit https://github.com/buildkite/agent/pull/3493

The commit added fallback logic that tries to create `checkoutRoot` when checkout is skipped, but the implementation may have timing or path resolution issues that cause vendored plugin validation to fail intermittently.

zhming0:
### Description

We have false negative `Vendored plugin path "..." must be a directory within the checked-out repository` error, when a checkout global hook or plugin hook recreated the checkout directory, and as part the process if the hook inject some plugins/hooks into the directory.

Closes #3530

### Context

PS-1302

### Changes

Refresh `e.checkoutRoot` after all `checkout` behavior. 

### Testing

The QA is mostly relying on local testing. It highlight the need of comprehensive integration test suite.

### Disclosures / Credits

I am human

### DIFF
--- a/internal/job/checkout.go
+++ b/internal/job/checkout.go
@@ -95,6 +95,28 @@ func (e *Executor) createCheckoutDir() error {
 		}
 	}
 
+	if err := e.refreshCheckoutRoot(); err != nil {
+		return err
+	}
+
+	if e.shell.Getwd() != checkoutPath {
+		if err := e.shell.Chdir(checkoutPath); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// refreshCheckoutRoot refreshes e.checkoutRoot
+func (e *Executor) refreshCheckoutRoot() error {
+	checkoutPath, _ := e.shell.Env.Get("BUILDKITE_BUILD_CHECKOUT_PATH")
+	if e.checkoutRoot != nil {
+		if err := e.checkoutRoot.Close(); err != nil {
+			// While it's unlikely, it's not a blocking error
+			e.shell.Warningf("unable to close existing checkoutRoot during refreshCheckoutRoot: %w", err)
+		}
+	}
 	root, err := os.OpenRoot(checkoutPath)
 	if err != nil {
 		return fmt.Errorf("opening checkout path as root: %w", err)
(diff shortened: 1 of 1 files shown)


internal/job/checkout.go
```diff
@@ -257,43 +312,12 @@ func (e *Executor) CheckoutPhase(ctx context.Context) error {
...
-	}
-
-	if err := e.executeLocalHook(ctx, "post-checkout"); err != nil {
-		return err
-	}
-
-	if err := e.executePluginHook(ctx, "post-checkout", e.pluginCheckouts); err != nil {
+	// After everything, we need to refresh checkout root.
+	// This is because checkout hook might re-create the checkout root folder entirely, deprecating e.checkoutRoot.
+	if err := e.refreshCheckoutRoot(); err != nil {
```
zhming0, line 323: This is pretty much the actual change. 

All other diff are just moving codes around for better readability.

internal/job/checkout.go
```diff
@@ -113,6 +108,20 @@ func (e *Executor) createCheckoutDir() error {
 	return nil
 }
 
+// refreshCheckoutRoot refreshes e.checkoutRoot
+func (e *Executor) refreshCheckoutRoot() error {
+	checkoutPath, _ := e.shell.Env.Get("BUILDKITE_BUILD_CHECKOUT_PATH")
+	root, err := os.OpenRoot(checkoutPath)
```
DrJosh9000, line 120: Perhaps close the existing checkout root if there is one?
The potential double close on runtime cleanup shouldn't crash.

```suggestion
	if e.checkoutRoot != nil {
		e.checkoutRoot.Close()
	}
	root, err := os.OpenRoot(checkoutPath)
```

PR #3547, opened 2025-10-21
More agent stopping / job cancellation review

DrJosh9000:
### Description

Fix things found while testing #3534.

### Context

https://linear.app/buildkite/issue/PS-1220

### Changes

- Defer close the k8s socket in kubernetes-bootstrap.
- Fix the formatting of the cancellation grace period in the "Forcefully stopping" log message, so that `60s` doesn't print as `60000000000 seconds`.
- In `AcquireAndRunJob`, upon agent stop, only cancel the context of `AcquireJob`, not the context of `RunJob`. Cancelling the `RunJob` context gets in the way of cleanup steps (uploading final logs and finishing).
- Add a whole pile of comments here and there.
- Remove `CancelAndStop` and instead give a "reason" enum arg to `Cancel`.
- Make `cancelled`, `agentStopping`, and `jobRunner` atomics, just in case.
- Because `Cancel` can block while it waits for the process to terminate, `Stop` all workers in a pool simultaneously.
- Because `Stop` should block while it waits for workers to stop, it should be concurrent with the sleep.
- The `SIGQUIT` handler should also force exit after the grace period.
- In case stopping takes the whole cancel-grace-period, add 1 second to the force exit to allow for agent disconnect.
- Split `Stop` into `StopGracefully` and `StopUngracefully`.
 
### Testing

### Disclosures / Credits

I did not use AI tools at all

### DIFF
--- a/agent/agent_worker.go
+++ b/agent/agent_worker.go
@@ -391,7 +391,7 @@ func (a *AgentWorker) runPingLoop(ctx context.Context, idleMonitor *IdleMonitor)
 
 		switch action {
 		case "disconnect":
-			a.Stop(false)
+			a.StopUngracefully()
 			return nil
 
 		case "pause":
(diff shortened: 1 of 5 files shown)


agent/agent_worker.go
```diff
@@ -502,7 +503,15 @@ func (a *AgentWorker) runPingLoop(ctx context.Context, idleMonitor *IdleMonitor)
...
+// running. Whether or not Stop also cancels the current job depends on the
+// `graceful` argument.
+//
+//   - *Graceful* stop: allow the current job to finish without interruption
+//     (but don't accept new jobs) before exiting
+//   - *Ungraceful* stop: cancel the current job (jobRunner.CancelAndStop),
+//     wait a grace period for it to exit, and then exit ourselves
+//
+// In both cases, the ping loop should cease running.
 func (a *AgentWorker) Stop(graceful bool) {
```
moskyb: In a bunch of places around the code in this PR, we have code like `Stop(true /* graceful */)` and `Cancel(true /* the agent is stopping */)`, which feels like a bit of a smell that we have to document what the boolean means every time.

perhaps something like 
```Go
func (a *AgentWorker) GracefulStop() {
  a.Stop(false)
}

func (a *AgentWorker) HardStop() {
  a.Stop(true)
}
```

or something (with a similar approach with `jobRunner.Cancel()`) might make it a little clearer?

DrJosh9000: Good point, I've used `StopGracefully` and `StopUngracefully`, and changed `jobRunner.Cancel` to take an "enum".

PR #3548, opened 2025-10-21
PS-1295: inject OpenTelemetry context to all child processes

zhming0:
### Description

This PR addresses a our long standing OpenTelemetry gap, mentioned in #1663: we don't tell our Agent's child processes what our current span id and trace id is. It means our current OpenTelemetry integration only offer observability to job level, nothing more granular below that. 

We had this feature implemented for the deprecated opentracing backend a long time ago, this PR filled the gap for OpenTelemetry. 

This PR conforms to [OTel env carrier standard](https://opentelemetry.io/docs/specs/otel/context/env-carriers/), whenever Agent starts an child processes, it will try to propagate current OpenTel Span context to child processes via environment variables. 

This PR takes inspiration from Intercom's implementation https://github.com/intercom/buildkite-agent/pull/1, but aim to be more standard compliant and sustainable. 

### Context

PS-1295

### Testing

Screenshot 2025-10-21 at 4 31 40 PM

### Disclosures / Credits

Human

### DIFF
--- a/internal/shell/shell.go
+++ b/internal/shell/shell.go
@@ -28,6 +28,9 @@ import (
 	"github.com/buildkite/shellwords"
 	"github.com/gofrs/flock"
 	"github.com/opentracing/opentracing-go"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/propagation"
+	"go.opentelemetry.io/otel/trace"
 )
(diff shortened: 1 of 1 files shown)


catkins (approved):
Amazing! Thanks @zhming0

PR #3551, opened 2025-10-22
Add cache save and restore using github.com/buildkite/zstash

wolfeidau:
### Description

Adding buildkite cache support to the agent.

### Context

https://linear.app/buildkite/issue/MDC-723/add-the-restore-and-save-commands-to-the-buildkite-agent

### Changes

This change introduces some new sub commands.

* buildkite-agent cache save
* buildkite-agent cache restore

It is currently using some conventions from zstash which will change as it is integrated more with the agent.

This change is just adding these subcommands to the agent for use from a plugin.

### Testing

Currently most of the testing for this feature is in github.com/buildkite/zstash given this is just a minimal integration currently.

### Disclosures / Credits

Claude Code was used to write and review code in this PR.

### DIFF
--- a/clicommand/global.go
+++ b/clicommand/global.go
@@ -182,18 +182,6 @@ type APIConfig struct {
 	NoHTTP2          bool   `cli:"no-http2"`
 }
 
-// CacheConfig includes cache-related shared options for easy inclusion across
-// cache command config structs (via embedding).
-type CacheConfig struct {
-	Ids             string `cli:"ids"`
-	Registry        string `cli:"registry"`
-	BucketURL       string `cli:"bucket-url" validate:"required"`
-	Branch          string `cli:"branch" validate:"required"`
-	Pipeline        string `cli:"pipeline" validate:"required"`
-	Organization    string `cli:"organization" validate:"required"`
-	CacheConfigFile string `cli:"cache-config-file"`
-}
-
 func globalFlags() []cli.Flag {
 	return []cli.Flag{
 		NoColorFlag,
@@ -211,53 +199,6 @@ func apiFlags() []cli.Flag {
 		NoHTTP2Flag,
 		DebugHTTPFlag,
 		TraceHTTPFlag,
-	}
-}
-
-func cacheFlags() []cli.Flag {
-	return []cli.Flag{
-		cli.StringFlag{
-			Name:   "ids",
-			Value:  "",
-			Usage:  "Comma-separated list of cache IDs (if empty, processes all caches)",
-			EnvVar: "BUILDKITE_CACHE_IDS",
-		},
-		cli.StringFlag{
-			Name:   "registry",
-			Value:  "~",
-			Usage:  "The slug of the cache registry to use, defaults to the default registry (~)",
-			EnvVar: "BUILDKITE_CACHE_REGISTRY",
-		},
-		cli.StringFlag{
-			Name:   "bucket-url",
-			Value:  "",
-			Usage:  "The URL of the bucket (e.g., s3://bucket-name)",
-			EnvVar: "BUILDKITE_CACHE_BUCKET_URL",
-		},
-		cli.StringFlag{
-			Name:   "branch",
-			Value:  "",
-			Usage:  "Which branch should the cache be associated with",
-			EnvVar: "BUILDKITE_BRANCH",
-		},
-		cli.StringFlag{
-			Name:   "pipeline",
-			Value:  "",
-			Usage:  "The pipeline slug for this cache",
-			EnvVar: "BUILDKITE_PIPELINE_SLUG",
-		},
-		cli.StringFlag{
-			Name:   "organization",
-			Value:  "",
-			Usage:  "The organization slug for this cache",
-			EnvVar: "BUILDKITE_ORGANIZATION_SLUG",
-		},
-		cli.StringFlag{
-			Name:   "cache-config-file",
-			Value:  ".buildkite/cache.yml",
-			Usage:  "Path to the cache configuration YAML file",
-			EnvVar: "BUILDKITE_CACHE_CONFIG_FILE",
-		},
 	}
 }
 
--- a/internal/cache/cache.go
+++ b/internal/cache/cache.go
@@ -27,8 +27,8 @@ type Config struct {
 	Organization string
 	// CacheConfigFile is the path to the cache configuration YAML file
 	CacheConfigFile string
-	// Ids is a comma-separated list of cache IDs (if empty, processes all caches)
-	Ids string
+	// Ids is a list of cache IDs (if empty, processes all caches)
+	Ids []string
 	// APIEndpoint is the Agent API endpoint
 	APIEndpoint string
 	// APIToken is the access token used to authenticate
@@ -128,10 +128,7 @@ func setupCacheClient(ctx context.Context, l logger.Logger, cfg Config) (*zstash
 	}
 
 	// Determine which cache IDs to process
-	var cacheIDs []string
-	if cfg.Ids != "" {
-		cacheIDs = strings.Split(cfg.Ids, ",")
-
+	if len(cfg.Ids) > 0 {
 		// Validate that specified cache IDs exist
 		validIDs := make(map[string]bool)
 		for _, cache := range cacheClient.ListCaches() {
@@ -139,7 +136,7 @@ func setupCacheClient(ctx context.Context, l logger.Logger, cfg Config) (*zstash
 		}
 
 		var invalidIDs []string
-		for _, id := range cacheIDs {
+		for _, id := range cfg.Ids {
 			if !validIDs[id] {
 				invalidIDs = append(invalidIDs, id)
 			}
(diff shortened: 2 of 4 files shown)


DrJosh9000:
Tagging myself so I'm subscribed

clicommand/cache_restore.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/urfave/cli"
+)
+
+const cacheRestoreHelpDescription = `Usage:
+
+    buildkite-agent cache restore [options]
+
+Description:
+
+Restores files from the cache for the current build based on the cache configuration
```
zhming0: I think it should be "job"?

```suggestion
Restores files from the cache for the current job based on the cache configuration
```

clicommand/cache_restore.go
```diff
@@ -0,0 +1,98 @@
...
+Configuration File Format:
+
+The cache configuration file should be in YAML format:
+
+    - id: node
+      key: '{{ id }}-{{ agent.os }}-{{ agent.arch }}-{{ checksum "package-lock.json" }}'
+	  fallback_keys:
+		- '{{ id }}-{{ agent.os }}-{{ agent.arch }}-'
+      paths:
+        - node_modules
```
zhming0: It might be wise to we document the behavior when it comes to symlink in the folder.

clicommand/cache_restore.go
```diff
@@ -0,0 +1,98 @@
...
+
+The command will report one of three outcomes for each cache:
+  - Cache hit: Exact key match found and restored
+  - Fallback used: No exact match, but a fallback key was found and restored
+  - Cache miss: No matching cache found
+
+The command automatically uses the following environment variables when available:
+  - BUILDKITE_BRANCH (for branch scoping)
+  - BUILDKITE_PIPELINE_SLUG (for pipeline scoping)
+  - BUILDKITE_ORGANIZATION_SLUG (for organization scoping)`
```
zhming0, line 62: [question] If these aren't available, are these information by default extracted from job token?

clicommand/cache_restore.go
```diff
@@ -0,0 +1,98 @@
...
+Example:
+
+    $ buildkite-agent cache restore
+
+This will restore all caches defined in .buildkite/cache.yml. You can also restore
+specific caches by providing their IDs:
+
+    $ buildkite-agent cache restore --ids "node"
+
+The cache will be retrieved from the bucket specified by --bucket-url or your
```
zhming0, line 37: perhaps we should be clear that it's S3(-compliant) buckets.

internal/cache/cache.go
```diff
@@ -0,0 +1,199 @@
...
+		},
+	})
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to create cache client: %w", err)
+	}
+
+	// Determine which cache IDs to process
+	var cacheIDs []string
+	if cfg.Ids != "" {
+		cacheIDs = strings.Split(cfg.Ids, ",")
```
zhming0: Do we need to assert that the cache exist for a cache id?

internal/cache/cache.go
```diff
@@ -0,0 +1,199 @@
...
+	var caches []cache.Cache
+	if err := yaml.Unmarshal(data, &caches); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal cache config file: %w", err)
+	}
+
+	return caches, nil
+}
+
+// setupCacheClient creates a cache client and determines which cache IDs to process
+func setupCacheClient(ctx context.Context, l logger.Logger, cfg Config) (*zstash.Cache, []string, error) {
```
zhming0, line 97: [non-blocking] as we will eventually introduce more ways to specify cache configuration, I am not sure if we should use cache id as the first level concept here. Maybe we should turn it into a concrete `CacheConfig` object  and pass that around.

wolfeidau: Good point, will add validation of cache ids.

clicommand/cache_restore.go
wolfeidau: Fixed 👍

clicommand/cache_restore.go
wolfeidau: @zhming0 at the moment this is just run from the CLI so it depends on the environment variables, this will change once we add some configuration to the pipeline yaml as we can then trigger it within the agent lifecycle without needing to invoke the cli.

clicommand/cache_restore.go
wolfeidau: @zhming0 yeah this will need to be captured as it also overlaps with security, which makes it very complicated.

clicommand/cache_restore.go
wolfeidau: @zhming0 We can capture all that once we start documenting usage IMO, I have added support for overriding path options and endpoint so it should work with a variaty of services. https://github.com/buildkite/zstash/blob/main/store/s3.go#L27

internal/cache/cache.go
wolfeidau: Good point fixed.

internal/cache/cache.go
zhming0: I think my main point is that maybe this method should return a list of cache configurations instead of cache ids. Cache ID as a concept doesn't seem to serve much purpose beyond system boundary. They might be more of an inconvenience when we want to refactor this down the track. 

Like this method, maybe it can just return 
```suggestion
func setupCacheClient(ctx context.Context, l logger.Logger, cfg Config) (*zstash.Cache, []CacheConfiguration, error) {
```

wolfeidau: @zhming0 sorry commented on the wrong change here, cache ids are used to selectively restore caches in a step. The common example is a rails app wiht ruby and node deps, being able to only restore node or ruby deps or both in a step is useful. 

This model follows other ci providers in their structure. 

A good example of a bigger cache is something we experimented here https://github.com/bk-playground/kitesocial/blob/main/.buildkite/cache.hosted.yml#L1-L9 which has a docker cache.

zhming0: I acknowledge the value of cache ids in boundary/edge (in people's pipeline yaml as you mentioned), but I don't see a great value preserving this concept beyond that. 

What I mean is that maybe we can just take the cache IDs, parse them into into config configuration at the earliest code path, so we don't need to pass cache ids around internall?

clicommand/global.go
```diff
@@ -202,6 +214,53 @@ func apiFlags() []cli.Flag {
 	}
 }
 
+func cacheFlags() []cli.Flag {
+	return []cli.Flag{
+		cli.StringFlag{
```
DrJosh9000: A `cli.StringSliceFlag` might be a better fit.

clicommand/global.go
```diff
@@ -182,6 +182,18 @@ type APIConfig struct {
 	NoHTTP2          bool   `cli:"no-http2"`
 }
 
+// CacheConfig includes cache-related shared options for easy inclusion across
+// cache command config structs (via embedding).
+type CacheConfig struct {
+	Ids             string `cli:"ids"`
```
DrJosh9000: Here's what I think:

```suggestion
	IDs             []string `cli:"ids"`
```

clicommand/global.go
DrJosh9000, line 1: Rather than adding more to `global.go`, maybe the new flags should go in e.g. `cache_shared.go` ?

clicommand/cache_restore.go
```diff
@@ -0,0 +1,99 @@
...
+Restores files from the cache for the current job based on the cache configuration
+defined in your cache config file (defaults to .buildkite/cache.yml).
+
+The cache configuration file defines which files or directories should be restored
+and their associated cache keys. Caches are scoped by organization, pipeline, and
+branch. If an exact cache match is not found, the command will attempt to use
+fallback keys if defined in your cache configuration.
+
+Note: This feature is currently in development and subject to change. It is not
+yet available to all customers.
```
moskyb: might be good to put an `EXPERIMENTAL` callout in here, as that's usually the verbiage we use to denote something that might change

clicommand/cache_save.go
```diff
@@ -0,0 +1,97 @@
...
+	Usage:       "Saves files to the cache",
+	Description: cacheSaveHelpDescription,
+	Flags:       slices.Concat(globalFlags(), apiFlags(), cacheFlags()),
+	Action: func(c *cli.Context) error {
+		ctx := context.Background()
+		ctx, cfg, l, _, done := setupLoggerAndConfig[CacheSaveConfig](ctx, c)
+		defer done()
+
+		l.Info("Cache save command executed")
+
```
moskyb:
```suggestion
```

likewise here

clicommand/global.go
wolfeidau: done.

clicommand/global.go
wolfeidau: done this was moved to `clicommand/cache_shared.go`

clicommand/global.go
wolfeidau: done and moved to `clicommand/cache_shared.go`

clicommand/cache_restore.go
zhming0: My main question was on the security aspect. If in the backend we validate `BUILDKITE_BRANCH` against  agent job token then I think it's ✅ .

clicommand/cache_restore.go
zhming0: yep, not a blocker.

zhming0 (approved):
Great starting point. I have doubt about some details but none of them are blocking. So it's a LGTM from me.

wolfeidau:
@DrJosh9000 did you have any concerns with me merging this?

PR #3554, opened 2025-10-27
Upgrade to AWS Go SDK v2

DrJosh9000:
### Description

Upgrade to the supported v2 of the AWS Go SDK (from v1, which is unsupported).

### Context

https://linear.app/buildkite/issue/PB-861/update-to-aws-go-sdk-v2

Fixes #3533 
Closes #3559

### Changes

Find v2 equivalents of v1 functionality. This is mostly fine but not always possible, particularly `NewChainCredentials`. Since we should be using a more standard approach to authenticating to AWS anyway (see #3559), I've changed the way that part works to be similar to #3559.

### Testing
- [x] Tests have run locally (with `go test ./...`). Buildkite employees may check this if the pipeline has run automatically.
- [x] Code is formatted (with `go fmt ./...`)
- [ ] Actually try running it against AWS

### Disclosures / Credits

No AI here

PR #3559, opened 2025-10-27
Standardize artifact S3 credential loading

csweeney-plaid:
### Description
- Rather than using an entirely custom credential chain, simply take the [default credential chain ](https://docs.aws.amazon.com/sdkref/latest/guide/standardized-credentials.html)& prepend the buildkite-specific provider (`BUILDKITE_S3` overrides)
- This expands support for currently unsupported options - like loading credentials from `~/.aws/config` and also normalizes the credential loading process here vs the AWS norms.

### Changes
- `buildkite-agent artifact upload/download` (not a visible api change)
- ⚠️  this is a breaking change to the extent that it changes the order of providers and adds currently unsupported providers that might be silently passing through

### Testing
- [ X ] Tests have run locally (with `go test ./...`). Buildkite employees may check this if the pipeline has run automatically.
- [ X ] Code is formatted (with `go fmt ./...`)

### Disclosures / Credits
 - I used Claude to help put this together.

### DIFF
--- a/internal/artifact/s3.go
+++ b/internal/artifact/s3.go
@@ -72,21 +72,20 @@ func (p *credentialsProvider) IsExpired() bool {
 func awsS3Session(region string, l logger.Logger) (*session.Session, error) {
 	// Chicken and egg... but this is kinda how they do it in the sdk
 
-	// Determine the profile to use
 	profile := os.Getenv("BUILDKITE_S3_PROFILE")
 	if profile == "" {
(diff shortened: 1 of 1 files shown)


DrJosh9000:
Thanks for raising this one @csweeney-plaid! The version with #3554 in it is [out now](https://github.com/buildkite/agent/releases/tag/v3.110.0)

PR #3561, opened 2025-11-02
chore: inject service name so children inherit buildkite-agent

mcncl:
### Description

Currently, the buildkite-agent service doesn't seem to be set as the service name for child processes such as plugins (docker pull in the docker-compose plugin).

### Changes

This sets the `OTEL_SERVICE_NAME` value to the default (`buildkite-agent`) unless it's been overwritten via config flags.

### Testing

### Caveats

There is a chance we may want to allow plugins to set their own service context in the future? However, plugins such as `docker-compose` often run commands on behalf of the agent, so having the service as the agent itself might be sensible...


mcncl:
Interesting, @pzeballos; my local even after a pull didn't have that. I'll close this PR and let the folks know

PR #3565, opened 2025-11-04
PS-1373 Update error message for hashing failures

moskyb:
### Description

When the FIPS140 mode is set to only, the agent produces a confusing error message:
```
fatal: failed to upload artifacts: collecting artifacts: building artifact: reading contents of /path/to/some/artifact: crypto/sha1: use of SHA-1 is not allowed in FIPS 140-only mode
```

the `reading contents of` in there makes it seem like we're failing to read the file, but we're actually failing to hash it. This failure to hash is also pointed out by the rest of the error message, but it's useful to be explicit.

### Context

PS-1373

### Testing

### Disclosures / Credits

This twelve-character change was generated entirely by my human brain.


DrJosh9000:
Quick question: do we use the SHA-1 hash for anything anymore?

DrJosh9000 (approved):

PR #3569, opened 2025-11-07
PB-899 Add --scope flag to `buildkite-agent annotate` command

moskyb:
### Description

We're currently building a feature where agents (and the jobs running on them) will be able to decide the "scope" for an annotation — ie, whether it's an annotation on the build (the current status quo), or an annotation on a particular job (new and shiny!).

This PR adds a flag to the agent to let callers decide the scope for their annotation. They can use `--scope build` (or omit the flag entirely) to scope to a build, or use `--scope job` to display their annotation on a job.

### Context

[PB-899](https://linear.app/buildkite/issue/PB-899/create-a-scope-flag-to-the-buildkite-agent-agent-annotate-command#comment-443ef163)

### Testing

### Disclosures / Credits

It's all me, baby.

### DIFF
--- a/agent/api.go
+++ b/agent/api.go
@@ -13,7 +13,7 @@ type APIClient interface {
 	AcceptJob(context.Context, *api.Job) (*api.Job, *api.Response, error)
 	AcquireJob(context.Context, string, ...api.Header) (*api.Job, *api.Response, error)
 	Annotate(context.Context, string, *api.Annotation) (*api.Response, error)
-	AnnotationRemove(context.Context, string, string) (*api.Response, error)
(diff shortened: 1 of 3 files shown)


clicommand/annotate.go
```diff
@@ -109,6 +116,11 @@ var AnnotateCommand = cli.Command{
 		ctx, cfg, l, _, done := setupLoggerAndConfig[AnnotateConfig](ctx, c)
 		defer done()
 
+		validScopes := []string{"build", "job"}
```
dabarrell: We may add a step scope in the future - could be good to leave this to server side validation to make that easier?

moskyb: ezpz

PR #3570, opened 2025-11-10
IdleMonitor-related fixes

DrJosh9000:
### Description

Various fixes and cleanups related to IdleMonitor. Particularly, if an agent worker stops or crashes, it should be treated as dead for the purposes of the disconnect-after-idle-timeout.

### Context

https://linear.app/buildkite/issue/PS-1390/3-buildkite-agent-crashingstopped-workers-breaks-idlemonitor

If multiple workers are spawned and then any of them either crash or are gracefully stopped by the API (or by clicking the Stop Agent button in UI), then idleMonitor.Idle() will never [evaluate](https://github.com/buildkite/agent/blob/v3.111.0/agent/idle_monitor.go#L31) to true.

### Changes

- Add a new agent state in IdleMonitor, "dead", which is when an agent has exited. "Dead" is like "idle" but is terminal (an agent can't exit the "dead" state).
- Move idleness timeout logic from the ping loop to the IdleMonitor. `idleMonitor` now decides when the agent should exit due to idleness.
- Un-export `idleMonitor`, `newIdleMonitor`, and all `idleMonitor` methods.
- Move `int` -> `time.Duration` conversion earlier (when producing `AgentConfiguration`).

### Testing
- [x] Some manual testing

### Disclosures / Credits

I did not use AI tools at all

### DIFF
--- a/agent/agent_pool.go
+++ b/agent/agent_pool.go
@@ -13,17 +13,15 @@ import (
 	"github.com/buildkite/agent/v3/status"
 )
 
-// AgentPool manages multiple parallel AgentWorkers
+// AgentPool manages multiple parallel AgentWorkers.
 type AgentPool struct {
-	workers     []*AgentWorker
-	idleMonitor *IdleMonitor
+	workers []*AgentWorker
 }
 
-// NewAgentPool returns a new AgentPool
+// NewAgentPool returns a new AgentPool.
 func NewAgentPool(workers []*AgentWorker) *AgentPool {
 	return &AgentPool{
-		workers:     workers,
-		idleMonitor: NewIdleMonitor(len(workers)),
+		workers: workers,
 	}
 }
 
@@ -57,12 +55,15 @@ func (r *AgentPool) Start(ctx context.Context) error {
 	defer done()
 	setStat("🏃 Spawning workers...")
 
+	idleMon := newIdleMonitor(len(r.workers))
+
 	errCh := make(chan error)
 
 	// Spawn each worker "in parallel" (in its own goroutine)
 	for _, worker := range r.workers {
 		go func() {
-			errCh <- r.runWorker(ctx, worker)
+			defer idleMon.markDead(worker)
+			errCh <- r.runWorker(ctx, worker, idleMon)
 		}()
 	}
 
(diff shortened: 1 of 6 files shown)


agent/idle_monitor.go
```diff
@@ -1,44 +1,94 @@
...
-	return len(i.idle) == i.totalAgents
+// shouldExit reports *all* have been idle for at least minIdle. It returns
+// false if minIdle == 0. Otherwise, if shouldExit returns true, it will return
+// true on all subsequent calls.
+func (i *idleMonitor) shouldExit(minIdle time.Duration) bool {
+	// If disconnect-after-idle-timeout is not set, don't exit because agents
+	// are idle.
+	if minIdle == 0 {
+		return false
+	}
```
zhming0: When `minIdle` <= 0, should we raise an error instead or even panic directly? 

This module should be agnostic about `disconnect-after-idle-timeout`.

agent/agent_pool.go
```diff
@@ -57,12 +55,15 @@ func (r *AgentPool) Start(ctx context.Context) error {
...
 
+	idleMon := newIdleMonitor(len(r.workers))
+
 	errCh := make(chan error)
 
 	// Spawn each worker "in parallel" (in its own goroutine)
 	for _, worker := range r.workers {
 		go func() {
-			errCh <- r.runWorker(ctx, worker)
+			defer idleMon.markDead(worker)
```
zhming0, line 65: If we use `idleMon.markIdle` here, it will work as well? do we need a concept of "dead" in idle monitor?

zhming0 (approved):
Looks good to me. I have two non-blocking questions. 

In addition, for the sake of transparency, should we describe the bug in PR description? 

> If multiple workers are spawned and then any of them either crash or are gracefully stopped by the API (or by clicking the Stop Agent button in UI), then idleMonitor.Idle() will never [evaluate](https://github.com/buildkite/agent/blob/v3.111.0/agent/idle_monitor.go#L31) to true.

agent/idle_monitor.go
DrJosh9000: I don't really think we need this to be agnostic, but the check was redundant, so I removed it.

agent/agent_pool.go
DrJosh9000: There is a distinction:

Suppose we have two workers and an idle timeout of 600s. Suppose worker 1 finished a job 300 seconds ago and is idle, and worker 2 has just died. No more jobs come in.

If we mark worker 2 as _idle_, then the agent will start the timeout from now and exit 600s from now (an extra 300s after all _alive_ agents became idle). If we mark worker 2 as _dead_, then the agent will exit in 300s, which I think is a bit more sensible.

zhming0: That makes sense.

PR #3571, opened 2025-11-10
Skip ticker after running job

DrJosh9000:
### Description

In the ping loop, skip waiting on the `pingTicker` when a job has just run. When a job has just run, the only wait will be the jitter.

### Context

Prior to rearranging and introducing jitter, the first thing the ping loop did was always to ping. `continue` (implied or explicit) thus led to an immediate ping. Per the comment near the end of the loop (#1567) we want to skip the ticker and ping immediately, but in exchange, wait the full ping interval the next time.

After rearranging and introducing jitter, the first thing the ping loop does is wait for the ticker, then the jitter. But the `pingTicker.Reset` call after running a job was written with the expectation that a ping would happen immediately. In the new arrangement, it instead causes the loop to wait the full ping interval first.

### Changes

- Rename the `first` channel to `skipTicker`
- Send a value on `skipTicker` when a job has just run
- Delete the redundant DisconnectAfterJob conditional

### Testing

### Disclosures / Credits

Good old fashioned wetware

### DIFF
--- a/agent/agent_worker.go
+++ b/agent/agent_worker.go
@@ -332,8 +332,10 @@ func (a *AgentWorker) runPingLoop(ctx context.Context, idleMon *idleMonitor) err
 		close(testTriggerCh)
 	}
 
-	first := make(chan struct{}, 1)
-	first <- struct{}{}
+	// On the first iteration, skip waiting for the pingTicker.
+	// This doesn't skip the jitter, though.
+	skipTicker := make(chan struct{}, 1)
+	skipTicker <- struct{}{}
 
 	a.logger.Info("Waiting for instructions...")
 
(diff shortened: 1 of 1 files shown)


agent/agent_worker.go
```diff
@@ -468,22 +470,16 @@ func (a *AgentWorker) runPingLoop(ctx context.Context, idleMon *idleMonitor) err
 		}
 
 		ranJob = true
-		if a.agentConfiguration.DisconnectAfterJob {
-			// Unless paused, this agent disconnects after the next ping.
-			// Do the ping immediately so we reduce the chances our agent is assigned a job
-			pingTicker.Reset(pingInterval)
-			continue
-		}
```
catkins, line 476: Image

hmm looks like this is at least load bearing to some tests currently, so we'll need to rework that little bit

DrJosh9000: I think I figured it out

catkins (approved):

PR #3579, opened 2025-11-13
Fix idle tracking for agents that never received jobs

scadu:
### Description

In #3570 idle logic has changed and `idleMon.markIdle` would get only triggered via `AcceptAndRunJob()`.
Agents that never received a job wouldn't be marked as idle.

### Context

Issue discovered in https://github.com/buildkite/elastic-ci-stack-for-aws/pull/1652 where the test stacks couldn't be destroyed. 
I found out that a single instance would prevent ASG from being destroyed.
I tested agent 3.112.0 on a separate stack and could confirm this behavior.
Agents that received jobs were terminated properly, while agents that never received job would never get marked as idle, handing indefinitely (or until job got assigned to it).

### Testing

I tested build 11168 of the agent on Elastic CI Stack for AWS instance. 
It now disconnects immediately when idle, following my agent's configuration.

### Disclosures / Credits

### DIFF
--- a/internal/mime/mime.go
+++ b/internal/mime/mime.go
@@ -779,6 +779,8 @@ var types = map[string]string{
 	".spq":         "application/scvp-vp-request",
 	".spx":         "audio/ogg",
 	".sql":         "application/x-sql",
+	".sqlite":      "application/vnd.sqlite3",
+	".sqlite3":     "application/vnd.sqlite3",
 	".src":         "application/x-wais-source",
 	".srt":         "application/x-subrip",
 	".sru":         "application/sru+xml",


DrJosh9000:
Just checking, this is deliberately a behaviour change, not just a fix? As far as I can tell, disconnect-after-idle-timeout has until now not timed out agents that haven't yet run a job?

DrJosh9000:
Actually, you're right, it seems that it used to set the `lastActionTime` at the start of the method, meaning that the agent effectively implicitly starts idle.

https://github.com/buildkite/agent/blob/1e8fa75d6293b2d8e90c071bdf9d3d6df1b5c7f0/agent/agent_worker.go#L271C2-L271C16

PR #3580, opened 2025-11-13
Fix the pipeline upload --reject-secrets flag not rejecting secrets

moskyb:
### Description

Long, long ago (https://github.com/buildkite/agent/pull/1523, https://github.com/buildkite/agent/pull/1589), we made it so that customers could opt into rejecting pipeline uploads that contained secret values -- eg, if you accidentally committed a pipeline like:
```yaml
steps:
  - command: "psql -c 'DROP TABLES *;'"
    env:
      DATABASE_PASSWORD: hunter2
```
the agent would recognise that there's a cleartext secret in there (`DATABASE_PASSWORD`) and refuse to upload it to buildkite.

However, much later on we refactored the redactor (https://github.com/buildkite/agent/pull/2277), and extracted the secrets detection into a function. That function would return an error if it thought the pipeline shouldn't be uploaded.

... and then we didn't check the error return of the function, meaning that secrets that probably shouldn't get uploaded were uploaded anyway.

This PR adds an error check to that function call, reinstating the behaviour that if a user adds `--reject-secrets` to their pipeline upload call, it'll refuse to upload sensitive values.

### Testing

### Disclosures / Credits

I didn't use AI.

### DIFF
--- a/clicommand/pipeline_upload.go
+++ b/clicommand/pipeline_upload.go
@@ -319,7 +319,11 @@ var PipelineUploadCommand = cli.Command{
 				if len(cfg.RedactedVars) > 0 {
 					// Secret detection uses the original environment, since
 					// Interpolate merges the pipeline's env block into `environ`.
-					searchForSecrets(l, &cfg, environ, result, input.name)
+					err := searchForSecrets(l, &cfg, environ, result, input.name)
+					if err != nil {
+						l.Error("%v", err)
+						return NewSilentExitError(1)
+					}
 				}
 
 				var (
@@ -509,6 +513,7 @@ func searchForSecrets(
 	if err != nil {
 		l.Warn("Couldn't match environment variable names against redacted-vars: %v", err)
 	}
+
 	for _, name := range short {
 		shortValues[name] = struct{}{}
 	}


internal/mime/mime.go
```diff
@@ -779,6 +779,8 @@ var types = map[string]string{
 	".spq":         "application/scvp-vp-request",
 	".spx":         "audio/ogg",
 	".sql":         "application/x-sql",
+	".sqlite":      "application/vnd.sqlite3",
+	".sqlite3":     "application/vnd.sqlite3",
```
wolfeidau, line 783: was this another change?

wolfeidau (commented):
just one query

internal/mime/mime.go
moskyb: yeah, although it also got merged in as part of #3579, so it should be gone when i rebase

moskyb: (which i have now done)

AliSoftware:
Ah, ended up finding the issue: that was because I since I set `BUILDKITE_AGENT_PIPELINE_UPLOAD_REJECT_SECRETS=1` to enable this feature on our agent… it considered that the value `1` (value of that env var)… was in fact a secret value that needed to be redacted 🤦 

I changed it to `export BUILDKITE_AGENT_PIPELINE_UPLOAD_REJECT_SECRETS=true` instead, and the extraneous redaction of all the `1`s in my logs have stopped 🎉 

That means though that now every mention of the word `true` in my logs are redacted… which is still not great… but at least it's not as frequent as having the character `1` redacted in all timestamps and lines 😅 

---

PS: I'm still surprised to see that the log redactor would consider the env var name `BUILDKITE_AGENT_PIPELINE_UPLOAD_REJECT_SECRETS` one that matches the `BUILDKITE_REDACTED_VARS` patterns (`*_PASSWORD,*_SECRET,*_TOKEN,*_PRIVATE_KEY,*_ACCESS_KEY,*_SECRET_KEY,*_CONNECTION_STRING`), given the env var name ends in `_SECRETS` not `_SECRET`. Also, I thought the log redactor was not supposed to redact strings less than 6 bytes… 🤔  I am filing a bug separately about that.

PR #3586, opened 2025-11-18
Delete pool & update Bazel

DrJosh9000:
### Description

Delete the pool package, and update everything Bazel.

### Context

Sporadic cleanup.

I thought about moving more libraries into `internal`. But the first one I found had only one use. I figured, why not delete it instead?

### Changes

- Delete pool
- Rewrite the concurrent artifact downloader's use of pool with channels, goroutines, and `sync.WaitGroup`
- Update Bazel module dependencies and Go version
- `bazel mod tidy`
- `bazel run :gazelle`

### Testing

### Disclosures / Credits

I did not use AI tools at all


internal/artifact/downloader.go
```diff
@@ -83,43 +83,100 @@ func (a *Downloader) Download(ctx context.Context) error {
...
+		}
+		errorsOutCh <- errors
+	}()
+
+	// A bunch of worker goroutines. Start the smaller of:
+	// - GOMAXPROCS (often equal to NumCPU)
+	// - the number of artifacts to download.
+	var wg sync.WaitGroup
+	artifactsCh := make(chan *api.Artifact)
+	for range min(runtime.GOMAXPROCS(0), len(artifacts)) {
```
zhming0: Artifact download seems to be an IO bound action, I am not sure if limiting the concurrency by CPU core count is right? (the old concurrency limit is 10 x CPU cores)

Merging this as it is will hinder single core agent's artifact download performance pretty heavily. 

Maybe we can use a different model: each connection can typically use X gbps bandwidth (in the case of s3 is 1 gbps typically), to saturate full 25 gbps, we allow Y max goroutine.

DrJosh9000: For now I've reverted to the old concurrency limit (10*CPU, but keeping min with # artifacts), since that discussion is not really the point of this PR (but could still be worth having).

zhming0 (approved):
There is suddenly a lot more bazel change, which I don't quite know how to review due to my lack of direct bazel file experience. Are they all output of `bazel mod tidy`?

Anyway I trust our new e2e tests  😁

PR #3589, opened 2025-11-20
Run gofumpt as part of CI

moskyb:
### Description

What if the formatting was a little bit nicer?

### Changes

- Replace `gofmt` with `gofumpt` in CI
- Run `gofumpt -extra -w .` across the entire repo

### Testing

### Disclosures / Credits

I guess a computer did most of the work here, but it was the old-fashioned kind.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,7 +23,6 @@ packaging/docker/*/hooks/
 .buildkite-agent.cfg
 
 .idea
-.vscode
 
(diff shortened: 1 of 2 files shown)


agent/api.go
DrJosh9000, line 1: What if we deleted this interface?

moskyb: i've thought about it many times. it's basically pointless.

PR #3599, opened 2025-11-27
Add end-to-end testing infra

DrJosh9000:
### Description

Adds a new package with some test helper code, and new pipeline yml.

So far it "end-to-end tests" the same agent version that runs the test, and does not inspect the job logs. These will be the next steps.

### Context

PB-698

### Changes

Similar to agent-stack-k8s integration test setup.

### Testing

### Disclosures / Credits

I did not use AI tools at all

### DIFF
--- a/.buildkite/docker-compose.yml
+++ b/.buildkite/docker-compose.yml
@@ -1,5 +1,3 @@
-version: '3.5'
-
 services:
   lint:
     build:
@@ -31,6 +29,23 @@ services:
       - GOCACHE=/gocache
       - GOMODCACHE=/gomodcache
       - BUILDKITE_TEST_ENGINE_API_ACCESS_TOKEN
+
+  e2e:
+    build:
+      context: .
+      dockerfile: Dockerfile-e2e
+    volumes:
+      - ../:/work:cached
+    working_dir: /work
+    environment:
+      - BUILDKITE_BUILD_NUMBER
+      - BUILDKITE_JOB_ID
+      - BUILDKITE_BUILD_CREATOR_EMAIL
+      - BUILDKITE_BUILD_CREATOR
+      - CI_E2E_TESTS_BUILDKITE_API_TOKEN
+      - CI_E2E_TESTS_AGENT_TOKEN
+      - CI_E2E_TESTS_TARGET_ORG
+      - CI_E2E_TESTS_TARGET_CLUSTER
 
   ruby:
     image: ruby:2.7.6
--- a/.buildkite/pipeline.e2e.yml
+++ b/.buildkite/pipeline.e2e.yml
@@ -1,0 +1,19 @@
+agents:
+  queue: agent-runners-linux-amd64
+
+env:
+  CI_E2E_TESTS_TARGET_ORG: agent-e2e-testing
+  CI_E2E_TESTS_TARGET_CLUSTER: e1773a06-b43c-4b77-84e5-5e088a737f6a
+
+steps:
+  - label: ":go::sleuth_or_spy: Run agent end-to-end tests"
+    command: .buildkite/steps/e2e-tests.sh
+    secrets:
+      - CI_E2E_TESTS_BUILDKITE_API_TOKEN
+      - CI_E2E_TESTS_AGENT_TOKEN
+    plugins:
+      - docker-compose#v4.14.0:
+          config: .buildkite/docker-compose.yml
+          cli-version: 2
+          mount-buildkite-agent: true
+          run: e2e
--- a/.buildkite/Dockerfile-e2e
+++ b/.buildkite/Dockerfile-e2e
@@ -1,0 +1,1 @@
+FROM public.ecr.aws/docker/library/golang:1.24.10@sha256:7b13449f08287fdb53114d65bdf20eb3965e4e54997903b5cb9477df0ea37c12
--- a/.buildkite/steps/e2e-tests.sh
+++ b/.buildkite/steps/e2e-tests.sh
@@ -1,0 +1,10 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# TODO: download the agent to test as an artifact
+#buildkite-agent artifact download buildkite-agent . --build "${BUILDKITE_TRIGGERED_FROM_BUILD_ID}"
+#export CI_E2E_TESTS_AGENT_PATH="${PWD}/buildkite-agent"
+
+# For now, e2e test the agent that's currently running
+export CI_E2E_TESTS_AGENT_PATH="$(which buildkite-agent)"
+go test -v -tags e2e ./internal/e2e/... | sed -e 's/^/>  /'


internal/e2e/basic_test.go
```diff
@@ -0,0 +1,37 @@
...
+	}
+
+	// TODO: add ability to inspect job logs
+
+	if err := agent.Process.Signal(syscall.SIGTERM); err != nil {
+		t.Errorf("agent.Process.Signal(%d) = %v", syscall.SIGTERM, err)
+	}
+	if err := agent.Wait(); err != nil {
+		t.Errorf("agent.Wait() = %v", err)
+	}
```
moskyb: this should get handled by the test completing, right? when the test finishes, its context gets cancelled, killing the process?

DrJosh9000: The agent is _killed_, so on the BK side it becomes lost. The queue can't be deleted because an agent is "still running".

Maybe this should go in a cleanup function registered in `startAgent` ? Thoughts?

DrJosh9000: Ugh, it's complicated by the ordering of cleanups and context cancellation. I've changed it to a deferred function that sends SIGQUIT.

moskyb: > Maybe this should go in a cleanup function registered in startAgent ? Thoughts?

yeah, i think that's a good call — there's a bit of nuance there that could be easy to miss, but much harder if there's a default cleanup function.

so i'm sure i have it right, `SIGTERM` is the one we don't catch, right?

.buildkite/docker-compose.yml
```diff
@@ -32,6 +30,23 @@ services:
       - GOMODCACHE=/gomodcache
       - BUILDKITE_TEST_ENGINE_API_ACCESS_TOKEN
 
+  e2e:
+    build:
+      context: .
+      dockerfile: Dockerfile-e2e
```
zhming0, line 36: Do we need build when the dockerfile only contain a single line of `FROM ...`?

.buildkite/pipeline.e2e.yml
```diff
@@ -0,0 +1,19 @@
+agents:
+  queue: agent-runners-linux-amd64
+
+env:
+  CI_E2E_TESTS_TARGET_ORG: agent-e2e-testing
+  CI_E2E_TESTS_TARGET_CLUSTER: e1773a06-b43c-4b77-84e5-5e088a737f6a
```
zhming0, line 6: Would it be good to log a task in linear to remove these two?

internal/e2e/testcase.go
```diff
@@ -0,0 +1,314 @@
...
+		// The arg is called "id" but it needs the build number...
+		state, _, err := tc.bkClient.Builds.Get(ctx, targetOrg, tc.pipeline.Slug, strconv.Itoa(build.Number), &buildkite.BuildGetOptions{})
+		if err != nil {
+			return "", err
+		}
+		switch state.State {
+		case "passed", "failed", "canceled", "canceling":
+			return state.State, nil
+
+		case "scheduled", "running":
```
zhming0, line 183: Without looking into how bk client work, this is a bit confusing? Why is there a "scheduled" state on build? I thought it's a job thing 🤔 .

internal/e2e/testcase.go
```diff
@@ -0,0 +1,314 @@
...
+	}
+	// The agent should be cancelled automatically by t.Context.
+	// But cancellation SIGKILLs the agent, so it doesn't disconnect cleanly.
+	// It is eventually marked lost on the backend, but not for a few minutes,
+	// which is long enough to block queue cleanup.
+	// This cleanup function SIGQUITs it, which forces an ungraceful
+	// but clean exit (jobs are cancelled but it has time to disconnect).
+	// This cleanup function has to happen *within* the test (use defer!),
+	// because the test context will otherwise kill the agent!
+	return cmd, func() {
```
zhming0: 👍🏿

zhming0 (approved):
No blocker, looks good to me 🚀

internal/e2e/basic_test.go
DrJosh9000: We catch `SIGINT`, `SIGTERM`, `SIGQUIT`, and `SIGHUP`. We're not allowed to catch `SIGKILL`. The difference between `SIGQUIT` and `SIGTERM` is that we proceed directly to cancelling jobs with `SIGQUIT`, but that only happens for the second `SIGTERM`.

https://github.com/buildkite/agent/blob/c1039ae975a77006442352a1731a63e3f8df97d5/clicommand/agent_start.go#L1385-L1456

internal/e2e/testcase.go
```diff
@@ -0,0 +1,314 @@
...
+		"--name", tc.fullName,
+		"--queue", tc.queue.Key,
+		"--build-path", buildPath,
+		"--hooks-path", hooksPath,
+		"--sockets-path", socketsPath,
+		"--plugins-path", pluginsPath,
+	}, extraArgs...)
+	tc.Logf("Starting agent with args: %q", args)
+
+	cmd := exec.CommandContext(tc.Context(), agentPath, args...)
```
zhming0, line 285: [totally not blocking] I wonder if we could intercept tc.Context()'s cancelation and change that to SIGTERM to the agent command? It would save the `defer` on the caller side.

.buildkite/docker-compose.yml
DrJosh9000: The Dockerfile mainly exists so that dependabot has something it can easily update. It might work with docker-compose.yml now, I haven't checked lately.

.buildkite/pipeline.e2e.yml
DrJosh9000: Remove and replace with what? Hard-code in the tests?

.buildkite/docker-compose.yml
zhming0: TIL!

internal/e2e/testcase.go
DrJosh9000: I copied this part with modifications from agent-stack-k8s. It seems there is a scheduled build state: https://github.com/buildkite/agent-stack-k8s/blob/eab654d3dda6574d1aa3fe9ba67c1dcb17987e8d/internal/integration/api/generated.go#L358-L359

.buildkite/pipeline.e2e.yml
zhming0: I am thinking replacing with automatically inspecting agent cluster token and find the org and cluster, like we did in the k8s stack

internal/e2e/testcase.go
DrJosh9000: I have a similar idea that might work... please stand by...

.buildkite/pipeline.e2e.yml
DrJosh9000: Oh ok. Sure.

internal/e2e/testcase.go
DrJosh9000: Now it replaces the command's `Cancel` func, and uses `WaitDelay` to ensure it is killed later. No special cleanup or defer needed!

Issue #3588, opened 2025-11-19
Buildkite logs redactor patterns are evaluated greedily, and short values still redacted


AliSoftware:
# TL;DR

The log redactor redacts vars named `FOO_SECRETS` (plural), as well as short values (< 6 bytes), while iiuc it's not supposed to do either.

**Buildkite Agent version**: `3.113.0` — though it seems the happen in earlier versions too (e.g. `3.107.0`)

# Expected Behavior

I'd expect that the log redactor would follow the same behavior as `pipeline upload --reject-secrets` and only redact values of env vars that match the `BUILDKITE_REDACTED_VARS` patterns _exactly_, anchoring them when evaluating them against the env var names.

That means that `BUILDKITE_REDACTED_VARS=*_SECRET` should redact the value of `echo $FAKE_SECRET` but not the value or `echo $FAKE_SECRETS`.

I'd expect that small values like `1` and `true` would not be redacted.

# Actual Behavior

The log redactor considers env vars named `$FOO_SECRETS` as being secret, while the `*_SECRET` pattern (from [the default value of `--redacted-vars` / `$BUILDKITE_REDACTED_VARS`](https://buildkite.com/docs/agent/v3/cli-pipeline#redacted-vars)) should match `$FOO_SECRET` but not `$FOO_SECRETS`.

💭  It seems to me that when it applies the patterns from `BUILDKITE_REDACTED_VARS` on the env var names—to know which one to consider needing redaction—it doesn't **anchor** the pattern on the env var name being tested?

As a result, even with `BUILDKITE_REDACTED_VARS`'s default value of `*_PASSWORD,*_SECRET,*_TOKEN,*_PRIVATE_KEY,*_ACCESS_KEY,*_SECRET_KEY,*_CONNECTION_STRING`, values of env vars with names like `MY_SECRETS` or `DISABLE_PASSWORD_PROMPT` will be matched and redacted in the logs.

Also, env vars considered secrets but having a value like `1` or `true` are **also** redacted in the logs, despite those values being shorter than 6 bytes.

## How I found out

See https://github.com/buildkite/agent/pull/3580#issuecomment-3554186310

I recently updated to `buildkite-agent` version `3.113.0` to benefit from [the recent fix about `buildkite-agent pipeline upload --reject-secrets`](https://github.com/buildkite/agent/pull/3580).

[...]

PR #3605, opened 2025-12-03
redactor add: Add optional filtering like env vars

DrJosh9000:
### Description

Adds an `--apply-vars-filter` flag to the `redactor add` subcommand, that enables the same filtering as normally happens for filtering env vars for secrets.

### Context

Relevant to #3588 

### Changes

- Adds `--apply-vars-filter` as described above, and the common `--redact-vars` flag.
- Updates the tests

Two small cleanups:

- JSON decoding should be able to work with a pointer to a map - the pointer to a pointer to a map should be unnecessary!
- Reordered the unknown format flag so it reads `unknown format "foo"` rather than `foo: unknown format`

### Testing

### Disclosures / Credits

I did it my way

### DIFF
--- a/clicommand/redactor_add.go
+++ b/clicommand/redactor_add.go
@@ -11,6 +11,8 @@ import (
 	"slices"
 	"strings"
 
+	"github.com/buildkite/agent/v3/env"
+	"github.com/buildkite/agent/v3/internal/redact"
 	"github.com/buildkite/agent/v3/jobapi"
 	"github.com/buildkite/agent/v3/logger"
 	"github.com/urfave/cli"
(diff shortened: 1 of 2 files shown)


clicommand/redactor_add.go
```diff
@@ -86,6 +90,12 @@ JSON does not allow duplicate keys. If you repeat the same key ("key"), the JSON
 			EnvVar: "BUILDKITE_AGENT_REDACT_ADD_FORMAT",
 			Value:  FormatStringNone,
 		},
+		cli.BoolFlag{
+			Name:   "apply-vars-filter",
+			Usage:  fmt.Sprintf("When the input is in 'json' format, filters the secrets to redact using the same rules used to detect secrets from environment variables: secrets must be at least %d characters long, and names must match the patterns defined by --redacted-vars or $BUILDKITE_REDACTED_VARS.", redact.LengthMin),
```
zhming0: Coming from an angle of not famlilar with agent redactor mechanism, I don't think I understand this feature without the code. Maybe we can use some touch up on the doc here, something alone the line of: 

"By default, all values in the JSON will be redacted. When apply-vars-filter is true, only ... gets redacted". 

But it's not a blocker.

zhming0 (approved):
Looks good, looks useful 👍🏿

PR #3606, opened 2025-12-03
Fetch and compare job logs

DrJosh9000:
### Description

Add ability to fetch job logs to the E2E test helpers, and show that it works in the basic E2E test.

### Context

https://linear.app/buildkite/issue/PB-990/add-mechanism-to-read-job-logs

### Changes

Per Description.

### Testing
- [x] It works

### Disclosures / Credits

Some bits copied from https://github.com/buildkite/agent-stack-k8s.

### DIFF
--- a/internal/e2e/basic_test.go
+++ b/internal/e2e/basic_test.go
@@ -2,7 +2,10 @@
 
 package e2e
 
-import "testing"
+import (
+	"strings"
+	"testing"
+)
 
 func TestBasicE2E(t *testing.T) {
 	ctx := t.Context()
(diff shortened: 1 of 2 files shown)


internal/e2e/basic_test.go
```diff
@@ -24,5 +27,11 @@ steps:
 		t.Errorf("Build state = %q, want %q", got, want)
 	}
 
-	// TODO: add ability to inspect job logs
+	logs, err := tc.fetchLogs(ctx, build)
+	if err != nil {
+		t.Errorf("tc.fetchLogs() error = %v", err)
+	}
```
zhming0: this error handling and the waitForBuild error handling can all happen in the callee side I think, I don't think they will ever be useful on caller side? 

Since we have a `t *testing.T` available, we can effectively "throw" error from callee side.

DrJosh9000: That's probably true. I'm not sure we'll need a case where they can fail, so I'll move that.

PR #3607, opened 2025-12-04
PS-1470 Fix issue where artifacts uploaded to custom s3 buckets would have a double slash prefix

moskyb:
### Description

Fix issue where artifacts uploaded to custom s3 buckets would have a double slash prefix

There's an edge case in URL construction for artifacts uploaded to custom S3 buckets where the key for the object would be generated as `/object-key` (note preceding slash), and then directly appended to a url like `https://my-definitely-realy-s3-bucket.s3.amazonaws.com/`, leading to a final object url of `https://my-definitely-real-s3-bucket.s3.amazonaws.com//object-key` (note the double slash!). See [the implementation](https://github.com/buildkite/agent/blob/2a95448832f3fa4c2eb08dabb69db06ff01782fb/internal/artifact/s3_uploader.go#L147) for more details — `u.BucketPath` is empty by default in custom artifact bucket cases.

When we were on the AWS Go SDK v1, this was fine, as the SDK [cleaned urls by default](https://github.com/aws/aws-sdk-go/issues/2559), leading to the object's key being normalized from `/object-key` to `object-key`. However, in an [undocumented breaking change](https://github.com/aws/aws-sdk-go-v2/issues/3095) (🫠), the SDK v2 doesn't do this cleaning, so the object would be uploaded with a leading slash on its key.

The artifact downloader relied on the URL cleaning functionality of the go SDK, so when attempting to download the artifact, it would generate a key of `object-key`, without the leading slash, leading to consistent 404s when trying to download artifacts from custom buckets.

This PR updates the artifact upload URL generation logic to do a path-join of the key rather than a crude append, essentially emulating the logic that was present in v1 of the AWS Go SDK.

### Context

[PS-1470](https://linear.app/buildkite/issue/PS-1470)

### Testing
- [x] Code is formatted (with `go tool gofumpt -extra -w .`)\

### Disclosures / Credits

I used Amp to investigate the issue initially, then wrote the fourteen-character fix all by myself.

### DIFF
--- a/.buildkite/steps/e2e-tests.sh
+++ b/.buildkite/steps/e2e-tests.sh
@@ -1,9 +1,10 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-if [[ -z "${BUILDKITE_TRIGGERED_FROM_BUILD_ID}" ]] ; then
+if [[ -z "${BUILDKITE_TRIGGERED_FROM_BUILD_ID:-}" ]] ; then
 	# For now, e2e test the agent that's currently running
-	export CI_E2E_TESTS_AGENT_PATH="$(which buildkite-agent)"
+	CI_E2E_TESTS_AGENT_PATH="$(which buildkite-agent)"
+	export CI_E2E_TESTS_AGENT_PATH
 else
 	# Download the artifact from the triggering build
 	ARTIFACT="pkg/buildkite-agent-$(go env GOOS)-$(go env GOARCH)"
--- a/internal/e2e/artifact_test.go
+++ b/internal/e2e/artifact_test.go
@@ -1,0 +1,21 @@
+//go:build e2e
+
+package e2e
+
+import (
+	"testing"
+)
+
+// Test that an agent can upload and download an artifact across different steps in the same build
+func TestArtifactUploadDownload(t *testing.T) {
+	ctx := t.Context()
+
+	tc := newTestCase(t, "artifact_upload_download.yaml")
+
+	tc.startAgent()
+	build := tc.triggerBuild()
+	state := tc.waitForBuild(ctx, build)
+	if got, want := state, "passed"; got != want {
+		t.Errorf("Build state = %q, want %q", got, want)
+	}
+}
--- a/internal/e2e/fixtures/artifact_upload_download.yaml
+++ b/internal/e2e/fixtures/artifact_upload_download.yaml
@@ -1,0 +1,11 @@
+agents:
+  queue: {{ .queue }}
+steps:
+  - key: upload
+    commands:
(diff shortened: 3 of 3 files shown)


moskyb:
oooh, good call!

internal/e2e/testcase.go
```diff
@@ -62,6 +63,8 @@ type testCase struct {
 	pipeline       *buildkite.Pipeline
 }
 
+var leadingTabsRe = regexp.MustCompile(`(?m)^(\t+)`)
```
DrJosh9000:
```suggestion
var leadingTabsRE = regexp.MustCompile(`(?m)^(\t+)`)
```

internal/e2e/artifact_test.go
```diff
@@ -0,0 +1,35 @@
...
+	ctx := t.Context()
+
+	tc := newTestCase(t, `
+agents:
+  queue: {{.queue}}
+steps:
+  - commands:
+    - echo "hello world" > artifact.txt
+    - buildkite-agent artifact upload artifact.txt
+  - commands:
```
DrJosh9000: Should this have a `depends_on` or `wait` to make the ordering clear?

internal/e2e/artifact_test.go
```diff
@@ -0,0 +1,35 @@
...
+	build := tc.triggerBuild()
+	state := tc.waitForBuild(ctx, build)
+	if got, want := state, "passed"; got != want {
+		t.Errorf("Build state = %q, want %q", got, want)
+	}
+
+	logs := tc.fetchLogs(ctx, build)
+	if !strings.Contains(logs, "\nsuccess") {
+		t.Errorf("tc.fetchLogs(ctx, build %q) logs as follows, did not contain 'success'\n%s", build.ID, logs)
+	}
```
DrJosh9000: Relying on the job to succeed or fail might be enough. Having to pull the job logs might make the test run longer for not much benefit.

internal/e2e/artifact_test.go
DrJosh9000, line 1: I assume you're working on a version that uses a custom S3 bucket?

.buildkite/steps/e2e-tests.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -euo pipefail
+set -eo pipefail
 
 if [[ -z "${BUILDKITE_TRIGGERED_FROM_BUILD_ID}" ]] ; then
```
DrJosh9000: What about:


```suggestion
set -euo pipefail

if [[ -z "${BUILDKITE_TRIGGERED_FROM_BUILD_ID:-}" ]] ; then
```

PR #3608, opened 2025-12-04
Read target org and cluster from agent token

DrJosh9000:
### Description

Instead of hard- or firmly-coding the org and cluster, interrogate Buildkite with the Agent token.

### Context

https://linear.app/buildkite/issue/PB-989/obtain-the-target-org-and-cluster-from-the-api-token

### Changes

Per description.

### Testing

### Disclosures / Credits

Copied and then rewrote parts of the API client from agent-stack-k8s.

### DIFF
--- a/.buildkite/pipeline.e2e.yml
+++ b/.buildkite/pipeline.e2e.yml
@@ -1,10 +1,6 @@
 agents:
   queue: agent-runners-linux-amd64
 
-env:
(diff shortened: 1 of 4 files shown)


zhming0 (approved):
Looks good. It might be a good moment to add some instruction to our README friends?

DrJosh9000:
What updates to the README did you have in mind?

PR #3610, opened 2025-12-04
Move pipeline YAML into fixtures dir

DrJosh9000:
### Description

YAML will be easier to edit in separate files than within Go string literals.

### Context

https://linear.app/buildkite/issue/PB-1011/store-test-pipelines-in-individual-files

### Changes

Similar to agent-stack-k8s

### Testing
- [x] It works

### Disclosures / Credits

I did not use AI tools at all

### DIFF
--- a/internal/e2e/basic_test.go
+++ b/internal/e2e/basic_test.go
@@ -9,12 +9,7 @@ import (
 
 func TestBasicE2E(t *testing.T) {
 	ctx := t.Context()
-	tc := newTestCase(t, `
-agents:
-  queue: {{.queue}}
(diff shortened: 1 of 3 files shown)


internal/e2e/basic_test.go
zhming0, line 1: I have conflicting feeling about this. I wonder if it's wise to keep both options? 

Maybe we can help a bit by injecting the agent/queue bit in the yaml?

zhming0: but my stance isn't very strong on this topic so I will approve

PR #3611, opened 2025-12-04
Log agent logs separately

DrJosh9000:
### Description

Instead of passing agent output directly to `os.Stderr`, capture in a buffer and log once during cleanup. This keeps the agent logs in one chunk separate from the test logs.

### Context

https://buildkite-corp.slack.com/archives/C061U5PG9K5/p1764821554921599?thread_ts=1764817320.578409&cid=C061U5PG9K5

### Changes

Per the description.

### Testing

### Disclosures / Credits

I did not use AI tools at all


moskyb (approved):
awesome!

PR #3614, opened 2025-12-05
Add branch_name to aws assume role session tags for beta release pipeline

moskyb:
### Description

Recently, we updated the trust policy on the role that the beta release pipeline assumes to do AWS stuff such that it ensures that the branch name is either `main` or ends with `-stable`, but we forgot to add the branch name aws session tag to the OIDC token we use to assume that role.

This PR adds the `branch_name` session tag to the assume role calls for the beta release pipeline.

### Context

[PLT-5041](https://linear.app/buildkite/issue/PLT-5041/agent-releases-borked)

### Testing

### Disclosures / Credits

Thanks to AWS support for pointing out our mistake after we sort of just assumed it was their fault 🫠

### DIFF
--- a/.buildkite/pipeline.release-unstable.yml
+++ b/.buildkite/pipeline.release-unstable.yml
@@ -17,6 +17,7 @@ steps:
             - organization_slug
             - organization_id
             - pipeline_slug
+            - branch_name
       - ecr#v2.7.0:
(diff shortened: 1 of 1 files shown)


duckalini:
Trust Policy updated in IAM

PR #3615, opened 2025-12-08
More custom artifact bucket fixes

moskyb:
### Description

In https://github.com/buildkite/agent/pull/3607, we partially fixed an issue where artifacts uploaded to a customer-managed artifacts bucket would be prefixed with an extra `/`, leading to the artifacts not being downloadable.

However, in that PR, there was a case that we missed. This PR adds code to cover the missed case, as well as adding an end-to-end test to verify that the behaviour is working as intended, preventing the issue from recurring.

### Context

PS-1470
[The original PR](https://github.com/buildkite/agent/pull/3607)
Some infra PRs to create the infra the E2E tests need:
- https://github.com/buildkite/terraform-modules/pull/70
- https://github.com/buildkite/aws-buildkite-dist/pull/151

### Testing

### Disclosures / Credits

As with the original PR, i used Amp as a tool to investigate the original issue, as well as several testing issues i encountered along the way. In some cases it was useful, and in others it led me on wild goose chases.

### DIFF
--- a/.buildkite/Dockerfile-e2e
+++ b/.buildkite/Dockerfile-e2e
@@ -1,1 +1,9 @@
 FROM public.ecr.aws/docker/library/golang:1.24.11@sha256:cf1272dbf972a94f39a81dcb9dc243a8d2f981e5dd3b5a5c965f6d9ab9268b26
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    unzip \
+    curl \
+    jq \
+    && curl "https://awscli.amazonaws.com/awscli-exe-linux-$(uname -m).zip" -o "awscliv2.zip" \
+    && unzip awscliv2.zip \
+    && ./aws/install


internal/e2e/fixtures/artifact_upload_download.yaml
zhming0, line 1: what is the background behind changing the artifact file name?

internal/e2e/fixtures/artifact_custom_s3_bucket.yaml
```diff
@@ -0,0 +1,31 @@
...
+  # NB: not using ACLs (which is what this ACL does) means artifacts won't be available from the buildkite UI.
+  # This isn't really a big deal because the jobs and pipelines that these tests generate are ephemeral
+  BUILDKITE_S3_ACL: bucket-owner-full-control
+agents:
+  queue: {{ .queue }}
+steps:
+  - key: upload
+    plugins:
+      - aws-assume-role-with-web-identity#v1.4.0:
+          role-arn: arn:aws:iam::172840064832:role/pipeline-agent-e2e-testing-buildkite-agent-e2e
```
zhming0, line 13: Did we do a terraform dance somewhere or click opsed the change? maybe worth to leave a small doc in our internal comms so we know where to look for things.

internal/e2e/artifact_test.go
```diff
@@ -19,3 +19,21 @@ func TestArtifactUploadDownload(t *testing.T) {
...
+	ctx := t.Context()
+	tc := newTestCase(t, "artifact_custom_bucket.yaml")
+
+	tc.startAgent()
+	build := tc.triggerBuild()
+	state := tc.waitForBuild(ctx, build)
+
+	logs := tc.fetchLogs(ctx, build)
+	t.Log("Job logs:")
+	t.Log(logs)
```
zhming0: [nit] YAGNI?

internal/e2e/testcase.go
```diff
@@ -106,22 +106,29 @@ func newTestCase(t testing.TB, file string) *testCase {
...
 	if err != nil {
 		t.Fatalf("Could not create pipeline with the following config in org %q: testHelper.createPipeline(%q, pipelineCfg) error = %v\n%s", targetOrg, name, err, pipelineCfg.String())
 	}
 	t.Cleanup(func() {
-		if err := cleanupPipeline(); err != nil {
-			t.Logf("Could not clean up pipeline %q (id = %s) in org %q: cleanup() = %v", pipeline.Slug, pipeline.ID, targetOrg, err)
-		}
+		// if err := cleanupPipeline(); err != nil {
+		// 	t.Logf("Could not clean up pipeline %q (id = %s) in org %q: cleanup() = %v", pipeline.Slug, pipeline.ID, targetOrg, err)
+		// }
```
zhming0: should we uncomment these?

internal/e2e/artifact_test.go
```diff
@@ -19,3 +19,21 @@ func TestArtifactUploadDownload(t *testing.T) {
 		t.Errorf("Build state = %q, want %q", got, want)
 	}
 }
+
+// Test that an agent can upload and download artifact to/from a customer-managed S3 bucket
+func TestArtifactUploadDownload_CustomBucket(t *testing.T) {
+	ctx := t.Context()
+	tc := newTestCase(t, "artifact_custom_bucket.yaml")
```
zhming0:
```suggestion
	tc := newTestCase(t, "artifact_custom_s3_bucket.yaml")
```

we are soon introducing GCP and Azure equivalent, might be worth to give a more specific name

zhming0 (commented):
I see this PR is in draft. But looks good, left some comments

internal/e2e/testcase.go
moskyb: oh yes, definitely

internal/e2e/fixtures/artifact_upload_download.yaml
moskyb: ah, i changed it on the custom bucket one so i did it on this one too, but i now realise that this one doesn't need it. i'll get rid of it.

internal/e2e/fixtures/artifact_custom_s3_bucket.yaml
moskyb: yeah, there's a couple of PRs around. I'll put them in a message in slack.

internal/e2e/artifact_test.go
DrJosh9000: If the pipeline is cleaned up automatically, this would be the next most accessible way of reading the job logs in case the test fails.

.buildkite/Dockerfile-e2e
```diff
@@ -1 +1,12 @@
...
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    unzip \
+    curl \
+    jq \
+    && curl "https://awscli.amazonaws.com/awscli-exe-linux-$(uname -m).zip" -o "awscliv2.zip" \
+    && unzip awscliv2.zip \
+    && ./aws/install \
+    && rm -rf awscliv2.zip aws \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
```
DrJosh9000: These cleanups don't seem like much of a win, because the image is only used once...

moskyb: hmm, good point!

PR #3627, opened 2025-12-11
feat: add support for concurrent save and restore operations

wolfeidau:
### Description

This may add a bit of contention but will help with situations where there are many small, sub 1gb caches.

I have followed the prevailing style with the concurrency implementation.

### Context

https://linear.app/buildkite/issue/MDC-822/add-support-for-concurrent-save-and-restore-of-caches-in-the-buildkite

### Changes

Adds a new option to cache CLI operations and some logic to save and restore to support concurrent operations.

### Testing

Updated tests.

### Disclosures / Credits

AMP Code was used to review concurrency options and draft this change, and update tests, then I adjusted a few things.

### DIFF
--- a/clicommand/cache_shared.go
+++ b/clicommand/cache_shared.go
@@ -20,7 +20,7 @@ func cacheFlags() []cli.Flag {
 		cli.StringSliceFlag{
 			Name:   "ids",
 			Value:  &cli.StringSlice{},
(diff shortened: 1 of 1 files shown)


CerealBoy (approved):
Would it be a good idea to have a couple of extra test cases, specifically for values like `0` and `100` / having many blobs?

Otherwise LGTM 🚀 🚀

PR #3628, opened 2025-12-11
PS-1491: Fix double retry issue for k8s mode bootstrap

zhming0:
### Description

Starting #3306 and subsequently worsen by #3398, we have created a double retry mechanism that will make k8s agent to spend 1200 seconds on retry when it's unable to connect to the coordinating agent container. 

This PR remove double retry mechanism, centralize the retry logic to socket.Connect.

### Context

PS-1491

### Testing

### Disclosures / Credits

Human

### DIFF
--- a/kubernetes/client.go
+++ b/kubernetes/client.go
@@ -23,16 +23,22 @@ type Client struct {
 
 var errNotConnected = errors.New("client not connected")
 
+// Connect establishes a connection to the Agent container in the same k8s pod and registers the client.
+// Because k8s might run the containers "out of order", the server socket might not exist yet,
+// so this method retries the connection with a 1-second interval until the context is cancelled.
+// Callers should use context.WithTimeout to control the connection timeout.
 func (c *Client) Connect(ctx context.Context) (*RegisterResponse, error) {
 	if c.SocketPath == "" {
 		c.SocketPath = defaultSocketPath
 	}
 
-	// Because k8s might run the containers "out of order", the server socket
-	// might not exist yet. Try to connect several times.
+	// Retry until the context is cancelled. The high maxAttempts is a safety net
+	// in case the caller forgets to set a context deadline - in practice the
+	// context deadline should be the limiting factor.
+	const retryInterval = time.Second
 	r := roko.NewRetrier(
-		roko.WithMaxAttempts(30),
-		roko.WithStrategy(roko.Constant(time.Second)),
+		roko.WithMaxAttempts(3600),
+		roko.WithStrategy(roko.Constant(retryInterval)),
 	)
 	client, err := roko.DoFunc(ctx, r, func(*roko.Retrier) (*rpc.Client, error) {
 		return rpc.DialHTTP("unix", c.SocketPath)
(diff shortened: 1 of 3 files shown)


internal/job/executor.go
```diff
@@ -1366,13 +1366,9 @@ func (e *Executor) setupRedactors(log shell.Logger, environ *env.Environment, st
...
-	rtr := roko.NewRetrier(
-		roko.WithMaxAttempts(7),
-		roko.WithStrategy(roko.Exponential(2*time.Second, 0)),
-	)
-	regResp, err := roko.DoFunc(ctx, rtr, func(rtr *roko.Retrier) (*kubernetes.RegisterResponse, error) {
-		return k8sAgentSocket.Connect(ctx)
-	})
+	// Use a default connection timeout of 120 seconds for kubernetes socket connection.
+	const connectionTimeout = 120 * time.Second
+	regResp, err := k8sAgentSocket.Connect(ctx, connectionTimeout)
```
DrJosh9000: What about using a context to supply the timeout?

```suggestion
	ctx, cancel := context.WithTimeout(ctx, 120 * time.Second)
	defer cancel()
	regResp, err := k8sAgentSocket.Connect(ctx)
```

zhming0: Yep, sure.

internal/job/executor.go
```diff
@@ -1366,13 +1366,8 @@ func (e *Executor) setupRedactors(log shell.Logger, environ *env.Environment, st
 }
 
 func (e *Executor) kubernetesSetup(ctx context.Context, environ *env.Environment, k8sAgentSocket *kubernetes.Client) error {
```
zhming0: This is dead code. To be removed in https://github.com/buildkite/agent/pull/3629

kubernetes/client.go
```diff
@@ -23,16 +23,26 @@ type Client struct {
...
-	// Because k8s might run the containers "out of order", the server socket
-	// might not exist yet. Try to connect several times.
+	timeout := 120 * time.Second
+	if deadline, ok := ctx.Deadline(); ok {
+		timeout = time.Until(deadline)
+	}
+
+	const retryInterval = time.Second
+	maxAttempts := max(int(timeout.Seconds()/retryInterval.Seconds()), 1)
+
```
DrJosh9000: This can be simplified a bit. It doesn't matter if `maxAttempts` is fixed at 120 - if the context times out after 120 seconds then the last few attempts won't happen (roko is context aware).

DrJosh9000 (approved):
Aside from my latest comment, LGTM

kubernetes/client.go
zhming0: Can do but the issue is that I can't predict how high people will set the timeout to be 😬 . I set a 3600 max attempt as safety net, wdyt?

DrJosh9000: 3600 sounds reasonable. The other path would be `roko.TryForever`, if a user wants to configure the option arbitrarily poorly we could allow that.

PR #3635, opened 2025-12-13
SUP-5826: Remove experiment from 'env' command

Mykematt:
### Description
Remove experiment from 'buildkite-agent env' command since it has been promoted since agent version v3.64.0

### Context
- Clarity in documentation
- Linear: https://linear.app/buildkite/issue/SUP-5826/docs-remove-experiment-label-from-buildkite-agent-env-command

### Changes
- `./buildkite-agent-dev env get` now outputs:
> buildkite-agent: fatal: Could not create Job API client: BUILDKITE_AGENT_JOB_API_SOCKET empty or undefined
> This command can only be used from hooks or plugins running under a job executor
> where the Job API (available in agent v3.64.0 and later) is available.

### Testing

### Disclosures / Credits

I did not use AI tools

### DIFF
--- a/clicommand/env_get.go
+++ b/clicommand/env_get.go
@@ -12,7 +12,7 @@ import (
 
 const envClientErrMessage = `Could not create Job API client: %w
 This command can only be used from hooks or plugins running under a job executor
-where the Job API (available in agent v3.64.0 and later) is available.`
+where the agent's job API is available (in version v3.64.0 and later of the Buildkite Agent).`
 
 const envGetHelpDescription = `Usage:
 


DrJosh9000 (approved):
LGTM

clicommand/env_get.go
```diff
@@ -12,7 +12,7 @@ import (
 
 const envClientErrMessage = `Could not create Job API client: %w
 This command can only be used from hooks or plugins running under a job executor
-where the "job-api" experiment is enabled.`
+where the Job API (available in agent v3.64.0 and later) is available.`
```
gilesgas: Is this a reference to the [Jobs API](https://buildkite.com/docs/apis/rest-api/jobs)?

If so, please implement these updates for consistency, and to be in line with how we've referenced Buildkite Agent versions for other CLI wording throughout this repo:


```suggestion
where the agent's job API is available (in version v3.64.0 and later of the Buildkite Agent).`
```

Also, this rewording removes duplication of the word 'available', which should hopefully make this a tad easier to scan.

DrJosh9000: This would be a reference to [the agent's internal/local API](https://buildkite.com/docs/agent/v3#promoted-experiments-job-api), not the Jobs API.

gilesgas: Thanks for clarifying this @DrJosh9000 ! I've tweaked the suggestion above and'll commit it. Will then approve this PR to be merged in.

PR #3676, opened 2026-01-21
A-1118: Replace addRepositoryHostToSSHKnownHosts with StrictHostKeyChecking=accept-new

zhming0:
### Description

Simplifies SSH host key verification by using OpenSSH's `StrictHostKeyChecking=accept-new` option instead of running ssh-keyscan before git operations.

* Set `GIT_SSH_COMMAND="ssh -o StrictHostKeyChecking=accept-new"` when `SSHKeyscan` is enabled
* Removed ssh-keyscan invocations and known_hosts file manipulation
* Deleted `ssh.go`, `knownhosts.go` and their tests (~450 lines removed)

* Eliminates dependency on ssh-keyscan binary
* Removes Windows-specific path hackery to find SSH tools bundled with Git
* Simpler, fewer moving parts

One important note is this PR raises our minimum SSH version requirement to ~2017. You can vote yes or no 🙂.

Re using `GIT_SSH_COMMAND`, I do have concern over whether env var is the right approach, but I am running out of ideas, I am open to suggestions 🙌🏿 . 

### Context

A-1118

### Testing

### Disclosures / Credits

LLM powered this time


internal/job/ssh.go
```diff
@@ -1,89 +1,50 @@
...
+		return false
+	}
+	return major > 7 || (major == 7 && minor >= 6)
+}
 
-func sshKeyScan(ctx context.Context, sh *shell.Shell, host string) (string, error) {
-	toolsDir, err := findPathToSSHTools(ctx, sh)
+// parseSSHVersion runs "ssh -V" and parses the version number.
+// Returns (major, minor, true) on success, (0, 0, false) on failure.
+func parseSSHVersion() (major, minor int, ok bool) {
```
DrJosh9000: I would expect `parseSSHVersion` to do what `extractSSHVersion` does, and for this function to just be called `sshVersion` - thoughts?

internal/job/ssh.go
```diff
@@ -1,89 +1,50 @@
...
-// Some more details on the relative paths at
-// https://stackoverflow.com/a/11771907
-func findPathToSSHTools(ctx context.Context, sh *shell.Shell) (string, error) {
-	sshKeyscan, err := sh.AbsolutePath("ssh-keyscan")
-	if err == nil {
-		return filepath.Dir(sshKeyscan), nil
+// extractSSHVersion parses an SSH version string like "OpenSSH_8.9p1" or "OpenSSH_7.6p1".
+// Returns (major, minor, true) on success.
+func extractSSHVersion(output string) (major, minor int, ok bool) {
+	re := regexp.MustCompile(`OpenSSH[_\s](\d+)\.(\d+)`)
```
DrJosh9000: Move this line in to a package-level var, so it doesn't recompile the regex for every call, and if there's a compilation problem it crashes the binary right away.

internal/job/ssh_host_key_checking.go
```diff
@@ -0,0 +1,32 @@
+package job
+
+// configureSSHKeyscan sets up GIT_SSH_COMMAND with host key checking options.
+// If GIT_SSH is set (a binary path), we skip configuration since we can't add flags.
+// For SSH >= 7.6, uses StrictHostKeyChecking=accept-new (TOFU: accept new, reject changed).
+// For older SSH, falls back to StrictHostKeyChecking=no with ephemeral known_hosts.
+func (e *Executor) configureSSHKeyscan() {
```
DrJosh9000: Naming things, again: this function now has nothing to do with `keyscan`, perhaps  `configureSSHKeyChecking` is better?

internal/job/ssh_host_key_checking.go
```diff
@@ -0,0 +1,32 @@
...
+	var sshOptions string
+	if sshSupportsAcceptNew() {
+		// OpenSSH 7.6+ supports accept-new: accept new host keys, reject changed ones
+		sshOptions = "-o StrictHostKeyChecking=accept-new"
+	} else {
+		// Older SSH: disable host key checking entirely
+		// Use /dev/null for known_hosts to avoid polluting the user's file
+		sshOptions = "-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null"
+		e.shell.Commentf("SSH version < 7.6 detected, using StrictHostKeyChecking=no")
+	}
```
DrJosh9000: I think "don't indent the main flow" applies here; "super old SSH" almost feels like an error case:

```
// OpenSSH 7.6+ supports accept-new: accept new host keys, reject changed ones
sshOptions := "-o StrictHostKeyChecking=accept-new"
if !sshSupportsAcceptNew() {
	// Older SSH: disable host key checking entirely
	// Use /dev/null for known_hosts to avoid polluting the user's file
	sshOptions = "-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null"
	e.shell.Commentf("SSH version < 7.6 detected, using StrictHostKeyChecking=no")
}
```

DrJosh9000 (changes requested):
Mainly style suggestions

DrJosh9000 (approved):
LGTM ✅

PR #3714, opened 2026-02-20
Surface secret fetch retry warnings in job output

blaknite:
### Description

The buffer logger (`logger.NewBuffer()`) passed to `secrets.FetchSecrets` was never flushed, so when secret fetching hit retryable errors (HTTP 502, 429, TLS timeouts), the `l.Warn()` calls about retries went into the buffer and vanished. Users would see their step appear to hang with no explanation of what was happening.

The buffer logger was originally introduced in #3453 to satisfy `api.NewClient`'s logger requirement and suppress debug HTTP header noise. When retry logic was added in #3706, `FetchSecrets` reused the same buffer logger without accounting for the fact that it now produces meaningful warnings.

After `FetchSecrets` returns, we now iterate the buffer's messages and forward any warnings to the shell logger so they appear in the job output.

### Context

[A-959](https://linear.app/buildkite/issue/A-959/secret-fetch-retry-warnings-are-silently-swallowed)

### Changes

After `FetchSecrets` completes, warning messages from the buffer logger are forwarded to the shell logger. The API client keeps its buffer logger to suppress debug noise.

### Testing

### Disclosures / Credits

Amp wrote the implementation.

### DIFF
--- a/internal/job/executor.go
+++ b/internal/job/executor.go
@@ -915,8 +915,19 @@ func (e *Executor) fetchAndSetSecrets(ctx context.Context) error {
 		Token:    e.shell.Env.GetString("BUILDKITE_AGENT_ACCESS_TOKEN", ""),
 	})
 
-	// Fetch all secrets
+	// Fetch all secrets. We pass secretLogger (a buffer) here because
+	// FetchSecrets takes a logger.Logger, but retry warnings within it need
+	// to reach the user. We flush those from the buffer afterwards.
 	fetchedSecrets, errs := secrets.FetchSecrets(ctx, secretLogger, apiClient, e.JobID, keys, 10)
+
+	// Surface any retry warnings that were buffered during fetching.
+	// The buffer logger prefixes warn messages with "[warn] ".
+	for _, msg := range secretLogger.Messages {
+		if after, ok := strings.CutPrefix(msg, "[warn] "); ok {
+			e.shell.Warningf("%s", after)
+		}
+	}
+
 	if len(errs) > 0 {
 		var errorMsg strings.Builder
 		for _, err := range errs {


internal/job/executor.go
```diff
@@ -915,8 +915,19 @@ func (e *Executor) fetchAndSetSecrets(ctx context.Context) error {
 		Token:    e.shell.Env.GetString("BUILDKITE_AGENT_ACCESS_TOKEN", ""),
 	})
 
-	// Fetch all secrets
+	// Fetch all secrets. We pass secretLogger (a buffer) here because
+	// FetchSecrets takes a logger.Logger, but retry warnings within it need
+	// to reach the user. We flush those from the buffer afterwards.
```
zhming0, line 920: This entire comment read a bit logically odd and misplaced 😅 
> We pass secretLogger (a buffer) here because FetchSecrets takes a logger.Logger, but retry warnings...


And i think this comment should just be added around the flushing code rather than here.

internal/job/executor.go
```diff
@@ -915,8 +915,19 @@ func (e *Executor) fetchAndSetSecrets(ctx context.Context) error {
...
 	fetchedSecrets, errs := secrets.FetchSecrets(ctx, secretLogger, apiClient, e.JobID, keys, 10)
+
+	// Surface any retry warnings that were buffered during fetching.
+	// The buffer logger prefixes warn messages with "[warn] ".
+	for _, msg := range secretLogger.Messages {
+		if after, ok := strings.CutPrefix(msg, "[warn] "); ok {
+			e.shell.Warningf("%s", after)
+		}
+	}
```
zhming0, line 929: Is this partially defeating the purpose of using Buffer logger? We haven't mutate redactor at this point, so we are running the risk of exposing secrets to log here.

zhming0 (changes requested):
I can see the rationale 👍🏿 , but I think it might be worth to re-look this implementation in security perspective.

PR #3720, opened 2026-02-25
Mandatory golangci lint

moskyb:
### Description

We've had golangci-lint as a soft-failing step in CI for a long time, and I think it's time we  turned it on for real. However, we have a lot of code that's not up to scratch.

That's ok though! I threw $10 at the sourcegraph corporation and they did a perfectly adequate job of fixing all of our GCIL errors.

### Changes

- errcheck: Added //nolint:errcheck pragmas with explanatory comments for deferred `Close()`, `os.Remove()`, logger/test writes, `http.Serve()` in goroutines, and similar best-effort operations
- staticcheck ST1005: Lowercased error string prefixes and removed trailing punctuation per Go conventions
- staticcheck SA1019: Removed deprecated `net.Error.Temporary()` call (timeout check already covers it)
- staticcheck QF1008: Simplified `p.Plugin.DisplayName()` → `p.DisplayName()` using promoted method
- unused: Removed unused constants and variables

### Testing

### Disclosures / Credits

Look, none of this was me, it was all Amp, but i'm so glad for that.

### DIFF
--- a/agent/integration/job_verification_integration_test.go
+++ b/agent/integration/job_verification_integration_test.go
@@ -347,23 +347,6 @@ var (
 		},
 		Token: "bkaj_job-token",
 	}
-
-	jobWithMismatchedSecrets = api.Job{
-		ChunksMaxSizeBytes: 1024,
-		ID:                 defaultJobID,
-		Step: pipeline.CommandStep{
-			Command: "echo hello world",
-			Secrets: pipeline.Secrets{
-				pipeline.Secret{Key: "DATABASE_URL", EnvironmentVariable: "DATABASE_URL"},
-			},
-		},
-		Env: map[string]string{
-			"BUILDKITE_COMMAND": "echo hello world",
-			"BUILDKITE_REPO":    defaultRepositoryURL,
-			"DEPLOY":            "0",
-		},
-		Token: "bkaj_job-token",
-	}
 )
 
 func TestJobVerification(t *testing.T) {
--- a/internal/artifact/bk_uploader.go
+++ b/internal/artifact/bk_uploader.go
@@ -337,7 +337,7 @@ func (m *multipartStreamer) WriteField(key, value string) error {
 // This can only be called once and must be the last thing written to the streamer
 func (m *multipartStreamer) WriteFile(key, artifactPath string, fh http.File) error {
 	if m.reader != nil {
-		return errors.New("WriteFile can't be called multiple times")
+		return errors.New("writeFile can't be called multiple times")
 	}
 
 	// Set up a reader that combines the body, the file and the closer in a stream
--- a/internal/artifact/s3.go
+++ b/internal/artifact/s3.go
@@ -141,10 +141,10 @@ func NewS3Client(ctx context.Context, l logger.Logger, bucket string) (*s3.Clien
 		const errorTitle = "could not authenticate to AWS S3 using any of the included credential providers."
 
 		if hasProxy && !hasNoProxyIdmsException {
-			return nil, fmt.Errorf("%s Your HTTP proxy settings do not grant a NO_PROXY=169.254.169.254 exemption for the instance metadata service, instance profile credentials may not be retrievable via your HTTP proxy.", errorTitle)
+			return nil, fmt.Errorf("%s your HTTP proxy settings do not grant a NO_PROXY=169.254.169.254 exemption for the instance metadata service, instance profile credentials may not be retrievable via your HTTP proxy", errorTitle)
 		}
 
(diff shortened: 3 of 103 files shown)


agent/integration/job_verification_integration_test.go
```diff
@@ -347,23 +347,6 @@ var (
 		},
 		Token: "bkaj_job-token",
 	}
-
-	jobWithMismatchedSecrets = api.Job{
```
DrJosh9000, line 351: This is unused? _Should_ it be used?

internal/artifact/bk_uploader.go
```diff
@@ -337,7 +337,7 @@ func (m *multipartStreamer) WriteField(key, value string) error {
 // This can only be called once and must be the last thing written to the streamer
 func (m *multipartStreamer) WriteFile(key, artifactPath string, fh http.File) error {
 	if m.reader != nil {
-		return errors.New("WriteFile can't be called multiple times")
+		return errors.New("writeFile can't be called multiple times")
```
DrJosh9000, line 340: `WriteFile` is referring to the method called `WriteFile`. Maybe the error should be rephrased:


```suggestion
		return errors.New("file already written")
```

internal/artifact/s3.go
```diff
@@ -141,10 +141,10 @@ func NewS3Client(ctx context.Context, l logger.Logger, bucket string) (*s3.Clien
...
 		}
 
-		return nil, fmt.Errorf("%s You can authenticate by setting Buildkite environment variables (BUILDKITE_S3_ACCESS_KEY_ID, BUILDKITE_S3_SECRET_ACCESS_KEY, BUILDKITE_S3_PROFILE), AWS environment variables (AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_PROFILE), Web Identity environment variables (AWS_ROLE_ARN, AWS_ROLE_SESSION_NAME, AWS_WEB_IDENTITY_TOKEN_FILE), or if running on AWS EC2 ensuring network access to the EC2 Instance Metadata Service to use an instance profile’s IAM Role credentials.", errorTitle)
+		return nil, fmt.Errorf("%s you can authenticate by setting Buildkite environment variables (BUILDKITE_S3_ACCESS_KEY_ID, BUILDKITE_S3_SECRET_ACCESS_KEY, BUILDKITE_S3_PROFILE), AWS environment variables (AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_PROFILE), Web Identity environment variables (AWS_ROLE_ARN, AWS_ROLE_SESSION_NAME, AWS_WEB_IDENTITY_TOKEN_FILE), or if running on AWS EC2 ensuring network access to the EC2 Instance Metadata Service to use an instance profile’s IAM Role credentials", errorTitle)
```
DrJosh9000, line 147: This is another example complicated by `errorTitle` being the actual start of the error string, with the rest being "hint to the user". 

What if we added an `ErrorWithHint` type that could be examined when the error is printed?

```go
type ErrorWithHint struct {
	Err error
	Hint string
}

func (e ErrorWithHint) Unwrap() error { return e.Err }
func (e ErrorWithHint) Error() string { return e.Err.String() }

//...

if h := new(ErrorWithHint); errors.As(err, h) {
	fmt.Fprintln(c.App.ErrWriter, h.Hint)
}
```

internal/artifact/s3_downloader.go
```diff
@@ -52,7 +52,7 @@ func NewS3Downloader(l logger.Logger, c S3DownloaderConfig) *S3Downloader {
 
 func (d S3Downloader) Start(ctx context.Context) error {
 	if d.conf.S3Client == nil {
-		return fmt.Errorf("S3Downloader for %s: S3Client is nil", d.conf.S3Path)
+		return fmt.Errorf("s3Downloader for %s: S3Client is nil", d.conf.S3Path)
```
DrJosh9000, line 55: Again this is capitalised because it's referring to the type. Not sure what the best solution is here, but this seems wrong.

internal/job/integration/git.go
```diff
@@ -118,7 +118,7 @@ func newGitRepository() (*gitRepository, error) {
 	// EvalSymlinks seems best? Maybe there's a better way?
 	tempDir, err := filepath.EvalSymlinks(tempDirRaw)
 	if err != nil {
-		return nil, fmt.Errorf("EvalSymlinks for temp dir: %w", err)
+		return nil, fmt.Errorf("evalSymlinks for temp dir: %w", err)
```
DrJosh9000, line 121:
```suggestion
		return nil, fmt.Errorf("evaluating symlinks for temp dir: %w", err)
```

DrJosh9000:
I think this was superseded by #3841 - please resuscitate if there's something else here.

PR #3736, opened 2026-03-05
feat: support server-hinted artifact batch sizing

lox:
**Disclosure:** Ampcode assisted, human curated. 

### Summary

This change adds server-driven artifact batching controls to the agent and introduces light adaptive behaviour for artifact creation when rate-limited.

### What Changed

1. Added optional ping response fields for artifact batching hints:
   1. `artifact_create_batch_size`
   1. `artifact_update_batch_size_max`
1. Stored ping-provided hint values in the agent worker and injected them into job env so bootstrap/subcommands inherit them:
   1. `BUILDKITE_ARTIFACT_CREATE_BATCH_SIZE`
   1. `BUILDKITE_ARTIFACT_UPDATE_BATCH_SIZE_MAX`
1. Added CLI/env wiring in `artifact upload` for both knobs.
1. Updated artifact uploader config to accept:
   1. create batch size
   1. update batch size max
1. Updated `CreateArtifacts` batching logic to:
   1. use configurable initial batch size
   1. increase batch size on `429` responses (doubling, capped)
1. Updated `UpdateArtifacts` logic to chunk large state updates by configured max batch size.
1. Added focused tests for:
   1. ping hint propagation into job env
   1. create-batch growth on `429`
   1. update-state chunking by configured max

### Why

1. Reduce control-plane request pressure during very large artifact workloads.
1. Let the server steer batching behaviour per workload without requiring static global defaults.
1. Improve resilience under rate-limiting by adapting create request shape.
1. Keep behaviour backwards-compatible when hints are absent.

### Notes

1. Defaults remain unchanged when no hint/env is provided.
1. Validated with:
   1. `go tool gofumpt -extra -w ...` on modified files
   1. `go test ./api ./agent ./internal/artifact ./clicommand`

### DIFF
--- a/internal/artifact/batch_creator.go
+++ b/internal/artifact/batch_creator.go
@@ -25,6 +25,10 @@ type BatchCreatorConfig struct {
 
 	// Whether to allow multipart uploads to the BK-hosted bucket.
 	AllowMultipart bool
+
+	// Number of artifacts in each CreateArtifacts request.
+	// If zero, a default is used.
+	CreateBatchSize int
 }
 
 type BatchCreator struct {
@@ -48,12 +52,16 @@ func NewArtifactBatchCreator(l logger.Logger, ac APIClient, c BatchCreatorConfig
 
 func (a *BatchCreator) Create(ctx context.Context) ([]*api.Artifact, error) {
 	length := len(a.conf.Artifacts)
-	chunks := 30
+	batchSize := a.conf.CreateBatchSize
+	if batchSize <= 0 {
+		batchSize = 30
+	}
+	const maxCreateBatchSize = 500
 
 	// Split into the artifacts into chunks so we're not uploading a ton of
 	// files at once.
-	for i := 0; i < length; i += chunks {
-		j := min(i+chunks, length)
+	for i := 0; i < length; {
+		j := min(i+batchSize, length)
 
 		// The artifacts that will be uploaded in this chunk
 		theseArtifacts := a.conf.Artifacts[i:j]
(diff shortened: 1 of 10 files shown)


internal/artifact/batch_creator.go
```diff
@@ -48,12 +52,16 @@ func NewArtifactBatchCreator(l logger.Logger, ac APIClient, c BatchCreatorConfig
 
 func (a *BatchCreator) Create(ctx context.Context) ([]*api.Artifact, error) {
 	length := len(a.conf.Artifacts)
-	chunks := 30
+	batchSize := a.conf.CreateBatchSize
+	if batchSize <= 0 {
+		batchSize = 30
+	}
+	const maxCreateBatchSize = 500
```
quinn-diesel: why are you thinking 500 for the max batch size?

moskyb (commented):
i like the idea here, but given the churn going on with pings at the moment with agent push, i think potentially it might be tricky to get remote config manipulation working through both push and polling ping at the same time.

that said though, i think responding to 429s with increased batch sizes is a great idea, and i'd be happy to merge that side of this PR as-is. perhaps we could bump the batch size limit even further? 1000?

DrJosh9000:
Hi @lox, I see this one is still draft as well, was this still something you wanted?

lox:
I still think the original problem here was real, but the customer workload that motivated it has gone away and this draft has gone pretty stale.

Given the extra agent/backend config surface and behaviour here, I don’t think it’s worth carrying speculative complexity for now.

I’m inclined to close this and, if the scaling case comes back, recut it from current `main` with fresh evidence from the newer artifact telemetry.

PR #3737, opened 2026-03-05
feat: honour Retry-After for artifact retries

lox:
### Summary

This follow-up teaches artifact retry loops to honour server-provided `Retry-After` headers, so retry pacing can be coordinated by the backend during pressure events.

### What Changed

1. Added a small shared helper in `internal/artifact/retry_after.go` to parse and apply `Retry-After` values to retry intervals.
1. Wired `Retry-After` handling into `CreateArtifacts` retries in `internal/artifact/batch_creator.go`.
1. Wired `Retry-After` handling into `UpdateArtifacts` retries in `internal/artifact/uploader.go`.
1. Added tests in `internal/artifact/batching_test.go` for valid and invalid `Retry-After` behaviour.

### Why

1. Improves stability by letting the server shape retry timing during load.
1. Reduces risk of synchronized retry bursts from many agents.
1. Keeps existing retry behaviour intact when `Retry-After` is absent or invalid.

### Validation

1. `go tool gofumpt -extra -w internal/artifact/retry_after.go internal/artifact/batch_creator.go internal/artifact/uploader.go internal/artifact/batching_test.go`
1. `go test ./internal/artifact -count=1`
1. `go test ./api ./agent ./clicommand -count=1`

### Notes

1. This PR is stacked on top of #3736.

### DIFF
--- a/internal/artifact/BUILD.bazel
+++ b/internal/artifact/BUILD.bazel
@@ -16,6 +16,7 @@ go_library(
         "downloader.go",
         "gs_downloader.go",
         "gs_uploader.go",
+        "retry_after.go",
         "s3.go",
         "s3_downloader.go",
         "s3_uploader.go",
--- a/internal/artifact/batch_creator.go
+++ b/internal/artifact/batch_creator.go
@@ -99,6 +99,9 @@ func (a *BatchCreator) Create(ctx context.Context) ([]*api.Artifact, error) {
 			if resp != nil && resp.StatusCode == 429 {
 				saw429 = true
 			}
+			if applyRetryAfterHeader(resp, r) {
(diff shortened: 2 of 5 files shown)


moskyb (approved):
looks good! good to bring this in line with many of our other retry loops.

DrJosh9000:
Hi @lox, just wondering what the state is here. It's draft, but it's been approved - time to take it out of draft and merge?

lox:
Same call here as `#3736`.

I still think honouring `Retry-After` was the right follow-up for the original scaling problem, but that workload has since gone away and this draft is now pretty stale as well.

I don’t think it’s worth carrying the extra complexity without an active need, so I’m going to close this too. If the artifact control-plane pressure case comes back, we can recut both from current `main` with fresh evidence.

PR #3791, opened 2026-04-01
Add `meta-data set-batch` command

pda:
### Description

Add a `buildkite-agent meta-data set-batch` command for setting multiple meta-data key/value pairs in a single API request.

```
buildkite-agent meta-data set-batch "foo=bar" "baz=qux"
```

The batch operation is transactional — all items succeed or none do.

### Context

Companion to a new `POST /v3/jobs/:job_id/data/set-batch` API route, which is not yet generally available.

### Changes

- **API client**: `MetaDataBatch` type and `SetMetaDataBatch` method (`POST /jobs/{id}/data/set-batch`)
- **CLI command**: `meta-data set-batch` subcommand accepting `key=value` arguments (no stdin support)
- **Tests**: table-driven argument parsing tests and HTTP-level tests covering success, error, and no-retry on 401/404
- **Docs**: minor correction to import organization note in `AGENT.md`

### Testing

### Disclosures / Credits

Amp (Claude Opus 4.6) wrote the implementation and tests with human direction and review.

### Shipping

- [ ] Server-side support shipped and tested

### DIFF
--- a/clicommand/meta_data_set_batch.go
+++ b/clicommand/meta_data_set_batch.go
@@ -1,0 +1,129 @@
+package clicommand
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"slices"
+	"strings"
+	"time"
+
+	"github.com/buildkite/agent/v3/api"
+	"github.com/buildkite/agent/v3/internal/redact"
+	"github.com/buildkite/agent/v3/logger"
+	"github.com/buildkite/roko"
+	"github.com/urfave/cli"
+)
+
+const metaDataSetBatchHelpDescription = `Usage:
+
+    buildkite-agent meta-data set-batch <key=value>... [options...]
+
+Description:
+
+Set multiple meta-data key/value pairs on a build in a single request.
+
+Each argument must be in key=value format.
+
+Keys and values must be non-empty strings, and strings containing only
+whitespace characters are not allowed.
+
+Example:
+
+    $ buildkite-agent meta-data set-batch foo=bar "greeting=hello world"
+    $ buildkite-agent meta-data set-batch duration:spec/a.rb=2.341 duration:spec/b.rb=5.672`
+
+type MetaDataSetBatchConfig struct {
+	GlobalConfig
+	APIConfig
+
+	Job          string   `cli:"job" validate:"required"`
+	RedactedVars []string `cli:"redacted-vars" normalize:"list"`
+}
+
+var MetaDataSetBatchCommand = cli.Command{
+	Name:        "set-batch",
+	Usage:       "Set multiple meta-data key/value pairs on a build",
+	Description: metaDataSetBatchHelpDescription,
+	Flags: slices.Concat(globalFlags(), apiFlags(), []cli.Flag{
+		cli.StringFlag{
+			Name:   "job",
+			Value:  "",
+			Usage:  "Which job's build should the meta-data be set on",
+			EnvVar: "BUILDKITE_JOB_ID",
+		},
+		RedactedVars,
+	}),
+	Action: func(c *cli.Context) error {
+		ctx := context.Background()
+		ctx, cfg, l, _, done := setupLoggerAndConfig[MetaDataSetBatchConfig](ctx, c)
+		defer done()
+
+		args := c.Args()
+		if len(args) == 0 {
+			return errors.New("at least one key=value argument is required")
+		}
+
+		items, err := parseMetaDataBatchArgs(args)
+		if err != nil {
+			return err
+		}
+
+		needles, _, err := redact.NeedlesFromEnv(cfg.RedactedVars)
+		if err != nil {
+			return err
+		}
+
+		for i := range items {
+			if redactedValue := redact.String(items[i].Value, needles); redactedValue != items[i].Value {
+				l.Warnf("Meta-data value for key %q contained one or more secrets from environment variables that have been redacted. If this is deliberate, pass --redacted-vars='' or a list of patterns that does not match the variable containing the secret", items[i].Key)
+				items[i].Value = redactedValue
+			}
+		}
+
+		return setMetaDataBatch(ctx, cfg, l, items)
+	},
+}
+
+func parseMetaDataBatchArgs(args []string) ([]api.MetaData, error) {
+	items := make([]api.MetaData, 0, len(args))
+	for _, arg := range args {
+		key, value, ok := strings.Cut(arg, "=")
+		if !ok {
+			return nil, fmt.Errorf("invalid argument %q: must be in key=value format", arg)
+		}
+		if strings.TrimSpace(key) == "" {
+			return nil, fmt.Errorf("invalid argument %q: key cannot be empty, or composed of only whitespace characters", arg)
+		}
+		if strings.TrimSpace(value) == "" {
+			return nil, fmt.Errorf("invalid argument %q: value cannot be empty, or composed of only whitespace characters", arg)
+		}
+		items = append(items, api.MetaData{Key: key, Value: value})
+	}
+	return items, nil
+}
+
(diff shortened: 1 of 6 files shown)


clicommand/meta_data_set_batch.go
```diff
@@ -0,0 +1,129 @@
...
+
+	batch := &api.MetaDataBatch{Items: items}
+
+	if err := roko.NewRetrier(
+		roko.WithMaxAttempts(10),
+		roko.WithStrategy(roko.ExponentialSubsecond(2*time.Second)),
+	).DoWithContext(ctx, func(r *roko.Retrier) error {
+		resp, err := client.SetMetaDataBatch(ctx, cfg.Job, batch)
+		if resp != nil && (resp.StatusCode == 401 || resp.StatusCode == 404) {
+			r.Break()
```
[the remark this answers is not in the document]
pda: 🤖 Valid observation, but this intentionally matches the existing `meta-data set` command, which has the same retry behaviour (only breaks on 401/404). Broadening the no-retry set for 4xx could be a good improvement, but it should be done consistently across both commands.

clicommand/meta_data_set_batch.go
```diff
@@ -0,0 +1,129 @@
...
+
+const metaDataSetBatchHelpDescription = `Usage:
+
+    buildkite-agent meta-data set-batch <key=value>... [options...]
+
+Description:
+
+Set multiple meta-data key/value pairs on a build in a single request.
+
+Each argument must be in key=value format.
```
DrJosh9000, line 26: Should we support JSON input (over stdin)?

pda: Maybe!
Do you think that would change the CLI design? 🤔

Would it look like one/some of these…

```
buildkite-agent meta-data set-batch < meta-data.json

buildkite-agent meta-data set-batch --json < meta-data.json

buildkite-agent meta-data set-batch --from-file meta-data.json
```

I mostly didn't add it because it contained more decisions 😅

If we think the current `argv` approach needs to change to make room for JSON/stdin input, I'll revise it. But if we think there's room at add it later, I'd rather defer it.

pda: (In my initial use-case, I have a bash script coordinating some build concerns, and it's rounding up lots of information and then setting ~8 meta-data key=values. I could push that into JSON structure, perhaps using `jq` as a builder, but it's a much better fit to just build an array of `key=value` strings and put them on argv)

DrJosh9000: I foresee it being useful if something other than Bash is orchestrating. We have JSON input in some other commands. But I agree it can be done later.

clicommand/meta_data_set_batch.go
```diff
@@ -0,0 +1,129 @@
...
+		key, value, ok := strings.Cut(arg, "=")
+		if !ok {
+			return nil, fmt.Errorf("invalid argument %q: must be in key=value format", arg)
+		}
+		if strings.TrimSpace(key) == "" {
+			return nil, fmt.Errorf("invalid argument %q: key cannot be empty, or composed of only whitespace characters", arg)
+		}
+		if strings.TrimSpace(value) == "" {
+			return nil, fmt.Errorf("invalid argument %q: value cannot be empty, or composed of only whitespace characters", arg)
+		}
```
DrJosh9000, line 118: Should we offer a way to "clear" a particular key? `key=` (nothing following) seems like a reasonable syntax for that.

pda: My robots suggest I reply like this… 😅

> acknowledge it's a good CLI ergonomic, but argue that clearing meta-data should be designed end-to-end as its own feature (server + single + batch + CLI), not smuggled in via an empty-value convention. File a follow-up issue.

And yeah… I like the idea of being able to clear like that, but I'd love to defer it as out of scope for right now. I think it's a clean non-breaking-change path to disallow it now, and then allow it later (as opposed to the inverse).

clicommand/meta_data_set_batch.go
```diff
@@ -0,0 +1,129 @@
...
+
+		needles, _, err := redact.NeedlesFromEnv(cfg.RedactedVars)
+		if err != nil {
+			return err
+		}
+
+		for i := range items {
+			if redactedValue := redact.String(items[i].Value, needles); redactedValue != items[i].Value {
+				l.Warn("Meta-data value for key %q contained one or more secrets from environment variables that have been redacted. If this is deliberate, pass --redacted-vars='' or a list of patterns that does not match the variable containing the secret", items[i].Key)
```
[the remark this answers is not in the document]
pda: 🤖 Fixed in 1f1c167d — both `l.Warn` calls in `meta_data_set_batch.go` now use `l.Warnf`.

pda:
Rebased, eliminated test assertion usage, updated for `Warnf`, etc.

clicommand/meta_data_set_batch.go
```diff
@@ -0,0 +1,129 @@
...
+		key, value, ok := strings.Cut(arg, "=")
+		if !ok {
+			return nil, fmt.Errorf("invalid argument %q: must be in key=value format", arg)
+		}
+		if strings.TrimSpace(key) == "" {
+			return nil, fmt.Errorf("invalid argument %q: key cannot be empty, or composed of only whitespace characters", arg)
+		}
+		if strings.TrimSpace(value) == "" {
+			return nil, fmt.Errorf("invalid argument %q: value cannot be empty, or composed of only whitespace characters", arg)
+		}
```
DrJosh9000, line 118: We can make it easier to check for specific errors in the tests without resorting to substring matching if these returned values of custom error types. One approach (another might be to wrap sentinel error values):

```go
type invalidFormatError struct { arg string }

func (e invalidFormatError) Error() string { return fmt.Sprintf("invalid argument %q: must be in key=value format", e.arg)

type emptyKeyError struct { arg string }

func (e emptyKeyError) Error() string { return fmt.Sprintf("invalid argument %q: key cannot be empty, or composed of only whitespace characters", e.arg)

type emptyValueError struct { arg string }

func (e emptyValueError) Error() string { return fmt.Sprintf("invalid argument %q: value cannot be empty, or composed of only whitespace characters", e.arg)

...

		key, value, ok := strings.Cut(arg, "=")
		if !ok {
			return nil, invalidFormatError{arg}
		}
		if strings.TrimSpace(key) == "" {
			return nil, emptyKeyError{arg}
		}
		if strings.TrimSpace(value) == "" {
			return nil, emptyValueError{arg}
		}

...


		wantErr error

...

wantErr: invalidFormatError{arg: "foobar"},

...

if !errors.Is(err, tc.wantErr) { 
	t.Errorf("parseMetaDataBatchArgs(%v) error = %v, want %v", tc.args, err, tc.wantErr)
}
```

`errors.Is` would help tighten up the logic in the test - it wouldn't need to branch on `tc.wantErr != ""`.

DrJosh9000 (approved):
This is pretty good, I have a style comment that will supercharge your table driven test

clicommand/meta_data_set_batch.go
pda: 🤖 Good call — applied in 3a94f797 using the custom-error-type variant you outlined (with the missing `}` typos fixed 😉). Each error type carries the offending `arg`, so `errors.Is` on a struct value verifies both the kind and the arg, and the test loop collapses to a single `errors.Is(err, tc.wantErr)` check (works for the happy path too, since `errors.Is(nil, nil)` is true).

pda: Don't mind the sass from Amp/Opus there 😜

DrJosh9000:
Hey @pda, is this ready to ship?

pda:
@DrJosh9000 Nope, still waiting on server-side support 😅
I'll give that a bump.

pda:
Server-side support has merged.
I'll do some validation prior to merging this PR though.

PR #3794, opened 2026-04-06
Add BUILDKITE_GIT_CHECKOUT_TIMEOUT timeout to add upper duration on the checkout hook

brandonnovick:
## Description

Adds `--git-checkout-timeout` / `BUILDKITE_GIT_CHECKOUT_TIMEOUT` (seconds) to bound each git checkout *attempt*. When a single attempt exceeds the timeout its context is cancelled and the error flows into the existing retry loop. `0` (default) preserves current behavior.

## Why

Hung git operations (clone, fetch, checkout, submodule update) can wedge a job indefinitely. A per-attempt timeout lets the retry path do its job instead of waiting forever.

## Changes

- New `--git-checkout-timeout` flag, wired through `BootstrapConfig`, `AgentStartConfig`, `AgentConfiguration`, and `ExecutorConfig`.
- `runDefaultCheckoutAttempt` wraps `defaultCheckoutPhase` in `context.WithTimeout` inside the existing roko retry loop.
- `ExecutorConfig.ReadFromEnvironment` now supports `reflect.Int`-typed env-tagged fields.

## Testing

- `go test ./...` passes locally.
- Formatted with `go tool gofumpt -extra -w .`.

## Notes

Largely generated by Claude Code, then reviewed and tweaked by hand.


internal/job/checkout.go
zhming0, line 1: Adding a global timeout to checkout is something that we are a bit hesitant about, given there is an retry mechanism underneath the checkout. 

However, we think it's very reasonable to add a timeout setting for each retry attempt, it would:

* Kill a hung individual git operation quickly
* Still allow the retry loop to kick in and try again

Unblocking CI is our ultimate goal.

zhming0 (changes requested):
Thank you so much for raising this PR @brandonnovick , we agree with the general direction but we think it's best if we be putting the timeout per attempt. 

Please let us know you want to give it an another shot, or alternatively raising a feature request to us. 

Lastly, for our curiosity, how often is such git hanging issue happening for you?

internal/job/checkout.go
brandonnovick: Apologies for the delay! Thank you for the feedback! I've applied this change in the scope of each retry as opposed to globally. Let me know if there's anything else you'd like to see!

clicommand/global.go
```diff
@@ -266,6 +266,13 @@ var (
 		EnvVar: "BUILDKITE_GIT_SUBMODULE_CLONE_CONFIG",
 	}
 
+	GitCheckoutTimeoutFlag = cli.IntFlag{
+		Name:   "git-checkout-timeout",
+		Value:  0,
+		Usage:  "Seconds to allow for the entire git checkout phase (0 means no timeout)",
```
zhming0: I think this doc is no longer accurate after the latest change. It's no longer about the "entire" git checkout phase :)

internal/job/checkout.go
```diff
@@ -326,6 +364,10 @@ func (e *Executor) checkout(ctx context.Context) error {
 
 			return err
 		}); err != nil {
+			if errors.Is(err, errCheckoutAttemptTimeout) {
```
zhming0: We probably do not need a specific exit code in this case. From the caller's perspective, "checkout timed out on every attempt" is just a flavor of "checkout failed." There's no action a CI system or user would take differently based on exit code 95 vs a generic failure code.

internal/job/checkout.go
```diff
@@ -268,6 +293,19 @@ func (e *Executor) checkout(ctx context.Context) error {
...
+				// the retrier decide whether to retry; clean the checkout dir so
+				// the next attempt starts fresh in case git left things in a bad
+				// state when killed.
+				e.shell.Warningf("Checkout attempt timed out (%s)", r)
+				if err := e.removeCheckoutDir(); err != nil {
+					e.shell.Warningf("Failed to remove checkout dir while cleaning up after a checkout timeout: %v", err)
+				}
+				if err := e.createCheckoutDir(); err != nil {
+					return err
+				}
```
zhming0: does it need its own case? The default branch (line 312–324) already does the same thing: `removeCheckoutDir` + `createCheckoutDir` + lets the retrier continue. The only difference is the log message: "Checkout attempt timed out" vs "Checkout failed!".

Since the wrapped errCheckoutAttemptTimeout message already contains "checkout attempt timed out" in the error string, it would show up in the generic "Checkout failed! %s" format anyway. So the separate case is probably not necessary.

zhming0 (commented):
Thank you very much for making the changes @brandonnovick , I've left a few more comments, but it's looking good overall 👍🏿
Appreciate your patience with the back-and-forth here, checkout is a critical path in the Agent, so we want to be extra careful 🙏🏿

clicommand/global.go
brandonnovick: Corrected!

internal/job/checkout.go
brandonnovick: Makes sense, fixed!

internal/job/checkout.go
brandonnovick: Addressed, I believe

brandonnovick:
So sorry for the delay @zhming0 , I have been on vacation for the last 2 weeks......addressed the comments!

zhming0 (approved):
Thank you @brandonnovick !

PR #3801, opened 2026-04-07
Add an option to fail startup if we cannot fetch instance tags

omehegan:
### Description

This is a customer-requested feature. They use EC2 instance tags to signal which AWS launch template should be used to launch an instance. In their workflow, when instance tags are unavailable (they've seen this due to AWS API rate-limiting), the agent will start up and connect to Buildkite, but isn't usable for their builds. They requested an option to make this conditional fatal during agent startup, to make it easier to diagnose and respond to the issue.

I considered using an agent-startup hook to check this and abort the startup if tags are not available, but this code change is pretty self-contained, and it felt low-risk and cleaner than the hook workaround.

### Context

PF-9576

### Changes

Adds a `--fail-on-missing-tags` argument to `buildkite-agent start`. 

> Exit the agent with an error if any enabled cloud tag source (EC2, ECS, GCP) fails to return tags (default: false) [$BUILDKITE_AGENT_FAIL_ON_MISSING_TAGS]

### Testing

### Disclosures / Credits

Claude wrote the code and tests. I reviewed the changes myself, to the best of my ability, and also used a separate Claude agent to poke holes, and made a couple of minor improvements based on that feedback.

### DIFF
--- a/agent/tags.go
+++ b/agent/tags.go
@@ -27,14 +27,17 @@ type FetchTagsConfig struct {
 	TagsFromGCPMetaDataPaths  []string
 	TagsFromGCPLabels         bool
 	TagsFromHost              bool
+	FailOnMissingTags         bool
 	WaitForEC2TagsTimeout     time.Duration
 	WaitForEC2MetaDataTimeout time.Duration
 	WaitForECSMetaDataTimeout time.Duration
 	WaitForGCPLabelsTimeout   time.Duration
 }
 
-// FetchTags loads tags from a variety of sources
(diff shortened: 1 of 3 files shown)


ss1909:
Test results:
1. --tags-from-ec2-meta-data-paths "my-tag=meta-data/instance-id" \
image

2. --tags-from-ec2-tags \
image

3. --tags-from-ec2-meta-data \      
image

4. --tags-from-ec2
image

ss1909 (approved):

ss1909:
Made one little change and did a bunch of testing, but lgtm!

PR #3812, opened 2026-04-13
A-1115: refactor ArtifactBatchCreator to use iterator

zhming0:
### Description

When there are many many small artifact files being uploaded, we currently get the presigned S3 url in the beginning. It could result in a situation, the combined round trip times are beyond S3 presidnged url TTL, causing artifact upload to fail. 

This PR refactor ArtifactBatchCreator to use iterator, so only obtain the upload right before yielding. 

### Context

A-1115

### Disclosures / Credits

My LLM minion

### DIFF
--- a/internal/artifact/batch_creator.go
+++ b/internal/artifact/batch_creator.go
@@ -2,6 +2,8 @@ package artifact
 
 import (
 	"context"
+	"iter"
+	"slices"
 	"time"
 
 	"github.com/buildkite/agent/v3/api"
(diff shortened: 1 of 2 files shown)


internal/artifact/batch_creator.go
```diff
@@ -46,82 +47,84 @@ func NewArtifactBatchCreator(l logger.Logger, ac APIClient, c BatchCreatorConfig
...
+// Batches yields successive chunks of artifacts after batch-creating them on
+// Buildkite (which assigns IDs and upload instructions). Each chunk is created
+// immediately before it is yielded, keeping presigned upload URLs fresh.
+func (a *BatchCreator) Batches(ctx context.Context) iter.Seq2[[]*api.Artifact, error] {
+	return func(yield func([]*api.Artifact, error) bool) {
+		length := len(a.conf.Artifacts)
+		const chunks = 30
+
+		for i := 0; i < length; i += chunks {
+			j := min(i+chunks, length)
```
moskyb: we can use the stdlib [`slices.Chunk`](https://pkg.go.dev/slices#Chunk) function here

```suggestion
		for _, batch := range slices.Chunk(a.conf.Artifacts, 30) {
```

moskyb (approved):
LGTM!

internal/artifact/uploader.go
```diff
@@ -107,21 +107,24 @@ func (a *Uploader) Upload(ctx context.Context) error {
 		artifact.URL = uploader.URL(artifact)
 	}
 
-	// Batch-create the artifact records on Buildkite
+	// Batch-create artifact records on Buildkite and upload each batch
+	// immediately. This keeps presigned upload URLs fresh — if all
+	// artifacts were batch-created upfront, URLs for artifacts near the
```
DrJosh9000: "batch-created" could mean the old way or the new way. Suggestion:

```suggestion
	// artifacts were created upfront in a single batch, URLs for artifacts near the
```

DrJosh9000 (approved):
LGTM with @moskyb's suggestion, here's a bonus one from me

PR #3814, opened 2026-04-13
Big process review & refactor

DrJosh9000:
### Description

Review and refactor of chunks of the `process` package, because it was getting pretty crusty, but also to solve a specific problem to do with interrupting programs on Windows.

### Context

https://linear.app/buildkite/issue/A-1075

The current mechanism used to interrupt processes on Windows is to send a "Ctrl-Break" event (as though someone was typing that key combination on the console window). This can be applied to all processes in a process group (created by an option on the Win32 `CreateProcess` call) and is understood to be vaguely similar to `SIGTERM` on *nixes...but it really isn't the same.

Go binaries seem to understand Ctrl-Break as a signal, so we don't need to change how the agent signals the bootstrap. But some processes (PowerShell, ping, probably others) absorb Ctrl-Break and do something other than exit.

Ctrl-C also exists as a concept on Windows and is supported as a generated console event. Unfortunately there are Windows-isms in the way - `GenerateConsoleCtrlEvent` understands `CTRL_C_EVENT`, but can't target it to a process group, only the current console. So we have to create the process in a new console, then detach console, attach to the other console, set an event handler, send the Ctrl-C, then detach and reattach to the original console and clean up the event handler. And I can't get that to work!

There's no particular reason we can't support `SIGKILL` as an "interrupt" signal either. Maybe the user just wants to kill stuff right away? But it isn't a handle-able signal, so we should avoid using it to interrupt the bootstrap (as a subprocess of `start` or `kubernetes-bootstrap`), otherwise post-command, pre-exit, artifact etc hooks won't run.

But... interrupting with `SIGKILL` will let us interrupt processes that absorb Ctrl-Break. So the solution is this PR, plus pass `--cancel-signal=SIGKILL`.

### Changes

- Add `SIGKILL` as an understood cancel signal
- For Windows, make `SIGKILL` call `terminateProcessGroup` - this isn't needed on *nixes because sending an actual `SIGKILL` to an actual process group _is_ how that works.
- Refactors:
  - Move `process` into `internal`, because it's not an intended API surface
  - Split out `setup`, `start`, `startWithPTY`, `startWithoutPTY`, `copyPTYToStdout`, `complete` from `Run`
  - Use `exec.CommandContext` and `Command.Cancel` to set up the interrupt-grace period-terminate handling, rather than running a goroutine waiting for context cancellation the whole time
  - Use `cmp.Or` a bit more
  - Remove the waitgroup, instead close the channel from `copyPTYToStdout`
  - Fix a few wrong comments
  - Fix some weak mutex coverage
  - Make some tests in `shell` table-driven to exercise the range of interrupt signals

### Testing

### Disclosures / Credits

I did not use AI tools, but Codex is probably going to review it when I mark it ready whether I want it to or not


process/process.go
```diff
@@ -233,7 +238,14 @@ func (p *Process) Run(ctx context.Context) error {
...
 		if err := p.postStart(); err != nil {
-			p.logger.Error("[Process] postStart failed: %v", err)
+			// postStart is used on Windows to associate the process to a job
+			// object. If postStart failed, then we must assume we cannot
+			// control the process or any children via the job object, so we
+			// should bail ASAP.
+			if err := p.command.Process.Kill(); err != nil {
+				p.logger.Error("[Process] failed to kill process after postStart failed: %v", err)
+			}
+			return fmt.Errorf("command post-start: %w", err)
```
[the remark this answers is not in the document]
DrJosh9000: > AssignProcessToJobObject [...] can fail in restricted job/container setups

I can't find anything to substantiate this.

agent/job_runner.go
```diff
@@ -340,6 +340,15 @@ func NewJobRunner(ctx context.Context, l logger.Logger, apiClient *api.Client, c
...
 		}
 
+		// We don't SIGKILL the bootstrap as a cancel signal, because that
+		// prevents post-checkout/command hooks running. (Change the signal
+		// grace period instead.) But the user may want the bootstrap to SIGKILL
+		// the command that it runs as an "interrupt".
+		cancelSignal := conf.CancelSignal
+		if cancelSignal == process.SIGKILL {
+			cancelSignal = process.SIGTERM
+		}
```
zhming0: The comment does not seem to explain why we reset `cancelSignal` to `SIGTERM` when `conf.CancelSignal` is `process.SIGKILL`?

The comment seem to suggest that should allow SIGKILL setting to be propagated to child process?

zhming0: After reading a bit more context, I think I get the gist but still feeling it's going to be a very hard to understand piece of logic for future travellers.

docs/images/agent-start.svg
zhming0, line 1: I played spot the diff for a bit. Can't find where things changed 😬

internal/shell/shell_test.go
```diff
@@ -134,81 +133,177 @@ func TestRunWithStdin(t *testing.T) {
...
+			started := make(chan struct{})
 
-	if err := sh.Command(sleepCmd.Path).Run(ctx); !shell.IsExitSignaled(err) {
-		t.Errorf("sh.Run(ctx, sleep) error = %v, want shell.IsExitSignaled(err) = true", err)
+			go func() {
+				select {
+				case <-time.After(1 * time.Minute):
+					t.Error("timeout waiting for process to start")
+				case <-started:
+					cancel()
```
zhming0, line 197: As a reader who didn't understand Process mechanism very much, it wasn't very obvious to me that this is where we sent the interrupt signal to children processes.

internal/process/process.go
zhming0, line 1: I think a commit stack or PR stack might be needed to make the diff here visible, same to `signal_windows.go`?

zhming0 (commented):
I think I'd need to proceed after some commit wrangling to make the diff viewable in the core file. 

But I am also happy to generate a diff locally if it's easier, lmk 🙏🏿

agent/job_runner.go
DrJosh9000: I've changed the comment a bit, let me know what you think!

docs/images/agent-start.svg
DrJosh9000: It's from the search and replace of "agent/v3/process" for "agent/v3/internal/process", probably a link URL.

internal/process/process.go
DrJosh9000: Done

internal/shell/shell_test.go
DrJosh9000: I've added some more comments.

agent/job_runner.go
```diff
@@ -340,6 +340,17 @@ func NewJobRunner(ctx context.Context, l logger.Logger, apiClient *api.Client, c
...
+		// CancelSignal == SIGKILL means the user wants the command to be killed
+		// instead of signaled more gracefully (SIGTERM, SIGINT, etc).
+		// We don't send SIGKILL to the bootstrap itself as a cancel signal,
+		// because that would kill the bootstrap immediately, which would
+		// prevent capturing the exit status of the command, executing various
+		// pre-exit hooks, and other cleanup.
+		cancelSignal := conf.CancelSignal
+		if cancelSignal == process.SIGKILL {
+			cancelSignal = process.SIGTERM
+		}
```
zhming0, line 352: Should we comment here that the only reasonable use case for specifying SIGKILL is for non-Linux systems?

internal/shell/shell_test.go
```diff
@@ -130,76 +132,183 @@ func TestRunWithStdin(t *testing.T) {
...
+			cmdWindows: ".\\fixtures\\sleep600.bat",
+			// The default interrupt signal is SIGTERM.
+			wantSignal: syscall.SIGTERM,
+		},
+		{
+			name: "SIGINT",
+			// SIGINT does nothing different on Windows.
+			cmdWindows:      ".\\fixtures\\sleep600.bat",
+			interruptSignal: process.SIGINT,
+			wantSignal:      syscall.SIGINT,
```
zhming0: This combining with the code below makes the intent of this test block quite ambiguous. I was thinking if it does nothing different why we set `wantSignal` to be sigint? 

It might be worth to add a `wantSignalWindows` field here to make everything explicit. 

```go
			wantSignal := test.wantSignal
			if runtime.GOOS == "windows" {
				wantSignal = -1
			}
```

zhming0 (approved):
LGTM 👍🏿 . Please correct me if I understood it wrong. For Windows user, they can set cancel signal to be SIGKILL, which will let `agent start` to simulate a `SIGKILL` to its subprocesses via `windows.CloseHandle(windows.Handle(p.winJobHandle))`.

The main benefit is that now cancel signal won't get swallowed by other mechanisms in Windows. 

But what I don't get is if this solves the original issue which is about job lifecycle hooks being missed? 

Nonetheless the change itself look good, I will give it tick.

DrJosh9000:
> But what I don't get is if this solves the original issue which is about job lifecycle hooks being missed?

If a command swallows Ctrl-Break and doesn't exit, then the bootstrap keeps waiting until the signal grace period, then kills the process. At the same time, the agent is _also_ waiting for the signal grace period, and then kills the bootstrap. Since the bootstrap is what executes the job lifecycle hooks, and the post-command and pre-exit hooks must run after the command, the bootstrap has no time to execute them before being killed.

agent/job_runner.go
DrJosh9000: I don't see why we should specifically discourage SIGKILL for Linux users, it just means there's effectively no signal/cancel grace period.

zhming0:
It all make senses how. 👍🏿

PR #3820, opened 2026-04-14
A-1099: Allow K8s container start timeout to be configured

zhming0:
### Description

In k8s mode, agent lifecycle was split into multiple containers within a single pod, one container is a coordinator waiting for other containers to connect. Currently the timeout is a static 5 minutes. 

This fine for 99% of the case except then people have container image that can exceed 50GB. With some contention, the timeout can be breached. 

This change make that 5 minutes configurable. 

### Context

part of A-1099

### Testing

### Disclosures / Credits

My LLM minion


DrJosh9000 (approved):
LGTM

PR #3822, opened 2026-04-14
A-1050 and A-1060: Make checkout retry attempt configurable, increasing default retry

zhming0:
### Description

Git checkout operation sometimes are subject to the mercy of network stability, a DNS blip can last for 30s+. Our current retry strategy is static and try 3 times, once every 2s. This may not capture some transient network failure scenarios and can be frustrating. 

Git checkout errors, apart from initial pipeline setup, can arguably all be categorized as transient errors. So I reckon we should do two things: 

1. Make the retry configurable.
2. Increase the default retry so that it will retry in the span of 30s (5 attempts, with exponential backoff)

### Context

A-1050 and A-1060 

### Testing

### Disclosures / Credits

Me and a LLM donkey

### DIFF
--- a/internal/job/checkout.go
+++ b/internal/job/checkout.go
@@ -239,9 +239,15 @@ func (e *Executor) checkout(ctx context.Context) error {
 			break
 		}
 
+		maxAttempts := e.CheckoutAttempts
+		if maxAttempts <= 0 {
+			maxAttempts = 6
+		}
+
 		if err := roko.NewRetrier(
-			roko.WithMaxAttempts(3),
-			roko.WithStrategy(roko.Constant(2*time.Second)),
+			roko.WithMaxAttempts(maxAttempts),
+			roko.WithStrategy(roko.Exponential(2*time.Second, 0)),
+			roko.WithJitter(),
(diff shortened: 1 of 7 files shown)


internal/job/checkout.go
```diff
@@ -239,9 +239,15 @@ func (e *Executor) checkout(ctx context.Context) error {
 			break
 		}
 
+		maxAttempts := e.CheckoutAttempts
+		if maxAttempts <= 0 {
+			maxAttempts = 5
+		}
```
swebb, line 245: Feels a little odd to me to prevent values < 5. Does this mean you can't disable retries?

swebb (approved):
LGTM 👍 

Low risk. 90% of this PR is simply argument handling. I had one question but I don't think it's a blocker.

internal/job/checkout.go
zhming0: This is preventing values <= 0, rather than 5. 

It's mainly to prevent `roko.WithMaxAttempts(0)` from panic. 

Since the setting is about "the number of checkout attempt" so I reckon we can see 0 attempt as an invalid input and override it with sane default.

PR #3825, opened 2026-04-14
fix: prevent snapshot logic if command phase isn't included

DrJosh9000:
### Description

Follow-up to #3789: require the command phase in the current executor to be included for the snapshot logic to trigger.

### Context

Codex actually made a good point in #3789.

https://linear.app/buildkite/issue/A-1073

### Testing

### Disclosures / Credits

I did the typing manually.

### DIFF
--- a/internal/job/checkout.go
+++ b/internal/job/checkout.go
@@ -565,8 +565,17 @@ func (e *Executor) updateGitMirror(ctx context.Context, repository string) (dir 
 // It's not impossible (perhaps we need a process to age-out existing checkouts)
 // but something to think about, and when we have a good implementation enable
 // it for more cases.
+//
+// Why no snapshots if the command phase isn't included:
+// The cleanup mechanism happens when this instance of the executor tears down,
+// not when the last executor among many tears down. In a split-phase setup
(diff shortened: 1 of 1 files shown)


internal/job/checkout.go
```diff
@@ -565,8 +565,13 @@ func (e *Executor) updateGitMirror(ctx context.Context, repository string) (dir
 // It's not impossible (perhaps we need a process to age-out existing checkouts)
 // but something to think about, and when we have a good implementation enable
 // it for more cases.
+//
+// Why no snapshots if the command phase isn't included:
+// The cleanup mechanism happens when this instance of the executor tears down,
```
zhming0, line 570: There is a bit of logical jump in this comment from command phase -> cleanup, I can understand why given the context I know, but not sure if the next dev will follow. Maybe we should say that in some situation command and checkout will happen in different container etc? in such case, if checkout uses snapshot and something will happen...?

zhming0 (approved):
LGTM with a comment on doc

internal/job/checkout.go
DrJosh9000: I wrote some more words

PR #3827, opened 2026-04-17
Normalise HTTP status-based retry policies across the agent

moskyb:
### Description

Across the agent, there are a bunch of interactions with the Buildkite Agent API that all have their own, bespoke retry policies to retry certain HTTP status codes. Many of these policies are very silly, retrying permission errors and the like.

This PR attempts to normalise these policies, adding a helper in the `api` package called `BreakOnNonRetryableStatus`, which breaks the roko retrier when the HTTP response has a status code which is non-retryable, and applying this helper across all calls to retried calls to the agent API.

### Context

I was working on a thing where in my local environment i was failing meta data gets with a permission error, and had this very silly situation:
CleanShot 2026-04-17 at 14 49 30@2x

where meta data gets failing with a permission error leads to them being retried for ~5 minutes.

### Changes

Where calls to the buildkite agent api are being retried through `roko`, ensure that when those calls receive a status that's non-retryable (any non-2xx that's _not_ one of these)
```Go
var retryableStatuses = map[int]bool{
	http.StatusTooManyRequests:     true, // 429
	529:                            true, // Buildkite-specific "still waiting" (used by pipeline upload)
	http.StatusInternalServerError: true, // 500
	http.StatusBadGateway:          true, // 502
	http.StatusServiceUnavailable:  true, // 503
	http.StatusGatewayTimeout:      true, // 504
}
```

that they break immediately and don't attempt to retry.

### Testing

### DIFF
--- a/api/retryable.go
+++ b/api/retryable.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"errors"
 	"io"
 	"net"
 	"net/http"
@@ -11,7 +12,7 @@ import (
 	"github.com/buildkite/roko"
 )
 
-var retrableErrorSuffixes = []string{
+var retriableErrorSuffixes = []string{
 	syscall.ECONNREFUSED.Error(),
 	syscall.ECONNRESET.Error(),
 	syscall.ETIMEDOUT.Error(),
@@ -37,22 +38,16 @@ func IsRetryableStatus(r *Response) bool {
 // Looks at a bunch of connection related errors, and returns true if the error
 // matches one of them.
 func IsRetryableError(err error) bool {
-	if neterr, ok := err.(net.Error); ok {
-		if neterr.Temporary() {
+	var neterr net.Error
+	if errors.As(err, &neterr) {
+		if neterr.Timeout() {
 			return true
 		}
 	}
 
-	if neterr, ok := err.(net.Error); ok && neterr.Timeout() {
-		return true
-	}
-
-	if urlerr, ok := err.(*url.Error); ok {
+	var urlerr *url.Error
+	if errors.As(err, &urlerr) {
 		if strings.Contains(urlerr.Error(), "use of closed network connection") {
-			return true
-		}
-
-		if neturlerr, ok := urlerr.Err.(net.Error); ok && neturlerr.Timeout() {
 			return true
 		}
 	}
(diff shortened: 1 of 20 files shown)


core/client.go
```diff
@@ -230,24 +230,9 @@ func (c *Client) FinishJob(ctx context.Context, job *api.Job, finishedAt time.Ti
...
-			// process).
-			// The API may also return a 401 when job tokens
-			// are enabled.
-			// In either case, we don't want to keep trying
-			// to finish the job forever so we'll just bail out and
-			// go find some more work to do.
-			if response != nil && (response.StatusCode == 422 || response.StatusCode == 401) {
-				c.Logger.Warn("Buildkite rejected the call to finish the job (%s)", err)
-				retrier.Break()
-				return err
```
zhming0, line 248: I think these are good comments, it shouldn't be be removed. ~Also the behavior seems to change now 422 will no longer break the retry loop.~

zhming0 (dismissed):

api/retryable.go
```diff
@@ -21,50 +24,78 @@ var retrableErrorSuffixes = []string{
 
 var retryableStatuses = map[int]bool{
 	http.StatusTooManyRequests:     true, // 429
+	529:                            true, // Buildkite-specific "still waiting" (used by pipeline upload)
```
zhming0, line 27: TIL!

api/retryable.go
```diff
@@ -21,50 +24,78 @@ var retrableErrorSuffixes = []string{
...
 	http.StatusGatewayTimeout:      true, // 504
 }
 
 // IsRetryableStatus returns true if the response's StatusCode is one that we should retry.
+// Success statuses (2xx) are not considered retryable — they are not errors.
 func IsRetryableStatus(r *Response) bool {
+	if r.StatusCode >= 200 && r.StatusCode <= 299 {
+		return false
+	}
 	return retryableStatuses[r.StatusCode]
```
zhming0, line 40: I wonder if we should just make all 5xx retriable?

api/retryable.go
moskyb: _sigh_

core/client.go
moskyb: done!

api/retryable.go
moskyb: done.

DrJosh9000 (approved):
sick ✅

zhming0 (approved):

PR #3830, opened 2026-04-17
Removal manual gates from deployment process

zhming0:
Remove the `block: "beta?"` and `block: "stable?"` manual gates from the release pipeline, so the release flow proceeds automatically: edge → beta → stable.

## Why

We never use these manual gates, every release we just click through them. The pipeline already has sufficient safeguards:

* Release steps only run on `main` or `*-*-stable` branches
* If the git tag for the version already exists, beta/stable release is skipped entirely
* A VERSION file change is required to trigger a new release

### DIFF
--- a/.buildkite/steps/upload-release-steps.sh
+++ b/.buildkite/steps/upload-release-steps.sh
@@ -39,11 +39,9 @@ trigger_step() {
 YAML
 }
 
-block_step() {
-  local label="$1"
+wait_step() {
   cat <<YAML
   - wait
-  - block: "${label}"
 YAML
 }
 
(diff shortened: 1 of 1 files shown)


.buildkite/steps/upload-release-steps.sh
DrJosh9000, line 67: If there's no block (or some other gap in time) between beta and stable, then there's no point in having both channels - 

Since v4 is coming (Real Soon Now™), and we might want to beta-test it, perhaps we should make this one or the other?
 
```shell
  wait_step

  if [[ "$agent_version" =~ (beta|rc) ]] ; then
    trigger_step \
      "beta ${agent_version}" \
      "agent-release-beta"
  else
    trigger_step \
      "stable ${agent_version}" \
      "agent-release-stable"
  fi
```

zhming0: yes 👍🏿

DrJosh9000 (approved):

PR #3832, opened 2026-04-19
feat: migrate to github.com/klauspost/compress/gzip for optimized deflate

wolfeidau:
### Description

Switch to a more optimized version of deflate in the log chunk uploader.

### Context

From the readme of this library:

> Typical speed is about 2x of the standard library packages.

This package is already an indirect dependency of the project as it is also used by connectrpc.

### Changes

This change switches to a more optimized version of deflate in the log chunk uploader.

### Testing

### Disclosures / Credits

I did not use AI tools at all

### DIFF
--- a/api/chunks.go
+++ b/api/chunks.go
@@ -2,9 +2,10 @@ package api
 
 import (
 	"bytes"
-	"compress/gzip"
 	"context"
 	"fmt"
+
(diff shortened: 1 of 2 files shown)


zhming0 (approved):
Drop in replacement, so nice 👍🏿

PR #3833, opened 2026-04-19
feat: add metrics to record histogram of chunk sizes and compressed size

wolfeidau:
### Description

Add some new metrics which cover log chunk compression and upload sizes.

Screenshot 2026-04-19 at 8 28 36 pm

### Context

These metrics are helpful to under the range of chunk sizes, and how the current compression algorythm is performing.

### Changes

Adds two new metrics.

* compressed_bytes_uploaded_total
* chunk_size_bytes

### Testing

### Disclosures / Credits

I did not use AI tools at all

### DIFF
--- a/agent/metrics.go
+++ b/agent/metrics.go
@@ -102,4 +102,17 @@ var (
 		// Log chunk upload can be retried for a while.
 		Buckets: prometheus.ExponentialBuckets(0.015625, 2, 16),
 	})
+	logCompressedBytesUploaded = promauto.NewCounter(prometheus.CounterOpts{
+		Namespace: metricsNamespace,
+		Subsystem: "logs",
+		Name:      "compressed_bytes_uploaded_total",
+		Help:      "Count of compressed log bytes sent over the wire",
+	})
+	logChunkSizeBytes = promauto.NewHistogram(prometheus.HistogramOpts{
+		Namespace: metricsNamespace,
+		Subsystem: "logs",
+		Name:      "chunk_size_bytes",
+		Help:      "Distribution of log chunk sizes in bytes at upload time",
+		Buckets:   prometheus.ExponentialBuckets(256, 4, 10), // 256B to ~64MB
+	})
 )
--- a/api/chunks.go
+++ b/api/chunks.go
@@ -9,10 +9,11 @@ import (
 
 // Chunk represents a Buildkite Agent API Chunk
 type Chunk struct {
-	Data     []byte
-	Sequence uint64
-	Offset   uint64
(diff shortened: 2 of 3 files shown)


agent/metrics.go
```diff
@@ -102,4 +102,17 @@ var (
 		// Log chunk upload can be retried for a while.
 		Buckets: prometheus.ExponentialBuckets(0.015625, 2, 16),
 	})
+	logCompressedBytesUploaded = promauto.NewCounter(prometheus.CounterOpts{
+		Namespace: metricsNamespace,
+		Subsystem: "logs",
+		Name:      "compressed_bytes_uploaded_total",
+		Help:      "Count of compressed log bytes sent over the wire",
+	})
```
zhming0, line 110: Would you want to consider a `compressed_chunk_size_bytes` histogram metric instead? I presume the goal of these two metrics is to construct a compression ratio across percentile? 

The histogram metric will cover the throughput signal for free.

api/chunks.go
```diff
@@ -27,6 +28,7 @@ func (c *Client) UploadChunk(ctx context.Context, jobId string, chunk *Chunk) (*
 	if err := gzipper.Close(); err != nil {
 		return nil, err
 	}
+	chunk.CompressedBytes = uint64(body.Len())
```
zhming0, line 31: do we want to consider letting UploadChunk to return something like: 

```golang
type UploadChunkResult struct {
    Response        *Response
    CompressedBytes int
}
```

so we won't have to mutate the input?

zhming0 (commented):
I have 2 questions 🙏🏿

agent/metrics.go
wolfeidau: @zhming0 I am using a counter like `bytes_uploaded_total` this way I can divide the two using an expression in prometheus.

```
rate(buildkite_agent_logs_bytes_uploaded_total[1m]) / rate(buildkite_agent_logs_compressed_bytes_uploaded_total[1m])
```

api/chunks.go
wolfeidau: This is already done above in [agent/job_runner.go‎](https://github.com/buildkite/agent/pull/3833/changes#diff-44911830fc5e3093b48306afe641877512f466b218317c812e12a5520edb7a58) `logBytesUploaded.Add(float64(chunk.Size))` if you trace this call down the only thing which is propagated back is the error and it passes through a retrier which seems messy to refactor.

The `Response` never comes back to this function.

zhming0 (approved):
LGTM ✅

PR #3834, opened 2026-04-20
Remove deprecated CLI flags

moskyb:
### Description

In the agent, we have a bunch of flags that have been deprecated (and emitting warnings whenever they're used) for a good long while. We're doing v4! Let's throw em out the window.

### Changes

Removed the following flags:
- from `agent_start.go`: 
  - `--meta-data`, 
  - `--meta-data-ec2`, 
  - `--meta-data-ec2-tags`, 
  - `--meta-data-gcp`, 
  - `--no-automatic-ssh-fingerprint-verification`, 
  - `--tags-from-ec2`, 
  - `--tags-from-gcp`, 
  - `--disconnect-after-job-timeout`, 
  - `--kubernetes-log-collection-grace-period`,
- from `artifact_upload.go`: `--follow-symlinks`
- from `global.go`: `--kubernetes-log-collection-grace-period`

### Testing

### Disclosures / Credits

Ctrl-F was instrumental to me in this work.

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -212,16 +212,6 @@ type AgentStartConfig struct {
 	Token     string `cli:"token" validate:"required"`
 	Endpoint  string `cli:"endpoint" validate:"required"`
 	NoHTTP2   bool   `cli:"no-http2"`
-	// Deprecated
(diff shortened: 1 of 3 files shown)


DrJosh9000 (approved):
Heck yeah ✅

PR #3836, opened 2026-04-20
feat: add tunable log streaming interval to reduce chunk upload API calls

wolfeidau:
### Description

Adds a `--log-streaming-interval` flag (`BUILDKITE_LOG_STREAMING_INTERVAL`, default 3s) that controls how long the log streamer accumulates output before uploading a chunk. Larger intervals mean fewer, bigger chunks,
reducing API calls, improving gzip compression ratios, and lowering egress costs for agents behind proxies.

The first chunk of each job is always uploaded within 1s so the job log has visible output promptly. Set to 0 to restore the original per-poll-tick behaviour.

### Testing

### DIFF
--- a/agent/agent_configuration.go
+++ b/agent/agent_configuration.go
@@ -77,4 +77,10 @@ type AgentConfiguration struct {
 	AllowMultipartArtifactUpload bool
 
 	PingMode string
+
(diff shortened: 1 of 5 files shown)


DrJosh9000:
Hey @wolfeidau just wondering is this still needed, or is it ready for review?

PR #3837, opened 2026-04-20
fix: set PTY raw mode before process start

lox:
## Summary
- apply PTY raw mode before starting the child process, so early output cannot be transformed by the default line discipline
- keep PTY sizing and controlling-terminal setup in the PTY helper
- add regression coverage for output written before raw mode used to be applied

## Problem
We were seeing a startup race in the `pty-raw` path.

The child process was being started before the PTY master was put into raw mode, so output written immediately on startup could still be processed with the default PTY line discipline. When that happened, `\n` was rewritten to `\r\n`, which made `TestProcessOutputPTY_PTYRawExperiment` fail intermittently.

## Testing
- go test ./process
- go test ./process -run TestProcessOutputPTY_PTYRawExperimentWritesBeforeRawMode -count=1
- go test ./process -failfast -count=100 -run TestProcessOutputPTY_PTYRawExperiment

### DIFF
--- a/process/pty.go
+++ b/process/pty.go
@@ -16,7 +16,7 @@ func StartPTY(c *exec.Cmd, raw bool) (*os.File, error) {
 	if err != nil {
 		return nil, err
 	}
-	defer func() { _ = tty.Close() }()
+	defer tty.Close() //nolint:errcheck // Best-effort cleanup.
 
 	if err := pty.Setsize(ptmx, &pty.Winsize{
 		Rows: 100,
@@ -24,13 +24,13 @@ func StartPTY(c *exec.Cmd, raw bool) (*os.File, error) {
 		X:    0, // unused
 		Y:    0, // unused
 	}); err != nil {
-		_ = ptmx.Close()
+		ptmx.Close() //nolint:errcheck // Best-effort cleanup.
 		return nil, err
 	}
 
 	if raw {
 		if _, err := term.MakeRaw(int(ptmx.Fd())); err != nil {
-			_ = ptmx.Close()
+			ptmx.Close() //nolint:errcheck // Best-effort cleanup.
 			return nil, err
 		}
 	}
@@ -52,7 +52,7 @@ func StartPTY(c *exec.Cmd, raw bool) (*os.File, error) {
 	c.SysProcAttr.Setctty = true
 
 	if err := c.Start(); err != nil {
-		_ = ptmx.Close()
+		ptmx.Close() //nolint:errcheck // Best-effort cleanup.
 		return nil, err
 	}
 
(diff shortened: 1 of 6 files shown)


process/pty.go
```diff
@@ -5,15 +5,56 @@ package process
...
 )
 
-func StartPTY(c *exec.Cmd) (*os.File, error) {
-	return pty.StartWithSize(c, &pty.Winsize{
+func StartPTY(c *exec.Cmd, raw bool) (*os.File, error) {
+	ptmx, tty, err := pty.Open()
+	if err != nil {
+		return nil, err
+	}
+	defer func() { _ = tty.Close() }()
```
DrJosh9000: [style suggestion] I would rather have a lint-disable comment that explains why an error is ignored, rather than continue with the `_ =` pattern, even if that comment is "best-effort cleanup" over and over:

```suggestion
	defer tty.Close() //nolint:errcheck // Best-effort cleanup
```

process/pty.go
```diff
@@ -5,15 +5,56 @@ package process
...
+	if c.SysProcAttr == nil {
+		c.SysProcAttr = &syscall.SysProcAttr{}
+	}
+	c.SysProcAttr.Setsid = true
+	c.SysProcAttr.Setctty = true
+
+	if err := c.Start(); err != nil {
+		_ = ptmx.Close()
+		return nil, err
+	}
```
DrJosh9000: [As above]

```suggestion
		ptmx.Close() //nolint:errcheck // Best-effort cleanup
		return nil, err
	}

	if raw {
		if _, err := term.MakeRaw(int(ptmx.Fd())); err != nil {
			ptmx.Close() //nolint:errcheck // Best-effort cleanup
			return nil, err
		}
	}

	if c.Stdout == nil {
		c.Stdout = tty
	}
	if c.Stderr == nil {
		c.Stderr = tty
	}
	if c.Stdin == nil {
		c.Stdin = tty
	}

	if c.SysProcAttr == nil {
		c.SysProcAttr = &syscall.SysProcAttr{}
	}
	c.SysProcAttr.Setsid = true
	c.SysProcAttr.Setctty = true

	if err := c.Start(); err != nil {
		ptmx.Close() //nolint:errcheck // Best-effort cleanup
		return nil, err
	}
```

DrJosh9000 (approved):
LGTM with one style comment split over two suggestions

process/pty.go
lox: Fixed in 713a1226: replaced the `_ =` cleanup ignores with explicit `//nolint:errcheck` best-effort cleanup comments at each PTY close site in this path.

process/pty.go
lox: Fixed in 713a1226: replaced the `_ =` cleanup ignores with explicit `//nolint:errcheck` best-effort cleanup comments.

PR #3838, opened 2026-04-20
fix: More consistent protected env vars

DrJosh9000:
### Description

Make the whole situation with env var protection clearer.

- Apply normalisation when checking protection
- Allow some env vars that are mutable from hooks and plugins to be modifiable from the Job API
- Prevent mutating some env vars via hook wrapper

### Context

https://linear.app/buildkite/issue/A-991, and others

### Testing

### Disclosures / Credits

The wet noodles of brain tissue curling around inside my skull


env/environment.go
```diff
@@ -349,26 +389,39 @@ var ProtectedEnv = map[string]struct{}{
...
 	"BUILDKITE_CONTAINER_COUNT":          {},
-	"BUILDKITE_GIT_CLEAN_FLAGS":          {},
-	"BUILDKITE_GIT_CLONE_FLAGS":          {},
-	"BUILDKITE_GIT_CLONE_MIRROR_FLAGS":   {},
-	"BUILDKITE_GIT_FETCH_FLAGS":          {},
+	"BUILDKITE_GIT_CHECKOUT_FLAGS":       {allowWriteFromJob: true},
+	"BUILDKITE_GIT_CLEAN_FLAGS":          {allowWriteFromJob: true},
+	"BUILDKITE_GIT_CLONE_FLAGS":          {allowWriteFromJob: true},
+	"BUILDKITE_GIT_CLONE_MIRROR_FLAGS":   {allowWriteFromJob: true},
+	"BUILDKITE_GIT_FETCH_FLAGS":          {allowWriteFromJob: true},
```
zhming0, line 417: is this change a feature? it feels like a big expansion to plugin capabilities.

env/environment.go
```diff
@@ -337,9 +356,30 @@ func (diff *Diff) Empty() bool {
...
+	// reconfigures the executor
+	// (see ReadFromEnvironment in internal/job/config.go).
+	allowWriteFromJob bool
+}
+
+// protectedEnv contains environment variables that can only be set by agent
+// configuration, or in some cases, from within the job.
+//
+// These variables cannot be overwritten by job-level environment variables or
+// secrets, but some may still be set in hooks or plugins.
```
zhming0, line 371: Just for my understanding, why would we disallow an env to be mutated by job env var but instead allow it to be mutated by hook and plugin?

env/environment.go
```diff
@@ -337,9 +356,30 @@ func (diff *Diff) Empty() bool {
...
+//
+// The actual enforcement of protected env within the agent level (overriding
+// job-level env vars based on agent configuration) happens implicitly rather
+// than relying on this map - see createEnvironment in agent/job_runner.go.
+// But this map is used in a couple of other places to prevent env var changes.
+//
+// Ensure allowWriteFromJob is enabled for the vars present in the env tags in
+// ExecutorConfig in internal/job/config.go.
+// Note that powerful-looking vars that are not in this map may not be present
+// if they are set/overridden on the backend.
```
zhming0: 👍🏿 TLDR, this is not the source of truth?

jobapi/env.go
```diff
@@ -141,7 +141,9 @@ func (s *Server) deleteEnv(w http.ResponseWriter, r *http.Request) {
 func checkProtected(candidates []string) []string {
 	protected := make([]string, 0, len(candidates))
 	for _, c := range candidates {
-		if _, ok := env.ProtectedEnv[c]; ok {
+		// The Job API is only accessible from within the job, so allow writes
+		// to vars that allow write from within job.
+		if env.IsProtectedFromWithinJob(c) {
```
zhming0, line 146: Is this change "Prevent mutating some env vars via hook wrapper"? 

If I am not mistaken, we didn't add any new protected env var  in this PR?

env/environment.go
zhming0: Might it be something worth writing down as well?

zhming0 (commented):
It looks mostly good, I can see the rationale. I mainly have questions to prevent me misunderstanding this

env/environment.go
DrJosh9000: Plugins are already able to set these, e.g. https://github.com/buildkite-plugins/git-clean-buildkite-plugin/blob/master/hooks/pre-checkout

zhming0: ah then is this change re envs that are created as part of Pipeline yaml??

env/environment.go
DrJosh9000: Good question. The main question I ask is: if someone has rights to start a build and provide env vars, could an env var they set be able to bypass other restrictions like no-command-eval? And would the env var be a reasonable way for a plugin to set config. Git flags are an example. There are some shell injections possible via Git flag, so they should be rejected from the build creation dialog, but plugins should be able to adjust Git flags prior to checkout. (Plugins can also implement the checkout hook, bypassing the default checkout process entirely.)

env/environment.go
DrJosh9000: There are many places where env vars are set, overridden, merged, combined.

env/environment.go
DrJosh9000: All these env vars are ignored from the pipeline YAML or build creation dialog - that was already the case. The change here is to make it clearer that some of them can be set from plugins, hooks, or the command.

env/environment.go
DrJosh9000: I've written some of this into the comment.

jobapi/env.go
DrJosh9000: The lines where you've commented, the change is to allow _more_ env vars to be changed via the Job API. Previously, the git flags vars would have been rejected here, even though they can be set via the hook wrapper.

The hook wrapper and Job API should have consistent logic, since they're both doing the same thing (allowing a script or command to change env vars), so the same check is now added to the hook wrapper.

env/environment.go
zhming0: Interesting, I thought someone who has rights to start a build are often the one that has code commit permission, so this permission mechanism seems quite narrow.
 
But thanks for clarifying it!

env/environment.go
```diff
@@ -349,26 +410,39 @@ var ProtectedEnv = map[string]struct{}{
...
 	"BUILDKITE_CONTAINER_COUNT":          {},
-	"BUILDKITE_GIT_CLEAN_FLAGS":          {},
-	"BUILDKITE_GIT_CLONE_FLAGS":          {},
-	"BUILDKITE_GIT_CLONE_MIRROR_FLAGS":   {},
-	"BUILDKITE_GIT_FETCH_FLAGS":          {},
+	"BUILDKITE_GIT_CHECKOUT_FLAGS":       {allowWriteFromJob: true},
+	"BUILDKITE_GIT_CLEAN_FLAGS":          {allowWriteFromJob: true},
+	"BUILDKITE_GIT_CLONE_FLAGS":          {allowWriteFromJob: true},
+	"BUILDKITE_GIT_CLONE_MIRROR_FLAGS":   {allowWriteFromJob: true},
+	"BUILDKITE_GIT_FETCH_FLAGS":          {allowWriteFromJob: true},
```
zhming0: Without reading more context, the `allowWriteFromJob` seems to be saying allow Job environment to modify these env vars. 

But the comments above say: 
> But we don't want the job environment to be able to set them, because git is riddled with shell injections

env/environment.go
zhming0: > All these env vars are ignored from the pipeline YAML or build creation dialog

TIL! I thought I got it working via Pipeline yaml before. 

I have doubt but all good for now 👍🏿

jobapi/env.go
zhming0: 👍🏿 yeah, my question was mainly about where is the code change for this item in PR description?

> Prevent mutating some env vars via hook wrapper

zhming0 (approved):
LGTM though I think it might be worth to re-clarify some names.

jobapi/env.go
DrJosh9000: The change in executor.go (https://github.com/buildkite/agent/blob/0aef51de69625f6630204383d6ee6ae22e9fa034/internal/job/executor.go#L631-L644)

process/BUILD.bazel
DrJosh9000, line 1: `bazel run :gazelle` generated this

PR #3840, opened 2026-04-20
chore: add cleanroom config

lox:
## Draft

This PR is stacked on top of #3839 and should not be merged until that PR lands first.

## Summary

Add a repo-level `cleanroom.yaml` based on the existing Buildkite Agent example in the `cleanroom` repo.

This gives the agent repo a checked-in cleanroom policy for running inside a sandbox with deny-by-default egress, explicit `mise` bootstrap, and cached Go module dependency warmup.

## Changes

- add `cleanroom.yaml` at the repo root
- copy the policy from `../cleanroom/examples/buildkite-agent/cleanroom.yaml`
- update the dependency cache key to use `mise.toml` instead of `.mise.toml`
- enable default repo bootstrap so `cleanroom exec -- mise ...` works from this checkout

## Why

This keeps the cleanroom policy alongside the repo it is meant to run, instead of only existing as an external example.

It also depends on the earlier `mise.toml` rename in #3839, since the dependency-stage cache key now points at `mise.toml`.

## Testing

- `cleanroom policy validate`
- `cleanroom exec -- mise x -- go version`

### DIFF
--- a/cleanroom.yaml
+++ b/cleanroom.yaml
@@ -1,0 +1,42 @@
+version: 1
+repository:
+  enabled: false
+backends:
+  darwin-vz:
+    memory_mib: 12288
+sandbox:
+  image:
+    ref: ghcr.io/buildkite/cleanroom-base/debian@sha256:28c3f638fabe1ed780f87b82cfb0c6dda2549c86b9e4edbe519e8250243411c5
+  dependencies:
+    command: |
+      mise settings ruby.compile=false
+      mise install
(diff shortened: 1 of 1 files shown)


DrJosh9000:
Hey @lox just wondering is this ready for review?

lox:
This has gone stale and I don’t think it’s worth carrying as an open draft.

If we still want repo-level cleanroom config here later, I’d rather recut it cleanly from current `main`.

PR #3842, opened 2026-04-20
perf: speed up job integration tests

lox:
## Summary
- reuse a prepared integration git repo template instead of rebuilding the seed repo for every tester
- batch local hook fixture commits in the integration harness
- make hook wrappers respect `BUILDKITE_OVERRIDE_SELF` for the ExecutorTester path so `env dump` uses the existing bintest mock instead of self-execing the full test binary
- add regression and perf coverage around the harness changes

## Results
- `mise run test -- -count=1 ./...`: about `93.98s` -> `80.58s` on macOS
- `internal/job/integration` in the full suite: about `1m28.5s` -> `1m16.5s`
- `mise exec go@1.25.9 -- go test ./internal/job/integration -count=1`: about `77.2s` -> `70.6s`

## Verification
- `mise exec go@1.25.9 -- go test ./internal/job/hook ./internal/job/integration -count=1`
- `mise run test -- -count=1 ./...`

### DIFF
--- a/internal/job/hook/wrapper.go
+++ b/internal/job/hook/wrapper.go
@@ -34,6 +34,7 @@ const (
 	hookExitStatusEnv = "BUILDKITE_HOOK_EXIT_STATUS"
 	hookWorkingDirEnv = "BUILDKITE_HOOK_WORKING_DIR"
 	hookWrapperDir    = "buildkite-agent-hook-wrapper"
+	testJobID         = "1111-1111-1111-1111"
 
 	batchWrapper = `@echo off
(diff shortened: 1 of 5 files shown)


jamiemonserrate (approved):
Nice, this looks good to me overall.

One small thing I noticed while reviewing: the new override-self test only covers the positive path. I have a local commit that turns it into a table test and also covers the fallback cases:
- BUILDKITE_OVERRIDE_SELF is ignored without the ExecutorTester job ID
- the real executable is used when the override is unset

Happy for me to push that commit onto this PR, or would you prefer to fold it in yourself?

PR #3856, opened 2026-04-27
fix: use single-quoted shell escaping for --format env output

ss1909:
### Description

Extracts the secretGet logic from the CLI action into a testable function, and adds a unit test suite for it. Also fixes the env format output to use single-quoted shell escaping (via shellQuote) instead of Go's %q double-quoting, so values are safe to source in shell without unintended interpolation.

### Context

The secretGet logic previously lived inside the cli.Command action, making it impossible to unit test without going through the CLI framework. The env format also used %q which produces Go-style double-quoted strings — not correct shell quoting for use with source or declare -x.

### Changes

- Extracted secretGet into a standalone function accepting 

- Fixed env format to use shellQuote (single quotes with '\'' escaping) instead of %q

- Added clicommand/secret_get_test.go 

### Testing

### Disclosures / Credits

Claude Code wrote some of the unit tests, then I changed most of it 😄

### DIFF
--- a/clicommand/secret_get.go
+++ b/clicommand/secret_get.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"maps"
 	"slices"
 	"strings"
@@ -12,6 +13,7 @@ import (
 	"github.com/buildkite/agent/v3/api"
 	"github.com/buildkite/agent/v3/internal/secrets"
 	"github.com/buildkite/agent/v3/jobapi"
+	"github.com/buildkite/agent/v3/logger"
 	"github.com/urfave/cli"
 )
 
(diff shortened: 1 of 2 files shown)


clicommand/secret_get_test.go
```diff
@@ -0,0 +1,130 @@
...
+		assert.NoError(t, json.Unmarshal(out.Bytes(), &result))
+		assert.Equal(t, map[string]string{"deploy_key": "supersecret"}, result)
+	})
+
+	t.Run("env format - sorts keys, uppercases and shellQuotes", func(t *testing.T) {
+		t.Parallel()
+		server := newSecretGetTestServer(t, map[string]string{
+			"deploy_key":     "secret1",
+			"github_api_key": "secret2",
+		})
```
zhming0, line 108: Should we add some sample data that contains single quote to test the escaping mechanism? also for things like `\n` etc..

clicommand/secret_get_test.go
zhming0, line 1: The agents team prefer not using `assert` although we don't have a linter for this. [doc](https://www.notion.so/buildkite/Assert-is-bad-33db8dbc2c8980ca8421e9ad47e0dddc?source=copy_link)

clicommand/secret_get.go
```diff
@@ -84,67 +86,74 @@ Examples:
...
-			_, _ = fmt.Fprintln(c.App.Writer, secrets[0].Value)
-			return nil
-
-		case cfg.Format == "json" || (cfg.Format == "default" && len(cfg.Keys) > 1):
-			if err := json.NewEncoder(c.App.Writer).Encode(secretsMap); err != nil {
-				return fmt.Errorf("failed to write JSON response: %w", err)
-			}
+	case cfg.Format == "env":
+		for _, key := range slices.Sorted(maps.Keys(secretsMap)) {
+			_, _ = fmt.Fprintf(w, "%s=%s\n", strings.ToUpper(key), shellQuote(secretsMap[key]))
```
zhming0, line 146: [blocking] It seems a "\n" in the secret value would break this output? I think it's worth a test case :)

zhming0 (changes requested):
There might be a blocker re `\n` handling but otherwise I think this PR is in the right direction 👍🏿

clicommand/secret_get_test.go
```diff
@@ -0,0 +1,144 @@
...
+		defer server.Close()
+
+		var out bytes.Buffer
+		err := secretGet(context.Background(), baseSecretGetConfig(server.URL, []string{"deploy_key", "github_api_key"}, "env"), &out, logger.NewBuffer())
+		require.NoError(t, err)
+		require.Equal(t, "DEPLOY_KEY='secret1'\nGITHUB_API_KEY='secret2'\n", out.String())
+	})
+
+	t.Run("env format - secret contains newline and single quote, returns valid shell syntax", func(t *testing.T) {
```
ss1909, line 117: thank you @zhming0, I've added this test case here to cover the newline and single quotes scenario

clicommand/secret_get.go
ss1909: Covered in the new test case! :)

zhming0 (approved):
LGTM, I hallucinated about the newline issue 😅

PR #3858, opened 2026-04-27
Remove and prevent assert libraries

DrJosh9000:
### Description

Remove uses of github.com/stretchr/testify and gotest.tools/v3. Add a command to the lint step to prevent new uses being added.

### Context

https://www.notion.so/buildkite/Assert-is-bad-33db8dbc2c8980ca8421e9ad47e0dddc?v=217b8dbc2c898063af47000cce90e7c4&source=copy_link

### Testing

### Disclosures / Credits

I wrote a tool to do the bulk of the changes automatically. Some still needed minor tweaking (e.g. because `cmp.Diff` doesn't like comparing unexported fields, but `testify/assert` uses `reflect.DeepEqual`).


moskyb:
👀👀👀👀

zhming0:
I thought our axe is mainly targeting `assert` but leave a mercy for `require`? 👀 🥹 

But i won't block this 😁

DrJosh9000:
> I thought our axe is mainly targeting assert but leave a mercy for require? 👀 

`require` _is_ `assert`, the difference being it fails tests immediately rather than continuing (but still failing). `require` receives all the same readability criticisms as `assert`, but also it discourages ["keep going"](https://google.github.io/styleguide/go/decisions.html#keep-going) style.

moskyb (approved):
you have my sword.

PR #3859, opened 2026-04-27
Add no-checkout-override flag

scadu:
### Description

This PR adds `no-checkout-override`, a flag that lets agent config lock checkout-related env vars.

Without the flag, checkout-scoped vars stay mutable and agent checkout config acts as a default. 
With the flag on, those vars become locked and agent checkout config acts as a hard override.

I looked at a couple of other approaches, like patching each surface separately or making checkout vars always protected, but both felt worse. One would have kept the behavior inconsistent, and the other would have changed the unlocked case.

Other approaches considered: adding `checkoutScoped` bool to the `protectedEnv` map, in addition to existing `mutableFromWithinJob` bool.

PR into `feat/git-checkout-features` which is meant to be used as a feature branch for all the checkout improvements.

### Context

https://linear.app/buildkite/issue/SUP-6423

### Changes

- add the `no-checkout-override` flag
- split checkout-scoped env vars out from the always-protected env set
- only lock checkout-scoped vars when `BUILDKITE_NO_CHECKOUT_OVERRIDE=true`
- apply that rule to:
  - hook and plugin env mutation
  - Job API env updates and deletes
  - pipeline secrets
  - bootstrap / job-runner env merging
- add regression coverage for:
  - bootstrap-boundary env behavior
  - hook behavior
  - secrets behavior

### Testing

### DIFF
--- a/env/protected.go
+++ b/env/protected.go
@@ -63,6 +63,11 @@ var protectedEnv = map[string]protection{
 	"BUILDKITE_SHELL":                       {},
 }
 
+// checkoutOverrideScope contains checkout-related vars that remain mutable in
+// hooks, plugins, Job API, and secrets by default so jobs can tailor checkout
+// behavior. When checkout override is enabled, those same vars become locked so
+// agent checkout config wins and git flags cannot be used to undermine
+// no-command-eval.
 var checkoutOverrideScope = map[string]struct{}{
 	"BUILDKITE_GIT_CHECKOUT_FLAGS":              {},
 	"BUILDKITE_GIT_CLEAN_FLAGS":                 {},
--- a/env/protected.go
+++ b/env/protected.go
@@ -22,16 +22,6 @@ type protection struct {
 // disables plugins, but even if it is changed by a hook, the agent doesn't
 // reconfigure no-command-eval based on any changes.)
 //
-// Git flags are an example of vars that are primarily driven by agent config,
-// but we want to allow plugins to set git flags to alter the default checkout
-// process (see for example the git-clean plugin). Doing this deliberately
-// reconfigures the executor (see ReadFromEnvironment and config struct tags in
-// internal/job/config.go).
-// But we don't want the job env from the backend to be able to set them,
-// because git is riddled with shell injections, and someone with privileges to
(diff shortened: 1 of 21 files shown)


env/protected.go
```diff
@@ -45,41 +35,47 @@ type protection struct {
...
+	"BUILDKITE_NO_CHECKOUT_OVERRIDE":        {},
+	"BUILDKITE_PLUGINS_ALWAYS_CLONE_FRESH":  {mutableFromWithinJob: true},
+	"BUILDKITE_PLUGINS_ENABLED":             {},
+	"BUILDKITE_PLUGINS_PATH":                {},
+	"BUILDKITE_REFSPEC":                     {mutableFromWithinJob: true},
+	"BUILDKITE_REPO":                        {mutableFromWithinJob: true},
+	"BUILDKITE_SHELL":                       {},
+}
+
+var checkoutOverrideScope = map[string]struct{}{
```
DrJosh9000, line 71: There ought to be a comment explaining what checkout override scope is about. I think this is a reasonable place because it is where future maintainers might have to add or remove new vars.

DrJosh9000 (commented):
I think it is necessary for `no-command-eval` to imply `no-checkout-override`, because of the previously-mentioned shell injections possible via Git flags.

I am also starting to think that maybe `no-checkout-config-override` / `BUILDKITE_NO_CHECKOUT_CONFIG_OVERRIDE` may be slightly clearer, but it's still not perfect (which config is not overriding which other config?)

env/protected.go
scadu: @DrJosh9000 it's certainly a bit rough, hence we'd first merge it to a feature branch and continue shaping it. Will add a comment regarding checkout override scope.

> I am also starting to think that maybe no-checkout-config-override / BUILDKITE_NO_CHECKOUT_CONFIG_OVERRIDE may be slightly clearer [...]

Do you have anything specific in mind?

scadu: @DrJosh9000 I added the description (hopefully it's a bit clearer now). Also, there's now a mechanism to enable `no-checkout-override=true` when `no-command-eval` is enabled.

PR #3866, opened 2026-04-29
fix: Add automatic tracing to CLI subcommands

ss1909:
### Description

When `BUILDKITE_TRACING_BACKEND=opentelemetry` is configured, the bootstrap process initialises a TracerProvider and creates spans for each job phase (artifact-upload, etc.). However, subcommands like `buildkite-agent artifact upload `run as separate OS processes and never initialised their own TracerProvider, so even though `TRACEPARENT` and `BUILDKITE_TRACING_BACKEND` were present in their environment, no spans were emitted. The executor's phase spans were not visible internally.

### Context
https://linear.app/buildkite/issue/A-1168/add-automatic-tracing-to-cli-subcommands-artifact-upload-pipeline

### Changes
- internal/job/tracing.go — extracted InitOTelTracerProvider as a standalone exported function; startTracingOpenTelemetry now delegates to it

- clicommand/global.go — setupLoggerAndConfig now initialises a TracerProvider and extracts TRACEPARENT/TRACESTATE into context when running as a subprocess of a traced job

Every subcommand (`artifact upload`, `pipeline upload`, etc.) gets tracing for free.

- `ctx, span := otel.Tracer("buildkite-agent").Start(ctx, "<sub-command-name>")`
create a new span named after the sub-command-name. Because ctx already has the parent trace context extracted from `TRACEPARENT`, OTel automatically makes this span a child of the executor's span. 

### Testing

Grafana screenshots:
image 
image

### Disclosures / Credits
Implementation was carried out in consultation with Claude

### DIFF
--- a/internal/job/tracing.go
+++ b/internal/job/tracing.go
@@ -181,7 +181,7 @@ func (e *Executor) startTracingOpenTelemetry(ctx context.Context) (tracetools.Sp
 	extras, warnings := toOpenTelemetryAttributes(GenericTracingExtras(e, e.shell.Env))
 	for k, v := range warnings {
 		e.shell.Warningf("Unknown attribute type (key: %v, value: %v (%T)) passed when initialising OpenTelemetry. This is a bug, submit this error message at https://github.com/buildkite/agent/issues", k, v, v)
-		e.shell.Warningf("OpenTelemetry will still work, but the attribute %v and its value above will not be included", v)
+		e.shell.Warningf("OpenTelemetry will still work, but the attribute %v and its value above will not be included", k)
 	}
 
 	tracerProvider, err := InitOTelTracerProvider(ctx, e.TracingServiceName, extras)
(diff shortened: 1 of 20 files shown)


internal/job/tracing.go
```diff
@@ -181,6 +174,22 @@ func (e *Executor) startTracingOpenTelemetry(ctx context.Context) (tracetools.Sp
...
 	))
 
+	return tracerProvider, nil
+}
+
+func (e *Executor) startTracingOpenTelemetry(ctx context.Context) (tracetools.Span, context.Context, stopper) {
+	extras, warnings := toOpenTelemetryAttributes(GenericTracingExtras(e, e.shell.Env))
+	for k, v := range warnings {
+		e.shell.Warningf("Unknown attribute type (key: %v, value: %v (%T)) passed when initialising OpenTelemetry. This is a bug, submit this error message at https://github.com/buildkite/agent/issues", k, v, v)
+		e.shell.Warningf("OpenTelemetry will still work, but the attribute %v and its value above will not be included", v)
```
buildkate:
```suggestion
		e.shell.Warningf("OpenTelemetry will still work, but the attribute %v and its value above will not be included", k)
```

buildkate (approved):
small nit in comments, but otherwise approved

PR #3867, opened 2026-04-29
Rip out opentracing tracing backend

moskyb:
### Description

Opentracing was deprecated even when we added opentelemetry. We should get rid of it, but we haven't been able to make breaking changes. We're doing v4 now! Goodbye, opentracing 🫡 

### Changes

Remove the `opentracing` tracing backend. Only `opentelemetry` will be a permitted tracing backend now.

### Testing

### Disclosures / Credits

This one was mostly Amp.


DrJosh9000 (approved):
Ooooh, good riddance!

PR #3868, opened 2026-04-29
Replace `cancel-grace-period` and `signal-grace-period-seconds` with `cancel-signal-timeout` and `cancel-cleanup-timeout`

moskyb:
### Description

Previously, `cancel-grace-period` (default 10s) was the *total* time budget covering both process shutdown and agent-side cleanup (log uploads, artifact uploads, disconnects). `signal-grace-period-seconds` (default -1) controlled how much of that budget went to the process, using negative-relative arithmetic: -1 meant "`cancel-grace-period` minus 1", so the process got 9s and the agent got 1s. This made configuration confusing — the flag that *sounded* like the process's grace period (`cancel-grace-period`) was actually the total, and the actual process grace period required subtracting a negative number from it. Validation was also complex because the two values had to be checked against each other, and invalid combinations (e.g.  `signal-grace-period-seconds` >= `cancel-grace-period`) returned errors.

The new model uses two independent, positive durations:

- `--cancel-signal-timeout` (default 10s): how long the subprocess gets to handle the cancel signal before receiving SIGKILL. This is (ofetn) the value users actually think about when configuring cancellation.
- `--cancel-cleanup-timeout` (default 5s): how long the agent gets after the process exits or is killed to upload logs and artifacts.

The total grace period is simply their sum. There is no validation logic because the values cannot conflict. Both flags accept Go duration syntax (e.g. "30s", "1m30s") via `cli.DurationFlag` instead of integer seconds, matching the convention used by other timeout flags in the agent (`wait-for-ec2-tags-timeout`, `kubernetes-container-start-timeout`, etc.).

The defaults produce the same effective behaviour as before: 9s for the process, 1s for agent cleanup, 10s total.

### Testing

### Disclosures / Credits

This one was mostly implemented with Amp, with some fine tuning and tweaking (and checking work) by moi

### DIFF
--- a/agent/agent_configuration.go
+++ b/agent/agent_configuration.go
@@ -58,8 +58,8 @@ type AgentConfiguration struct {
 	DisconnectAfterJob           bool
 	DisconnectAfterIdleTimeout   time.Duration
 	DisconnectAfterUptime        time.Duration
-	CancelGracePeriod            int
-	SignalGracePeriod            time.Duration
+	CancelSignalTimeout          time.Duration
+	CancelCleanupTimeout         time.Duration
 	EnableJobLogTmpfile          bool
 	JobLogPath                   string
 	WriteJobLogsToStdout         bool
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -360,7 +360,7 @@ func NewJobRunner(ctx context.Context, l logger.Logger, apiClient *api.Client, c
 			Stdout:            r.jobLogs,
 			Stderr:            r.jobLogs,
 			InterruptSignal:   cancelSignal,
-			SignalGracePeriod: conf.AgentConfiguration.SignalGracePeriod,
+			SignalGracePeriod: conf.AgentConfiguration.CancelSignalTimeout,
 		})
 	}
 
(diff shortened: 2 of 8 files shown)


DrJosh9000 (approved):
YES... HA HA, HA... **_YES!_**

moskyb:
@DrJosh9000 what do you think of changing the defaults:
- `cancel-signal-timeout`: 10s
- `cancel-cleanup-timeout`: 5s

i feel like it makes the numbers more normal, gives the cleanup a wee bit of extra time, but won't make things take too much longer.

DrJosh9000:
Now's definitely the time to change it. 10s/5s sounds reasonable. I think agent-stack-k8s sets it to 50s/10s by default, bigger may be better.

(A thought to make things more complicated: a separate timeout per cleanup action?)

moskyb:
> A thought to make things more complicated: a separate timeout per cleanup action?

~i wondered about this, but there are like four phases in the cleanup (pre-exit hook, artifact uploads, flush logs, finalize the job), and having four separate knobs feels like perhaps a little too much customisation~

moskyb:
hmm, wait, i've conflated a couple of things there — pre-exit and artifact upload are covered by signal timeout, logs and finalization are covered by cleanup timeout.

that in mind, i still don't think it makes much sense to have per-phase cleanup timeouts. i doubt there would be much utility in it

PR #3872, opened 2026-04-30
Only log first stream connection

moskyb:
### Description

Currently, every time the ping stream is re-established (approximately once a minute), it logs a message `Ping stream connection established`. This gets kind of annoying at times.

This PR makes it so that the message is posted only if either:
- It's the first stream connection ever
- It's the first stream connection after switching modes

If it's a regular reconnection, then the message won't be logged.

### Context

It was annoying me, in a dream

### Testing

### Disclosures / Credits

I did it, it's me.


DrJosh9000 (approved):
LGTM, thanks 👍

PR #3873, opened 2026-04-30
Slay the Basilisk: Remove Bazel from the agent

moskyb:
### Description

We don't actually use Bazel, we added it a little while ago as learning exercise. We're not using it, so let's get rid of it. We can always revert this PR if we decide we want it.

### Context

https://buildkite-corp.slack.com/archives/C07KJ0ZSNN9/p1777531659165559

### Testing

### Disclosures / Credits

Cmd + F "bazel" | rm


DrJosh9000 (approved):
Sounds good to me. No point keeping a bunch of config around if we ain't using it.

PR #3875, opened 2026-04-30
feat[artifact]: add multipart s3 download

buildkate:
### Description

Improve artifact download speed for custom s3 buckets by supporting multipart download. Previously this was done via a single stream

### Context

Resolves [A-1172](https://linear.app/buildkite/issue/A-1172/safety-culture-delay-in-artifact-upload)

### Changes

Uses the feature/s3/manager to fan out a single s3 GET request into parallel ranged requests.
On by default; opt out via
--no-s3-multipart-download.

### Testing
- [ ] Added new e2e test suite that validates behaviour with real custom s3 bucket
- [ ] Test on hosted agent by overriding agent version

### Disclosures / Credits

**R**esearched **P**lanned and **I**mplemented with claude
 Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>


EXPERIMENTS.md
zhming0, line 1: When can we say that we have enough confidence that this can graduate from experiment? Can we use more extensive test coverage to avoid the experiment?  

Our current stance is avoiding experiment if possible (though I notice that we just added the `zip-plugins` experiment 😅 )

buildkate: Yea that's fair. I'd at least like the reporting customer to try it, but I might defer to more seasoned Agents Stewards.

internal/artifact/s3_downloader.go
```diff
@@ -55,6 +75,16 @@ func (d S3Downloader) Start(ctx context.Context) error {
 		return fmt.Errorf("S3Downloader for %s: S3Client is nil", d.conf.S3Path)
 	}
 
+	if experiments.IsEnabled(ctx, experiments.S3MultipartDownloads) {
+		return d.startMultipart(ctx)
+	}
```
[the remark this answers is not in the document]
buildkate: The multipart upload was intentionally not wrapped in a the retrier because the AWS SDK already handles per part retries.  Retrying the outer call could result in 5 (configured retries) x 3 (AWS SDK default retry per part) = 15 total retries.

As a compromise I will wire in `conf.Retries` into the S3 client's per part max retries.

EXPERIMENTS.md
moskyb: one of the tricky things about experiments is that it's often really hard to get customers to actually use them, even when it would fix their issues. saying "try this feature that we're not sure works" is often kind of a hard sell, for obvious reasons.

on the other hand, one of the wonderful things about the agent is that if we break stuff, customers can very easily revert their version and let us know that something broke. so in reality, every change to the agent is opt-in (or opt-out).

obviously this doesn't give us permission to ship stuff that's broken, but it does allow us to Boldly Go a little bit. so long as we've done our due diligence (and based on the tests here, we have), it's okay to ship it and fix it if it breaks.

moskyb: my general rules of thumb on experiments are:
- if at some point we want to say "this feature was a bad idea, i think we should just get rid of it", then we should use an experiment. `zip-plugins` and `pty-raw` are good examples of this — we're not sure whether or not the feature is going to make it to primetime, so we give ourselves permission to remove them if we want.
- if a feature is unfinished at the time of an agent release, it _might_ be worth making an experiment, but it's often going to be easier to have things behind a flag and say "experimental" somewhere in the flag description

there was a time where we were using experiments as escape hatches for backwards compatibility fixes (eg, `normalised-upload-paths`), but i think we're moving away from that as time goes on, and just using regular CLI flags instead. the user interface is more or less the same (the difference between `--experiment normalised-upload-paths` and `--normalised-upload-paths` is small), but having flags is generally more discoverable

zhming0: I agree with @moskyb. I'd vote let's ship without experiment flag 🙏🏿

buildkate: Cheers, thanks for the feedback @zhming0 and @moskyb . I'll remove the experimental flag then, and have it as the default download behaviour with an additional `--no-s3-multipart-download` to have parity with `--no-multipart-artifact-upload` for uploads

internal/artifact/s3_downloader.go
```diff
@@ -54,7 +77,17 @@ func (d S3Downloader) Start(ctx context.Context) error {
 	if d.conf.S3Client == nil {
 		return fmt.Errorf("S3Downloader for %s: S3Client is nil", d.conf.S3Path)
 	}
+	if d.conf.DisableMultipart {
+		return d.startSingle(ctx)
+	}
+	return d.startMultipart(ctx)
```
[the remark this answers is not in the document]
buildkate: Thanks, but this was a deliberate decision as per chats with real humans. https://github.com/buildkite/agent/pull/3875#discussion_r3171579698

clicommand/artifact_download.go
```diff
@@ -80,6 +81,7 @@ var ArtifactDownloadCommand = cli.Command{
 			EnvVar: "BUILDKITE_AGENT_INCLUDE_RETRIED_JOBS",
 			Usage:  "Include artifacts from retried jobs in the search (default: false)",
 		},
+		NoS3MultipartDownloadFlag,
```
moskyb: if this is the only place the flag is being used, just define it inline rather than having it out in globals i reckon

internal/artifact/downloader.go
```diff
@@ -36,6 +36,10 @@ type DownloaderConfig struct {
 	DebugHTTP    bool
 	TraceHTTP    bool
 	DisableHTTP2 bool
+
+	// Disable parallel multipart downloads for s3:// artifacts and fall back
+	// to a single-stream presigned GET. Has no effect on other backends.
+	NoS3MultipartDownload bool
```
moskyb: the general pattern we have for `--no-whatever` flags is to have the flag be negated (eg `--no-s3-multipart-download`), but then have any references to that behaviour happen in the positive (eg `if cfg.S3MultipartDownload { ... }`) as it makes the logic a lot easier to read. `if cfg.S3MultipartDownload` is a lot easier to read than `if !cfg.NoS3MultipartDownload` etc.

see [this block](https://github.com/buildkite/agent/blob/main/clicommand/agent_start.go#L1069-L1083) in agent start — all of the plumbing in the agent uses positive bools, which are negated from the flag values, which are negative bools.

this one is a bit of a persnicket — happy to chat more about it on monday.

internal/artifact/s3_downloader_test.go
moskyb, line 1: these tests are 😍

moskyb (changes requested):
a couple little nits, but it's looking really good!

moskyb (approved):
🤠

PR #3876, opened 2026-05-01
Flatten tracing config to be a single bool flag

moskyb:
### Description

In https://github.com/buildkite/agent/pull/3867, we removed the opentracing tracing backend, leaving us with only opentelemetry (or none) as our available tracing backends. This PR updates the `buildkite-agent start` command such that:
- the `--tracing-backend` flag has been removed
- replaced with a `--opentelemetry-tracing` flag
- the `--tracing-propagate-traceparent` flag has been removed, and enabled by default when tracing is turned on

The internals of the agent's tracing bits and bobs have intentionally been left untouched. there's a bunch of code in there that exists to swap between interfaces for tracing that doesn't really need to be there anymore, but those changes will be internal-only, so i'll leave them for a later PR.

### Context

You'll never guess where it came to me.

### Testing

### Disclosures / Credits

A bit of Amp, a bit of me.

### DIFF
--- a/agent/agent_configuration.go
+++ b/agent/agent_configuration.go
@@ -69,7 +69,7 @@ type AgentConfiguration struct {
 	Profile                      string
 	RedactedVars                 []string
 	AcquireJob                   string
-	TracingBackend               string
+	OpenTelemetryTracing         bool
 	TracingServiceName           string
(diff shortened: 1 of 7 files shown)


DrJosh9000 (approved):
I like it

PR #3877, opened 2026-05-01
Preserve percent sequences in JSON job logs

friel-openai:
### Description

`JSONJobLogger.Write` currently passes job output directly to `Logger.Info`. `Logger.Info` treats its first argument as a Go format string, so job output containing a percent sequence can be rewritten before it reaches the JSON `msg` field.

This PR changes `JSONJobLogger.Write` to pass the job output as a `%s` argument instead. That keeps literal output such as `Receiving objects:   0% (1/140)` unchanged.

### Context

We observed JSON job logs where literal percent sequences in command output were changed by Go formatting before log consumers saw them.

### Changes

- Preserve literal percent sequences in `JSONJobLogger.Write`.
- Add a regression test for a Git-style progress line containing `%`.

### Testing

### Disclosures / Credits

I used Codex to inspect the logging path, make this small patch, and run the local tests.

### DIFF
--- a/agent/json_job_logger.go
+++ b/agent/json_job_logger.go
@@ -56,6 +56,6 @@ func (l JSONJobLogger) Write(data []byte) (int, error) {
 	// When writing as a structured log, trailing newlines and carriage returns
 	// generally don't make sense.
 	msg := strings.TrimRight(string(data), "\r\n")
-	l.log.Info(msg)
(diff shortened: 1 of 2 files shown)


nsuma8989 (approved):
Looks good to me

DrJosh9000 (approved):

PR #3878, opened 2026-05-03
chore: make logger.Logger methods end in f

DrJosh9000:
### Description

It's easy to forget that the first arg to Debug/Info/Warn/Error/Fatal/Notice is a format arg, because it didn't use the convention of putting an `f` on the end of the method name.

### Context

#3877

### Changes

- Change the method names
- Add a compatibility wrapper for `signature`

### Testing

### Disclosures / Credits

Search and replace - 10 minute job.


pda (approved):
Good idea 👍🏼

PR #3880, opened 2026-05-04
fix: separate OTel provider provisioning from span generation

ss1909:
#### Problem

`setupLoggerAndConfig` only initialised the OpenTelemetry TracerProvider when `TRACEPARENT` was set in the environment. This was unreliable: `TRACEPARENT` is injected into every subprocess the agent spawns (git, bootstrap, hooks) so its presence is not a reliable signal that the current process is a traced subcommand.

This also caused a double-initialisation bug in the bootstrap flow: `setupLoggerAndConfig` created one provider, and `startTracingOpenTelemetry` created a second, loosing the first one.

#### Fix

Separate the two concerns:

- `setupLoggerAndConfig` always initialises the provider when `BUILDKITE_TRACING_BACKEND=opentelemetry`, and owns its lifecycle (`ForceFlush` + `Shutdown` in `done()`)
- `startTracingOpenTelemetry` uses `otel.GetTracerProvider()` to obtain the already-registered provider and only creates the root job span, adding job-specific attributes (pipeline slug, job ID, etc.) at span level rather than resource level since they are not available at provider init time

### Testing

screenshots on grafana to show the subcommands:
image

### DIFF
--- a/internal/job/tracing.go
+++ b/internal/job/tracing.go
@@ -127,7 +127,7 @@ func (e *Executor) otRootSpanName() string {
 	return base + "job"
 }
 
-// initOTelTracerProvider creates and globally registers an OpenTelemetry TracerProvider
+// InitOTelTracerProvider creates and globally registers an OpenTelemetry TracerProvider
 // and text map propagator. Caller must call ForceFlush and Shutdown
 // on the returned provider before the process exits.
 func InitOTelTracerProvider(ctx context.Context, serviceName string, extraAttrs []attribute.KeyValue) (*sdktrace.TracerProvider, error) {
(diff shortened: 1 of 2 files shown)


internal/job/tracing.go
```diff
@@ -178,38 +178,30 @@ func InitOTelTracerProvider(ctx context.Context, serviceName string, extraAttrs
...
-	stop := func() {
-		ctx := context.Background()
-		_ = tracerProvider.ForceFlush(ctx)
-		_ = tracerProvider.Shutdown(ctx)
-	}
-
-	return tracetools.NewOpenTelemetrySpan(span), ctx, stop
+	// The provider ForceFlush+Shutdown is owned by setupLoggerAndConfig's done()
+	// closure. This stopper only needs to end the span so it is exported before
+	// the flush runs.
```
zhming0: These comments seem to be unnecessary, it's good as a PR comment but without this PR this comments will read out of context.

internal/job/tracing.go
```diff
@@ -178,38 +178,30 @@ func InitOTelTracerProvider(ctx context.Context, serviceName string, extraAttrs
 }
 
 func (e *Executor) startTracingOpenTelemetry(ctx context.Context) (tracetools.Span, context.Context, stopper) {
+	// The TracerProvider has already been initialised and globally registered
+	// by setupLoggerAndConfig when the bootstrap process started.
```
zhming0: Same to this comment.

zhming0 (approved):
LGTM ✅ with some nits. Well done 👍🏿

PR #3899, opened 2026-05-07
Remove header times, ANSI timestamps only and always

DrJosh9000:
### Description

Header times no longer bring joy.
Timestamps disabled no longer brings joy.
Plaintext timestamps never brought joy in the first place.

ANSI timestamps only and always!

### Context

https://buildkite-corp.slack.com/archives/C06J07P1TKM/p1778112393905589

### Changes

Remove the header times streamer.
Remove the flags for disabling ANSI timestamps and enabling plaintext timestamps.
Remove the somewhat confusing IO indirection in the job runner that enabled having three different timestamping modes with or without header times.

### Testing

### Disclosures / Credits

I used a Keyboardio Atreus to type the changes.


agent/job_runner.go
```diff
@@ -287,22 +280,11 @@ func NewJobRunner(ctx context.Context, l logger.Logger, apiClient *api.Client, c
 			}
 		}()
 
-	default:
-		// processWriter -> {pw, outputWriter};
-		// pw -> pr -> process.Scanner -> headerTimesStreamer
+	default: // neither ANSI timestamps nor plaintext timestamps
```
moskyb: should we enforce that everything have ANSI timestamps?

moskyb (approved):
Love to see it

agent/job_runner.go
DrJosh9000: Done!

PR #3900, opened 2026-05-07
feat: add jobsRunning gauge and labels for Prometheus metrics

buildkate:
### Description

- Adds priority and queue labels to Prometheus job start/ended counters
- Adds jobsRunning gauge for Prometheus with priority and queue labels

### Context

Resolves [A-1100](https://linear.app/buildkite/issue/A-1100/add-job-metadata-labels-to-agent-prometheus-metrics)
Corresponding PR in buildkite/buildkite https://github.com/buildkite/buildkite/pull/29440

### Testing
- [x] Unit tests added to cover changes

### Disclosures / Credits
In consultation with Claude Opus 4.7

### DIFF
--- a/agent/metrics.go
+++ b/agent/metrics.go
@@ -57,6 +57,9 @@ var (
 		Buckets:   prometheus.LinearBuckets(1, 1, 20),
 	})
 
+	// Original unlabelled counters — kept for backwards compatibility with
+	// existing scrape consumers. New code should prefer the labelled
+	// jobsStartedWithLabels / jobsEndedWithLabels counters below.
 	jobsStarted = promauto.NewCounter(prometheus.CounterOpts{
 		Namespace: metricsNamespace,
 		Subsystem: "jobs",
@@ -69,6 +72,29 @@ var (
 		Name:      "ended_total",
 		Help:      "Count of jobs ended (any outcome)",
 	})
+
+	// Labelled counters added for A-1100. These coexist with the unlabelled
+	// jobsStarted / jobsEnded above; both are incremented at the same call
+	// site so the unlabelled total always equals sum() over the labelled.
+	jobsStartedWithLabels = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace: metricsNamespace,
+		Subsystem: "jobs",
+		Name:      "started_with_labels_total",
+		Help:      "Count of jobs started, labelled by job priority and agent queue",
+	}, []string{"priority", "queue"})
+	jobsEndedWithLabels = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace: metricsNamespace,
+		Subsystem: "jobs",
+		Name:      "ended_with_labels_total",
+		Help:      "Count of jobs ended (any outcome), labelled by job priority and agent queue",
+	}, []string{"priority", "queue"})
+
+	jobsRunning = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: metricsNamespace,
+		Subsystem: "jobs",
+		Name:      "running",
+		Help:      "Number of jobs currently running on this agent, labelled by job priority and agent queue",
+	}, []string{"priority", "queue"})
 
 	logChunksUploaded = promauto.NewCounter(prometheus.CounterOpts{
 		Namespace: metricsNamespace,
--- a/agent/metrics_test.go
+++ b/agent/metrics_test.go
@@ -1,0 +1,177 @@
+package agent
+
+import (
+	"strconv"
+	"testing"
+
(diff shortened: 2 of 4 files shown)


agent/metrics.go
```diff
@@ -57,18 +57,24 @@ var (
...
 		Name:      "ended_total",
 		Help:      "Count of jobs ended (any outcome)",
-	})
+	}, []string{"priority", "queue"})
+	jobsRunning = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: metricsNamespace,
+		Subsystem: "jobs",
+		Name:      "running",
+		Help:      "Number of jobs currently running on this agent",
+	}, []string{"priority", "queue"})
```
blaknite: Important: Changing jobsStarted and jobsEnded from Counter to CounterVec is a breaking change for anyone scraping buildkite_agent_jobs_started_total and buildkite_agent_jobs_ended_total today. Previously these were flat
counters; after this change they become labelled, so existing PromQL queries like buildkite_agent_jobs_started_total will stop returning a single scalar and instead return a vector. Existing dashboards and alerts that
reference these metrics will need updating.

Is this intentional and expected? If so it's probably worth calling out in the PR description (or release notes) so folks consuming these metrics know what's coming.

If the goal was to avoid breakage, the labelled
variants could be separate metrics alongside the existing unlabelled ones.

api/jobs.go
```diff
@@ -27,6 +27,7 @@ type Job struct {
 	RunnableAt            string                     `json:"runnable_at,omitempty"`
 	ChunksFailedCount     int                        `json:"chunks_failed_count,omitempty"`
 	TraceParent           string                     `json:"traceparent"`
+	Priority              int                        `json:"priority"`
```
blaknite, line 30: Nit: The other int fields on the Job struct (ChunksFailedCount, ChunksIntervalSeconds) use omitempty. Priority doesn't. This means the priority field will always be serialized in JSON even when zero, which might matter if
this struct is ever sent back to the API. Since the zero value (0) is a meaningful default priority, I think omitting omitempty is actually correct here — just wanted to flag it as an intentional deviation from the
pattern on the neighbouring fields.

agent/metrics_test.go
blaknite, line 1: Minor: The tests validate that the Prometheus wiring is correct (label order, independence of label combinations, gauge inc/dec) but they do this by directly calling the metric vars rather than exercising RunJob.

Without than having to scaffold the whole thing, is it easy to check RunJob calls the underlying functions?

blaknite (changes requested):
I'm an absolute Go novice and what you see here is the result of me asking an agent questions.

Looks like the change from Counter to CounterVec is a legit regression worth understanding.

api/jobs.go
buildkate: This was intentional since the Job db table itself has a non-nil directive `priority integer DEFAULT 0 NOT NULL`  https://github.com/buildkite/buildkite/blob/main/db/pipelines_structure.sql#L2517

Please let me know if I've understood that incorrectly.

agent/metrics.go
buildkate: I did read into this and that the breakage would be visual (as in some queries might return more than one series), but to be totally safe I'll create duplicates.  I might add a PR to v4 to remove the unlabelled metrics

api/jobs.go
DrJosh9000: `omitempty` is used when the struct is marshalled into JSON, not when it is unmarshalled from JSON. I don't think the agent ever marshals `Job` back into JSON, so I think that `omitempty` could be removed from all its fields.

agent/metrics_test.go
buildkate: That's a fair observation. The tests deliberately stop at verifying the metric mechanics rather than driving RunJob itself.

The blocker for going further is that CounterVec/GaugeVec in the prometheus/client_golang are concrete structs, not interfaces, so to assert "RunJob called Inc on this specific metric" we'd need to either wrap each metric in a private interface (production indirection added for the sake of testing).

api/jobs.go
buildkate: Good shout @DrJosh9000 .  I'll defer the broader cleanup to a separate PR to keep this one focused.

agent/metrics_test.go
blaknite: Sounds like you've landed on the right choice. Happy to leave it where it is.

api/jobs.go
blaknite: Works for me!

agent/metrics.go
blaknite: Excellent!

blaknite (approved):

PR #3905, opened 2026-05-07
Add support for git commit verification

omehegan:
### Description

Adds commit-on-branch verification to the checkout phase. 

As part of our agent checkout improvement work, we are adding support for git commit verification. This addresses a potential security issue: if an attacker adds a malicious commit to the repo, and is then able to trigger a build which specifies `commit: 3v1l5ha123` `branch: main` we would build that commit as if it were on the `main` branch, without verifying that this is the case, which could potentially lead to a production deployment of malicious code. 

This change introduces a `BUILDKITE_GIT_COMMIT_VERIFICATION` environment variable, which defaults to `false`. If set to `true`, when we are given a commit and branch, before checkout and build, we do the following:

1. Use `git merge-base --is-ancestor` to check if the commit really belongs to the specified branch. 
2. If it does, proceed; if it definitely does not, fail and exit.
3. In ambiguous situations, it's possible that we only have a shallow clone of the repo. In that case we:
4. Deepen the clone by 50 commits, with the assumption that this should be enough to find the specified SHA.
5. If we still get an ambiguous result, to a full `unshallow` and check again.
6. Proceed if the commit matches, fail and exit if it does not.

In terms of alternatives, the only change I considered was just going straight to `unshallow` before verifying any commit. But I didn't want users to have to completely give up shallow commits to get verification, so I went with the "deepen first, then unshallow if necessary" approach.

### Context

See SUP-6535.

### Changes

Adds a `BUILDKITE_GIT_COMMIT_VERIFICATION` env var to enable to functionality. When set, we perform various git operations in between fetch and checkout, to make sure the provided commit is valid on the provided branch.

### Testing

### Disclosures / Credits

I used Claude to help me plan the feature and teach me some Go fundamentals. I coded the core functionality myself, and let Claude write the tests.

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -167,6 +167,7 @@ type AgentStartConfig struct {
 	GitMirrorsLockTimeout       int      `cli:"git-mirrors-lock-timeout"`
 	GitMirrorsSkipUpdate        bool     `cli:"git-mirrors-skip-update"`
 	GitCheckoutTimeout          int      `cli:"git-checkout-timeout"`
+	GitCommitVerification       string   `cli:"git-commit-verification"`
 	NoGitSubmodules             bool     `cli:"no-git-submodules"`
 	GitSubmoduleCloneConfig     []string `cli:"git-submodule-clone-config"`
 	SkipCheckout                bool     `cli:"skip-checkout"`
@@ -371,7 +372,8 @@ var AgentStartCommand = cli.Command{
 	Name:        "start",
 	Usage:       "Starts a Buildkite agent",
 	Description: startDescription,
-	Flags: append(globalFlags(),
+	Flags: append(
+		globalFlags(),
 		cli.StringFlag{
 			Name:   "config",
 			Value:  "",
@@ -534,6 +536,7 @@ var AgentStartCommand = cli.Command{
 		GitCheckoutFlagsFlag,
 		GitCloneFlagsFlag,
 		GitCleanFlagsFlag,
+		GitCommitVerificationFlag,
 		GitFetchFlagsFlag,
 		GitCloneMirrorFlagsFlag,
 		GitMirrorsPathFlag,
@@ -863,6 +866,11 @@ var AgentStartCommand = cli.Command{
 			}
 		}
 
+		// Validate the commit verification option input
+		if v := cfg.GitCommitVerification; v != "" && v != "strict" && v != "warn" {
+			return fmt.Errorf("invalid value for --git-commit-verification: %q (must be \"strict\" or \"warn\")", v)
+		}
+
 		// Force some settings if on Windows (these aren't supported yet)
 		if runtime.GOOS == "windows" {
 			cfg.NoPTY = true
@@ -1070,6 +1078,7 @@ var AgentStartCommand = cli.Command{
 			GitCloneFlags:                   cfg.GitCloneFlags,
 			GitCloneMirrorFlags:             cfg.GitCloneMirrorFlags,
 			GitCleanFlags:                   cfg.GitCleanFlags,
+			GitCommitVerification:           cfg.GitCommitVerification,
 			GitFetchFlags:                   cfg.GitFetchFlags,
 			GitSubmodules:                   !cfg.NoGitSubmodules,
 			GitSubmoduleCloneConfig:         cfg.GitSubmoduleCloneConfig,
--- a/internal/job/checkout.go
+++ b/internal/job/checkout.go
@@ -968,6 +968,10 @@ func (e *Executor) defaultCheckoutPhase(ctx context.Context) (retErr error) {
 		return err
 	}
 
+	if err := e.verifyCommit(ctx); err != nil {
+		return err
+	}
+
 	gitCheckoutFlags := e.GitCheckoutFlags
 
 	if e.Commit == "HEAD" {
--- a/internal/job/commit_verification.go
+++ b/internal/job/commit_verification.go
@@ -1,0 +1,134 @@
+package job
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/buildkite/agent/v3/internal/shell"
+)
+
+// ErrCommitVerificationFailed indicates that git has definitively determined
+// the commit is not on the specified branch. This is the security-relevant case.
+var ErrCommitVerificationFailed = errors.New("commit verification failed")
+
+// ErrCommitVerificationUnavailable indicates that git was unable to perform
+// the ancestry check (e.g. due to repo corruption, misconfigured remotes, etc).
+// This is NOT evidence of an attack — it's an infrastructure problem.
+var ErrCommitVerificationUnavailable = errors.New("commit verification unavailable")
+
+// checkCommitOnBranch performs the actual git ancestry check, handling shallow
+// clones by deepening or unshallowing as needed. It returns:
+//   - nil if the commit is verified on the branch
+//   - ErrCommitVerificationFailed if the commit is definitively not on the branch
+//   - ErrCommitVerificationUnavailable if the check cannot be performed
+func (e *Executor) checkCommitOnBranch(ctx context.Context) error {
+	e.shell.Commentf("Verifying commit %q is on branch %q", e.Commit, e.Branch)
+
+	for _, fetchFlag := range []string{"", "--deepen=50", "--unshallow"} {
+		if fetchFlag != "" {
+			// After the first iteration, try to unshallow the repo (a bit or a lot)
+			e.shell.Commentf("Deepening checkout to verify commit (%s)...", fetchFlag)
+			fetchErr := e.shell.Command("git", "fetch", fetchFlag).Run(ctx)
+			if fetchErr != nil {
+				return fmt.Errorf("%w: unable to verify commit %q on branch %q: %w", ErrCommitVerificationUnavailable, e.Commit, e.Branch, fetchErr)
+			}
+		}
+
+		// Try the ancestry check
+		err := e.shell.Command("git", "merge-base", "--is-ancestor", e.Commit, e.Branch).Run(ctx)
+
+		switch shell.ExitCode(err) {
+		case 0:
+			return nil // verified!
+		case 1:
+			return fmt.Errorf("%w: commit %q is not on branch %q", ErrCommitVerificationFailed, e.Commit, e.Branch)
+		case 128:
+			// unclear — continue below
+		default:
+			return fmt.Errorf("%w: unable to verify commit %q on branch %q: %w", ErrCommitVerificationUnavailable, e.Commit, e.Branch, err)
+		}
+
+		// On the first iteration, check if the checkout is shallow. If it is
+		// not, the 128 exit code reflects some other error.
+		if fetchFlag != "" {
+			continue
+		}
+		output, shallowErr := e.shell.Command("git", "rev-parse", "--is-shallow-repository").RunAndCaptureStdout(ctx)
+		if shallowErr != nil {
+			return fmt.Errorf("%w: unable to verify commit %q on branch %q: %w", ErrCommitVerificationUnavailable, e.Commit, e.Branch, shallowErr)
+		}
+
+		if strings.TrimSpace(output) != "true" {
+			// Not shallow — this is a genuine error
+			return fmt.Errorf("%w: unable to verify commit %q on branch %q: %w", ErrCommitVerificationUnavailable, e.Commit, e.Branch, err)
+		}
+	}
+
+	// All attempts exhausted — verification is unavailable
+	return fmt.Errorf("%w: unable to verify commit %q on branch %q after exhausting fetch strategies", ErrCommitVerificationUnavailable, e.Commit, e.Branch)
+}
+
+// verifyCommit is called if the user has commit verification enabled. It ensures that the commit we are
+// asked to build exists and is reachable on the branch we are given.
+func (e *Executor) verifyCommit(ctx context.Context) error {
+	// Skip if not enabled
+	if e.GitCommitVerification == "" {
+		return nil
+	}
+
+	// Skip if commit is HEAD (nothing to verify)
+	if e.Commit == "HEAD" {
+		e.shell.Commentf("Skipping commit verification: commit is HEAD")
+		return nil
+	}
+
+	// Skip if we haven't been given a branch - e.g. it's a tag push event
+	if e.Branch == "" {
+		e.shell.Commentf("Skipping commit verification: no branch specified")
+		return nil
+	}
+
+	// Skip if this is a tag build — tags are not branch-specific
+	if e.Tag != "" {
+		e.shell.Commentf("Skipping commit verification: tag build (%s)", e.Tag)
+		return nil
+	}
+
+	// Skip if this is a PR build — the commit may be on a merge ref, not the target branch
+	if e.PullRequest != "" {
+		e.shell.Commentf("Skipping commit verification: pull request build (#%s)", e.PullRequest)
+		return nil
+	}
+
+	// Skip if a custom refspec is set — the fetch may not populate standard branch refs,
+	// making ancestry verification unreliable
+	if e.RefSpec != "" {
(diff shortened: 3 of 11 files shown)


internal/job/checkout.go
```diff
@@ -806,6 +806,116 @@ func (e *Executor) fetchSource(ctx context.Context) error {
...
+	// Skip if this is a PR build — the commit may be on a merge ref, not the target branch
+	if e.PullRequest != "" {
+		return nil
+	}
+
+	// Skip if a custom refspec is set — the fetch may not populate standard branch refs,
+	// making ancestry verification unreliable
+	if e.RefSpec != "" {
+		return nil
+	}
```
zhming0: While I totally see the value of this change adding a level of protection, I worry that it's a bit narrow given all these bypass conditions. 

Also on the direction, it's ultimately saying: "under X case, do not trust Buildkite backend". It's a questionable direction for us to invest in though we have prior arts (I think). But I do wonder if we could/should strengthen the backend to not send false information. 

It's not strictly a blocker, but something worth a number of words to explain about.

internal/job/checkout.go
zhming0, line 1: Given this change is largely isolated from the happy path of `checkout.go` and `checkout.go` is getting really big. Would it be possible move them to maybe e.g. `checkout_verify_commit.go`? 

This comment applies to the test code in `checkout_test.go` too

internal/job/checkout.go
```diff
@@ -806,6 +806,116 @@ func (e *Executor) fetchSource(ctx context.Context) error {
...
+		if retryCode == 0 {
+			return nil // Found a valid commit after deepening
+		}
+		if retryCode == 1 {
+			return fmt.Errorf("commit %q is not on branch %q", e.Commit, e.Branch)
+		}
+
+		// Still 128 - full unshallow as last resort
+		e.shell.Commentf("Deepening insufficient, performing a full unshallow...")
+		_ = e.shell.Command("git", "fetch", "--unshallow").Run(ctx)
```
zhming0: I wonder if the cost outweigh the gain here? Now the equation become, if someone put a random commit  SHA in the build api, our agent will basically do a repo clone -> which depending on situation can take a long time. 

An realistically, building for a old commit in CI is pretty rare 🤔

zhming0 (commented):
I see the value in this change, but I left some questions re overall directions and the cost / gain asymmetry. Also I recommend relocating some codes to another file.

internal/job/checkout.go
DrJosh9000: > "under X case, do not trust Buildkite backend"

There's plenty of prior art for client-side protections: pre-bootstrap hook for job validation, Signed Pipelines, allowed-env-vars flag, allowed-plugins flag, etc.

In the original Buildkite model, where the backend doesn't have access to the code, the backend can't determine whether or not a particular commit is on a particular branch. I'm not sure how we would strengthen the backend to send only commits on the intended branch without requiring code access.

internal/job/checkout.go
omehegan: I'm not sure what you're suggesting - do you think we should never go through the unshallow step, because it will make builds slower if an attacker sends a random commit via the build API? If this person already has API access, we have bigger things to worry about. The more likely attack scenario is someone putting a legitimate malicious commit on a branch and tricking us into building it as though it was on `main` (i.e. the pipeline runs with main branch permissions, could deploy, etc.).

You're thinking about the legitimate scenario of building an "old" commit where the "deepen by 50 commits" might not be enough. But in a busy monorepo, a rebuild from even a day ago could be more than 50 commits behind, so we'd have to unshallow to do the verification. A user is going to expect that. I'm just trying to avoid scenarios where we block legitimate builds.

internal/job/checkout.go
omehegan: I agree with Josh about the client-side protections. There's no way for us to verify that a commit is or isn't on the claimed branch without having a checkout of the repo. Theoretically there are ways that we could do this with the GitHub API before triggering a build, but such a solution has potential access issues, race conditions, and rate limit implications.

In terms of the bypass conditions you mentioned, those are all scenarios in which commit verification isn't meaningful. For example:

`HEAD` commit - we're not building a specific commit, we're building whatever is at HEAD, we're not being told to build something specific.
Empty branch - if we're not given a branch name, then there are no branch-based conditions to exploit.
PR builds - If it's a PR build, there's no risk of pretending that it's on main.

internal/job/checkout.go
zhming0: Ah soz I should be clearer, I think there is pretty big jump between check 50 commits to a full repo scan.

I wonder if it's an legitimate commit in a busy repo, how likely will we trigger the full repo scan.

This isn't a blocker, just something I am trying to understand.

internal/job/checkout.go
zhming0: 👍 I don't see another way out either for this particular defense so I have no objection. My main point is mainly around holistically if this prevention mechanism is meaningful enough when an attacker gained BK API access.

Could the attacker specify refspec (which bypass the protection) + branch at the same time? I didn't trace that code path so a bit unsure.

So not a blocker, mainly a question for my understanding.

omehegan: Yeah I got you - so what I would say is that there are two attack vectors here. The first is webhook manipulation - someone creates a branch and commits malicious code to it, then spoofs a webhook that includes that commit SHA, with `branch: main`. Refspec isn't part of this payload, so that concern isn't valid here. Without verification, this is enough for us to build a malicious commit with `main` permissions.

The second vector is what you're thinking about, BK API access. As I mentioned in my other comment, if someone compromises an API token, we have bigger things to worry about. They can do more damage than what verification is going to prevent. You're right that with that access they could bypass verification by sending a bogus refspec. But we have to skip verification in the custom refspec case, because we have to assume that `main` resolves to something meaningful locally, in order to check if a provided commit SHA legitimately exists on it or not. A custom refspec breaks that assumption, so we have to skip attempting verification or else risk false failures.

Let me know if all that makes sense!

internal/job/checkout.go
omehegan: Got it! Yeah it's a tradeoff, and I did think about what the right balance was. First, this only matters for people using shallow clones, which AFAIK is a small percentage of current users. Second, if you do shallow clones on a repo that isn't very busy, `deepen 50` is probably going to be enough - especially for people who are doing something reasonable like `depth 10` rather than `depth 1`. It's only very busy branches where, if I understand your concern, `deepen 50` might not be enough, but a full `unshallow` would be expensive. In that scenario you might wish for an intermediate `deepen 250` step or something, before the full `unshallow`. Is that what you have in mind? If so, I think what I'd suggest is that we stick with the current solution and see how it goes. If this starts to become a problem, a potential solution would be to make the `deepen` amount configurable, instead of hard-coded to 50 commits. So for a busy repo, if this keeps coming up, they could change it to a value that suits them.

internal/job/checkout.go
omehegan: Done, thanks for that suggestion :)

internal/job/checkout.go
zhming0: 👍 happy to give this a shot.

zhming0:
Another thing we raised in the Agent Office hour is the big feature branch approach. We recommend targeting the main branch: 1. to derisk the big bang deployment. 2. V4 is happening in parallel, we worry that you would hit non-trivial conflicts if we hold a long running feature branch. Oz and Ben might be able to share a bit more context, they were on the meeting.

What do you think?

omehegan:
@zhming0 yeah I'm catching up on our internal thread about that topic. I was just following the example we were already using for this work, but since it sounds like folks would rather have us merge to main in discrete chunks, I'm fine with that.

internal/job/config.go
```diff
@@ -104,6 +104,9 @@ type ExecutorConfig struct {
 	// Flags to pass to "git clean" command
 	GitCleanFlags string `env:"BUILDKITE_GIT_CLEAN_FLAGS"`
 
+	// Enable git commit verification
+	GitCommitVerification string `env:"BUILDKITE_GIT_COMMIT_VERIFICATION"`
```
DrJosh9000: This is setting the `env` tag on the field, which is used to update the field if a plugin/hook changes it. But the env var is added in `env/protected.go` without setting `mutableFromWithinJob: true`. If it's not meant to be mutable from within the job, then remove the tag here.

clicommand/agent_start.go
DrJosh9000, line 1: Consider adding a check to the Action that `GitCommitVerification` has one of the accepted values, so that it fails early if there's a configuration mistake.

internal/job/commit_verification.go
```diff
@@ -0,0 +1,150 @@
...
+	// Skip if this is a PR build — the commit may be on a merge ref, not the target branch
+	if e.PullRequest != "" {
+		return nil
+	}
+
+	// Skip if a custom refspec is set — the fetch may not populate standard branch refs,
+	// making ancestry verification unreliable
+	if e.RefSpec != "" {
+		return nil
+	}
```
DrJosh9000, line 110: Each of these cases should generate a log - at least at the debug level but maybe a higher level - because the user may be confused why verification isn't running in their case.

internal/job/commit_verification.go
```diff
@@ -0,0 +1,150 @@
...
+		// Verification was unavailable — warn but don't block
+		e.shell.Warningf("Commit verification unavailable: %v", err)
+		return nil
+	case "warn":
+		e.shell.Warningf("Commit verification failed: %v", err)
+		return nil
+	default:
+		e.shell.Warningf("Unknown git-commit-verification value %q, skipping verification", e.GitCommitVerification)
+		return nil
+	}
```
DrJosh9000, line 127: `// Verification passed` could be misleading, since it's placed before the `if` that checks `err`.

More importantly, `if err == nil { ... }` is unusual enough that you should consider the opposite or at least a comment flagging it (https://google.github.io/styleguide/go/decisions.html#indent-error-flow). 

Thirdly, distinguishing the two failure types is probably just as important for the "warn" case as the "strict" case, so perhaps the nesting should be swapped around? 

Combining these lets us do the `errors.Is` test first, returning early and logging based on the mode, and then catch other errors.

Fourthly, I'm not 100% convinced on the value of about half of the comment...

```suggestion
	if errors.Is(err, ErrCommitVerificationFailed) {
		switch e.GitCommitVerification {
		case "strict":
			// In strict mode, only a definitive failure (ErrCommitVerificationFailed) blocks
			// the build. 
			return err
		case "warn":
			e.shell.Warningf("Commit verification failed: %v", err)
			return nil
		}
	}
	if err != nil {
		// If verification is unavailable, we warn but allow the build to proceed.
		// This avoids users disabling verification entirely due to infrastructure false
		// positives, which would leave them with no protection at all.
		e.shell.Warningf("Commit verification unavailable: %v", err)
	}
	return nil
```

DrJosh9000 (commented):
🚴

internal/job/commit_verification.go
```diff
@@ -0,0 +1,150 @@
...
+	exitCode := shell.ExitCode(err)
+
+	switch exitCode {
+	case 0:
+		return nil // verified!
+	case 1:
+		return fmt.Errorf("%w: commit %q is not on branch %q", ErrCommitVerificationFailed, e.Commit, e.Branch)
+	case 128:
+		// We might have a shallow clone, try to deepen or unshallow to find the commit
+		output, _ := e.shell.Command("git", "rev-parse", "--is-shallow-repository").RunAndCaptureStdout(ctx)
```
DrJosh9000: Why is the error value from this command being ignored? It might be vanishingly unlikely to happen (`git` somehow disappears from `$PATH` in between the `merge-base` call and here?), but I think that's an argument in favour of handling the error (report and make clear unlikely conditions when they occur). https://google.github.io/styleguide/go/decisions.html#handle-errors

internal/job/commit_verification.go
```diff
@@ -0,0 +1,150 @@
...
+
+		if strings.TrimSpace(output) != "true" {
+			// Not shallow — this is a genuine error
+			return fmt.Errorf("%w: unable to verify commit %q on branch %q: %w", ErrCommitVerificationUnavailable, e.Commit, e.Branch, err)
+		}
+
+		// Try deepening by 50 commits first
+		e.shell.Commentf("Shallow clone detected, deepening by 50 commits...")
+		_ = e.shell.Command("git", "fetch", "--deepen=50").Run(ctx)
```
DrJosh9000: Same here - if `--deepen` fails then we can at least return early and avoid the subsequent code that is just as likely to fail.

internal/job/commit_verification.go
```diff
@@ -0,0 +1,150 @@
...
+		retryCode := shell.ExitCode(retryErr)
+
+		if retryCode == 0 {
+			return nil // Found a valid commit after deepening
+		}
+		if retryCode == 1 {
+			return fmt.Errorf("%w: commit %q is not on branch %q", ErrCommitVerificationFailed, e.Commit, e.Branch)
+		}
+
+		// Still 128 - full unshallow as last resort
```
DrJosh9000: _Is_ it still 128?

Perhaps it would be good to figure out how to make this function into a loop, so the exit code handling from `merge-base --is-ancestor` is consistent.

Here's my attempt at this (including more error handling):

```go
	for _, fetchFlag := range []string{"", "--deepen=50", "--unshallow"} {
		if fetchFlag != "" {
			// After the first iteration, try to unshallow the repo (a bit or a lot)
			fetchErr := e.shell.Command("git", "fetch", fetchFlag).Run(ctx)
			if fetchErr != nil {
				return fmt.Errorf("%w: unable to verify commit %q on branch %q: %w", ErrCommitVerificationUnavailable, e.Commit, e.Branch, fetchErr)
			}
		}

		// Try the ancestry check
		err := e.shell.Command("git", "merge-base", "--is-ancestor", e.Commit, e.Branch).Run(ctx)

		switch shell.ExitCode(err) {
		case 0:
			return nil // verified!
		case 1:
			return fmt.Errorf("%w: commit %q is not on branch %q", ErrCommitVerificationFailed, e.Commit, e.Branch)
		case 128:
			// unclear - continue below
		default:
			return fmt.Errorf("%w: unable to verify commit %q on branch %q: %w", ErrCommitVerificationUnavailable, e.Commit, e.Branch, err)
		}

		// On the first iteration, check if the checkout is shallow. If it is
		// not, the 128 exit code reflects some other error.
		if fetchFlag != "" {
			continue
		}
		output, shallowErr := e.shell.Command("git", "rev-parse", "--is-shallow-repository").RunAndCaptureStdout(ctx)
		if shallowErr != nil {
			return fmt.Errorf("%w: unable to verify commit %q on branch %q: %w", ErrCommitVerificationUnavailable, e.Commit, e.Branch, err)
		}

		if strings.TrimSpace(output) != "true" {
			// Not shallow — this is a genuine error
			return fmt.Errorf("%w: unable to verify commit %q on branch %q: %w", ErrCommitVerificationUnavailable, e.Commit, e.Branch, err)
		}
	}
```

internal/job/config.go
omehegan: Ah, I misunderstood how this works - thanks for explaining.

internal/job/commit_verification.go
omehegan: Took me a few gear turns, but I see what you mean now. However I think the suggested commit was in the wrong spot; revised it to what I think was your intent.

internal/job/commit_verification.go
omehegan: Also fixed it for the unshallow case.

internal/job/commit_verification.go
omehegan: Oh yeah, that is way cleaner. I've integrated that!

internal/job/commit_verification.go
DrJosh9000: That's looking a lot better, though I still think you should also change

```go
if err == nil {
	return nil
}

...snip...

e.shell.Warningf("Commit verification unavailable: %v", err)
return nil
```

to

```go
...snip...

if err != nil {
	e.shell.Warningf("Commit verification unavailable: %v", err)
}
return nil
```

(noting that this works with the other part because `errors.Is(nil, ErrSomething) == false`)

mcncl (approved):

PR #3907, opened 2026-05-10
fix: surface permission errors when checking plugin directories

ss1909:
### Description

Previously, osutil.FileExists silently returned false for any stat error, including permission denied. This caused the agent to attempt a re-clone of an already-checked-out plugin when the plugin directory was inaccessible. Now os.Stat is used directly so permission errors are returned immediately with a clear message, rather than falling through to a doomed re-clone.

### Context

https://linear.app/buildkite/issue/A-1065/plugin-code-does-not-distinguish-between-permission-errors-and-non

### Testing

### DIFF
--- a/internal/job/plugin.go
+++ b/internal/job/plugin.go
@@ -16,7 +16,6 @@ import (
 	"github.com/buildkite/agent/v3/agent/plugin"
 	"github.com/buildkite/agent/v3/internal/experiments"
 	"github.com/buildkite/agent/v3/internal/job/hook"
-	"github.com/buildkite/agent/v3/internal/osutil"
 	"github.com/buildkite/roko"
 	"github.com/buildkite/shellwords"
 )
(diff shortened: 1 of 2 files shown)


internal/job/plugin.go
```diff
@@ -400,6 +403,8 @@ func (e *Executor) checkoutPlugin(ctx context.Context, p *plugin.Plugin) (*plugi
 			return nil, fmt.Errorf("%q was not a directory within the %q plugin: %w", checkout.HooksDir, checkout.Plugin.Name(), err)
 		}
 		return checkout, nil
+	} else if !errors.Is(err, os.ErrNotExist) {
+		return nil, fmt.Errorf("checking plugin git directory %q: %w", pluginGitDirectory, err)
 	}
```
moskyb: probably worth swapping these around for readability — the happy path should be left-aligned

```Go
_, err := os.Stat(pluginGitDirectory)
if err != nil && !errors.Is(err, os.ErrNotExist) {
  return nil, ...
}

headCommit, err := ...
```

internal/job/plugin_zip.go
```diff
@@ -39,11 +39,13 @@ func (e *Executor) checkoutZipPlugin(ctx context.Context, p *plugin.Plugin, chec
...
 
 	// Remove existing directory if present, as of right now caching is not supported.
-	if osutil.FileExists(pluginDirectory) {
+	if _, err := os.Stat(pluginDirectory); err == nil {
 		if err := os.RemoveAll(pluginDirectory); err != nil {
 			e.shell.Errorf("Failed to remove existing plugin directory %s", pluginDirectory)
 			return err
 		}
+	} else if !errors.Is(err, os.ErrNotExist) {
+		return fmt.Errorf("checking zip plugin directory %q: %w", pluginDirectory, err)
```
moskyb: as above

moskyb (approved):
🎉

bdstar81 (approved):

PR #3908, opened 2026-05-11
GIT LFS Checkout feature 

lizrabuya:
- **Pave road to no-checkout-override**
- **Define checkout override scope for runtime env mutations**
- **Checkout override in executor**
- **Add unit tests for no-checkout-override**
- **Fix env set panic on Job API errors**
- **Add setCheckoutEnv**
- **Add integration tests**
- **Add comment re checkoutOverrideScope**
- **Set no-checkout-override=true when command-eval disabled**
- **Add test for no command eval enabling no checkout override**

### Testing


lizrabuya:
Closing to create a cleaner PR for GIT LFS Checkout changes

PR #3924, opened 2026-05-14
internal: [TE-5828] Add bktec plan step to collect commit metadata on agent builds

gchan:
## Description

Adds a discardable `bktec plan --collect-git-metadata` step to the
agent build pipeline so git commit metadata is sent to Buildkite Test
Engine on every build. The plan output is discarded -- existing test
steps continue to use their declared `parallelism:` values, so this is
a metadata-only side-channel with no run-time contract change.

`bktec` v2.5.0 is downloaded from the `test-engine-client` GitHub
release at step time. The agent repo pins `test-engine-client v1.6.0`
in `go.mod` for use as `go tool test-engine-client` in
`.buildkite/steps/tests.sh`, which predates the `plan` subcommand
(v2.0.0) and the `--collect-git-metadata` flag (v2.4.0). A coordinated
bump of both call sites is tracked as follow-up.

The step is `soft_fail: true`, has no `depends_on` and no dependents,
so a Test Engine API hiccup never blocks the build. It runs inside the
existing `agent` docker-compose service to pick up `go` on PATH (bktec
shells out to `go list ./...` for package discovery) and the
`BUILDKITE_TEST_ENGINE_API_ACCESS_TOKEN` env var.

## Context

Internal ticket: [TE-5828](https://linear.app/buildkite/issue/TE-5828).

## Changes

Best reviewed commit-by-commit:

- Add bktec plan script to collect git commit metadata
- Wire bktec metadata step into agent build pipeline
- Run bktec metadata step inside agent compose service
- Simplify metadata script for linux-only compose service

## Verification

The new `:test_tube: Collect commit metadata` step should:

- Print `bktec 2.5.0` after the download.
- Exit 0 with the log line `Plan request issued -- git commit metadata sent to Test Engine`.
- Not affect the Tests and Coverage group's timing or behaviour.

AI assisted.

## Deployment

Low risk. New step is `soft_fail: true` and has no dependents.

## Rollback

Yes -- single revert.

PR #3925, opened 2026-05-14
internal: Bump test-engine-client to v2.5.1 and unify bktec call sites

gchan:
## Description

This repo currently calls `bktec` from two places at two different versions: `.buildkite/steps/tests.sh` uses `go tool test-engine-client` at v1.6.0 (pinned in `go.mod`), and `.buildkite/steps/collect-commit-metadata.sh` (added in #3924) downloads the v2.5.0 binary from the GitHub release at step time. v1.6.0 predates the `plan` subcommand and the `--collect-git-metadata` flag, which is why the metadata step had to bring its own bktec.

This PR unifies both call sites onto a single `go tool test-engine-client` install pinned in `go.mod`, dropping the curl bootstrap and the runtime dependency on GitHub releases.

bktec v2.0.0 split the previous bare-command shape into `run` / `plan` / `tools` subcommands, so the runtime test path migrates from `go tool test-engine-client` to `go tool test-engine-client run` at the same time.

## Changes

Best reviewed commit-by-commit:

- Bump `test-engine-client` from v1.6.0 to v2.5.1 in `go.mod`.
- Migrate `tests.sh` to the v2 `run` subcommand shape.
- Switch the metadata step to `go tool test-engine-client plan` so both call sites share the same install.

## Verification

- `test-linux-amd64`, `test-linux-arm64`, `test-windows`, and `test-race-linux-arm64` keep passing and continue to emit the same `junit-*` artifacts.
- The metadata step keeps emitting `Plan request issued -- git commit metadata sent to Test Engine`.

## Deployment

Low risk. The runtime test contract is unchanged (same suite slug, runner, result path, retry count, test command -- only the entry-point shape moves from v1 to v2). The metadata step is `soft_fail: true` so any regression on that path doesn't block the build.

## Rollback

Yes -- revert the PR.


gchan:
This PR is waiting on a 2.5.1 release which addresses Go's [module versioning rules](https://go.dev/ref/mod#major-version-suffixes)

buildkate (approved):

PR #3930, opened 2026-05-19
fix: resume after job dispatch drops job

DrJosh9000:
## Description

Fix cases where previously dispatched jobs could be silently dropped.

## Context

[A-1266](https://linear.app/buildkite/issue/A-1266/low-agent-edge-debouncer-may-silently-drop-jobs)

## Changes

Debounce incoming messages more thoughtfully. In short:

* Preserve incoming jobs in a queue, so that changes in pause state don't drop them.
* Don't allow debouncing away from "disconnect"

## Testing

## Disclosures / Credits

I did not use any AI tools.


CerealBoy (approved):
LGTM 🚀 🚀

PR #3932, opened 2026-05-20
Fix error variable shadowing causing incorrect OTEL spans in hooks

calebfroese:
## Description

The observed problem is that the internals of [Tracing in the Buildkite agent](https://buildkite.com/docs/agent/v3/tracing#using-opentelemetry-tracing) incorrectly report `OK` spans that should be `ERROR` spans.

### Issue 1: Variable shadowing in child spans

| Before | After |
|--------|-------|
| broken | fixed |

The OTEL span capture is done with a predeclared `err` value accessed in a `defer` block, but not all error returning lines set the `error` that the `defer` block accesses, and are instead creating their own shadowed version of `err`.

Here's the problem in practice:

```go
func stuff() error {
  var err error
  defer func() { span.FinishWithError(err) }() // <-- this err is always nil, because
  if err := someFunction(); err != nil { // <-- err := here creates a new variable called err
    return err
  }
}
```

A fixed version of this code either looks like :

```go
func stuff() error {
  var err error
  if err = someFunction(); err != nil { // <-- note err = , not err :=
    return err
  }
}
```

or, equally valid

```go
func stuff() (err error) {
  if err = someFunction(); err != nil {
    return err
  }
}
```

### Issue 2: Command span ignoring command errors

| Before | After |
|--------|-------|
| broken | fixed |

The `command` span in `CommandPhase` was only finishing with `hookErr`, ignoring `commandErr`. When the command itself failed (not a pre/post hook), the span showed OK while its child `default command hook` span showed ERROR.

```go
func CommandPhase() (hookErr, commandErr error) {
    defer func() { span.FinishWithError(hookErr) }() // <-- ignores commandErr
    // ...
    commandErr = runCommand() // this error never reaches the span, so the span is OK
    return nil, commandErr
}
```

Fixed to check both:

```go
defer func() {
    if hookErr != nil {
        span.FinishWithError(hookErr)
    } else {
        span.FinishWithError(commandErr)
    }
}()
```

### Alternatives considered

I considered not using a named `err` return value, and instead continue to use `var err error`. The downside of this is that you need to remember at every error-returning line to set the variable:

```go
func stuff() (err error) {
  var err error
  // ...
  err = fmt.Errorf("something bad: %w", err)
  return err
}
```

Failure to do the above will result in a mismatch between the function returned error and the span finish with error.

It's far more ergonomic to simply have the following, and then you don't have to worry about forgetting to set `err`.

```go
func stuff() (err error) {
  return fmt.Errorf("something bad: %w", err)
}
```

## Context

N/A

## Changes

- Fix variable shadowing so child spans (hooks, checkout, etc.) show `ERROR` when they fail
- Fix `command` span to reflect command errors, not just hook errors
- Rename non-fatal non-OTEL-capturing errors for clarity, e.g. [here](https://github.com/buildkite/agent/pull/3932/changes#diff-ab2257bcff4d75670c1e896bba230e26db13910cc84f23621d7528518d211c48R178-R181) and [here](https://github.com/buildkite/agent/pull/3932/changes#diff-ab2257bcff4d75670c1e896bba230e26db13910cc84f23621d7528518d211c48R967-R970). This isn't a change in behavior, but the now renamed non-shadowed errors makes it clear they are unrelated to our span capture `error` variable. This is correct (and unchanged) behavior per the OTEL spec that these non-fatal errors should not capture exceptions:

> An exception SHOULD be recorded as an Event on the span during which it occurred if and only if it remains unhandled when the span ends and causes the span status to be set to ERROR.

## Testing
- Have build the binary locally and tested. Expand for how to reproduce the testing I did

<details>
<summary>End-to-end local validation process (expand me)</summary>

````bash
# Start Jaeger
docker run -d --name jaeger -p 4543:16686 -p 4344:4317 jaegertracing/all-in-one:latest

# Create failing hook
mkdir -p /tmp/bk-test/hooks /tmp/bk-test/builds /tmp/bk-test/checkout
echo -e '#!/bin/bash\nexit 1' > /tmp/bk-test/hooks/pre-command
chmod +x /tmp/bk-test/hooks/pre-command

# Build both versions
git checkout main && go build -o /tmp/agent-broken .
git checkout refactor-error-handling-for-tracing && go build -o /tmp/agent-fixed .

# Run broken (span shows OK)
OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4344 BUILDKITE_BUILD_CHECKOUT_PATH=/tmp/bk-test/checkout \
  /tmp/agent-broken bootstrap --command "echo hi" --job broken --repository x --commit HEAD \
  --branch main --build-path /tmp/bk-test/builds --hooks-path /tmp/bk-test/hooks \
  --phases command --tracing-backend opentelemetry --tracing-service-name broken \
  --agent a --organization o --pipeline p --pipeline-provider github

# Run fixed (span shows ERROR)
OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4344 BUILDKITE_BUILD_CHECKOUT_PATH=/tmp/bk-test/checkout \
  /tmp/agent-fixed bootstrap --command "echo hi" --job fixed --repository x --commit HEAD \
  --branch main --build-path /tmp/bk-test/builds --hooks-path /tmp/bk-test/hooks \
  --phases command --tracing-backend opentelemetry --tracing-service-name fixed \
  --agent a --organization o --pipeline p --pipeline-provider github

# View traces at http://localhost:4543
# Compare "agent pre-command hook" span - broken has no error, fixed has otel.status_code=ERROR
````

</details>

http://github.com/Canva

## Disclosures / Credits

I identified the problem and an opportunity to use named return values, then used Claude Opus 4.5 to implement and validate the desired fix.

### DIFF
--- a/internal/job/checkout.go
+++ b/internal/job/checkout.go
@@ -131,9 +131,8 @@ func (e *Executor) refreshCheckoutRoot() error {
 
 // CheckoutPhase creates the build directory and makes sure we're running the
 // build at the right commit.
-func (e *Executor) CheckoutPhase(ctx context.Context) error {
+func (e *Executor) CheckoutPhase(ctx context.Context) (err error) {
 	span, ctx := tracetools.StartSpanFromContext(ctx, "checkout", e.TracingBackend)
-	var err error
 	defer func() { span.FinishWithError(err) }()
 
 	if err = e.executeGlobalHook(ctx, "pre-checkout"); err != nil {
@@ -168,17 +167,17 @@ func (e *Executor) CheckoutPhase(ctx context.Context) error {
 	}
 
 	// Make sure the build directory exists
-	if err := e.createCheckoutDir(); err != nil {
+	if err = e.createCheckoutDir(); err != nil {
 		return err
 	}
 
-	if err := e.checkout(ctx); err != nil {
+	if err = e.checkout(ctx); err != nil {
 		return err
 	}
 
-	err = e.sendCommitToBuildkite(ctx)
-	if err != nil {
-		e.shell.OptionalWarningf("git-commit-resolution-failed", "Couldn't send commit information to Buildkite: %v", err)
+	// sendCommitToBuildkite is non-critical, don't propagate to span
+	if sendErr := e.sendCommitToBuildkite(ctx); sendErr != nil {
+		e.shell.OptionalWarningf("git-commit-resolution-failed", "Couldn't send commit information to Buildkite: %v", sendErr)
 	}
 
 	// Store the current value of BUILDKITE_BUILD_CHECKOUT_PATH, so we can detect if
@@ -189,15 +188,15 @@ func (e *Executor) CheckoutPhase(ctx context.Context) error {
 	}
 
 	// Run post-checkout hooks
-	if err := e.executeGlobalHook(ctx, "post-checkout"); err != nil {
+	if err = e.executeGlobalHook(ctx, "post-checkout"); err != nil {
 		return err
 	}
 
-	if err := e.executeLocalHook(ctx, "post-checkout"); err != nil {
+	if err = e.executeLocalHook(ctx, "post-checkout"); err != nil {
 		return err
 	}
 
-	if err := e.executePluginHook(ctx, "post-checkout", e.pluginCheckouts); err != nil {
+	if err = e.executePluginHook(ctx, "post-checkout", e.pluginCheckouts); err != nil {
 		return err
 	}
 
@@ -208,7 +207,7 @@ func (e *Executor) CheckoutPhase(ctx context.Context) error {
 	if previousCheckoutPath != "" && previousCheckoutPath != newCheckoutPath {
 		e.shell.Headerf("A post-checkout hook has changed the working directory to \"%s\"", newCheckoutPath)
 
-		if err := e.shell.Chdir(newCheckoutPath); err != nil {
+		if err = e.shell.Chdir(newCheckoutPath); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: internal/job/checkout.go [hunk 1]
```diff
@@ -131,9 +131,8 @@ func (e *Executor) refreshCheckoutRoot() error {
 
 // CheckoutPhase creates the build directory and makes sure we're running the
 // build at the right commit.
-func (e *Executor) CheckoutPhase(ctx context.Context) error {
+func (e *Executor) CheckoutPhase(ctx context.Context) (err error) {
```
aeijdenberg-canva: I'm struggling to find a canonical reference for it, but a pattern I've seen in other codebases (such as podman) is to use the name `retErr` for the named return error value, then only ever refer to it within the `defer` statement. Often this is done to capture the error from closing a file, for example: https://github.com/containers/podman/blob/6ef4d68da48dcc6575eab38875096efc2cba6078/pkg/bindings/artifacts/extract.go#L100-L120

That way the rest of the code can happily look normal and use `if err := xxx; err != nil { ... }` pattern without needing to worry about masking anything (and would mean you could touch many fewer lines in this PR).

internal/job/executor.go
```diff
@@ -134,10 +134,10 @@ func (e *Executor) Run(ctx context.Context) (exitCode int) {
 		e.shell = sh
 	}
 
-	var err error
+	var retErr error
```
aeijdenberg-canva, line 137: This bit breaks the pattern... (e.g. `retErr` is referred to and modified outside of the simple `defer`)  I wonder if we can wrap the section that the span applies to in it's own function, then we can use the same pattern?

aeijdenberg-canva (commented):
Look good to me except for the one funky bit that breaks the pattern...

internal/job/checkout.go
calebfroese: Good call 👍 I've done some refactoring to use that pattern, reduces the diff a lot and makes it easier, nice!

internal/job/executor.go
aeijdenberg-canva: Also if this same pattern is repeated a bunch of times, might be easy enough to make a simple wrapper function that Does The Right Thing for all of them?

calebfroese: The function signature here is an exit code integer, not an error

https://github.com/calebfroese/agent/blob/b0c9c90550d7935beefaafb4ea3059f4551f27ef/internal/job/executor.go#L98

It could alternatively just remain unchanged as `err` to reduce the PR diff size, but then you have the "did I remember to set `err` and not shadow it" problem again. The name `retErr` gives more consistency imo

Will leave wrapper idea for buildkite maintainers to comment on if desired. My 2c it's simple enough with a named error that a wrapper wouldn't justify its value

aeijdenberg-canva: SGTM.

aeijdenberg-canva (approved):

DrJosh9000 (approved):
Thanks @calebfroese (and @aeijdenberg-canva)! 

I'm happy to proceed without the wrapper for Run, but that's a nice idea, might do that later.

Good work, LGTM ✅ ✅

PR #3934, opened 2026-05-21
Update versioned brew formula based on version being released

DrJosh9000:
## Description

On release, update the versioned Brew formulas. 

## Context

Updating the original formula won't work following buildkite/homebrew-buildkite#54.

## Changes

The key changes are

```diff
-FORMULA_FILE=./pkg/buildkite-agent.rb
-UPDATED_FORMULA_FILE=./pkg/buildkite-agent-updated.rb
+BASE_FORMULA_FILE="buildkite-agent@${AGENT_VERSION%%.*}.rb"
+FORMULA_FILE="./pkg/${BASE_FORMULA_FILE}"
+UPDATED_FORMULA_FILE=./pkg/${BASE_FORMULA_FILE}.updated
```

and similarly using $BASE_FORMULA_FILE within URLs to fetch/write to GitHub.

The remainder are shell style tweaks.

## Testing

## Disclosures / Credits

I did not use AI for this.

### DIFF
--- a/.buildkite/steps/release-homebrew.sh
+++ b/.buildkite/steps/release-homebrew.sh
@@ -19,107 +19,116 @@ function to_json {
 
 echo '--- Getting agent version from build meta data'
 
-export FULL_AGENT_VERSION=$(buildkite-agent meta-data get "agent-version-full")
-export AGENT_VERSION=$(buildkite-agent meta-data get "agent-version")
-export BUILD_VERSION=$(buildkite-agent meta-data get "agent-version-build")
-export IS_PRERELEASE=$(buildkite-agent meta-data get "agent-is-prerelease")
+export FULL_AGENT_VERSION="$(buildkite-agent meta-data get "agent-version-full")"
+export AGENT_VERSION="$(buildkite-agent meta-data get "agent-version")"
+export BUILD_VERSION="$(buildkite-agent meta-data get "agent-version-build")"
+export IS_PRERELEASE="$(buildkite-agent meta-data get "agent-is-prerelease")"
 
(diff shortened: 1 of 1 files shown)


.buildkite/steps/release-homebrew.sh
```diff
@@ -19,115 +19,124 @@ function to_json {
...
-echo "Release SHA256: $RELEASE_SHA256_ARM64"
+echo "Release SHA256: ${RELEASE_SHA256_ARM64}"
 
 echo "--- :octocat: Fetching current Homebrew formula from GitHub Contents API"
 
-FORMULA_FILE=./pkg/buildkite-agent.rb
-UPDATED_FORMULA_FILE=./pkg/buildkite-agent-updated.rb
+BASE_FORMULA_FILE="buildkite-agent@${AGENT_VERSION%%.*}.rb"
+FORMULA_FILE="./pkg/${BASE_FORMULA_FILE}"
+UPDATED_FORMULA_FILE=./pkg/${BASE_FORMULA_FILE}.updated
```
CerealBoy, line 93: This is going to yield `./pkg/buildkite-agent@3.rb.updated` as an example, is that intended?

DrJosh9000: That should be fine. The "updated" file should only exist on disk locally. It gets read back into `UPDATED_FORMULA_BASE64`, which is incorporated into `pkg/github_post_data.json`, which is the POST body back to GitHub.

CerealBoy (approved):
LGTM 🚀 🚀 

Big fan of the cleaning up too!

PR #3935, opened 2026-05-21
fix: nil deref if HTTP/2 is disabled

DrJosh9000:
## Description

Fix potential nil dereference when disabling HTTP/2.

## Context

https://slopcannon.tail952194.ts.net/lachlan/deepsec-buildkite-agent/HIGH_BUG/agent-other-logic-bug-ec621ce249.md

## Changes

Ensure `TLSClientConfig` is not nil before setting `NextProtos`. Creating a blank `tls.Config` is fine, it's what Go itself does (see https://cs.opensource.google/go/go/+/refs/tags/go1.26.3:src/net/http/h2_bundle.go;l=7537) 

## Testing

## Disclosures / Credits

The bug was found with an LLM (deepsec), I made the fix myself.

### DIFF
--- a/internal/agenthttp/client.go
+++ b/internal/agenthttp/client.go
@@ -102,6 +102,10 @@ func newTransport(conf *clientConfig) *http.Transport {
 		}
 	} else {
 		transport.TLSNextProto = make(map[string]func(string, *tls.Conn) http.RoundTripper)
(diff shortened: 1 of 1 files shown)


wolfeidau (approved):
👍🏻 🧹

PR #3938, opened 2026-05-21
Add SSH config and env var

mcncl:
## Description

This allows for folks to set the location of an SSH key, with a view to later support that setting via YAML.

## Context

Part of the agent checkout improvements project.

## Changes

- add bootstrap/env wiring for BUILDKITE_GIT_SSH_KEY
- add GitSSHKey to internal/job.ExecutorConfig
- update default checkout to:
- create a private temp directory adjacent to the checkout path
- write the SSH key to id with 0600 permissions
- normalize the key to a trailing newline
- set GIT_SSH_COMMAND=ssh -i -o IdentitiesOnly=yes
- restore any previous GIT_SSH_COMMAND
- remove the temp key directory on success/failure
- add *_SSH_KEY to default redaction patterns
- add unit/integration coverage for key creation, permissions, cleanup, and checkout behavior

## Testing

Ran locally:

```
go test ./...
```

## Disclosures / Credits

Used AI to highlight any shortcomings or lacking test coverage; it fixed the initial issue with the tests running on Windows where explicitly closing the checkoutRoot was required.

I also used it to help write this description; scanning the files for changes to remind me what I'd done.


clicommand/bootstrap.go
```diff
@@ -489,7 +496,8 @@ var BootstrapCommand = cli.Command{
 		defer cancel()
 
 		signals := make(chan os.Signal, 1)
-		signal.Notify(signals, os.Interrupt,
+		signal.Notify(
```
mcncl, line 507: This isn't really a change, it's just a result of using `gofumpt` for my formatting rather than `gofmt`, I'm happy to default to the latter, though

internal/job/checkout.go
```diff
@@ -838,6 +838,23 @@ func (e *Executor) defaultCheckoutPhase(ctx context.Context) (retErr error) {
...
+	}
+	if cleanupSSHKey != nil {
+		defer func() {
+			if cleanupErr := cleanupSSHKey(); cleanupErr != nil {
+				cleanupErr = fmt.Errorf("cleaning up git ssh key %q: %w", sshKeyPath, cleanupErr)
+				if err == nil {
+					err = cleanupErr
+				} else {
+					err = errors.Join(err, cleanupErr)
+				}
```
DrJosh9000: `err` has (meanwhile) become `retErr`, and `errors.Join` does the right thing with `nil` (i.e. drops it), so:

```suggestion
				retErr = errors.Join(retErr, cleanupErr)
```

internal/job/checkout.go
```diff
@@ -1065,6 +1082,77 @@ func (e *Executor) defaultCheckoutPhase(ctx context.Context) (retErr error) {
...
+		return "", nil, fmt.Errorf("creating ssh key parent directory %q: %w", parentDir, err)
+	}
+
+	pattern := ".buildkite-ssh-key-"
+	if e.PipelineSlug != "" {
+		pattern += badCharsRE.ReplaceAllString(e.PipelineSlug, "-") + "-"
+	}
+
+	// os.MkdirTemp creates the directory with mode 0o700 on Unix, giving the
+	// key file inside an additional layer of protection from sibling builds.
```
DrJosh9000: Sibling builds would have to be running _under a different uid_ to be prevented from accessing inside a dir with mode 0o700, right?

internal/job/checkout.go
```diff
@@ -1065,6 +1082,77 @@ func (e *Executor) defaultCheckoutPhase(ctx context.Context) (retErr error) {
...
+// prepareGitSSHKey materialises the configured GitSSHKey into a private
+// directory next to the build checkout and points GIT_SSH_COMMAND at it for
+// the duration of the checkout phase. It returns the path to the key file, a
+// cleanup function that removes the key and restores any previous
+// GIT_SSH_COMMAND, and an error. If no key is configured both the path and
+// cleanup are zero.
+//
+// Only the default checkout phase invokes this; custom checkout hooks must
+// arrange their own credentials.
+func (e *Executor) prepareGitSSHKey() (_ string, _ func() error, retErr error) {
```
DrJosh9000: Naming the return args has clear readability and documentation upsides over using `_`:

```suggestion
func (e *Executor) prepareGitSSHKey() (sshKeyPath string, cleanup func() error, retErr error) {
```

quinn-diesel:
rather than tag Buildsworth in this -  i did some side investigating it's looking good. approving and can always fix up if there are any issues

quinn-diesel (approved):

PR #3940, opened 2026-05-21
fix: remove eval from install.sh

DrJosh9000:
## Description

Remove a HackerOne magnet (and prevent command injections that depend on how you hold the script).

## Context

https://linear.app/buildkite/issue/A-1267

## Changes

Replace `eval` calls with fixed commands stored as arrays (because)

## Testing

## Disclosures / Credits

I did it

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -72,9 +72,9 @@ else
 fi
 
 if command -v curl >/dev/null 2>&1 ; then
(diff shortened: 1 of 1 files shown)


CerealBoy (approved):
LGTM 🚀 🚀

PR #3941, opened 2026-05-22
Document step parallelism updates

catkins:
## Description

Adds a `buildkite-agent step update` help example for updating a step's `parallelism` attribute.

The agent itself did not need a behavior change. `step update` already accepts arbitrary step update attributes and sends them through the Agent API; this PR only updates the command help doc-string so the supported usage is easier to discover.

## Verification

```sh
go test ./clicommand
```


DrJosh9000:
Hey @catkins just wondering if this is ready for review or still needed?

PR #3957, opened 2026-05-26
internal: Add comments about env files

DrJosh9000:
## Description

Make it easier to understand what `BUILDKITE_ENV_FILE` is for.

## Context

https://linear.app/buildkite/issue/A-1236

## Changes

Add some comments.

## Testing

## Disclosures / Credits

These are just some words I wrote

### DIFF
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -491,6 +491,15 @@ BUILDKITE_AGENT_JWKS_KEY_ID`
 		}
 
 		for key, value := range env {
+			// Note that the value below is %q-quoted, and not shell-escaped.
+			// Env files are designed to be validated by the pre-bootstrap hook.
+			// Because the job env may contain untrusted or even dangerous env
+			// vars (suppose a user adds an env var in the "New Build" UI with a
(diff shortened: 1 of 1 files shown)


agent/job_runner.go
```diff
@@ -775,7 +784,11 @@ func (r *JobRunner) executePreBootstrapHook(ctx context.Context, hook string) (b
 
 	// This (plus inherited) is the only ENV that should be exposed
 	// to the pre-bootstrap hook.
-	// - Env files are designed to be validated by the pre-bootstrap hook
+	// - Env files are designed to be validated by the pre-bootstrap hook.
+	//   Because the job env may contain untrusted or even dangerous env vars,
```
zhming0: Should we give an example here?

zhming0 (approved):
Looks good with a suggestion 👍🏿

agent/job_runner.go
DrJosh9000: I've added some more words in parentheses that explain how such an env var might be added, and how it could be dangerous, but I would rather not be overly specific.

PR #3958, opened 2026-05-26
PF-9591-traceset added  end-to-end W3C tracestate propagation to the agent

SorchaAbel:
### Description

Completes the end-to-end W3C `tracestate` propagation started in buildkite/buildkite#29329, #29355, #29474. Reads the `tracestate` field the control plane now sends on the job acquire/accept payload, threads it through the agent → bootstrap → executor, and attaches it to OpenTelemetry spans alongside the existing `traceparent`.

Once this ships and is deployed, agent-emitted spans (`job.run` and descendants) will carry the same tracestate the control plane stamps on `buildkite.build` / `buildkite.job` spans, completing the customer-configured trace context (e.g.  `oai=t.oai_ft:1` retention bit) across the full backend + agent trace.

There is **no impact on existing customers**:
- Orgs without tracestate configured → payload omits the field → agent sets nothing → spans unchanged.
- Orgs with tracestate configured but agent not opted into `--tracing-propagate-traceparent` → value is plumbed through but never extracted, same as `traceparent` today.
- Only when the server provides a tracestate **and** the operator has opted into traceparent propagation does it get attached to spans.

### Context

- Linear: [PF-9591]
- Server side: buildkite/buildkite#29329, #29355, #29474
- The agent currently silently drops the `tracestate` field on the JSON payload because there's no field on the `Job` struct to decode it into. This PR adds that field and wires it through the existing tracing plumbing.

### Changes

Six files, structurally identical to the existing `TraceParent` flow:

- **`api/jobs.go`** — adds `TraceState string` (`json:"tracestate"`) to the `Job` struct...
- **`agent/job_runner.go`** — when `Job.TraceState` is non-empty, sets `BUILDKITE_TRACING_TRACESTATE`...
- **`clicommand/bootstrap.go`** — adds `TracingTraceState` config field...
- **`internal/job/config.go`** — adds `TracingTraceState string` to `ExecutorConfig`.
- **`internal/job/tracing.go`** — in `contextWithTraceparentIfEnabled`, includes `"tracestate"` in the `propagation.MapCarrier` when set.
- **`internal/job/tracing_test.go`** — unit tests for `contextWithTraceparentIfEnabled` covering propagation disabled, empty traceparent, traceparent only, and traceparent + tracestate.

### Verification

- `go build ./...` — clean.
- `go test ./api/... ./agent/... ./internal/job/... ./clicommand/...` — all pass, including the new `TestContextWithTraceparentIfEnabled_*` cases that confirm the OTel span context carries the configured tracestate end-to-end

End-to-end manual verification once deployed: configure a tracestate in OTel service settings on a test org, run a build with an agent running this branch, confirm the agent-emitted `job.run` (OTel) span carries the same `tracestate` value as the server-emitted `buildkite.job` span — using the same local OTLP debug collector setup from buildkite/buildkite#29329 / #29355.

### Deployment

Safe.

- Backward compatible for orgs without OTel tracestate configured (no field present in payload → `Job.TraceState` defaults to `""` → no env var, no carrier entry, no span change).
- Backward compatible for operators not opted into `--tracing-propagate-traceparent` (same opt-in gate as `traceparent` today — value is plumbed through but never read by the propagator).
- No new dependencies, no flag default changes, no config-file changes required.

### Rollback

Safe to revert.

### DIFF
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -480,6 +480,7 @@ BUILDKITE_TRACE_CONTEXT_ENCODING
 BUILDKITE_TRACING_BACKEND
 BUILDKITE_TRACING_SERVICE_NAME
 BUILDKITE_TRACING_TRACEPARENT
+BUILDKITE_TRACING_TRACESTATE
 BUILDKITE_TRACING_PROPAGATE_TRACEPARENT
 BUILDKITE_AGENT_AWS_KMS_KEY
 BUILDKITE_AGENT_GCP_KMS_KEY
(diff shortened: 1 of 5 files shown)


DrJosh9000 (approved):
LGTM - small focused PR that looks correct ✅

PR #3959, opened 2026-05-27
A-1270 part 3: refactor cache package tighten seperation of concern

zhming0:
## Description

This change refactors cache package so we get:

* to reduce the public surface
* better separation of concern. 

Overall hopefully make things more maintainable.

None of these code are used by users yet, so really low risk, buildsworth if you are reading this.

## Context

part of A-1270

## Changes

Changes
- One config struct. Merged ClientConfig into Config; the library-only knobs (Format, Platform, Registry, OnProgress, Env, Caches) are now internal defaults instead of caller responsibilities.
- Unexported the handle. Client → client, NewClient → newClient, CacheClient interface → unexported cacheOps. The package now exports only what clicommand consumes: Save, Restore, Config, result types, and sentinel errors.
- One construction path. setupCacheClient folded into newClient; API client creation, config load + expand + validate, and cache-ID resolution all happen in one place.
- Config loading lives with config types. loadCacheConfiguration + FileConfig moved out of cache.go and into internal/cache/configuration as configuration.LoadFile(path) ([]Cache, error). The on-disk YAML wrapper struct is inlined and no longer leaks out of the loader.
- Tightened doc comments explaining that client is a configured handle (API client + bucket + expanded caches), not a network connection.

## Testing

## Disclosures / Credits

AI and AI 
Actual and Artificial Intelligence


zhming0:
@buildsworth-bk show up and do review works!

buildkate (approved):

PR #3960, opened 2026-05-27
A-1270 part 4: consolidate cache and agent API layer

zhming0:
## Description

Folds the cache HTTP client into the existing top-level api package, replacing the bespoke transport (originally from zstash) with the agent's standard one.

### Why?

- Reuse the token and header logic -> this is very important so we get to respect sharding. 
- Consistent transport stack with rest of the code (timeout, etc).
- No more `zstash/version` User-Agent

## Context

part of A-1270

## Changes

- Moved most of api code from `cache/api/*` to `api/cache.go`.
- make the `cache_save` and `cache_restore` consistent with other subcommands. 

## Testing

## Disclosures / Credits

LLM and brain.


api/cache.go
```diff
@@ -0,0 +1,347 @@
...
+	err := fmt.Errorf(format, args...)
+	span.RecordError(err)
+	span.SetStatus(codes.Error, err.Error())
+	return err
+}
+
+// isJSONContent reports whether contentType represents JSON, including media
+// types with suffix (e.g. application/problem+json) or parameters (e.g.
+// application/json; charset=utf-8).
+func isJSONContent(contentType string) bool {
```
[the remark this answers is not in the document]
zhming0: ported over the old test coverage.

zhming0:
@buildsworth-bk review again?

api/cache.go
```diff
@@ -0,0 +1,346 @@
...
+
+// ErrCacheEntryNotFound is reserved for callers that want a sentinel for the
+// "no entry" condition. The cache methods themselves report it via the
+// (resp, exists, err) return shape; this value is exported for parity.
+var ErrCacheEntryNotFound = errors.New("cache entry not found")
+
+var cacheTracer = otel.Tracer("github.com/buildkite/agent/v3/api/cache")
+
+// CacheCreateReq is the request body for creating a cache entry.
+type CacheCreateReq struct {
```
buildkate, line 36: [nonblocking] Looking at the different endpoints here, one operates on the registry and four operate on the cache entry itself.  I'm wondering if we should be more explicit with naming  so `CacheCreate` -> `CacheEntryCreate` etc.

buildkate (approved):

api/cache.go
zhming0: Good idea, I implemented this change in another PR: https://github.com/buildkite/agent/pull/3961

PR #3962, opened 2026-05-27
Retry git checkout when per-attempt timeout fires

scadu:
## Description

`git-checkout-timeout` (added in #3794) is supposed to retry, but on timeout the retry never happens. The per-attempt context cancellation kills git with SIGTERM, which returns `exec.ExitError` with `ExitCode() == -1`. 
The retrier switch in `checkout()` matched that as a generic signal interruption and called `r.Break()`, so the loop exited after one attempt regardless of `--checkout-attempts`.

Fix: join a sentinel `errCheckoutAttemptTimedOut` into the error from `runDefaultCheckoutAttempt` on deadline, and match it in the switch before the signal-kill case. The new branch cleans the checkout dir and returns the error so the retrier can back off and try again.

## Context

https://linear.app/buildkite/issue/PS-2047

## Changes

- `internal/job/checkout.go`: add `errCheckoutAttemptTimedOut` sentinel; join it into the timeout error returned by `runDefaultCheckoutAttempt`, add a matching branch in the retrier switch before the signal interruption case that cleans the checkout dir and falls through to retry.

## Testing

Verified with a hanging git endpoint (e.g.: `socat -u TCP-LISTEN:9418,reuseaddr,fork EXEC:'sleep 3600'` plus `--git-checkout-timeout 30 --checkout-attempts 3`: three clone attempts with backoff, not one.

## Disclosures / Credits

[Club Mate](https://www.club-mate.de/en/) helped me in bringing this PR to life.

### DIFF
--- a/internal/job/checkout.go
+++ b/internal/job/checkout.go
@@ -213,10 +213,16 @@ func (e *Executor) CheckoutPhase(ctx context.Context) (retErr error) {
 	return nil
 }
 
+// errCheckoutAttemptTimedOut is a sentinel marker, joined into the error
+// returned by runDefaultCheckoutAttempt when the per-attempt timeout fires.
+// The retry loop matches it explicitly so a timeout-killed git process is
+// retried instead of being treated like a user signal.
+var errCheckoutAttemptTimedOut = errors.New("checkout attempt timed out")
+
 // runDefaultCheckoutAttempt runs defaultCheckoutPhase, applying a per-attempt
-// timeout if BUILDKITE_GIT_CHECKOUT_TIMEOUT is set. On timeout the underlying
-// error is wrapped so the retry loop's generic failure path can log a
-// meaningful message and clean up before retrying.
+// timeout if BUILDKITE_GIT_CHECKOUT_TIMEOUT is set. On timeout the returned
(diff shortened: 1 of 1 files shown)


CerealBoy (approved):
LGTM 🚀 🚀

internal/job/checkout.go
```diff
@@ -228,7 +234,8 @@ func (e *Executor) runDefaultCheckoutAttempt(ctx context.Context) error {
 
 	err := e.defaultCheckoutPhase(attemptCtx)
 	if err != nil && attemptCtx.Err() == context.DeadlineExceeded && ctx.Err() == nil {
-		return fmt.Errorf("checkout attempt timed out after %s: %w", timeout, err)
+		return errors.Join(errCheckoutAttemptTimedOut,
+			fmt.Errorf("checkout attempt timed out after %s: %w", timeout, err))
```
DrJosh9000: You could use multiple `%w` verbs in `fmt.Errorf` for a similar effect and a less repetitive, more compact error message:

```suggestion
		return fmt.Errorf("%w after %s: %w", errCheckoutAttemptTimedOut, timeout, err)
```

PR #3963, opened 2026-05-28
A-1320 retry cache commands

buildkate:
## Description
Adds retry coverage to the cache CLI commands (`cache restore` and `cache save`), matching the consolidated retry pattern from `api/retryable.go` and sibling commands. Each API call is wrapped individually with `roko.NewRetrier(...).DoWithContext(...)` + `api.BreakOnNonRetryable` so transient 429/5xx/network errors retry with backoff while non-retryable 4xx breaks immediately.
## Context
- Linear: [A-1320](https://linear.app/buildkite/issue/A-1320/add-retry-coverage-to-cache-restore-and-cache-save-cli-commands)
## Changes
- All five cache API methods now return `*api.Response` (positioned second-to-last to match the sibling convention in `api/`) so callers can drive `BreakOnNonRetryable` decisions; `cacheAPI` interface and `mockAPIClient` updated in lockstep.
- Wrapped each individual cache API call in `internal/cache/restore.go` and `internal/cache/save.go` with retriers using the `meta_data_get` config (`WithMaxAttempts(10)`, `Constant(5*time.Second)`, `WithJitter()`).
- Verified `CacheEntryCreate` and `CacheEntryCommit` are retry-safe by reading the server-side code in `buildkite/buildkite`; short comments at each wrap site cite the verified server behaviour.
## Key decisions
- **Retry lives at the per-API-call boundary inside `internal/cache`, not at the CLI or as one outer wrap around `Save`/`Restore`** — outer-wrapping would replay already-succeeded caches in the fan-out and redo expensive non-API work (archive build, upload) when only the final API call failed.
- **Closure-capture pattern over `roko.DoFunc[N]`** — matches the 28 other retry call sites in the codebase and sidesteps the `DoFunc3` arity ceiling.
## Testing
## Disclosures / Credits
Used opencode (Claude) to walk the codebase, evaluate retry-scoping alternatives, verify server-side idempotency, and write the implementation. All decisions and final code reviewed by me.

### DIFF
--- a/internal/cache/restore.go
+++ b/internal/cache/restore.go
@@ -15,6 +15,7 @@ import (
 	"github.com/buildkite/agent/v3/api"
 	"github.com/buildkite/agent/v3/internal/cache/archive"
 	"github.com/buildkite/agent/v3/internal/cache/store"
+	"github.com/buildkite/roko"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
@@ -34,6 +35,11 @@ import (
 //
 // The operation respects context cancellation and will stop immediately when
 // ctx is cancelled, cleaning up any temporary resources (downloaded archives).
+//
+// Transient API errors (429, 5xx, connection reset/timeout/EOF) on
+// CacheEntryRetrieve are retried automatically per api.BreakOnNonRetryable.
+// Non-retryable errors (4xx other than 429) cause the operation to fail
+// immediately.
 //
 // Progress callbacks (if configured) are invoked at each stage with the
 // following stages: "validating", "checking_exists", "downloading", "extracting",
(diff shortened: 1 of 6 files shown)


internal/cache/restore.go
```diff
@@ -94,11 +100,31 @@ func (c *client) Restore(ctx context.Context, cacheID string) (RestoreResult, er
...
-		FallbackKeys: strings.Join(cacheConfig.FallbackKeys, ","),
+	var (
+		apiResp      *api.Response
+		retrieveResp api.CacheEntryRetrieveResp
+		exists       bool
+	)
+
+	err = roko.NewRetrier(
+		roko.WithMaxAttempts(10),
+		roko.WithStrategy(roko.Constant(5*time.Second)),
```
zhming0: Would it be better to use exponential instead? Cache restore is latency sensitive, 5s cache restore might as well be considered cache miss 😬 .

zhming0 (approved):
Looks good to me with one suggestion on using exponential retry strategy instead.

internal/cache/restore.go
buildkate: I lowered retry count to 5, and applied exponential strategy, so max 5 retries will be ~3.5 seconds.

Also applied to the API calls in save for consistency.

internal/cache/save.go
```diff
@@ -93,9 +100,29 @@ func (c *client) Save(ctx context.Context, cacheID string) (SaveResult, error) {
 	c.callProgress(cacheID, "checking_exists", "Checking if cache already exists", 0, 0)
 
 	// Check if cache already exists
-	_, exists, err := c.api.CacheEntryPeekExists(ctx, c.registry, api.CacheEntryPeekReq{
-		Key:    cacheConfig.Key,
-		Branch: c.branch,
+	var (
+		peekApiResp *api.Response
```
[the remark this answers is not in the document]
buildkate: noted

internal/cache/restore.go
zhming0: [non-blocking] I don't have a hard evidences against 3.5s total wait time, but it's worth keep in mind that it's better to have a green build than red build. 

A red build usually result in support tickets

PR #3967, opened 2026-05-29
refactor: replace interface{} with any (Go 1.18+)

MD-Mushfiqur123:
Replaces `interface{}` with the `any` alias (available since Go 1.18) in `internal/cache/configuration/configuration.go`.

This is the only file in the repo still using the old `interface{}` type literal.


DrJosh9000:
Thanks @MD-Mushfiqur123, but your code change doesn't do what the PR description says.

PR #3968, opened 2026-05-29
feature: Add Job API + CLI support for setting the working directory from polyglot hooks

francoiscampbell:
## Description

Today only wrapped POSIX shell hooks can change the working directory for subsequent hooks/commands — the hook wrapper exports `BUILDKITE_HOOK_WORKING_DIR=$PWD` after sourcing the hook and the executor `chdir`s to it.

Binary and polyglot (non-shell) hooks can't do this. They run via `runUnwrappedHook` in a child process, so any `chdir()` inside the child is lost.

This PR gives unwrapped hooks a first-class way to change the working directory for subsequent phases, via a new Job API endpoint and a matching `buildkite-agent workdir set <path>` CLI command.

**How it works:**
- The CLI resolves relative paths against its own cwd (the hook's actual working directory) and validates the path exists and is a directory.
- The Job API endpoint (`PUT /workdir`) accepts only absolute paths and records the request as pending (it does not stat the filesystem).
- Because the Job API handler runs in the server goroutine while `shell.wd` is single-writer, the endpoint only *records* the request. The executor consumes the pending workdir after the hook process exits and applies it through the existing `applyEnvironmentChanges` → `shell.Chdir` path — single-threaded, consistent with the wrapped-hook path.
- The change persists across subsequent hooks and the command phase via `shell.wd`; it is not reset between hooks.

Scope is intentionally the unwrapped hook path. Wrapped POSIX shell hooks continue to use the wrapper's `$PWD` capture and are unaffected.

## Context

Internal Linear ticket (Buildkite).

## Changes

- **Job API**: new `PUT /api/current-job/v0/workdir` endpoint (`jobapi/workdir.go`), `WorkdirSetRequest`/`WorkdirSetResponse` payloads, `Server.pendingWorkdir` + take-on-read `TakePendingWorkdir`, and `Client.SetWorkdir`.
- **Executor**: retain the `*jobapi.Server` on the `Executor`; `runUnwrappedHook` now consumes any pending workdir and applies it after the hook exits. New `hook.EnvChangesForWorkdir` constructor for workdir-only changes.
- **CLI**: new `workdir` parent command with a `set` subcommand (leaving room for a future `get`).

```bash
$ buildkite-agent workdir --help
Usage:

  buildkite-agent workdir <command> [options...]

Available commands are:

  set   Sets the working directory for subsequent phases of the job
  help  Shows a list of commands or help for one command
```

```bash
$ buildkite-agent workdir set --help
Usage:

    buildkite-agent workdir set <path>

Description:

Sets the working directory for subsequent phases of the current job. The change
persists across later hooks and the command phase.

This is intended for binary and polyglot (non-shell) hooks, which run in a child
process whose own directory changes are otherwise lost. Wrapped POSIX shell
hooks can change the working directory simply by `cd`-ing.

Relative paths are resolved against the current working directory of this command
(i.e. the hook's actual working directory). The path must exist and be a
directory.
```

## Testing

New tests:
- `jobapi/workdir_test.go`: absolute path → 200 + pending workdir set; relative → 422; empty → 422; `TakePendingWorkdir` returns once then clears.
- `clicommand/workdir_set_test.go`: relative paths resolved against cwd; nonexistent path and non-directory both error before any API call.
- `internal/job/integration/hooks_integration_test.go` (`TestBinaryHookCanSetWorkdir`): a binary `pre-command` hook sets the workdir via the Job API; asserts both the `command` and `post-command` hooks run in the requested directory, proving the change persists across hooks.

## Disclosures / Credits

Claude Code (Opus) implemented this change from a written plan I authored and reviewed, including the tests; I reviewed and verified the result.

### DIFF
--- a/internal/job/integration/test-binary-hook-workdir/main.go
+++ b/internal/job/integration/test-binary-hook-workdir/main.go
@@ -20,7 +20,7 @@ import (
 // records the expected directory in EXPECTED_WORKDIR so the test's command and
 // post-command hooks can assert they run there.
 func main() {
-	ctx := context.TODO()
+	ctx := context.Background()
 
 	c, err := jobapi.NewDefaultClient(ctx)
 	if err != nil {
--- a/internal/job/integration/test-binary-hook-workdir/main.go
+++ b/internal/job/integration/test-binary-hook-workdir/main.go
@@ -1,0 +1,58 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"os"
+	"path/filepath"
+
+	"github.com/buildkite/agent/v3/jobapi"
+)
+
+// This file gets built and used as part of the hooks integration test to ensure
+// that an unwrapped (binary) hook can change the working directory of subsequent
+// phases via the Job API /workdir endpoint.
+//
+// It creates a subdirectory of its own working directory (the checkout dir),
+// resolves it to an absolute, symlink-free path, then asks the executor to use
+// it as the working directory for subsequent hooks and the command. It also
+// records the expected directory in EXPECTED_WORKDIR so the test's command and
+// post-command hooks can assert they run there.
+func main() {
+	ctx := context.TODO()
+
+	c, err := jobapi.NewDefaultClient(ctx)
+	if err != nil {
+		log.Fatalf("error: %v", fmt.Errorf("creating job api client: %w", err))
+	}
+
+	cwd, err := os.Getwd()
+	if err != nil {
+		log.Fatalf("error: %v", fmt.Errorf("getting working directory: %w", err))
+	}
+
+	target := filepath.Join(cwd, "binary-hook-subdir")
+	if err := os.MkdirAll(target, 0o777); err != nil {
+		log.Fatalf("error: %v", fmt.Errorf("creating target directory: %w", err))
+	}
+
+	// Resolve symlinks so the path matches what the command hook observes as its
+	// working directory (on macOS /tmp and /var are symlinks).
+	resolved, err := filepath.EvalSymlinks(target)
+	if err != nil {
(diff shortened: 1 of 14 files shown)


internal/job/integration/test-binary-hook-workdir/main.go
```diff
@@ -0,0 +1,58 @@
...
+// that an unwrapped (binary) hook can change the working directory of subsequent
+// phases via the Job API /workdir endpoint.
+//
+// It creates a subdirectory of its own working directory (the checkout dir),
+// resolves it to an absolute, symlink-free path, then asks the executor to use
+// it as the working directory for subsequent hooks and the command. It also
+// records the expected directory in EXPECTED_WORKDIR so the test's command and
+// post-command hooks can assert they run there.
+func main() {
+	ctx := context.TODO()
```
DrJosh9000: This is at the top of a `main()`, `context.Background()` seems totally appropriate

```suggestion
	ctx := context.Background()
```

jobapi/workdir.go
```diff
@@ -0,0 +1,52 @@
...
+			s.Logger.Errorf("Job API: couldn't write error: %v", err)
+		}
+		return
+	}
+
+	s.mtx.Lock()
+	s.pendingWorkdir = req.Workdir
+	s.mtx.Unlock()
+
+	resp := WorkdirSetResponse{Workdir: req.Workdir}
```
DrJosh9000: Will need a //nolint directive to make golangci-lint shut up.


```suggestion
	resp := WorkdirSetResponse{Workdir: req.Workdir} //nolint:staticcheck // struct literal is clearer than conversion here
```

DrJosh9000 (approved):
Thanks @francoiscampbell , this is looks pretty good! I have two comments that I'll apply directly, then we should be able to merge

PR #3970, opened 2026-06-02
chore: run go1.26.3 fix ./...

DrJosh9000:
## Description

Automatically update some anachronistic Go using `go fix` from Go 1.26.

## Context

Some `interface{}` has snuck in, reported in #3967 but that PR deleted a bunch of code instead of fixing it.

## Changes

As generated by `go1.26.3 fix ./...`.

## Testing

## Disclosures / Credits

The Go team for creating `go fix`.

### DIFF
--- a/internal/cache/configuration/configuration.go
+++ b/internal/cache/configuration/configuration.go
@@ -102,7 +102,7 @@ func loadTemplates() (map[string]Cache, error) {
 
 	decoder := json.NewDecoder(file)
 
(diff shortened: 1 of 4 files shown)


moskyb (approved):

moskyb:
Should we update the agent to use 1.26 as its default toolchain?

DrJosh9000:
> Should we update the agent to use 1.26 as its default toolchain?

Yeah maybe!

PR #3977, opened 2026-06-04
lint: Prefer t.Context in tests

lox:
Tests should use the test lifecycle when creating root contexts so work is cancelled before test cleanup runs. The repo already targets a Go version with `testing.T.Context`, but there was no lint rule preventing new `context.Background()` calls in tests.

This enables the `usetesting` linter for `context.Background()` and updates the existing lint-reported test call sites to use `t.Context()` instead. For example, test operations now pass `t.Context()` directly or derive timeouts and cancellable contexts from it with `context.WithTimeout(t.Context(), ...)`.

The remaining `context.Background()` uses are limited to places where no `testing.T` exists, such as `TestMain` setup and helper-process code, with comments explaining why they intentionally use a root context. Other `usetesting` suggestions such as `t.TempDir` and `t.Setenv` remain disabled so this PR only enforces the context rule.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,14 @@
+version: "2"
+
+linters:
+  enable:
+    - usetesting
+  settings:
+    usetesting:
+      os-create-temp: false
+      os-mkdir-temp: false
+      os-setenv: false
+      os-temp-dir: false
+      os-chdir: false
+      context-background: true
+      context-todo: false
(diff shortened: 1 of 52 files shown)


.golangci.yml
catkins, line 1: Did we not already have a golangci-lint config file 🤔

lox: Nope, apparently just defaults.

CerealBoy (approved):
Nice clean-up and comments where appropriate. I'm also a little surprised we didn't have a `golangci-lint` config file already 😅 

LGTM 🚀 🚀

PR #3978, opened 2026-06-04
fix: Prevent agent HTTP auth redirect leaks

lox:
Authenticated agent HTTP clients currently inject credentials in the transport layer, after Go's redirect sanitizer has already decided whether to strip `Authorization`. If a trusted Buildkite endpoint ever redirected to a different host or scheme, the redirected request could receive the agent token or bearer credential.

This keeps the normal redirect behavior for unauthenticated clients, but authenticated clients now stop at redirects that change scheme or host and return the redirect response instead. Same-origin redirects still follow normally and continue to include the expected authorization header.

For example, an authenticated request from `https://agent.buildkite.com/...` to a `Location: https://example.invalid/...` response will no longer follow that redirect with credentials attached, while `/old` to `/new` on the same origin continues to work.

### DIFF
--- a/internal/agenthttp/client.go
+++ b/internal/agenthttp/client.go
@@ -55,7 +55,21 @@ func NewClient(opts ...ClientOption) *http.Client {
 			Token:    conf.Token,
 			Delegate: transport,
 		},
+		CheckRedirect: checkAuthenticatedRedirect,
 	}
+}
+
+func checkAuthenticatedRedirect(req *http.Request, via []*http.Request) error {
+	if len(via) == 0 {
+		return nil
+	}
+
+	origin := via[0].URL
+	if req.URL.Scheme != origin.Scheme || req.URL.Host != origin.Host {
+		return http.ErrUseLastResponse
+	}
+
+	return nil
 }
 
 // Various NewClient options.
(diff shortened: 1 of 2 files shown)


quinn-diesel:
@buildsworth-bk review please

@lox sorry didn't mean for Codex to review as well

DrJosh9000:
I had a thought about this, since I had a quick look yesterday, but didn't respond. Here it is.

> If a trusted Buildkite endpoint ever redirected to a different host or scheme

On the surface this seems very unlikely. Redirecting HTTPS to HTTP seems remarkably unlikely. A different host is more plausible. To be safe, the backend would have to limit any such redirects to agent versions after this change, or until after all old agents are no longer used, which would limit the utility of such a redirect.

I don't think this is an argument against this change.

lox:
I think this is probably not worth landing.

The issue is real in the abstract, but it relies on a pretty weird redirect from a trusted Buildkite endpoint for authenticated agent traffic. That seems unlikely enough that I don’t reckon the extra client behaviour is justified.

The main cost here isn’t the patch size, it’s that authenticated clients now behave differently around redirects and start surfacing raw `3xx`s in cases where they previously would have just followed them. That’s another subtle rule to own and explain later.

I’m inclined to close this rather than carry the complexity for a fairly hypothetical case.

PR #3979, opened 2026-06-05
Add OTLP job log export

catkins:
## Description

Adds an opt-in job log OTLP sink. When `--job-logs-otlp` / `BUILDKITE_JOB_LOGS_OTLP` is enabled, the agent emits job output as OpenTelemetry log records using the existing OTLP exporter environment configuration. This is an **additional** sink: the normal Buildkite job log is still streamed to the control plane unchanged. The difference from that stream is that OTLP records carry native OTLP timestamps rather than timestamps encoded into the ANSI/OSC job-log body.

The OTLP endpoint and transport are intentionally inherited from the OpenTelemetry exporter configuration rather than adding Buildkite-specific endpoint flags. For logs, set `OTEL_EXPORTER_OTLP_LOGS_ENDPOINT` for a log-specific endpoint or `OTEL_EXPORTER_OTLP_ENDPOINT` for the generic endpoint; protocol selection follows `OTEL_EXPORTER_OTLP_LOGS_PROTOCOL` / `OTEL_EXPORTER_OTLP_PROTOCOL`.

The log records carry native OTLP timestamps plus a small set of Buildkite attributes for correlation: organization, pipeline, branch, queue, agent, build, job, current phase, and current hook scope/plugin where known. Trace correlation uses the native OTLP `LogRecord` trace context fields rather than duplicating `trace_id` / `span_id` as log attributes.

## Parity with the Buildkite job log

The OTLP sink mirrors the **same visible content** customers see in the Buildkite UI / downloadable job log, so an OTLP destination is not confusingly different from Buildkite:

* **Child-process output** (hook/command stdout & stderr) is mirrored via the shell output interceptor.
* **Bootstrap control output** — section headers (`~~~`), prompts (`$`), comments (`#`) and warnings — is mirrored by teeing the redacted shell logger output into the exporter.

Both paths emit the same bytes (including ANSI colour codes) that land in the downloadable Buildkite log. The **only** intended difference is the timestamp transport: OTLP records use the native `LogRecord` timestamp instead of the `\x1b_bk;t=…` OSC markers Buildkite encodes into the raw log body. The human-visible content is identical in both destinations.

Control output is bootstrap narration rather than the output of a specific traced hook/command, so those records carry the base `buildkite.*` attributes but no per-hook span context; child-process records remain trace-correlated to their hook/command span.

## Architecture: bootstrap-only

OTLP job log export lives entirely in the **bootstrap process**, which is the single home for this feature. This is a deliberate choice: the bootstrap is the process that actually runs the hooks and command, so it has structured phase/hook metadata and the active hook/command span directly, with no need to reconstruct them from the job-log text downstream. Keeping emission here also means a single emitter, a single `buildkite.*` attribute schema, and behaviour that does not vary with the (otherwise unrelated) tracing backend.

The bootstrap emits records whenever `--job-logs-otlp` is enabled, regardless of tracing backend:

* When OpenTelemetry tracing is enabled, child-process records are emitted around the actual hook/command execution contexts, so each record carries the same native OTLP `TraceId`/`SpanId` as the corresponding exported hook/command span.
* When tracing is disabled, records are still emitted but remain intentionally uncorrelated (empty trace context).

`--tracing-propagate-traceparent` only controls accepting the Buildkite control-plane traceparent; it is not required for local agent trace/log correlation.

```mermaid
flowchart TD
    subgraph agent["Agent process"]
        AS["buildkite-agent start<br/>--job-logs-otlp"]
    end

    AS -->|"BUILDKITE_JOB_LOGS_OTLP=true"| RUN

    subgraph bootstrap["Bootstrap process — single home for OTLP export"]
        RUN["Executor.Run<br/>JobLogsOTLP enabled?"] -->|"yes (any tracing backend)"| WIRE

        HC["hook / command<br/>stdout + stderr"] --> SH["shell OutputInterceptor<br/>(per hook/command, structured attrs)"]
        CTRL["bootstrap control output<br/>headers / prompts / comments / warnings"] --> SL["shell logger"]

        WIRE(("wire sinks"))
        WIRE --> SH
        WIRE --> SL

        SH --> TEE{{"tee raw output"}}
        TEE --> JR["job-log redactor<br/>replacer.Mux"]
        TEE --> OR["OTLP redactor<br/>replacer seeded with live needles"]
        JR --> JLOG["Buildkite job log stream"]
        OR --> EM["line emitter<br/>+ buildkite.* attributes"]

        SL --> LR["logger redactor"]
        LR --> LTEE{{"tee redacted control output"}}
        LTEE --> JLOG
        LTEE --> CEM["control line emitter<br/>+ base buildkite.* attributes"]

        EM --> EXP["OTLP log exporter<br/>(batch)"]
        CEM --> EXP
    end

    SPAN["active hook/command span<br/>only when OTel tracing enabled"] -. "native TraceId / SpanId" .-> EM
    EXP -->|"OTLP/gRPC or HTTP"| BE[("OTLP backend")]

    classDef redact fill:#fde2e2,stroke:#d33;
    class JR,OR,LR redact;
```

Both sinks share the same redaction (`replacer`, highlighted), so secrets are `[REDACTED]` in the OTLP records exactly as in the job log. Control output is teed **after** the logger redactor, so it is never re-redacted and never leaks pre-redaction bytes. Trace context is attached to child-process OTLP records only when OpenTelemetry tracing is enabled (dotted edge); otherwise records are emitted uncorrelated.

## Redaction

OTLP records are redacted with the **same secret needles as the customer-facing job log**. The child-process OTLP copy is routed through its own `replacer.Replacer`, seeded from the job's live redactor `Mux`, so secret values are replaced with `[REDACTED]` before they reach the OTLP backend (including secrets split across multiple writes). Control output is mirrored from the post-redaction side of the shell logger redactor, so it inherits the same redaction without a second pass.

Relevant OpenTelemetry references:

* [OTLP exporter configuration environment variables](<https://opentelemetry.io/docs/concepts/sdk-configuration/otlp-exporter-configuration/>)
* [Logs data model: trace context fields](<https://opentelemetry.io/docs/specs/otel/logs/data-model/#trace-context-fields>)
* [Logs specification: log correlation](<https://opentelemetry.io/docs/specs/otel/logs/#log-correlation>)
* [Semantic convention naming guidelines](<https://opentelemetry.io/docs/specs/semconv/general/naming/>)
* [CI/CD log semantic conventions](<https://opentelemetry.io/docs/specs/semconv/cicd/cicd-logs/>)

## Context

Slack context: [https://buildkite-corp.slack.com/archives/C05R3MTRK38/p1780282420246339](<https://buildkite-corp.slack.com/archives/C05R3MTRK38/p1780282420246339>)

This is a spike for moving high-rate job-log consumption toward agent-side OTLP log export, using the existing OpenTelemetry SDK plumbing in the agent.

## Changes

* Adds `--job-logs-otlp` and `BUILDKITE_JOB_LOGS_OTLP` to `buildkite-agent start`, propagated to the bootstrap subprocess.
* Adds an OTLP job log writer in the bootstrap that line-buffers process output and emits OpenTelemetry `LogRecord`s with native timestamps and structured Buildkite hook/command attributes.
* Mirrors bootstrap control output (section headers, prompts, comments, warnings) into OTLP by teeing the redacted shell logger, giving the OTLP destination parity with the downloadable Buildkite log and the UI stream.
* Activates OTLP export in the bootstrap whenever the flag is set, regardless of tracing backend; trace correlation is present only when OpenTelemetry tracing is enabled.
* Redacts OTLP record bodies through the same live secret needle set as the job log redactor.
* Uses the existing OpenTelemetry exporter environment configuration for endpoint/protocol selection instead of adding Buildkite-specific endpoint flags.
* Emits trace/span correlation through native OTLP `LogRecord` fields, not duplicated log attributes.
* Keeps Buildkite control-plane traceparent acceptance behind the existing `--tracing-propagate-traceparent` opt-in.
* Detaches stored OTLP emit contexts from command cancellation while preserving trace context, so cleanup flushes do not inherit a canceled command context.

CLI help excerpt:

```text
--job-logs-otlp                                 Export job logs directly as OpenTelemetry log records using the OTEL_EXPORTER_OTLP_LOGS_* / OTEL_EXPORTER_OTLP_* environment configuration (default: false) [$BUILDKITE_JOB_LOGS_OTLP]
```

## Testing

Additional verification:

* `go build ./...`, `golangci-lint run ./internal/job/`
* `go test -race ./internal/job/ ./internal/replacer/ ./internal/shell/` (one pre-existing, unrelated git failure — `TestVerifyCommit/fails_when_commit_is_not_on_branch` — reproduces on a clean tree).
* Regression tests: `TestOTLPJobLoggerRedactsSecrets`, `TestOTLPJobLoggerRedactsSecretsSplitAcrossWrites`, `TestOTLPJobLoggerControlWriter`, `TestOTLPJobLoggerControlWriterReused`.

Validated end-to-end with a **real build on a local Buildkite** (`buildkite.localhost`) exporting to a local ClickStack (ClickHouse + OpenTelemetry collector). A locally-built agent ran a job with a secret env var, then the downloadable Buildkite log (via the v2 REST API) was compared against `default.otel_logs` / `default.otel_traces` for the same job id:

* **Control-line parity:** the downloadable Buildkite log's section headers, prompts and comments (`~~~ Preparing working directory`, `$ git clone …`, `# Creating …`, `~~~ Running commands`) all appear verbatim — including ANSI colour codes — in the OTLP records. The only difference is the OTLP records use native timestamps instead of the `\x1b_bk;t=…` OSC markers in the raw Buildkite log body.
* **Redaction:** a command printing the secret produced `leak=[REDACTED]` in both the downloadable Buildkite log and the OTLP record bodies.
* **Correlation:** child-process records correlated to the exported `checkout` and `default command hook` spans via native OTLP trace context; control-narration records were emitted with base attributes and no span context, as intended.

Buildkite.

## Disclosures / Credits

Codex and Amp assisted with implementation, adversarial review, local verification, and drafting this PR description.


internal/job/otlp_job_logger.go
```diff
@@ -0,0 +1,187 @@
...
+	for len(data) > 0 {
+		i := bytes.IndexByte(data, '\n')
+		if i < 0 {
+			w.buf = append(w.buf, data...)
+			return n, err
+		}
+
+		line := append(w.buf, data[:i]...)
+		line = bytes.TrimSuffix(line, []byte{'\r'})
+		w.emit(string(line))
```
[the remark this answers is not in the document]
quinn-diesel: @catkins this popped up with another review i was doing, any issue with it?

catkins:
@codex review

catkins:
@codex review

Please focus on the new control-line parity change (latest commit `2d124cc3`): the `teeWriter` in `internal/job/executor.go` that mirrors the redacted shell logger output into the OTLP `controlWriter` in `internal/job/otlp_job_logger.go`. Specifically: redaction correctness (no pre-redaction leak), concurrency/races between the cancellation goroutine and Close/flush, and that there is no double-emission of child-process output.

catkins:
@codex review

Updated since the last review with commit `b7e9eb9e` addressing adversarial review findings:
- The per-command OTLP redactor now syncs with the live redactor `Mux` on each write (`internal/job/otlp_job_logger.go`), so secrets added mid-command (e.g. via the Job API) are redacted in OTLP output, not just secrets known at writer creation.
- `teeWriter.Write` now holds the lock across the whole write (`internal/job/executor.go`), so detaching the OTLP control sink can't race with an in-flight write.

Validated end-to-end against a real build on `buildkite.localhost` exporting to ClickStack: control lines have parity with the downloadable Buildkite log, secrets are `[REDACTED]` in both destinations, and child-process records correlate to the `checkout`/`default command hook` spans.

DrJosh9000:
Hey @catkins just wondering if this is ready for review or still needed?

internal/job/executor.go
```diff
@@ -74,10 +74,54 @@ type Executor struct {
...
+	// down) cannot complete while a write is still in flight. This guarantees
+	// no control output lands on the OTLP emitter after it has been flushed.
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
+	n, err := w.primary.Write(p)
+	if w.secondary != nil {
+		// The secondary sink is best-effort: failures must not affect the
+		// primary (customer-facing) job log output.
+		_, _ = w.secondary.Write(p)
```
DrJosh9000: Why is the secondary sink considered best-effort? I can imagine a user totally ignoring the BK UI and instead expecting 100% of logs to always be present in their OTel system.

internal/job/executor.go
```diff
@@ -156,6 +200,36 @@ func (e *Executor) Run(ctx context.Context) (exitCode int) {
...
+			defer func() {
+				if e.loggerTee != nil {
+					e.loggerTee.setSecondary(nil)
+				}
+				if err := jobLogger.Close(); err != nil {
+					e.shell.Warningf("Failed to close OTLP job log exporter: %v", err)
+				}
+				e.otlpJobLogger = nil
+			}()
+		}
```
DrJosh9000: The large `else` branch here makes me think this might be better as a method with an early return:

```go
func (e *Executor) setupOTLPJobLogger(ctx context.Context) func() {
	jobLogger, err := newOTLPJobLogger(ctx, e)
	if err != nil {
		e.shell.Warningf("Failed to initialize OTLP job log exporter: %v", err)
		return func() {}
	}
	e.otlpJobLogger = jobLogger
	e.shell.SetOutputInterceptor(jobLogger.Wrap)
	// Mirror the redacted shell logger control output (section headers,
	// prompts, comments, warnings) into OTLP so the exported records
	// match the downloadable Buildkite job log and the UI stream.
	if e.loggerTee != nil {
		e.loggerTee.setSecondary(jobLogger.controlWriter())
	}
	return func() {
		if e.loggerTee != nil {
			e.loggerTee.setSecondary(nil)
		}
		if err := jobLogger.Close(); err != nil {
			e.shell.Warningf("Failed to close OTLP job log exporter: %v", err)
		}
		e.otlpJobLogger = nil
	}
}
```

```suggestion
		cleanup := e.setupOTLPJobLogger(ctx)
		defer cleanup()
```

internal/job/otlp_job_logger.go
```diff
@@ -0,0 +1,375 @@
...
+func (l *otlpJobLogger) Close() error {
+	if l.control != nil {
+		l.control.flush()
+	}
+	l.FlushRedactors()
+	if l.provider == nil {
+		return nil
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
DrJosh9000: I don't think this `Close` method needs to satisfy any interfaces like `io.Closer`. So we should probably prefer to pass in a context arg over using `context.Background()`. Even if that arg will probably need to be detached (`context.WithoutCancel`).

internal/job/otlp_job_logger.go
```diff
@@ -0,0 +1,375 @@
...
+	av, bv := reflect.ValueOf(a), reflect.ValueOf(b)
+	return av.Type() == bv.Type() && av.Comparable() && av.Interface() == bv.Interface()
+}
+
+// otlpLineEmitter line-buffers (already redacted) output and emits each line as
+// an OpenTelemetry log record with a native timestamp. Write and flush are
+// guarded by mu because control output may be written from multiple goroutines
+// (e.g. the cancellation handler emitting a comment).
+type otlpLineEmitter struct {
+	ctx   context.Context
```
DrJosh9000, line 260: Why does a context need to be stored in this struct? (The answer should probably be a comment here, perhaps with a suggestion for a future refactor.)

internal/job/otlp_job_logger.go
```diff
@@ -0,0 +1,375 @@
...
+	return w.stream.emitter.Write(data)
+}
+
+func sameOTLPWriter(a, b io.Writer) bool {
+	if a == nil || b == nil {
+		return a == nil && b == nil
+	}
+	av, bv := reflect.ValueOf(a), reflect.ValueOf(b)
+	return av.Type() == bv.Type() && av.Comparable() && av.Interface() == bv.Interface()
+}
```
DrJosh9000: I believe this sort of thing is only necessary if there is a concrete type implementing `io.Writer` in use that is not comparable...are there any? All calls to `Wrap` I can see pass a `*bytes.Buffer`, which is comparable (as a pointer).

DrJosh9000 (commented):
I have thoughts

internal/job/executor.go
catkins: thats reasonable

internal/job/otlp_job_logger.go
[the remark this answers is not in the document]
catkins: 🤖 Fixed. OTLP output now passes through persistent per-stream redactors seeded from—and synchronized with—the customer-facing redactor set. The final local Buildkite/ClickStack run confirmed the synthetic secret was redacted in both sinks and absent from OTLP logs and trace attributes.

internal/job/executor.go
```diff
@@ -1293,7 +1319,11 @@ func (e *Executor) defaultCommandPhase(ctx context.Context) (retErr error) {
 		e.shell.Promptf("%s", cmdToExec)
 	}
 
-	err = e.shell.Command(cmd[0], cmd[1:]...).Run(ctx, shell.ShowPrompt(false))
+	err = e.shell.Command(cmd[0], cmd[1:]...).Run(ctx, shell.ShowPrompt(false), shell.WithOutputAttributes(map[string]string{
+		"buildkite.phase":      "hook",
+		"buildkite.hook.name":  "command",
+		"buildkite.hook.scope": "default",
```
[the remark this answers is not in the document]
catkins: 🤖 Fixed. Default command output now carries `buildkite.phase=command`, with the command hook name and default scope supplied as structured attributes rather than inferred from log text. The final integration run confirmed those attributes on stdout and stderr records.

internal/shell/shell.go
```diff
@@ -482,6 +494,27 @@ func (c Command) Run(ctx context.Context, opts ...RunCommandOpt) error {
 		}()
 	}
 
+	var flushers []interface{ Flush() }
+	if c.shell.outputInterceptor != nil {
+		if cfg.captureStdout == nil && stdout != nil && stdout != io.Discard {
+			stdout = c.shell.outputInterceptor(ctx, stdout, cfg.outputAttrs)
```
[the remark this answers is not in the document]
catkins: 🤖 The runtime shell-logger path is now covered: already-redacted headers, prompts, comments, and warnings are teed into the OTLP control writer. Warnings emitted before the exporter can be constructed necessarily remain outside that exporter. Final local verification found 31 OTLP records including the runtime control lines and matching child output.

internal/job/otlp_job_logger.go
```diff
@@ -0,0 +1,224 @@
...
+		attrs: appendLogAttrs(l.attrs, attrs),
+	}
+	// Redact OTLP output using the same secret needles as the job-log
+	// redactor, so secrets never reach the OTLP backend. A fresh per-command
+	// Replacer feeds the line emitter; it is seeded with the needles known at
+	// the time the command starts (which already include env-, secret- and Job
+	// API-derived needles added during earlier phases).
+	return &otlpJobLogWriter{
+		out:      out,
+		redactor: replacer.New(emitter, l.redactors.Needles(), redact.Redacted),
```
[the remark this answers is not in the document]
catkins: 🤖 Fixed. Each process-output write synchronizes the persistent OTLP redactor with the live redactor Mux before emitting, so secrets added through the Job API during a command are included. The regression test covers a needle added after the writer was created.

internal/job/executor.go
```diff
@@ -74,6 +74,42 @@ type Executor struct {
...
+type teeWriter struct {
+	mu        sync.Mutex
+	primary   io.Writer
+	secondary io.Writer
+}
+
+func (w *teeWriter) Write(p []byte) (int, error) {
+	w.mu.Lock()
+	secondary := w.secondary
+	w.mu.Unlock()
```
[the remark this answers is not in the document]
catkins: 🤖 Fixed. `teeWriter` now holds its mutex through both sink writes, and cleanup detaches the secondary under that same lock before flush/shutdown. This waits for in-flight control writes and prevents post-shutdown emission.

internal/job/otlp_job_logger.go
```diff
@@ -0,0 +1,269 @@
...
+
+func (e *otlpLineEmitter) Write(data []byte) (int, error) {
+	e.mu.Lock()
+	defer e.mu.Unlock()
+
+	origLen := len(data)
+	for len(data) > 0 {
+		i := bytes.IndexByte(data, '\n')
+		if i < 0 {
+			e.buf = append(e.buf, data...)
```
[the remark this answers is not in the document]
catkins: 🤖 Fixed. Unterminated output is emitted in bounded 64 KiB records while retaining normal line behavior and CRLF handling. The regression test writes more than two chunks without a newline and verifies bounded pre-flush emission plus byte-for-byte reconstruction.

internal/job/executor.go
catkins: 🤖 Agreed and fixed. `teeWriter` now returns secondary write failures, converts secondary short writes to `io.ErrShortWrite`, and preserves the primary sink byte count. OTLP also uses the synchronous processor so exporter back-pressure cannot overflow and silently drop records from a batch queue.

internal/job/executor.go
catkins: 🤖 Applied this refactor. OTLP initialization and teardown now live in `setupOTLPJobLogger`, use the early-return no-op cleanup on initialization failure, and are called with the explicit `cleanup := ...; defer cleanup()` form. Cleanup also detaches the interceptor before shutdown.

internal/job/otlp_job_logger.go
catkins: 🤖 Fixed. `Close` now accepts the caller context and derives a detached five-second shutdown context with `context.WithoutCancel`, so cancellation does not prevent the final force-flush and provider shutdown.

internal/job/otlp_job_logger.go
catkins: 🤖 Added the explanation here. The writer API does not carry a context, while line and redaction buffering can emit after the supplying call, so the detached context is retained to preserve command trace correlation through delayed emission. The comment also points toward a future record-aware buffer that could retain context per chunk.

internal/job/otlp_job_logger.go
catkins: 🤖 Fixed. The reflective OTLP writer-identity helper is gone. The persistent stream table now keys directly on the comparable `io.Writer` interface values; executor production writers and the test writers are pointer-based.

internal/job/executor.go
```diff
@@ -74,10 +74,84 @@ type Executor struct {
...
 	jobAPI *jobapi.Server
+
+	// loggerTee carries the redacted shell logger output (section headers,
+	// prompts, comments, warnings) to stderr and, when OTLP job logging is
+	// enabled, mirrors it into the OTLP exporter so the exported records match
+	// the customer-facing Buildkite job log.
+	loggerTee *teeWriter
+}
+
+// teeWriter writes to a primary writer and, when configured, also to a
```
DrJosh9000, line 94: This file is already huge, can we have `teeWriter` in a separate file?

clicommand/bootstrap.go
```diff
@@ -367,12 +368,16 @@ var BootstrapCommand = cli.Command{
...
-
 		cli.BoolFlag{
 			Name:   "no-job-api",
 			Usage:  "Disables the Job API, which gives commands in jobs some abilities to introspect and mutate the state of the job (default: false)",
 			EnvVar: "BUILDKITE_AGENT_NO_JOB_API",
 		},
+		cli.BoolFlag{
+			Name:   "job-logs-otlp",
+			EnvVar: "BUILDKITE_JOB_LOGS_OTLP",
+			Hidden: true,
```
DrJosh9000, line 379: I don't think we bother hiding flags, even in bootstrap, unless they are deprecated or not yet ready.

internal/job/otlp_job_logger.go
```diff
@@ -0,0 +1,372 @@
...
+
+	if len(e.buf) == 0 {
+		return
+	}
+	e.emit(string(e.buf))
+	e.buf = e.buf[:0]
+}
+
+func (e *otlpLineEmitter) emit(line string) {
+	now := time.Now()
```
DrJosh9000, line 325: The ANSI timestamper writes timestamp at the start of each line, and also injects additional timestamps every so often for lines that take a long time to be produced by the command. The `now` computed here is at the end of each line. That might be a discrepancy the especially log-curious customers notice.

internal/job/otlp_job_logger.go
```diff
@@ -0,0 +1,372 @@
...
+
+func (w *otlpJobLogWriter) Write(data []byte) (int, error) {
+	w.stream.mu.Lock()
+	defer w.stream.mu.Unlock()
+
+	n, err := w.stream.out.Write(data)
+	// Keep the OTLP redactor's needles in sync with the live job redactor
+	// before redacting, so secrets added mid-command (e.g. via the Job API)
+	// are redacted in OTLP output too. Replacer.Add deduplicates, so re-adding
+	// the full needle set each write is safe.
```
DrJosh9000, line 196: This problem is solved elsewhere by adding all redactors the executor's mux, which is updated with new needles only when they change. The replacer's `Add` method is not engineered to be called super frequently (worst case is quadratic in number of needles).

zhming0:
Closed in favor of #4121 , but thanks @catkins!

PR #3981, opened 2026-06-05
chore: Add artifact upload timing summary

lox:
Large artifact uploads currently leave us guessing which part of the pipeline is slow. The job log shows the overall artifact upload step and per-artifact activity, but it does not give a compact breakdown of collection, Buildkite artifact creation, upload worker time, or state update calls.

This adds a single info-level summary after successful artifact uploads so operators can compare runs without enabling HTTP tracing or reconstructing timings from many log lines. Example:

```text
Artifact upload timings: collect=1.456s create=82ms upload=334ms state_update=11ms artifacts=1296 bytes=5.1 MiB batches=44 work_units=1296 max_workers=30 state_updates=44
```

The change also adds debug timing breadcrumbs for path discovery, collection worker drain time, artifact creation batches, upload batches, and state update calls. The upload behavior and default concurrency are unchanged.

### DIFF
--- a/internal/artifact/uploader.go
+++ b/internal/artifact/uploader.go
@@ -78,6 +78,65 @@ type Uploader struct {
 	apiClient APIClient
 }
 
+type artifactUploadTimings struct {
+	collectDuration time.Duration
+	createDuration  time.Duration
+	uploadDuration  time.Duration
+	stateDuration   time.Duration
+
+	artifactCount    int
+	artifactBytes    int64
+	batches          int
+	workUnits        int
+	maxWorkerCount   int
+	stateUpdateCount int
+}
+
+type artifactUploadBatchStats struct {
+	workUnits      int
+	workerCount    int
+	uploadDuration time.Duration
+
+	stateUpdateDuration time.Duration
+	stateUpdateCount    int
+}
+
+func (t *artifactUploadTimings) addArtifacts(artifacts []*api.Artifact) {
+	t.artifactCount = len(artifacts)
+	for _, artifact := range artifacts {
+		t.artifactBytes += artifact.FileSize
+	}
+}
+
+func (t *artifactUploadTimings) addCreateBatch(duration time.Duration) {
+	t.createDuration += duration
+	t.batches++
+}
+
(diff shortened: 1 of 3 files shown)


lox:
@catkins do you think we should include the low-cardinality OTel side of this in this PR as well?

The current patch adds the job-log summary, which is useful for customers who do not have tracing enabled. For OTel users, we could attach the same diagnostic shape to the existing `artifact-upload` span as low-cardinality attributes, roughly:

- `artifact.count`
- `artifact.bytes`
- `artifact.batch_count`
- `artifact.work_unit_count`
- `artifact.max_workers`
- `artifact.collect_ms`
- `artifact.create_ms`
- `artifact.upload_ms`
- `artifact.state_update_ms`

I would avoid per-artifact/per-work-unit spans or anything path/destination/URL-like. This would just make the same phase breakdown available in traces without log scraping. Worth folding in here, or better as a follow-up?

internal/artifact/uploader.go
```diff
@@ -607,6 +690,7 @@ func (a *Uploader) upload(ctx context.Context, artifacts []*api.Artifact, upload
 
 	// Wait for the workers to finish
 	wg.Wait()
+	stats.uploadDuration = time.Since(uploadStart)
```
catkins, line 693: I _was_ going to ask whether this needed any mutex-es, but it looks like the writing to the stats struct is all after the fanout. 👌

catkins:
> [...] we could attach the same diagnostic shape to the existing artifact-upload span as low-cardinality attributes [...] I would avoid per-artifact/per-work-unit spans or anything path/destination/URL-like. 

By "low cardinality attributes", are you referring to the idea that we'd capture them at the level of the `buildkite-agent artifact upload ...` call as opposed to each individual uploaded object? I agree that avoiding the per-artifact/work-unit spans is the right level of detail.

> Worth folding in here, or better as a follow-up?

My (soft) preference is a small follow up or stacked PR, but I endorse the idea either way!

catkins (approved):
Nice, this is useful debugging info to expose!

lox:
@catkins followed on in https://github.com/buildkite/agent/pull/3982

PR #3982, opened 2026-06-06
chore: Add artifact upload span attributes

lox:
The artifact upload timing summary is now visible in job logs, but tracing users still need to scrape those logs to get the same phase breakdown in their trace tooling.

This adds command-level attributes to the existing `artifact-upload` span using the same low-cardinality summary fields. The attributes describe the upload as a whole, not individual artifacts or work units, so traces get useful aggregate diagnostics without adding span volume or sensitive path/destination data.

The span now includes attributes like:

```text
artifact.count
artifact.bytes
artifact.batch_count
artifact.work_unit_count
artifact.max_workers
artifact.state_update_count
artifact.collect_ms
artifact.create_ms
artifact.upload_ms
artifact.state_update_ms
```

Artifact upload behavior and the existing job-log summary are unchanged.

### DIFF
--- a/internal/artifact/uploader.go
+++ b/internal/artifact/uploader.go
@@ -25,6 +25,8 @@ import (
 	"github.com/buildkite/agent/v3/logger"
 	"github.com/buildkite/roko"
 	"github.com/dustin/go-humanize"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
 )
 
 const ArtifactFallbackMimeType = "binary/octet-stream"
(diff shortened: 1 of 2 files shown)


internal/artifact/uploader.go
```diff
@@ -121,6 +123,21 @@ func (t *artifactUploadTimings) addUploadBatch(stats artifactUploadBatchStats) {
 	t.maxWorkerCount = max(t.maxWorkerCount, stats.workerCount)
 }
 
+func (t *artifactUploadTimings) setSpanAttributes(ctx context.Context) {
+	trace.SpanFromContext(ctx).SetAttributes(
```
catkins, line 127: https://pkg.go.dev/go.opentelemetry.io/otel/trace#SpanFromContext

> If no Span is currently set in ctx an implementation of a Span that performs no operations is returned.

My only comment here was going to be asking whether this needed `nil` guards, but looks like we are all good here 🎉

catkins (approved):
🚀 excellent

PR #3987, opened 2026-06-09
Fix env set/unset panic on Job API errors

petetomasik:
## Description

`buildkite-agent env set` and `buildkite-agent env unset` logged the error returned by the Job API client but did not return it, then fell through to use the response value. `EnvUpdate`/`EnvDelete` return a nil response on error, so:

- `env set`, with the default `plain` output format, dereferenced the nil response (`resp.Added`) and panicked.
- `env unset` silently swallowed the error and exited 0, printing "No variables unset."

Both now return the error, so the command fails cleanly with a non-zero exit code and a useful message instead of panicking or hiding the failure.

## Changes

- `clicommand/env_set.go`: return the `EnvUpdate` error instead of logging and continuing; drop the now-unused logger binding.
- `clicommand/env_unset.go`: same for the `EnvDelete` error.

## Testing

## Disclosures / Credits

The original fix was implemented by @scadu in commit 33546381 on the `git-checkout-features` branch. During a code review of that branch using Claude Code (Opus 4.8), the same bug was found still present on `main`, so the fix was extracted to this standalone branch for an isolated fix.

### DIFF
--- a/clicommand/env_set.go
+++ b/clicommand/env_set.go
@@ -69,7 +69,7 @@ var EnvSetCommand = cli.Command{
 
 func envSetAction(c *cli.Context) error {
 	ctx := context.Background()
-	ctx, cfg, l, _, done := setupLoggerAndConfig[EnvSetConfig](ctx, c)
+	ctx, cfg, _, _, done := setupLoggerAndConfig[EnvSetConfig](ctx, c)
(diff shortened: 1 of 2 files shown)


DrJosh9000 (approved):
LGTM, makes sense ✅

PR #3989, opened 2026-06-10
fix: Test build failure due to missing context import

DrJosh9000:
## Description

Replace more `context.Background()` calls in tests with `t.Context()`, particularly where the "context" import had been removed from a file.

Also test-engine-client doesn't seem to do the right thing with build failures (the test is retried and passes? Investigating...), so disable retries.

## Context

Breakage added when #3977 was merged

## Changes

As described above.

## Testing

## Disclosures / Credits

I did it myself.


DrJosh9000:
@buildsworth-bk review

PR #3990, opened 2026-06-10
fix: flaky TestCollect_Literal

DrJosh9000:
## Description

`TestCollect_Literal` is flaky, because the artifact collector is concurrent, sometimes producing output in a different order. The failure was masked by the Test Engine client automatic retry.

## Context

Noticed while doing the previous fix.

## Changes

Sort the data.

## Testing

## Disclosures / Credits

I did not use AI tools at all


DrJosh9000:
@buildsworth-bk review

PR #3991, opened 2026-06-10
fix: flaky TestContextCancelInterrupts

DrJosh9000:
## Description

(Try to) fix another flaky test, this time `TestContextCancelInterrupts/SIGINT`. To be honest I don't have a theoretical explanation for why this is flaky. There must be a command and an `*os.Process` after `Start`, but perhaps there is a small window in which the subprocess isn't able to handle `SIGINT`?

## Context

Found after merging the last fix.

## Changes

Add a small sleep after the process `started` channel is closed, to be even more sure the subprocess has started (?)  I hate it

## Testing

## Disclosures / Credits

I was rained on today while coming home from lunch.

### DIFF
--- a/internal/shell/shell_test.go
+++ b/internal/shell/shell_test.go
@@ -202,7 +202,7 @@ func TestContextCancelInterrupts(t *testing.T) {
 			defer cancel()
 
 			// Wait for the process to start before cancelling the context.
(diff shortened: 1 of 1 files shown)


DrJosh9000:
@buildsworth-bk review

isaacsu (approved):

PR #3992, opened 2026-06-11
Add `buildkite-agent job promise-failure` command

123sarahj123:
## Description

Add a `buildkite-agent job promise-failure <exit-status>` command. It lets a running job promise up front that it will finish with a failing exit status, via the new `PUT jobs/:id/promise_failure` Agent API endpoint, so a build can begin failing before the job actually finishes.

The promise is binding: it sets a floor on the job's outcome. If the job is later reported as a success, the promised exit status is recorded instead, so the job still fails. Likewise, if a hard failure was promised but the job reports a soft-failure status, the promised status is kept. Any other reported failure is recorded as reported. (See `effective_exit_status_for` in the agent jobs finish path.)

This follows the existing `buildkite-agent job update` command shape: `roko` retries, secret redaction of `--reason`, and `--job` defaulting to `BUILDKITE_JOB_ID`. The exit status is sent as a JSON integer and validated server-side. Repeated calls with the same status are idempotent; declaring a *different* status once one is already recorded is rejected.

The command is **hidden** until the early-failure feature is generally available (the same approach as the `cache` commands), so we can enable it per-organization without advertising it broadly.

If the API doesn't accept the promise, the command **exits non-zero** with a clear message rather than collapsing to success — that keeps it composable in scripts, where a caller who considers a rejection acceptable can append `|| true`. Each known case gets a tailored error:

- **404** — promised failures aren't enabled for the organization
- **409** — a different promised status was already declared
- **422** — the job is no longer running (cancel/timeout race)
- **401/403** and exhausted retries on **429/5xx/network** also fail

## Context

Implements PB-1911.

## Changes

- `api`: add `api.Client.PromiseFailure` and `JobPromiseFailureRequest` (PUT `jobs/:id/promise_failure`), plus client tests.
- `clicommand`: add the `job promise-failure` command (hidden), register it under the `job` subcommand group, and add it to the config-completeness test.
- `.gitignore`: ignore the `buildkite-agent` build artifact produced by `go build -o buildkite-agent .`.

`buildkite-agent job promise-failure --help`:

```
Usage:

    buildkite-agent job promise-failure <exit-status> [options...]

Description:

Promise the current job will finish with a failing exit status. This records
a non-zero exit status that the job is expected to finish with, allowing the
build to begin failing before the job actually completes.

The promise is binding: it sets a floor on the job's outcome. If the job is
later reported as a success, the promised exit status is recorded instead, so
the job still fails. Likewise, if a hard failure was promised but the job
reports a soft-failure status, the promised status is kept. Any other
reported failure is recorded as reported.

Repeated calls with the same exit status are idempotent. Declaring a
different exit status once one is already recorded is rejected.

The command exits non-zero if the promise is not accepted (for example, if
the job is no longer running, or a different exit status was already
promised). Append '|| true' if you would prefer to ignore that in a script.

Example:

    $ buildkite-agent job promise-failure 1
    $ buildkite-agent job promise-failure 42 --reason "detected failing tests"
```

## Testing

Added unit tests for the new API client method covering the success (204), feature-disabled (404), conflict (409), and job-not-running (422) responses.

## Disclosures / Credits

I used Amp to research the server-side endpoint, design the status handling, and write the implementation and tests, with GPT-5 for a code review pass. I reviewed all of the changes myself.

### DIFF
--- a/clicommand/job_promise_failure.go
+++ b/clicommand/job_promise_failure.go
@@ -20,17 +20,22 @@ const jobPromiseFailureHelpDescription = `Usage:
 
 Description:
 
-Declare a promised (early) failure for the current job. This records a
-non-zero exit status that the job is expected to finish with, allowing the
-build to begin failing before the job actually completes. The job keeps
-running and finishes normally; only the build-failing cascade starts early.
+Promise the current job will finish with a failing exit status. This records
+a non-zero exit status that the job is expected to finish with, allowing the
+build to begin failing before the job actually completes.
+
+The promise is binding: it sets a floor on the job's outcome. If the job is
+later reported as a success, the promised exit status is recorded instead, so
+the job still fails. Likewise, if a hard failure was promised but the job
+reports a soft-failure status, the promised status is kept. Any other
+reported failure is recorded as reported.
 
 Repeated calls with the same exit status are idempotent. Declaring a
 different exit status once one is already recorded is rejected.
 
-This command requires the early-failure feature to be enabled for your
-organization. If it is not, the command logs a warning and exits
-successfully, so it can safely be added to pipelines ahead of rollout.
+The command exits non-zero if the promise is not accepted (for example, if
+the job is no longer running, or a different exit status was already
+promised). Append '|| true' if you would prefer to ignore that in a script.
 
 Example:
 
@@ -50,8 +55,9 @@ type JobPromiseFailureConfig struct {
 
 var JobPromiseFailureCommand = cli.Command{
 	Name:        "promise-failure",
-	Usage:       "Declare a promised (early) failure for a job",
+	Usage:       "Promise a job will finish with a failing exit status",
 	Description: jobPromiseFailureHelpDescription,
+	Hidden:      true, // hidden until the early-failure feature is generally available
 	Flags: slices.Concat(globalFlags(), apiFlags(), []cli.Flag{
 		cli.StringFlag{
 			Name:   "job",
@@ -109,25 +115,18 @@ var JobPromiseFailureCommand = cli.Command{
 			return nil
 		})
 		if err != nil {
-			// These statuses are expected during a gradual rollout and aren't
-			// worth failing the build over, so warn and exit successfully.
+			// The promise wasn't accepted. Exit non-zero so the outcome is
+			// visible to scripts; callers who consider a given case acceptable
+			// can append '|| true'.
 			switch {
 			case api.IsErrHavingStatus(err, http.StatusNotFound):
-				// The early-failure feature is not enabled for this organization.
-				l.Warnf("Promised job failure is not enabled for this organization; skipping declaration")
-				return nil
+				return fmt.Errorf("promised failures are not enabled for this organization: %w", err)
 
 			case api.IsErrHavingStatus(err, http.StatusConflict):
-				// A different promised exit status was already declared; the
-				// first declaration already took effect.
-				l.Warnf("A different promised exit status has already been declared for this job; ignoring")
-				return nil
+				return fmt.Errorf("a different promised exit status has already been declared for this job: %w", err)
 
 			case api.IsErrHavingStatus(err, http.StatusUnprocessableEntity):
-				// Most likely the job is no longer running (cancelled or timed
-				// out) by the time this declaration arrived: a benign race.
-				l.Warnf("Could not declare promised failure (the job may no longer be running); ignoring")
-				return nil
+				return fmt.Errorf("the job is no longer running and cannot accept a promised failure: %w", err)
 			}
 
 			return fmt.Errorf("failed to declare promised job failure: %w", err)
--- a/clicommand/job_promise_failure.go
+++ b/clicommand/job_promise_failure.go
@@ -80,8 +80,10 @@ var JobPromiseFailureCommand = cli.Command{
 		if err != nil {
 			return fmt.Errorf("exit status must be an integer: %w", err)
 		}
-		if exitStatus == 0 {
-			return fmt.Errorf("exit status must be non-zero: a promised failure cannot have a successful (0) exit status")
+		// Only positive exit statuses are meaningful here: 0 is success, and
+		// negative values (such as -1) are reserved for internal use.
+		if exitStatus <= 0 {
+			return fmt.Errorf("exit status must be a positive integer: a promised failure cannot have a zero (successful) or negative exit status")
 		}
 
 		client := api.NewClient(l, loadAPIClientConfig(cfg, "AgentAccessToken"))
--- a/clicommand/job_promise_failure.go
+++ b/clicommand/job_promise_failure.go
@@ -93,7 +93,7 @@ var JobPromiseFailureCommand = cli.Command{
 			return err
 		}
 		if redactedValue := redact.String(cfg.Reason, needles); redactedValue != cfg.Reason {
-			l.Warnf("The reason for job %q contained one or more secrets from environment variables that have been redacted. If this is deliberate, pass --redacted-vars='' or a list of patterns that does not match the variable containing the secret", cfg.Job)
+			l.Warnf("The promise-failure reason for job %q contained one or more secrets from environment variables that have been redacted. If this is deliberate, pass --redacted-vars='' or a list of patterns that does not match the variable containing the secret", cfg.Job)
 			cfg.Reason = redactedValue
 		}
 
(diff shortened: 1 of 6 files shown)


jordandcarter:
@buildsworth-bk review

clicommand/job_promise_failure.go
```diff
@@ -0,0 +1,139 @@
...
+const jobPromiseFailureHelpDescription = `Usage:
+
+    buildkite-agent job promise-failure <exit-status> [options...]
+
+Description:
+
+Promise the current job will finish with a failing exit status. This records
+a non-zero exit status that the job is expected to finish with, allowing the
+build to begin failing before the job actually completes. The job keeps
+running and finishes normally; only the build-failing cascade starts early.
```
moskyb: i think it's probably worth capturing here (if this is true?) that if you promise a non-zero status and then return a zero status, then the job is still treated as a failure by the control plane. likewise with promising `exit 1` but returning `exit 2` etc.

clicommand/job_promise_failure.go
```diff
@@ -0,0 +1,139 @@
...
+
+Repeated calls with the same exit status are idempotent. Declaring a
+different exit status once one is already recorded is rejected.
+
+This command requires the early-failure feature to be enabled for your
+organization. If it is not, the command logs a warning and exits
+successfully, so it can safely be added to pipelines ahead of rollout.
+
```
moskyb: we generally try to avoid talking about backend rollouts in the agent code/docs. obviously we still do them, but paragraphs like this one are very easy to forget about, and then all of a sudden it's two years later and the command is still marked experimental (ask me how i know)

the approach we generally take is to hide the command à la [the `cache` commands](https://github.com/buildkite/agent/blob/main/clicommand/commands.go#L58) until they're fully rolled out. This way we can communicate to customers that we've enabled it for "hey, you can use this now" without letting all but the most dedicated changelog watchers know about it.

clicommand/job_promise_failure.go
```diff
@@ -0,0 +1,139 @@
...
+    $ buildkite-agent job promise-failure 1
+    $ buildkite-agent job promise-failure 42 --reason "detected failing tests"
+`
+
+type JobPromiseFailureConfig struct {
+	GlobalConfig
+	APIConfig
+
+	ExitStatus   string   `cli:"arg:0" label:"exit status" validate:"required"`
```
moskyb, line 50: i _think_ if you set this to 
```suggestion
	ExitStatus   int   `cli:"arg:0" label:"exit status" validate:"required"`
```

it should parse the int for you, so you won't have to do the `strconv.Atoi` call below

clicommand/job_promise_failure.go
```diff
@@ -0,0 +1,139 @@
...
+		client := api.NewClient(l, loadAPIClientConfig(cfg, "AgentAccessToken"))
+
+		needles, _, err := redact.NeedlesFromEnv(cfg.RedactedVars)
+		if err != nil {
+			return err
+		}
+		if redactedValue := redact.String(cfg.Reason, needles); redactedValue != cfg.Reason {
+			l.Warnf("The reason for job %q contained one or more secrets from environment variables that have been redacted. If this is deliberate, pass --redacted-vars='' or a list of patterns that does not match the variable containing the secret", cfg.Job)
+			cfg.Reason = redactedValue
+		}
```
moskyb, line 98: love to see this 👍

clicommand/job_promise_failure.go
```diff
@@ -0,0 +1,139 @@
...
+				// first declaration already took effect.
+				l.Warnf("A different promised exit status has already been declared for this job; ignoring")
+				return nil
+
+			case api.IsErrHavingStatus(err, http.StatusUnprocessableEntity):
+				// Most likely the job is no longer running (cancelled or timed
+				// out) by the time this declaration arrived: a benign race.
+				l.Warnf("Could not declare promised failure (the job may no longer be running); ignoring")
+				return nil
+			}
```
moskyb, line 129: i have slightly mixed feelings about this. i can see that it would be useful for people to add this to their pipeline prior to us enabling it on the backend, but realistically we have enough problems getting customers to use new features _after_ they're in general availability. Additionally, for us, the act of enabling a feature for a customer is usually a one minute job, and something that we can do pre-emptively (ie, before they add code to their build scripts) very easily.

my other observation here is that by handling these errors for the customer (returning status 0 for all of these cases) the caller of `buildkite-agent job promise-failure 123` actually loses context.

If this command exits non-zero for a `404`, `409`, or `422`, a customer who considers those cases acceptable can still opt out very easily:

```sh
buildkite-agent job promise-failure 123 || true
```

but if the command treats these API responses as success, then customers who _do_ care have to inspect stdout/stderr to know whether the promise was actually registered, which is a more awkward interface than just checking the process exit code.

all this in mind, i think i'd prefer that the command fail when the API did not accept the promised failure, even for "expected" rollout/race cases. We can make those failures non-scary and well explained, but i'd be hesitant to collapse them into success because it makes the command less composable in scripts.

moskyb (changes requested):
this looks good! i have a couple of nits about docs and stuff, and one juicier comment about error handling, but i'm really psyched to get this merged

clicommand/job_promise_failure.go
jordandcarter: I tried that but cliconfig doesn't coerce positional args — `cli:"arg:0"` assigns the raw string straight onto the field.

moskyb: ah bugger. as you were then!

clicommand/job_promise_failure.go
```diff
@@ -0,0 +1,138 @@
...
+	}),
+	Action: func(c *cli.Context) error {
+		ctx, cfg, l, _, done := setupLoggerAndConfig[JobPromiseFailureConfig](context.Background(), c)
+		defer done()
+
+		exitStatus, err := strconv.Atoi(cfg.ExitStatus)
+		if err != nil {
+			return fmt.Errorf("exit status must be an integer: %w", err)
+		}
+		if exitStatus == 0 {
```
moskyb: probably worth also ensuring that the exit code is positive — we only want to handle positive exit codes, and the `-1` exit code is reserved for internal use

jordandcarter: Ah yes, good point

clicommand/job_promise_failure.go
```diff
@@ -0,0 +1,140 @@
...
+		}
+
+		client := api.NewClient(l, loadAPIClientConfig(cfg, "AgentAccessToken"))
+
+		needles, _, err := redact.NeedlesFromEnv(cfg.RedactedVars)
+		if err != nil {
+			return err
+		}
+		if redactedValue := redact.String(cfg.Reason, needles); redactedValue != cfg.Reason {
+			l.Warnf("The reason for job %q contained one or more secrets from environment variables that have been redacted. If this is deliberate, pass --redacted-vars='' or a list of patterns that does not match the variable containing the secret", cfg.Job)
```
moskyb: often (like when getting called from within a script) the command that emits a warning isn't visible in the output, so it's useful to include as much context as possible in warnings/error messages:

```suggestion
			l.Warnf("The promise-failure reason for job %q contained one or more secrets from environment variables that have been redacted. If this is deliberate, pass --redacted-vars='' or a list of patterns that does not match the variable containing the secret", cfg.Job)
```

clicommand/job_promise_failure.go
```diff
@@ -0,0 +1,140 @@
...
+			}
+			return nil
+		})
+		if err != nil {
+			// The promise wasn't accepted. Exit non-zero so the outcome is
+			// visible to scripts; callers who consider a given case acceptable
+			// can append '|| true'.
+			switch {
+			case api.IsErrHavingStatus(err, http.StatusNotFound):
+				return fmt.Errorf("promised failures are not enabled for this organization: %w", err)
```
moskyb: a 404 isn't always going to mean that the feature is disabled, right? like, if agents are behind a proxy (many such cases) that doesn't recognise this call, it might return a 404 even though the feature is enabled for this org. 

i might try to soften the message a little bit, or even just remove this case entirely. if we tell a customer to try using `promise-failure`, and they say all the calls are 404ing, then it's going to be relatively clear what's going on

moskyb (approved):
two minor error message nits, but this looks really good! awesome work @123sarahj123 @jordandcarter

clicommand/job_promise_failure.go
jordandcarter: Good point!

PR #3993, opened 2026-06-11
feat: migrate agent cache CLI to cache v2 protocol

ss1909:
## Description

Implements the agent side of the v2 cache redesign (backend: [A-1289](https://linear.app/buildkite/issue/A-1289/migrate-agent-cache-api-to-v2-addressing-target-paths-keys-cluster)). The cache CLI now parses v2 `cache.yml`, speaks the v2 wire protocol, and uses content-addressed S3 storage.

## Context

https://linear.app/buildkite/issue/A-1293/agent-cli-parse-v2-cacheyml-and-speak-v2-wire-protocol

## Changes

- **Parse v2 `cache.yml`:** `caches:` with a structured `cache_key` (literal, `{agent: os/arch}`, `{checksum: <file>}`, `{env: VAR}`) and `target_paths`; removed `templates.json` and the `html/template` key engine.
- **v2 wire protocol** (`api/cache.go`): `target_paths` + structured `cache_key` + `blobs[]` with a typed digest; `peek`/`retrieve` change `GET → POST`.
- **Content-addressed storage:** the S3 object name is the blob digest; upload/download go directly to S3 with ambient credentials.
- **Env vars:** registry slug from `BUILDKITE_AGENT_CACHE_REGISTRY` (default `~`); store URL from `BUILDKITE_AGENT_CACHE_STORE_URL`.
- Map the `agent_managed` store adapter to the S3 backend.
- Update unit/integration tests and `cache save`/`restore` help text.

## Testing

**End-to-end tested locally** against my test org (`queue-test`), local backend + DynamoDB + MinIO:

1. Created `.buildkite/cache.yml`:

   ```yaml
   caches:
     - id: deps
       cache_key: [ v1, { agent: os }, { agent: arch }, { checksum: go.sum } ]
       target_paths: [ vendor ]
   ```

2. Ran `cache save --ids deps` against a local build of the agent with the cache env vars set:
image

3. Ran `cache restore --ids deps` (after wiping `vendor/`) and verified the files were restored:

image

### What `cache save --ids deps` does

- **Load + validate** — reads `.buildkite/cache.yml`, finds the `deps` entry.
- **Resolve the key → the address:** `v1`, `{agent: os}`→`darwin`, `{agent: arch}`→`arm64`, `{checksum: go.sum}`→ sha256 of `go.sum`. Becomes `cache_key: [{value, mandatory:true}, …]`.
- **Peek** (`POST …/peek`) — "does an entry already exist at `(target_paths=[vendor], cache_key)`?" If yes → bail early (don't re-upload). If no (404) → continue.
- **Archive** — tar+zstd the `target_paths` (`vendor/`), compute the archive's SHA-256 (that's the blob digest).
- **Store** (`PUT …/store`) — send the metadata (`target_paths`, `cache_key`, `blobs:[{digest, file_size, compression}]`, branch/pipeline/owner) → backend writes a temp DynamoDB entry, returns an `upload_id`.
- **Upload** — push the archive bytes to `s3://<bucket>/<digest>` (MinIO) with ambient credentials.
- **Commit** (`PUT …/commit`) — finalise the entry so it's retrievable.

**Net effect:** package up `vendor/`, address it by `(vendor + the resolved key)`, store the bytes in S3 under their digest, and record the metadata entry in the backend — so a later `cache restore --ids deps` with the same key can fetch it back. The env vars supply the "where/who".

### What `cache restore --ids deps` does

- **Load + validate** — reads `.buildkite/cache.yml`, finds the `deps` entry.
- **Resolve the key → the address** (identical to save): same inputs ⇒ same address ⇒ it can find what save stored.
- **Retrieve** (`POST …/retrieve`) — "is there an entry at `(target_paths=[vendor], cache_key)`?" **Exact match only** in M1 (no fallback walking). On a **hit (200)** the backend bumps the entry's TTL and returns `blobs[]`, `store`, `expires_at`, `fallback=false`. On a **miss (404)** it's a normal cache miss — restore returns `CacheRestored=false` and the job continues *without* a cache (not an error).
- **Validate store config** — confirm `BUILDKITE_AGENT_CACHE_STORE_URL` is set for the returned store adapter (`agent_managed` → S3).
- **Download** — read `blobs[0].digest.value` (the digest = the object name) → `GET s3://<bucket>/<digest>` from MinIO with ambient credentials → a temp file.
- **Clean paths** — remove the existing `target_paths` (`vendor/`) first, so extraction lands on a clean tree (handles read-only dirs like module caches).
- **Extract** — unarchive the downloaded zstd into the `target_paths` (`vendor/`).

**Net effect:** compute the same address from the resolved key, look up the entry in the backend, download the blob by its content digest from S3, then wipe and re-extract `vendor/` — reconstructing exactly what `cache save` stored.

### How it mirrors save (the round trip)

```
save:    key → address;  archive vendor/ → digest;  store metadata;  upload bytes @ digest;  commit
restore: key → SAME address;  retrieve metadata → digest;  download bytes @ digest;  extract → vendor/
```

The **resolved key** is the shared join point (find the same entry), and the **blob digest** is the shared pointer to the bytes (find the same S3 object). That's why the restore run showed `checking_exists` (retrieve hit) → `downloading` (by digest) → `cleaning` → `extracting` → "Cache restored successfully", with `vendor/a.txt` + `vendor/example/b.txt` restored.

### Two differences vs save

- **No peek/archive/upload** — restore is just retrieve → download → extract (1 backend call + 1 storage read), whereas save is peek → archive → store → upload → commit (3 backend calls + 1 storage write).
- **A miss is success-with-no-cache**, not a failure — the job proceeds and (typically) a later `save` populates it.

## Disclosures / Credits

Claude was used as a pairing tool throughout — for the implementation, tests, and this PR description. I directed the design decisions, reviewed every change, and verified the behaviour end-to-end against a local backend + DynamoDB + MinIO.

## Notes:
-  the backend owns the cache_key separator invariant - it rejects any resolved value containing '#' (CACHE_KEY_SEPARATOR) with a clear error , so we don't duplicate that check (and its hardcoded separator) here.


ss1909:
@buildsworth-bk review

clicommand/cache_shared.go
```diff
@@ -26,14 +26,14 @@ func cacheFlags() []cli.Flag {
...
+			Usage:  "The slug of the cache registry to use; '~' selects the cluster's default registry",
+			EnvVar: "BUILDKITE_AGENT_CACHE_REGISTRY",
 		},
 		cli.StringFlag{
 			Name:   "bucket-url",
 			Value:  "",
-			Usage:  "The URL of the bucket (e.g., s3://bucket-name)",
-			EnvVar: "BUILDKITE_CACHE_BUCKET_URL",
+			Usage:  "The URL of the cache store bucket (e.g., s3://bucket-name); uploads/downloads use ambient credentials",
+			EnvVar: "BUILDKITE_AGENT_CACHE_STORE_URL",
```
buildkate, line 36: nit: the env var name doesn't match the cli flag `bucket-url` vs `..._CACHE_STORE_URL`

api/cache.go
```diff
@@ -32,75 +31,81 @@ var ErrCacheEntryNotFound = errors.New("cache entry not found")
...
+	Fallback             bool           `json:"fallback"`
+	Multipart            bool           `json:"multipart"`
+	DownloadInstructions []string       `json:"download_instructions"`
+	Message              string         `json:"message"`
 }
 
-// CacheEntryCreateResp describes where and how to upload the new cache entry.
+// CacheEntryCreateResp describes how to upload the new cache entry. The storage
+// object name is not echoed back — the agent derives it from the blob digest.
 type CacheEntryCreateResp struct {
```
buildkate: nit: can we move this under `CacheEntryCreateReq`

clicommand/cache_save.go
```diff
@@ -36,20 +35,25 @@ specific caches by providing their IDs:
 
     $ buildkite-agent cache save --ids "node"
 
-The cache will be stored in the bucket specified by --bucket-url or your
-cache configuration. If a cache with the same key already exists, it will
-not be overwritten.
+The cache is stored at BUILDKITE_AGENT_CACHE_STORE_URL (or --cache-store-url).
```
buildkate, line 38: probably for another PR but this is duplicated across cache_restore.go and here. Not sure where's the right home for it.

internal/cache/configuration/key_part.go
```diff
@@ -0,0 +1,162 @@
...
+		return "", fmt.Errorf("cache_key: unknown source %q", k.Source)
+	}
+	if err != nil {
+		return "", err
+	}
+	// The backend joins cache_key values with '#' to form the DynamoDB sort key
+	// (CACHE_KEY_SEPARATOR), so a '#' in any resolved value would corrupt the
+	// address and is rejected server-side.
+	if strings.Contains(v, "#") {
+		return "", fmt.Errorf("cache_key: %s resolved to a value containing '#': %q", k.Source, v)
```
buildkate: maybe we need stronger language here.. something along the lines of  "cache_key: %s cannot contain #"

clicommand/cache_save.go
ss1909: the line in cache_restore.go reads as "The cache is retrieved from BUILDKITE_AGENT_CACHE_STORE_URL (or --cache-store-url)". To me, having it in both places makes sense cause one is about where its saved and the other is about where its restored from.

buildkate (approved):

internal/cache/store/blob.go
```diff
@@ -16,7 +16,7 @@ type Blob interface {
 
 func NewBlobStore(ctx context.Context, store, bucketURL string) (Blob, error) {
 	switch store {
-	case LocalS3Store:
+	case LocalS3Store, AgentManaged:
```
zhming0: [not-blocking] do we need to keep the LocalS3Store and LocalFileStore?

internal/cache/cache_test.go
```diff
@@ -370,14 +372,16 @@ func TestNewClient_InvalidCacheIDs(t *testing.T) {
 func TestNewClient_ValidCacheIDs(t *testing.T) {
 	t.Parallel()
 
-	config := `dependencies:
+	config := `caches:
   - id: cache1
```
zhming0, line 376: It can come in a subsequent PR. But the Linear card wasn't clear that it should be `name:` instead of `id:`.

internal/cache/configuration/configuration.go
```diff
@@ -1,257 +1,16 @@
...
-
-* Expands cache.Paths using templatable arguments (such as id, agent.os, agent.arch, env, checksum etc)
-
-Uses the OS environment variables for template expansion.
-*/
+// ExpandCacheConfiguration is the load-time pass over the parsed cache
+// definitions.
+//
+// In v1 this expanded `template:` references and ran the html/template key
+// engine. In v2 there is nothing to expand at load time: a cache_key is an
```
zhming0: Same with the comments in your prior PRs, let's avoid editorial comments that are more suitable as PR comments instead of code comments 🙏🏿 

(if your agent is making the same issue many times, i might be worth considering a root AGENTS.md directive)

internal/cache/restore.go
```diff
@@ -252,9 +264,15 @@ func (c *client) downloadCache(ctx context.Context, retrieveResp api.CacheEntryR
 	ctx, span := tracer.Start(ctx, "Client.downloadCache")
 	defer span.End()
 
+	// Content-addressed storage: the object name is the blob digest.
+	if len(retrieveResp.Blobs) == 0 {
+		return "", "", nil, fmt.Errorf("cache entry has no blobs to download")
+	}
+	storeObjectName := retrieveResp.Blobs[0].Digest.Value
```
zhming0, line 275: [not blocking] It might deserve a comment to explain why we only consider `blobs[0]` for now.

internal/cache/integration_test.go
```diff
@@ -159,37 +159,18 @@ func (m *mockAPIClient) CacheEntryRetrieve(ctx context.Context, registry string,
 		return api.CacheEntryRetrieveResp{}, false, nil, fmt.Errorf("registry not found: %s", registry)
 	}
 
-	// Try exact key match first
-	if entry, exists := reg.cache[req.Key]; exists && entry.committed {
+	// M1: exact address match only (backend fallback walking is out of scope).
```
zhming0: This comment isn't accurate 🙂

internal/cache/configuration/key_part.go
```diff
@@ -0,0 +1,162 @@
...
+			if val.Kind != yaml.ScalarNode {
+				return fmt.Errorf("cache_key: agent source must be a string")
+			}
+			if _, ok := supportedAgentArgs[val.Value]; !ok {
+				return fmt.Errorf("cache_key: unsupported agent argument: %s", val.Value)
+			}
+			k.Source, k.Arg = SourceAgent, val.Value
+		case SourceChecksum:
+			if val.Kind != yaml.ScalarNode {
+				return fmt.Errorf("cache_key: checksum takes a single file path in M1 (no arrays)")
```
zhming0: Let's try to avoid exposing terms like `M1` to users 🙏🏿 , if necessary, we can leave a code comment here instead.

internal/cache/configuration/key_part.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+	if err != nil {
+		return "", err
+	}
+	// The backend joins cache_key values with '#' to form the DynamoDB sort key
+	// (CACHE_KEY_SEPARATOR), so a '#' in any resolved value would corrupt the
+	// address and is rejected server-side.
+	if strings.Contains(v, "#") {
+		return "", fmt.Errorf("cache_key: %s cannot contain '#'", k.Source)
+	}
```
zhming0: Defense in depth is good, but it comes at a cost of maintenance, what happen if we change the separator in backend? 

Another train of thought is that, the separator should occur rare enough that we don't need to set a specific defense for. 

Is it worth to reconsider if we need this defense?

internal/cache/configuration/key_part.go
```diff
@@ -0,0 +1,162 @@
...
+			return "", fmt.Errorf("cache_key: unsupported agent fact %q", k.Arg)
+		}
+	case SourceEnv:
+		if env != nil {
+			v = env[k.Arg]
+		} else {
+			v = os.Getenv(k.Arg)
+		}
+	case SourceChecksum:
+		v, err = sha256File(k.Arg)
```
zhming0, line 135: It might be worth adding a comment that it's temporary that we only support single file. 

Also the fact that we didn't consider anything like directory, symlink, very big file, etc.

zhming0 (approved):
LGTM structurally, I left some minor non-blocking comments.

internal/cache/store/blob.go
ss1909: 👍 
backend only sends agent_managed and local_hosted_agents (ref: https://github.com/buildkite/buildkite/blob/b095b51a95fce4f33161802d27e0f48f51aa4d4d/app/models/cache_registry/cache_store.rb#L26-L40). Some of the tests rely on the existence of LocalFileStore, so I'll keep that and remove LocalS3Store

internal/cache/cache_test.go
ss1909: I'll change this in a subsequent PR

internal/cache/restore.go
ss1909: done

internal/cache/integration_test.go
ss1909: fixed

internal/cache/configuration/key_part.go
ss1909: missed this one. I tried to remove all of them 😅

internal/cache/configuration/key_part.go
ss1909: done

internal/cache/configuration/key_part.go
ss1909: yeah, makes sense. Removing this as it is already handled in the backend here https://github.com/buildkite/buildkite/blob/b095b51a95fce4f33161802d27e0f48f51aa4d4d/app/models/cache_registry/entry/key_part.rb#L9

PR #3994, opened 2026-06-12
Fix delayed disconnect-after-job in streaming ping mode

maleadt:
## Description

In `--disconnect-after-job` mode with the streaming ping protocol (the default since ping mode `auto` prefers streaming), the agent can idle for up to a minute after its job finishes before actually disconnecting.

The action handler loop deliberately evaluates the disconnect-after-job condition only when handling the *next* action after a job, so that a pending "pause" can keep the agent alive (93768548). The classical ping loop accounts for this by skipping the ping interval and pinging immediately after a job (605ff871), producing that next action promptly. Streaming mode has no equivalent: nothing arrives until the server pushes another message, and the client has no read deadline on the ping stream, so the agent just sits there — in our production setup, consistently ~57–60 seconds per job, which dominates the runtime of short jobs on ephemeral one-job-per-container agents.

Observed timeline (agent v3.127.2):

```
13:43:55  Ping stream connection established
13:43:58  Finished job ... (exit 0)
13:44:55  Job ran, and disconnect-after-job is enabled. Disconnecting...
```

This PR fixes it in the debouncer, which is the component that knows whether another action (such as a pause received during the job) is already pending: when a job action completes successfully in disconnect-after-job mode and nothing else is pending, it synthesises an idle action so the action handler loop can disconnect immediately. This mirrors poll mode's "ping once right after the job" behaviour, so the two modes now have aligned semantics — including the (pre-existing, identical in poll mode) race where a pause pushed after the job finishes may arrive too late.

Alternatives considered:

- Returning from the action loop directly after `AcceptAndRunJob` — simpler, but loses the deliberate pause-keeps-agent-alive semantics from 93768548.
- A client-side read deadline / local timer on the stream — heavier-handed, and changes behaviour for all agents rather than just disconnect-after-job.
- (As a user-side workaround, `--ping-mode=poll-only` avoids the issue on current releases.)

## Context

Found while operating the JuliaGPU Buildkite cluster, which runs untrusted open-source PR workloads on self-hosted hardware. To isolate jobs from one another, each agent runs inside a Docker container and `--disconnect-after-job` is what drives the recycle: the agent exits after its single job, the container exits with it, and systemd brings up a fresh container for the next job. With the streaming linger, that per-job recycle gained a ~60s tax, reducing a single-queue agent to roughly one 3-second job per minute. No GitHub issue filed; happy to open one if preferred.

## Changes

- `agent/agent_worker_debouncer.go`: track whether the most recently dispatched action carried a job; when it completes without error in disconnect-after-job mode and no other action is pending, synthesise an idle action. Pause/resume debouncing and ignore-in-dispatches behaviour are unchanged.
- `agent/agent_worker_test.go`: add `TestAgentWorker_Streaming_DisconnectAfterJob_PromptDisconnect`, which streams a job assignment and then goes quiet. Without the fix the worker never disconnects (in production it waits for the server's next keepalive); with it, it disconnects immediately. The existing streaming disconnect-after-job test didn't catch this because its fake server pushes follow-up messages right after the job.

No CLI changes.

## Testing

JuliaGPU / JuliaHub — this affects the CI infrastructure for the Julia GPU ecosystem.

## Disclosures / Credits

Claude Code (Claude Fable 5) diagnosed the issue from production logs, and wrote the fix and the regression test; I reviewed the change. The fix is verified by the regression test (which hangs without it), not yet by a production deployment.

### DIFF
--- a/agent/agent_worker_debouncer.go
+++ b/agent/agent_worker_debouncer.go
@@ -59,6 +59,10 @@ func (a *AgentWorker) runDebouncer(ctx context.Context, bat *baton, outCh chan<-
 	// pending is true when the action is pending, false otherwise
 	var pending bool
 
+	// lastActionWasJob is true when the most recently sent action included a
+	// job to run. Used for disconnect-after-job, below.
+	var lastActionWasJob bool
+
 	// Is the stream healthy?
 	// If so, take the baton (which blocks the ping loop).
 	// If not, return the baton (unblocking the ping loop).
@@ -139,12 +143,26 @@ func (a *AgentWorker) runDebouncer(ctx context.Context, bat *baton, outCh chan<-
 
 			// continue below to send it
 
-		case <-lastActionResult: // most recent action has completed
+		case err := <-lastActionResult: // most recent action has completed
 			a.logger.Debugf("[runDebouncer] Last action has completed")
 			// Set the channel variable to nil so we don't spinloop.
 			// (Operations on a nil channel block forever.)
 			lastActionResult = nil
 			actionInProgress = false
+
+			// In disconnect-after-job mode, the action handler loop only
+			// disconnects when it handles the next action after running a job
+			// (this gives a pending "pause" the chance to keep the agent
+			// alive). The ping loop pings the backend immediately after a job,
+			// so it produces that next action promptly, but the stream might
+			// not deliver another message for a while. So if a job just ran
+			// and nothing else is pending, synthesise an idle action to let
+			// the action handler loop disconnect promptly.
+			if a.agentConfiguration.DisconnectAfterJob && lastActionWasJob && err == nil && !pending {
+				a.logger.Debugf("[runDebouncer] Job ran in disconnect-after-job mode; enqueueing a synthetic idle action")
+				pending = true
+				pendingAction = ""
+			}
 
 			// continue below to send a pending message
 		}
@@ -169,6 +187,7 @@ func (a *AgentWorker) runDebouncer(ctx context.Context, bat *baton, outCh chan<-
 		case outCh <- newMsg:
 			// sent!
 			pending = false
+			lastActionWasJob = newMsg.jobID != ""
 			if newMsg.jobID != "" {
 				pendingJobs = pendingJobs[1:]
 				pending = len(pendingJobs) > 0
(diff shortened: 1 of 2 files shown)


agent/agent_worker_debouncer.go
```diff
@@ -139,13 +143,27 @@ func (a *AgentWorker) runDebouncer(ctx context.Context, bat *baton, outCh chan<-
...
+			// alive). The ping loop pings the backend immediately after a job,
+			// so it produces that next action promptly, but the stream might
+			// not deliver another message for a while. So if a job just ran
+			// and nothing else is pending, synthesise an idle action to let
+			// the action handler loop disconnect promptly.
+			if a.agentConfiguration.DisconnectAfterJob && lastActionWasJob && err == nil && !pending {
+				a.logger.Debugf("[runDebouncer] Job ran in disconnect-after-job mode; enqueueing a synthetic idle action")
+				pending = true
+				pendingAction = ""
+			}
```
DrJosh9000: This argument also applies to acquire-job. Also the `err == nil` is stylistically a bit weird - I think we should break out of the case early. So:

```suggestion
			if err != nil {
				break
			}
			ephemeral := a.agentConfiguration.AcquireJob != "" || a.agentConfiguration.DisconnectAfterJob
			if ephemeral && lastActionWasJob && !pending {
				a.logger.Debugf("[runDebouncer] Job ran in an ephemeral mode; enqueueing a synthetic idle action")
				pending = true
				pendingAction = ""
			}
```

DrJosh9000 (approved):
LGTM, good find and thank you for jumping in with a fix. I have one comment.

agent/agent_worker_debouncer.go
maleadt: AFAIU extending the condition would be a no-op, since acquire-job agents run their single job in AcquireAndRunJob before the loops start?

DrJosh9000: Right, yes, and we ensure there's a message when the stream starts. So it is just the disconnect-after-job scenario.

maleadt: OK, I've added a comment and moved `err != nil`.

DrJosh9000 (approved):

Issue #2790, opened 2024-05-23
Buildkite Agent PTY randomly failed with fork/exec... operation not permitted


leakingtapan:
## The problem
The buildkite agent could randomly fail the job with following error message:

```
2024-05-20 11:15:52 PDT $ git clone -v -- ssh...
...
...
2024-05-20 11:15:52 PDT 🚨 Error: Failed to checkout plugin docker: error running "/usr/bin/git checkout -f e9efccb": error starting pty: fork/exec /usr/bin/git: operation not permitted
2024-05-20 11:15:52 PDT Running global pre-exit hook
```

```
2024-05-21 09:28:12 PDT ⚠️ Warning: Checkout failed! cleaning repository post-checkout: error running "/usr/bin/git clean -ffxdq": error starting pty: fork/exec /usr/bin/git: operation not permitted (Attempt 1/3 Retrying in 2s)
```

```
2024-05-20 10:37:40 PDT 🚨 Error: error running "/bin/bash -e -c \"trap 'kill -- $$' INT TERM QUIT; if [ -n \\\"$(git diff HEAD~1 --exit-code 'cruise/ai_platform/search')\\\" ]; then\\n  cat <<- ...\"": error starting pty: fork/exec /bin/bash: operation not permitted
```

I worked with BK support on this issue for long time, and we are able to pin down the issue being the [github.com/creack/pty](https://github.com/creack/pty/) PTY package that's used. Although v1.1.21 reverted a bug v1.1.20 that's mostly related to this bug. The issue is not fully resolved in v1.1.21.

I tested following BK agent version:
* v3.59.0 uses pty v1.1.20
* v3.63.1 uses pty v1.1.21
all has the same issue, just v3.3.60 has a lower frequency. 

I tested v3.63.1 with pty v1.1.19 and the problem no longer happens. So there should be something something wrong with pty v1.1.21 still.

## Proposed solution
pin agent's pty package version to v1.1.19

cc @DrJosh9000 @steveh

PR #3995, opened 2026-06-12
pin creack/pty version in go mod

jeremybumsted:
## Description
Pins the version of `github.com/creak/pty` to v1.1.19 to fix issues with 

## Context
This particular package has caused problems in the past resulting in failures during hook execution (see #2790).
Looks like we may have bumped to a later version at some point, though there seems to be evidence this is not resolved:

https://github.com/creack/pty/issues/196

and as of `v1.1.23` (not `v1.1.24`) there is only a partial revert of the change introduced in https://github.com/creack/pty/pull/167 that seems to be the cause of the errors.

This PR should pin the dependency version back to a known working version; there are no changes in later versions of this package that would affect our usage of the pinned version, so this should be safe to pin.

## Testing
- [ x ] Tests have run locally (with `go test ./...`). Buildkite employees may check this if the pipeline has run automatically.
- [ x ] Code is formatted (with `go tool gofumpt -extra -w .`)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -14,7 +14,7 @@ updates:
       container-images:
         patterns:
           - "*"
-          
+
   - package-ecosystem: docker
     directories:
       - /.buildkite
@@ -30,6 +30,8 @@ updates:
     schedule:
       interval: weekly
     open-pull-requests-limit: 10
+    ignore:
+      - dependency-name: "github.com/creack/pty"
     groups:
       otel:
         patterns:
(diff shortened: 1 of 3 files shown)


DrJosh9000 (approved):
This changes the version depended on, but won't stop dependabot opening a PR to update it again. You might need to add to the dependabot config.

jeremybumsted:
@DrJosh9000 Oh yeah nice catch - updated dependabot so it will ignore this dependency

DrJosh9000 (approved):

PR #3996, opened 2026-06-15
Debounce `buildkite-agent job promise-failure` via the Job API

jordandcarter:
## Description

Debounce repeated `buildkite-agent job promise-failure` calls in the agent so each exit status results in at most one Buildkite API call per job.

Customers are likely to call `promise-failure` from their test suite — potentially on every failing test. Each call spawns a new `buildkite-agent` process and would otherwise make its own API request, all of which the server must authenticate and process before discovering there's nothing new to record. For a suite with thousands of failures that's a lot of useless load on the Agent API.

This change moves the debouncing into the agent. The Job API server owns the declaration: each `promise-failure` invocation POSTs its exit status to the Job API, and the server declares the failure to the Buildkite API at most once per exit status. The first caller for a given exit status declares it, **blocking on the API** so it can return an accurate result; concurrent callers for the same exit status wait and share that outcome; later callers return from the cached success. Terminal failures (4xx other than 429, e.g. 409 or 422) are cached too, so repeated calls don't keep re-sending a declaration the API will always reject; transient failures (5xx, network, 429) aren't cached, so a later call can retry. Because the caller blocks until the result is known, the command exits with an accurate code rather than guessing.

The response reports an `outcome` of `declared` (this call reached the Buildkite API) or `debounced` (it shared an earlier declaration), and the CLI logs the distinction.

The Job API is job-scoped and enabled by default, so the state is naturally per-job (no keying by job ID) and there's no experiment to opt into. When the Job API is unavailable (`--no-job-api`, unsupported Windows) or can't be reached, the command falls back to declaring directly; the endpoint is idempotent for the same exit status, so a duplicate is safe.

This PR also drops the `--job` flag added to the command in #3992. The endpoint requires the job's own token and rejects declarations for any other job, so a promised failure only ever makes sense for the current job — the command now always targets `BUILDKITE_JOB_ID`.

## Context

Follows on from #3992, which added the `buildkite-agent job promise-failure` command.

## Changes

- `jobapi`: the `POST /api/current-job/v0/promise-failure` endpoint now declares the promised failure to the Buildkite API itself, coalescing concurrent and repeated calls so each exit status is declared at most once successfully. The declarer is injected via `WithPromiseFailureDeclarer`. The response carries an `outcome` (`declared`/`debounced`).
- `clicommand`: declare the promised failure through the Job API (blocking for an accurate result) and log whether it was declared or debounced; fall back to declaring directly when the Job API is unavailable or unreachable. Drop the `--job` flag and always target the current job.

## Testing

Added server tests over a real socket — sequential (first call declares, repeat is debounced, a different status declares on its own), concurrent (many simultaneous calls of the same status yield exactly one API declaration, all callers see success), error handling (terminal failures are cached, transient failures aren't and can retry; status normalization to 502), and panic safety — plus a client test for `DeclarePromiseFailure`.

## Disclosures / Credits

I used Amp to design and implement the debouncing and tests, with GPT-5 for a code review pass. I reviewed all of the changes myself.

### DIFF
--- a/clicommand/job_promise_failure.go
+++ b/clicommand/job_promise_failure.go
@@ -2,6 +2,7 @@ package clicommand
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"os"
@@ -11,6 +12,7 @@ import (
 
 	"github.com/buildkite/agent/v3/api"
 	"github.com/buildkite/agent/v3/internal/redact"
+	"github.com/buildkite/agent/v3/internal/socket"
 	"github.com/buildkite/agent/v3/jobapi"
 	"github.com/buildkite/agent/v3/logger"
 	"github.com/buildkite/roko"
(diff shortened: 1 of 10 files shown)


jordandcarter:
@buildsworth-bk review

jordandcarter:
@buildsworth-bk review

clicommand/job_promise_failure.go
```diff
@@ -86,14 +83,37 @@ var JobPromiseFailureCommand = cli.Command{
...
+		// API is unavailable (--no-job-api, unsupported Windows, or run outside
+		// a job), we fall back to declaring directly.
+		if jobAPI, err := jobapi.NewDefaultClient(ctx); err != nil {
+			l.Debugf("Job API unavailable, declaring promised failure without debouncing: %v", err)
+		} else if claimed, err := jobAPI.PromiseFailureClaim(ctx, exitStatus); err != nil {
+			l.Warnf("Couldn't reach the Job API to debounce the promised failure; declaring it directly: %v", err)
+		} else if !claimed {
+			l.Debugf("Promised exit status %d has already been claimed; skipping duplicate", exitStatus)
+			return nil
+		}
```
moskyb: this error flow is pretty confusing and not very idiomatic for go — i'd probably extract this into a function like

```Go
func shouldSkip(ctx context.Context, l logger.Logger status int) bool {
	jobAPI, err := jobapi.NewDefaultClient(ctx)
	if err != nil {
		l.Debugf("Job API unavailable, declaring promised failure without debouncing: %v", err)
		return false
	}

	claimed, err := // ...
```

then just return nil if `shouldSkip` returns true

moskyb (commented):
love love love this. it's the exact kind of thing we should be using the job API for. i have one readability thing, but other than that this looks great.

jordandcarter:
@buildsworth-bk review

clicommand/job_promise_failure.go
```diff
@@ -131,7 +146,49 @@ var JobPromiseFailureCommand = cli.Command{
...
+// The status is only recorded after a successful declaration, so a failed
+// attempt isn't cached and a later call will retry.
+func promiseFailureDebounce(ctx context.Context, l logger.Logger, exitStatus int) (jobAPI *jobapi.Client, alreadyDeclared bool) {
+	client, err := jobapi.NewDefaultClient(ctx)
+	if err != nil {
+		l.Debugf("Job API unavailable, declaring promised failure without debouncing: %v", err)
+		return nil, false
+	}
+
+	declared, err := client.PromiseFailureCheck(ctx, exitStatus)
```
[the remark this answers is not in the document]
jordandcarter: I've reverted back to the claim-before-attempt approach

jordandcarter:
@buildsworth-bk review

clicommand/job_promise_failure.go
jordandcarter: Before seeing this comment I went on a journey and switched from claim-before-attempt (the code you reviewed) to claim-on-success, then decided it wasn't worth the complexity and have reverted back to what you reviewed.

Plus I addressed your error flow improvement 👍

moskyb:
as written it seems like the flow here is (and please correct me if i'm wrong):
- Job calls `promise-failure`
- `promise-failure` checks with the job API to see if the exit status has been claimed for this job
- if the exit hasn't been claimed, it sends the HTTP request off to the agent API

in this situation, if the agent API returns an error for the first promised failure, doesn't that mean that all subsequent `promise-failure` calls will fail with a "exit status has already been claimed" message, even though it hasn't? if so, that would be a nasty surprise to callers.

i'm pretty sure you've already gone back and forth on this at least once, but if that is the case then i think we'll probably need to change the flow a bit. my recommendation would probably be to have the job API handler call the agent API, and conditionally set the cached value only if the request to the agent API succeeds.

jordandcarter:
>if the agent API returns an error for the first promised failure, doesn't that mean that all subsequent promise-failure calls will fail with a "exit status has already been claimed" message

Yes, if all 10 retries failed then future attempts would incorrectly be told the claim is already made. In such a case the job would continue until finished and then exit normally. I thought this rare case wasn't worth the complexity of moving the bk API call within the jobapi and handling simultaneous callers.

>my recommendation would probably be to have the job API handler call the agent API, and conditionally set the cached value only if the request to the agent API succeeds.

Should the caller wait for the jobapi to make the API request? Does this cause simultaneous callers to be blocked until the jobapi has made the bk API call?

jordandcarter:
@buildsworth-bk review

jordandcarter:
@buildsworth-bk review

moskyb (changes requested):
this looks really good in concept, but there's a couple suggestions i have. i've split them out into a stacked PR in https://github.com/buildkite/agent/pull/4027 -- feel free to give it a look!

moskyb (approved):
yeehaw!

PR #3997, opened 2026-06-15
Clarify redaction failures without Job API

jamiemonserrate:
## Description

Older Windows versions such as Windows Server 2016 cannot use the Job API because Unix domain sockets are unavailable. OIDC token and secret redaction both depend on the Job API, so this change keeps failing closed by default while making the failure explicit and actionable.

## Context

Linear: https://linear.app/buildkite/issue/A-1290/elastic-oidc-support-for-older-windows-versions-2016

## Changes

- Return a sentinel `jobapi.ErrJobAPIUnavailable` when the current machine cannot support the Job API.
- Check socket availability before reading the default Job API socket/token environment variables.
- Improve `oidc request-token` and `secret get` errors so users know output was not printed and can opt out with `--skip-redaction` or the command-specific environment variable.
- Add coverage for OIDC token redaction setup failures and skipped redaction, plus secret redaction setup failures.

## Testing

Ran focused tests locally:

```sh
mise exec -- go test ./clicommand ./jobapi
```

Formatted touched files with `gofmt`.

Manually smoke-tested OIDC token requests on Unix and Windows Server 2016. Verified the Unix redaction path, the Windows fail-closed error path, and the explicit `--skip-redaction` opt-out path.

Buildkite

## Disclosures / Credits

I used OpenAI Codex to help implement and refactor this change.

### DIFF
--- a/clicommand/oidc_request_token.go
+++ b/clicommand/oidc_request_token.go
@@ -3,6 +3,7 @@ package clicommand
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"slices"
 	"time"
@@ -155,7 +156,15 @@ var OIDCRequestTokenCommand = cli.Command{
 		if !cfg.SkipRedaction {
 			jobClient, err := jobapi.NewDefaultClient(ctx)
 			if err != nil {
-				return fmt.Errorf("failed to create Job API client: %w", err)
+				reason := "the Job API client could not be created"
+				if errors.Is(err, jobapi.ErrJobAPIUnavailable) {
+					reason = "the Job API is unavailable on this machine. On older Windows versions, including Windows Server 2016, the Job API is unavailable because Unix domain sockets are not supported"
+				}
+				return fmt.Errorf(
+					"automatic OIDC token redaction requires the Job API, but %s. OIDC token was not printed. To request it anyway, explicitly opt out of redaction with --skip-redaction or BUILDKITE_AGENT_OIDC_REQUEST_TOKEN_SKIP_TOKEN_REDACTION=true: %w",
+					reason,
+					err,
+				)
 			}
 
 			if err := AddToRedactor(ctx, l, jobClient, token.Token); err != nil {
--- a/clicommand/oidc_request_token.go
+++ b/clicommand/oidc_request_token.go
@@ -156,13 +156,12 @@ var OIDCRequestTokenCommand = cli.Command{
 		if !cfg.SkipRedaction {
 			jobClient, err := jobapi.NewDefaultClient(ctx)
 			if err != nil {
-				reason := "the Job API client could not be created"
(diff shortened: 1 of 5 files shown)


jamiemonserrate:
@buildsworth-bk - can you review please?

clicommand/oidc_request_token_test.go
```diff
@@ -0,0 +1,94 @@
...
+		if err == nil {
+			t.Fatal("runOIDCRequestTokenCommand() error = nil, want non-nil")
+		}
+
+		if out != "" {
+			t.Fatalf("runOIDCRequestTokenCommand() output = %q, want empty", out)
+		}
+
+		for _, want := range []string{
+			"automatic OIDC token redaction requires the Job API, but the Job API client could not be created",
```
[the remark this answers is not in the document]
jamiemonserrate: This has been addressed.

clicommand/oidc_request_token.go
```diff
@@ -155,7 +156,15 @@ var OIDCRequestTokenCommand = cli.Command{
...
-				return fmt.Errorf("failed to create Job API client: %w", err)
+				reason := "the Job API client could not be created"
+				if errors.Is(err, jobapi.ErrJobAPIUnavailable) {
+					reason = "the Job API is unavailable on this machine. On older Windows versions, including Windows Server 2016, the Job API is unavailable because Unix domain sockets are not supported"
+				}
+				return fmt.Errorf(
+					"automatic OIDC token redaction requires the Job API, but %s. OIDC token was not printed. To request it anyway, explicitly opt out of redaction with --skip-redaction or BUILDKITE_AGENT_OIDC_REQUEST_TOKEN_SKIP_TOKEN_REDACTION=true: %w",
+					reason,
+					err,
+				)
```
moskyb: i think we should be a little more explicit about why the command is failing, and where specifically token redaction is and isn't available. my pitch:

```suggestion
				err = fmt.Errorf("the Job API client could not be created (error: %w)", err)
				if errors.Is(err, jobapi.ErrJobAPIUnavailable) {
					err = fmt.Errorf("the Job API is unavailable on this machine, as it requires Unix domain sockets to function. On Windows, Unix domain sockets require Windows build 17063 or newer (Windows 10 version 1803 / Windows Server 2019 onwards)")
				}
				return fmt.Errorf("automatic OIDC token redaction requires the Job API, but %w. The command failed instead of outputting the token because it could leak in logs without redaction. To output the token anyway, explicitly opt out of redaction with --skip-redaction", err)
			}
```
(this version also has a bunch of my specific style preferences built in — feel free to ignore them).

this leaves us with error messages like:

when you're on an old windows version:
Image

when the job API is unavailable for some other reason:

Image

jamiemonserrate: Thank you for the review @moskyb! I've addressed your feedback.

moskyb (approved):
looking good!

PR #3998, opened 2026-06-15
feat: Change ID to Name in the agent to match v2 (follow up on #3993)

ss1909:
## Description

follow up on this comment on https://github.com/buildkite/agent/pull/3993#discussion_r3410483454
## Context

https://linear.app/buildkite/issue/A-1293/agent-cli-parse-v2-cacheyml-and-speak-v2-wire-protocol 
## Changes

renamed the cache entry's id: field to name: and the --ids flag/BUILDKITE_CACHE_IDS env to --names/BUILDKITE_CACHE_NAMES (with matching internal Name/Names/resolveCacheNames renames).

## Testing

### DIFF
--- a/internal/cache/configuration/cache.go
+++ b/internal/cache/configuration/cache.go
@@ -7,8 +7,8 @@ import (
 )
 
 type Cache struct {
-	// ID of the cache entry to save.
-	ID string `yaml:"id"`
+	// Name of the cache entry to save.
+	Name string `yaml:"name"`
 	// Key of the cache entry to save, this can be a template string.
 	CacheKey []KeyPart `yaml:"cache_key"`
 	// Target Paths to remove.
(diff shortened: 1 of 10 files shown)


internal/cache/configuration/cache.go
```diff
@@ -19,11 +19,11 @@ type Cache struct {
 func (c Cache) Validate() error {
 	var errors []string
 
-	// ID validation: alphanumeric and underscore only
-	if strings.TrimSpace(c.ID) == "" {
-		errors = append(errors, "id cannot be empty")
-	} else if !regexp.MustCompile(`^[a-zA-Z0-9_]+$`).MatchString(c.ID) {
-		errors = append(errors, fmt.Sprintf("id '%s' can only contain letters, numbers, and underscores", c.ID))
+	// Name validation: alphanumeric and underscore only
```
ss1909, line 22: I've assumed this rule here. Does this still stand @zhming0 ?

buildkate (approved):

internal/cache/configuration/cache.go
zhming0: it's reasonable to me. this is a two way door, we can change this anytime so 👍🏿

zhming0 (approved):

PR #4007, opened 2026-06-15
feat: Support fallbackLimit for agent keyPart

ss1909:
## Description

Building on the changes introduces in [A-1293](https://linear.app/buildkite/issue/A-1293/agent-cli-parse-v2-cacheyml-and-speak-v2-wire-protocol)

The v2 cache wire protocol marks each `cache_key` part as `mandatory` or optional: mandatory parts must match for a cache hit, while optional parts can be dropped in a fallback lookup so a near-match can still be restored. Until now the agent hard-coded every resolved part to `mandatory: true`, so users had no way to express partial fallback (e.g. "reuse yesterday's `node_modules` even if the lockfile changed").

This adds a `fallbackLimit: true` modifier to `cache.yml` that lets users draw the fallback boundary themselves. The marked part is the *last mandatory* part; every part after it is optional. With no `fallbackLimit` anywhere, all parts remian mandatory. 

## Context

https://linear.app/buildkite/issue/A-1342/agent-keypart-support-fallbacklimit

## Changes

- **Parser** (`internal/cache/configuration/key_part.go`): `fallbackLimit` is accepted as an optional modifier alongside the single source key on any object-form `KeyPart`. `fallbackLimit: false` is a documented no-op; a non-bool value is a parse error; an entry with zero or more-than-one source key is rejected.
- **Validation** (`internal/cache/configuration/cache.go`): at most one part may declare `fallbackLimit` — this cross-part rule lives in `Validate()` since `UnmarshalYAML` only sees one part at a time.
- **Config→wire normalization** (`ResolveCacheKey`): computes the boundary index from the part declaring `fallbackLimit` (default: last part) and sets `Mandatory: i <= limit` per part.
- **Docs/comments**: refreshed the stale "fallback not enabled / always mandatory" notes in `api/cache.go` and the `cache restore --help` text, and added a `fallbackLimit` example to the help.
- **Tests**: parse cases (`true`/`false`/non-bool/source-less), boundary normalization (middle and first part), and the multiple-`fallbackLimit` validation error.

## Testing

## Disclosures / Credits

Implemented in consultation with Claude.

### DIFF
--- a/clicommand/cache_restore.go
+++ b/clicommand/cache_restore.go
@@ -21,8 +21,9 @@ defined in your cache config file (defaults to .buildkite/cache.yml).
 
 The cache configuration file defines which files or directories should be restored
 and their associated cache key. An entry is restored when its target_paths
-(an order-insensitive set) and cache_key match exactly. Fallback matching is not
-enabled yet — every key part is treated as mandatory.
+(an order-insensitive set) and the mandatory parts of its cache_key match. By
+default every key part is mandatory; mark a part with fallbackLimit: true to make
+every part after it optional.
 
 Note: This feature is currently in development and subject to change. It is not
 yet available to all customers.
@@ -45,14 +46,19 @@ Configuration File Format:
 
 The cache configuration file should be in YAML format. cache_key is an ordered
 list of parts; each part is a literal string or one of { agent: os },
-{ agent: arch }, { checksum: <file> }, or { env: <VAR> }:
+{ agent: arch }, { checksum: <file> }, or { env: <VAR> }. Any one part may also
+set fallbackLimit: true to make every part after it optional for fallback
+matching (the marked part itself stays mandatory). In the example below an exact
+match is preferred, but if the lockfile changed, an entry matching node + os + arch
+is still restored, as the fallback is specified on arch, making node + os + arch mandatory, 
+but making checksum optional:
 
     caches:
       - name: node
         cache_key:
           - node
           - { agent: os }
-          - { agent: arch }
+          - { agent: arch, fallbackLimit: true }
           - { checksum: package-lock.json }
         target_paths:
           - node_modules
--- a/internal/cache/configuration/key_part.go
+++ b/internal/cache/configuration/key_part.go
@@ -15,6 +15,9 @@ import (
 type KeyPart struct {
 	Arg    string
 	Source Source
+	// FallbackLimit marks the fallback boundary: this part and every part before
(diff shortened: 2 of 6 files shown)


clicommand/cache_restore.go
```diff
@@ -45,14 +46,18 @@ Configuration File Format:
 
 The cache configuration file should be in YAML format. cache_key is an ordered
 list of parts; each part is a literal string or one of { agent: os },
-{ agent: arch }, { checksum: <file> }, or { env: <VAR> }:
+{ agent: arch }, { checksum: <file> }, or { env: <VAR> }. Any one part may also
+set fallbackLimit: true to make every part after it optional for fallback
+matching (the marked part itself stays mandatory). In the example below an exact
```
buildkate, line 51: Suggestion to expand on the example: 

In the example below node + os + arch are mandatory and lockfile is optional.  An exact
match is preferred, but if the lockfile changed, an entry matching node + os + arch
is still restored as a fallback:

ss1909: thanks @buildkate 💟 
I guess because i have context about this, the example makes sense to me. Could you help me understand what in the example is not clear, so I can expand on it?

buildkate: It helps to explicitly state which ones are mandatory / optional.  Phrases like "every part after it is optional" can be a bit confusing, because I have to look at the list of parts and think which direction am I going in to determine optional or not, and was it inclusive of the one where I marked it `fallbackLimit: true`. Does that make sense?

internal/cache/configuration/key_part.go
```diff
@@ -90,19 +111,31 @@ func (k *KeyPart) unmarshalMapping(node *yaml.Node) error {
 }
 
 // ResolveCacheKey resolves each KeyPart to its concrete value and returns the
-// flat wire shape the backend expects. Every part is mandatory in M1 (fallback
-// semantics land separately), which is why Mandatory is hard-coded true.
+// flat wire shape the backend expects. A part is mandatory iff it is at or
```
buildkate, line 114:
```suggestion
// flat wire shape the backend expects. A part is mandatory if it is at or
```

clicommand/cache_restore.go
buildkate: Just a suggestion, btw. Non blocking.

buildkate (approved):

clicommand/cache_restore.go
ss1909: makes sense. I've updated it.

PR #4009, opened 2026-06-16
A-1376 part 1: e2e test for basic cache read/write

zhming0:
## Description

Added a basic e2e test to assert the happy path of buildkite cache save/restore working

## Context

part of A-1376

## Testing

## Disclosures / Credits

LLM and brain

### DIFF
--- a/internal/e2e/fixtures/cache_save_restore.yaml
+++ b/internal/e2e/fixtures/cache_save_restore.yaml
@@ -1,0 +1,37 @@
+# cache save/restore round-trip against the customer-managed S3 bucket.
+# Region is pinned in the URL (cache S3 client defaults to us-east-1).
+# The cache config is shared via a base64 env var; its cache_key embeds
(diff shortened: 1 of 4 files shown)


internal/e2e/fixtures/cache_save_restore.yaml
```diff
@@ -0,0 +1,51 @@
+# Exercises `buildkite-agent cache save` / `cache restore` against the same
+# customer-managed S3 bucket used by the artifact e2e fixtures.
```
DrJosh9000: Should we have a separate bucket, or rename the existing bucket to be multi-purpose?

zhming0: We should! I was looking for if these were terraformed or clickops.

zhming0: https://github.com/buildkite/aws-buildkite-dist/pull/163

buildkate (approved):

PR #4010, opened 2026-06-16
Add `checkout-override-mode` option

petetomasik:
## Description

The agent exposes a number of checkout-related environment variables (git clone/fetch/checkout/clean flags, sparse-checkout paths, submodules, skip-fetch-existing-commits, skip-checkout, commit verification). Historically the protections around these were implicit and inconsistent: some were silently overwritten by agent config in `createEnvironment`, others were overridable from the backend job env, hooks, plugins, the Job API, or secrets, with the rules spread across a single `protectedEnv` map and scattered logic. Because git flags are a shell-injection vector (for example `--upload-pack`, `-c core.sshCommand`, `ext::` transports), letting a job set them can be equivalent to arbitrary command execution and a way to bypass `no-command-eval`.

This PR replaces that implicit handling with an explicit, tri-state `--checkout-override-mode` flag (env `BUILDKITE_CHECKOUT_OVERRIDE_MODE`), shared by `agent start` and `bootstrap`, controlling which sources may override the agent's checkout-scoped configuration:

- `strict` — the agent is authoritative against every source: pipeline/step env, secrets, hooks, plugins, and the Job API.
- `from-job` (default) — hooks, plugins, and the Job API may set checkout-scoped vars (so a job can still tailor its own checkout), but the backend job env (pipeline/step env) and secrets may not. This matches the agent's historical behavior.
- `none` — any source, including the backend job env and secrets, may set them.

Disabling command evaluation via `no-command-eval=true` forces the mode to `strict` regardless of the configured value, so git flags can never be used to undermine `no-command-eval`.

The protection model lives in `env/protected.go` and is split into two disjoint tiers (disjointness enforced by a test):

- `protectedEnv` — always agent-authoritative in every mode: agent infrastructure, all mirror configuration (`BUILDKITE_GIT_MIRRORS_PATH`, `BUILDKITE_GIT_MIRRORS_LOCK_TIMEOUT`, `BUILDKITE_GIT_MIRRORS_SKIP_UPDATE`, `BUILDKITE_GIT_MIRROR_CHECKOUT_MODE`, `BUILDKITE_GIT_CLONE_MIRROR_FLAGS`), and `BUILDKITE_GIT_SUBMODULE_CLONE_CONFIG`.
- `checkoutOverrideScope` — the checkout-scoped vars whose lock depends on the mode: `BUILDKITE_GIT_CHECKOUT_FLAGS`, `BUILDKITE_GIT_CHECKOUT_TIMEOUT`, `BUILDKITE_GIT_CLEAN_FLAGS`, `BUILDKITE_GIT_CLONE_FLAGS`, `BUILDKITE_GIT_COMMIT_VERIFICATION`, `BUILDKITE_GIT_FETCH_FLAGS`, `BUILDKITE_GIT_SKIP_FETCH_EXISTING_COMMITS`, `BUILDKITE_GIT_SPARSE_CHECKOUT_PATHS`, `BUILDKITE_GIT_SUBMODULES`, `BUILDKITE_SKIP_CHECKOUT`. `BUILDKITE_GIT_COMMIT_VERIFICATION` is an enum (`""`/`warn`/`strict`), not an injection vector, but the backend exposes it under `checkout:` alongside the flag vars, so it is governed by the mode too.

## Context

[Linear](https://linear.app/buildkite/issue/SUP-6423/agent-implement-no-checkout-override-in-agent-config)

## Changes

New CLI flag, shared by `agent start` and `bootstrap`:

```
--checkout-override-mode value   Controls which sources may override the agent's checkout settings;
                                 one of [strict from-job none]. 'strict' makes the agent authoritative
                                 against pipeline/step env, secrets, hooks, plugins, and the Job API.
                                 'from-job' (default) lets hooks, plugins, and the Job API set checkout
                                 vars, but blocks pipeline/step env and secrets. 'none' additionally
                                 lets pipeline/step env and secrets set them. All mirror configuration
                                 and submodule clone config stay agent-authoritative in every mode. The
                                 checkout SSH key and Git LFS toggle are not governed by this flag and
                                 stay job-settable in every mode. Disabling command-eval forces this to
                                 'strict'.
                                 (default: "from-job") [$BUILDKITE_CHECKOUT_OVERRIDE_MODE]
```

- Adds a `CheckoutOverrideMode` type in `env` (`strict`/`from-job`/`none`) with `ParseCheckoutOverrideMode`, `String`, and `RestrictedForCommandEval` (forces `strict` when command-eval is off).
- Wires the mode through `AgentConfiguration`, `AgentStartConfig`, `BootstrapConfig`, `ExecutorConfig`, and the Job API server (`WithCheckoutOverrideMode`), via a shared `resolveCheckoutOverrideMode` helper that applies the command-eval floor. `ExecutorConfig.CheckoutOverrideMode` has no `env` tag, so a hook cannot relax the mode mid-job.
- Splits `env/protected.go` into `protectedEnv` (always protected) and `checkoutOverrideScope` (mode-dependent), and adds predicates `IsCheckoutOverrideScoped`, `IsCheckoutLocked` (within-job sources: hooks, plugins, and the Job API; `strict` only), and `IsCheckoutLockedForSecrets` (secrets; every mode except `none`). The backend job env follows the same rule as secrets (locked in every mode except `none`), enforced in `createEnvironment`. The two maps are asserted disjoint.
- Scopes `BUILDKITE_GIT_COMMIT_VERIFICATION` to the mode: it was fully protected, so a pipeline's `checkout.commit_verification` was silently dropped with an "Ignored" warning even under `--checkout-override-mode=none`. Because the backend exposes it under `checkout:` alongside the flag vars, it moves from `protectedEnv` to `checkoutOverrideScope` and is emitted via `setCheckoutEnv` in `createEnvironment`. Under `none` the job env now sets it (so verification can actually run); `from-job` and `strict` keep the agent authoritative.
- Enforces the mode at every point a job can reach checkout config: backend job-env precedence in `createEnvironment` (`setCheckoutEnv`), runtime config refresh in `ReadFromEnvironment`, hook env changes in `applyEnvironmentChanges`, secret-to-env mappings in `fetchAndSetSecrets`, and Job API patch/delete (`checkProtected`).
- Makes all mirror configuration and `BUILDKITE_GIT_SUBMODULE_CLONE_CONFIG` always agent-authoritative, including moving `BUILDKITE_GIT_MIRRORS_SKIP_UPDATE` and `BUILDKITE_GIT_CLONE_MIRROR_FLAGS` out of the mode-dependent scope (the mirror is shared infrastructure and clone-mirror flags are a cross-job injection vector).
- The Job API rejection message names the checkout-override lock when that is the reason for rejection.
- Adds unit and integration coverage across `env`, `internal/job` config, the `agent` job runner, and the hooks/plugins/secrets/Job API paths, including: the job-env override matrix across all three modes, infra vars staying agent-authoritative even under `none`, secrets rejected for the protected checkout-infra vars, commit verification following the mode (enabled from job env under `none`, locked under `from-job`/`strict`), and an end-to-end test that disabling command-eval auto-forces `strict`.

### Agent behavior changes
Under `from-job` (the default), checkout-scoped vars remain overridable from **within-job** sources (hooks, plugins, and the Job API) but not from the backend job env or secrets, matching the agent's historical behavior. `strict` offers a more protected behavior: hooks and plugins can no longer set the checkout-scoped or mirror vars (on main several were mutable from within the job). `none` is the only mode that opens checkout-scoped vars to the backend job env and secrets. As part of this, `BUILDKITE_GIT_COMMIT_VERIFICATION` is now mode-governed rather than always dropped: a pipeline's `checkout.commit_verification` takes effect under `none`, while `from-job` and `strict` keep the agent authoritative.

## Testing
- [x] Tests have run locally (with `go test ./...`).

`go test ./...` passes locally, including the full checkout-override-mode unit and integration coverage across `env`, `internal/job`, `agent`, `jobapi`, and `clicommand`, plus the commit-verification cases added to `agent/integration/job_environment_integration_test.go`.

## Disclosures / Credits

Code generated by Claude Code (Opus 4.8) under my guidance. Multiple reviews were conducted to establish the correctness and completeness of the implementation of this new agent config option and protection model. The core feature and its conversion from the earlier binary design to the tri-state mode were implemented in earlier revisions.


DrJosh9000:
Just to let you know I plan to continue reviewing next week

agent/job_runner.go
```diff
@@ -440,6 +441,14 @@ func (r *JobRunner) createEnvironment(ctx context.Context) ([]string, error) {
 		}
 		env[name] = value
 	}
+	setCheckoutEnv := func(name, value string) {
```
DrJosh9000, line 456:
```suggestion
	// For some checkout env vars, we allow the Job env (if present) to
	// have higher precedence than the agent configuration, unless
	// NoCheckoutOverride is set.
	setCheckoutEnv := func(name, value string) {
```

agent/job_runner.go
```diff
@@ -620,18 +642,18 @@ BUILDKITE_AGENT_JWKS_KEY_ID`
...
+	setCheckoutEnv("BUILDKITE_GIT_SPARSE_CHECKOUT_PATHS", strings.Join(r.conf.AgentConfiguration.GitSparseCheckoutPaths, ","))
+	setCheckoutEnv("BUILDKITE_GIT_CLONE_MIRROR_FLAGS", r.conf.AgentConfiguration.GitCloneMirrorFlags)
 	setEnv("BUILDKITE_GIT_MIRROR_CHECKOUT_MODE", r.conf.AgentConfiguration.GitMirrorCheckoutMode)
-	setEnv("BUILDKITE_GIT_CLEAN_FLAGS", r.conf.AgentConfiguration.GitCleanFlags)
+	setCheckoutEnv("BUILDKITE_GIT_CLEAN_FLAGS", r.conf.AgentConfiguration.GitCleanFlags)
 	setEnv("BUILDKITE_GIT_MIRRORS_LOCK_TIMEOUT", strconv.Itoa(r.conf.AgentConfiguration.GitMirrorsLockTimeout))
 	if r.conf.AgentConfiguration.GitCheckoutTimeout > 0 {
 		setEnv("BUILDKITE_GIT_CHECKOUT_TIMEOUT", strconv.Itoa(r.conf.AgentConfiguration.GitCheckoutTimeout))
 	}
-	setEnv("BUILDKITE_GIT_SUBMODULE_CLONE_CONFIG", strings.Join(r.conf.AgentConfiguration.GitSubmoduleCloneConfig, ","))
```
DrJosh9000, line 637: The other changes mostly make sense to me, but what is the use-case for setting `BUILDKITE_GIT_SUBMODULE_CLONE_CONFIG` (specifically) from the job? I recall that it was added so that some of the integration tests could be made to work when there was a Git security vulnerability https://github.com/buildkite/agent/pull/1895

DrJosh9000 (commented):
I have questions

agent/job_runner.go
petetomasik: Agreed. This change was inadvertently grouped in with the others. Let me remedy this back to the agent being authoritative for this config.

agent/job_runner.go
```diff
@@ -620,16 +645,13 @@ BUILDKITE_AGENT_JWKS_KEY_ID`
...
-	setEnv("BUILDKITE_GIT_FETCH_FLAGS", r.conf.AgentConfiguration.GitFetchFlags)
-
-	if len(r.conf.AgentConfiguration.GitSparseCheckoutPaths) > 0 {
-		setEnv("BUILDKITE_GIT_SPARSE_CHECKOUT_PATHS", strings.Join(r.conf.AgentConfiguration.GitSparseCheckoutPaths, ","))
-	}
-	setEnv("BUILDKITE_GIT_CLONE_MIRROR_FLAGS", r.conf.AgentConfiguration.GitCloneMirrorFlags)
+	setCheckoutEnv("BUILDKITE_GIT_CHECKOUT_FLAGS", r.conf.AgentConfiguration.GitCheckoutFlags)
+	setCheckoutEnv("BUILDKITE_GIT_CLONE_FLAGS", r.conf.AgentConfiguration.GitCloneFlags)
+	setCheckoutEnv("BUILDKITE_GIT_FETCH_FLAGS", r.conf.AgentConfiguration.GitFetchFlags)
+	setCheckoutEnv("BUILDKITE_GIT_SPARSE_CHECKOUT_PATHS", strings.Join(r.conf.AgentConfiguration.GitSparseCheckoutPaths, ","))
```
isaacsu, line 671: Previously, we only set this envvar when `len(r.conf.AgentConfiguration.GitSparseCheckoutPaths) > 0` (line 627-629).

Does it matter that, with this change, we're now setting it regardless?

DrJosh9000 (approved):
It's taken a while to review, mainly because of size, but I think this is fine.

clicommand/global.go
```diff
@@ -205,6 +205,12 @@ var (
 		EnvVar: "BUILDKITE_SKIP_CHECKOUT",
 	}
 
+	NoCheckoutOverrideFlag = cli.BoolFlag{
+		Name:   "no-checkout-override",
+		Usage:  "Don't allow pipeline/step env, hooks, plugins, the Job API, or secrets to override the agent's checkout settings (default: false)",
+		EnvVar: "BUILDKITE_NO_CHECKOUT_OVERRIDE",
+	}
```
DrJosh9000: On second thought, I think we can't proceed with this unless no-checkout-override is enabled by default. Allowing pipeline/step env to override checkout settings is too much behaviour change. We can consider making no-checkout-override disabled by default in v4 or a later major version.

DrJosh9000 (changes requested):

clicommand/global.go
petetomasik: @DrJosh9000 We can certainly default `no-checkout-override` to enabled for now and revisiting this default at a later major release. Before we change the default behavior I wanted to line the scope of this change, because I think the security-critical piece is already covered and the default becomes a behavior-change decision. Enabling `no-checkout-override=true` as the default would alter two different things relative to the current behavior in the agent:

- Pipeline/step env and secrets: already blocked for the git flags (pipeline/step env because `createEnvironment` overwrites them with agent config, secrets via `IsProtected`), so defaulting the lock on just keeps that as-is.
- Hooks, plugins, and the Job API: agent allows updates (`mutableFromWithinJob: true`), specifically so checkout-related plugins can set flags. For example, the `git-clean` plugin sets `BUILDKITE_GIT_CLEAN_FLAGS` from a `pre-checkout` hook. With `no-checkout-override=true` that export gets stripped in `applyEnvironmentChanges` (the job just sees an "env vars were blocked" line). So defaulting `no-checkout-override=true` would have the effect of silently no-op'ing checkout-related plugins unless a user explicitly sets `no-checkout-override=false` in the agent.

Do we want to re-think how `no-checkout-override` is implemented to minimize behavior changes or potentially have this new security control wait for the next major release? What do you think is best here?

DrJosh9000: Yeah, I think we're agreed on the constraints for defaults for new config options in the same major version:

* avoid breaking existing pipelines (e.g. the various git plugins continue to work by setting checkout-related env vars)
* avoid introducing new security holes (e.g. git flags passed from the backend with shell injections that could bypass no-command-eval)

Which means we're a bit stuck if we want to ship either case in v3.

> Do we want to re-think how no-checkout-override is implemented to minimize behavior changes or potentially have this new security control wait for the next major release? What do you think is best here?

There's another path: make the new option string-valued, and let it accept 3 choices (the names are straight off the top of my head, feel free to come up with better ones):

* `strict` - equivalent to the current `true`
* `none` - equivalent to the current `false`
* `from-job` (new default) which is the existing behaviour.

petetomasik: I've reworked the protection mechanism to change from a boolean to a tri-state value which allows preserving the agent's current behavior by default (`from-job`), while introducing new stricter (`strict`) and looser (`none`) options.

agent/job_runner.go
petetomasik: No, it's benign and now deliberate. An empty `BUILDKITE_GIT_SPARSE_CHECKOUT_PATHS` is treated the same as unset: `cleanGitSparseCheckoutPaths` drops empty entries and `resolveSparseCheckout` returns `nil` (full checkout) when none survive (`internal/job/checkout_sparse.go:18-22,40-49`).

It also has to be unconditional now that it flows through `setCheckoutEnv`: under `strict`/`from-job` the `len > 0` guard was a leak, since with no agent-configured paths a backend-supplied value would survive past the lock. Under `none`, `setCheckoutEnv` still defers to the job env. Covered by the empty-agent-config cases in `TestCheckoutScopedJobEnvOverrideHonorsCheckoutOverrideMode`.

internal/job/executor.go
```diff
@@ -648,7 +648,8 @@ func (e *Executor) applyEnvironmentChanges(changes hook.EnvChanges) {
 	// breaks the rest of the job.
 	var protected []string
 	for k := range changes.Diff.Keys {
-		if env.IsProtectedFromWithinJob(k) {
+		if env.IsProtectedFromWithinJob(k) ||
+			env.IsCheckoutLocked(k, e.CheckoutOverrideMode) {
```
DrJosh9000: Don't break the line: it's not uncomfortably long as 1 line, there's no "max line length" in Go, and breaking like this can make the second condition appear to be a statement within the if block due to indentation.

```suggestion
		if env.IsProtectedFromWithinJob(k) || env.IsCheckoutLocked(k, e.CheckoutOverrideMode) {
```

jobapi/env.go
```diff
@@ -138,14 +138,29 @@ func (s *Server) deleteEnv(w http.ResponseWriter, r *http.Request) {
...
 
-func checkProtected(candidates []string) []string {
+func (s *Server) checkProtected(candidates []string) []string {
 	protected := make([]string, 0, len(candidates))
 	for _, c := range candidates {
 		// The Job API is only accessible from within the job, so allow writes
 		// to vars that allow write from within job.
-		if env.IsProtectedFromWithinJob(c) {
+		if env.IsProtectedFromWithinJob(c) ||
+			env.IsCheckoutLocked(c, s.checkoutOverrideMode) {
```
DrJosh9000:
```suggestion
		if env.IsProtectedFromWithinJob(c) || env.IsCheckoutLocked(c, s.checkoutOverrideMode) {
```

jobapi/env.go
```diff
@@ -138,14 +138,29 @@ func (s *Server) deleteEnv(w http.ResponseWriter, r *http.Request) {
...
+// always-protected var.
+func (s *Server) protectedEnvMessage(protected []string) string {
+	msg := fmt.Sprintf("the following environment variables are protected, and cannot be modified: % v", protected)
+	for _, p := range protected {
+		if env.IsCheckoutLocked(p, s.checkoutOverrideMode) {
+			msg += fmt.Sprintf(". Checkout-related variables are locked because BUILDKITE_CHECKOUT_OVERRIDE_MODE=%s", s.checkoutOverrideMode)
+			break
+		}
+	}
+	return msg
```
DrJosh9000: The `stringsbuilder` lint will complain that this is building a string in a loop with `+=`, even though it breaks on the first locked var it finds.

There's no problem with switching to `strings.Builder` and `fmt.Fprintf`, though:

```suggestion
	var msg strings.Builder
	fmt.Fprintf(&msg, "the following environment variables are protected, and cannot be modified: % v", protected)
	for _, p := range protected {
		if env.IsCheckoutLocked(p, s.checkoutOverrideMode) {
			fmt.Fprintf(&msg, ". Checkout-related variables are locked because BUILDKITE_CHECKOUT_OVERRIDE_MODE=%s", s.checkoutOverrideMode)
			break
		}
	}
	return msg.String()
```

env/protected.go
```diff
@@ -102,3 +151,141 @@ func IsProtectedFromWithinJob(name string) bool {
...
+// IsCheckoutLocked (the job's own config sources) and IsCheckoutLockedForSecrets.
+func IsCheckoutOverrideScoped(name string) bool {
+	_, exists := checkoutOverrideScope[normalizeKeyName(name)]
+	return exists
+}
+
+// CheckoutOverrideMode controls how much of the agent's checkout configuration a
+// job may override. It applies only to checkoutOverrideScope vars; protectedEnv
+// membership is independent of the mode.
+type CheckoutOverrideMode int
```
DrJosh9000: I'm not sure why we need both an int-style enum and strings, given that the string form is the one used the most. If this were a string type:

```suggestion
type CheckoutOverrideMode string
```

then `CheckoutOverrideFromJob`, `CheckoutOverrideStrict`, `CheckoutOverrideNone` could take their string values, and `ParseCheckoutOverrideMode` would simplify down to:

```go
if !slices.Contains(CheckoutOverrideModeNames, s) { 
	return CheckoutOverrideFromJob, ... 
}
return CheckoutOverrideMode(s), nil
```

DrJosh9000 (approved):

PR #4011, opened 2026-06-17
A-1376 part 3: e2e test for filesystem integrity

zhming0:
## Description

Add end to end test coverage for fs integrity after cache restore, it was an issue in v1

## Context

part of A-1376

## Disclosures / Credits

Clanker and minion.


zhming0:
@buildsworth-bk review again?

PR #4015, opened 2026-06-19
Skip tracking unchanged slices in ReadFromEnvironment (suppress BUILDKITE_GIT_SUBMODULE_CLONE_CONFIG log noise)

lizrabuya:
## Description

When git-submodule-clone-config isn't set at the agent level, the job runner was still calling setEnv("BUILDKITE_GIT_SUBMODULE_CLONE_CONFIG", "") unconditionally. The env var would then be present (as an empty  string) in the job's process environment, which means it shows up in every hook's env dump  adding noise for a value that was never configured.

This guards the setEnv call on len(...) > 0, matching the pattern used a few lines above for BUILDKITE_GIT_CHECKOUT_TIMEOUT. When the slice has any entries — including a single empty string from a malformed config like --git-submodule-clone-config= — we still export the value, so misconfiguration remains visible in logs rather than being silently swallowed.

## Context

Customer reported that they are seeing `BUILDKITE_GIT_SUBMODULE_CLONE_CONFIG is now ""` was repeatedly logged in a job's log when they never set this env var and expects it to be empty.

## Changes

- agent/job_runner.go: only call setEnv("BUILDKITE_GIT_SUBMODULE_CLONE_CONFIG", ...) when AgentConfiguration.GitSubmoduleCloneConfig is non-empty.

## Testing

## Disclosures / Credits

Claude Code helped trace the code paths and draft this description; I reviewed and authored the change.

### DIFF
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -631,7 +631,9 @@ BUILDKITE_AGENT_JWKS_KEY_ID`
 	if r.conf.AgentConfiguration.GitCheckoutTimeout > 0 {
 		setEnv("BUILDKITE_GIT_CHECKOUT_TIMEOUT", strconv.Itoa(r.conf.AgentConfiguration.GitCheckoutTimeout))
 	}
-	setEnv("BUILDKITE_GIT_SUBMODULE_CLONE_CONFIG", strings.Join(r.conf.AgentConfiguration.GitSubmoduleCloneConfig, ","))
+	if len(r.conf.AgentConfiguration.GitSubmoduleCloneConfig) > 0 {
+		setEnv("BUILDKITE_GIT_SUBMODULE_CLONE_CONFIG", strings.Join(r.conf.AgentConfiguration.GitSubmoduleCloneConfig, ","))
+	}
 	setEnv("BUILDKITE_GIT_COMMIT_VERIFICATION", r.conf.AgentConfiguration.GitCommitVerification)
 
 	setEnv("BUILDKITE_SHELL", r.conf.AgentConfiguration.Shell)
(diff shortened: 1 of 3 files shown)


agent/job_runner.go
```diff
@@ -631,7 +631,9 @@ BUILDKITE_AGENT_JWKS_KEY_ID`
 	if r.conf.AgentConfiguration.GitCheckoutTimeout > 0 {
 		setEnv("BUILDKITE_GIT_CHECKOUT_TIMEOUT", strconv.Itoa(r.conf.AgentConfiguration.GitCheckoutTimeout))
 	}
-	setEnv("BUILDKITE_GIT_SUBMODULE_CLONE_CONFIG", strings.Join(r.conf.AgentConfiguration.GitSubmoduleCloneConfig, ","))
+	if len(r.conf.AgentConfiguration.GitSubmoduleCloneConfig) > 0 {
+		setEnv("BUILDKITE_GIT_SUBMODULE_CLONE_CONFIG", strings.Join(r.conf.AgentConfiguration.GitSubmoduleCloneConfig, ","))
+	}
```
DrJosh9000: `BUILDKITE_GIT_SUBMODULE_CLONE_CONFIG` should not be overridable from the job env, because it is too dangerous (a number of command injections are possible). Can we solve this on the bootstrap side instead?

DrJosh9000 (changes requested):

agent/job_runner.go
lizrabuya: I see. I found the issue on the `config.go` comparison for env var slices like BUILDKITE_GIT_SUBMODULE_CLONE_CONFIG to always get marked as changed and fixed that instead.

DrJosh9000 (approved):
LGTM, thanks

PR #4016, opened 2026-06-20
Support post-registration hook and inject agent env into lifecycle hooks

catkins:
### Motivation

- Add the ability to run a `post-registration` hook after an agent registers, with the agent-specific environment available to the hook. 
- Ensure lifecycle hook execution can receive custom environment variables so hooks can use registration metadata like agent ID, name, and tags.

### Description

- Added import of the `env` package and a new `agentPostRegistrationHook` which builds an environment and delegates to the generic lifecycle handler. 
- Call `agentPostRegistrationHook` immediately after successful agent registration and use a per-agent logger (`agentLog`) for context. 
- Changed `agentLifecycleHook` signature to accept `hookEnv *env.Environment` and merge that environment into the shell before running hooks. 
- Added helper `agentPostRegistrationHookEnv` to populate `BUILDKITE_AGENT_ID`, `BUILDKITE_AGENT_NAME`, and `BUILDKITE_AGENT_META_DATA` from the registration response. 
- Updated `agentStartupHook` and `agentShutdownHook` calls to pass `nil` for the new env parameter and added test helpers (`writeAgentHookScript`) and a new unit test `TestAgentPostRegistrationHook` to validate the behavior.

### Testing

- Ran unit tests with `go test ./...` and the test suite passed. 
- Exercised `TestAgentPostRegistrationHook`, `TestAgentStartupHook`, and `TestAgentShutdownHook` which succeeded and validated hook output and environment injection.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_b_6a370dca7918832f9c58bfabef357635)

### DIFF
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -26,6 +26,7 @@ import (
 	"github.com/buildkite/agent/v3/agent"
 	"github.com/buildkite/agent/v3/api"
 	"github.com/buildkite/agent/v3/core"
+	"github.com/buildkite/agent/v3/env"
 	"github.com/buildkite/agent/v3/internal/agentapi"
 	"github.com/buildkite/agent/v3/internal/awslib"
 	"github.com/buildkite/agent/v3/internal/concurrently"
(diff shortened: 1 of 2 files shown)


catkins:
@codex review

catkins:
I might switch this to just add the ids into the existing agent startup hook for now

PR #4017, opened 2026-06-21
chore: put cache logs in collapsed group

buildkate:
## Description

Put cache execution logs in their own collapsed group for better visibility.

## Context

Solves A-1400

## Testing

### DIFF
--- a/clicommand/cache_restore.go
+++ b/clicommand/cache_restore.go
@@ -2,6 +2,7 @@ package clicommand
 
 import (
 	"context"
+	"fmt"
 	"slices"
 
 	"github.com/buildkite/agent/v3/api"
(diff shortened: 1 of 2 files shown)


clicommand/cache_restore.go
```diff
@@ -93,7 +94,9 @@ var CacheRestoreCommand = cli.Command{
 		ctx, span := otel.Tracer("buildkite-agent").Start(ctx, "cache-restore")
 		defer span.End()
 
-		l.Infof("Cache restore command executed")
+		// Emit a Buildkite group header as raw job-log output so the cache
+		// output is collapsed into its own group.
+		fmt.Println("--- :package: Cache restore command executed")
```
moskyb: given that header logs are kinda "titles" for a group of logs, it might be good to make the log something like `Restoring cache...` or something more forward-looking

mitchbne: `📦 Restoring cache...` has my vote 👍

buildkate: The tribe as spoken. ✅

zhming0 (approved):

PR #4024, opened 2026-06-24
fix(cache): set flag as name singular

buildkate:
## Description

(cache restore/save commands) changes the flag `--names` to `--name`

## Context

Solves A-1413

## Testing

## Disclosures / Credits

Claude

### DIFF
--- a/clicommand/cache_restore.go
+++ b/clicommand/cache_restore.go
@@ -35,9 +35,9 @@ Example:
     $ buildkite-agent cache restore
 
 This will restore all caches defined in the cache configuration file.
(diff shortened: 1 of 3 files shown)


DrJosh9000 (approved):
I think this is new enough that we don't need to worry about it being a breaking change?

buildkate:
> I think this is new enough that we don't need to worry about it being a breaking change?

@DrJosh9000 we're only using it internally in the agent-stack-k8s ci, which I will update after [rel 3.129.0](https://github.com/buildkite/agent/pull/4026)

PR #4027, opened 2026-06-24
Stacked review changes for #3996

moskyb:
Stacked on #3996 

This PR suggests a couple of changes to #3996:
- Split job api promise failure logic into HTTP stuff (which the handler handles) and coordination (a small, dedicated coordinator object)
- Job API HTTP errors are now reserved for errors in the job API itself — a 5xx response from the Buildkite Agent API no longer returns an error (though still has to be handled by the caller
- Makes it so that the HTTP client constructed for the job API to call the agent API retains settings like `NO_HTTP2`, has the right user agent, etc

### DIFF
--- a/jobapi/payloads.go
+++ b/jobapi/payloads.go
@@ -89,4 +89,18 @@ const (
 type PromiseFailureResponse struct {
 	// Outcome is PromiseFailureDeclared or PromiseFailureDebounced.
 	Outcome string `json:"outcome"`
+
+	// Accepted reports whether the Buildkite API accepted the promised failure.
+	Accepted bool `json:"accepted"`
+
+	// UpstreamStatus is the Buildkite API status, when one was received.
+	// Network errors that never received a response leave this as 0.
+	UpstreamStatus int `json:"upstream_status,omitempty"`
+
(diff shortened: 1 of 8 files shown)


jobapi/payloads.go
```diff
@@ -89,4 +89,18 @@ const (
...
+	// UpstreamStatus is the Buildkite API status, when one was received.
+	// Network errors that never received a response leave this as 0.
+	UpstreamStatus int `json:"upstream_status,omitempty"`
+
+	// Error is the Buildkite API declaration error, if Accepted is false.
+	Error string `json:"error,omitempty"`
+
+	// Terminal reports whether repeating this declaration is expected to return
+	// the same error, so the result can be cached.
+	Terminal bool `json:"terminal,omitempty"`
```
jordandcarter: What's the purpose of `terminal`? The client never uses it?

moskyb: yeah, good call — i think the original purpose was that the caller could check it to see what kind of action they need to take, but you're right, it isn't needed. i'll pull it out

jordandcarter (approved):

PR #4028, opened 2026-06-25
A-1414: support hosted cache

zhming0:
## Description

Support hosted cache via Namespace artifact storage, driven by an `nsc://<namespace>` cache store URL under the `agent_managed` store type.

What this does:

- **Scheme dispatch:** `NewBlobStore` routes `agent_managed` URLs by scheme — `nsc://` → `NscStore`, otherwise S3.
- **Namespace from URL:** the namespace is parsed from `nsc://<namespace>` and passed as `--namespace` to the `nsc` CLI (errors if the namespace is missing).
- **Validation:** `validateCacheStore` now accepts `s3://`, `nsc://`, and `file://` for `agent_managed`, and rejects unknown schemes.
- **Default expiry:** uploads pass `--expires_in 24h` (temporary stopgap; TTL refresh-on-access parity with S3 tracked in A-1437).
- **Context cancellation:** the nsc CLI now runs via `exec.CommandContext`, so uploads/downloads abort on context cancel/timeout.
- **Cleanup:** removed the unused `local_hosted_agents` store type (NSC is now reachable only via `nsc://`).

Tests are lean and don't require the `nsc` binary: URL parsing, namespace passthrough via an injected command-runner seam, validation short-circuiting, and `nsc://` dispatch. The existing `NSC_INTEGRATION_TEST`-gated end-to-end test is retained.

Follow-ups: A-1437 (TTL refresh, pending upstream), A-1438 (cache busting / invalidation).

## Context

A-1414

## Testing

## Disclosures / Credits

LLM and me.

### DIFF
--- a/internal/cache/store/blob.go
+++ b/internal/cache/store/blob.go
@@ -3,6 +3,7 @@ package store
 import (
 	"context"
 	"fmt"
+	"strings"
 )
 
 // Blob interface defines the operations for blob storage
@@ -17,9 +18,16 @@ type Blob interface {
 func NewBlobStore(ctx context.Context, store, bucketURL string) (Blob, error) {
 	switch store {
 	case AgentManaged:
-		return NewS3Blob(ctx, bucketURL)
-	case LocalHostedAgents:
-		return NewNscStore()
+		scheme, _, _ := strings.Cut(bucketURL, "://")
+		switch scheme {
+		case nscScheme:
+			return NewNscStore(bucketURL)
+		case "file":
+			// Supported only for local testing, kept consistent with validateCacheStore.
+			return NewLocalFileBlob(ctx, bucketURL)
+		default:
+			return NewS3Blob(ctx, bucketURL)
+		}
 	case LocalFileStore:
 		return NewLocalFileBlob(ctx, bucketURL)
 	default:
--- a/internal/cache/store/nsc.go
+++ b/internal/cache/store/nsc.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"net/url"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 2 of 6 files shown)


internal/cache/store/nsc.go
```diff
@@ -16,13 +17,61 @@ import (
...
+type NscStore struct {
+	// namespace is taken from the nsc://<namespace> cache store URL and passed
+	// as --namespace to the nsc CLI.
+	namespace string
+	run       commandRunner
+}
 
-func NewNscStore() (*NscStore, error) {
-	return &NscStore{}, nil
+// NewNscStore creates a store backed by the nsc CLI. namespace comes from the
```
buildkate: this comment says namespace may be empty but parseNscURL will return an error if it's empty. they kind of contradict each other

internal/cache/store/nsc_test.go
```diff
@@ -211,107 +214,116 @@ func TestRunCommandValidation(t *testing.T) {
...
+	tests := []struct {
+		name      string
+		url       string
+		namespace string
+		ok        bool
+		wantErr   bool
+	}{
+		{name: "nsc with namespace", url: "nsc://my-namespace", namespace: "my-namespace", ok: true},
+		{name: "not nsc", url: "s3://my-bucket", ok: false},
+		{name: "nsc without namespace", url: "nsc://", ok: true, wantErr: true},
```
buildkate: Can we add a case for invalid URL

zhming0: Added 👍🏿

internal/cache/store/nsc.go
zhming0: Yeah I agree it's wise to universally reject empty namespace. 👍🏿

internal/cache/store/blob.go
```diff
@@ -17,9 +18,19 @@ type Blob interface {
 func NewBlobStore(ctx context.Context, store, bucketURL string) (Blob, error) {
 	switch store {
 	case AgentManaged:
-		return NewS3Blob(ctx, bucketURL)
-	case LocalHostedAgents:
-		return NewNscStore()
+		namespace, isNsc, err := parseNscURL(bucketURL)
```
buildkate: It feels like parseNscURL could be a generalised URL parser, that returns `scheme` ,`host`, `err` and then we switch on `scheme` here. 

added benefit of reusing the parser in save.go  validateCacheStore

zhming0: Ah good catch! yep I've refactored this dispatch logic to use a generic scheme parser.

buildkate (approved):

PR #4029, opened 2026-06-25
Avoid overriding BUILDKITE_GIT_SPARSE_CHECKOUT_PATHS when agent has no sparse config

lizrabuya:
## Description

When the backend ships `BUILDKITE_GIT_SPARSE_CHECKOUT_PATHS` in the job env, the agent  was silently replacing it with the agent-side config value — which, on a default   agent, is an empty string. The bootstrap then read `""`, found no paths, and fall back to a full checkout.
 
 After this change:
  - Agent operator sets `--git-sparse-checkout-paths` → agent config wins (unchanged).
  - Backend sends the env var, agent has no sparse config → server value flows
    through to the bootstrap as intended.

## Testing

## Disclosures / Credits

Used claude to review the code change.


omehegan (approved):
Good catch on this!

PR #4031, opened 2026-06-29
feat: Add `--reject-parse-warnings` flag to fail pipeline upload on p…

ss1909:
…arse warnings

## Description

This adds an opt-in `--reject-parse-warnings` flag (env: `BUILDKITE_AGENT_PIPELINE_UPLOAD_REJECT_PARSE_WARNINGS`) that makes the upload fail (exit 1) when parsing or interpolation produces any warning, instead of logging and continuing. Defaults to `false`, so existing behaviour is unchanged.

## Context

https://linear.app/buildkite/issue/A-1238/add-config-option-to-fail-pipeline-upload-on-parsing-warnings-as-well

## Changes

- New `--reject-parse-warnings` flag on `buildkite-agent pipeline upload`.
- Fix `parseAndInterpolate` so parse warnings are preserved through interpolation (and combined with interpolation warnings).
- Extract the warning/error handling into `handleParseError` for clarity and testability.
- Tests: `TestParseWarningsSurviveInterpolation` (warning survives with and without interpolation) and `TestHandleParseError` (hard errors always abort; warnings abort only when the flag is set, otherwise log-and-continue).

## Testing

## Disclosures / Credits

Claude helped write the test and with naming the flag 🎏

### DIFF
--- a/clicommand/pipeline_upload.go
+++ b/clicommand/pipeline_upload.go
@@ -622,7 +622,11 @@ func (cfg *PipelineUploadConfig) handleParseError(l logger.Logger, src string, e
 	}
 
 	if cfg.RejectParseWarnings {
-		return NewExitError(1, fmt.Errorf("pipeline %q had parsing warnings, and --reject-parse-warnings is enabled:\n%v", src, w))
+return NewExitError(1, fmt.Errorf(
+	"pipeline %q parsed with warnings; refusing upload because --reject-parse-warnings is enabled:\n%v",
+	src,
+	w,
+))
 	}
 
 	l.Warnf("There were some issues with the pipeline input - pipeline upload will proceed, but might not succeed:\n%v", w)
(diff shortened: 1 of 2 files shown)


clicommand/pipeline_upload.go
```diff
@@ -607,6 +609,26 @@ func searchForSecrets(
...
+		return nil
+	}
+
+	w := warning.As(err)
+	if w == nil {
+		return err
+	}
+
+	if cfg.RejectParseWarnings {
+		return NewExitError(1, fmt.Errorf("pipeline %q had parsing warnings, and --reject-parse-warnings is enabled:\n%v", src, w))
```
moskyb: non-blocking, but may maybe something like this would be clearer:

```suggestion
return NewExitError(1, fmt.Errorf(
	"pipeline %q parsed with warnings; refusing upload because --reject-parse-warnings is enabled:\n%v",
	src,
	w,
))
```

moskyb (approved):
awesome 💪

PR #4037, opened 2026-06-30
A-1453: tune cache restore s3 backend concurrency and part size

zhming0:
## Description

Restore downloads were running on the SDK's **upload** defaults (5 streams × 5 MB) and the default HTTP client only kept `MaxIdleConnsPerHost=10` warm, so the download phase was 2–4× slower than `s5cmd`/`aws` and concurrency above ~10 just re-established connections.

- Give downloads their own defaults (c32 / 32 MB parts) instead of inheriting upload defaults. Uploads keep the SDK defaults (c5 / 5 MB). Per the benchmark, c32/p32 reached ~757 MB/s vs ~179 at the default.
- Supply a tuned HTTP client with `MaxIdleConnsPerHost >= concurrency` so parallel parts reuse warm connections.
- Extracted the settings resolution into a pure `resolveTransferSettings` helper and added unit tests.

## Context

A-1453

## Disclosures / Credits

LLM

### DIFF
--- a/internal/cache/store/s3.go
+++ b/internal/cache/store/s3.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log/slog"
+	"net/http"
 	"net/url"
 	"os"
 	"path"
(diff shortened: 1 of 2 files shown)


zhming0:
@buildsworth-bk review again, it's additive, clearly a L1

buildkate (approved):

PR #4044, opened 2026-07-02
Set Partial Clone flags on sparse-checkout

lizrabuya:
## Description

When `BUILDKITE_GIT_SPARSE_CHECKOUT_PATHS` is set, the agent now injects `--sparse` and `--filter=blob:none` into git clone and `--filter=blob:none` into git fetch, so only the paths listed in `BUILDKITE_GIT_SPARSE_CHECKOUT_PATHS` are materialised and blobs outside the sparse set aren't downloaded up front.

Each auto-injected flag respects a user-supplied equivalent — the agent won't add `--sparse` if the user's clone flags already contain it, and won't add `--filter=blob:none` if the user has `--filter=…` in either `BUILDKITE_GIT_CLONE_FLAGS` or `BUILDKITE_GIT_FETCH_FLAGS` (git takes the last `--filter` on the command line, so stacking ours after would silently override theirs; on fetch, `git clone --filter=X` also stores the filter in `remote.origin.partialclonefilter` and subsequent fetches inherit it automatically, so re-passing our own filter would override the user's choice for that fetch).

The whole flow is gated on the local git actually supporting `sparse-checkout set --cone` (>= 2.27). On older git, the agent falls back to a plain full checkout — no partial clone, no `--sparse`, no `--cone` attempt.

The sparse-checkout decision is resolved once at the top of `defaultCheckoutPhase` via `planSparseCheckout`, which runs and validates `git --version` at most once per checkout and only when sparse paths are configured. 
 
## Changes

- add a new `sparseCheckoutPlan` struct and `planSparseCheckout` function to perform git version check only once before clone starts
- version floor bumped from 2.26 → 2.27 (`single-call sparse-checkout set --cone` landed in 2.27; 2.26 only accepted `--cone` on `init`).
- In `defaultCheckoutPhase`, when `sparsePlan.supported`:
  - Append `--sparse` to `gitCloneFlags` unless the user already supplied it.
  - Append `--filter=blob:none` to `gitCloneFlags` unless the user already
    supplied a `--filter` (checked via `hasPartialCloneFilter`).
- `fetchSource` now takes an `autoAddBloblessFilter bool` and prepends   `--filter=blob:none` to fetch flags when   set. The caller computes this as `sparsePlan.supported` && no `--filter` in clone flags && no `--filter` in fetch flags, so a user-supplied filter on either side is respected.
- Add `hasPartialCloneFilter` and `hasSparseCheckoutFlag` helpers in `internal/job/git.go`.
- Added unit and integration tests, including:
  - `TestCheckingOutLocalGitProjectWithSparseCheckoutAutoAddsBlobNoneFilter`
    — verifies both auto-adds happen when the user supplies neither.
  - `TestCheckingOutLocalGitProjectWithSparseCheckoutPreservesUserFilter`
    — verifies a user-supplied `--filter=tree:0` on clone is preserved and no `--filter=blob:none` is prepended to fetch.
  - `TestCheckingOutLocalGitProjectWithSparseCheckoutFallsBackOnOldGit`
    — verifies the git < 2.27 fallback: no auto-added flags, no `sparse-checkout set --cone`, warning appears.

## Testing

## Disclosures / Credits

Claude code drafted all the tests and helped diagnose build failures. I applied and reviewed all changes.


lizrabuya:
@buildsworth-bk , please re-review

internal/job/git.go
```diff
@@ -92,6 +92,32 @@ func gitCheckout(ctx context.Context, sh *shell.Shell, gitCheckoutFlags, referen
...
+	return false
+}
+
+func hasSparseCheckoutFlag(flags []string) bool {
+	for _, f := range flags {
+		if f == "--sparse" {
+			return true
+		}
+	}
+	return false
```
DrJosh9000: There's a standard library function for this:

```suggestion
	return slices.Contains(flags, "--sparse")
```

internal/job/checkout_sparse.go
```diff
@@ -12,6 +12,42 @@ import (
...
+	// shipped in 2.26 but `--cone` was only accepted on the `init` subcommand
+	// then; `set --cone` landed in 2.27. On 2.26.x that call fails, so we
+	// fall back to a full checkout rather than carrying the older two-step.
+	ok, err := gitVersionAtLeast(ctx, e.shell, 2, 27)
+	if err != nil {
+		e.shell.Warningf("Sparse checkout requires git >= 2.27; falling back to full checkout (%v)", err)
+		return sparseCheckoutPlan{paths: paths}
+	}
+	if !ok {
+		e.shell.Warningf("Sparse checkout requires git >= 2.27; falling back to full checkout")
```
DrJosh9000: Might it be useful to log the git version detected?

internal/job/checkout_sparse.go
```diff
@@ -12,6 +12,42 @@ import (
 	"github.com/buildkite/agent/v3/internal/shell"
 )
 
+// sparseCheckoutPlan is the decision, made once at the top of the checkout,
+// about whether sparse checkout applies to this build. Data only — all state
+// mutation happens in setupSparseCheckout.
+type sparseCheckoutPlan struct {
```
DrJosh9000: I'm leaning against this needing to be a struct. `planSparseCheckout` could return `(paths []string, supported bool)`, with both of those passed to `setupSparseCheckout`. Or perhaps only `paths` needs to be passed to `setupSparseCheckout`, and `planSparseCheckout` could return nil for `paths`.

internal/job/checkout.go
```diff
@@ -991,15 +1014,18 @@ func (e *Executor) defaultCheckoutPhase(ctx context.Context) (retErr error) {
 		}
 	}
 
-	if err := e.fetchSource(ctx); err != nil {
+	autoAddBloblessFilter := sparsePlan.supported &&
+		!strings.Contains(e.GitCloneFlags, "--filter") &&
+		!strings.Contains(e.GitFetchFlags, "--filter")
```
DrJosh9000: It's at least theoretically possible that the flags strings could contain `--filter` without it being interpreted as a flag. 

There's a refactoring opportunity here. Strings containing multiple flags being passed somewhere else (to git) could be split into string slices (`shellwords.Split`) earlier on, e.g. when the bootstrap subcommand is starting/parsing its own configuration.

internal/job/checkout_sparse.go
isaacsu: ah we thought of the same thing! See https://github.com/buildkite/agent/pull/4067 😄

isaacsu (commented):
I've raised a PR with suggestions: https://github.com/buildkite/agent/pull/4067

Feel free to incorporate all/some/none 😀

internal/job/checkout.go
```diff
@@ -991,15 +1015,18 @@ func (e *Executor) defaultCheckoutPhase(ctx context.Context) (retErr error) {
 		}
 	}
 
-	if err := e.fetchSource(ctx); err != nil {
+	addBloblessFilter := len(sparsePaths) > 0 &&
+		!strings.Contains(e.GitCloneFlags, "--filter") &&
+		!strings.Contains(e.GitFetchFlags, "--filter")
```
isaacsu: Can we reuse `hasPartialCloneFilter` for this?

isaacsu (commented):
Just one more, then I think we should be good to go.

isaacsu (approved):
Nice work @lizrabuya 🥳

PR #4050, opened 2026-07-06
Conditionally refresh cache TTL on restore (A-1454)

zhming0:
## Description

The S3 cache restore path refreshes a cache object's lifecycle expiration by copying the object to itself (source == dest, `MetadataDirective: REPLACE`), which resets its `LastModified` so the bucket lifecycle rule doesn't expire it.

**Before:** that self-copy ran on _every_ restore. It's a synchronous full-object server-side rewrite — about 27s of blocking work for a 2GB cache — and it hard-errors on objects above S3's 5GB CopyObject limit, failing the whole restore (and the pipeline).

**After:** the self-copy still fires on every restore, but it now carries a `CopySourceIfUnmodifiedSince` precondition set to `now - restoreRefreshInterval`. S3 evaluates the precondition server-side and returns `412 PreconditionFailed` — skipping the copy — whenever the object was modified within the interval. So a hot cache is rewritten at most about once per `restoreRefreshInterval` instead of on every restore. And if the copy does run and fails for any other reason — including the >5GB case — the restore continues normally; the failure is logged as a warning instead of aborting.

**How:** the decision is entirely server-side via the precondition; there's no `HeadObject` and no dependency on the backend `expires_at` (which is decoupled from the actual S3 lifecycle policy and so isn't a reliable signal). The `CopyObject` result is classified into three branches: success (refreshed), `412 PreconditionFailed` (skipped, modified recently — a small `isPreconditionFailed` helper detects it via smithy's `APIError` code), and any other error (soft-fail: log a warning and continue). All three paths return normally.

`restoreRefreshInterval` is a single tunable const, currently `12 * time.Hour`. The actual bucket lifecycle window isn't known to the agent, so this is a heuristic — keep it comfortably below that window.

## Context

Linear ticket: [A-1454](https://linear.app/buildkite/issue/A-1454). The per-restore CopyObject was identified as the largest single cost in the cache-restore investigation.

## Changes

- `internal/cache/store/s3.go`: add the `restoreRefreshInterval` const and the `isPreconditionFailed` helper; in `Download`, issue the self-copy with a `CopySourceIfUnmodifiedSince` precondition and branch on the result (success / 412 skipped / soft-fail on any other error).
- `internal/cache/store/s3_test.go`: table-driven unit tests for `isPreconditionFailed`.

## Testing
- [x] Tests have run locally (`go test ./internal/cache/...` passes; `go build ./...` and `go vet ./internal/cache/...` clean).
- [x] Code is formatted.

## Disclosures / Credits

Implemented with Claude Code.

### DIFF
--- a/internal/cache/store/s3.go
+++ b/internal/cache/store/s3.go
@@ -20,6 +20,7 @@ import (
 	"github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/feature/s3/manager"
 	"github.com/aws/aws-sdk-go-v2/service/s3"
+	smithy "github.com/aws/smithy-go"
 	smithymiddleware "github.com/aws/smithy-go/middleware"
 	"github.com/buildkite/agent/v3/internal/cache/internal/trace"
 	"github.com/buildkite/roko"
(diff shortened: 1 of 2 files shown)


isaacsu:
@zhming0 Some suggestions: https://github.com/buildkite/agent/pull/4052. Feel free to incorporate all/some/none.

zhming0:
@isaacsu they all look good. Merged  👍🏿 .

isaacsu (approved):

PR #4061, opened 2026-07-07
feat: implement cache busting mechanism (part 2/2)

ss1909:
## Description

**Part 2 of 2** (agent) — depends on the backend `POST /expire` endpoint ([[buildkite#...](https://claude.ai/epitaxy/local_43a53550-8fc7-4793-8d50-d185a9e7013b#)](#)) - has been shipped.

#### Problem

A cache entry can go **split-brain**: the registry entry still exists, but its blob was deleted from storage (S3 lifecycle / NSC TTL). Restore then treats the missing-blob download as a hard error (failing the step), and nothing invalidates the entry, so `save`'s `peek` keeps short-circuiting and never re-uploads.

#### Solution

On a genuinely-missing blob, restore now degrades to a clean **cache miss** and best-effort invalidates the stale entry so the next `save` re-uploads it — the cache self-heals.

- **Stores classify not-found distinctly** via a new `store.ErrBlobNotFound` sentinel, so restore can tell a missing blob from a transient transfer error: S3 (typed `*types.NoSuchKey`), NSC (CLI "not found"), local file (`os.IsNotExist`).
- **`CacheEntryExpire` API client** → `POST /cache_registries/:registry/expire`. Any 2xx is success (the backend is idempotent); other statuses return an error.
- **Restore wiring** — on `ErrBlobNotFound`: log, best-effort `expire` the resolved entry (retried with the same curve as `retrieve` — a transient blip shouldn't leave the entry poisoned), clear hit/fallback/restored state, and return a **miss, not an error**.

**Behavior:** a missing blob is a green cache miss (exit 0), never a failed step. The expire is best-effort — failures are logged and swallowed (the restore already degraded to a miss). Fails safe against a backend without the endpoint: the call errors, gets logged, and restore still returns a clean miss (auto-heal just activates once the backend ships).

## Context

https://linear.app/buildkite/issue/A-1438/cache-busting-mechanism

## Changes

**Store: classify "not found" distinctly**
- `internal/cache/store/blob.go` — add `ErrBlobNotFound` sentinel.
- `internal/cache/store/s3.go` — add `isNotFound` (typed `*types.NoSuchKey`); `Download` maps it to `ErrBlobNotFound`.
- `internal/cache/store/nsc.go` — a "not found" CLI failure maps to `ErrBlobNotFound`.
- `internal/cache/store/file.go` — `os.IsNotExist` maps to `ErrBlobNotFound`.

**API client**
- `api/cache.go` — add `CacheEntryExpireReq` and `CacheEntryExpire` (`POST /cache_registries/:registry/expire`; any 2xx = success, else error).
- `internal/cache/client.go` — add `CacheEntryExpire` to the `cacheAPI` interface.

**Restore: degrade to miss + invalidate**
- `internal/cache/restore.go` — on `ErrBlobNotFound`: return a clean miss (not an error), clear `CacheHit`/`FallbackUsed`/`CacheRestored`, and best-effort `invalidateStaleEntry` (expire the resolved entry, retried with the same curve as `retrieve`, failures logged and swallowed).

**Tests**
- `internal/cache/store/{s3,nsc,file}_test.go` — `isNotFound` classification + NSC/local `Download` → `errors.Is(err, ErrBlobNotFound)`.
- `api/cache_test.go` — `CacheEntryExpire` success (method/path/2xx) and non-2xx-is-error.
- `internal/cache/integration_test.go` — end-to-end `TestCacheIntegration_RestoreMissingBlobInvalidates`: save → delete blob → restore degrades to miss + fires one expire → next save re-uploads.

---

**E2E test for the cache-busting flow**

Added an end-to-end test (`TestCacheMissingBlobReupload`, `internal/e2e/`) that exercises the whole split-brain recovery against real Buildkite + S3 + the deployed `/expire` endpoint. As with the other e2e cache tests, the Go test is a thin wrapper (`newTestCase → startAgent → triggerBuild → waitForBuild`, asserting the build passes); the behaviour lives in the fixture pipeline `cache_missing_blob_reupload.yaml`.

**How it simulates a missing blob:** rather than deleting the S3 object (which would need an `s3:DeleteObject` grant the e2e IAM role doesn't have), the `save` step writes to one store prefix (`…/${BUILDKITE_BUILD_ID}/original`) and the later steps use a different, empty prefix (`…/${BUILDKITE_BUILD_ID}/fresh`). The registry entry (in DynamoDB, keyed by cache_key) is shared across steps, but restore against `/fresh` looks for the blob where it isn't → the exact same `NoSuchKey` / missing-blob path, with no delete permission required. Prefixes are per-build so concurrent runs don't collide.

**The four steps** (each `depends_on` the previous; any non-zero exit fails the build):
1. `save` (`/original`) — writes `cache-src/file.txt`, `cache save` → creates the entry + uploads the blob.
2. `restore_miss` (`/fresh`) — `cache restore`: the entry exists but its blob isn't in `/fresh`, so restore must **exit 0 (clean miss, not a hard failure)**, must **not** populate the target, and best-effort invalidates the stale entry via `/expire`.
3. `resave` (`/fresh`) — recreate the file, `cache save`: because the entry was invalidated, the existence peek misses and it **re-uploads**.
4. `restore_final` (`/fresh`) — `cache restore` → now a **hit**; asserts the file is restored with the original content.

So it proves the full loop: detect missing blob → degrade to a clean miss → invalidate → re-upload → cache heals.

---

## Testing

## Disclosures / Credits

Implemented in consultation with my buddy, Claude


api/cache.go
```diff
@@ -235,6 +243,28 @@ func (c *Client) CacheEntryRetrieve(ctx context.Context, registry string, retrie
...
+	if err != nil {
+		return nil, cacheSpanErr(span, "failed to create request: %w", err)
+	}
+
+	apiResp, err := c.cacheDo(req, &struct{}{})
+	if err != nil {
+		return apiResp, cacheSpanErr(span, "%w", err)
+	}
+	switch apiResp.StatusCode {
+	case http.StatusOK, http.StatusNoContent, http.StatusNotFound:
```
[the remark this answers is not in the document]
ss1909: fixed

internal/cache/restore.go
```diff
@@ -176,6 +176,23 @@ func (c *client) Restore(ctx context.Context, cacheID string) (RestoreResult, er
 	// Download cache
 	tmpDir, archiveFile, transferInfo, err := c.downloadCache(ctx, retrieveResp, c.bucketURL)
 	if err != nil {
+		if errors.Is(err, store.ErrBlobNotFound) {
+			slog.Warn("cache blob missing, treating as miss and invalidating entry",
+				"cache_id", cacheID, "err", err)
+			c.invalidateStaleEntry(ctx, retrieveResp)
+			result.CacheHit = false
```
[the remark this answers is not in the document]
ss1909: fixed

internal/e2e/fixtures/cache_missing_blob_reupload.yaml
zhming0, line 1: Do we want to consider using yaml anchors? 

https://buildkite.com/docs/pipelines/integrations/plugins/using#using-yaml-anchors-with-plugins-overriding-yaml-anchors

internal/cache/store/nsc.go
zhming0, line 1: I understand we can't write a test for this bit just yet, just checking if you have tested this CLI return format in a HA build/step?

internal/cache/restore.go
```diff
@@ -176,6 +176,26 @@ func (c *client) Restore(ctx context.Context, cacheID string) (RestoreResult, er
...
+			// state set earlier so callers and the span reflect a clean miss.
+			result.CacheHit = false
+			result.FallbackUsed = false
+			result.CacheRestored = false
+			result.TotalDuration = time.Since(startTime)
+			span.SetAttributes(
+				attribute.Bool("cache.hit", false),
+				attribute.Bool("cache.restored", false),
+				attribute.Bool("cache.invalidated", invalidated),
+				attribute.Int64("cache.duration_ms", result.TotalDuration.Milliseconds()),
```
zhming0: Otel span comes with an duration attribute, this might be redundent?

zhming0 (commented):
LGTM 👍🏿 I just have a question.

internal/cache/store/nsc.go
ss1909: yep - https://buildkite.com/buildkite/sneha-test-pipeline/builds/2/canvas?tab=output&jid=019f4975-dff4-4ce3-9704-2ea2fef2471d#L61
I've kept it generic to "not found".

internal/e2e/fixtures/cache_missing_blob_reupload.yaml
ss1909: Ooo, I've never used it before - thanks for the suggestion! I've implemented it :)

internal/cache/restore.go
ss1909: I was just following the save.go to be consistent - https://github.com/buildkite/agent/blob/7c597f5a2c23e3a33ceec8210212aed49138ca7f/internal/cache/save.go#L145
happy to remove it.

internal/cache/store/nsc.go
zhming0: (I can't access this build log, it may be good to have a screenshot) 
But I trust you verified it so ✅

zhming0 (approved):

internal/cache/store/nsc.go
ss1909: image

PR #4067, opened 2026-07-07
Suggestions for #4044

isaacsu:
Suggestions for #4044

- Replace **sparseCheckoutPlan** struct with a plain **[]string**
- Rename **planSparseCheckout** → **resolveSparseCheckout**, which now returns just **[]string** or **nil** when no paths have been requested or git is too old.
- Update **setupSparseCheckout** to take sparsePaths **[]string** instead of a **plan**. Empty/nil means full checkout (disabling any prior sparse config).
- Replace **hasSparseCheckoutFlag** with inline **slices.Contains(gitCloneFlags, "--sparse")**  
- Update tests and tighten up commentary

### DIFF
--- a/internal/job/checkout.go
+++ b/internal/job/checkout.go
@@ -762,7 +762,7 @@ func (e *Executor) getOrUpdateMirrorDir(ctx context.Context, repository string) 
 
 // fetchSource fetches the git source for the job. If GitSkipFetchExistingCommits is
 // enabled and the commit already exists locally, the fetch is skipped entirely.
-func (e *Executor) fetchSource(ctx context.Context, autoAddBloblessFilter bool) error {
+func (e *Executor) fetchSource(ctx context.Context, addBloblessFilter bool) error {
 	// If configured, skip the fetch when the commit already exists locally.
 	// This is useful when a pre-populated git mirror is used with --reference,
 	// as the commit objects are already reachable and fetching is redundant.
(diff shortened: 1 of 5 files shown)


internal/job/integration/checkout_integration_test.go
```diff
@@ -250,8 +250,8 @@ func TestCheckingOutLocalGitProjectWithSparseCheckoutFallsBackOnOldGit(t *testin
 // sparse checkout paths are configured but the user did NOT supply any
 // --filter in BUILDKITE_GIT_CLONE_FLAGS. The other sparse integration tests
 // all pre-supply --filter=blob:none, which short-circuits the auto-add and
-// therefore doesn't exercise enableCloneFlagsForSparseCheckout's "add it"
```
isaacsu, line 253: Couldn't find any reference to `enableCloneFlagsForSparseCheckout`

internal/job/checkout.go
```diff
@@ -762,7 +762,7 @@ func (e *Executor) getOrUpdateMirrorDir(ctx context.Context, repository string)
 
 // fetchSource fetches the git source for the job. If GitSkipFetchExistingCommits is
 // enabled and the commit already exists locally, the fetch is skipped entirely.
-func (e *Executor) fetchSource(ctx context.Context, autoAddBloblessFilter bool) error {
+func (e *Executor) fetchSource(ctx context.Context, addBloblessFilter bool) error {
```
isaacsu, line 765: The "auto" decision happens outside the function

PR #4076, opened 2026-07-10
Improve log URL to point to specific job

mitchhentgesspotify:
Without this change, the user has to manually create the URL, which is tedious when debugging agents.

## Testing

Spotify

### DIFF
--- a/agent/run_job.go
+++ b/agent/run_job.go
@@ -77,7 +77,7 @@ func (r *JobRunner) Run(ctx context.Context, ignoreAgentInDispatches *bool) (err
 		return errors.New("job already cancelled before running")
 	}
 
-	r.agentLogger.Infof("Starting job %s for build at %s", r.conf.Job.ID, r.conf.Job.Env["BUILDKITE_BUILD_URL"])
+	r.agentLogger.Infof("Starting job at %s?jid=%s", r.conf.Job.Env["BUILDKITE_BUILD_URL"], r.conf.Job.ID)
 
 	ctx, done := status.AddItem(ctx, "Job Runner", "", nil)
 	defer done()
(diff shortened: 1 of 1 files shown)


agent/run_job.go
```diff
@@ -77,7 +77,7 @@ func (r *JobRunner) Run(ctx context.Context, ignoreAgentInDispatches *bool) (err
 		return errors.New("job already cancelled before running")
 	}
 
-	r.agentLogger.Infof("Starting job %s for build at %s", r.conf.Job.ID, r.conf.Job.Env["BUILDKITE_BUILD_URL"])
+	r.agentLogger.Infof("Starting job at %s?jid=%s", r.conf.Job.Env["BUILDKITE_BUILD_URL"], r.conf.Job.ID)
```
petetomasik: Two other places in the agent already build a link to a specific job using the URL fragment form (`<build_url>#<job-id>`):

- [`agent/json_job_logger.go:30`](https://github.com/buildkite/agent/blob/f2e8269f0b9554d63b8701bde7bb81d2111644a0/agent/json_job_logger.go#L30)
- [`internal/job/tracing.go:231`](https://github.com/buildkite/agent/blob/9b8f49df7389305d0b399955d5f1a8c40fbf4bdb/internal/job/tracing.go#L231)

This change introduces a second convention (`?jid=<job-id>`) for the same thing. Both forms resolve to the job on the classic and new build UIs, so this is about consistency rather than correctness, but it'd be good to match the existing pattern so all three call sites read the same. The same swap applies to the "Finished job" line (line 460).

```suggestion
	r.agentLogger.Infof("Starting job at %s#%s", r.conf.Job.Env["BUILDKITE_BUILD_URL"], r.conf.Job.ID)
```

mitchhentgesspotify: Good call!
(Interestingly, when I verified this locally, adding `#<job-id>` _doesn't work_ when the URL is at `<build-url>/canvas#<job-id>`, but rather only when `/canvas` isn't specified: `<build-url>#<job-id>`).

isaacsu:
Heya @mitchhentgesspotify, thanks for your contribution 💚. I've merged it as part of  https://github.com/buildkite/agent/pull/4083

PR #4084, opened 2026-07-14
Constant-time token comparison

DrJosh9000:
## Description

Reduce spurious vuln reports by comparing the socket token in constant time.

## Context

https://linear.app/buildkite/issue/FODA-1611

## Changes

Use `subtle.ConstantTimeCompare`.

An alternative would be to hash (e.g. SHA) both tokens and compare them in variable time, as that would destroy the relationship between comparison time and token value.

## Testing

## Disclosures / Credits

As suggested in the ticket.

### DIFF
--- a/internal/socket/middleware.go
+++ b/internal/socket/middleware.go
@@ -1,6 +1,7 @@
 package socket
 
 import (
+	"crypto/subtle"
 	"maps"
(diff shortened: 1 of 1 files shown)


zhming0 (approved):
LoL :shipit:

PR #4087, opened 2026-07-15
fix: Map expired artifact to ErrBlobNotFound

ss1909:
## Description

Map nsc's "the artifact has expired" download error to ErrBlobNotFound (alongside "not found"), so an expired HA cache entry is treated as a cache miss and invalidated instead of failing the build.

## Context

https://linear.app/buildkite/issue/A-1541/namespace-artifact-expiry-breaks-ha-cache-bust-mechanism

## Testing


zhming0 (approved):
LGTM with a really minor note that this is a OSS repo. Some people might audit our changes, so we should make PR description descriptive for 3rd party reader 🙂

PR #4088, opened 2026-07-15
Fix redaction leak from ssh_secret within checkout

JoeColeman95:
## Description

When a checkout uses `ssh_secret`, the fetched key is set as `BUILDKITE_GIT_SSH_KEY` in the job environment. That var is also an executor config var, so after any hook runs the executor logs it as a changed config value. That log line formats values with `%q`, which escapes the key's newlines into literal backslash n sequences.

## Context

[Linear
](https://linear.app/buildkite/issue/SUP-7552/private-ssh-key-leaked-in-logs)

Reported via support. A customer's build log showed their SSH deploy key in plaintext after a global environment hook ran.

## Changes

- Redact known secrets from env change log values before `%q` formatting in `applyEnvironmentChanges`, via a new `redactedValue` helper
- Add `Mux.Needles` to `internal/replacer`, returning the deduplicated needle set across replacers
- Add a regression test mapping a multiline secret onto `BUILDKITE_GIT_SSH_KEY` with an environment hook present, plus unit tests for `Mux.Needles`

## Testing
- [ x ] Tests have run locally (with `go test ./...`). Buildkite employees may check this if the pipeline has run automatically.
- [ x ] Code is formatted (with `go tool gofumpt -extra -w .`)

## Disclosures / Credits

Fable 5 traced the leak path for me and explained what was occurring, also wrote the secrets integration test so I could validate.

### DIFF
--- a/clicommand/meta_data_set_batch.go
+++ b/clicommand/meta_data_set_batch.go
@@ -81,7 +81,8 @@ var MetaDataSetBatchCommand = cli.Command{
 			}
 		}
 
-		return setMetaDataBatch(ctx, cfg, l, items)
+		client := api.NewClient(l, loadAPIClientConfig(cfg, "AgentAccessToken"))
+		return setMetaDataBatch(ctx, client, cfg.Job, l, items)
 	},
 }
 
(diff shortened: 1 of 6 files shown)


clicommand/meta_data_set_batch.go
DrJosh9000, line 1: These changes under clicommand (to `meta-data set-batch`) seem totally unrelated?

clicommand/meta_data_set_batch_test.go
```diff
@@ -90,14 +90,38 @@ func TestParseMetaDataBatchArgs(t *testing.T) {
...
 	t.Parallel()
 
 	t.Run("success", func(t *testing.T) {
 		t.Parallel()
 
+		l := logger.NewBuffer()
+
 		var receivedBatch api.MetaDataBatch
-		server := httptest.NewServer(http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+		client := stubAPIClient(l, func(req *http.Request) (*http.Response, error) {
```
DrJosh9000, line 124: Please don't do this. Using `httptest` and having an actual network round-trip (even if it's the same host) makes the test more faithful than stubbing out the client.

DrJosh9000:
I've taken a stab at an alternative in #4091

zhming0:
Since we merged https://github.com/buildkite/agent/pull/4091, I will close this PR for now. 🙏🏿

PR #4089, opened 2026-07-15
fix(core): restore FinishJob retries on transport errors

ivannalisetska:
- Only stop finish retries on 401/422 API responses
- Keep retrying unknown transport errors such as http2 connection lost

SUP: #7554

## Description
Successful jobs can exit 0 but then fail the Agent API finish call with a transient transport error such as http2: client connection lost. After the retry-policy cleanup in 3.124.0, FinishJob used BreakOnNonRetryable, which treated that error as non-retryable and gave up after one attempt. The agent then idled/disconnected and the job showed as agent lost (-1).

This PR restores the previous FinishJob retry behavior: only stop on 401/422 responses, and keep retrying unknown transport errors.

Alternative considered: adding http2: client connection lost to the shared IsRetryableError allowlist. That would help other API callers, but would not restore the original finish semantics (retry any non-401/422 failure). Finish is critical enough to keep the more resilient path.

## Context
SUP-7554
Customer report after upgrade from agent 3.114.1 → 3.130.0: finish fails with http2: client connection lost, then idle disconnect marks the job agent lost (-1)
## Changes
Restore FinishJob retry break conditions to 401/422 only in core/client.go
Add regression tests for retrying http2: client connection lost and not retrying 401
## Testing

 Tests have run locally (with go test ./...). Buildkite employees may check this if the pipeline has run automatically.
## Disclosures / Credits
I compared the previous FinishJob retry logic with the post-cleanup version. AI helped revert the deleted behavior and add regression tests.

### DIFF
--- a/core/client.go
+++ b/core/client.go
@@ -229,24 +229,11 @@ func (c *Client) FinishJob(ctx context.Context, job *api.Job, finishedAt time.Ti
 	).DoWithContext(ctx, func(retrier *roko.Retrier) error {
 		response, err := c.APIClient.FinishJob(ctx, job, ignoreAgentInDispatches)
 		if err != nil {
-			// If the API returns with a 422, that means that we successfully tried to
-			// finish the job, but Buildkite rejected the finish for some reason. This
-			// can sometimes mean that Buildkite has cancelled the job before we get a
-			// chance to send the final API call (maybe this agent took too long to kill
-			// the process).
-			// The API may also return a 401 when job tokens are enabled.
-			// In either case, we don't want to keep trying to finish the job forever so
-			// we'll just bail out and go find some more work to do.
-			//
-			// Unlike other API calls, we intentionally do not use
-			// api.BreakOnNonRetryable here: finish is critical, so unknown transport
-			// errors (e.g. "http2: client connection lost") must keep retrying.
-			if response != nil && (response.StatusCode == 422 || response.StatusCode == 401) {
-				c.Logger.Warnf("Buildkite rejected the call to finish the job (%s)", err)
-				retrier.Break()
-				return err
+			// Non-retryable responses (e.g. 422, or 401 when job tokens are being used) mean the job has been cancelled or
+			// otherwise already finished. We should stop trying and go find more work to do.
+			if !api.BreakOnNonRetryable(retrier, response, err) {
+				c.Logger.Warnf("%s (%s)", err, retrier)
 			}
-			c.Logger.Warnf("%s (%s)", err, retrier)
 		}
 
 		return err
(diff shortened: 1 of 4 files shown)


isaacsu:
Screenshot 2026-07-17 at 10 54 14 am

I think this is what your change will result in (the yellow rows).

If I understand correctly, the only change we really want is Row 10. Rows 2-4 and Row 6 would be regressions unless we truly consider those status codes to be retriable in this specific case.

I think keeping the current use of `api.BreakOnNonRetryable` and just adding `http2: client connection lost` as retriable could be sufficient to solve the customer issue while maintaining the desired current behaviour for Rows 2-4 and 6?

core/client.go
zhming0, line 1: _Agent sweep drives-by_. 

I agree with @isaacsu's feedback. 

We should broadly consider `http2: client connection lost` transient in the shared error classification path.

core/client.go
```diff
@@ -229,11 +229,24 @@ func (c *Client) FinishJob(ctx context.Context, job *api.Job, finishedAt time.Ti
...
+			// the process).
+			// The API may also return a 401 when job tokens are enabled.
+			// In either case, we don't want to keep trying to finish the job forever so
+			// we'll just bail out and go find some more work to do.
+			//
+			// Unlike other API calls, we intentionally do not use
+			// api.BreakOnNonRetryable here: finish is critical, so unknown transport
+			// errors (e.g. "http2: client connection lost") must keep retrying.
+			if response != nil && (response.StatusCode == 422 || response.StatusCode == 401) {
+				c.Logger.Warnf("Buildkite rejected the call to finish the job (%s)", err)
```
zhming0: This log is probably from past commits, and it isn't the most accurate description for this situation. 

There are many fatal cases that isn't caused by Buildkite rejecting the call.

zhming0 (dismissed):

core/client.go
ivannalisetska: Updated  with : 
Keep BreakOnNonRetryable in FinishJob (avoids regressions on 402–421 / 423–428)
Classify http2: client connection lost as retriable in the shared error path
Also removed the old "Buildkite rejected the call to finish the job" log from the earlier approach.

core/client_finish_job_test.go
zhming0, line 1: Since this PR is no longer about fixing this one endpoint anymore, this test case might not be needed either. 

And the PR title/description can use a bit update too after the direction change 🙏🏿

zhming0 (commented):
LGTM with a minor comment 🙏🏿

zhming0:
@ivannalisetska I've taken your changes into #4110  (your authorship remains 👍🏿 ). We will take it from here 🙏🏿 . Thanks for leading the way!

PR #4090, opened 2026-07-16
feat: add --job-context-dir to configure job coordination file locations

aeijdenberg-canva:
## Description

When running under k8s we would like more control over where the `start` command writes it's socket and other job information which the other containers need to talk about to ti.

## Context

The agent creates several files to coordinate with the processes running a job: the job env files (`BUILDKITE_ENV_FILE` / `BUILDKITE_ENV_JSON_FILE`), the job timeout marker file (`BUILDKITE_AGENT_JOB_TIMEOUT_FILE`), and, in Kubernetes mode, the socket that `kubernetes-bootstrap` connects to. The directory for these was previously hardcoded to `/workspace` in Kubernetes mode.

## Changes

This adds a `--job-context-dir` flag (`BUILDKITE_JOB_CONTEXT_DIR`) to both `agent start` and `kubernetes-bootstrap`. All coordination files are created in that directory, with the socket at a predictable name within it (`buildkite.sock`). 

Defaults are unchanged (os.TempDir(), or /workspace under --kubernetes-exec).

The env var is included in `kubernetes-bootstrap`'s `existingEnvPriority` so that each container's own value survives the `registration-env` overlay (the shared volume may be mounted at different paths per container), and in `protectedEnv `so job-level env cannot redirect it.

## Testing

Also deployed into our k8s environment with the new var set to a different directory and jobs ran fine.

Canva

## Disclosures / Credits

- Claude helped write this change and test, I reviewed and asked it to make a bunch of changes along the way.

### DIFF
--- a/agent/agent_configuration.go
+++ b/agent/agent_configuration.go
@@ -45,6 +45,7 @@ type AgentConfiguration struct {
 	RunInPty                        bool
 	KubernetesExec                  bool
 	KubernetesContainerStartTimeout time.Duration
+	JobContextDir                   string
 
 	SigningJWKSFile  string // Where to find the key to sign pipeline uploads with (passed through to jobs, they might be uploading pipelines)
 	SigningJWKSKeyID string // The key ID to sign pipeline uploads with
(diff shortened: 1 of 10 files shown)


isaacsu:
Hi @aeijdenberg-canva, just wanted to acknowledge and thank you for your contribution. We look forward to reviewing it soon!

zhming0:
We included this change in the latest agent release 👍🏿

PR #4091, opened 2026-07-16
Redact Go-escaped needles

DrJosh9000:
## Description

Alternative to #4088, that also catches multiline secrets that might have passed through Go `%q` formatting within commands, etc.

Buildsworth also liked to point out that the JSON encoder defaults to escaping a couple of HTML-sensitive chars, which could interfere with the secret redactor, so I have gone around disabling the HTML escaping in situations where the output is more likely to end up in log output than in a JS context.

## Context

https://linear.app/buildkite/issue/SUP-7552/private-ssh-key-leaked-in-logs

## Changes

- Add `GoEscaped` and `AppendGoEscaped` helpers to `redact`
- When creating a replacer with `redact.New`, add `GoEscaped` versions of the needles
- When adding new needles in executor / pipeline upload, also add `GoEscaped` versions
- Disable HTML escaping in several places

## Testing

## Disclosures / Credits

Straight from the top of my dome

### DIFF
--- a/clicommand/env_dump.go
+++ b/clicommand/env_dump.go
@@ -56,6 +56,7 @@ var EnvDumpCommand = cli.Command{
 		}
 
 		enc := json.NewEncoder(c.App.Writer)
+		enc.SetEscapeHTML(false) // HTML escapes may interfere with secret redaction
 		if cfg.Format == "json-pretty" {
 			enc.SetIndent("", "  ")
 		}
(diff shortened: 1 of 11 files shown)


DrJosh9000:
@buildsworth-bk review, it's ok to approve for L2 risk

DrJosh9000:
@buildsworth-bk review, approval OK at L2 risk

buildkate:
@DrJosh9000 what do you think about copying over the integration test from #4088  (TestSecretsIntegration_ConfigVarSecretNotLoggedByEnvChange)?

buildkate (approved):

PR #4094, opened 2026-07-16
Fix commit verification with non-default branches

petetomasik:
## Description

`checkout.commit_verification` (`BUILDKITE_GIT_COMMIT_VERIFICATION`) verifies that a build's commit is an ancestor of the build's branch before checkout, catching a build pointed at a commit that isn't actually on the claimed branch. Under `strict` it should fail the build; under `warn` it logs and continues.

The check only ever enforced on a repository's default branch. `checkCommitOnBranch` ran `git merge-base --is-ancestor <commit> <branch>` against the bare `BUILDKITE_BRANCH` name, but at verification time only the default branch exists as a local ref (the clone checks it out; the build's own branch is fetched as a detached commit, not a local branch ref). Git does not resolve a bare name like `feature` to `refs/remotes/origin/feature`, so merge-base exited 128 for any non-default branch. That 128 was classified as "unavailable", which is warn-only even under `strict`, so a build on a non-default branch with a commit genuinely not on that branch ran to completion.

The fix fetches the branch tip, pins it to a SHA, and checks ancestry against that SHA instead of the unresolvable bare name. The branch is fetched as `refs/heads/<branch>` so a tag sharing the branch's name can't shadow it, using a direct `git fetch` rather than the shared fetch helper (which word-splits its refspecs and would corrupt a branch name containing shell metacharacters that are legal in git refs). The fetch is retried a few times before giving up, so a transient failure doesn't silently skip verification. Shallow clones are handled by trusting a "not an ancestor" result only once the repo is non-shallow (a shallow clone can report a genuine ancestor as "not an ancestor" when the connecting history is beyond the shallow boundary), deepening as needed. A merge-base that fails to run at all is treated as "unavailable" rather than a definitive failure. The checkout retry loop now fast-breaks on a definitive failure so a provably-off-branch commit fails immediately instead of re-cloning through the full backoff.

A second bug disabled verification on nearly every build: `verifyCommit` skipped whenever `BUILDKITE_PULL_REQUEST` was non-empty, but the agent sets that variable to the string `"false"` (not empty) on every non-PR build. The check therefore returned early before ever running. It now skips only for real pull-request builds, matching how the rest of the checkout treats the `"false"` sentinel.

One behavior was left deliberately unchanged: if fetching the branch keeps failing (branch deleted or renamed on the remote, or the network stays down through the retries), the check reports "unavailable" and does not block, even under `strict`. This matches the existing documented design that infrastructure failures never block verification, so users don't disable it over false positives.

## Context

[Linear](https://linear.app/buildkite/issue/SUP-7566/checkoutcommit-verification-not-blocking-job-when-commit-not-in-branch). Found while testing `checkout.commit_verification: strict` in a pipeline: a build created from a non-default branch with a commit SHA not on that branch allowed the jobs to complete instead of failing.

## Changes

- `internal/job/commit_verification.go`: rewrite `checkCommitOnBranch` to fetch and pin the branch tip (as `refs/heads/<branch>`, via a direct `git fetch` so the ref isn't word-split, and with a bounded retry), check ancestry against the SHA, trust "not an ancestor" only when the repo is non-shallow, and treat a non-exit merge-base error as unavailable.
- `internal/job/commit_verification.go`: skip verification only for real pull-request builds (`BUILDKITE_PULL_REQUEST` is the string `"false"`, not empty, on ordinary builds), and drop a duplicated prefix from the warn-mode log messages.
- `internal/job/checkout.go`: fast-break the checkout retry loop on `ErrCommitVerificationFailed` so a provably-off-branch commit fails immediately.
- Tests: shallow off-branch failure, fast-break (verification runs once, not per retry), unavailable-stays-warn under `strict`, a strengthened warn-mode assertion, a regression test for a build on a non-default branch, a valid commit on a non-default branch that must pass, a non-PR build (`BUILDKITE_PULL_REQUEST=false`) that must still verify, a commit reachable only after `--unshallow`, a transient branch-fetch failure that must retry rather than degrade, and a branch name containing shell metacharacters.
- Tests: the `file://`-backed cases share a fixture helper and use a bare repo served over `file://`, because shallow deepening over the HTTP test server's stateless-rpc transport is not reliable across git versions.

No CLI argument changes.

## Testing

## Disclosures / Credits

I used Claude Code (Opus 4.8) to diagnose the bug, implement the fix and tests, run a structured code review, and address the findings, all under my direction and review. I made the design decisions (including keeping the "unavailable never blocks" behavior under `strict`) and verified the changes fixed the issue with non-default branches.

### DIFF
--- a/internal/job/checkout.go
+++ b/internal/job/checkout.go
@@ -165,6 +165,12 @@ func (e *Executor) checkout(ctx context.Context) error {
 			var errGit *gitError
 
 			switch {
+			case errors.Is(err, ErrCommitVerificationFailed):
+				// A commit that is provably not on its branch won't become valid by
+				// retrying, so fail fast instead of re-cloning through the whole backoff.
+				e.shell.Warningf("Checkout failed! %s", err)
+				r.Break()
+
 			case errors.Is(err, errCheckoutAttemptTimedOut):
 				// The per-attempt timeout fired and git was signal-killed.
 				// Treat this like a generic transient failure: warn, clean
--- a/internal/job/commit_verification.go
+++ b/internal/job/commit_verification.go
@@ -5,8 +5,11 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/buildkite/agent/v3/internal/shell"
+	"github.com/buildkite/roko"
+	"github.com/buildkite/shellwords"
 )
 
 // ErrCommitVerificationFailed indicates that git has definitively determined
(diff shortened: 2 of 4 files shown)


dahtey-bk (approved):

omehegan:
@petetomasik I think this is good to merge now, but I would suggest squashing your commits down to 1, as this repo doesn't have squash merge enabled. I was going to do that for you and merge it, but there are other commits mixed in with yours and I decided better to wait and let you do it.

petetomasik:
@omehegan - all sorted. 👍

PR #4100, opened 2026-07-20
Nice the command subprocess so k8s boostrap can't be starved of CPU

omehegan:
## Description

When a CPU-intensive command consumes all available CPU within a K8s command container, it can starve the kubernetes-bootstrap process of scheduler cycles. The StatusLoop goroutine can't complete its once-per-second RPC call to the Runner, LastHeardFrom exceeds the 30-second ClientLostTimeout, and the Runner marks the client as StateLost. The customer sees "container stopped communicating without exiting normally" even though nothing was OOM-killed.

To address this, we launch the `buildkite-agent bootstrap` process at `nice 5` so that the root process has priority on the CPU. That should help prevent a heavy command workload from starving out the status check and causing the agent to be "lost."

## Context

SUP-7567

## Changes

As above.

## Testing

## Disclosures / Credits

I used Claude to plan, implement, and test the change.

### DIFF
--- a/clicommand/kubernetes_bootstrap.go
+++ b/clicommand/kubernetes_bootstrap.go
@@ -225,6 +225,7 @@ var KubernetesBootstrapCommand = cli.Command{
 			Stderr:            io.MultiWriter(os.Stderr, socket),
 			Dir:               buildPath,
 			PTY:               runInPTY,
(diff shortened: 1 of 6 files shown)


DrJosh9000 (approved):
LGTM

zhming0 (approved):

PR #4113, opened 2026-07-21
Dedupe needles in replacer using map

DrJosh9000:
## Description

Make deduplicating needles in the string replacer more efficient (in big-O terms).

```plain
goos: darwin
goarch: arm64
pkg: github.com/buildkite/agent/v3/internal/replacer
cpu: Apple M5 Pro
            │ replacer-bench-main.txt │       replacer-bench-map.txt       │
            │         sec/op          │   sec/op     vs base               │
Replacer-15               7.857µ ± 2%   7.800µ ± 1%  -0.73% (p=0.000 n=30)
```

## Context

While reviewing #3979, I remembered that `unsafeAdd` is worst-case quadratic in the number of needles being added. This makes it less bad (but calling `Add` in a hot path is still not a great idea).

## Changes

Add a map for tracking needles in the replacer. Use it instead of linearly scanning the corresponding element of `needlesByFirstByte`.

## Testing

## Disclosures / Credits

I happen to know how all this works.

### DIFF
--- a/internal/replacer/replacer.go
+++ b/internal/replacer/replacer.go
@@ -4,6 +4,7 @@ package replacer
 import (
 	"fmt"
 	"io"
+	"maps"
 	"slices"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


DrJosh9000:
@buildsworth-bk review, approve at L2 risk

moskyb (approved):
looks great!

PR #4115, opened 2026-07-22
Fix unknown Kubernetes container exit statuses

jamiemonserrate:
## Description

Fix a race during Kubernetes job shutdown where the runner can terminate before the command container sends its `Exit` RPC. In that case, the container remains connected but its zero-value exit status was reported to Buildkite as a genuine success.

The runner now returns a synthetic exit status of `137` when the container's terminal status is unavailable.

This follows Kubernetes' convention for previously running containers that can no longer be located:

- [Kubelet synthesising exit code 137](https://github.com/kubernetes/kubernetes/blob/faba7e6a94df833524a619874172cd77fdeaf760/pkg/kubelet/kubelet_pods.go#L2393-L2403)
- [`ContainerStatusUnknown` definition](https://github.com/kubernetes/kubernetes/blob/faba7e6a94df833524a619874172cd77fdeaf760/pkg/kubelet/container/runtime.go#L307-L308)

Explicitly reported exit statuses, including exit `0`, remain unchanged.

## Context
- https://buildkite.com/jamie-21-july-2026/buildkite-testing/builds/7/summary shows a sample passing build, that should have failed

## Changes

- Return synthetic exit status `137` when a connected Kubernetes client never reports its terminal status.
- Preserve explicit successful and non-zero exit statuses.
- Add single- and multi-container regression coverage.

## Testing

- [x] Full test suite passed locally: 2,706 tests passed, 10 skipped.
- [x] Kubernetes package tests passed.
- [x] `golangci-lint run` passed with zero issues.
- [x] Code is formatted with `gofumpt`.
- [x] https://buildkite.com/jamie-21-july-2026/buildkite-testing/builds/10/summary

## Deployment / Rollback

No configuration or migration is required. This ships with the agent binary and can be rolled back by reverting this change.

## Disclosures / Credits

OpenAI Codex investigated the failure mode, implemented the change and regression tests, and ran verification under Jamie Monserrate's direction.

### DIFF
--- a/kubernetes/runner.go
+++ b/kubernetes/runner.go
@@ -23,7 +23,10 @@ func init() {
 	gob.Register(new(syscall.WaitStatus))
 }
 
-const defaultSocketPath = "/workspace/buildkite.sock"
+const (
+	defaultSocketPath          = "/workspace/buildkite.sock"
+	unknownContainerExitStatus = 137
+)
 
 type RunnerConfig struct {
 	SocketPath         string
@@ -227,16 +230,22 @@ func (r *Runner) WaitStatus() process.WaitStatus {
 		exitStatus, state := client.ExitStatus, client.State
 		client.mu.Unlock()
 
-		if exitStatus != 0 {
-			return waitStatus{Code: exitStatus}
-		}
-
-		// use an unusual status code to distinguish unusual states
 		switch state {
+		case StateExited:
+			if exitStatus != 0 {
+				return waitStatus{Code: exitStatus}
+			}
 		case StateLost:
 			return waitStatus{Code: -7}
 		case StateNotYetConnected:
 			return waitStatus{Code: -10}
+		default:
+			// During agent shutdown, a long-running post-command hook can outlast
+			// the signal grace period. The runner then stops before the connected
+			// client reports its terminal status. Match Kubernetes' synthetic exit
+			// status rather than treating the zero-value exit status as success.
+			// Any future client state is also unknown by default.
+			return waitStatus{Code: unknownContainerExitStatus}
 		}
 	}
 	return waitStatus{}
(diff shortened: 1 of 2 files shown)


kubernetes/runner.go
```diff
@@ -233,6 +236,11 @@ func (r *Runner) WaitStatus() process.WaitStatus {
 
 		// use an unusual status code to distinguish unusual states
 		switch state {
+		case StateConnected:
+			// The runner stopped before the client reported its terminal status.
+			// Match Kubernetes' synthetic exit status for an unknown container
+			// status rather than treating the zero-value exit status as success.
+			return waitStatus{Code: unknownContainerExitStatus}
```
zhming0: What do you think about deny by default? so if we ever introduce new state, this can cover it? 

```
  default: 
    return waitStatus{Code: unknownContainerExitStatus}
```

kubernetes/runner.go
```diff
@@ -233,6 +236,11 @@ func (r *Runner) WaitStatus() process.WaitStatus {
 
 		// use an unusual status code to distinguish unusual states
 		switch state {
+		case StateConnected:
+			// The runner stopped before the client reported its terminal status.
+			// Match Kubernetes' synthetic exit status for an unknown container
+			// status rather than treating the zero-value exit status as success.
```
zhming0: Would it be good to document the scenario for this happen? (post-command runs for a long time etc)

zhming0 (approved):
LGTM with two non-blocking recommendation!

kubernetes/runner.go
jamiemonserrate: Good suggestion. Will do.

kubernetes/runner.go
jamiemonserrate: I see your point. So by default, we assume failure - unless we get an explicit message/return value that the container exited with 0.... I like it. Let me give it a try.

PR #4118, opened 2026-07-23
fix (cache): Migrate cache S3 manager to AWS transfermanager

ss1909:
## Description

The agent cache S3 store used the deprecated aws-sdk-go-v2 `feature/s3/manager` uploader/downloader for cache save and restore. AWS has superseded that package with `feature/s3/transfermanager`, which the agent already uses for artifact multipart downloads. This migrates cache save/restore onto `transfermanager`, aligning cache transfers with the maintained API and letting us drop the `manager`-related `//nolint:staticcheck` (SA1019) suppressions from the cache package.

Restore uses explicit range-based downloads (`GetObjectType = GetObjectRanges`) instead of the SDK's default part-number fan-out. Part-number fan-out only parallelises objects uploaded as multipart — a single `PutObject` (or a `CopyObject`, e.g. our TTL refresh) collapses to a single stream. Range mode fans out for any object, so restore parallelism is a property of our configured concurrency/part size rather than of how each object happened to be written. Save uses `transfermanager.UploadObject`.

Unlike the old `manager`, which streamed parts straight from the file, `transfermanager.UploadObject` buffers each in-flight part in memory, eagerly allocating `(concurrency+1) × partSize`. To keep this bounded, upload settings are decoupled from the S3 URL (`concurrency`/`part_size_mb` now tune restore only), and upload concurrency is chosen per object from its size so the pool stays under a 256 MiB budget — full parallelism for normal caches, stepping down to 1 for multi-TiB objects. The residual floor is S3's 10,000-part limit: a buffering uploader must hold at least one part of `size/10_000`, which only a streaming uploader could avoid.

## Context

Linear - https://linear.app/buildkite/issue/A-1470/migrate-cache-s3-transfers-to-aws-transfermanager

## Changes

- `internal/cache/store/s3.go`: replace `manager.Uploader`/`manager.Downloader` with two `transfermanager.Client`s (upload- and download-tuned) for save and restore.
- Restore: `transfermanager.DownloadObject` with `GetObjectType = GetObjectRanges`.
- Save: `transfermanager.UploadObject`.
- Keep the 412-retry loop (`downloadWithRetry`); update the `objectDownloader` interface to the `DownloadObject` signature.
- Replace the smithy `PartCounter` middleware with a computed part count (`ceil(bytes / partSize)`); remove the now-unused `sync/atomic` and `smithy-go/middleware` imports.
- Decouple upload from the URL: `concurrency`/`part_size_mb` now tune restore only; uploads use fixed defaults (`manager.DefaultUploadConcurrency`/`DefaultUploadPartSize` replaced by local 5 × 5 MiB constants).
- Bound upload memory: `uploadConcurrencyForSize` picks per-object concurrency so the eager part-buffer pool `(concurrency+1) × partSize` stays under a 256 MiB budget; pin `MaxUploadParts` and `MultipartUploadThreshold` for deterministic part sizing.
- Remove the `feature/s3/manager` import and all `//nolint:staticcheck // SA1019` suppressions in the package.
- Tests: update `TestResolveTransferSettings`, `TestDownloadWithRetry`, and the `fakeDownloader` double to the new API; add `TestNewS3Blob` (custom-endpoint + path-style construction) and `TestUploadConcurrencyForSize` (size→concurrency bounding).

## Testing


internal/cache/store/s3.go
```diff
@@ -227,14 +231,20 @@ func NewS3Blob(ctx context.Context, s3url string) (*S3Blob, error) {
...
 
-	// Create the uploader and downloader with their resolved settings
-	uploader := manager.NewUploader(client, func(u *manager.Uploader) { //nolint:staticcheck // SA1019: pending migration to transfermanager
-		u.Concurrency = settings.uploadConcurrency
-		u.PartSize = settings.uploadPartSize
+	// Create the transfer-manager clients with their resolved settings. Uploads
+	// and downloads are tuned differently, so each gets its own client.
+	uploader := transfermanager.New(client, func(o *transfermanager.Options) {
+		o.Concurrency = settings.uploadConcurrency
+		o.PartSizeBytes = settings.uploadPartSize
```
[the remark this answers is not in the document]
ss1909: fixed - I have decoupled the the upload(save) settings from the restore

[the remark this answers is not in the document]
ss1909: Good catch on both. Fixed:

- Upload settings are now fully decoupled from the URL — concurrency/part_size_mb tune restore (download) only, so the 505 GiB case can no longer happen.
- For the size-driven path, upload concurrency is now chosen per object from its size (uploadConcurrencyForSize) to keep the eager part-buffer pool (concurrency+1) × partSize under a 256 MiB budget. Normal caches keep full parallelism; it steps down to concurrency 1 for multi-TiB objects.
- MaxUploadParts and MultipartUploadThreshold are pinned so part sizing is deterministic and the budget math is exact.
- The one thing this can't beat is the S3 10,000-part floor: a multi-TiB object forces parts of size/10_000, and a buffering uploader must hold at least one, so RAM at concurrency 1 is ~2 × size/10_000 (~1 GiB at 5 TiB). Going below that needs a streaming uploader, which only the deprecated manager provides — so we've bounded it as far as the non-deprecated API allows. Added test coverage in uploadConcurrencyForSize.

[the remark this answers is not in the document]
zhming0: If `uploadMemoryBudget` doesn't enforce any limit, I wonder what the point is behind that setting? :g_thinking:

internal/cache/store/s3.go
```diff
@@ -35,6 +33,24 @@ import (
...
+	defaultUploadConcurrency   = 5
+	defaultUploadPartSizeBytes = 5 * 1024 * 1024
+	// defaultUploadMultipartThreshold is pinned rather than inherited from the
+	// SDK so upload behaviour is explicit and stable across SDK versions.
+	defaultUploadMultipartThreshold = 16 * 1024 * 1024
+	// uploadMemoryBudget is the target ceiling for a single upload's part-buffer
+	// pool. It holds for all normal object sizes; multi-TiB objects bottom out at
+	// concurrency 1, where one part buffer (size/uploadMaxParts) is unavoidable
+	// and may exceed this.
+	uploadMemoryBudget = 256 * 1024 * 1024
```
zhming0, line 50: Just a question for my understanding, what was this memory budget before this change?

internal/cache/store/s3.go
```diff
@@ -227,14 +247,26 @@ func NewS3Blob(ctx context.Context, s3url string) (*S3Blob, error) {
...
-	downloader := manager.NewDownloader(client, func(d *manager.Downloader) { //nolint:staticcheck // SA1019: pending migration to transfermanager
-		d.Concurrency = settings.downloadConcurrency
-		d.PartSize = settings.downloadPartSize
+	downloader := transfermanager.New(client, func(o *transfermanager.Options) {
+		o.Concurrency = settings.downloadConcurrency
+		o.PartSizeBytes = settings.downloadPartSize
+		// Use Range-based fan-out rather than partNumber-based. The SDK default
+		// (PART) only fans out for objects uploaded as multipart; Ranges work on any object
+		// regardless of how it was uploaded, so restore parallelism
+		// is determined by our config (Concurrency × PartSizeBytes), not by upload history
```
zhming0, line 269: Can you help me understanding this? What doesn't work with SDK default?

internal/cache/store/s3.go
```diff
@@ -267,21 +300,18 @@ type transferSettings struct {
...
+	downloadPartSize := int64(defaultDownloadPartSizeMB) * 1024 * 1024
 	if opts.Concurrency > 0 {
-		uploadConcurrency = opts.Concurrency
 		downloadConcurrency = opts.Concurrency
 	}
-
-	uploadPartSize := manager.DefaultUploadPartSize
-	downloadPartSize := int64(defaultDownloadPartSizeMB) * 1024 * 1024
 	if opts.PartSizeMB > 0 {
-		uploadPartSize = int64(opts.PartSizeMB) * 1024 * 1024
```
zhming0, line 284: Sorry, can you help me understand again (using your words) why we need to set different concurrency and part size for upload and download? I did read the buildsworth thread, that doesn't quite make sense to my brain :g_thinking:. <br><br>The thing is, our current url scheme allows a generic currency and part size setting, if we were to make this change, our URL params will become misleading.

zhming0 (commented):
I have left some questions mainly for myself to understand this better 🙏

internal/cache/store/s3.go
ss1909: The SDK default here is GetObjectParts, which is PartNumber based. The problem with that is that when there is a only a single part (which is highly likely due to the default used by transferManager which happens to be 16MiB). So PartNumber=1, which is a single part. But fans out only happens if the object was uploaded as multipart. Objects written by a single PutObject (any cache ≤ 16 MiB threshold) report PartsCount=1, and our TTL-refresh CopyObject also rewrites objects as single-part  - so with the default those restore as a single sequential stream, ignoring our concurrency. 

Basically, this would just mean parallelism could get affected and restore could be slower if we used the default.

internal/cache/store/s3.go
ss1909: So there was no explicit "budget" before - manager was inherently bounded as it was streaming rather than buffering (which is waht transferManager does).

internal/cache/store/s3.go
ss1909: save and restore have differing memory behaviours:
- Save buffers each in-flight part fully in RAM
- Restore streams each range through a small (32 KiB) buffer
Save/Upload buffers (memory grows with part size); Restore/download streams (memory doesn't). That mismatch is why this need for different concurrency/part-size handling arises. 

I very much agree with the generic names for concurrency/part_size_mb, are misleading, so I propose we could do one of these:
1. Document them as restore/download only tuning
2. Rename to download_concurrency / download_part_size_mb

internal/cache/store/s3.go
zhming0: This makes sense. Thanks 👍

internal/cache/store/s3.go
zhming0: > Save buffers each in-flight part fully in RAM

I can see that the buffer behave a bit differently, but why does the the configuration need to differ I wonder? I read the other thread it's a bit avoiding excessive memory? But does the upload memory budget setting cap the memory?

ss1909: the reason why the configs need to be different is because the numbers that make restore fast are exactly the ones that make save allocate big chunks (gbs), so a single shared value can't serve both! it's either too small (slow restore) or too large (save OOM). 

Restore streams, so memory is flat regardless of part size/concurrency - the best config is high concurrency + large parts (c32/p32) purely for speed. Save buffers, so memory = (concurrency+1) × partSize - the best config is low concurrency + small parts to stay bounded. More on this here - https://docs.aws.amazon.com/AmazonS3/latest/userguide/qfacts.html and https://github.com/aws/aws-sdk-go-v2/tree/50fd8e78e0780bc66f360bc50d59aa314d164d54/feature/s3/transfermanager 

does that make sense?

internal/cache/store/s3.go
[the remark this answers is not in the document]
ss1909: it doesnt enforce a limit on upload/save but it does on restore.

internal/cache/store/s3.go
zhming0: I see what this is trying to achieve now. But I am not sure if hard coding to 256MB is the right path:

1. We should at least detect the available RAM. 
2. But even then, we don't know how many concurrent upload is happening at the same time. 
3. We don't seem to clearly understand the implication of this knob → e.g. by turning it up and down, are we expecting a speed up/down? It feels to me that we are trying to use this knob to serve as a workaround for a pitfall in AWS transfer manager. <br> 

At this point, I think it'd be fair to pause to consider if transfer manager is mature enough to justify this change.

internal/cache/store/s3.go
```diff
@@ -294,6 +325,36 @@ func resolveTransferSettings(opts *Options) transferSettings {
...
+	// solve (concurrency+2) × partSize <= budget.
+	concurrency := int(uploadMemoryBudget/partSize) - 2
+	if concurrency > maxConcurrency {
+		concurrency = maxConcurrency
+	}
+	if concurrency < 1 {
+		concurrency = 1
+	}
+	return concurrency
+}
```
zhming0, line 356: \[blocking\] I believe the arithmetic is based on some internal logic of transfer function, this isn't ideal because we don't want to be coupled to something that will change anytime.

zhming0 (changes requested):
I think we should reconsider the upgrade. My quick read (I could be wrong), is that the upstream upload manager is quite flawed and we had to build extra finicky harness to prevent its causing OOM issue. <br><br>I am leaning towards not upgrading to transfer manager now (at least for the upload path) after seeing the efforts here. What do you think @ss1909?

ss1909:
Yeah I concur. I reckon we could move this to later, maybe after the next milestone even.

zhming0:
yeah, I agree. Let's move it off the current milestone back to backlog.

PR #4120, opened 2026-07-27
feat(cache): namespace archive entries by anchor via a manifest

ss1909:
## Description

Moves the cache archive format from **v1** (a single flat namespace where the zip entry name is the sole routing key) to **v2** (each configured `target_path` gets its own numbered namespace `_N/`, plus a reserved `.buildkite/cache-manifest.json` that maps each namespace to its **anchor** — a symbol `~` / `/` / `.`, not a resolved path).

**Why:** in v1 every target path shares one flat namespace, so at restore the agent has to *guess* which configured path an entry belongs to via prefix matching. That is collision-prone (two chroots can produce identical entry names) and can't represent absolute paths outside `$HOME` (quickzip can't chroot at `/`) — which blocks A-1533. v2 gives each path its own namespace and records its anchor, so routing is exact, collisions are impossible by construction, and absolute/home/relative paths each restore correctly.

**Key design points / trade-offs:**
- **Anchors are symbols, not resolved paths.** `~` is re-expanded against the restoring `$HOME` (portable); `/` (POSIX) or the drive root `C:\` (Windows) stays pinned; `.` resolves against the job's working directory. This is what makes home/relative caches portable across agents while absolute paths stay put.
- **Absolute paths are pinned — including under `$HOME`.** Only a literal leading `~` is portable (per A-1584). An absolute path is addressed by the backend by its literal string, so it must not silently follow a changed `$HOME`.
- **Windows support via volume-encoded anchors.** The pinned-absolute anchor is the path's volume root, so a drive is retained (`C:\`) rather than lost; POSIX archives are byte-identical to before.
- **quickzip constraint.** quickzip (a vendored dep we can't patch) offers no entry-name prefix and can't chroot at `/`. So each mapping is archived into a temporary quickzip archive and its entries are **raw-copied** (`CreateRaw`/`OpenRaw`, no recompression) into the final zip with the `_N/` prefix — keeping quickzip's compression and byte-determinism (the archive's sha256 is its content-addressed storage name).
- **Clean break.** New agents only read/write v2; there is no v1 fallback. Any archive an agent can't make sense of — an old/foreign format *or* a corrupt blob — **soft-fails to a cache miss and invalidates the entry**, leaving the target paths untouched. A bad cache never blocks a build.

## Context

- Linear — https://linear.app/buildkite/issue/A-1584/cache-archive-format-v2-manifest-based-layout
- Notion — https://app.notion.com/p/buildkite/Cache-Archive-Format-from-V1-V2-3a3b8dbc2c898009b6d5c0e2f04ad1be

## Changes

- **`internal/cache/archive/mappings.go`** — replaced v1's `{Chroot, RelativePath, Relative}` model with `{Namespace, Anchor, ResolvedPath}`. Added `classifyPath` (single source of truth for save classification and restore matching), volume-aware root anchor (`volumeRoot`/`isRootAnchor`), normalized `homeDir`/`workingDir`, `resolveAnchor`, and exported `ResolveConfigPath`. Dropped `chrootPath` and `ResolveHomeDir`.
- **`internal/cache/archive/archive.go`** — added `Manifest`, `ManifestPath`, `ManifestVersion`, `ErrUnrecognizedFormat`, `writeManifest`/`readManifest` (with version *and* anchor validation), and `DetectFormat`. Removed the now-dead `isUnderHome`.
- **`internal/cache/archive/create.go`** — `BuildArchive` rewritten: per-mapping temp-zip + raw-copy merge with `_N/` prefixes, manifest written first, `saveLayout` (chroot-at-target + root-entry retention so empty dirs / dir symlinks / mode survive), and **overlap rejection** on both lexical and canonical (symlink-resolved, case-canonical) paths.
- **`internal/cache/archive/extract.go`** — `ExtractFiles` rewritten: manifest-driven routing, anchor re-resolution, zip-slip containment (`isUnder`, which handles volume roots), longest-match config filtering, and hash-named discard routing for the manifest / unconfigured namespaces (so a hostile entry name can't escape the throwaway dir on any OS).
- **`internal/cache/restore.go`** — pre-clean `DetectFormat` gate: **any** unreadable archive (unrecognized format *or* corrupt blob) soft-fails to a cache miss and invalidates the stale entry, dirs untouched. Cleanup resolves paths via `ResolveConfigPath` (same classifier as save), and `cleanPath` guards root/cwd/home by **file identity** (`os.SameFile`) rather than string, and handles file (not just directory) targets.
- **Tests** — `mappings_test.go`/`create_test.go` updated to v2 names + a determinism check; all v2 archive tests consolidated in `archive_test.go` (round-trips for home/absolute/relative/file/empty-dir/symlink, portability, pinning, spelling-variant matching, overlap rejection incl. symlink/lexical cases, discard containment, `isUnder`). Integration tests cover the full client path: regular-file round trip, and miss-and-invalidate for both unrecognized-format and corrupt archives.

## Testing

## Disclosures / Credits

Implemented in consultation with Claude.


internal/cache/restore.go
```diff
@@ -213,6 +213,31 @@ func (c *client) Restore(ctx context.Context, cacheID string) (RestoreResult, er
...
+			result.CacheRestored = false
+			result.TotalDuration = time.Since(startTime)
+			span.SetAttributes(
+				attribute.Bool("cache.hit", false),
+				attribute.Bool("cache.restored", false),
+				attribute.Bool("cache.unrecognized_format", true),
+			)
+			span.SetStatus(codes.Ok, "cache miss (unrecognized format)")
+			c.callProgress(cacheID, "complete", "Cache miss (unrecognized archive format)", 0, 0)
+			return result, nil
```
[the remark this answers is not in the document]
ss1909: Fixed - an unreadable archive is a miss and the entry is expired, so the next save re-uploads a v2 archive instead of being blocked

internal/cache/archive/mappings.go
```diff
@@ -7,70 +7,80 @@ import (
...
 
-		chroot, err := chrootPath(mapping.ResolvedPath)
-		if err != nil {
-			return nil, fmt.Errorf("failed to get chroot path: %w", err)
-		}
+// classifyPath determines a configured path's anchor and the absolute path it
+// resolves to in the current environment.
+func classifyPath(path, home, cwd string) (anchor, resolved string) {
+	switch {
+	case path == "~" || strings.HasPrefix(path, "~/"):
```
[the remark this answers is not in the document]
ss1909: Fixed - both restore and save resolve ~ to the real home

internal/cache/archive/create.go
```diff
@@ -110,3 +122,117 @@ func BuildArchive(ctx context.Context, paths []string, key string) (*ArchiveInfo
...
+
+	reader, err := zip.NewReader(tmp, stat.Size())
+	if err != nil {
+		return 0, 0, fmt.Errorf("failed to read namespace archive: %w", err)
+	}
+
+	for _, f := range reader.File {
+		// Skip the chroot root entry (quickzip names it "."); its directory is
+		// recreated on extract when its children are written.
+		if path.Clean(f.Name) == "." {
```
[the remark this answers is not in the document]
ss1909: Fixed - chroot is not skipped anymore

internal/cache/archive/create.go
```diff
@@ -110,3 +122,117 @@ func BuildArchive(ctx context.Context, paths []string, key string) (*ArchiveInfo
...
+func saveLayout(m Mapping, isDir bool, home, cwd string) (chroot, prefix string, err error) {
+	switch m.Anchor {
+	case AnchorHome:
+		return home, m.Namespace + "/", nil
+	case AnchorCWD:
+		return cwd, m.Namespace + "/", nil
+	case AnchorRoot:
+		// A single "/" anchor cannot represent Windows volumes, so absolute-path anchoring is POSIX-only.
+		// Home and cwd-anchored paths still work on Windows.
+		if runtime.GOOS == "windows" {
```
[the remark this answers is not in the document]
ss1909: Fixed - classified abs-under-$HOME as the home anchor

internal/cache/archive/extract.go
```diff
@@ -36,37 +38,102 @@ func ListArchive(ctx context.Context, zipFile *os.File, zipFileLen int64) ([]str
...
+				best = i
+			}
+		}
+		if best == -1 {
+			return discard(file.Name), nil
 		}
+		foundPaths[paths[best]] = true
 
-		return "", fmt.Errorf("failed to find path mapping for: %s", file.Name)
+		return dest, nil
```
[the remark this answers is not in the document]
ss1909: fixed

ss1909:
@buildsworth-bk review

internal/cache/archive/extract.go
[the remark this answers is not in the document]
ss1909: fixed - covering aliases as well

ss1909:
@buildsworth-bk review

ss1909:
@buildsworth-bk review

internal/cache/archive/archive.go
```diff
@@ -27,33 +46,76 @@ type ArchiveInfo struct {
...
+}
+
+// DetectFormat reports whether an archive is a readable archive. It returns
+// ErrUnrecognizedFormat for unknown-version archives.
+func DetectFormat(zipFile *os.File, zipFileLen int64) error {
+	reader, err := zip.NewReader(zipFile, zipFileLen)
+	if err != nil {
+		return fmt.Errorf("failed to open zip reader: %w", err)
+	}
+	_, err = readManifest(reader)
```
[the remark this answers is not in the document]
ss1909: fixed - verify the downloaded blob's digest before cleanup

internal/cache/archive/create.go
```diff
@@ -33,61 +38,91 @@ func BuildArchive(ctx context.Context, paths []string, key string) (*ArchiveInfo
...
+
+	// Reject overlapping targets: nested/equal paths archive shared files twice
+	// and collide on the same restore destination (a race, or a symlink-vs-dir
+	// conflict). Check lexical and canonical paths — lexical catches nesting at
+	// the restore destination ("/cache" + "/cache/link/sub"), canonical catches
+	// symlink aliases and Windows case variants. EvalSymlinks falls back to
+	// lexical on error (plans only hold existing paths).
+	canonical := make([]string, len(plans))
+	for i, p := range plans {
+		r, err := filepath.EvalSymlinks(p.mapping.ResolvedPath)
```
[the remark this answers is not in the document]
ss1909: fixed - resolve symlinked parents but preserve a final symlink

internal/cache/archive/extract.go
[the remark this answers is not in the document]
ss1909: made the check more file system aware - Fixed

internal/cache/restore.go
```diff
@@ -415,26 +500,32 @@ func cleanPath(ctx context.Context, dir string) error {
...
-	if home, err := os.UserHomeDir(); err == nil {
-		if clean == filepath.Clean(home) {
-			return fmt.Errorf("cleanPath: refusing to remove home directory %q", clean)
+	// Identity guards: refuse if the target resolves — through symlinked parents
+	// or case-insensitive spellings — to a protected directory (working dir,
+	// home, or filesystem root), even when the lexical spelling differs.
+	info, statErr := os.Stat(clean)
+	if statErr == nil {
+		for _, p := range protectedDirs() {
+			if pinfo, err := os.Stat(p); err == nil && os.SameFile(info, pinfo) {
```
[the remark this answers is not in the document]
ss1909: fixed

internal/cache/restore.go
```diff
@@ -415,26 +500,34 @@ func cleanPath(ctx context.Context, dir string) error {
...
-	// Refuse to delete home directory
-	if home, err := os.UserHomeDir(); err == nil {
-		if clean == filepath.Clean(home) {
-			return fmt.Errorf("cleanPath: refusing to remove home directory %q", clean)
+	// Identity guards: refuse if the target is, or is an ancestor of, a
+	// protected directory (working dir, home, or filesystem root) — removing it
+	// would recursively delete that directory too. Comparison is by file
+	// identity (walking each protected dir up to the root), so symlinked parents
+	// and case-insensitive spellings are handled, not just an exact match.
+	info, statErr := os.Stat(clean)
```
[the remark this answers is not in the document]
ss1909: fixed - using Lstat instead

internal/cache/archive/extract.go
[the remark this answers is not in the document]
ss1909: yes we can

internal/cache/archive/archive.go
```diff
@@ -27,33 +46,75 @@ type ArchiveInfo struct {
...
-// isUnderHome checks if the given path is under the user's home directory.
-// It first gets the absolute path of the given path, then gets the user's
-// home directory, and finally checks if the absolute path starts with the
-// home directory path.
-func isUnderHome(path string) (bool, error) {
-	if path == "" {
-		return false, fmt.Errorf("path is empty")
+// writeManifest converts manifest into json, creates a
+// new zip/archive file and writes the manifest to that file.
+func writeManifest(zw *zip.Writer, manifest Manifest) error {
```
zhming0: It might be worth to open a archive/manifest.go to contain all manifest related logic.

internal/cache/restore.go
```diff
@@ -367,6 +431,49 @@ func (c *client) downloadCache(ctx context.Context, retrieveResp api.CacheEntryR
...
+
+	h := sha256.New()
+	if _, err := io.Copy(h, f); err != nil {
+		return fmt.Errorf("failed to hash archive file: %w", err)
+	}
+	if got := hex.EncodeToString(h.Sum(nil)); got != expected {
+		return fmt.Errorf("archive digest mismatch: got %s, want %s", got, expected)
+	}
+	return nil
+}
```
zhming0: These seem to be for another task. Could the clanker mixing things up?

internal/cache/restore.go
```diff
@@ -213,14 +216,75 @@ func (c *client) Restore(ctx context.Context, cacheID string) (RestoreResult, er
...
+	if len(retrieveResp.Blobs) > 0 && retrieveResp.Blobs[0].Digest.Algorithm == "sha256" {
+		expectedDigest = retrieveResp.Blobs[0].Digest.Value
+	}
+	if err := verifyDownloadedArchive(archiveFile, transferInfo.BytesTransferred, expectedDigest); err != nil {
+		// A bad cache must never block a build: any unreadable/corrupt archive
+		// degrades to a miss and invalidates the entry (targets untouched — this
+		// runs before cleaning).
+		unrecognized := errors.Is(err, archive.ErrUnrecognizedFormat)
+		slog.Warn("cache archive unreadable, treating as miss and invalidating entry",
+			"cache_id", cacheID, "unrecognized_format", unrecognized, "err", err)
```
zhming0: I am not sure if  "unreadable" is accurate, I think there are a number of error case in it. ,

internal/cache/restore.go
```diff
@@ -213,14 +216,75 @@ func (c *client) Restore(ctx context.Context, cacheID string) (RestoreResult, er
...
+		result.TotalDuration = time.Since(startTime)
+		span.SetAttributes(
+			attribute.Bool("cache.hit", false),
+			attribute.Bool("cache.restored", false),
+			attribute.Bool("cache.targets_overlap", true),
+		)
+		span.SetStatus(codes.Ok, "cache miss (targets overlap at restore)")
+		c.callProgress(cacheID, "complete", "Cache miss (targets overlap at restore)", 0, 0)
+		return result, nil
+	}
```
zhming0, line 268: \[blocking\] I think these cases are fatal for a cache entry, we should do cache busting.

But as you can see these are secondary to the goal of this PR, given the size and gravity of this PR, would you want to split validation work into another PR? <br><br>Secondly, a `archive.Validate` to encapsulate all validation logics might be helpful?

zhming0 (changes requested):
Still reading :green_circle: , I left some structural comments first.

internal/cache/restore.go
ss1909: - I want to understand what you mean by this - why do you think its fatal? To me this reads like - when targets overlap at restore, it doesn't make the archived data incorrect/wrong. So instead of busting, we should be sodt failing because at restore time, its a particular machine's situation makes it unusable, right? 

> [blocking] I think these cases are fatal for a cache entry, we should do cache busting.

- I think because I've already done the work, here, I'll just do archive.Validate to encapsulate all the validation stuff

internal/cache/restore.go
ss1909: yes! :oldManYells:

ss1909: I've removed this

internal/cache/restore.go
ss1909: I agree, chnaged it

internal/cache/archive/archive.go
ss1909: done

internal/cache/restore.go
zhming0: That's a fair point. I guess it ***can*** be fatal when the path overlapping isn't as a result of local OS. e.g. if we have two `~/foo`, or `~/foo/../foo`, then I suppose it's fatal? <br><br>----<br><br>Back to whether we split the PR:<br>👆 The nuance above is partly why I'd suggest a split, so we can we have one focused review on each areas. It's easier to review two 400 LoC PRs than to review a single 800 LoC PR. (these day clankers can handle the split pretty well)<br><br>But I can see that you might find splitting a PR annoying (very understandable!), so it's okay if you wish to continue as it is.

internal/cache/archive/validate.go
```diff
@@ -0,0 +1,36 @@
...
+
+	if expectedDigest != "" {
+		h := sha256.New()
+		if _, err := io.Copy(h, f); err != nil {
+			return fmt.Errorf("failed to hash archive file: %w", err)
+		}
+		if got := hex.EncodeToString(h.Sum(nil)); got != expectedDigest {
+			return fmt.Errorf("archive digest mismatch: got %s, want %s", got, expectedDigest)
+		}
+	}
```
zhming0: Hmmm were you going to remove this?  <br><br>There is a fair bit of nuances in digest checking so it's a separated task, we'd want to consider stream approach to avoid reading the file twice.

internal/cache/restore.go
ss1909: > That's a fair point. I guess it can be fatal when the path overlapping isn't as a result of local OS. e.g. if we have two ~/foo, or ~/foo/../foo, then I suppose it's fatal? 

Yes, this is fair point. But, in this case, the save will reject it so it never really gets to the restore. 

> But I can see that you might find splitting a PR annoying (very understandable!), so it's okay if you wish to continue as it is.

🙇‍♀️ 🙇‍♀️ 🙇‍♀️ 🙇‍♀️

internal/cache/archive/validate.go
ss1909: oh yes, thanks! forgot about this one. 🙇‍♀️

internal/cache/archive/manifest.go
```diff
@@ -0,0 +1,99 @@
...
+// DetectFormat reports whether an archive is a readable archive. It returns
+// ErrUnrecognizedFormat for unknown-version archives.
+func DetectFormat(zipFile *os.File, zipFileLen int64) error {
+	reader, err := zip.NewReader(zipFile, zipFileLen)
+	if err != nil {
+		return fmt.Errorf("failed to open zip reader: %w", err)
+	}
+	_, err = readManifest(reader)
+	return err
+}
```
zhming0: \[nit\] the name DetectFormat imply that this function will return a format. Shall we rename this?

internal/cache/archive/mappings.go
```diff
@@ -7,86 +7,147 @@ import (
...
+)
+
+// Mapping describes how one configured target path maps into an archive: a
+// numbered namespace ("_0", "_1", ...) whose entries are relative to the anchor.
 type Mapping struct {
-	Path         string
+	// Path is the target path exactly as configured.
+	Path string
+	// Namespace is the archive prefix that isolates this mapping's entries,
+	Namespace string
```
zhming0: \[nit\] is Prefix a better name?

internal/cache/archive/mappings.go
```diff
@@ -7,86 +7,147 @@ import (
...
+	// working directory.
+	AnchorCWD = "."
+)
+
+// Mapping describes how one configured target path maps into an archive: a
+// numbered namespace ("_0", "_1", ...) whose entries are relative to the anchor.
 type Mapping struct {
-	Path         string
+	// Path is the target path exactly as configured.
+	Path string
```
zhming0, line 28: \[Question\] Should this be the exact target path? or the path striped off the anchor? I imagine the purpose Mapping object to take anchor and path apart?

internal/cache/archive/mappings.go
```diff
@@ -7,86 +7,147 @@ import (
...
+	}
 
-		mapping.Chroot = chroot
+	return mappings, nil
+}
 
-		pathMappings = append(pathMappings, mapping)
+// classifyPath determines a configured path's anchor and the absolute path it
+// resolves to in the current environment.
+func classifyPath(path, home, cwd string) (anchor, resolved string) {
```
zhming0: `classifyAndResolvePath`?

internal/cache/archive/create.go
```diff
@@ -110,3 +131,211 @@ func BuildArchive(ctx context.Context, paths []string, key string) (*ArchiveInfo
...
+		}
+		rel, err := filepath.Rel(m.Anchor, m.ResolvedPath)
+		if err != nil {
+			return "", "", fmt.Errorf("failed to compute root-relative path for %q: %w", m.ResolvedPath, err)
+		}
+		return m.ResolvedPath, m.Namespace + "/" + filepath.ToSlash(rel) + "/", nil
+	default:
+		return "", "", fmt.Errorf("unknown anchor %q", m.Anchor)
+	}
+}
```
zhming0: This function dilute the value of the Mapping abstraction. It looks me that we are calculating two levers of Mappings, a.k.a it become difficult to read this bit of code without reading mapping and wise versa. This is usually a sign of abstraction leak.  Would you want to consider consolidate zip archive mapping logic into Mapping? Hence reconsider the structure of the Mapping struct per my comments above and refactor here?

zhming0 (changes requested):
My second round of reviews find the Mapping abstraction can use some polishing. We should be able to make the overall separation of concern clearer.

internal/cache/archive/create.go
```diff
@@ -110,3 +131,211 @@ func BuildArchive(ctx context.Context, paths []string, key string) (*ArchiveInfo
...
+		// quickzip can't chroot at a bare root, so chroot at the target itself;
+		// archiveMapping renames the "." entry onto the namespace.
+		if m.ResolvedPath == m.Anchor {
+			return "", "", fmt.Errorf("cannot archive the volume/filesystem root %q", m.ResolvedPath)
+		}
+		rel, err := filepath.Rel(m.Anchor, m.ResolvedPath)
+		if err != nil {
+			return "", "", fmt.Errorf("failed to compute root-relative path for %q: %w", m.ResolvedPath, err)
+		}
+		return m.ResolvedPath, m.Namespace + "/" + filepath.ToSlash(rel) + "/", nil
```
zhming0: @ss1909 FYI I patched quickzip so it can take `/` as chroot now, we can probably simplify here  [https://github.com/wolfeidau/quickzip/releases/tag/v1.0.3](<https://github.com/wolfeidau/quickzip/releases/tag/v1.0.3>)

internal/cache/archive/create.go
ss1909: I have done a bunch of refactoring to address this: 
- folded it all onto Mapping: saveLayout is gone, replaced by an env-free Mapping.archiveLayout(). 
- The Mapping struct now takes apart anchor and path 
- Plus, thanks to your quickzip /-chroot patch, the whole root-anchor special case collapsed - every anchor just chroots at its base now.

internal/cache/archive/mappings.go
ss1909: Its the exact target path. We only use it for displaying logs and errors. Happy to rename it to ConfiguredPath if that reads better?

internal/cache/archive/mappings.go
ss1909: I would actually like to keep this "Namespace" for two reasons:
1. Prefix and anchor could allude to the same thing 
2. I actually think Namespace is the more accurate name as it conveys sort of a container that groups many entries under it, where "prefix" only describes the leading string

internal/cache/archive/mappings.go
```diff
@@ -7,86 +7,162 @@ import (
...
+
+// archiveLayout returns the directory quickzip archives from (chroot) and the
+// entry-name prefix prepended so each stored entry is "<namespace>/..". Every
+// anchor chroots at its base (home, cwd, or the volume root), so quickzip names
+// entries relative to that base and the prefix only adds the namespace.
+func (m Mapping) archiveLayout() (chroot, prefix string, err error) {
+	// Refuse to archive a base directory itself
+	if isRootAnchor(m.Anchor) && m.resolved == m.base {
+		return "", "", fmt.Errorf("cannot archive the volume/filesystem root %q", m.resolved)
+	}
```
zhming0: \[nit, not blocking\] I wonder if this is the right place to do validation.

internal/cache/archive/create.go
```diff
@@ -33,61 +37,68 @@ func BuildArchive(ctx context.Context, paths []string, key string) (*ArchiveInfo
...
+		manifest.Mappings[mapping.Namespace] = mapping.Anchor
+	}
+
+	// Reject overlapping targets: they archive shared files twice and collide on
+	// one restore destination. Restore re-checks the re-resolved paths.
+	resolved := make([]string, len(plans))
+	for i, p := range plans {
+		resolved[i] = p.mapping.ResolvedPath()
+	}
+	if i, j, ok := OverlappingPaths(resolved); ok {
```
zhming0, line 81: \[not blocking\] should OverlappingPaths become a method of mapping directly? or some place where it can be more apparent that it's shared by create and extract at the same time?

internal/cache/archive/create.go
```diff
@@ -110,3 +121,187 @@ func BuildArchive(ctx context.Context, paths []string, key string) (*ArchiveInfo
...
+
+		rc, err := f.OpenRaw()
+		if err != nil {
+			return 0, 0, fmt.Errorf("failed to open raw entry %q: %w", f.Name, err)
+		}
+
+		if _, err := io.Copy(w, rc); err != nil {
+			return 0, 0, fmt.Errorf("failed to copy entry %q: %w", f.Name, err)
+		}
+	}
```
zhming0, line 208: \[not blocking\] @ss1909 I figure we could patch quickzip upstream to make it support prefix. What do you think? We can probably add a task?

internal/cache/archive/extract.go
```diff
@@ -36,37 +38,102 @@ func ListArchive(ctx context.Context, zipFile *os.File, zipFileLen int64) ([]str
...
+				continue
 			}
+			if best == -1 || len(cfg) > len(configPaths[best]) {
+				best = i
+			}
+		}
+		if best == -1 {
+			return discard(file.Name), nil
 		}
+		foundPaths[paths[best]] = true
```
zhming0: \[Question\] Can you help me understand this better? What case is this trying to prevent?

zhming0 (approved):
I no longer see a blocker. Let's give it a go 👍 💚

internal/cache/archive/mappings.go
ss1909: agree - I've moved it to PathsToMappings, so this archiveLayout stays, as the name suggests, purely about the archive layout.

internal/cache/archive/create.go
ss1909: agree on this too - I've moved it and made it explicit that its shared by save and restore 🙏

internal/cache/archive/create.go
ss1909: 🙌 yess. I can create a task

internal/cache/archive/extract.go
ss1909: this loops helps where there's multiple paths, which could overlap. Taking the example in the comments - where there is ~/.cache" and "~/.cache/sub", we want to resolve to the longest (ie the more specific one). But I'm second gueesing this because of the way foundPaths is used later on here https://github.com/buildkite/agent/blob/33caa8aaad01f940bd8a87a0c897a993d144493c/internal/cache/archive/extract.go#L75, it really doesn't seem to matter. So I'm going to change this and return the first match instaed of doing the longest string match.

internal/cache/archive/mappings.go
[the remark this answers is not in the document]
ss1909: fixed - this will now error

PR #4121, opened 2026-07-27
A-1591: Add OTLP job log export

zhming0:
## Description

Continuation of #3979 (thanks @catkins). That branch is squashed into the first commit here; see the original PR for the full design discussion and verification notes. The second commit reworks the correlation design, per the [which span owns a log line](https://www.notion.so/buildkite/3aab8dbc2c898051a08ec6bd65c348d9) design note.

Adds an opt-in job log OTLP sink. When `--job-logs-otlp` / `BUILDKITE_JOB_LOGS_OTLP` is enabled, the bootstrap emits job output as OpenTelemetry log records using the existing OTLP exporter environment configuration. This is an additional sink: the normal Buildkite job log stream is unchanged.

## What each record carries

* Body: one job log line, with the same `[REDACTED]` markers as the customer-facing job log.
* Trace correlation: the trace ID and the span ID of the nearest enclosing phase or hook span at the time the line is emitted, falling back to the root job span. Records within a phase deliberately carry the phase or hook span, not child op spans such as `git.clone`. With tracing disabled, records are emitted without correlation.
* Base attributes: org, pipeline, branch, queue, agent, build and job identifiers.
* Timestamp: the arrival of the line's first byte, matching the start-of-line semantics of the job log's ANSI timestamper.

We dropped the per-record `buildkite.phase` and `buildkite.hook.*` attributes from the earlier iteration. The agent only defines `plugin`, `checkout` and `command` as phases, and inventing phase names for other stages risks exporting wrong metadata. Richer attribute propagation can follow in a dedicated change.

## How it differs from the original PR

* Child-process output and bootstrap narration (section headers, prompts, comments, warnings) are mirrored into OTLP by two tees downstream of the existing job log redactors. Records inherit the job log's redaction by construction, including secrets split across writes and secrets added mid-job via the Job API, so the parallel OTLP redaction pipeline (needle syncing, cross-command partial-match state, flush-boundary alignment) is deleted.
* The shell `OutputInterceptor` plumbing is removed entirely.
* Correlation moved from per-command context plumbing to a single emit context that the executor updates at span boundaries (`defer e.otlpLogSpan(ctx)()` right after starting a phase or hook span).

Also addresses review feedback from #3979: the `job-logs-otlp` bootstrap flag is no longer hidden ([comment](https://github.com/buildkite/agent/pull/3979#discussion_r3619308256)), and `teeWriter` lives in its own file ([comment](https://github.com/buildkite/agent/pull/3979#discussion_r3619280325)).

## Testing

- [x] Tests have run locally (with `go test ./internal/... ./clicommand/ ./agent/` and `go test -race ./internal/job/`). Buildkite employees may check this if the pipeline has run automatically.
- [x] `golangci-lint run` clean on touched packages


internal/job/executor.go
```diff
@@ -74,10 +74,26 @@ type Executor struct {
...
+	// stdoutTee carries the redacted child-process output (hook/command stdout
+	// and stderr) to stdout and, when OTLP job logging is enabled, mirrors it
+	// into the OTLP exporter.
+	stdoutTee *teeWriter
+
+	// loggerTee carries the redacted shell logger output (section headers,
+	// prompts, comments, warnings) to stderr and, when OTLP job logging is
+	// enabled, mirrors it into the OTLP exporter so the exported records match
+	// the customer-facing Buildkite job log.
+	loggerTee *teeWriter
```
DrJosh9000: Since the main difference between `stdoutTee` and `loggerTee` is the `primary`, perhaps `loggerTee` would be better named `stderrTee` ?

internal/job/executor.go
```diff
@@ -738,6 +777,10 @@ func (e *Executor) addOutputRedactors() {
 	}
 }
 
+func (e *Executor) flushOutputRedactors() error {
+	return e.redactors.Flush()
+}
```
DrJosh9000: Since there's no other code in here, we can go back to calling `e.redactors.Flush()` directly, simplifying away this new method.

internal/job/otlp_job_logger.go
```diff
@@ -0,0 +1,345 @@
...
+// setSpanContext at span boundaries (phases and hooks), and every record
+// emitted until the next boundary carries that span's context (when tracing
+// is enabled).
+type otlpJobLogger struct {
+	log      otellog.Logger
+	provider *sdklog.LoggerProvider
+
+	baseAttrs []otellog.KeyValue
+
+	// mu guards the current emit context.
```
DrJosh9000, line 42: Add a comment explaining why the context is being stored in the struct. (This is necessary because as a general rule, contexts do not belong stored in structs: https://go.dev/blog/context-and-structs).

zhming0: added a comment 👍🏿

internal/job/otlp_job_logger.go
```diff
@@ -0,0 +1,351 @@
...
+	)
+
+	return newOTLPJobLoggerWithLogger(provider.Logger(
+		"buildkite-agent",
+		otellog.WithInstrumentationVersion(version.Version()),
+		otellog.WithSchemaURL(semconv.SchemaURL),
+	), provider, otlpJobAttributes(e)), nil
+}
+
+func newOTLPJobLoggerWithLogger(log otellog.Logger, provider *sdklog.LoggerProvider, baseAttrs []otellog.KeyValue) *otlpJobLogger {
```
DrJosh9000: Why does `newOTLPJobLogger` take a context arg, but this function does not?

DrJosh9000 (approved):
Happy to ship it and see how it goes

internal/job/otlp_job_logger.go
zhming0: It's because this function isn't context-aware. The function exist to reduce a bit of duplicated code. Maybe it's not worth the indirection 🤔

DrJosh9000: It stores `context.Background()` into the ctx field. That could be passed in?

zhming0: True, done. I also added a comment to explain that the passed in ctx isn't about lifetime control.

PR #4142, opened 2026-07-30
A-1547: Add --job-context-dir to configure job coordination file locations

zhming0:
Adds a `--job-context-dir` flag (`BUILDKITE_JOB_CONTEXT_DIR`) to `agent start` and `kubernetes-bootstrap` to configure the directory for job coordination files. Based on #4090 by @aeijdenberg-canva, thanks Adam for the contribution. Tracked as [A-1547](https://linear.app/buildkite/issue/A-1547).

Closes #4090.

## Context

The agent and the processes running a job coordinate through a small set of files: the job env snapshot (`BUILDKITE_ENV_FILE` / `BUILDKITE_ENV_JSON_FILE`), the job timeout marker (`BUILDKITE_AGENT_JOB_TIMEOUT_FILE`), and, in Kubernetes mode, the socket that `kubernetes-bootstrap` connects to (`buildkite.sock`). Their directory was hardcoded: `/workspace` in Kubernetes mode, the system temp dir otherwise.

## Who this is for

* **agent-stack-k8s users**: no change. The stack manages the shared `/workspace` volume that Kubernetes mode defaults to; leave the flag unset.
* **Self-managed `--kubernetes-exec` orchestration**: set `BUILDKITE_JOB_CONTEXT_DIR` on every container to the path where the shared volume is mounted in that container. This removes the last hardcoded `/workspace` in the agent binary, which can collide with images that claim that path.
* **Ordinary self-hosted agents**: rarely needed. It relocates the env and timeout files out of the system temp dir (previously only movable via `TMPDIR`).

Defaults are unchanged in all modes.

## Changes

Two commits:

1. The original commit from #4090, rebased onto main. Conflicts with the checkout override work in `env/protected.go` and the container exit status change in `kubernetes/runner.go` were resolved during the rebase.
2. Follow-ups:
   * `kubernetes-bootstrap` now rebases `BUILDKITE_ENV_FILE`, `BUILDKITE_ENV_JSON_FILE` and `BUILDKITE_AGENT_JOB_TIMEOUT_FILE` onto its own context directory. The agent sends these as absolute paths computed in its own mount namespace, so a container mounting the shared volume at a different path would connect to the socket fine and then fail on unreachable env file paths.
   * Reuse the `contextDir` computed in `NewJobRunner` for the socket path, and rename the `tempDir` parameters to `contextDir` since the directory is no longer necessarily the system temp dir.
   * Document the intended audience in the flag help, and correct the `existingEnvPriority` comment (agent-stack-k8s does not set `BUILDKITE_JOB_CONTEXT_DIR`).

Note on scope: mounting the shared volume at the same path in every container remains the supported configuration, since `BUILDKITE_BUILD_PATH` also crosses containers as an absolute path. The env file rebasing is defence in depth and a no-op when paths match.

## Testing

* New unit test `TestRebaseJobContextPaths` covers differing mount paths, same path no-op, and absent vars
* Tests have run locally (with `go test ./...`)
* Code is formatted (with `go tool gofumpt -extra -w .`) and `golangci-lint run` reports no issues on the touched packages

### DIFF
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -90,6 +90,14 @@ type JobRunnerConfig struct {
 	// It's useful to be configured in situations like huge container image cold download.
 	KubernetesContainerStartTimeout time.Duration
 
+	// JobContextDir is the directory for files the agent uses to coordinate
+	// with the processes running the job: the job env files, the job timeout
+	// marker file, and, in Kubernetes mode, the coordination socket. If
+	// empty, it defaults to kubernetes.DefaultContextDir in Kubernetes mode,
+	// or the system temporary directory otherwise. In Kubernetes mode it must
+	// be on a volume shared by all containers in the pod.
+	JobContextDir string
+
 	// Stdout of the parent agent process. Used for job log stdout writing arg, for simpler containerized log collection.
 	AgentStdout io.Writer
 }
@@ -212,7 +220,9 @@ func NewJobRunner(ctx context.Context, l logger.Logger, apiClient *api.Client, c
 		},
 	)
 
-	r.envShellFile, r.envJSONFile, err = createJobEnvFiles(r.agentLogger, r.conf.Job.ID, conf.KubernetesExec)
+	contextDir := jobContextDir(conf)
+
+	r.envShellFile, r.envJSONFile, err = createJobEnvFiles(r.agentLogger, r.conf.Job.ID, contextDir)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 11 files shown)


agent/job_runner.go
```diff
@@ -980,28 +991,37 @@ func (r *JobRunner) onUploadHeaderTime(ctx context.Context, cursor, total int, t
...
+	}
+	if conf.KubernetesExec {
+		return kubernetes.DefaultContextDir
 	}
+	return os.TempDir()
+}
 
-	// tempDir is not guaranteed to exist
-	if _, err := os.Stat(tempDir); os.IsNotExist(err) {
+func createJobEnvFiles(l logger.Logger, jobID, contextDir string) (shellFile, jsonFile *os.File, err error) {
```
jamiemonserrate, line 1009: Nit: because createJobEnvFiles creates an explicit context directory when absent, forgetting the shared-volume mount results in a directory and socket being created only in the agent container.

It just means users will take more time to debug a misconfiguration. Maybe worth adding some explicit logging?

jamiemonserrate (approved):
LGTM! 🚀

agent/job_runner.go
zhming0: I think this behavior is pre-existing. If there is a misconfiguration the agent will fail loudly. 

If are to add logging, I struggle to find wording that adds value, because creating the directory is not wrong in itself (it is normal outside Kubernetes mode). Something like "Creating job context directory because it does not exist" would log on healthy runs too, and since the job context directory is not a prominent concept, I suspect it would be more distracting than helpful.

PR #4147, opened 2026-07-31
fix(cache): make expire scope aware (part 2/2)

ss1909:
## Description

On a scoped cache registry (policy with branch/build/pipeline save scopes), entries live at a distinct scoped address, so that expire request can never match the entry the agent actually restored. The invalidation silently no-ops, the dangling entry keeps being served, and every subsequent restore re-downloads a missing blob — an unrecoverable loop for that key until TTL.

This PR completes the fix by echoing the matched entry's scopes through the round-trip: the paired server change returns the matched entry's scopes on retrieve, and the agent now sends them back on expire so the backend can address (and delete) the exact scoped entry.

## Context

PR on the bk/bk repo: https://github.com/buildkite/buildkite/pull/31863
Linear: https://linear.app/buildkite/issue/A-1625/make-expire-scope-aware

## Changes

- api.CacheEntryRetrieveResp gains a Scopes map[string]string field (json:"scopes,omitempty") — the matched entry's resolved scope labels, nil when unscoped.
- api.CacheEntryExpireReq gains the same Scopes field (omitempty, so unscoped expire requests are byte-identical on the wire to today).
- invalidateStaleEntry copies retrieveResp.Scopes into the expire request, so the missing-blob self-heal addresses the exact scoped entry that was restored.
- Tests: a focused invalidateStaleEntry test asserting the scopes are echoed into the expire request; the API client wire test asserts scopes round-trips over HTTP. No CLI/argument changes.

## Testing

## Disclosures / Credits

I used claude to write the code

### DIFF
--- a/api/cache.go
+++ b/api/cache.go
@@ -79,22 +79,24 @@ type CacheEntryRetrieveReq struct {
 
 // CacheEntryRetrieveResp describes the cache entry to download.
 type CacheEntryRetrieveResp struct {
-	TargetPaths          []string       `json:"target_paths"`
-	CacheKey             []CacheKeyPart `json:"cache_key"`
-	Blobs                []CacheBlob    `json:"blobs"`
-	ExpiresAt            time.Time      `json:"expires_at"`
-	Store                string         `json:"store"`
-	Fallback             bool           `json:"fallback"`
-	Multipart            bool           `json:"multipart"`
-	DownloadInstructions []string       `json:"download_instructions"`
-	Message              string         `json:"message"`
+	TargetPaths          []string          `json:"target_paths"`
+	CacheKey             []CacheKeyPart    `json:"cache_key"`
+	Blobs                []CacheBlob       `json:"blobs"`
+	ExpiresAt            time.Time         `json:"expires_at"`
+	Store                string            `json:"store"`
+	Fallback             bool              `json:"fallback"`
+	Multipart            bool              `json:"multipart"`
+	DownloadInstructions []string          `json:"download_instructions"`
+	Message              string            `json:"message"`
+	Scopes               map[string]string `json:"scopes,omitempty"`
 }
 
 // CacheEntryExpireReq is the request body for invalidating a cache entry.
 // The address should be the resolved entry, as echoed by the retrieve response.
 type CacheEntryExpireReq struct {
-	TargetPaths []string       `json:"target_paths"`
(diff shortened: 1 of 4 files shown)


internal/cache/restore_test.go
```diff
@@ -145,3 +148,32 @@ func TestCleanPathWindowsDriveRoot(t *testing.T) {
...
+	ctx := t.Context()
+
+	cacheClient, _, _ := setupTestCache(t, "local_file")
+	mockClient := cacheClient.api.(*mockAPIClient)
+
+	scopes := map[string]string{"branch": "main", "pipeline": "app"}
+	retrieveResp := api.CacheEntryRetrieveResp{
+		TargetPaths: []string{"node_modules"},
+		CacheKey:    []api.CacheKeyPart{{Value: "v1-key"}},
+		Scopes:      scopes,
```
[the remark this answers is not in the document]
ss1909: added a new test to cover this

api/cache.go
```diff
@@ -88,13 +88,19 @@ type CacheEntryRetrieveResp struct {
 	Multipart            bool           `json:"multipart"`
 	DownloadInstructions []string       `json:"download_instructions"`
 	Message              string         `json:"message"`
+	// Scopes are the matched entry's resolved scope labels (nil when unscoped).
+	// A scoped entry lives at a distinct address, so these must be echoed back on
+	// expire to invalidate this exact entry.
```
zhming0: Hmm, this comment doesn't quite make sense from agent side. Do we need it?

api/cache.go
```diff
@@ -88,13 +88,19 @@ type CacheEntryRetrieveResp struct {
...
+	Scopes map[string]string `json:"scopes,omitempty"`
 }
 
 // CacheEntryExpireReq is the request body for invalidating a cache entry.
-// The address should be the resolved entry, as echoed by the retrieve response.
+// The address should be the resolved entry, as echoed by the retrieve response —
+// including Scopes, which select the exact scoped entry (omit for an unscoped one).
```
zhming0: I don't feel we need to specifically mention "including Scopes" because it's obvious from the code change. I'd recommend a small tweak in our AGENTS.md to make it smarter about what code comment is appropriate vs not.

zhming0 (approved):
LGTM 🚀

ss1909:
closing this PR and tracking an alternative change here - https://github.com/buildkite/buildkite/pull/32088

PR #4148, opened 2026-07-31
Fix stray log timestamps under terminal redraws

petetomasik:
## Description

When an agent runs with `--no-ansi-timestamps --timestamp-lines`, job logs end up holding stray timestamps: some on rows of their own, some with only a tail surviving. `--timestamp-lines` prefixes each line at position 0, ahead of any carriage returns or cursor movement the process has already emitted. Programs that redraw progress in place (Bazel's curses UI here, but anything using `\r` or `ESC[1A` behaves the same) then move the cursor before the rest of the line is written, so the text lands on a different row and the timestamp is left behind, or the next redraw writes over it.

That residue is what produced the reported symptom. The agent writes `^^^ +++` after a command failure, and when that directive lands on a row still holding an orphaned timestamp it overwrites the first seven characters of it, producing `^^^ +++7-31T15:56:23Z]`. The frontend read that as an expand-and-rename directive and retitled the enclosing group with the fragment.

The rename itself is already fixed. buildkite/buildkite#31875 tightened `CURRENT_GROUP_REGEX` so residue on the row can no longer be read as a rename, and that shipped. This PR removes the cause rather than the symptom, so the residue stops being written in the first place.

The fix is to put the timestamp where it survives on the row rather than always at position 0. A process that redraws without ever ending a line arrives here as one line, so the timestamp goes after the last carriage return and the control run following it, beside the text that is actually left on the row. Where a line only opens with control, the timestamp goes after that run when the run would displace it: cursor movement relocates the rest of the line to another row, and an erase wipes a timestamp written in front of it. Colour codes, forward movement, private `?` sequences and anything the renderer draws as literal text do not displace anything, so those keep the timestamp at position 0 and aligned with every other line's. Lines a process used only to move the cursor are left untimestamped, since prefixing them puts visible output on a row the process left empty.

Header time behaviour is deliberately untouched. `agent/header_times_streamer.go` is not modified. The timestamp is always inserted immediately before whatever survives on the row, so any group directive on the line is pushed off column zero exactly as position-0 prefixing did, and the only lines left untimestamped consist entirely of control sequences, which contain no directive. Group counting and header times are therefore identical to `main`.

## Context

[Linear SUP-7705](https://linear.app/buildkite/issue/SUP-7705/job-log-groups-render-with-a-timestamp-fragment-as-their-title)

Found while investigating a customer job log with a malformed group title.

Frontend companion, already merged: buildkite/buildkite#31875.

## Changes

`agent/job_runner.go` adds `timestampLine`, which replaces the unconditional `fmt.Sprintf("[%s] %s", ts, line)` in the `--timestamp-lines` path, plus two regexes. `leadingTerminalControlRE` finds where a run of control ends and is deliberately wider than what any renderer acts on, though `ESC D`, `ESC E` and `ESC c` stay out of it because terminal-to-html draws those as literal text and so they end the run. `displacingControlRE` decides whether that run would displace a timestamp written in front of it.

`agent/job_runner_test.go` adds table-driven coverage, 36 shapes using control sequences taken verbatim from the affected log.

`clicommand/agent_start.go` updates the `--timestamp-lines` help text, which said the flag prepends a timestamp to every line and no longer does.

No CLI arguments change. The default ANSI-timestamps path is untouched, since it never reaches this code.

## Testing

`go mod tidy`, `go vet`, `golangci-lint run` and `git diff --check` are all clean.

## Disclosures / Credits

Claude Code (Opus 5) did the investigation and wrote the change, and I directed it, reviewed the result and made the call on scope. It traced the symptom from the raw job log through the agent, terminal-to-html and the frontend log parser, and identified the root cause.


DrJosh9000:
Thanks @petetomasik. Both `--no-ansi-timestamps` and `--timestamp-lines` are removed in V4, which is scheduled to be released to stable on 1st Sept. Is this worth doing now?

petetomasik:
> Thanks @petetomasik. Both `--no-ansi-timestamps` and `--timestamp-lines` are removed in V4, which is scheduled to be released to stable on 1st Sept. Is this worth doing now?

Thanks @DrJosh9000, good to know! With the fix for this particular issue already live on the frontend, I'll close this. We can visit after V4 is out and settled if it still presents any weirdness.

PR #4161, opened 2026-08-01
feat: Populate on-host mirrors from remote mirrors

pda:
## Description

Populate and refresh the shared on-host mirror from the server-provided remote mirror before falling back to canonical. This is PR 2 based on #4160.

## Context

The shared mirror’s durable `origin` and canonical heads remain canonical for mixed-version agents. Remote objects are pinned in a namespaced ref without moving canonical refs.

## Changes

- Create mirrors through shared-permission staging, canonicalize `origin`, and atomically publish.
- Reclaim interrupted staging and fail open to canonical if optional cleanup fails.
- Refresh by exact SHA with `--no-tags` and hidden URL prompts.
- Require warm objects to be reachable from a durable ref, preventing failed ref writes from becoming retry hits.
- Reconcile canonical `origin` on remote and warm hits before shared rename maintenance.
- Skip the redundant canonical exact-SHA checkout fetch after an on-host hit.
- Bypass mirrors containing an unpinned build object and dissociate reused checkouts from stale alternates before canonical fetch.
- Continue canonical checkout without an on-host reference on lock timeout.
- Preserve locks, snapshots, submodules, skip-update, cancellation, and reference/dissociate modes.

## Testing

Real Git-over-HTTP tests cover atomic creation, permissions, staging, lag, timeout, namespaced refs, tag/head isolation, fallback, warm/remote collision repair, failed ref write plus canonical force-push plus reused-checkout retry, lock contention, prompt hiding, and cancellation. The retry regression verifies the stale alternate is removed before canonical fetch; test executors close checkout roots for Windows cleanup.

## Disclosures / Credits

Implemented with Cursor under human direction. Independent review drove atomic publication, durable reachability, retry alternate cleanup, warm-hit maintenance, and contention fallback.

[...]

PR #4162, opened 2026-08-01
feat: Refresh reused checkouts from remote mirrors

pda:
## Description

Fetch the immutable build commit from the remote mirror into a reused checkout before falling back to canonical. This is PR 3 based on #4161.

A mirror miss never deletes the checkout. Unsafe promisor-cleanup failure quarantines the checkout's Git state, recreates an empty workdir for teardown hooks, and lets the next mirror-ineligible attempt clone canonically.

## Context

The explicit mirror URL makes the bounded source deterministic despite customer `insteadOf` rules. That requires filter inheritance and one-shot promisor cleanup without mutating user-owned promisor configuration.

## Changes

- Attempt only immutable commit fetches and skip canonical fetch only after a local hit.
- Preserve effective fetch flags, including Git-accepted `--fi*` and `--no-fi*` abbreviations.
- Inherit origin filters only without an explicit override.
- Establish canonical promisor ownership before removing mirror ownership on every outcome.
- Write a credential-free digest marker before mirror fetches and repair only the matching agent-created promisor after interruption, URL removal, or rotation.
- Leave unmarked and unrelated user promisor configuration untouched; clear marker-only crash windows safely.
- Preserve pre-existing unmarked config for the exact mirror URL by skipping the mirror and using canonical.
- Fall back to canonical when marker setup or ownership inspection fails.
- Route cleanup failure through clean canonical retry, prioritising quarantine over cancellation handling.
- Keep gitfile-backed reused checkouts canonical because external Git directories cannot be safely quarantined with the worktree.
- Recreate an empty checkout directory after quarantine so teardown hooks retain a valid working directory.
- Run mirror-keyed config commands silently with checkout PATH resolution, preventing URL-bearing argv/stderr leaks.
- Preserve cancellation without starting canonical fallback.

Filtered Git fetches can transiently write URL-named promisor keys. Completed fetches remove the keys and marker; the next fetch repairs only matching marked state left by process interruption.

F2 mirror-routed lazy materialisation remains outside this stack; lazy reads use canonical `origin`.

## Testing

Real Git-over-HTTP tests cover hit, lag, failure, timeout, cancellation, automatic/inherited/abbreviated filters, marked repair after URL removal, preservation of unmarked user promisors including an exact-URL section with extra keys, marker-only crash recovery, marker privacy and setup failure, canonical materialisation, credential-safe config failures, checkout PATH, state preservation, and debug suppression. Regression tests additionally force cancellation plus config-lock cleanup failure, verify unsafe Git state is removed while the teardown workdir remains, and keep real `--separate-git-dir` checkouts mirror-ineligible.

## Disclosures / Credits

Implemented with Cursor under human direction. Independent review drove all-outcome and restart recovery, scoped ownership marking, exact-URL collision handling, option parity, safe retry, gitfile/PATH correctness, teardown invariants, and logging hardening.

### DIFF
--- a/internal/job/checkout_fetch.go
+++ b/internal/job/checkout_fetch.go
@@ -212,6 +212,22 @@ func (e *Executor) fetchExistingCheckoutFromRemoteMirror(
 	attempt *remoteMirrorAttempt,
 	gitFetchFlags string,
 ) (bool, error) {
+	hasURLConfig, err := e.hasRemoteMirrorURLConfig(ctx, attempt.url)
+	if err != nil {
+		if ctx.Err() != nil {
+			return false, ctx.Err()
+		}
+		attempt.outcome = remoteMirrorOutcomeError
+		e.shell.Warningf("Could not inspect remote Git mirror state; falling back to canonical")
+		return false, nil
+	}
+	if hasURLConfig {
+		attempt.outcome = remoteMirrorOutcomeSkipped
+		attempt.skipReason = remoteMirrorSkipURLConfigConflict
+		e.shell.Commentf("Remote Git mirror URL already has user-owned config; falling back to canonical")
+		return false, nil
+	}
+
 	if err := e.markRemoteMirrorPromisor(ctx, attempt.url); err != nil {
 		if ctx.Err() != nil {
 			return false, ctx.Err()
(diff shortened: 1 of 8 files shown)


internal/job/checkout_fetch.go
```diff
@@ -168,6 +201,282 @@ func (e *Executor) fetchSource(ctx context.Context, addBloblessFilter bool, atte
...
+		if ctx.Err() != nil {
+			return false, ctx.Err()
+		}
+		attempt.outcome = remoteMirrorOutcomeError
+		e.shell.Warningf("Could not mark remote Git mirror state; falling back to canonical")
+		return false, nil
+	}
+
+	e.shell.Commentf("Fetch commit from remote Git mirror")
+	// C21: preserve the caller's effective fetch flags. Explicit ref-mutating
```
catkins, line 241: Is this a plan reference?

internal/job/git.go
```diff
@@ -150,16 +150,14 @@ func gitCheckout(ctx context.Context, sh *shell.Shell, gitCheckoutFlags, referen
...
-// hasPartialFilterFlags returns true if flags contains a
-// --filter=<spec> or --filter <spec> option.
+// hasPartialFilterFlags returns true if flags contains a filter option,
+// including Git's accepted --fi* long-option abbreviations.
 func hasPartialFilterFlags(flags []string) bool {
 	for i, f := range flags {
-		// Check for --filter=<spec>
-		if strings.HasPrefix(f, "--filter=") {
+		if strings.HasPrefix(f, "--fi") && strings.Contains(f, "=") {
 			return true
```
catkins, line 158: Will this match invalid too?

catkins (approved):

internal/job/checkout_fetch.go
pda: yep… https://github.com/buildkite/agent/blob/01c20f6b38521fdff2cbe344041a2b9b21b47161/docs/remote-git-mirrors.md#L1445-L1452

maybe shouldn't be cited from a code comment, although that plan does exist on `main` for now.

internal/job/git.go
pda: Seems reasonable 🤔

PR #4165, opened 2026-08-03
A-1641: Consume control-plane OTLP exporter configuration from registration

lox:
Agent-side companion to buildkite/buildkite#31880. Replaces #4155 (and the earlier #4149/#4150) with a fresh, deliberately minimal implementation.

Design doc: [A-1641](https://linear.app/buildkite/issue/A-1641/consume-control-plane-otlp-exporter-config-from-registration) (Oracle-reviewed, revised 2026-08-04 after product review; previously in `docs/plans/`, now moved out of the repo).

## Threat model

The security bar is explicit and modest: **no worse than baking an OTLP endpoint and auth token into the agent image**. In that baseline the credential is in the agent process env, inherited by bootstrap, and inherited by every hook and job command — which is also exactly how userspace tracing works today. This PR intentionally omits the machinery that caused thrash in #4155 — no FD passing, no transport pinning, no debug-HTTP redesign, no custom-bootstrap gating, and (as of the latest revision) no post-tracer-setup scrubbing.

## What it does

**1. Registration contract.** The agent unconditionally advertises the `control-plane-otlp-tracing` capability at registration. The server may respond with a `tracing` block:

```json
{
  "tracing": {
    "backend": "opentelemetry",
    "propagate_traceparent": true,
    "exporter": {
      "endpoint": "https://collector.example/v1/traces",
      "protocol": "http/protobuf",
      "headers": {"Authorization": "Bearer …"}
    }
  }
}
```

Wire types live in `api/tracing.go`.

**2. Local-wins merge.** `agent.ApplyControlPlaneTracing` merges the server policy into each worker's config copy. Explicit local configuration always wins: a locally configured non-OTel backend ignores the policy entirely; a local OTLP destination in the agent's env (endpoint or headers, generic or traces-specific, presence counting even when empty) blocks the server exporter; an explicit local `propagate_traceparent` (including `false`) wins. A server policy for an unsupported backend is ignored in full — even when the local backend is already `opentelemetry`. One sanitized (scheme+host only, never headers) log line per worker.

**3. Delivery via standard OTel vars, passed through to the job.** The job runner injects `OTEL_EXPORTER_OTLP_TRACES_{ENDPOINT,PROTOCOL,HEADERS}` into the bootstrap process environment (endpoint always; protocol and headers only when supplied, never as empty-valued vars — the OTel spec treats empty as unset but not every SDK does; header values are percent-escaped, including `+` as `%2B` for form-style SDK parsers). Bootstrap uses them for its own tracer (traces-specific protocol now correctly takes precedence over the generic var), and **hooks, plugins, and the job command inherit them** — so userspace OTel tooling can join the agent's trace via the propagated `TRACEPARENT` and export its spans to the same collector with the distributed credential. That pass-through is the point of the feature (per product review; an earlier revision scrubbed the vars after tracer setup, which broke userspace span export — the marker/restore/cleanup machinery that scrub required is deleted).

**4. Per-job collision rule, all-or-nothing.** If the backend job env (pipeline/step) already sets *any* OTLP destination var — `OTEL_EXPORTER_OTLP_{ENDPOINT,HEADERS}` or the traces-specific pair, presence counting even when empty — injection is skipped entirely for that job (logged). Endpoint and headers must travel together: no partial merge, so a server credential is never sent to a pipeline-chosen endpoint or vice versa. The same four-name list (`env.OTLPDestinationVars`) backs both this check and the agent-local one, and the check runs against the case-normalized job env so case-variant names can't dodge it on Windows.

**5. Env-file hygiene.** Injection happens after `BUILDKITE_ENV_FILE`/`BUILDKITE_ENV_JSON_FILE` are written, so the credential never lands on disk — matching the baseline, where baked env isn't in the env files either. Known consequence: Docker/Compose `propagate-environment` uses the env file as a name allowlist, so containers those plugins create need explicit `environment` propagation (documented in the design doc; a cross-repo follow-up, not a reason to write credentials to disk).

## Security properties

- Credential never enters the on-disk env files.
- Pipeline-chosen destinations are never mixed with the server credential (all-or-nothing skip).
- Traces-specific env vars prevent the credential reaching logs/metrics exporters via generic-var fallback.
- Logging is sanitized: endpoint scheme+host only, never headers.
- Hooks/commands/Job API **do** see the credential — by design, they need it to export spans; identical exposure to a baked-in token. `env` in a step prints it to the build log, same as baked-in today; operators can add `OTEL_EXPORTER_OTLP_TRACES_HEADERS` to `--redacted-vars` if they care.

## Validation

- `go build ./...`, `go vet`, `GOOS=windows go build`, `golangci-lint run` (0 issues)
- `go test ./agent/ ./api/ ./env/ ./internal/job/ ./clicommand/` — all pass
- `JobRunner.createEnvironment`-level tests cover the contract directly: injected values reach the bootstrap env and stay out of both env files; each of the four destination vars (present and present-empty) blocks all injection while pipeline values survive; protocol-only and logs/metrics-signal vars don't block; case-variant collision covered on Windows
- End-to-end bootstrap smoke test: with the exporter vars in the bootstrap process env, the job command sees all three `OTEL_EXPORTER_OTLP_TRACES_*` values (pass-through confirmed)
- Two earlier Oracle passes found and fixed: Windows case-normalization of job env keys, unsupported-server-backend acceptance when local backend was OTel; a third Oracle pass reviewed the pass-through pivot (verdict: go-with-changes, all changes incorporated)
- A final Oracle pass before marking ready fixed: empty-valued protocol/headers vars are no longer injected (omitted instead), `+` in header values is escaped as `%2B` for form-style SDK parsers, and Windows key normalization is scoped to the collision check instead of rewriting the whole job env (env-file key spelling unchanged)

## Notes for reviewers

Choices you may want to scrutinize:

- The capability is advertised unconditionally (no opt-in flag); the server decides whether to send a policy.
- An unsupported server backend ignores the whole policy with a warning, rather than partial application.
- When a pipeline sets its own OTLP destination, only exporter injection is skipped — the server-applied backend and traceparent propagation still stand.
- FD passing, transport pinning, debug-HTTP changes, and scrubbing exporter vars from the job env are deliberately out of scope per the threat model and product intent above.


agent/agent_configuration.go
```diff
@@ -61,27 +62,33 @@ type AgentConfiguration struct {
...
+	RedactedVars                []string
+	AcquireJob                  string
+	TracingBackend              string
+	TracingServiceName          string
+	TracingPropagateTraceparent bool
+	// ControlPlaneTracingExporter is an OTLP trace exporter destination
+	// supplied by the control plane at registration (see
+	// ApplyControlPlaneTracing). Its headers may hold credentials: it is
+	// delivered only to the bootstrap process environment, never job env
+	// files, and bootstrap strips it before hooks and the command run.
```
catkins: I thought we wanted to propagate it to the job env and bootstrap

agent/control_plane_tracing.go
```diff
@@ -0,0 +1,151 @@
...
+
+	"github.com/buildkite/agent/v3/api"
+	envutil "github.com/buildkite/agent/v3/env"
+	"github.com/buildkite/agent/v3/logger"
+	"github.com/buildkite/agent/v3/tracetools"
+)
+
+// LocalTracingConfig captures the operator's explicit local tracing choices,
+// which always take precedence over control-plane (registration) tracing
+// policy. See docs/plans/2026-08-03-control-plane-otlp-exporter.md.
```
catkins: Will this plan be here after the pr?

agent/control_plane_tracing.go
```diff
@@ -0,0 +1,151 @@
...
+		}
+	}
+	return false
+}
+
+// ApplyControlPlaneTracing merges control-plane tracing policy from an agent
+// registration response into a single worker's copy of the agent
+// configuration. Local configuration always wins: a local non-OTel backend
+// disables the whole policy, a local OTLP destination disables the server
+// exporter, and an explicitly-set local propagate-traceparent keeps its
```
catkins, line 49: We should probably make propagate traceparent default to true in v4

env/control_plane_otlp.go
catkins, line 1: I'm not sure why we aren't just propagating them normally?

agent/control_plane_tracing.go
DrJosh9000: Already is: https://github.com/buildkite/agent/pull/3807/changes/9285a2730f526a3f8ceb07bd706eaa54596c452f

agent/agent_configuration.go
lox: Yes — this comment was stale from the earlier revision that scrubbed the vars after tracer setup. Since 6d8b99a the exporter config is passed through: bootstrap uses it for its own tracer and hooks/plugins/command inherit the `OTEL_EXPORTER_OTLP_TRACES_*` vars, so job-level tooling can export spans to the same collector. Doc comment fixed in d8f6abe.

env/control_plane_otlp.go
lox: As of 6d8b99a we do propagate them to hooks and the command — the earlier scrub/restore machinery is deleted. The two remaining deliberate deviations from fully "normal" propagation: (1) traces-specific vars rather than generic `OTEL_EXPORTER_OTLP_*`, so the credential can't be picked up by a logs/metrics exporter via the generic-var fallback; (2) they're set on the bootstrap process env after the env files are written rather than added to `Job.Env`, which keeps the credential out of `BUILDKITE_ENV_FILE` on disk (trade-off: docker's `propagate-environment` name-allowlist won't auto-forward them into containers — noted in the design doc). Happy to flatten either of those if you think full normal propagation is better.

agent/control_plane_tracing.go
lox: Currently yes — it's committed under `docs/plans/` (a new directory) as the design record for the feature, and code comments reference it. If you'd rather the repo not accumulate plan docs, I can drop it from the branch and keep the content in the PR description only — your call.

agent/control_plane_tracing.go
catkins: oh perfect

clicommand/agent_start.go
```diff
@@ -275,6 +275,11 @@ func (asc AgentStartConfig) Features(ctx context.Context) []string {
 		features = append(features, "propagate-traceparent")
 	}
 
+	// This agent can consume control-plane OTLP exporter configuration from
+	// the registration response (see
+	// docs/plans/2026-08-03-control-plane-otlp-exporter.md).
+	features = append(features, "control-plane-otlp-tracing")
```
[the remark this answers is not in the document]
catkins: Mr buildsworth I think you are misunderstanding the intent of that contract, that is about how the exporter config is persisted. We are intentionally providing it to the agent.

env/control_plane_otlp.go
catkins: 👏

catkins (approved):
🚀

PR #4175, opened 2026-08-04
Add systemd watchdog notification support

scadu:
## Description

Adds opt-in systemd watchdog support to the agent. When enabled through systemd's `WatchdogSec`, the agent only sends notifications while its workers are healthy and successfully contacting Buildkite.

Scaler-side recovery was considered, but without agent lifecycle events it has to guess which EC2 instance is unhealthy.

## Context

[SUP-6917](https://linear.app/buildkite/issue/SUP-6917)

## Changes

- Add a small cross-platform systemdnotify package.
- Notify systemd at half the configured watchdog interval.
- Stop notifying after a worker fails or misses successful Buildkite contact for two heartbeat intervals.
- Keep the feature inactive when the systemd watchdog isn't configured.

## Testing
- [x] Tested with a custom agent built from this branch on EC2.

In test build with an instance under memory, CPU and IO pressure, last watchdog registered about three minutes before system shutdown has started (that was the configured action in my systemd unit), indicating watchdog functioning as expected.

<details>
<summary>Exact systemd unit config</summary>

`buildkite-agent.service`
```
[Unit]
Description=Buildkite Agent
Documentation=https://buildkite.com/agent
After=syslog.target
After=network.target
After=docker.service

[Service]
Type=simple
User=buildkite-agent
Environment="HOME=/var/lib/buildkite-agent"
Environment="PATH=/usr/local/bin:/usr/bin"
Environment="USER=buildkite-agent"
Environment="TMPDIR=/var/tmp"
# The =- (rather than just = ) in the line below means that systemd won't complain if it can't find the file
EnvironmentFile=-/var/lib/buildkite-agent/env
ExecStart=/usr/bin/buildkite-agent start
ExecStopPost=/usr/local/bin/terminate-instance
RestartSec=5
Restart=on-failure
RestartForceExitStatus=SIGPIPE
TimeoutStartSec=10
TimeoutStopSec=0
KillMode=process
# https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#OOMScoreAdjust=
# Set to -1000 to disable OOM killing of processes of buildkite-agent service
# managing EC2 instance lifecycle
OOMScoreAdjust=-1000

[Install]
WantedBy=multi-user.target
```

`20-systemd-watchdog.conf` to override items like `OOMScoreAdjust=-1000`
```
[Unit]
FailureAction=poweroff-force

[Service]
WatchdogSec=3min
WatchdogSignal=SIGKILL
TimeoutAbortSec=10s
TimeoutStopSec=10s
ExecStopPost=
Restart=no
OOMScoreAdjust=0
OOMPolicy=kill
```

</details>

## Disclosures / Credits

Medikinet CR 20


agent/agent_watchdog.go
```diff
@@ -0,0 +1,109 @@
...
+	age := now.Sub(lastSuccessfulContact)
+	if age > maxContactAge {
+		return fmt.Errorf("worker %d has not successfully contacted Buildkite for %s", a.spawnIndex, age.Round(time.Second))
+	}
+	return nil
+}
+
+func (ap *AgentPool) runWatchdog(ctx context.Context, l logger.Logger, watchdogInterval time.Duration) {
+	// systemd recommends notifying every half of the configured interval.
+	ticker := time.NewTicker(watchdogInterval / 2)
```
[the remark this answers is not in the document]
scadu: Addressed in dbce73b

agent/agent_pool.go
```diff
@@ -20,13 +21,18 @@ import (
 type AgentPool struct {
 	workers     []*AgentWorker
 	idleTimeout time.Duration
+	watchdog    watchdogNotifier
+	watchdogErr error
```
DrJosh9000: It would be more idiomatic to change `NewAgentPool` to return `(*AgentPool, error)` and return the error at that point.

agent/agent_watchdog.go
```diff
@@ -0,0 +1,111 @@
...
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/buildkite/agent/v4/logger"
+)
+
+type watchdogNotifier interface {
```
DrJosh9000, line 11: Good use of an interface 😄 (particularly how it is defined on the using side, not the implementation side)

agent/agent_pool.go
```diff
@@ -89,13 +118,25 @@ func runWorker(ctx context.Context, worker *AgentWorker, idleMon *idleMonitor) e
...
 
 	// Starts the agent worker and wait for it to finish.
-	return worker.Start(ctx, idleMon)
+	worker.watchdogMarkRunning()
+	err := worker.Start(ctx, idleMon)
+	if err == nil || ctx.Err() != nil {
+		worker.watchdogMarkStopped()
+	} else {
+		worker.watchdogMarkFailed(err)
+	}
```
DrJosh9000: This looks a bit strange. Should we instead condition on the error returned? If `ctx.Err() != nil` then probably `errors.Is(err, context.Canceled)` (in particular, `err != nil`).

Perhaps `watchdogMarkStopped` could preserve any existing error, meaning it can safely be deferred right after `watchdogMarkRunning`?

agent/agent_pool.go
scadu: Hmmm, righto. Will remodel it ;)

agent/agent_pool.go
scadu: Fair point. Checking `ctx.Err()` could hide a real worker error if cancellation races with the worker returning. I'll check the returned error directly, treating `context.Canceled` as an intentional stop, and make `watchdogMarkStopped` preserve any existing failure so it can be safely deferred after `watchdogMarkRunning` as suggested. Unless you've other ideas @DrJosh9000 🕵️

agent/agent_pool.go
scadu: Addressed in 14d5a99

agent/agent_pool.go
scadu: Addressed in a1ac821

DrJosh9000 (approved):
LGTM, thank you! ✅ ✅

PR #4179, opened 2026-08-05
fix: Trust warm mirror hits without reachability scans

pda:
## Description

Checkout of large repositories with on-host mirrors stalled for minutes on `git for-each-ref --contains` reachability scans in `updateGitMirror` — observed at 3m15s per job on the buildkite/buildkite mirror, both with and without a remote mirror URL configured. The two prior hotfixes ([#4171](https://github.com/buildkite/agent/pull/4171), [#4172](https://github.com/buildkite/agent/pull/4172)) only exempted the no-URL paths.

This removes the reachability guard entirely and returns warm hits to the presence-only check the agent shipped with for years. That deliberately re-accepts a narrow pre-existing risk: an object present in the mirror but reachable from no mirror ref can be lent to a `--reference` checkout and later pruned by mirror gc. The failure needs an unpinned object (interrupted fetch, or rebuilding a force-pushed-away commit), a prune past `gc.pruneExpire` (default two weeks), and a borrowing checkout to coincide — and it fails one job with a retryable `bad object` error that the existing retry-clean path heals with a fresh canonical clone. The alternative (proving reachability) is inherently a global scan whose cost scales with the repository size the mirror feature exists to serve. The decision is recorded as caveat C22 in `docs/remote-git-mirrors.md`, with guidance to verify only expected refs if stronger pinning is ever wanted.

## Context

- [Production job with remote mirror enabled: 3m17s gap between mirror fetch and clone](https://buildkite.com/buildkite/buildkite/builds/208214/canvas?sid=019fd0b1-e294-4817-a047-efbc458f14db&tab=output)
- Follow-up to [#4171](https://github.com/buildkite/agent/pull/4171) and [#4172](https://github.com/buildkite/agent/pull/4172), which fixed the same stall for jobs without `BUILDKITE_GIT_REMOTE_MIRROR_URL`

## Changes

- Delete `hasGitCommitReachableFromRef` and both call sites (warm-hit check and post-fetch guard).
- Delete both no-URL carve-outs from the prior hotfixes — presence-only now applies uniformly, removing the per-path reasoning that caused three incident rounds.
- Warm hits now flow through the shared `updateRemoteURL` tail instead of returning early, so repository-rename repair runs on warm hits. This also fixes the stale-origin edge Codex flagged on #4171.
- Gate the "Updating existing repository mirror" log message so warm hits no longer log a misleading update.
- Replace the deleted unpinned-retry regression with a table-driven warm-hit test (no-URL and URL-configured) asserting no `for-each-ref` runs, plus a unit test preserving coverage of the bypassed-mirror alternate dissociation.
- Record the accepted risk as caveat C22 in `docs/remote-git-mirrors.md`.

## Testing

The full `TestUpdateGitMirror*`, `TestPrepareCheckoutWorkdir*`, and `TestGetOrUpdateMirrorDir*` suites pass under the race detector.

Known trade-offs called out by review, both inside the accepted decision: a present-but-unpinned object is no longer re-pinned by a mirror fetch (it stays unpinned for its gc window), and warm hits on mirrors with colliding canonical URLs (lossy `dirForRepository`) now pay the rename-repair `fsck`/`gc` that the early return previously skipped.

## Disclosures / Credits

Cursor Agent diagnosed the production stalls from job-log timestamps, implemented the removal after the trade-off was decided by a human, and ran independent ship-risk, maintainability, and simplicity reviews.

### DIFF
--- a/docs/remote-git-mirrors.md
+++ b/docs/remote-git-mirrors.md
@@ -922,8 +922,9 @@ no mirror ref reaches; mirror `gc` may later prune it out from under a
 `--reference` checkout. Proving reachability requires
 `for-each-ref --contains`, a global scan that cost minutes per job on
 production mirrors with hundreds of thousands of refs. Accept the narrow risk
-instead of paying that scan on every warm hit; the namespaced destination ref
-written on every mirror hit still shrinks the unpinned population.
+instead of paying that scan on every warm hit. The namespaced destination ref
+pins every commit the mirror fetch downloads; a commit that is already present
+but unpinned stays unpinned, which is the accepted exposure.
 
 Do not return directly on the remote hit. Record it in the local boolean and let
 the common maintenance and snapshot tail run.
--- a/docs/remote-git-mirrors.md
+++ b/docs/remote-git-mirrors.md
@@ -916,12 +916,14 @@ booleans, skip only the applicable network fetch, and share the existing
 maintenance and snapshot tail. Keep the remote attempt inside the
 `isMainRepository` guard so submodule mirrors remain untouched.
 
-The warm-cache check requires both object presence and reachability from a
-durable ref. A previous fetch can download an object and then fail its
-destination-ref write; if canonical fallback also fails, a presence-only retry
-would skip both sources and hand reference checkouts an object that `gc` may
-remove. Reachability forces the retry to publish the namespaced ref or complete
-the canonical fetch.
+The warm-cache check is presence-only by design (C22). A previous fetch can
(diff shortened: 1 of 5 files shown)


pda:
@buildsworth-bk review approve L2

lox (approved):

catkins (approved):

pda:
Running some integration / perf testing across this before merge.

pda:
Integration / performance testing is good 👍🏼

PR #4187, opened 2026-08-06
A-1494: Verify cache checksum on restore

zhming0:
## What

Verifies a downloaded cache blob's SHA256 digest against its content addressed name before anything touches the target folders. A mismatch is treated as a clean miss: nothing is unpacked, existing files stay intact, and the build continues. The entry is also expired so a subsequent save replaces the bad blob. A warning and tracing attributes record the mismatch.

Takes over #4169. The first commit is that PR rebased onto main (module path updated to v4). The second commit refines it:

* Digest verification lives in its own `digest.go` with a `verifyBlobDigest` span, keeping `restore.go` focused on the restore workflow.
* `verifyBlobDigest` hashes via `io.Copy` through a context aware reader. Memory use is a fixed 32 KiB buffer regardless of archive size.
* A mismatch expires the entry, the same path as a missing blob. This goes beyond the ticket, which only asked for a clean miss: a blob that never verifies would otherwise cost a wasted download and hash on every restore until TTL. Worst case with a systematic digest bug is one expire and re-upload per build, visible via the mismatch warning and span attributes. Both sides of the comparison are computed by the agent itself (save names the blob with the same hash), so drift risk is low, and unknown algorithms skip verification entirely.
* The flagship integration test is a true cache poisoning case: a valid archive containing a backdoor file is planted under the legitimate blob's content addressed name. Extraction of it would succeed silently, so the digest check is the only defence. The test asserts the backdoor never lands on disk, the pre existing sentinel survives, the entry is expired, and a re-save re-creates it.

## Why

A stored blob's name is a fingerprint of its contents, but we never checked it on restore. A corrupted or substituted blob would be silently unpacked. See A-1494.

On hashing during download instead of after: the ticket suggested hashing the bytes as they arrive, but the S3 backend downloads parts concurrently and out of order, and the nsc backend shells out to a CLI that writes the file itself. Neither exposes a sequential byte stream. The verify pass reads a file written moments earlier, so it is served from the page cache, and the SHA256 cost is the same either way.

## Testing

* `go test ./internal/cache/...` passes.
* Confirmed the poisoning test fails when the digest check is disabled: the backdoor file gets planted and the assertion catches it.

closes #4169

### DIFF
--- a/internal/cache/digest.go
+++ b/internal/cache/digest.go
@@ -1,0 +1,85 @@
+package cache
+
+import (
+	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/buildkite/agent/v4/api"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/codes"
+)
+
+// ErrDigestMismatch means a downloaded blob's bytes don't hash to the
+// content-addressed name it was fetched under (e.g. a corrupt or half-written
+// upload). Restore treats it as a clean miss and never unpacks the file.
+var ErrDigestMismatch = errors.New("cache blob digest mismatch")
+
+// verifyBlobDigest re-hashes the downloaded archive and confirms it matches
+// the content-addressed name it was fetched under.
+//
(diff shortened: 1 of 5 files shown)


ss1909:
LGTM ❤️

ss1909 (approved):

PR #4190, opened 2026-08-06
[SUP-7744]: Update artifact retention period from six months to 90 days

stephanieatte:
## Description

Corrects the artifact retention period stated in the `buildkite-agent artifact upload` CLI help text. It said "six months," but the actual Buildkite-managed artifact storage retention policy is 90 days.

## Changes

- `clicommand/artifact_upload.go`: updated retention text from "six months" to "90 days"

### DIFF
--- a/clicommand/artifact_upload.go
+++ b/clicommand/artifact_upload.go
@@ -27,7 +27,7 @@ You can specify an alternate destination on Amazon S3, Google Cloud Storage
 or Artifactory as per the examples below. This may be specified in the
 'destination' argument, or in the 'BUILDKITE_ARTIFACT_UPLOAD_DESTINATION'
 environment variable.  Otherwise, artifacts are uploaded to a
-Buildkite-managed Amazon S3 bucket, where they’re retained for six months.
+Buildkite-managed Amazon S3 bucket, where they’re retained for 90 days.
 
 Example:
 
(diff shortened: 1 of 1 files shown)


clicommand/artifact_upload.go
```diff
@@ -27,7 +27,7 @@ You can specify an alternate destination on Amazon S3, Google Cloud Storage
 or Artifactory as per the examples below. This may be specified in the
 'destination' argument, or in the 'BUILDKITE_ARTIFACT_UPLOAD_DESTINATION'
 environment variable.  Otherwise, artifacts are uploaded to a
-Buildkite-managed Amazon S3 bucket, where they’re retained for six months.
+Buildkite-managed Amazon S3 bucket, where they’re retained for 90 days.
```
moskyb: i wonder if it would be better to not mention the retention period at all — it's the kind of thing that we could change at some point.

omehegan: @moskyb I'm inclined to agree, as we see it has already changed once and now requires an agent release to correct this help. 

@stephanieatte I propose that we remove the retention comment entirely from the agent source, and retain it in the non-generated parts of the docs site.

stephanieatte: Sure @omehegan will make the change now

DrJosh9000 (approved):

PR #4193, opened 2026-08-07
feat: Use remote Git mirrors for pull request head builds

pda:
## Description

Remote Git mirrors currently skip every pull request build (`skip_reason=pull-request`), which is most builds on active repositories. That guard was over-conservative: mirror interactions are keyed entirely by the build's immutable commit SHA — the mirror never serves `refs/pull/*` — and eligibility already requires a full object ID, so a lagging mirror can only miss, never serve wrong content.

PR **head** builds now use the mirror at every site. The canonical `refs/pull/N/head` fetch remains the fallback on any miss, with retry semantics unchanged. This is F3 from `docs/remote-git-mirrors.md`, unblocked by the provider confirming `refs/pull/*` replication (Q3).

Two PR shapes deliberately stay canonical:

- **Merge-ref builds** (`refs/pull/N/merge`): the speculative merge commit is regenerated by the provider rather than pinned to the build's object ID, and its canonical fetch doubles as the merge-conflict fail-fast. New skip reason `pull-request-merge-ref`.
- **`--refmap` fetch flags on PR builds**: that is the one supported way the PR-ref fetch produces durable local refs (covered by `TestCheckingOutGitHubPullRequestAndCustomRefmap`); a mirror hit would leave them silently stale. New skip reason `pull-request-refmap`.

## Context

- F3 / Q3 in `docs/remote-git-mirrors.md`; the provider confirmed `refs/pull/*` refs are mirrored asynchronously like any other ref.
- Fork PR heads are often not reachable from any branch, so the tests pin the fork shape: commit reachable only through `refs/pull/*` and not an advertised tip, which exercises protocol v2's unadvertised reachable wants — the transport property the mirror probe relies on.

## Changes

- Eligibility (`resolveRemoteMirrorAttempt`): replace the blanket PR skip with the merge-ref and refmap guards. PR builds without a known commit stay ineligible via the existing full-object-ID clause.
- `fetchSource`: extract the existing-checkout mirror probe into `tryExistingCheckoutRemoteMirror` and call it from the PR-head arm as well as the commit arm. A hit skips the canonical PR-ref fetch; a miss falls through unchanged.
- On-host mirror (`updateGitMirror`): no code change needed — its exact-SHA probe already runs ahead of the refspec-specific canonical fetch arm.
- `githttptest`: forward the `Git-Protocol` header like `git http-backend` (previously every test negotiation silently downgraded to protocol v0) and deliver upload-pack's in-band ERR pkt-lines instead of replacing them with HTTP 500. Without this the fork-shape tests cannot exercise the v2 behavior production relies on.
- `docs/remote-git-mirrors.md`: correct the stale "Status: plan" header (the stack is on `main`), document the new eligibility including the refmap rationale, note fork branch names joining the namespaced-ref keyspace under C7, and mark F3 landed for head builds with tags/merge refs as remaining scope.

Telemetry note: the `pull-request` skip reason no longer exists; dashboards keyed on it should watch `pull-request-merge-ref` / `pull-request-refmap`, and PR head builds now report ordinary hit/miss outcomes.

## Testing

Full `internal/job` package passes (478s), plus race-detector runs of the mirror, fetch, and eligibility suites. New coverage: eligibility table rows (head eligible, merge-ref skipped, refmap skipped including the pre-2.36 `--ref` abbreviation, no-commit PR still ineligible, non-PR refmap still eligible); on-host and existing-checkout hit tests with canonical closed and the commit reachable only through `refs/pull/*` (asserting the namespaced ref is written, `refs/pull/*` never enters the on-host mirror, and promisor config is cleaned); lagging-mirror miss tests asserting canonical PR-ref fallback.

## Deployment

Behavior changes only for jobs where the backend supplies `BUILDKITE_GIT_REMOTE_MIRROR_URL`. Expected steady-state: brand-new PR pushes often miss (webhook beats async replication) at roughly one negotiation round-trip of cost, with hits on later jobs, retries, and rebuilds.

## Rollback

Revert the commit; the previous blanket PR skip returns. No persistent state format changes.

## Disclosures / Credits

Cursor Agent implemented this end-to-end from a human-approved plan, including the test-harness protocol fix, and ran independent ship-risk, maintainability, and simplicity reviews (two passes).


internal/job/checkout_fetch.go
```diff
@@ -208,6 +212,26 @@ func isExistingCheckoutRemoteMirrorAttempt(attempt *remoteMirrorAttempt) bool {
 		attempt.outcome == remoteMirrorOutcomeNotReached
 }
 
+// tryExistingCheckoutRemoteMirror probes the remote mirror for the build
+// commit when the attempt targets this existing checkout. On a hit the caller
+// can skip its canonical fetch: the checkout proceeds by exact object ID.
+// (C4: FETCH_HEAD then records the non-secret mirror URL. No mirror-eligible
+// flow reads it, so retain Git's normal write.)
```
zhming0: I see the good intention but things like C4, F3 in the repo will quickly become cryptic for both human and clankers.

internal/job/checkout_fetch.go
```diff
@@ -208,6 +212,26 @@ func isExistingCheckoutRemoteMirrorAttempt(attempt *remoteMirrorAttempt) bool {
...
+func (e *Executor) tryExistingCheckoutRemoteMirror(
+	ctx context.Context,
+	attempt *remoteMirrorAttempt,
+	gitFetchFlags string,
+) (bool, error) {
+	if !isExistingCheckoutRemoteMirrorAttempt(attempt) {
+		return false, nil
+	}
```
zhming0, line 267: Hmm, I don't fully understand this. 🤔 Is this to prevent us pulling the remote mirror repeatedly? Just wondering, if the remote mirror is assumed to be reasonably fast to catch up, maybe we don't need to be so conservative.

zhming0 (commented):
Just started reading related code, I left a question to start with, still reading 👀

internal/job/checkout_fetch.go
pda: agree

pda:
@buildsworth-bk review approve L2

zhming0:
I made a simplified PR in https://github.com/buildkite/agent/pull/4216. 🙏 
My main changs are so 1 we don't over commit on edge case and 2 make code comments easier to read.

PR #4195, opened 2026-08-07
A-1494: Verify cache checksum on restore (backport to v3)

ss1909:
## What

Backports #4187 to `v3`.

Verifies a downloaded cache blob's SHA256 digest against its content-addressed name before anything touches the target folders. A mismatch is treated as a clean miss: nothing is unpacked, existing files stay intact, and the build continues. The entry is also expired so a subsequent save replaces the bad blob. A warning and tracing attributes record the mismatch.

## Translation notes

- Rewrote `github.com/buildkite/agent/v4/...` import paths (and the OpenTelemetry tracer name string) to `v3`.
- `v3`'s `internal/cache/integration_test.go` lacks several v4-only tests added between the two backported commits' anchor points (symlink-target restore, single-file targets, home-directory overlap detection, unrecognized-format/corrupt-archive invalidation). Cherry-pick's 3-way merge had no matching context for those, so I took only the digest-verification test out of each conflict and left the unrelated v4-only tests out, rather than pulling them into this backport.
- No other v3-vs-v4 behavioral differences applied here — this change is internal to `internal/cache` and doesn't touch CLI/urfave code.

## Testing

- `go test ./internal/cache/...` passes.
- `go vet ./internal/cache/...` and `gofmt -l` are clean.
- Final diffstat matches upstream PR #4187 exactly (273 insertions, 1 deletion across `digest.go`, `digest_test.go`, `integration_test.go`, `restore.go`).

Closes nothing on its own; backport of #4187 (closes #4169 upstream).

### DIFF
--- a/internal/cache/digest.go
+++ b/internal/cache/digest.go
@@ -1,0 +1,85 @@
+package cache
+
+import (
+	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/buildkite/agent/v3/api"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/codes"
+)
+
+// ErrDigestMismatch means a downloaded blob's bytes don't hash to the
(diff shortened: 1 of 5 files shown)


jamiemonserrate (approved):
LGTM 🚀

PR #4197, opened 2026-08-07
Enable container-aware parallel Git checkout

lox:
## Description

Enable Git's parallel checkout for fresh worktree clones, while bounding its worker count to the agent's effective CPU parallelism:

```text
-v -c checkout.workers=<runtime.GOMAXPROCS(0)>
```

The repository builds with Go 1.25.12, whose default `GOMAXPROCS` is aware of Linux cgroup CPU bandwidth limits. Passing that positive value prevents Git's `checkout.workers=0` host-logical-CPU detection from over-parallelizing checkout in CPU-limited containers. This uses the Go runtime's existing behavior without bespoke cgroup parsing, an `automaxprocs` dependency, or changing the agent scheduler.

## Context

This revision addresses the cgroup correctness concern raised in review on #4197. Go 1.25's behavior is documented in the [release notes](https://go.dev/doc/go1.25#container-aware-gomaxprocs) and [Go blog](https://go.dev/blog/container-aware-gomaxprocs).

## Changes

- Fresh worktree clones persist a positive `checkout.workers` value derived from `runtime.GOMAXPROCS(0)`.
- Explicit `BUILDKITE_GIT_CLONE_FLAGS` and `--git-clone-flags` values still replace the default entirely.
- Bare mirror clone flags remain unchanged because mirror clones do not materialize a worktree.
- Integration subprocesses inherit the test process's effective `GOMAXPROCS`, keeping worker-count assertions deterministic.

On non-Linux and unconstrained systems, `GOMAXPROCS` safely resolves from the runtime's normal CPU/affinity behavior. Go uses cgroup CPU quota limits, not shares/weights (or Kubernetes CPU requests). Explicit `GOMAXPROCS` or `GODEBUG=containermaxprocs=0` configuration remains authoritative. The clone default is a process-start snapshot, so a later in-place quota resize does not rewrite it; Go may also floor its cgroup-derived default at two workers.

## Testing

Passed locally with Go 1.25.12:

- `go test ./clicommand -count=1`
- `go test ./internal/job/... -skip '^TestPolyglotScriptHooksCanBeRun$' -count=1`
- `GOMAXPROCS=2 go test ./clicommand -run '^TestGitCloneFlagsFlag$' -count=1`
- `GOMAXPROCS=2 go test ./internal/job/integration -run '<affected checkout tests>' -count=1`
- `golangci-lint run` (0 issues)
- `go tool gofumpt -extra -d .` (clean)

A full local run was attempted. The orb lacks Ruby for `TestPolyglotScriptHooksCanBeRun`; it also encountered the existing transient agent-shutdown pipe test failure on the first broad run, which passed when rerun as the full `clicommand` package. CI will provide the complete environment-backed run.

## Disclosures / Credits

Amp implemented and tested the update under Lachlan Donald's direction, including a separate expert review of the container/runtime behavior and test robustness.

### DIFF
--- a/clicommand/global.go
+++ b/clicommand/global.go
@@ -6,6 +6,7 @@ import (
 	"iter"
 	"os"
 	"reflect"
+	"runtime"
 	"strconv"
 	"time"
 
@@ -234,7 +235,7 @@ var (
 
 	GitCloneFlagsFlag = &cli.StringFlag{
 		Name:    "git-clone-flags",
-		Value:   "-v -c checkout.workers=0",
+		Value:   fmt.Sprintf("-v -c checkout.workers=%d", runtime.GOMAXPROCS(0)),
 		Usage:   "Flags to pass to \"git clone\" command",
 		Sources: cli.EnvVars("BUILDKITE_GIT_CLONE_FLAGS"),
 	}
--- a/clicommand/global.go
+++ b/clicommand/global.go
@@ -234,7 +234,7 @@ var (
 
 	GitCloneFlagsFlag = &cli.StringFlag{
 		Name:    "git-clone-flags",
-		Value:   "-v",
+		Value:   "-v -c checkout.workers=0",
 		Usage:   "Flags to pass to \"git clone\" command",
 		Sources: cli.EnvVars("BUILDKITE_GIT_CLONE_FLAGS"),
 	}
--- a/clicommand/global_test.go
+++ b/clicommand/global_test.go
@@ -2,6 +2,8 @@ package clicommand
 
 import (
 	"context"
+	"fmt"
+	"runtime"
 	"slices"
 	"testing"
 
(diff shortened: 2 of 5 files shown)


pda (approved):
Nice, especially the auto `0` → one per CPU/core/thread 👌🏼

DrJosh9000 (approved):

zhming0 (changes requested):
Per our chat in Slack, we have concerns over the correctness of this setting under cgroup and whether the tradeoff is worth it for small and mid size repo. 

I'll block this PR for now until we resolve the concerns. (sorry!)

lox:
Closing this for now. Parallel checkout can improve isolated checkout performance when CPU and storage have spare capacity, but we don’t have enough evidence that it is a safe universal default. On busy multi-job nodes, increasing per-checkout concurrency may worsen shared I/O contention, aggregate throughput, and tail latency; the isolated ~0.6s improvement does not justify that risk.

The investigation was still useful: Git’s automatic worker detection is not cgroup-aware, and Go 1.25’s GOMAXPROCS gives us a reasonable effective worker count when we explicitly enable this. Parallel checkout remains available to users through BUILDKITE_GIT_CLONE_FLAGS, and we can revisit a default after realistic node-level experiments across repository shapes, concurrency levels, and storage classes.

PR #4205, opened 2026-08-08
Run Windows batch hooks through cmd.exe

lox:
## Summary

- execute `.bat` and `.cmd` hook scripts explicitly through `cmd.exe` instead of passing the batch-file path directly to `CreateProcess`
- use a raw Windows command line with `/D /V:OFF /S /C` so cmd's quoting rules preserve hook paths containing spaces and metacharacters
- add a Windows regression test that runs a batch hook from a path containing spaces, `&`, and parentheses while stdout/stderr are copied through a non-file writer

## Why

`TestAgentStartupHookWithRegisteredAgentsEnv` has repeatedly flaked in Windows CI in two closely related ways:

- the batch hook returned success but neither of its two output lines reached the lifecycle-hook logger
- process creation failed with `The handle is invalid` while setting up the batch hook's redirected output handles

The lifecycle hook already closes its `io.Pipe` writer and joins the scanner goroutine before returning, and `exec.Cmd.Wait` waits for the internal output-copy goroutine. The missing lines therefore are not an assertion or scanner timing race.

The shared problem boundary is that `shell.Script` treated `.bat` and `.cmd` files as executables. Go passed the batch path to Windows process creation while also constructing inherited handles for the non-file lifecycle output writer. Batch files are command-processor input, and Go explicitly documents that `cmd.exe`/batch quoting does not follow its normal `CommandLineToArgvW` argument encoding. The fix makes that interpreter boundary explicit and carries the raw cmd command line through the existing process configuration before Windows installs its process-group attributes.

This changes production hook execution rather than serializing the test or adding sleeps/retries. It also applies to pre-bootstrap and wrapped job hooks that share `shell.Script`.

This PR is based independently on `main`; it does not modify or stack on #4197 or #4203.

## Validation

- `go test ./internal/shell ./internal/process ./clicommand`
- `go test -race ./internal/shell ./internal/process ./clicommand`
- focused lifecycle test: `go test ./clicommand -run '^TestAgentStartupHookWithRegisteredAgentsEnv$' -count=500`
- Windows cross-compilation of `internal/shell`, `internal/process`, and `clicommand` test binaries
- `golangci-lint run ./internal/shell ./internal/process ./clicommand`
- `go tool gofumpt -extra -w` on changed files
- `git diff --check`

## Remaining uncertainty

The orb cannot execute Windows binaries, so the new test and repeated Windows suite are the decisive checks in CI. A high-load Linux whole-package repetition also produced an existing lifecycle output-pipe error while many parallel subprocess tests were active; that indicates there may be a second, platform-independent process-pipe stress issue. It does not invalidate the unsupported direct-batch launch fixed here, but if Windows still flakes after this change that lower-level issue should be reduced separately rather than adding retries here.

### DIFF
--- a/internal/process/process.go
+++ b/internal/process/process.go
@@ -86,19 +86,20 @@ func ParseSignal(sig string) (Signal, error) {
 
 // Configuration for a Process
 type Config struct {
-	PTY               bool
-	Path              string
-	Args              []string
-	Env               []string
-	Stdin             io.Reader
-	Stdout            io.Writer
-	Stderr            io.Writer
-	Dir               string
-	Nice              int // Unix nice value for the child process (0 = default)
-	InterruptSignal   Signal
-	SignalGracePeriod time.Duration
-	Started           chan struct{}
-	Done              chan struct{}
+	PTY                bool
+	Path               string
+	Args               []string
+	WindowsCommandLine string
+	Env                []string
+	Stdin              io.Reader
+	Stdout             io.Writer
+	Stderr             io.Writer
+	Dir                string
+	Nice               int // Unix nice value for the child process (0 = default)
+	InterruptSignal    Signal
+	SignalGracePeriod  time.Duration
+	Started            chan struct{}
+	Done               chan struct{}
 }
 
 // Process is an operating system level process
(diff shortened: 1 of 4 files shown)


lox:
Closing because this investigation did not establish that direct batch execution was the root cause of the recurring lifecycle-hook flake, particularly the `ERROR_INVALID_HANDLE` failure. The underlying Windows batch-execution defect is real, but it is a separate product bug with broader compatibility impact. I’ll replace this with a product-focused PR that addresses interpreter resolution and path semantics and treats any flake improvement only as a possible secondary benefit.

PR #4239, opened 2026-08-12
Describe and retry undecodable API responses

Temikus:
## Description

An intermediary in front of the Agent API (proxy, load balancer, CDN) can answer a request with an HTML error page and a 2xx status. When that happened to us this morning, affected jobs failed on the first attempt with:

```
buildkite-agent: fatal: failed to decode JSON response: invalid character '<' looking for beginning of value
🚨 Error: Error setting up job executor: running "agent environment" shell hook: the agent environment hook exited with status 1
```

Two problems, both in `api/`:

1. The error doesn't name anything, so there is no way to tell an API fault from a broken hook or a bad IAM role without reproducing it under `--debug-http`.
2. A 2xx was read as "settled", so nothing retried. Looks like `doRequest` returns the `*Response` alongside the decode error, so `BreakOnNonRetryable` asked `IsRetryableStatus(200)`, got false, and broke immediately, so the agent's own retrier never fired.

The same response class reaches `Ping` and `Heartbeat`, where a non-retryable status means "stop the agent", so an undecodable body could terminate agents during the kind of blip they should ideally be able to recover from.

## Changes

- **`api/client.go`** - new `UndecodableResponseError` carrying the method, the final URL after redirects, the status, the content type, and the start of the body. Non-2xx responses (`checkResponse`) gain the same body detail, which was previously read and discarded.
- **`api/retryable.go`** - `IsRetryableError` returns true for an undecodable response, whatever status it arrived with, and `BreakOnNonRetryable` classifies the error before the response status. This also covers the hand-rolled retry loops in `core/client.go` and `internal/secrets/secret.go`.
- **`agent/agent_worker.go`, `agent/agent_worker_ping.go`** - an undecodable response is no longer an unrecoverable "stop the agent" signal.
- **`clicommand/oidc_request_token.go`** - fold the audience into the returned error, so the audience and the cause land on one line instead of two.

The same failure should now log something like:

```
WARN  POST https://agent.buildkite.com/v3/jobs/.../oidc/tokens: 200 OK: could not decode
response body (content-type "text/html"): invalid character '<' looking for beginning of
value: body starts with "<html><head><title>502 Bad Gateway</title></head></html>"
(Attempt 1/5 Retrying in 1s)
```

Some things I want your guidance on, as these are design decisions:

- **Quoting response bodies.** A body is quoted only when its content type is not JSON - my reasoning was - a malformed JSON body could contain things like tokens, so it stays out of the error. Quoted bodies are capped at 512 bytes.
- **Retrying what we could not decode** theoretically means we can retry a non-idempotent POSTs.
- **"not retrying: <reason>" log line** wherever `BreakOnNonRetryable` breaks. It would help but touches all 26 `BreakOnNonRetryable` call sites, so I'm not sure if it's worth it, happy to add if you think so.

## Testing

<details><summary><b>Verification</b></summary>

New tests in `api/undecodable_response_test.go`:
- the error names the request, status, content type, and body
- a malformed JSON body is reported but not quoted
- a non-2xx HTML body is reported, and an API JSON error message is unchanged
- quoted bodies are capped

New test in `clicommand/oidc_request_token_test.go`: `oidc request-token` against a server that returns an HTML 200 once and then a token. The command now succeeds and the server sees 2 requests. This test fails without the `api/retryable.go` change.

`go test ./...` is green apart from `TestUpdateGitMirrorCreatesFromRemoteMirrorAndKeepsCanonicalOrigin` in `internal/job`, which fails the same way on a clean checkout of `main` on macOS (`shared mirror mode = drwxr-xr-x, want setgid preserved`). Unrelated to this change.

`golangci-lint` was not run: it is not installed locally and is not a `go tool` in this repo. `go vet ./...` and `gofumpt -extra -l` are clean.
</details>

## Disclosures / Credits

Claude Code wrote the tests and the implementation under my direction and review, based on our incident investigation this morning.

### DIFF
--- a/agent/agent_worker.go
+++ b/agent/agent_worker.go
@@ -461,7 +461,9 @@ func (a *AgentWorker) Heartbeat(ctx context.Context) error {
 	beat, err := roko.DoFunc(ctx, r, func(r *roko.Retrier) (*api.Heartbeat, error) {
 		b, resp, err := a.apiClient.Heartbeat(ctx)
 		if err != nil {
-			if resp != nil && !api.IsRetryableStatus(resp) {
+			// A response we couldn't decode is not the API telling us to stop,
+			// whatever status it arrived with, so check the error as well.
+			if resp != nil && !api.IsRetryableStatus(resp) && !api.IsRetryableError(err) {
 				r.Break()
 				return nil, &errUnrecoverable{action: "Heartbeat", response: resp, err: err}
 			}
--- a/agent/agent_worker_ping.go
+++ b/agent/agent_worker_ping.go
@@ -223,7 +223,9 @@ func (a *AgentWorker) Ping(ctx context.Context) (jobID, action string, err error
 		// If the ping has a non-retryable status, we have to kill the agent, there's no way of recovering
 		// The reason we do this after the disconnect check is because the backend can (and does) send disconnect actions in
 		// responses with non-retryable statuses
-		if resp != nil && !api.IsRetryableStatus(resp) {
+		// A response we couldn't decode is not the API telling us to stop, whatever
+		// status it arrived with, so check the error as well
+		if resp != nil && !api.IsRetryableStatus(resp) && !api.IsRetryableError(pingErr) {
 			return "", action, &errUnrecoverable{action: "Ping", response: resp, err: pingErr}
 		}
 
--- a/api/client.go
+++ b/api/client.go
@@ -24,6 +24,9 @@ import (
 const (
 	defaultEndpoint  = "https://agent-edge.buildkite.com/v3"
 	defaultUserAgent = "buildkite-agent/api"
+
+	// Maximum bytes of a response body to quote in an error message.
+	maxErrorBodySnippet = 512
 )
 
 // Config is configuration for the API Client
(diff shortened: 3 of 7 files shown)


DrJosh9000:
Thanks @Temikus, we'll take a look

Temikus:
Thanks @DrJosh9000 - all feedback very welcome, happy to make any changes necessary 🤝

DrJosh9000:
Overall I think this is good. Here's some thoughts on what you asked:

> * **Quoting response bodies.** A body is quoted only when its content type is not JSON - my reasoning was - a malformed JSON body could contain things like tokens, so it stays out of the error. Quoted bodies are capped at 512 bytes.

That sounds sensible to me, though if the intermediary is outside of our control (collectively, Buildkite and the customer. Or say another incident occurs with our own intermediary layers and they happen to be misconfigured), I don't know if we can promise that tokens won't be leaked in non-JSON responses. Maybe the secret redactor is good enough in that case?

> * **Retrying what we could not decode** theoretically means we can retry a non-idempotent POSTs.

I'm not too worried - I don't think there are that many non-idempotent requests made by the agent.

> * **"not retrying: " log line** wherever `BreakOnNonRetryable` breaks. It would help but touches all 26 `BreakOnNonRetryable` call sites, so I'm not sure if it's worth it, happy to add if you think so.

I'm fine with doing this later if we decide it is worth it.

---

The only other thing I'll add is that (according to internal Slack chatter) the incident where this happened should only have affected agent versions prior to v3.97.0. I think this PR is worthwhile regardless.

DrJosh9000 (approved):

Temikus:
Amazing, thank you, Josh! Have a great weekend :) @DrJosh9000

PR #4251, opened 2026-08-18
PS-2073: Validate GitHub pull request merge heads

SorchaAbel:
## Description

GitHub can briefly serve a stale `refs/pull/<PR>/merge` after a force-push. When the agent resolves that ref with `BUILDKITE_COMMIT=HEAD`, it can therefore check out and persist a merge commit whose second parent is not the build’s expected pull request head.

Validate the fetched merge commit before checkout. If its direct second parent differs from `BUILDKITE_PULL_REQUEST_HEAD_COMMIT`, treat it as a fetch failure and retry through the existing checkout retry loop.

## Context

- [PS-2073](https://linear.app/buildkite/issue/PS-2073/3-nvidia-stale-merge-commit-reuse-in-pr-builds-using-github-test-merge)
- Companion Buildkite server change: [buildkite/buildkite#30177](https://github.com/buildkite/buildkite/pull/30177)

## Changes

- Capture `BUILDKITE_PULL_REQUEST_HEAD_COMMIT` during bootstrap.
- Keep the captured expected head immutable from environment hooks.
- After fetching a GitHub PR merge ref with `BUILDKITE_COMMIT=HEAD`, read the raw merge commit headers and compare its direct second-parent object ID with the expected head. This works with shallow fetches where the parent objects are intentionally absent.
- Return a retryable checkout fetch error when the values differ or the fetched object is not a valid merge commit.
- Preserve existing behaviour when the expected head is absent, the build uses an explicit commit, or checkout uses a custom/head refspec.

## Intentional scope

This change only prevents an unresolved GitHub merge ref from being accepted when it targets the wrong pull request head. It intentionally does not:

- Validate the merge commit’s first parent or merge base.
- Re-resolve or replace an explicit merge SHA already persisted for the build. Jobs and retries remain pinned to that SHA so a retry tests the same checkout as the original job.
- Change custom checkout hooks, custom refspecs, ordinary pull request head refspecs, or checkouts without the backend-provided expected head.
- Repair stale merge SHAs previously persisted by older agents.

## Testing

- [ ] Tests have run locally (with `go test ./...`).
- [x] Code is formatted (with `go tool gofumpt -extra -w .`).

Focused and race-enabled coverage passes:

    go test ./internal/job/integration \
      -run '^TestCheckingOutGitHubPullRequestMergeRefspec' \
      -count=1

    go test -race ./internal/job/integration \
      -run '^TestCheckingOutGitHubPullRequestMergeRefspec' \
      -count=1

The successful merge-ref integration path uses a depth-1 clone and fetch to cover shallow repositories. CLI wiring, executor configuration, checkout-path, and package-level tests also pass.

The complete `go test ./...` run encounters two unrelated existing macOS/full-parallel failures: setgid preservation in `TestUpdateGitMirrorCreatesFromRemoteMirrorAndKeepsCanonicalOrigin`, and an artifact test whose temporary HTTP server disappears under full parallel execution. Their focused tests pass.

## Deployment

The validation activates only when the backend supplies `BUILDKITE_PULL_REQUEST_HEAD_COMMIT`.

Older agents safely ignore the new environment variable but remain vulnerable to the stale-ref race. In a mixed fleet, only unresolved merge-ref checkouts running on updated agents receive protection. Once a merge SHA is persisted, later jobs and retries remain pinned to it.

## Rollback

Revert this change. There are no migrations or persistent agent-side state changes; checkout returns to accepting the merge ref without head validation.

## Disclosures / Credits

Amp implemented the tests and code under human direction following investigation of the customer escalation and existing checkout behaviour.


lox (approved):
Seems like a pragmatic approach to me 🚢

PR #4252, opened 2026-08-19
[A-1638] Dogfood Cache v2 in Agent CI

jamiemonserrate:
## Description

This PR dogfoods Cache v2 in the Agent repository's own CI. The design is centred on cache ownership and archive shape: cache the work that is expensive and reusable, while keeping each entry safe for its consumers.

### Design choices and trade-offs
- **Give each cache key one producer whose output matches its consumers.** 

  Lint analysis facts, compiled test objects, race-instrumented objects, cross-compiled target objects, and E2E objects have different shapes, so they do not share writable entries. This creates more keys and some duplicated data, but avoids save races and prevents an exact entry from being permanently frozen in the shape produced by the wrong job.
- **Prefer narrow, purpose-built caches over broad restores.** 
  * Cross-compiles use per-target build caches 
  * bk CLI caches the pinned executable
  * protobuf generation caches versioned tools
  * acknowledgements cache the generated target-specific file using mandatory target, tool, and module inputs. 

  This adds cache entries and invalidation plumbing, but avoids transferring unrelated Go state. 
- **Accept a narrow stale-attributions risk to preserve acknowledgements cache reuse.** 

  We are deliberately using `go.mod` as the source of truth for `go-licenses` rather than hashing every Go source file. The risk is the cache can omit a required notice when a source-only change starts importing a package from a module already in `go.mod` and that package requires licence attribution not already represented in the cached file. We consider the case unlikely: it requires a newly imported subpackage of an existing module to carry distinct attribution. We accept that limitation for this dogfood iteration in exchange for retaining useful cross-commit cache reuse.

## Context

- [A-1638: Dogfood on agent repo](https://linear.app/buildkite/issue/A-1638/dogfood-on-agent-repo)
- [Dogfood: Cache in Agent repo — jMe notes](https://app.notion.com/p/buildkite/Dogfood-Cache-in-Agent-repo-3bab8dbc2c89803cbbf2f08450194498?source=copy_link#3c1b8dbc2c8980fba04df193d3a14295)

## Disclosures / Credits

Jamie directed the design and dogfood measurements. Claude Opus contributed implementation and analysis during the cache-design iterations. Codex helped establish the baseline, prepare the branch for review, and draft this description.


.buildkite/cache.yml
```diff
@@ -0,0 +1,164 @@
+# Cache v2 definitions for the Agent's own CI.
+#
+# Every key ends at the dependency checksum and carries no per-commit part, so
```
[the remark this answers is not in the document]
jamiemonserrate: The cache registry is created with the following policy

```
# Entries are isolated by both branch and pipeline.
save:
  scopes: { branch: true, pipeline: true }
# Restore candidates are tried top to bottom: current branch, then main.
restore:
  scopes:
    - { branch: $current, pipeline: $current }
    - { branch: main, pipeline: $current }
# The rule grants access to both saving and restoring.
rules:
  - name: allow-save-and-restore
    effect: allow
    action: [save, restore]
```

So this should mean that PRs should not be able to write an entry to the main cache.

.buildkite/cache.yml
```diff
@@ -0,0 +1,166 @@
...
+    # under-partition. Every key part is mandatory because reusing an
+    # attribution file from a different go.mod or generator version could
+    # produce incomplete or outdated acknowledgements.
+    cache_key:
+      - agent-acknowledgements-v1
+      - { env: TARGET_GOOS }
+      - { env: TARGET_GOARCH }
+      - { env: TARGET_GO_VERSION }
+      - { checksum: .buildkite/go-licenses-version }
+      - { checksum: go.mod }
```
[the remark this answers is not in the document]
jamiemonserrate: Yup, you’re right that go.mod isn’t a complete representation of the import graph.
We’ve decided to accept this as a known limitation. For it to result in incomplete acknowledgements, a source-only change would need to import a previously unreachable package from a module already in go.mod, and that package would also need distinct licence attribution that isn’t already present. Adding a new module or changing its version still updates go.mod and invalidates the cache.
We expect that remaining case to be very rare, so I think this is a reasonable trade-off for this dogfood iteration. I’ve documented it in the cache definition and the PR description.

.buildkite/cache.yml
```diff
@@ -0,0 +1,173 @@
...
+# fallback_limit marks where matching may stop: parts after it are optional on
+# restore, so a dependency bump still lands on a compatible entry.
+#
+# One producer per key, and the producer must leave behind what the consumers
+# need. Exact keys make saves free but remove accumulation: once an entry
+# exists it is frozen until the checksum changes, so a key with two
+# differently-shaped producers keeps whichever one happened to win the first
+# build after a dependency bump, and no later build can add to it. Jobs that
+# compile this module but leave different artifacts behind therefore get
+# separate keys rather than sharing one.
```
moskyb: this (and many of the long prose comments in this PR) is super dense LLM-ese, and IMO makes it _harder_ to understand what the cache yaml is doing rather than easier. when i see a super long comment, my instinct is that it's important, but when it's hard to read it ends up being confusing.

i'd really appreciate it if you went through the comments on this PR and rewrote (or removed) them, using your judgement as to whether or not they're useful for the flow of the PR

jamiemonserrate: that's fair. I'll take a pass at humanising the comments / description.

buildkate: I'd lean towards removing this whole comment block from 2 to 22 inclusive.  It is a lot of info about how cache works in general vs being actual specific things about _this_ cache.

.buildkite/cache.yml
```diff
@@ -0,0 +1,120 @@
+# Cache v2 definitions for the Agent's own CI.
```
buildkate:
```suggestion
# Cache definitions for the Agent's own CI.
```

.buildkite/steps/test-bk.sh
```diff
@@ -1,9 +1,27 @@
...
-echo "--- :package: Downloading bk binary"
-go install github.com/buildkite/cli/v2/cmd/bk@cdcc5fa4b6e209f5ffa79469dad04938d6eed0cd
+mkdir -p "$GOBIN"
+
+echo --- :inbox_tray: Restoring bk CLI
+buildkite-agent cache restore --name bk_cli
+
+if [[ ! -x "$GOBIN/bk" ]]; then
+	echo "--- :package: Installing bk CLI"
+	go install "github.com/buildkite/cli/v2/cmd/bk@${BK_CLI_COMMIT}"
```
moskyb: [non-blocking]

😱 we're still using v2 of the cli? at time of writing we're on [v3.55](https://github.com/buildkite/cli/releases/tag/v3.55.0)

don't fix this as part of this PR

.buildkite/cache.yml
```diff
@@ -0,0 +1,120 @@
...
+      - agent-go-build-race-v2
+      - { agent: os }
+      - { agent: arch }
+      - { env: AGENT_GO_VERSION, fallback_limit: true }
+      - { checksum: go.mod }
+    target_paths:
+      - ~/.gocache_race
+
+  - name: lint_gocache
+    # Lint gets its own cache because no other job writes its analyzer results.
```
moskyb, line 38: :chefskiss:

.buildkite/pipeline.yml
```diff
@@ -17,25 +19,66 @@ steps:
       - go.{mod,sum}
       - "**.go"
       - .buildkite/steps/check-code-committed.sh
+      - .buildkite/cache.yml
+      - .buildkite/docker-compose.yml
     plugins:
+      - &cache_v2_role
```
moskyb:
```suggestion
      - &assume_cache_role
```

.buildkite/pipeline.yml
```diff
@@ -17,25 +19,66 @@ steps:
...
+    env:
+      GOBIN: /root/.protobuf-tools/bin
+      GOCACHE: /root/.protobuf-tools/gocache
+      GOMODCACHE: /root/.protobuf-tools/gomodcache
+    command:
+      - . .buildkite/steps/protobuf-tool-versions.sh
+      - 'export PROTOBUF_GO_VERSION="$(go env GOVERSION | cut -d. -f1,2)"'
+      - buildkite-agent cache restore --name protobuf_tools
+      - .buildkite/steps/check-protobuf-generation.sh
+      - buildkite-agent cache save --name protobuf_tools
```
moskyb: given that this is the only(?) step with a multiline command, i wonder if it's worth integrating all of this into `check-protobuf-generation.sh`

.buildkite/cache.yml
```diff
@@ -0,0 +1,120 @@
...
+      - agent-go-modules-v1
+      - { agent: os }
+      - { agent: arch }
+      - { env: AGENT_GO_VERSION, fallback_limit: true }
+      - { checksum: go.mod }
+    target_paths:
+      - ~/.gomodcache
+
+  - name: gocache
+    # Shard 0 is the only writer for each platform.
```
zhming0: what is a shard? 👀

.buildkite/cache.yml
```diff
@@ -0,0 +1,120 @@
...
+      - { agent: arch }
+      - { env: AGENT_GO_VERSION, fallback_limit: true }
+      - { checksum: go.mod }
+    target_paths:
+      - ~/.gocache_race
+
+  - name: lint_gocache
+    # Lint gets its own cache because no other job writes its analyzer results.
+    cache_key:
+      - agent-go-build-lint-v1
```
zhming0: I sometime use this trick: `- { env: BUILDKITE_STEP_KEY }` but perhaps static string is better 👍🏿

zhming0: But I am curious behind the use of `v1` and `v2` here

.buildkite/cache.yml
```diff
@@ -0,0 +1,120 @@
...
+
+  - name: protobuf_tools
+    cache_key:
+      - protobuf-tools-v1
+      - { agent: os }
+      - { agent: arch }
+      - { env: PROTOBUF_GO_VERSION }
+      - { env: PROTOBUF_BUF_VERSION }
+      - { env: PROTOBUF_PROTOC_GEN_GO_VERSION }
+      - { env: PROTOBUF_PROTOC_GEN_CONNECT_GO_VERSION }
```
zhming0, line 107: These makes me think how we can improve the experience 🤔

zhming0 (commented):
left some questions 🤩

.buildkite/pipeline.yml
```diff
@@ -98,7 +144,16 @@ steps:
...
               propagate-environment: true
+              mount-buildkite-agent: true
               run: agent
+              env:
+                - BUILDKITE_AGENT_CACHE_REGISTRY
+                - BUILDKITE_AGENT_CACHE_STORE_URL
+                - AWS_ACCESS_KEY_ID
+                - AWS_SECRET_ACCESS_KEY
+                - AWS_SESSION_TOKEN
+                - BUILDKITE_BRANCH
```
buildkate: Are we actually using `BUILDKITE_BRANCH` and `BUILDKITE_COMMIT` in the cache keys?   Saves us 8 lines in the pipeline file if we can remove them.

.buildkite/steps/test-bk.sh
jamiemonserrate: Good point. I didn't even notice 😅 

Created https://linear.app/buildkite/issue/A-1760/bump-bk-cli-we-use-to-test-the-agent to track this

.buildkite/cache.yml
jamiemonserrate: Good point. It means the first step of the multi step job (I think that's the terminology). But I can rewrite this comment without using this terminology.

.buildkite/cache.yml
jamiemonserrate: Good point again. This was me going through multiple iterations of the cache key. And bumping the version when i changed what was cached. Its superfluous now. Will delete.

.buildkite/cache.yml
jamiemonserrate: Its a good question - and I don't have a good answer. I feel these are inputs we want from the user. But I would need to do some research to form an opinion on how users are used to telling us how to invalidate a cache.

.buildkite/pipeline.yml
jamiemonserrate: Good catch. Let me validate.

.buildkite/cache.yml
```diff
@@ -0,0 +1,110 @@
...
+      - agent-acknowledgements
+      - { env: TARGET_GOOS }
+      - { env: TARGET_GOARCH }
+      - { env: TARGET_GO_VERSION }
+      - { checksum: .buildkite/go-licenses-version }
+      - { checksum: go.mod }
+    target_paths:
+      - clicommand/ACKNOWLEDGEMENTS.md.gz
+
+  - name: e2e_gomodcache
```
ss1909, line 77: Suggestion: This and gomodcache can be clubbed cause they both resolve to the same go.mod version

.buildkite/cache.yml
```diff
@@ -0,0 +1,110 @@
...
+    # The host is part of the key because this also contains host-built tools.
+    cache_key:
+      - agent-go-target-build
+      - { agent: os }
+      - { agent: arch }
+      - { env: TARGET_GOOS }
+      - { env: TARGET_GOARCH }
+      - { env: TARGET_GOARM }
+      - { env: TARGET_BUILD_MODE }
+      - { env: TARGET_GO_VERSION, fallback_limit: true }
```
ss1909, line 58: Question: do we want to be so strict with the fallback implementation for all of these cache entries?

.buildkite/cache.yml
jamiemonserrate: This was an explicit choice. Otherwise the cache ends up too big, and the gains of caching are negated by the size of the download. 

Its a good catch, and I should add a comment about this.

.buildkite/cache.yml
jamiemonserrate: in the sense? Each architecture and go version will generate different artifacts - so I'm thinking its worth keeping it separate. What were you thinking?

ss1909: Sorry, wasn't clear - I meant the fallback_limit on the Go-version part specifically. 

Not a biggie, just curious.

ss1909 (approved):

PR #4256, opened 2026-08-23
[A-1756] Apply request headers from streaming pings

moskyb:
## Description

Ping-stream responses can now include updated server-specified request headers. Without applying these updates, a streaming agent may continue using stale routing information for subsequent Agent API requests.

This change applies headers received from the stream and reconnects when the effective header set changes, ensuring the next streaming request includes them. Any action accompanying the update is delivered before reconnecting.

Polling pings already apply server-specified request headers. This brings streaming pings in line with that behaviour, including distinguishing between:

- absent headers, which leave the current headers unchanged; and
- an explicitly empty header set, which clears them.

## Context

https://github.com/buildkite/buildkite/pull/32972
INC-1228
A-1756

## Changes

- Update the protobuf definitions for streaming pings to include request headers to be reflected
- Restart the ping stream when its effective request headers change.
- Synchronise access to request headers shared by concurrent API requests.
- Return header snapshots rather than exposing mutable internal state.
- Cover adding, retaining, and clearing headers across stream connections.

## Testing

- [x] Tests have run locally (with `go test ./...`). Relevant packages pass with `go test ./api ./agent ./agent/integration`; the full suite currently has an unrelated failure in `internal/job`.

## Disclosures / Credits

Me + Amp

### DIFF
--- a/agent/integration/job_environment_integration_test.go
+++ b/agent/integration/job_environment_integration_test.go
@@ -183,23 +183,19 @@ func TestNoCacheSettings_DoesNotLogMessage(t *testing.T) {
 }
 
 func TestBuildkiteRequestHeaders(t *testing.T) {
-	t.Parallel()
-
 	// create a mock agent API
 	e := createTestAgentEndpoint()
 	server := e.server()
 	defer server.Close()
 
 	// create a client with server-specified headers
+	t.Setenv("BUILDKITE_REQUEST_HEADER_BUILDKITE_HELLO", "world")
 	l := logger.NewConsoleLogger(logger.NewTestPrinter(t), func(int) {})
 	client := api.NewClient(l, api.Config{
 		Endpoint:  server.URL,
 		Token:     "llamasrock",
 		DebugHTTP: true,
 	})
-	headers := client.ServerSpecifiedRequestHeaders()
-	// That getter isn't designed to modify the headers, but all's fair in test setup code and war.
-	headers.Set("Buildkite-Hello", "world")
 
 	mb := mockBootstrap(t)
 	defer mb.CheckAndClose(t) //nolint:errcheck // bintest logs to t
--- a/api/pings_streaming.go
+++ b/api/pings_streaming.go
@@ -38,7 +38,7 @@ func (c *Client) StreamPings(ctx context.Context, agentID string, opts ...connec
 	h := callInfo.RequestHeader()
 
 	// Add any request headers specified by the server during register/ping
-	for k, values := range c.requestHeaders {
+	for k, values := range c.requestHeadersSnapshot() {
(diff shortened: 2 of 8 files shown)


blaknite (approved):

JuanitoFatas (approved):
🚢

agent/integration/job_environment_integration_test.go
zhming0, line 1: [not blocking] I am not following what these diff are about 🤔

api/pings_streaming.go
```diff
@@ -61,7 +61,22 @@ func (c *Client) StreamPings(ctx context.Context, agentID string, opts ...connec
...
+				}
+				restart = c.setRequestHeaders(values)
+			}
+
+			// Always yield the action before restarting so that a message
+			// containing both an action and new headers doesn't lose the action.
+			if !yield(msg, nil) {
+				return
+			}
+			if restart {
```
zhming0: Please correct me if I am wrong: reading this, it feels to me that it would cause the stream to always restart after the initial server-side ping (when there is cache-miss), regardless if the shard need to change or not?

zhming0 (changes requested):
Due to the gravity of this change, leaving a blocking comment so we can clarify first 🙏🏿

agent/integration/job_environment_integration_test.go
moskyb: this test was doing crimes by modifying the request headers in place. the `api.Client.ServerSpecifiedRequestHeaders()` method now returns a snapshot of the headers, so modifying it in place will no longer have an effect.

that said, i think setting this through an envar is pretty yuck. i'll make it so we can pass an initial set through the `api.Config` struct instead.

api/pings_streaming.go
moskyb: the `restart` return value of `c.setRequestHeaders` is only true if there's been a change in effective header set — so if my headers are
```
Buildkite-Mountain: Cotopaxi
Buildkite-River: Guayas
```

and the ping returns those headers, `restart` will be false.

if it returns
```
Buildkite-Mountain: Chimborazo
```

then it'll be true, and we'll restart the stream

zhming0: ah cool, thanks for clarifying! 👍

zhming0 (approved):

agent/integration/job_environment_integration_test.go
moskyb: actually, doing this turned out to be a bit of a rabbit hole. i'm going to leave it for another time.

api/pings_streaming.go
moskyb: i've changed the name of this variable to be `headersDidChange`, which should make things clearer

PR #4269, opened 2026-08-25
Automatically acquire jobs from JAT tokens

swebb:
## Description

Recognize `bkjat_` Job Acquisition Tokens and automatically enter acquire mode using the job UUID from the JWT `sub` claim. The token is decoded locally only to select acquire mode; Buildkite still validates its signature and authorization during registration.

An explicitly configured `--acquire-job` remains supported, but must match the JAT's job UUID.

Fixes: A-1697

## Context

- https://github.com/buildkite/buildkite/pull/32347

## Changes

- Decode and validate the job UUID from JAT payloads.
- Configure acquire mode automatically after resolving direct or indirect registration tokens.
- Reject malformed JATs and conflicting explicit job UUIDs.
- Cover regular tokens, valid JATs, malformed JATs, and conflicting acquisition settings.

## Testing

Additional focused validation:
- `go test ./clicommand -count=1`
- `go vet ./clicommand`

Buildkite

## Disclosures / Credits

Amp implemented and tested this change under my direction.

### DIFF
--- a/clicommand/register_token.go
+++ b/clicommand/register_token.go
@@ -1,11 +1,15 @@
 package clicommand
 
 import (
+	"encoding/base64"
+	"encoding/json"
 	"fmt"
 	"io"
 	"os"
 	"strconv"
 	"strings"
+
+	"github.com/google/uuid"
 )
 
 // The agent registration token is a long-lived secret, so we try to avoid
@@ -26,7 +30,10 @@ import (
 // fd://N instead (see reexecToScrubRegistrationToken), which replaces the
 // exec-time cmdline and environ with token-free versions.
 
-const registrationTokenEnvVar = "BUILDKITE_AGENT_TOKEN"
+const (
+	registrationTokenEnvVar   = "BUILDKITE_AGENT_TOKEN"
+	jobAcquisitionTokenPrefix = "bkjat_"
+)
 
 // maxTokenFileSize is a sanity limit when reading tokens from files or file
 // descriptors.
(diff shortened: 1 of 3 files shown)


clicommand/register_token.go
```diff
@@ -76,6 +83,37 @@ func resolveRegistrationToken(token string) (string, error) {
...
+		return "", true, fmt.Errorf("expected a three-part JWT")
+	}
+
+	payload, err := base64.RawURLEncoding.DecodeString(parts[1])
+	if err != nil {
+		return "", true, fmt.Errorf("couldn't decode JWT payload: %w", err)
+	}
+
+	var claims struct {
+		Subject string `json:"sub"`
```
blaknite, line 105: Non-blocking: could we also decode and validate the JWT’s `typ` field before using `sub` as a job UUID? `bkjat_` is a useful routing hint, but checking the token type here would ensure a UUID-shaped subject from another token type cannot switch the agent into acquire mode.

blaknite (approved):

clicommand/register_token.go
```diff
@@ -76,6 +83,37 @@ func resolveRegistrationToken(token string) (string, error) {
 	}
 }
 
+// jobAcquisitionTokenJobUUID extracts the job UUID from a Job Acquisition
+// Token. The token's signature is validated by Buildkite during registration;
+// decoding it here only selects the agent's acquisition mode.
+func jobAcquisitionTokenJobUUID(token string) (string, bool, error) {
```
moskyb, line 89: i think we shouldn't be pulling the acquisition job ID out of the token, because it means that the token is no longer opaque, and we can never change the shape of the JWT, or make it not a JWT (idk why we'd want to do that, but we might). Additionally, the value of the job in the JWT is inherently untrustworthy, because we aren't (and can't) verify the signature of the token.

if someone gave us a criminal token and changed the id in the JWT, the agent still wouldn't be able to acquire that job because of how jatz work, but i think it's still important to have defence in depth here.

i think my preferred way of implementing this would be something like:
- When the agent is launched with a JAT (as decided by the token prefix), it always registers itself excluded from dispatch
- When an agent registers with buildkite using a JAT, the backend returns the `job_acquisition_job_id` in the register response
- From there, we switch the agent into acquire mode

i'm not 100% hard against decoding the JWT, because the backend will protect us from most of the ways that it could go wrong, so i'm happy to disagree and commit here, but i also want to be able to iterate on token structure etc if we need to

clicommand/register_token.go
```diff
@@ -76,6 +83,37 @@ func resolveRegistrationToken(token string) (string, error) {
...
+	if !strings.HasPrefix(token, jobAcquisitionTokenPrefix) {
+		return "", false, nil
+	}
+
+	parts := strings.Split(strings.TrimPrefix(token, jobAcquisitionTokenPrefix), ".")
```
moskyb, line 94: if we're decoding the jwt (see above), we definitely shouldn't be doing it by hand this way. you should be able to use the `github.com/lestrrat-go/jwx/jwt` package for safe JWT decoding

moskyb (changes requested):

swebb:
I'm going to close this for the time being as we're short on time and don't have a consensus. It feels like bit of a one way door as it locks us into a JWT format (yes we can change the format, but we still need to support any existing agent versions for 12 months. As this is a stretch goal let's not commit ourselves at this time.

PR #4270, opened 2026-08-25
Report graceful agent stops to Buildkite

scadu:
## Description

Graceful shutdowns currently drain the worker locally, but the control plane shows the agent as `CONNECTED` until it disconnects.

This reports a non-forced `/stop` using the worker’s existing session. The agent moves to `STOPPING` while jobs drain, then `STOPPED` after disconnecting.

Reporting is best-effort and limited to five seconds. Ungraceful shutdown paths, including standard agent-stack-k8s termination, remain unchanged.

I considered doing this in the scaler or through the Stacks API, but either option would need broader credentials and agent-to-instance mapping. The worker already has everything it needs.

## Context

https://github.com/buildkite/elastic-ci-stack-for-aws/issues/1861

## Changes

- Report graceful stops exactly once before disconnecting.
- Wait for the stop report before sending `/disconnect`.
- Keep reporting failures from blocking shutdown for longer than five seconds.
- Cover session authentication, request payload, ordering, failures, and non-graceful shutdown paths.

## Testing

Also tested the lifecycle on Elastic Stack with three workers: all moved to `STOPPING` while their jobs continued, then reached `STOPPED` after finishing.


clicommand/agent_start.go
```diff
@@ -413,6 +414,12 @@ var AgentStartCommand = &cli.Command{
 			Usage:   "The maximum uptime in seconds before the agent stops accepting new jobs and shuts down after any running jobs complete. The default of 0 means no timeout",
 			Sources: cli.EnvVars("BUILDKITE_AGENT_DISCONNECT_AFTER_UPTIME"),
 		},
+		&cli.BoolFlag{
+			Name:    "report-graceful-stop",
+			Value:   true,
+			Usage:   "Report graceful stops to Buildkite before disconnecting (default: true)",
+			Sources: cli.EnvVars("BUILDKITE_AGENT_REPORT_GRACEFUL_STOP"),
+		},
```
moskyb: if we're pretty sure that this won't affect behaviour in the k8s stack (and i'm pretty sure it won't, the k8s stack [always skips graceful terminations](https://github.com/buildkite/agent/blob/main/clicommand/agent_start.go#L1339), i'm inclined to make this not configurable and just have it always be on.

moskyb (commented):
LGTM, but i think we can remove the configurability.

additionally, once this is ready to to go, can you open a backport PR to ship this change to v3? you should be able to ask your agent of choice to do it by asking, and it'll read [cherrypicking-pr-to-v3.md](https://github.com/buildkite/agent/blob/main/.agents/skills/cherrypicking-pr-to-v3/SKILL.md).

you are, of course, also welcome to do it by hand.

clicommand/agent_start.go
scadu: Done in 3f9a86e

moskyb (approved):
love your work!

PR #4271, opened 2026-08-26
[A-1755] Die cleanly when heartbeats fail

moskyb:
## Description

In `ping-mode=auto`, an unrecoverable heartbeat error could leave the agent process running indefinitely instead of exiting.

The heartbeat and four worker loops (count 'em, five messages) reported their results into a four-slot channel. If the heartbeat occupied a slot before the worker loops stopped, one worker loop blocked while sending its result. `Start` then waited forever for that loop before it could drain the channel.

This removes the shared result channel. Each worker loop now records its error independently, while heartbeat shutdown is explicitly cancelled and awaited on every return path.

## Context

Discovered while investigating agents that remained alive after the Agent API returned 404 responses during a network partition.

A-1755
INC-1230

## Changes

- Remove the fixed-capacity error channel from agent worker shutdown.
- Collect each worker loop's error without a blocking channel send.
- Always cancel and await the heartbeat loop before `Start` returns.
- Add regression coverage confirming an agent using `ping-mode=auto` exits after an unrecoverable heartbeat 404.

## Testing

## Disclosures / Credits

Me + Amp

### DIFF
--- a/agent/agent_worker.go
+++ b/agent/agent_worker.go
@@ -268,16 +268,20 @@ func (a *AgentWorker) Start(ctx context.Context, idleMon *idleMonitor) (startErr
 	}
 	defer a.metricsCollector.Stop() //nolint:errcheck // Best-effort cleanup
 
-	// There are as many as 4 different loops that send 1 error here each.
-	errCh := make(chan error, 4)
-
 	// Use this context to control the heartbeat loop.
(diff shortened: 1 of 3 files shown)


ss1909:
Makes sense! LGTM 🚀

ss1909 (approved):

PR #4276, opened 2026-08-27
Skip mirrors for relative submodule URLs

davido:
Submodule URLs from .gitmodules may be relative to the superproject remote. The mirror checkout path currently reads those raw values and runs from the mirror cache directory, where ../java-prettify is treated as a local filesystem path instead of resolving against the superproject remote.

Skip only the mirror optimization for relative submodule URLs and let git submodule update resolve them with Git's normal semantics. Absolute submodule URLs still use mirrors.

Add regression tests that fail on the current behavior by observing the attempted mirror clone of ../java-prettify, and pass once that relative URL falls back to the normal submodule update path.

Refs bazelbuild/continuous-integration#2815.

## Testing


swebb:
Thanks for opening this PR @davido. I’ve created a ticket to track it internally.

buildkate:
Closing in favour of #4287

PR #4285, opened 2026-08-28
Add --git-fetch-base-branch to fetch the base branch at checkout

### LINKED ISSUE #4284: Option to fetch the base branch during checkout, not just for pipeline upload's if_changed diff
## Is your feature request related to a problem? Please describe.

A job's checkout fetches `refs/pull/N/head` plus the commit (or the commit alone for a branch build). None of those update a branch's remote-tracking ref, so `refs/remotes/origin/<base>` in a reused checkout directory holds whatever an earlier build left there — stale by an unbounded amount, with nothing in the log to say so.

`--fetch-diff-base` / `BUILDKITE_FETCH_DIFF_BASE` solves this for `buildkite-agent pipeline upload`'s own `if_changed` diff. Anything else in the job that asks what changed — a pipeline generator, a test selector, a script running `git merge-base HEAD origin/main` — still reads the stale ref. A stale base only ever widens a diff, so nothing fails and nobody notices.

The fix today is per step: a plugin, or an explicit `git fetch` in the command. Every new step has to remember it, and a step that forgot looks identical to one that didn't.

## Describe the solution you'd like

An agent-level option, off by default, that fetches the base branch during checkout so every job in the build starts from its current tip: `--git-fetch-base-branch` / `BUILDKITE_GIT_FETCH_BASE_BRANCH`, on `agent start` and `bootstrap`.

## Describe alternatives you've considered

- `BUILDKITE_GIT_FETCH_FLAGS`: flags land before `--`, so it cannot add a refspec.
- `BUILDKITE_REFSPEC`: replaces the PR fetch rather than adding to it.
- `--fetch-diff-base`: scoped to `pipeline upload`, and only to the diff it computes itself.

## Affiliation (optional)

Altana.

## Additional context

Happy to send a PR — I have one ready.

jasonwbarnett:
## Description

A checkout fetches `refs/pull/N/head` plus the commit, or just the commit for a branch build. Neither updates a branch's remote-tracking ref, so `refs/remotes/origin/<base>` in a reused checkout directory holds whatever an earlier build left there. `--fetch-diff-base` covers `pipeline upload`'s own `if_changed` diff, but anything else in the job that asks what changed — a pipeline generator, a test selector, a script running `git merge-base HEAD origin/main` — still reads the stale ref. A stale base only widens a diff, so nothing fails and nobody notices.

`--git-fetch-base-branch` fetches the base branch during checkout, so every job starts from its current tip. It takes a mode, following `BUILDKITE_GIT_COMMIT_VERIFICATION`:

| Mode | Behaviour |
| --- | --- |
| `off` (default) | No extra fetch. |
| `optimistic` | Best-effort: rides out a brief blip, then warns without failing the job. |
| `strict` | Retries as hard as the job's own fetch, and fails the checkout when the ref doesn't land. Guaranteeing freshness requires failing closed. |

The base branch is the first non-empty value of `BUILDKITE_PULL_REQUEST_BASE_BRANCH` or `BUILDKITE_PIPELINE_DEFAULT_BRANCH` — the precedence `pipeline upload` uses for `--git-diff-base`, minus its final literal `main`, since guessing would mean a failed fetch on every repo whose default branch is named something else. Building the base branch itself is a skip in every mode. Under `strict`, a job that names no base branch at all fails too: skipping quietly there leaves the job diffing against whatever an earlier build left behind, which is what the mode exists to prevent.

It's on `agent start` as well as `bootstrap`, so a fleet can turn it on once, and it's in `env/protected.go` so a pipeline can't turn it back off.

Alternatives: `BUILDKITE_GIT_FETCH_FLAGS` can't carry a refspec (flags land before `--`); `BUILDKITE_REFSPEC` replaces the PR fetch rather than adding to it; a per-step plugin is what we do today, and it's an opt-in every new step has to remember.

Six things are deliberate, each pinned by a test:

- **A separate fetch, not an extra refspec on the job's own.** `rev-parse FETCH_HEAD` resolves to the first ref fetched, and the checkout of the job's commit depends on that, so the base branch fetch must not leave `FETCH_HEAD` on the base tip.
- **An explicit, forced refspec** (`+refs/heads/<base>:refs/remotes/origin/<base>`), so the ref lands regardless of `remote.origin.fetch` and follows a force-push instead of stopping at a non-fast-forward rejection.
- **The refspec is passed to git verbatim,** via a new `gitFetchArgs.LiteralRefSpecs`. `gitFetch` word-splits `RefSpecs` because that field carries operator-supplied strings which may hold several refspecs, but a refspec the agent builds from a ref name must not be split: a quote is legal in a git ref name, so `release'candidate` becomes `releasecandidate` — a fetch that succeeds against a different branch wherever that one exists, leaving the intended ref stale. Quoting the refspec instead would only fix posix; `shellwords.QuoteBatch` escapes neither quote character while `SplitBatch` treats both as quotes, so the round trip drops them on Windows.
- **It runs before the `GitSkipFetchExistingCommits` early return.** A job that asked for the base branch asked for it either way.
- **Retried, but not on a ref the remote doesn't have, and not equally in both modes.** An outage at the remote is the usual reason this fails in CI, and one attempt would silently leave the stale ref in place under `optimistic`, so both modes retry. They differ in how long that's worth: `strict` gets the same budget as the job's own fetch (10 attempts, ~90s) because the alternative is failing the job, while `optimistic` has already decided the job proceeds either way and takes 3 (~2s). A remote answering "couldn't find remote ref" ends the retries immediately in both, so a deleted base branch costs one fetch. That's also why this can't simply set `Retry` on `gitFetch`: that retrier deliberately keeps retrying a missing ref, to wait out GitHub creating `refs/pull/N/head` asynchronously, and a base branch is either already there or gone for good. Under `strict` the exhausted error is returned and the checkout's own retrier gets a further go; a job naming no base branch breaks that retrier instead, since retrying can't discover one.
- **`strict` rejects a `BUILDKITE_GIT_FETCH_FLAGS` that writes no ref.** git exits zero for `--dry-run`, `--prefetch` and `--negotiate-only`, so any of them would leave `strict` promising a current base branch and delivering whatever the checkout directory already held. Those flags are the operator's configuration for every fetch in the job and `strict` can't honour both them and its own guarantee, so it reports the conflict rather than stripping a `--dry-run` that explicitly asked for nothing to be written. `optimistic` promises nothing about the ref, so the flags stand there. The flag matching that commit verification already does for these three is shared rather than duplicated.

## Context

Fixes #4284.

## Changes

```
$ buildkite-agent start --help
  --git-fetch-base-branch string                                                     Also fetch the base branch during checkout, so every command in the job can diff against its current tip; one of [strict optimistic off]. ′optimistic′ warns and carries on when the fetch fails, ′strict′ fails the job, and ′off′ (default) skips the fetch. A transient failure is retried briefly under ′optimistic′ and on the same budget as the job's own fetch under ′strict′; a base branch the remote does not have fails without retrying. The base branch is the first non-empty value of {$BUILDKITE_PULL_REQUEST_BASE_BRANCH, $BUILDKITE_PIPELINE_DEFAULT_BRANCH}; nothing is fetched when it is the branch being built, and under ′strict′ a job with neither set fails. ′strict′ also rejects a $BUILDKITE_GIT_FETCH_FLAGS that asks for a fetch writing no ref (--dry-run, --prefetch, --negotiate-only), which git reports as success (default: "off") [$BUILDKITE_GIT_FETCH_BASE_BRANCH]
```

- `internal/job/checkout_fetch_base_branch.go`: the modes, `baseBranchToFetch`, `fetchBaseBranch`.
- `internal/job/checkout_fetch.go`: one new span attribute (`git.fetch_base_branch`) and the call from `fetchSource`.
- `internal/job/git.go`: `gitFetchArgs.LiteralRefSpecs`, for refspecs that must reach git unsplit.
- `internal/job/checkout.go`: fail fast on an invalid mode, alongside the existing sparse-mode check; break the retrier on the two `strict` failures retrying can't change.
- `internal/job/commit_verification.go`: its ref-suppressing flag check is now a shared predicate, so verification strips those flags and the base branch fetch reports them.
- `internal/job/config.go`, `clicommand/{global,bootstrap,agent_start}.go`, `agent/{agent_configuration,job_runner}.go`: the flag, validated like `--git-commit-verification` and wired like `--git-skip-fetch-existing-commits` in both checkout-override modes.
- `env/protected.go`: `BUILDKITE_GIT_FETCH_BASE_BRANCH`.
- Tests: 25 cases in `internal/job/checkout_fetch_base_branch_test.go` against a real `githttptest` remote, plus env precedence and flag validation. `githttptest` grows a `FailNextRequests` knob so a test can watch a fetch survive a remote that is briefly unavailable.
- `docs/remote-git-mirrors.md`: one line in §4.1. Drop that hunk if you'd rather own that doc.

## Public documentation
**Status:** publish — the flag reference lives in `buildkite/docs`; happy to open a PR there once this lands.

## Testing

`go test ./...` is clean — 60 packages, 38 of them with tests — and `golangci-lint` reports 0 issues on the touched packages.

Each behaviour was checked by breaking it and confirming a test catches it: removing the call, moving it after the skip-fetch return, reordering it after the job's own fetch, dropping the built-branch guard, dropping the default-branch fallback, using a plain refspec, dropping the leading `+`, putting the refspec back through the word-splitting `RefSpecs`, making `strict`'s fetch failure non-fatal, letting `strict` tolerate a job with no base branch, letting `strict` accept ref-suppressing fetch flags, dropping the retrier, treating a missing ref as retryable, and swapping the two modes' retry budgets in each direction.

Altana.

## Disclosures / Credits

Claude Code (Opus) wrote the implementation and the tests and drafted this description, working from my problem statement and reviewing decisions with me as it went.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01EMT5Lt9633Ny8a3QQZ4hni

### DIFF
--- a/internal/job/checkout_fetch.go
+++ b/internal/job/checkout_fetch.go
@@ -86,16 +86,25 @@ func (e *Executor) fetchSource(ctx context.Context, addBloblessFilter bool, atte
 	skipFetch := (e.GitSkipFetchExistingCommits || mirrorHit) && e.Commit != "HEAD" &&
 		hasGitCommit(ctx, e.shell, ".git", e.Commit)
 
-	span.SetAttributes(attribute.Bool("git.skipped", skipFetch))
-
-	if skipFetch {
-		e.shell.Commentf("Commit %q already exists locally, skipping fetch", e.Commit)
-		return nil
-	}
+	span.SetAttributes(
+		attribute.Bool("git.skipped", skipFetch),
+		attribute.Bool("git.fetch_base_branch", e.GitFetchBaseBranch),
+	)
 
 	gitFetchFlags := e.GitFetchFlags
 	if addBloblessFilter {
 		gitFetchFlags = "--filter=blob:none " + gitFetchFlags
+	}
+
+	// A job that asked for the base branch wants it whether or not its own commit
+	// still needs fetching, so this precedes the skip below.
+	if e.GitFetchBaseBranch {
+		e.fetchBaseBranch(ctx, gitFetchFlags)
+	}
+
+	if skipFetch {
+		e.shell.Commentf("Commit %q already exists locally, skipping fetch", e.Commit)
+		return nil
 	}
 
 	switch kind {
@@ -263,6 +272,59 @@ func prMergeRefspecHint(isMergeRefspec bool) string {
 		return ""
 	}
 	return "\nThis is possibly due to a merge conflict, or GitHub being unable to create the merge ref automatically"
+}
+
+// baseBranchToFetch returns the branch a diff in this job would be taken against.
+//
+// The precedence is the one `pipeline upload` already uses for --git-diff-base,
+// minus its final literal "main": guessing a name would mean a failed fetch on
+// every repository whose default branch is called something else.
+func (e *Executor) baseBranchToFetch() string {
+	built := strings.TrimPrefix(e.Branch, "refs/heads/")
+	for _, name := range []string{"BUILDKITE_PULL_REQUEST_BASE_BRANCH", "BUILDKITE_PIPELINE_DEFAULT_BRANCH"} {
+		branch, _ := e.shell.Env.Get(name)
+		branch = strings.TrimPrefix(branch, "refs/heads/")
+		if branch == "" {
+			continue
+		}
+		if branch == built {
+			// The job's own fetch already brought the branch under test.
+			return ""
+		}
+		return branch
+	}
+	return ""
+}
+
+// fetchBaseBranch updates refs/remotes/origin/<base>, which the job's own fetch
+// does not: it asks for refs/pull/N/head and the commit, and checkout directories
+// are reused, so that ref otherwise holds whatever an earlier build left there.
+//
+// It is separate from that fetch because FETCH_HEAD resolves to the first ref
+// fetched and the checkout depends on it. The refspec is explicit and forced so the
+// remote-tracking ref lands whatever remote.origin.fetch says, and follows a
+// force-pushed base branch instead of stopping at a non-fast-forward rejection.
+//
+// Neither fatal nor retried: the checkout is complete without it, and the retry on
+// the job's own fetch waits out asynchronous creation of the ref being built, which
+// a deleted base branch never gets.
+func (e *Executor) fetchBaseBranch(ctx context.Context, gitFetchFlags string) {
+	base := e.baseBranchToFetch()
+	if base == "" {
+		return
+	}
+
+	e.shell.Commentf("Fetch base branch %q", base)
+
+	refspec := fmt.Sprintf("+refs/heads/%s:refs/remotes/origin/%s", base, base)
+	if err := gitFetch(ctx, gitFetchArgs{
+		Shell:         e.shell,
+		GitFetchFlags: gitFetchFlags,
+		Repository:    "origin",
+		RefSpecs:      []string{refspec},
+	}); err != nil {
+		e.shell.Warningf("Couldn't fetch base branch %q, continuing: %v", base, err)
+	}
 }
 
 func isExistingCheckoutRemoteMirrorAttempt(attempt *remoteMirrorAttempt) bool {
--- a/agent/agent_configuration.go
+++ b/agent/agent_configuration.go
@@ -36,6 +36,7 @@ type AgentConfiguration struct {
 	GitSubmoduleCloneConfig         []string
 	SkipCheckout                    bool
 	GitSkipFetchExistingCommits     bool
+	GitFetchBaseBranch              bool
 	CheckoutOverrideMode            env.CheckoutOverrideMode
 	CheckoutAttempts                int
 	AllowedRepositories             []*regexp.Regexp
--- a/agent/job_runner.go
+++ b/agent/job_runner.go
@@ -615,6 +615,7 @@ BUILDKITE_AGENT_JWKS_KEY_ID`
 		setEnv("BUILDKITE_GIT_SUBMODULES", fmt.Sprint(r.conf.AgentConfiguration.GitSubmodules))
 		setEnv("BUILDKITE_SKIP_CHECKOUT", fmt.Sprint(r.conf.AgentConfiguration.SkipCheckout))
 		setEnv("BUILDKITE_GIT_SKIP_FETCH_EXISTING_COMMITS", fmt.Sprint(r.conf.AgentConfiguration.GitSkipFetchExistingCommits))
+		setEnv("BUILDKITE_GIT_FETCH_BASE_BRANCH", fmt.Sprint(r.conf.AgentConfiguration.GitFetchBaseBranch))
 		// A zero timeout means no checkout timeout; emit it anyway under strict so
 		// a job-supplied value can't reintroduce one past the agent config.
 		setEnv("BUILDKITE_GIT_CHECKOUT_TIMEOUT", strconv.Itoa(r.conf.AgentConfiguration.GitCheckoutTimeout))
@@ -632,6 +633,9 @@ BUILDKITE_AGENT_JWKS_KEY_ID`
 		}
 		if r.conf.AgentConfiguration.GitSkipFetchExistingCommits {
 			setCheckoutEnv("BUILDKITE_GIT_SKIP_FETCH_EXISTING_COMMITS", "true")
+		}
+		if r.conf.AgentConfiguration.GitFetchBaseBranch {
+			setCheckoutEnv("BUILDKITE_GIT_FETCH_BASE_BRANCH", "true")
 		}
 		if r.conf.AgentConfiguration.GitCheckoutTimeout > 0 {
 			setCheckoutEnv("BUILDKITE_GIT_CHECKOUT_TIMEOUT", strconv.Itoa(r.conf.AgentConfiguration.GitCheckoutTimeout))
--- a/clicommand/agent_start.go
+++ b/clicommand/agent_start.go
@@ -174,6 +174,7 @@ type AgentStartConfig struct {
 	GitSubmoduleCloneConfig     []string `cli:"git-submodule-clone-config"`
 	SkipCheckout                bool     `cli:"skip-checkout"`
 	GitSkipFetchExistingCommits bool     `cli:"git-skip-fetch-existing-commits"`
+	GitFetchBaseBranch          bool     `cli:"git-fetch-base-branch"`
 	CheckoutOverrideMode        string   `cli:"checkout-override-mode"`
 	CheckoutAttempts            int      `cli:"checkout-attempts"`
 
@@ -546,6 +547,7 @@ var AgentStartCommand = &cli.Command{
 		GitCheckoutTimeoutFlag,
 		GitSubmoduleCloneConfigFlag,
 		GitSkipFetchExistingCommitsFlag,
+		GitFetchBaseBranchFlag,
 		CheckoutAttemptsFlag,
 
 		&cli.StringFlag{
@@ -1006,6 +1008,7 @@ var AgentStartCommand = &cli.Command{
 			GitSubmoduleCloneConfig:         cfg.GitSubmoduleCloneConfig,
 			SkipCheckout:                    cfg.SkipCheckout,
 			GitSkipFetchExistingCommits:     cfg.GitSkipFetchExistingCommits,
+			GitFetchBaseBranch:              cfg.GitFetchBaseBranch,
 			CheckoutOverrideMode:            checkoutMode,
 			CheckoutAttempts:                cfg.CheckoutAttempts,
 			SSHKeyscan:                      !cfg.NoSSHKeyscan,
--- a/clicommand/bootstrap.go
+++ b/clicommand/bootstrap.go
@@ -75,6 +75,7 @@ type BootstrapConfig struct {
 	SkipCheckout                 bool          `cli:"skip-checkout"`
 	GitCheckoutTimeout           int           `cli:"git-checkout-timeout"`
 	GitSkipFetchExistingCommits  bool          `cli:"git-skip-fetch-existing-commits"`
+	GitFetchBaseBranch           bool          `cli:"git-fetch-base-branch"`
 	GitCheckoutFlags             string        `cli:"git-checkout-flags"`
 	GitCloneFlags                string        `cli:"git-clone-flags"`
 	GitFetchFlags                string        `cli:"git-fetch-flags"`
@@ -284,6 +285,7 @@ var BootstrapCommand = &cli.Command{
 		GitSubmoduleCloneConfigFlag,
 		GitCheckoutTimeoutFlag,
 		GitSkipFetchExistingCommitsFlag,
+		GitFetchBaseBranchFlag,
 		CheckoutAttemptsFlag,
 
 		&cli.StringFlag{
@@ -487,6 +489,7 @@ var BootstrapCommand = &cli.Command{
 			SkipCheckout:                 cfg.SkipCheckout,
 			GitCheckoutTimeout:           cfg.GitCheckoutTimeout,
 			GitSkipFetchExistingCommits:  cfg.GitSkipFetchExistingCommits,
+			GitFetchBaseBranch:           cfg.GitFetchBaseBranch,
 			CheckoutOverrideMode:         checkoutMode,
 			Command:                      cfg.Command,
 			CommandEval:                  cfg.CommandEval,
--- a/clicommand/global.go
+++ b/clicommand/global.go
@@ -334,6 +334,12 @@ var (
 		Sources: cli.EnvVars("BUILDKITE_GIT_SKIP_FETCH_EXISTING_COMMITS"),
 	}
 
+	GitFetchBaseBranchFlag = &cli.BoolFlag{
+		Name:    "git-fetch-base-branch",
+		Usage:   "Also fetch the base branch during checkout, so every command in the job can diff against its current tip. The base branch is the first non-empty value of {$BUILDKITE_PULL_REQUEST_BASE_BRANCH, $BUILDKITE_PIPELINE_DEFAULT_BRANCH}, and nothing is fetched when it is the branch being built (default: false)",
+		Sources: cli.EnvVars("BUILDKITE_GIT_FETCH_BASE_BRANCH"),
+	}
+
 	CheckoutAttemptsFlag = &cli.IntFlag{
 		Name:    "checkout-attempts",
 		Value:   6,
--- a/docs/remote-git-mirrors.md
+++ b/docs/remote-git-mirrors.md
@@ -176,7 +176,9 @@ point a reader would otherwise file a bug. §10 lists each one and where it goes
 4. `git clean`, optional `git lfs install --local`.
 5. `fetchSource` → `git fetch <flags> -- origin <refspec>`, where the refspec is
    the custom refspec, or `refs/pull/N/{head,merge}`, or the branch (when
-   `BUILDKITE_COMMIT=HEAD`), or — the common case — the commit SHA.
+   `BUILDKITE_COMMIT=HEAD`), or — the common case — the commit SHA. With
+   `--git-fetch-base-branch`, an additional fetch runs first and updates
+   `refs/remotes/origin/<base>`.
 6. Commit verification, sparse setup, `git checkout`, submodules, LFS, clean.
 
 Four things are easy to get wrong, and the plan depends on all of them:
--- a/env/protected.go
+++ b/env/protected.go
@@ -115,6 +115,7 @@ var checkoutOverrideScope = map[string]struct{}{
 	"BUILDKITE_GIT_CLEAN_FLAGS":                 {},
 	"BUILDKITE_GIT_CLONE_FLAGS":                 {},
 	"BUILDKITE_GIT_COMMIT_VERIFICATION":         {},
+	"BUILDKITE_GIT_FETCH_BASE_BRANCH":           {},
 	"BUILDKITE_GIT_FETCH_FLAGS":                 {},
 	"BUILDKITE_GIT_SKIP_FETCH_EXISTING_COMMITS": {},
 	"BUILDKITE_GIT_SPARSE_CHECKOUT_MODE":        {},
--- a/internal/job/checkout_fetch_base_branch_test.go
+++ b/internal/job/checkout_fetch_base_branch_test.go
@@ -1,0 +1,303 @@
+package job
+
+import (
+	"errors"
+	"os"
+	"os/exec"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/buildkite/agent/v4/internal/job/githttptest"
+	"github.com/buildkite/agent/v4/internal/shell"
+)
+
+// The reused checkout is the case GitFetchBaseBranch exists for: the job's own fetch
+// asks for refs/pull/N/head and the commit, so refs/remotes/origin/<base> keeps
+// whatever an earlier build in the same directory left there.
+func TestFetchSourceBaseBranch(t *testing.T) {
+	tests := []struct {
+		name            string
+		fetchBaseBranch bool
+		branch          string
+		skipExisting    bool
+		env             map[string]string
+		// Whether origin/main should have moved to the remote's new tip.
+		wantFetched bool
+	}{
+		{
+			name:            "fetches the pull request base branch",
+			fetchBaseBranch: true,
+			branch:          "feature-branch",
+			env:             map[string]string{"BUILDKITE_PULL_REQUEST_BASE_BRANCH": "main"},
+			wantFetched:     true,
+		},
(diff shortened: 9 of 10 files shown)


jasonwbarnett:
hello. real life human being here. I reviewed the output from Claude and I'm going to be deploying this out to our production fleet shortly as it's urgent and we need this functionality to guarantee base branches are up-to-date for use when answering the question: "What changed in this PR?" or "What changed in this multi-commit push to main from the merge queue?"

Altana is a customer of Buildkite and we would like to see this get merged and released, although we're having issues with the latest release and are stuck on v3.133.0 and we haven't yet diagnosed why our fleet went to hell when we released Elastic CI Stack v6.71.0 which bundles buildkite-agent v3.136.2. I didn't diagnose if it was the stack or the agent, but this came to my mind as I was typing this so I likely need to rollout a test queue using the newest CI stack and agent to see if those issues are resolved.

jasonwbarnett:
This (https://github.com/altana-ai/buildkite-agent/releases/tag/v3.133.0-altana.1) is now deployed to production and working beautifully.

For BK employee(s) to see it in action: https://buildkite.com/altana/monorepo-build/builds/152465/canvas?sid=01a05596-b6c2-4ce5-918f-85f665d5e8ca&tab=output

swebb:
Thanks for opening this PR @jasonwbarnett. I’ve created a ticket to track it internally.

buildkate:
> hello. real life human being here. I reviewed the output from Claude and I'm going to be deploying this out to our production fleet shortly as it's urgent and we need this functionality to guarantee base branches are up-to-date for use when answering the question: "What changed in this PR?" or "What changed in this multi-commit push to main from the merge queue?"
> 
> Altana is a customer of Buildkite and we would like to see this get merged and released, although we're having issues with the latest release and are stuck on v3.133.0 and we haven't yet diagnosed why our fleet went to hell when we released Elastic CI Stack v6.71.0 which bundles buildkite-agent v3.136.2. I didn't diagnose if it was the stack or the agent, but this came to my mind as I was typing this so I likely need to rollout a test queue using the newest CI stack and agent to see if those issues are resolved.

@jasonwbarnett thanks for the PR and putting in the time to write a real like human comment! 😄 
Overall no big issues with this as implemented. I do want to call out that your comment describes `--git-fetch-base-branch` as **guaranteeing** that the base branch ref is up to date, but the fetch is intentionally non-fatal and leaves an existing stale ref intact when it fails. 

I would describe this as a best-effort refresh instead.  For example, “attempts to update `origin/<base>` before commands run”?  Otherwise guaranteeing freshness requires failing closed.

internal/job/checkout_fetch.go
```diff
@@ -265,6 +274,59 @@ func prMergeRefspecHint(isMergeRefspec bool) string {
...
+func (e *Executor) baseBranchToFetch() string {
+	built := strings.TrimPrefix(e.Branch, "refs/heads/")
+	for _, name := range []string{"BUILDKITE_PULL_REQUEST_BASE_BRANCH", "BUILDKITE_PIPELINE_DEFAULT_BRANCH"} {
+		branch, _ := e.shell.Env.Get(name)
+		branch = strings.TrimPrefix(branch, "refs/heads/")
+		if branch == "" {
+			continue
+		}
+		if branch == built {
+			// The job's own fetch already brought the branch under test.
```
buildkate, line 291: Skipping the fetch when the base branch is also the branch being built seems reasonable: there is no distinct base branch to prepare for comparison. However two nits:
1. The existing `skips the branch being built` case still uses the shared executor configured as a GitHub PR (`PullRequest: "124"`), so it does not test a normal default-branch build.  Could we add or replace it with a case using `PullRequest: "false"`, `Branch: "main"`, a commit SHA, and `BUILDKITE_PIPELINE_DEFAULT_BRANCH=main`, asserting that `origin/main` remains unchanged?
2. The comment here is slightly incorrect: `The job's own fetch already brought the branch under test.`.  The agent fetches the requested commit by SHA. That downloads the commit but does not update the `origin/<BUILDKITE_BRANCH>` ref.
I'd change it to something like: `There is no distinct base branch to prepare when building the base itself.`

jasonwbarnett:
I am actually thinking about adding another environment variable which controls strictness, similar to the recently added verification that the git commit sha is on the branch. Because I can imagine some people want to fetch optimistically and are ok if it fails and others (like me) want strong guarantees it actually fetches.

buildkate:
> I am actually thinking about adding another environment variable which controls strictness, similar to the recently added verification that the git commit sha is on the branch. Because I can imagine some people want to fetch optimistically and are ok if it fails and others (like me) want strong guarantees it actually fetches.

 That's a good approach.  Default to optimistic with the option to enforce.  Did you want to extend this PR or open a separate one?

jasonwbarnett:
> > I am actually thinking about adding another environment variable which controls strictness, similar to the recently added verification that the git commit sha is on the branch. Because I can imagine some people want to fetch optimistically and are ok if it fails and others (like me) want strong guarantees it actually fetches.
> 
> That's a good approach. Default to optimistic with the option to enforce. Did you want to extend this PR or open a separate one?

I want to extend this one. I'd like to get it feature complete now since I have engagement and this is on a good path.

internal/job/checkout_fetch.go
jasonwbarnett: Both done, thanks.

1. Added `skips a default branch build` — `PullRequest: "false"`, `Branch: "main"`, the commit SHA, `BUILDKITE_PIPELINE_DEFAULT_BRANCH=main`, asserting `origin/main` is untouched. I kept the existing PR-shaped case alongside it rather than replacing it, since the two reach the guard by different paths (`BUILDKITE_PULL_REQUEST_BASE_BRANCH` vs. the default-branch fallback). Both fail if the guard is dropped.
2. Comment is now `There is no distinct base branch to prepare when building the base itself.` — you're right that the job's own fetch brings the commit by SHA and never updates `origin/<BUILDKITE_BRANCH>`.

The mode work from the other thread is in the same push: `off` (default) / `optimistic` / `strict`, shaped like `BUILDKITE_GIT_COMMIT_VERIFICATION`. `optimistic` is the best-effort behaviour you described; `strict` fails closed, including when nothing names a base branch at all — skipping quietly there would leave the job diffing against whatever an earlier build left in the checkout directory, which is the case the mode exists to catch. Description updated to match, so it no longer claims a guarantee for the best-effort path.

jasonwbarnett:
Extended this PR as promised. `--git-fetch-base-branch` now takes a mode instead of a boolean, following `BUILDKITE_GIT_COMMIT_VERIFICATION`:

- `off` (default) — no extra fetch.
- `optimistic` — the behaviour you reviewed, described as best-effort now: it attempts to update `origin/<base>` before commands run and warns without failing when it cannot.
- `strict` — fails closed, which is the only way to actually guarantee freshness.

Two things about `strict` worth a look, since they're judgement calls rather than mechanics:

- It fails when **nothing names a base branch** (neither `BUILDKITE_PULL_REQUEST_BASE_BRANCH` nor `BUILDKITE_PIPELINE_DEFAULT_BRANCH`), not just when the fetch itself fails. Skipping quietly there would leave the job diffing against whatever `origin/<something>` an earlier build left in the reused checkout — exactly the silent-stale-base failure the mode is for. Retrying can't discover a base branch, so that case breaks the checkout retrier instead of spending the whole attempt budget, the same treatment `ErrCommitVerificationFailed` gets.
- Building the base branch itself is still a skip in **every** mode, `strict` included: there's no distinct base branch to prepare.

The fetch is still not retried inside either mode, for the reason in the original description. Under `strict` the failure is returned instead, and the checkout's own retrier handles a transient one, classifying it exactly as it does a failure to fetch the job's own source.

One breaking note for the changelog if this lands: `BUILDKITE_GIT_FETCH_BASE_BRANCH=true` is no longer valid and must become `optimistic` or `strict`. The value is validated, so an agent carrying the old boolean fails to start rather than silently skipping the fetch — loud, which seems right, and since the flag has never shipped I left the enum clean rather than accepting `true` as an alias. Happy to add the alias if you'd prefer.

Also rebased onto `main` (994fd2cd). `go test ./...` is clean across all 38 packages — including `clicommand.TestAgentStartupHookWithAdditionalPaths`, which I'd flagged as failing on a pristine `main` in the first round and no longer reproduces.

jasonwbarnett:
Pushed `1c0deecb`: the base branch fetch now retries. Remote outages are a frequent cause of CI failures for us, and one attempt was all this got — under `optimistic` that silently left the stale ref in place with nothing but a warning, which defeats the point of turning the flag on.

It retries in both modes, on the same budget as the job's own fetch (10 attempts, subsecond exponential backoff, about a minute and a half if every one fails). The exception is a remote that answers `couldn't find remote ref`: that ends the retries at once, so a deleted base branch still costs a single fetch. That exception is also why this can't just set `Retry` on `gitFetch` — that retrier deliberately keeps retrying a missing ref, to wait out GitHub creating `refs/pull/N/head` asynchronously, and a base branch is either already there or gone for good.

`githttptest` grows a `FailNextRequests(n)` knob for this, so a test can watch a fetch survive a remote that's briefly unavailable. Both directions are covered: a fetch that succeeds after two rejected requests, and a time bound on the missing-ref case that fails if it ever starts exhausting the budget.

jasonwbarnett:
Correction to the comment above, in `f13832f8`: the two modes no longer share one retry budget.

`strict` keeps the job's own (10 attempts, ~90s) because the alternative is failing the job. `optimistic` gets 3 (~2s): it has already decided the job proceeds either way, so a minute and a half of waiting buys nothing it's willing to act on — but one attempt was still too few, since a single reset would silently widen every diff in the job and report it as a warning. Fail-fast on `couldn't find remote ref` is unchanged and applies to both.

clicommand/global_test.go
```diff
@@ -26,6 +26,73 @@ func TestAllFlagEnvs(t *testing.T) {
...
+		},
+		{
+			name: "can be set to strict",
+			args: []string{"--git-fetch-base-branch", "strict"},
+			want: "strict",
+		},
+		{
+			// A boolean is what the flag took before it grew modes, and what anyone
+			// would guess from its name, so the value must be rejected rather than
+			// quietly enabling or disabling the fetch.
```
buildkate: Could you please remove this comment as original PR was never merged and the historical boolean value only really exists in the PR threads.

buildkate:
> One breaking note for the changelog if this lands: `BUILDKITE_GIT_FETCH_BASE_BRANCH=true` is no longer valid and must become `optimistic` or `strict`. The value is validated, so an agent carrying the old boolean fails to start rather than silently skipping the fetch — loud, which seems right, and since the flag has never shipped I left the enum clean rather than accepting `true` as an alias. Happy to add the alias if you'd prefer.
> 

This was only in a PR comment but want to clarify that there will be no breaking changes since the PR introducing the flag was never merged.

internal/job/checkout_fetch_base_branch.go
```diff
@@ -0,0 +1,204 @@
...
+
+	refspec := fmt.Sprintf("+refs/heads/%s:refs/remotes/origin/%s", base, base)
+	err := roko.NewRetrier(
+		roko.WithMaxAttempts(attempts),
+		roko.WithStrategy(roko.ExponentialSubsecond(time.Second)),
+		roko.WithJitter(),
+	).DoWithContext(ctx, func(r *roko.Retrier) error {
+		err := gitFetch(ctx, gitFetchArgs{
+			Shell:         e.shell,
+			GitFetchFlags: gitFetchFlags,
```
buildkate, line 179: We preserve the configured fetch flags here, so `--dry-run` or `--prefetch` can make the fetch succeed without updating `origin/<base>`, even in strict mode. Using this PR’s existing fixture, the following test fails for both flags:

```go
func TestStrictBaseFetchRefreshesRef(t *testing.T) {
    for _, flags := range []string{"--dry-run", "--prefetch"} {
        t.Run(flags, func(t *testing.T) {
             f := newBaseBranchFixture(t) // stale origin/main
            e := newBaseBranchFetchExecutor(t, f)
            e.Branch = "feature-branch"
            e.GitFetchBaseBranch = GitFetchBaseBranchStrict
            e.GitFetchFlags = flags
            e.shell.Env.Set("BUILDKITE_PULL_REQUEST_BASE_BRANCH", "main")

            if err := e.fetchSource(t.Context(), false, nil); err != nil {
                return // Failing closed is acceptable.
            }
            if got := gitRevParseForBaseBranchTest(
                t, f.checkout, "origin/main",
            ); got != f.currentMain {
                t.Fatal("strict fetch succeeded but origin/main remains stale")
            }
        })
    }
}
```

Given your use case relies on ref freshness, what behavior would you expect here? We could preserve the flags and document the exception to strict mode’s freshness guarantee. Alternatively, we could strip ref-suppressing flags for this fetch,[ as commit verification does](https://github.com/buildkite/agent/blob/9643022ea69adb1b2fcf9f8b5603de176feddab3/internal/job/commit_verification.go#L224), or reject conflicting flags in strict mode.

My preference is to preserve the flags and document the exception, since `--dry-run` explicitly requests a non-mutating operation; silently stripping it would go against that intention. But I’d appreciate your perspective here. 

We should also  add a test for whichever behavior we choose.

internal/job/checkout_fetch_base_branch.go
```diff
@@ -0,0 +1,204 @@
...
+			Repository:    "origin",
+			RefSpecs:      []string{refspec},
```
buildkate: Could you please pass the generated refspec through `shellwords.Quote` to escape characters before passing it to `gitFetch`?

```go
RefSpecs: []string{shellwords.Quote(refspec)},
```

`gitFetch` calls `shellwords.Split` on each refspec, which otherwise interprets valid characters in branch names as shell syntax. So `release'candidate` becomes `releasecandidate`. If both branches exist, strict mode can successfully fetch the wrong branch and leave the intended ref stale.

Could we also add a regression test with both branch names present, asserting that the intended remote-tracking ref is updated?

buildkate:
@jasonwbarnett Hey mate, appreciate your collaboration on this PR. I like the addition of the mode 👍

clicommand/global_test.go
jasonwbarnett: Removed. Agreed — the boolean only ever existed in these threads, so the comment pointed at something no reader of the repo can see. I've taken the same pass over the PR description: the "breaking if you're already running the first revision" note and the running commentary on earlier review rounds are gone, and it now just describes the change as it stands.

internal/job/checkout_fetch_base_branch.go
jasonwbarnett: Good catch, and your test does reproduce it. I've gone with your third option — `strict` rejects those flags — rather than preserve-and-document, and I want to put the case to you because I think it lands closer to your own reasoning than the middle option does.

I agree that silently stripping `--dry-run` is wrong: the operator asked for a fetch that mutates nothing, and overriding that is the agent deciding it knows better. But documenting the exception has the same problem pointed the other way — the operator also asked for a guaranteed-current base branch, and they'd get a mode that reports success while `origin/<base>` stays exactly as stale as it was. `strict` has no other reason to exist; `optimistic` already covers "fetch if you can". An exception in the docs makes the guarantee conditional on a flag usually set in agent config, far from the pipeline that relies on the diff.

The two requests genuinely contradict, so neither can be quietly honoured at the other's expense. `strict` now says so:

```
git fetch flags write no base branch ref: BUILDKITE_GIT_FETCH_BASE_BRANCH=strict cannot guarantee a current base branch while BUILDKITE_GIT_FETCH_FLAGS contains --dry-run
```

`--dry-run` is still never stripped, and `optimistic` keeps the flags untouched — it promises nothing about the ref, and a stale base is the outcome it already tolerates. The check sits on the fetch path rather than at flag-parse time, so an agent configured this way only fails the jobs that actually wanted a base branch fetched, and it breaks the checkout retrier, since another clone won't rewrite the flags.

Taking your pointer to `commit_verification.go` as the precedent: rather than duplicating the flag matching I pulled `isRefSuppressingFetchFlag` out of `stripRefSuppressingFetchFlags`, so verification still strips those flags for its probe and the base branch fetch reports them, both getting the prefix handling (`--dry` is git's abbreviation of `--dry-run`) from one place. Say the word if you'd rather I left `commit_verification.go` alone and kept a local copy in the base-branch file.

Tests in `TestFetchSourceBaseBranchStrictRejectsFetchFlagsThatWriteNoRef` (`--dry-run`, `--prefetch`, `--negotiate-only`, `-v --dry`) and `TestFetchSourceBaseBranchOptimisticKeepsFetchFlagsThatWriteNoRef`. With the guard removed all four strict cases fail, three of them the way you predicted — and `--negotiate-only` takes 92s to do it, because it fails the fetch rather than no-opping and so burns the whole strict retry budget first. A small extra argument for catching the conflict before the fetch rather than after.

Happy to switch to preserve-and-document if you'd still rather; it's a smaller diff and I don't feel strongly enough to hold up the PR over it.

internal/job/checkout_fetch_base_branch.go
jasonwbarnett: Done, though not with `shellwords.Quote` — it turns out that only fixes posix, so I've passed the refspec through verbatim instead.

`QuoteBatch` escapes only `^&;,=%`; neither `'` nor `"` is in that set, yet both *are* in `SplitBatch`'s `QuoteChars`. So on Windows the quoted refspec is byte-identical to the unquoted one and gets mangled exactly as you described:

```
batch  "release'candidate"    *** BROKEN ***
  quoted = "+refs/heads/release'candidate:refs/remotes/origin/release'candidate"
  split  = ["+refs/heads/releasecandidate:refs/remotes/origin/releasecandidate"]
```

Because the branch name appears twice in the refspec the quote count is always even, so it mangles silently rather than erroring — and the regression test would have passed on Linux while the bug survived on the Windows AMD64 step.

So instead of round-tripping through a shell quoter, `gitFetchArgs` grows `LiteralRefSpecs`, appended to the argv without `shellwords.Split`:

```go
// LiteralRefSpecs are refspecs passed to git exactly as given, one argument
// each. RefSpecs is word-split because it carries operator-supplied strings
// that may hold several refspecs, but a refspec the agent builds from a ref
// name must not be [...]
```

That keeps this fetch going through `gitFetch`, which matters because `baseBranchFetchIsRetryable` classifies on the `gitError` types `gitFetch` produces — a remote answering "couldn't find remote ref" has to end the retries, while git's generic exit 128 has to keep them. Building the argv directly the way `checkCommitOnBranch` does would have given up that classification.

Regression test is `TestFetchSourceBaseBranchFetchesABranchNameNeedingQuoting`, as you asked: both `release'candidate` and `releasecandidate` exist on the remote at different tips, `BUILDKITE_PULL_REQUEST_BASE_BRANCH=release'candidate`, `strict`. It asserts `origin/release'candidate` lands the intended tip and `origin/releasecandidate` is absent. Switching the call back to `RefSpecs` makes it fail as the bug you reported, on any platform:

```
origin/release'candidate = "", want "5f1cb8bac410f6d117b528ca43a9c891271b13ce"
origin/releasecandidate = "e6d4bacbfb0d732dacba78b64c059a2fbbd08d4f", want it to be absent
```

`strict` reported success there, which was the worst part of it.

jasonwbarnett:
All three threads answered inline. I rewrote the branch into topical commits rather than stacking fixups, so the diff against your review is gone — apologies for that. What changed:

**`global_test.go` comment** — removed, and I took the same pass over the PR description: the "breaking if you're on the first revision" note and the commentary on earlier review rounds are gone.

**`shellwords.Quote`** — it only fixes posix, so I went a different way. `QuoteBatch` escapes neither `'` nor `"`, yet `SplitBatch` treats both as quote characters, so on Windows the quoted refspec is byte-identical to the unquoted one and mangles exactly as you described. The regression test would have passed on Linux and failed the Windows AMD64 step. Instead `gitFetchArgs` grows `LiteralRefSpecs`, appended to argv without `shellwords.Split` — nothing round-trips through a shell quoter, and the fetch still goes through `gitFetch` so `baseBranchFetchIsRetryable` keeps its error classification.

**`--dry-run` / strict** — took your third option (reject the conflict) over preserve-and-document. Full reasoning in that thread; it's a small diff either way and I'll switch if you'd still rather document the exception. `--negotiate-only` is in the matched set too, since it also leaves the ref stale.

Commits on top of the original four:

- `f1051c2f` Pass the base branch refspec to git verbatim
- `c899d78f` Reject fetch flags that write no ref under strict
- `542f8aa6` Pin that the base branch fetch leaves FETCH_HEAD alone

That last one is unprompted: reviewing my own work I found two tests that weren't pinning what they claimed — the separate-fetch/`FETCH_HEAD` rationale had no test at all, and the optimistic flag test passed against a no-op. Both now fail against the mutations they're meant to catch. Worth knowing if you write tests in that file: the fixture's `feature-branch` and advanced `main` share a sha (same tree, parent and message, same author second), which is now called out in `newBaseBranchFixture`.

`go test ./...` clean, `golangci-lint` 0 issues, and each commit builds and passes `internal/job` on its own.

PR #4286, opened 2026-08-28
Set last hook exit status for unwrapped hooks

francoiscampbell:
## Description

Unwrapped hooks (non-shell scripts and binaries) currently do not update `BUILDKITE_LAST_HOOK_EXIT_STATUS`. This differs from wrapped shell hooks and can leave the variable unset or stale for subsequent hooks, even when the unwrapped hook succeeded.

Set the variable from the unwrapped process result so later hooks observe the same exit-status behavior regardless of hook type.

## Context

This surfaced after converting a plugin command hook from Bash to Python. Its pre-exit hook used `BUILDKITE_LAST_HOOK_EXIT_STATUS` to report the command result, but the value was no longer populated after the successful Python hook.

## Changes

- Set `BUILDKITE_LAST_HOOK_EXIT_STATUS` after running an unwrapped hook, using the same `shell.ExitCode` conversion as wrapped hooks.
- Add regression coverage for successful and failing unwrapped hooks.

## Testing

Passed:

```text
go test ./internal/job -run '^TestRunUnwrappedHookSetsLastHookExitStatus$' -count=1
```

Instacart

## Disclosures / Credits

OpenAI Codex via Pi helped diagnose the behaviour, implement the change, and write the regression test. The PR was then reviewed and refined by a human.

### DIFF
--- a/internal/job/executor_test.go
+++ b/internal/job/executor_test.go
@@ -416,7 +416,7 @@ func TestRunUnwrappedHookSetsLastHookExitStatus(t *testing.T) {
 		t.Run(test.name, func(t *testing.T) {
 			e := newCancelTestExecutor(t)
 			hookPath := filepath.Join(t.TempDir(), "hook")
-			hook := []byte(fmt.Sprintf("#!/bin/sh\nexit %d\n", test.exitStatus))
+			hook := []byte(fmt.Sprintf("#!/usr/bin/env python\nimport sys; sys.exit(%d)\n", test.exitStatus))
 			if err := os.WriteFile(hookPath, hook, 0o755); err != nil {
 				t.Fatalf("os.WriteFile(%q) error = %v", hookPath, err)
 			}
--- a/internal/job/executor_test.go
+++ b/internal/job/executor_test.go
@@ -139,41 +139,6 @@ func newCancelTestExecutor(t *testing.T) *Executor {
 	return e
 }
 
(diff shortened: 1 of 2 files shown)


francoiscampbell:
Could a maintainer please add the `bug` label? This fixes `BUILDKITE_LAST_HOOK_EXIT_STATUS` not being updated for unwrapped hooks. External contributors do not have permission to add labels in this repository.

internal/job/executor_test.go
```diff
@@ -397,3 +400,38 @@ func TestConfigureRepositoryProviderGitCredentials(t *testing.T) {
...
+		name       string
+		exitStatus int
+	}{
+		{name: "success", exitStatus: 0},
+		{name: "failure", exitStatus: 7},
+	} {
+		t.Run(test.name, func(t *testing.T) {
+			e := newCancelTestExecutor(t)
+			hookPath := filepath.Join(t.TempDir(), "hook")
+			hook := []byte(fmt.Sprintf("#!/usr/bin/env python\nimport sys; sys.exit(%d)\n", test.exitStatus))
```
buildkate, line 419:
```suggestion
			hook := []byte(fmt.Sprintf("#!/usr/bin/env ruby\nexit(%d)\n", test.exitStatus))
```

buildkate:
@francoiscampbell Thanks for catching the bug and opening the PR. It looks good. Just a small change request in the test case because we don't have python available in our CI env.

buildkate (approved):

PR #4287, opened 2026-08-29
Resolve relative submodule URLs for mirrors

### LINKED ISSUE #4282: Git mirror checkout fails for relative submodule URLs
## Describe the bug

When git mirrors are enabled (`git-mirrors-path`), checkout of a repository whose submodules use **relative URLs** in `.gitmodules` fails before any job commands run.

Git allows submodule URLs to be relative to the superproject remote, e.g.:

```ini
[submodule "modules/java-prettify"]
    path = modules/java-prettify
    url = ../java-prettify
```

Git resolves `../java-prettify` against the superproject remote — `https://gerrit.googlesource.com/gerrit.git` + `../java-prettify` = `https://gerrit.googlesource.com/java-prettify`. The agent's mirror-submodule path instead reads the **raw** `.gitmodules` value and runs `git clone --mirror` from the mirror cache directory (`/var/lib/gitmirrors`), where `../java-prettify` is treated as a local filesystem path and does not exist.

## Steps To Reproduce

1. Configure an agent with git mirrors enabled:
   ```
   git-mirrors-path="/var/lib/gitmirrors"
   git-clone-mirror-flags="-v --bare"
   ```
2. Run a pipeline against a repository with an absolute clone URL (e.g. `https://gerrit.googlesource.com/gerrit.git`) whose `.gitmodules` declares a submodule with a relative `url` such as `../java-prettify`.
3. Checkout runs `git submodule sync --recursive`, then attempts to mirror-clone the raw relative URL from the mirror cache directory:
   ```
   $ cd /var/lib/gitmirrors
   # Cloning a mirror of the repository to "/var/lib/gitmirrors/---java-prettify"
   $ git clone --mirror -v --bare -- ../java-prettify /var/lib/gitmirrors/---java-prettify
   fatal: repository '../java-prettify' does not exist
   Warning: Checkout failed! getting/updating mirror dir for submodules: exit status 128
   ```

Minimal, network-free reproducer: https://github.com/davido/bk-rel-submodule-repro-issue-2815

## Expected behavior

Relative submodule URLs should not break checkout. They should either be resolved against the superproject remote before creating/updating the mirror, or the mirror optimization should be skipped for that submodule so normal `git submodule update` performs Git's standard relative-URL resolution:

```
../java-prettify -> https://gerrit.googlesource.com/java-prettify
```

Absolute submodule URLs should continue to use mirrors.

## Actual behaviour

The mirror path clones the unresolved relative URL from inside the mirror cache directory and fails:

```
git clone --mirror -v --bare -- ../java-prettify /var/lib/gitmirrors/---java-prettify
fatal: repository '../java-prettify' does not exist
```

The failure repeats until checkout gives up with `getting/updating mirror dir for submodules: exit status 128`. The mirror-submodule path is in `internal/job/checkout.go` (`updateGitSubmodules`).

## Stack parameters (please complete the following information)

 - Version: v4.0.0-beta.16 (also present on `main`)

## Affiliation (optional)

External contributor, filing on behalf of the Gerrit project's Bazel downstream Buildkite CI pipeline.

## Additional context

- Root-cause discussion and reproducer: bazelbuild/continuous-integration#2815
- Proposed fix (minimal, fail-open — skips the mirror only for relative submodule URLs, with a regression test): #4276

This is not a request to change repositories to absolute submodule URLs — relative submodule URLs are valid Git behavior, so the fix belongs in the checkout/mirror layer, not in affected repositories.

#4276 was a minimal first step that simply skipped the mirror for relative submodule URLs. The more complete fix is now open as #4287: it resolves relative submodule URLs against the superproject remote before creating/updating the mirror, so relative submodules keep the mirror optimization instead of falling back to a plain clone.

To keep it safe, resolution is fail-open for relative submodules — if URL resolution or the resolved mirror update fails, checkout warns and falls back to a plain git submodule update, letting Git apply its own authoritative relative-URL semantics. Absolute submodule mirror behavior is unchanged.

Example: `https://gerrit.googlesource.com/gerrit.git` + `../java-prettify` → `https://gerrit.googlesource.com/java-prettify`.

davido:
## Description

When `git-mirrors-path` is enabled, the agent pre-populates a Git mirror for each
submodule enumerated from `.gitmodules` before running `git submodule update`.
Submodule URLs in `.gitmodules` may be **relative to the superproject remote**,
e.g. `url = ../java-prettify`. The mirror prefetch path passed those raw values
directly to `git clone --mirror` while `cd`'d into the mirror cache directory, so
`../java-prettify` was interpreted as a local filesystem path under
`git-mirrors-path` instead of a repository relative to the superproject remote —
and checkout failed before any job command ran.

This PR resolves relative submodule URLs against `e.Repository` **only for mirror
selection/population**. The actual `git submodule update` command is left
unchanged (aside from whether it receives `--reference <mirrorDir>`), so Git
remains the authority on submodule checkout semantics.

Resolution is fail-open for relative submodules: if URL resolution or the
resolved mirror update fails, the agent warns and falls back to a plain
`git submodule update` (no `--reference`), letting Git apply its own
authoritative relative-URL handling. Absolute submodule mirror failures keep the
existing hard-fail behavior.

**Alternatives considered:** a minimal fix that simply *skipped* mirrors for any
relative submodule URL. It was safe but gave up the mirror optimization even when
the agent can resolve the URL correctly. The chosen approach keeps that safety net
(via the fail-open fallback) while still getting mirror performance in the common
case where the resolver agrees with Git.

## Context

- Closes #4282
- Refs bazelbuild/continuous-integration#2815

Example resolution: `https://gerrit.googlesource.com/gerrit.git` + `../java-prettify`
→ `https://gerrit.googlesource.com/java-prettify`.

## Changes

- **`internal/job/checkout.go`** — in `updateGitSubmodules`, resolve relative
  submodule URLs before the mirror update; add fail-open fallback for relative
  submodules (resolver error or resolved-mirror failure → plain submodule update);
  preserve hard-fail for absolute submodules.
- **`internal/job/git.go`** — add `resolveGitSubmoduleURL` and
  `isRelativeSubmoduleURL`. Resolution preserves the original URL form:
  `https`/`ssh` via `url.URL`, scp-like (`git@host:path`, including absolute
  `git@host:/abs/path`) via string reconstruction, and local paths via
  `filepath.Join`. `scpLikeURLPattern` now captures the leading slash after the
  colon so absolute scp remote paths survive resolution; `parseGittableURL` is
  updated to preserve that slash and is behavior-identical for all other inputs.
- **Tracing/logging** — the submodule mirror span reports `git.repo` as the
  resolved URL and adds `git.raw_repo` (the raw `.gitmodules` value) only when it
  differs; a comment logs the resolution. Repository URLs are credential-redacted
  in logs and span attributes.
- **Tests** — resolver unit coverage (absolute/relative × HTTPS/SSH/scp-like/scp
  absolute-path/local, cleaned paths, bare name unchanged, malformed-URL parse
  error, relative-URL detection) plus a `parseGittableURL` absolute-scp case; two
  integration tests (`..._WithRelativeSubmoduleURL_WithGitMirrors` and
  `..._WithRelativeSubmoduleURLFallsBackWhenMirrorFails_WithGitMirrors`) sharing a
  `setupRelativeSubmoduleFixture` helper.

## Testing

Focused runs:

```sh
go test ./internal/job -run 'TestResolveGitSubmoduleURL|TestIsRelativeSubmoduleURL|TestParseGittableURL'
go test ./internal/job/integration -run 'TestCheckingOutLocalGitProjectWith(Submodules|RelativeSubmoduleURL|RelativeSubmoduleURLFallsBackWhenMirrorFails|SubmodulesDisabled)_WithGitMirrors'
go vet ./internal/job/...
go build ./...
git diff --check
```

External contributor, filing on behalf of the Gerrit project's Bazel downstream Buildkite CI pipeline.

### DIFF
--- a/internal/job/git.go
+++ b/internal/job/git.go
@@ -8,6 +8,7 @@ import (
 	"net"
 	"net/url"
 	"os/exec"
+	"path"
 	"path/filepath"
 	"regexp"
 	"slices"
@@ -449,6 +450,35 @@ func gitEnumerateSubmoduleURLs(ctx context.Context, sh *shell.Shell) ([]string, 
 	return urls, nil
 }
 
+func resolveGitSubmoduleURL(superprojectURL, submoduleURL string) (string, error) {
+	if !isRelativeSubmoduleURL(submoduleURL) {
+		return submoduleURL, nil
+	}
+
+	if hasSchemePattern.MatchString(superprojectURL) {
+		base, err := url.Parse(superprojectURL)
+		if err != nil {
+			return "", err
+		}
+		base.Path = path.Join(base.Path, submoduleURL)
+		return base.String(), nil
+	}
+
+	if scpLikeURLPattern.MatchString(superprojectURL) {
+		matched := scpLikeURLPattern.FindStringSubmatch(superprojectURL)
+		user := matched[1]
+		host := matched[2]
+		repoPath := matched[3]
+		return fmt.Sprintf("%s%s:%s", user, host, path.Join(repoPath, submoduleURL)), nil
+	}
+
+	return filepath.Join(superprojectURL, filepath.FromSlash(submoduleURL)), nil
+}
+
+func isRelativeSubmoduleURL(repository string) bool {
+	return strings.HasPrefix(repository, "./") || strings.HasPrefix(repository, "../")
+}
+
 func gitRevParseInWorkingDirectory(ctx context.Context, sh *shell.Shell, workingDirectory string, extraRevParseArgs ...string) (string, error) {
 	gitDirectory := filepath.Join(workingDirectory, ".git")
 
@@ -459,8 +489,10 @@ func gitRevParseInWorkingDirectory(ctx context.Context, sh *shell.Shell, working
 }
 
 var (
-	hasSchemePattern  = regexp.MustCompile("^[^:]+://")
-	scpLikeURLPattern = regexp.MustCompile("^([^@]+@)?([^:]{2,}):/?(.+)$")
+	hasSchemePattern = regexp.MustCompile("^[^:]+://")
+	// Require at least two non-colon characters before ":" so Windows drive
+	// paths like C:\repo are handled as local filesystem paths, not scp-like URLs.
+	scpLikeURLPattern = regexp.MustCompile("^([^@]+@)?([^:]{2,}):(.+)$")
 )
 
 // parseGittableURL parses and converts a git repository url into a url.URL
@@ -470,8 +502,12 @@ func parseGittableURL(ref string) (*url.URL, error) {
 			matched := scpLikeURLPattern.FindStringSubmatch(ref)
 			user := matched[1]
 			host := matched[2]
-			path := matched[3]
-			ref = fmt.Sprintf("ssh://%s%s/%s", user, host, path)
+			repoPath := matched[3]
+			if strings.HasPrefix(repoPath, "/") {
+				ref = fmt.Sprintf("ssh://%s%s%s", user, host, repoPath)
+			} else {
+				ref = fmt.Sprintf("ssh://%s%s/%s", user, host, repoPath)
+			}
 		} else {
 			normalizedRef := strings.ReplaceAll(ref, "\\", "/")
 			ref = fmt.Sprintf("file:///%s", strings.TrimPrefix(normalizedRef, "/"))
--- a/internal/job/git.go
+++ b/internal/job/git.go
@@ -8,7 +8,6 @@ import (
 	"net"
 	"net/url"
 	"os/exec"
-	"path"
 	"path/filepath"
 	"regexp"
 	"slices"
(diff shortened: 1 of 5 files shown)


swebb:
Thanks for opening this PR. I’ve created a ticket to track it internally.

internal/job/git.go
```diff
@@ -449,6 +450,35 @@ func gitEnumerateSubmoduleURLs(ctx context.Context, sh *shell.Shell) ([]string,
 	return urls, nil
 }
 
+func resolveGitSubmoduleURL(superprojectURL, submoduleURL string) (string, error) {
```
buildkate, line 452: Thanks for the thorough fix and fallback handling. I’d like to proceed with this approach (instead of #4276) because it preserves mirror the acceleration for relative submodules.

I had a look into how git does the resolution and have one concern: Git resolves these URLs lexically using [`relative_url()` in `remote.c`](https://github.com/git/git/blob/master/remote.c#L3026-L3095), whereas this implementation uses `net/url` and `path.Join`. This differs for some valid Git URLs, for example:

```text
Superproject: https://host/team%2Frepos/main.git
Submodule:    ../child

Git resolves:
https://host/team%2Frepos/child

Current Go resolver produces:
https://host/team/repos/child
```

Could we adjust `resolveGitSubmoduleURL` to follow Git’s behavior? Git’s relevant compatibility cases are in [`t/t0060-path-utils.sh`](https://github.com/git/git/blob/master/t/t0060-path-utils.sh#L390-L436) for reference.

The existing mirror/fallback design and integration tests otherwise look good.

davido: Thanks, good catch. I updated `resolveGitSubmoduleURL` to avoid
`net/url` and `path.Join` and instead resolve relative submodule URLs
lexically like Git's `relative_url()` in `remote.c`.

That preserves escaped path bytes, so the example now resolves as Git
does:

`https://host/team%2Frepos/main.git` + `../child`
=> `https://host/team%2Frepos/child`

I also added unit coverage for the Git core compatibility vectors from
`t/t0060-path-utils.sh`, plus a focused test for the `%2F` case.

buildkate (approved):

davido:
@buildkate 

Thanks for merging #4287.

Could this be included in the next Buildkite agent release?

This is blocking Gerrit's Bazel downstream Buildkite pipeline:
bazelbuild/continuous-integration#2815

Bazel CI consumes the agent from released packages/artifacts when rebuilding
worker images, so the downstream image refresh can only pick up the fix once it
is available in a tagged release.

PR #4294, opened 2026-08-31
Avoid retaining hook file handles during type detection

scadu:
## Description

Hook type detection opens the hook without closing it, which can prevent Git from replacing that file on Windows.

Remove the redundant `open` - the binary and shebang checks already manage their own handles, and add a regression test.

## Context

SUP-7980

## Changes

- Remove the redundant `os.Open` from hook type detection.
- Add a regression test confirming the hook file can be removed afterwards

## Testing

Reproduced on Windows with the released agent v4.0.0 (failed job 01a05c06-1544-4f90-b228-bdf7254d0011). Retrying with this branch's binary allowed the hook to replace itself and pass (job 01a05c0c-11f0-4da4-a84c-96f5b3b7babd).

## Disclosures / Credits

Lavazza ☕


moskyb (approved):
love it

PR #4309, opened 2026-09-01
Clarify OIDC status 1 retry behavior

karensawrey:
Human comment: This is an AI-generated PR that is a result of reviewing https://github.com/buildkite/docs-private/pull/2057.
Feel free to close but looks like a valid little nitpick from the machine.

## Description

Status 1 from `oidc request-token` does not always mean the command exhausted its retries. Non-retryable 4xx responses such as 408, 421, and 425 can stop after the first attempt.

Remove the inaccurate retry-exhaustion claim while preserving the guidance that retrying may succeed.

## Context

Follow-up to buildkite/docs-private#2057 and buildkite/agent#4272.

## Testing

`go test ./clicommand`

## Disclosures / Credits

Amp identified the source wording, made the one-line edit, and ran the targeted test under Karen Sawrey's direction.

### DIFF
--- a/clicommand/oidc_request_token.go
+++ b/clicommand/oidc_request_token.go
@@ -75,7 +75,7 @@ Exit statuses:
   410, or 422 response, such as a disallowed audience). Retrying will not
   succeed.
 - 1: any other failure, including transient API or network errors (timeouts,
-  5xx, other 4xx) after exhausting retries. Retrying may succeed.`
+  5xx, other 4xx). Retrying may succeed.`
 )
 
 var OIDCRequestTokenCommand = &cli.Command{


clicommand/oidc_request_token.go
```diff
@@ -75,7 +75,7 @@ Exit statuses:
   410, or 422 response, such as a disallowed audience). Retrying will not
   succeed.
 - 1: any other failure, including transient API or network errors (timeouts,
-  5xx, other 4xx) after exhausting retries. Retrying may succeed.`
+  5xx, other 4xx). Retrying may succeed.`
```
moskyb, line 78: i think we can still say it exhausted its retries if we only tried once. i think this docstring is clearer as is

karensawrey: Makes sense. Thanks for taking a look, @moskyb!

karensawrey:
A single retry is still a retry. Closing.

PR #4321, opened 2026-09-03
Skip blob retention refresh on a fallback restore

ss1909:
## Description

The DynamoDB cache-entry metadata already skips its TTL bump on a fallback restore (a request whose exact key doesn't exist, served by an older, looser-matching entry instead) -- `unless entry_result.fallback_used?` in the backend. The reasoning: a fallback hit means the entry was saved under a shorter key sequence, and refreshing it would reset the clock on an entry the caller didn't explicitly target.

The two blob stores didn't follow the same rule. `NscStore.Download` ran `nsc artifact extend --ensure_minimum 72h` on every download, and `S3Blob.Download` self-`CopyObject`'d to reset `LastModified` (the field S3 lifecycle rules key off) on every download -- both unconditionally, regardless of exact-vs-fallback. `restore.go` already reads `retrieveResp.Fallback` right where it calls into the blob store; it just never got passed down, since `Blob.Download`'s signature had no fallback parameter.

Alternative considered: none -- this mirrors an already-decided policy (metadata already does this), so the fix is just applying the same rule to the two places that didn't.

## Context

https://linear.app/buildkite/issue/A-1777/align-fallback-restore-expiry-refresh-for-cache-metadata-and-blobs

## Changes

- `internal/cache/store/blob.go` -- `Blob.Download` gains a `fallback bool` parameter.
- `internal/cache/store/nsc.go` -- skips the TTL-extend call on a fallback download.
- `internal/cache/store/s3.go` -- skips the self-`CopyObject` refresh on a fallback download. Extracted the copy into `refreshObjectExpiry`, parameterized on a new `objectCopier` interface mirroring the existing `objectDownloader`/`downloadWithRetry` seam in the same file, so the refresh (and the fallback gate) is directly unit-testable.
- `internal/cache/store/file.go` -- `LocalFileBlob.Download` accepts and ignores the parameter (no retention concept for local files; still required for interface conformance).
- `internal/cache/restore.go` -- threads `retrieveResp.Fallback` into the one production `Download` call site.
- Test coverage: `TestNscStore_SkipsRefreshOnFallbackDownload`, `TestRefreshObjectExpiry`'s fallback case, plus existing call sites updated to preserve exact-match behavior.

## Testing

## Disclosures / Credits

Claude Code implemented this change end-to-end (TDD: failing test first, then the fix), under my direction and review, including a couple of follow-up simplification passes after I flagged unnecessary indirection.

### DIFF
--- a/internal/cache/restore.go
+++ b/internal/cache/restore.go
@@ -440,7 +440,7 @@ func (c *client) downloadCache(ctx context.Context, retrieveResp api.CacheEntryR
 	archiveFile = filepath.Join(tmpDir, storeObjectName)
 
 	// Download archive
-	transferInfo, err = blobStore.Download(ctx, storeObjectName, archiveFile, retrieveResp.Fallback)
+	transferInfo, err = blobStore.Download(ctx, storeObjectName, archiveFile)
 	if err != nil {
 		// Clean up temporary directory on failure
 		_ = os.RemoveAll(tmpDir)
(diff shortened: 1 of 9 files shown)


jamiemonserrate (commented):
So I think the PR makes sense - but I have a design qualm. `Blob.Download` taking a `fallback` bool doesn't make sense. Its muddling unrelated concepts. 

Could we have the restore flow decide whether to call something like `RefreshRetention` after a successful download? I think that keeps the responsibility clearer and avoids teaching each blob store what a fallback means.

ss1909:
thanks @jamiemonserrate , that makes sense especially since LocalFileBlob had not need for the fallback arg. 

I've made the following changes:

1. abstracted it away in a RetentionRefresher interface
2. RetentionRefresher is now implemented only by NscStore/S3Blob
3. restore.go type-asserts for it and decides whether to call it after a successful download
4. LocalFileBlob doesn't have to know anything about this anymore

jamiemonserrate (approved):
LGTM 🚀

PR #4322, opened 2026-09-03
Run windows tests on hosted compute + macOS tests in CI

moskyb:
## Description

We recently added support for windows hosted agents as an enterprise-only feature. it'd be good to dogfood this internally, and have our own builds be tied up in it.

Additionally, macOS tests have been running on developer laptops by proxy for a long time. While we're here, add steps running tests on macos, also against a hosted queue.

## Context

A-1864

## Changes

- Change windows tests to run on a windows hosted queue
- Add macOS tests, running on a macOS hosted queue
- Add `pipeline.yml` to all `if_changed` blocks, ensuring that all steps are run if someone modifies the pipeline definition

## Testing

## Disclosures / Credits

C'est moi


CerealBoy (approved):
Love it, LGTM 🚀 🚀

PR #4339, opened 2026-09-07
Honor the server's configured cache retention for backing blobs

ss1909:
## Summary

Makes the agent honor the cache registry's **configured retention** when it stamps a backing blob's lifetime, instead of a hard-coded `72h`.

Companion to the backend work in buildkite/buildkite#33814 (A-1831, cache retention configurable via a service quota). The backend now returns `retention_seconds` on the cache **store** (upload) and **retrieve** responses; this PR consumes it.

## Why

`NscStore` set the Namespace artifact lifetime to `nscDefaultExpiry = "72h"` on both upload (`--expires_in`) and refresh (`--ensure_minimum`). If an org's registry TTL is raised above 72h, the registry metadata outlives its backing blob: a later exact hit returns from the registry, the blob download 404s, and it collapses to a (self-invalidating) miss. So raising the quota bought hosted customers nothing. This threads the real retention to the blob so the two clocks stay aligned.

## Changes

- `api/cache.go`: add `RetentionSeconds` to `CacheEntryCreateResp` and `CacheEntryRetrieveResp`.
- `store.Blob.Upload` and `store.RetentionRefresher.RefreshRetention` take a `retention time.Duration`.
- `NscStore` uses it for `--expires_in` (upload) and `--ensure_minimum` (refresh) via a new `nscExpiry` helper that rounds up to whole hours and **falls back to `72h` when the server sends nothing** (older server) — backward compatible in both directions.
- `S3Blob` / `LocalFileBlob` accept the duration but ignore it (S3 lifetime is governed by the bucket lifecycle policy; local has no TTL).
- `save.go` passes the store response's retention on upload; `restore.go` passes the retrieve response's retention on the exact-hit refresh (fallback hits still skip the refresh).

## Verification

- `go build ./...`, `go vet ./internal/cache/...`, `go test ./internal/cache/...` all pass.
- New `TestNscExpiry` (fallback, whole-hours, multi-day, round-up); `nsc_test` asserts the formatted `--expires_in` / `--ensure_minimum`; `restore_test` asserts the retention propagates through `maybeRefreshRetention`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


blaknite:
@buildsworth-bk review and approve if L2

PR #4348, opened 2026-09-08
Avoid partial clones for Kubernetes sparse checkouts

### LINKED ISSUE #4199: Sparse checkout in agent-stack-k8s can break git in command containers (esp. with mirrors)
## Describe the bug
On agent ≥3.133.0, setting BUILDKITE_GIT_SPARSE_CHECKOUT_PATHS auto-injects `--sparse` and `--filter=blob:none` on clone/fetch (https://github.com/buildkite/agent/pull/4044). That turns the workspace into a promisor partial clone.

With agent-stack-k8s, only the checkout container has git credentials. The command container typically does not, so any later git command that needs a missing object may try to lazy-fetch from origin and then auth-fail or hang.

This is especially problematic with git mirrors: the mirror already avoids most blob network transfer, so `blob:none` adds little benefit but still enables promisor/lazy-fetch in the uncredentialed command container.

## Steps To Reproduce
- Run a job on agent-stack-k8s with a git mirror (--reference) and BUILDKITE_GIT_SPARSE_CHECKOUT_PATHS set.
- Confirm checkout logs show `--sparse --filter=blob:none`.
- In the command container (no git credentials), run a git command that touches an object not already local like `git cat-file -e <sha-not-in-checkout>^{commit}`, or any history/blob access outside what checkout materialized.

## Expected behavior
When using agent-stack-k8s with sparse checkout (and likely also whenever git mirrors are enabled), the agent should not auto-inject `--filter=blob:none` (or otherwise leave promisor lazy-fetch enabled for the command container). Missing-object checks should fail locally so callers can fetch explicitly from a credentialed context if needed.

## Actual behaviour
Checkout produces a partial clone. Command-container git can attempt promisor fetches without credentials, causing a hang.

## Stack parameters (please complete the following information)
 - Version v3.134.0
 - Orchestrator: agent-stack-k8s (separate checkout vs command containers)

## Additional context
Potential workaround is to put a `--filter` in BUILDKITE_GIT_CLONE_FLAGS/BUILDKITE_GIT_FETCH_FLAGS to suppress the behaviour from https://github.com/buildkite/agent/pull/4044, or use GIT_NO_LAZY_FETCH=1 in the command container

buildkate:
## Description

Sparse checkout currently makes every fresh checkout a blobless partial clone. In Agent Stack for Kubernetes, checkout and command phases run in separate containers, and Git credentials are normally available only during checkout. A later command that needs an omitted object can therefore attempt an unauthenticated lazy fetch and fail or hang.

When `BUILDKITE_KUBERNETES_EXEC=true`, this stops automatically adding `--filter=blob:none` to clone and fetch while retaining `--sparse`. Non-Kubernetes checkouts keep the existing optimization, and an explicitly supplied filter remains authoritative in Kubernetes.

## Context

Fixes https://github.com/buildkite/agent/issues/4199

https://linear.app/buildkite/issue/A-1667/sparse-checkout-in-agent-stack-k8s-can-break-git-in-command-containers

## Changes

- avoid automatic partial clones for Kubernetes sparse checkouts
- cover ordinary, explicit-filter, and mirror-backed Kubernetes checkouts
- update the remote mirror design notes for the Kubernetes exception

Existing persisted partial-clone worktrees are unchanged; Agent Stack's default workspace is fresh per job.

### Public documentation
- [ ] Ready for public documentation - document and publish
- [ ] Not ready for public docs - document and hold
- [x] Not applicable - docs not needed

## Testing

## Disclosures / Credits

Amp helped research, implement, review, and test this change.

### DIFF
--- a/docs/remote-git-mirrors.md
+++ b/docs/remote-git-mirrors.md
@@ -20,6 +20,12 @@ happens to share the word "mirror". This document uses:
 | **on-host mirror** | `--git-mirrors-path/<dir>` — the local bare clone shared between jobs on a host |
 | **checkout** | the working directory the job runs in |
 
+Throughout this document, statements that sparse checkout automatically adds
+`--filter=blob:none` apply only outside Kubernetes execution. Kubernetes
+checkout still uses `--sparse`, but does not add a filter unless the user
+supplies one, because later command containers may not have the credentials
+needed for lazy fetches.
+
 §§1–5 and §10 are durable: they describe the feature and the decisions behind
 it. §§6–9 and §11 are delivery-time content — once the stack has landed, fold
 anything still true into the durable sections and delete the rest rather than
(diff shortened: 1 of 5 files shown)


buildkate:
> I didn't find a code issue in this pass. The agent Buildkite check is currently failing, and this changes default Git checkout behavior for Kubernetes jobs, so I'm leaving approval to a human reviewer.

This does **not** change default git checkout behaviour:

- Normal Kubernetes checkout is **unchanged**.
- Sparse checkout is opt-in.
- Only Kubernetes jobs opting into sparse checkout—and not supplying an explicit filter—change behavior.

swebb (approved):
👍

PR #4354, opened 2026-09-10
Bound cache retrieval retry duration

ss1909:
## Description

A cache retrieval currently allows each of its five retry attempts to consume the server's full request timeout. During DynamoDB throttling, those nested retry budgets turned one retrieval into several minutes of waiting.

Give the complete cache-entry retrieval retry loop one two-minute deadline. This leaves flexibility for existing backend tail latency while preventing five slow requests from multiplying the operation to roughly five minutes. Fast retryable responses can still retry within that shared budget. Existing fatal failure semantics remain unchanged.

## Context

- [A-1905: Reddit Buildkite cache timeouts/throttling](https://linear.app/buildkite/issue/A-1905/3-reddit-buildkite-cache-timeoutsthrottling)
- Companion server PR: https://github.com/buildkite/buildkite/pull/34080
- E2E build 2578 showed a successful cache retrieval taking 12.99 seconds and other cache metadata operations taking 25–26 seconds, so the original 15-second deadline was below current legitimate backend latency.

## Changes

- Set a two-minute cache retrieval timeout on constructed clients.
- Share one child context across the Roko retry loop and every retrieval HTTP request.
- Distinguish parent cancellation, retrieval deadline exhaustion, and other failures in tracing.
- Test that a blocked request gets one attempt and that quick retryable 503s can still succeed within the deadline.

## Public documentation
**Status:** not needed

## Testing

Targeted checks passed with `go test ./api ./internal/cache`, `go test -race ./internal/cache`, and `golangci-lint run ./internal/cache/...`. Full `go test ./...` could not run to completion in this checkout because packages imported as `github.com/buildkite/agent/v4/agent` and `github.com/buildkite/agent/v4/agent/plugin` are absent; CI should provide the authoritative full-suite result.

Buildkite

## Disclosures / Credits

Amp assisted with investigation, implementation, tests, and PR preparation under human direction.

### DIFF
--- a/internal/cache/client.go
+++ b/internal/cache/client.go
@@ -52,14 +52,15 @@ var (
 // bucket and the expanded, validated cache definitions used by every call.
 // Safe for concurrent use; honours context cancellation.
 type client struct {
-	api        cacheAPI
-	bucketURL  string
-	format     string
-	platform   string
-	registry   string
-	force      bool
-	caches     []configuration.Cache
-	onProgress ProgressCallback
+	api             cacheAPI
+	bucketURL       string
+	format          string
+	platform        string
+	registry        string
+	force           bool
+	caches          []configuration.Cache
+	onProgress      ProgressCallback
+	retrieveTimeout time.Duration
 }
 
 // newClient builds a client from apiClient and cfg: loads and expands the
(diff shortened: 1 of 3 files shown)


ss1909:
Existing retries are sufficient; deferring an overall retrieval deadline, accepting up to ~5 minutes of delay during sustained timeouts.

PR #4358, opened 2026-09-11
Avoid blob filtering sparse reference-mirror checkouts

maschwenk:
## Description

Avoid automatically adding `--filter=blob:none` to sparse clones and fetches that use an on-host mirror in `reference` mode.

A reference clone already reads objects from the mirror through Git alternates. Combining the automatic blobless filter with `--reference` can instead force expensive local object processing without reducing the transfer from the canonical repository.

This keeps the existing behavior for remote mirrors, checkouts without a mirror, and `dissociate` mode. User-supplied clone and fetch filters are still preserved.

## Context

We observed this on a large monorepo with agent v3.137.0. A sparse reference clone transferred about 596 KiB, then locally enumerated 1,500,947 objects and compressed 353,810 objects. The checkout took 11m45s. A full reference-mirror checkout on the same pipeline and agent fleet completed in about 21s.

## Changes

- Suppress the agent-added blobless filter for sparse on-host reference-mirror clones and fetches.
- Add an integration regression that exercises a real sparse reference checkout and verifies that it is not configured as a promisor clone.

## Public documentation

**Status:** not needed

## Testing

- [x] Tests have run locally. `go test ./internal/job/...` passes. `go test ./...` passed every package except one unrelated macOS temp-directory cleanup failure in `clicommand`; that individual failing test passed immediately when rerun.

Owner.com

## Disclosures / Credits

OpenAI Codex helped investigate the production checkout, implement the change, and write the regression test. I reviewed the diff and ran the tests locally.

### DIFF
--- a/internal/job/checkout.go
+++ b/internal/job/checkout.go
@@ -439,6 +439,7 @@ func (e *Executor) defaultCheckoutPhase(ctx context.Context, previousAttempts in
 
 	addBloblessFilter := sparse.active() &&
 		e.shell.Env.GetString("BUILDKITE_KUBERNETES_EXEC", "") != "true" &&
+		!e.usesOnHostReferenceMirror(mirrorDir) &&
 		!userSuppliedCloneFilter &&
 		!hasPartialFilterFlags(gitFetchFlags)
 	if err := e.fetchSource(ctx, addBloblessFilter, &attempt); err != nil {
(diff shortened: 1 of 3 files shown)


buildkate (approved):
Hi @maschwenk, thanks for the PR.  Approved.

maschwenk:
@buildkate how do I merge? or will you

PR #4371, opened 2026-09-14
Scrub container registration tokens before starting tini

blaknite:
## Description

The existing agent re-exec scrubs registration tokens from the agent process, but official Docker entrypoints pass the original environment and arguments to tini first. The parent process therefore retains token material after the agent has scrubbed itself.

This change moves token preparation ahead of tini. A hidden Linux launcher replaces plaintext registration-token inputs with an inherited pipe reference, then execs tini as PID 1. The agent consumes the pipe before registration, leaving inherited read descriptors empty. No temporary secret files are created.

## Context

[A-1868](https://linear.app/buildkite/issue/A-1868/cluster-agent-token-still-readable-via-proc1environ-on-official-agent)

## Changes

- Update the six Alpine and Ubuntu entrypoints to prepare token inputs before tini, retaining startup scripts, SSH configuration, signal forwarding, and orphan reaping.
- Preserve CLI-over-environment and duplicate-option precedence, root CLI options, positional boundaries, and help behavior using the existing CLI flag definitions.
- Consume FD inputs before the parent starts and close overridden token descriptors. Keep operator-managed file:// references persistent, including files generated by startup scripts.
- Add unit coverage and a 32-case real-entrypoint regression suite, with architecture-specific test executables wired into image CI.

Startup scripts now receive an indirect registration-token reference rather than plaintext. Operator-managed token files and config-file exposure remain outside this change; this is not general same-UID isolation. The sidecar image is unchanged.

## Public documentation

**Status:** hold

Document the startup-script compatibility change with the release.

## Testing

- [ ] Full local suite (`go test ./...`) or successful CI pipeline.
- [x] Code is formatted with gofumpt.

Completed locally:
- Full command-package tests, including a final rerun through the declared mise toolchain.
- Focused token race tests and Linux/arm64 token tests.
- All 32 container cases twice on a cached official Alpine/arm64 image, with the rebuilt agent and updated entrypoint mounted read-only. These exercise real registration and authenticated polling against a container-local fake API, with networking disabled and synthetic credentials only.
- Linux-targeted lint, shell syntax, and whitespace checks.
- Negative control: the old entrypoint fails the process-environment leak assertion with the same rebuilt agent.

The container cases cover environment/argv and file/FD inputs, precedence, parsing boundaries, failure/help paths, process and descriptor inspection, SSH setup, tini PID 1, orphan reaping, and graceful SIGTERM disconnect. They do not dispatch a build job. The full six-image/two-architecture matrix remains for CI.

Independent review identified three parsing/help regressions; each was reproduced, fixed, and covered before this draft.

## Rollback

Safe to revert operationally; reverting restores the previous token exposure. No data or configuration migration is required. Keep the image entrypoints paired with the matching agent binary when rolling back.

## Disclosures / Credits

OpenCode and its subagents implemented the change, wrote and ran the tests, and performed the independent code review under my direction. This draft is for my review.

@buildsworth-bk permission to approve if risk L2


blaknite:
Superseded by two scoped draft PRs: #4373 for the minimal environment-token fix, and #4374 for argv protection stacked on top. The original branch and implementation are preserved for comparison.

PR #4373, opened 2026-09-15
Hide container environment tokens before starting tini

blaknite:
## Description

Official Docker images pass BUILDKITE_AGENT_TOKEN through their startup processes to tini. The agent scrubs its own environment, but cannot remove the token from its parent. Builds sharing the container UID and PID namespace can recover it from /proc/1/environ.

The entrypoints now write the token to a pipe, wait for the writer to exit, and re-exec Bash with an fd:// reference before running startup scripts. The agent drains the pipe before command parsing and retains the value in memory. This also removes an unused environment token when a CLI token overrides it, leaving parent processes with an empty pipe.

## Context

[A-1868](https://linear.app/buildkite/issue/A-1868/cluster-agent-token-still-readable-via-proc1environ-on-official-agent)

## Changes

- Add the pipe handoff to all six Alpine and Ubuntu entrypoints, retaining tini as PID 1.
- Consume the handoff at agent startup, including for bootstrap, help, and error paths.
- Withhold the consumption marker from startup scripts so commands such as buildkite-agent --version do not drain the final agent’s pipe.
- Run environment-token regression tests in the image CI matrix.

## Compatibility

Plaintext environment tokens are limited to 4096 bytes so the writer can finish without blocking on a full pipe. This limit also applies when displaying help or overriding the token.

Startup scripts receive a single-use fd:// reference instead of plaintext. They must not consume or persist it. Operator-managed file:// and fd:// inputs keep their existing behavior. Raw --token arguments remain visible in tini’s command line.

The updated entrypoints must ship with the matching agent binary.

## Public documentation

**Status:** hold

Document the startup-script behavior and size limit with the release.

## Testing

- [ ] Full local suite (`go test ./...`) or successful CI pipeline.
- [x] Code is formatted with gofumpt.

All 21 container cases passed on Bash 5.0.17 and 5.3.9 using representative ARM64 fixtures, the rebuilt agent, and actual entrypoints. Tests use dummy credentials and a loopback API with networking disabled. They inspect process environments and inherited descriptors, and cover token overrides, startup scripts, bootstrap, file/FD inputs, help/errors, SSH setup, signals, and reaping.

The old agent fails the negative control because tini retains unread token bytes. Full clicommand tests and Go lint passed. Shell checks passed apart from two existing SC2155 warnings in build-binary.sh.

The full six-image/two-architecture CI matrix has not been verified.

## Rollback

Safe to revert operationally, but reverting restores environment-token exposure. Roll back the entrypoints and agent binary together. No data migrations.

## Disclosures / Credits

OpenCode and its subagents implemented the change, wrote and ran tests, and performed static review under my direction.

@buildsworth-bk permission to approve if risk L2


blaknite:
The argv-protection follow-up is #4374, stacked on this branch. This environment-only PR can be reviewed and merged independently; it does not add an argument scanner or production helper executable.

blaknite:
@buildsworth-bk review

PR #4379, opened 2026-09-18
Avoid SSH version warnings when SSH is not installed

buildkate:
## Description

**What**
Don’t show the warning when SSH isn’t installed. 

**Why**
Bootstrap probes the SSH version even in command-only containers. A container running `npm test` without SSH installed therefore logs “Failed to check SSH version”, despite not needing SSH.

## Context

Fixes [A-1928](https://linear.app/buildkite/issue/A-1928/avoid-ssh-version-warning-in-command-containers-that-do-not-need-ssh).

## Public documentation

**Status:** not needed

## Testing

- [x] Code is formatted (with `go tool gofumpt -extra -w` on the changed files).

`go test ./internal/job -count=1` passes. The missing-SSH regression test failed before the fix; coverage also checks strict host-key checking, old and modern SSH, and warnings for unknown versions and failed probes.

## Disclosures / Credits

Amp investigated the issue, implemented the fix, and wrote and ran the tests.


zhming0 (approved):
LGTM but I recommend asking the clanker to trace one step up. Why are we doing ssh for command phase? maybe there is a cleaner fix there. <br><br>but this change LGTM

buildkate:
> LGTM but I recommend asking the clanker to trace one step up. Why are we doing ssh for command phase? maybe there is a cleaner fix there. but this change LGTM

@zhming0 The warning emit happens during [bootstrap](https://github.com/buildkite/agent/blob/9643022ea69adb1b2fcf9f8b5603de176feddab3/internal/job/executor.go#L979), which runs before all phases. 	It needs to run `ssh -V` to determine which configs to apply based on the version. 

The issue is that some containers don't require ssh, but they will still see the warning.

PR #4399, opened 2026-09-22
Add experimental job error capture

hannahcancode:
## Description

When a job fails, its exit status tells us little about what went wrong. Plugins often know more—for example, whether Docker failed to pull an image or the container command failed—but currently that distinction only appears in the job log.

Adds `buildkite-agent job capture-error <code> --message <message> [--context '<json>']` so job commands, hooks and plugins can report those observed failures as structured errors. Requests go through the authenticated Local Job API; the parent assigns a timestamp and delivery key, then submits the error using the job token.

Gated behind the `capture-error` experiment. Payload redaction and delivery retries are not implemented yet; this is not ready for general use.

Local JSON requests and raw context input are capped at **32 KiB**, as agreed in the [size-limit discussion](https://buildkite-corp.slack.com/archives/C0BNKKZ1PJA/p1790051330103419). Oversized reports are rejected.

## Deferred

- **Payload redaction:** this slice uses fixed plugin messages rather than log excerpts. Redaction of customer-controlled messages and context will happen before customer use.
- **Delivery retries:** reporting currently makes one attempt. Bounded retries will use the idempotency support implemented in Rails.
- **Log capture:** we want to validate error reporting first. Log capture can be done in another slice.

## Public documentation

**Status:** not needed — experimental, not ready for public documentation.

## Testing

- [x] `go test ./...`
- [x] Changed Go files are formatted with `gofumpt`.

## Disclosures / Credits

Amp implemented the changes and tests.


internal/job/api.go
```diff
@@ -77,6 +96,24 @@ We'll continue to run your job, but you won't be able to use the Job API`)
 	}, nil
 }
 
+func (e *Executor) reportCapturedError(ctx context.Context, capturedError *jobapi.CapturedError) error {
+	apiClient := api.NewClient(logger.Discard, api.Config{
+		Endpoint:     e.shell.Env.GetString("BUILDKITE_AGENT_ENDPOINT", ""),
+		Token:        e.shell.Env.GetString("BUILDKITE_AGENT_ACCESS_TOKEN", ""),
+		DisableHTTP2: e.shell.Env.GetBool("BUILDKITE_NO_HTTP2", false),
+		UserAgent:    version.UserAgent(),
+	})
```
123sarahj123, line 109: the capture uses the default 60sec timeout because this creates an API client without a specific timeout (so it inherits the 60sec http timeout). The CLI will wait on this synchronously for that request. A stalled request could hold up the plugin. 

If we notice reporting having troubles, we can give it a shorter timeout so we aren't holding up the plugin to report the error

jobapi/captured_errors.go
```diff
@@ -0,0 +1,116 @@
...
+
+func (s *Server) handleCapturedError(w http.ResponseWriter, r *http.Request) {
+	r.Body = http.MaxBytesReader(w, r.Body, maxCapturedErrorBody)
+	payload := new(CapturedError)
+	dec := json.NewDecoder(r.Body)
+	dec.DisallowUnknownFields()
+	dec.UseNumber()
+	if err := dec.Decode(payload); err != nil {
+		s.writeCapturedError(w, fmt.Errorf("failed to decode request body: %w", err), http.StatusBadRequest)
+		return
```
123sarahj123, line 71: so our whole report will be rejected if the size is above 16 KiB

Rails accepts up to 8MiB then truncates to 4KiB and will replace the oversized content

if we have a 20KiB error it willnever reach rails even though rails could retain a shorter error. 

So we will reject some errors before rails gets the chance to truncate. Did we have a yarn about this? I can't remember so will have to remind me :)

123sarahj123: we set on line 17

> const maxCapturedErrorBody = 16 << 10

internal/job/api.go
123sarahj123: group session: should we make this a synchronous or async background thread?
Amp is on the loose

jobapi/captured_errors.go
123sarahj123: gosh, sorry Hannah for all the pings

For me to investigate

internal/job/api.go
hannahcancode: split this out into a follow up ticket

clicommand/job_capture_error.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+var JobCaptureErrorCommand = &cli.Command{
+	Name:        "capture-error",
+	Usage:       "Capture a structured error for the current job (experimental)",
+	Hidden:      true,
+	Description: jobCaptureErrorHelpDescription,
+	Flags: append(globalFlags(),
+		&cli.StringFlag{Name: "message", Usage: "A human-readable description of the error", Required: true},
+		&cli.StringFlag{Name: "context", Usage: "Additional error context as a JSON object"},
```
moskyb: these should also have env vars available, eg `BUILDKITE_AGENT_JOB_CAPTURE_ERROR_MESSAGE` etc

clicommand/job_capture_error.go
```diff
@@ -0,0 +1,76 @@
...
+Send a structured error to the running parent agent through the authenticated
+Local Job API. The parent validates and normalizes the error. There is no
+direct network fallback when the Local Job API is unavailable.
+
+Note: This feature is currently in development and subject to change. It is not
+yet available to all customers.`
+
+type JobCaptureErrorConfig struct {
+	GlobalConfig
+	Code    string `cli:"arg:0" label:"error code"`
```
moskyb: presuming the value of code is arbitrary, it might better be named `identifier` or something similar? that way there's not overlap with eg HTTP status codes or command exit codes,

moskyb: there may be product context i'm missing here, however

clicommand/job_capture_error.go
```diff
@@ -0,0 +1,76 @@
...
+		defer done()
+
+		if c.Args().Len() != 1 || strings.TrimSpace(cfg.Code) == "" {
+			return errors.New("exactly one error code is required")
+		}
+		if strings.TrimSpace(cfg.Message) == "" {
+			return errors.New("message must not be blank")
+		}
+		capturedError := jobapi.CapturedError{Code: cfg.Code, Message: cfg.Message}
+		if c.IsSet("context") {
```
moskyb: i _think_ this is probably going to be both clearer and do what you want it to:

```suggestion
		if cfg.Context != "" {
```

clicommand/job_capture_error.go
```diff
@@ -0,0 +1,82 @@
...
+
+const jobCaptureErrorHelpDescription = `Usage:
+
+    buildkite-agent job capture-error <code> --message <message> [--context '<json>']
+
+Description:
+
+Send a structured error to the running parent agent through the authenticated
+Local Job API. The parent validates and normalizes the error. There is no
+direct network fallback when the Local Job API is unavailable.
```
moskyb: This usage string is kind of focussed on the mechanics of what's going on here, but CLI help strings are for humans (and agents) to read, so it's best to focus on the what and the why rather than the how. maybe something like:

```
Reports a structured error on the current job, which will be shown in the Buildkite UI and API. Requires the local job API.
```

as an aside, i'm glad we're requiring the job API and not falling back without it. good call.

clicommand/job_capture_error.go
```diff
@@ -0,0 +1,82 @@
...
+			return errors.New("message must not be blank")
+		}
+		capturedError := jobapi.CapturedError{Code: cfg.Code, Message: cfg.Message}
+		if c.IsSet("context") {
+			var input io.Reader = strings.NewReader(cfg.Context)
+			if cfg.Context == "-" {
+				input = c.Root().Reader
+			}
+			dec := json.NewDecoder(input)
+			dec.UseNumber()
```
moskyb, line 89: is there any particular reason to use `UseNumber` here? if not, it's probably simpler to have this in one `Unmarshal` call, eg
```Go
err := json.Unmarshal([]byte(cfg.Context), &capturedError.Context)

File: clicommand/job_capture_error.go (hunk 4 above)
Comment (hannahcancode) [reply clicommand/job_capture_error.go]: ooh that's a reasonable call. It's copied `code` in the rails side, error has a code is the idea - cc @123sarahj123 @pinzonjulian thoughts?

File: clicommand/job_capture_error.go (hunk 7 above)
Comment (hannahcancode) [reply clicommand/job_capture_error.go:89]: I'm told by Amp that with `Unmarshal` - large integer values in context can silently lose precision. The decoder also supports the new stdin path. ? No strong opinion from me, shockingly.

File: clicommand/job_capture_error.go (hunk 6 above)
Comment (hannahcancode) [reply clicommand/job_capture_error.go]: Ah yeah good call 👍

File: clicommand/job_capture_error.go (hunk 4 above)
Comment (123sarahj123) [reply clicommand/job_capture_error.go]: from [PB-3158](https://linear.app/buildkite/issue/PB-3158/define-initial-job-error-codes-and-message-conventions)

> For each initial producer (Docker, Git checkout, agent lost), identify an observable fact, a stable code meaning and a useful message grounded in what that producer actually knows.
>
> * Treat codes as the machine-readable contract; messages provide human/agent diagnostic context, not a stable string for automation to parse.
> * Document how codes can grow without changing existing meanings, and decide whether unknown/private codes are allowed initially. Do not silently invent an allowlist policy.

and from the [tech spec](<https://www.notion.so/buildkite/Tech-Spec-Job-Errors-Slice-1-Emitting-Error-Events-f3bb8dbc2c8983c0842d81ae6efa854a?source=copy_link>)

```
create table job_errors (
  uuid              uuid primary key default uuid_generate_v7(),
  organization_uuid uuid not null,
  build_uuid        uuid not null,
  job_uuid          uuid not null,
  code              text not null,                       -- error code classification (e.g. ref_not_found)
  message           text not null,                       -- human readable error message (e.g. "error: failed to push some refs")
  timestamp         timestamp(6) not null,               -- when the producer says the error occurred.
  created_at        timestamp(6) not null default now(), -- authoritative persistence time
  context           jsonb,                               -- extra metadata on the error (e.g. target, operation, status etc.)
  idempotency_key   text                                 -- buildkite-agent-generated per emission; null for backend emissions
);
```

I don't remember specifically if we chose code with conviction, but agree it's an overloaded term. <br><br>maybe we update to `error_code`? But that also feels vague<br><br>`job_error_code`?

Comment (123sarahj123) [reply clicommand/job_capture_error.go]: thinking about this a bit more, I'm a bit torn on `code`. It describes the kind of error, like `network.timeout` rather than identifying one particular occurrence. `identifier` might sound more like the record’s uuid?<br><br>naming is hard

Comment (pinzonjulian) [reply clicommand/job_capture_error.go]: Hm that's an interesting perspective. The noun _Error Code_ is used in more contexts than HTTP (e.g. [stripe error codes](https://docs.stripe.com/error-codes) which uses strings, like we expect) and `exit codes` already have `exit` in them to disambiguate.

No product context @moskyb; just feels and taste. 

I personally prefer _error code_ with some docs to disambiguate. @hannahcancode @123sarahj123 any thoughts?

Comment (hannahcancode) [reply clicommand/job_capture_error.go]: updated to use `error code` - I think that matches how we've been talking about it?

File: jobapi/captured_errors.go (hunk 2 above)
Comment (hannahcancode) [reply jobapi/captured_errors.go:71]: _Amp, reporting back from the team's size-limit discussion._

The team unanimously agreed on **32 KiB (32,768 bytes) for the complete Local Job API request**, replacing 16 KiB. [Slack discussion](https://buildkite-corp.slack.com/archives/C0BNKKZ1PJA/p1790051330103419) · [research and decision](https://ampcode.com/threads/T-5f190464-4b42-4bb6-8a6f-f48c4a8aabad).

The original 16 KiB value was an Amp implementation choice without a documented sizing requirement. The current Compose plugin sends only a few hundred bytes for ordinary service names; 32 KiB allows more diagnostic detail and possible future log lines without accepting MiB-sized reports. It is a transport budget, not a promise to carry everything Rails could retain or a guarantee about aggregate memory usage.

The limit counts the whole local JSON body, including escaping, before the parent adds its timestamp/key. A 20 KiB message can now reach Rails, but a 32 KiB context plus its envelope cannot. Truncation stays exclusively in Rails: its 4 KiB message retention still matters even with this transport cap. We are not changing Rails' request or retention limits.

The local follow-up also bounds raw context/stdin before decoding, checks the complete serialized client request, and independently returns 413 for oversized direct socket requests. Boundary tests cover exactly 32 KiB, one byte over, trailing whitespace, JSON expansion and unchanged forwarding.

Implementation and tests are complete locally, with Sarah as author and committer for this commit only. **Not pushed yet**, respecting the earlier instruction; leaving this thread open until the follow-up is published.

File: internal/job/api.go (hunk 1 above)
Comment (123sarahj123) [reply internal/job/api.go:109]: ticket created
https://linear.app/buildkite/issue/PB-3390/bound-how-long-job-error-reporting-can-hold-up-a-plugin

File: jobapi/captured_errors.go (hunk 2 above)
Comment (hannahcancode) [reply jobapi/captured_errors.go:71]: _Amp update: pushed with Sarah’s approval._

Implemented in [b45c1a3](https://github.com/buildkite/agent/commit/b45c1a3d01be8196f07899b0a0425fa963fa7d83), with the clearer error in [5e13cf3](https://github.com/buildkite/agent/commit/5e13cf31852b9ce8fffd1dd34144d86875a79006):

- Raises the local whole-request limit to **32 KiB**.
- Bounds raw context/stdin before JSON decoding and checks the full encoded client request, including escaping.
- Independently checks direct socket requests before parsing; oversized bodies return **413: `captured error request exceeds 32768 bytes`**. Truncation remains in Rails.

The additional tests cover byte boundaries, whitespace, JSON expansion and unchanged forwarding. Relevant package tests, targeted race tests and lint passed. Both commits use Sarah’s author/committer identity without changing Git defaults; the PR description is now shortened to two sentences about the limit.

Comment (123sarahj123) [reply jobapi/captured_errors.go:71]: these are very large comments @hannahcancode

File: jobapi/captured_errors.go [hunk 8]
```diff
@@ -0,0 +1,128 @@
...
+	"io"
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/buildkite/agent/v4/internal/socket"
+)
+
+// MaxCapturedErrorBody is the local request limit in bytes, before the parent
+// adds a timestamp and idempotency key. Truncation remains the backend's job.
```
pinzonjulian: We don't need to poison future readers (the agents) here with the word `truncation`. that's a backend decision that the client shouldn't know of. we might even get rid of it

jobapi/captured_errors.go
123sarahj123: FYI to travellers, I told Hannah's amp (multiplayer) to leave some comments, but it left them in her name!

jobapi/captured_errors.go
hannahcancode: Amp: fixed in [b97154e](https://github.com/buildkite/agent/commit/b97154e7bc0c49f75922fd46ba5e37d71dcaacd3)—removed backend truncation-policy references from the comment, CLI help, tests and PR description. No behavior change.

pinzonjulian (approved):

clicommand/job_capture_error.go
```diff
@@ -0,0 +1,103 @@
...
+
+    buildkite-agent job capture-error <error-code> --message <message> [--context '<json>']
+
+Description:
+
+Reports a structured error on the current job for display in the Buildkite UI
+and API. Requires the Local Job API.
+
+The error code is a string identifying the kind of error, such as
+image_pull_failed, not an HTTP status or command exit code.
```
moskyb:
```suggestion
The error code is a string identifying the kind of error, such as
image_pull_failed, not (necessarily) an HTTP status or command exit code.
```

clicommand/job_capture_error.go
```diff
@@ -0,0 +1,103 @@
...
+
+Optionally include additional details as a JSON object using --context, or use
+--context - to read them from standard input.
+
+The complete JSON request and raw context input are each limited to 32 KiB.
+The request includes the code, message, context and JSON escaping. Oversized
+reports are rejected.
```
moskyb: The complete JSON being limited to 32KiB implies that the context is also limited to (slightly less than) 32KiB, right? On the other hand, the non-context part of the payload is going to be ~a couple hundred bytes max, so maybe it's worth only limiting the context.

Separately, if these limits are enforced at the serverside rather than the agent, it's probably worth not quoting a number at all — that way, if we want to raise or lower the limit, these descriptions don't become retroactive lies.

moskyb: > the non-context part of the payload is going to be ~a couple hundred bytes max

oh, i'm making an assumption here. do we have independent limits on the lengths of message and error code? if not, we should.

clicommand/job_capture_error.go
moskyb: yeah, now that we have stdin involved i think there's more reason to use a decoder, and who knows, maybe people need absolutely enormous numbers. makes sense to me.

clicommand/job_capture_error.go
```diff
@@ -0,0 +1,103 @@
...
+			}
+			dec := json.NewDecoder(bytes.NewReader(data))
+			dec.UseNumber()
+			if err := dec.Decode(&capturedError.Context); err != nil {
+				return fmt.Errorf("invalid context JSON: %w", err)
+			}
+			var extra any
+			if err := dec.Decode(&extra); !errors.Is(err, io.EOF) || capturedError.Context == nil {
+				return errors.New("invalid context JSON: context must contain exactly one object")
+			}
```
moskyb, line 96: ugh, my least favourite """"""feature""""" of `json.Decoder` (trailing values being accessible).

i think if we didn't do the decoder flush, we'd just silently ignore any trailing data, right? so do we need to do it at all?

clicommand/job_capture_error.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+
+		client, err := jobapi.NewDefaultClient(ctx)
+		if err != nil {
+			return fmt.Errorf("the Local Job API is required to safely capture diagnostics: %w", err)
+		}
```
moskyb: i would probably hoist this to the top of the function — if the job API is unavailable, there's nothing we can do, so we might as well avoid the work of consuming stdin, parsing json etc if we can.

internal/job/api_test.go
```diff
@@ -0,0 +1,207 @@
...
+	"github.com/buildkite/agent/v4/internal/replacer"
+	"github.com/buildkite/agent/v4/internal/shell"
+	"github.com/buildkite/agent/v4/internal/socket"
+	"github.com/buildkite/agent/v4/jobapi"
+)
+
+func TestCapturedErrorJobCancellation(t *testing.T) {
+	if !socket.Available() {
+		t.Skip("Local Job API unavailable")
+	}
```
moskyb: We shouldn't be skipping these tests if the job API is unavailable -- if it's unavailable, there's been some problem in our CI setup, which we need to fix

clicommand/job_capture_error.go
moskyb: after coming back to this and thinking about it a bit, i think my preferred limit structure would be something like:
```
code -> max 64 chars (or something)
message -> max 4KiB  (or something)
context -> max 32KiB (or something)
```
(with no limit on request size other than that implied by the combination of the member limits)

that way, everything has an intelligible limit, and if customers ever hit any of those limits, there's an easy thing for them to do — if they hit the context limit, they can feed in less context.

the issue with request body size limits is that we're imposing a limit on customers that they have only limited control over (because we add things to the payload), but if we provide specific, actionable numbers, they're always empowered to do the right thing.

hannahcancode: @123sarahj123 you have the most context here :)

clicommand/job_capture_error.go
hannahcancode: I think this is more a validation check - so at least we'll flag malformed scripts to the user? Could go either way though, if you'd prefer to rip it out.

internal/job/api_test.go
hannahcancode: 🤖 muttered something about windows but we've made the change and will see if the tests pass 🤞

moskyb: if our windows agents can't run the job API, that's something we need to fix. iirc though they should be on Server 2022, which has support.

clicommand/job_capture_error.go
moskyb: i'm not really fussed, follow your heart

clicommand/job_capture_error.go
pinzonjulian: The thing is @moskyb , until we actually instrument a few errors, we don't know exactly what'll be most helpful. 

For now, we're thinking that most of the useful stuff will be raw text in the `message` (i.e. put a handful of log lines that explain the error so you don't have to pull the whole log and consume LLM context). 

`context` is optional in our current view. It's akin to error tracking context which is just extra bits of data that you might want to have handy. 

Since we're not sure yet what'll be more helpful, I feel like leaving it open might be easier?. I do like enforcing the `code` to be a certain length.

123sarahj123: yep, fair point that separate limits would be easier to understand 🙂

when we had a dig yesterday, 32 KiB is one shared budget for the whole request, so context gets whatever is left after the message code and JSON encoding. It’s enforced in the CLI and Local Job API, rather than just Rails. The timestamp/key are added afterwards, so they don’t count towards it

code already has a 255-byte limit I’m wondering if we keep the shared budget for now while we instrument a few errors and see what’s useful like mentioned here [slack](https://buildkite-corp.slack.com/archives/C0BNKKZ1PJA/p1790136963500869?thread_ts=1790132167.159999&cid=C0BNKKZ1PJA)

This is just our first slice and an experiment so we can refine based on usage and I'll raise a follow up ticket to revisit this

Even with separate field limits, we’d still want a cap on the request before parsing it

with putting log lines in the message though rails still is still truncating that to 4 KiB so we won’t retain everything just because the agent accepts it

I can make the text clearer that we are just using the one shared limit. Since the agent enforces it, I’d keep the number in there

123sarahj123: I updated the comment. 

4a5ae6ba3b78cfa601ea0fd8275a3a7b9f05a452

123sarahj123: Follow up linear ticket:
[Review error capture size limits before customer use](https://linear.app/buildkite/issue/PB-3466/review-error-capture-size-limits-before-customer-use)

PR #4400, opened 2026-09-22
A-1869: Mask username-only Git tokens in logs and spans

zhming0:
## Description

Git credentials can occupy the username field of a URL, but `redact.URLCredentials` previously masked only passwords. Replace the entire userinfo with `xxxxx` so logs and checkout span attributes do not retain username tokens, including when a password is also present.

For example, the synthetic URL `https://test-token@example.com/repo.git` now becomes `https://xxxxx@example.com/repo.git`. Password-bearing URLs remain masked, with their usernames now hidden too. This also masks usernames in `ssh://` URLs; scp-style Git references and URLs without userinfo remain unchanged.

Debug environment output now applies the same masking to `BUILDKITE_REPO` as to `BUILDKITE_GIT_REMOTE_MIRROR_URL`. Existing log and span callers use the shared fix without changes.

## Context

[A-1869](https://linear.app/buildkite/issue/A-1869)

## Public documentation

**Status:** not needed — restores credential masking; no configuration changes.

## Testing

- [x] Tests have run locally (`go test ./...`).
- [x] Changed Go files formatted with `go tool gofumpt -extra -w`.
- Regression cases failed against the original helper and pass with the fix. Coverage includes username-only and encoded credentials, empty passwords, debug environment output, remote-mirror logs, and recorded OpenTelemetry attributes from a local Git checkout.
- Targeted regression tests pass with `-race`.
- `golangci-lint run ./internal/redact/... ./internal/job/...`: 0 issues.

All added credential examples are synthetic.

## Disclosures / Credits

Amp investigated the issue, implemented the fix and tests, and ran validation.


zhming0:
@buildsworth-bk review again. This is not spicy at all, it's a good security boost.

moskyb (approved):

